Issue #16, opened 2015-03-30
add support for "serial1" (TX only)


Links2004:
TX is working RX is disabled (only posible on GPIO 8 but we have no access to it)

in HardwareSerial.cpp in uart0_interrupt_handler add:
```CPP
    status = READ_PERI_REG(UART_INT_ST(1));
    if (status & UART_TXFIFO_EMPTY_INT_ST)
    {
        WRITE_PERI_REG(UART_INT_CLR(1), UART_TXFIFO_EMPTY_INT_CLR);
        Serial1._tx_empty_irq();
    }
```

HardwareSerial1.cpp
```CPP
/* 
  HardwareSerial1.cpp - esp8266 UART support

  Copyright (c) 2014 Ivan Grokhotkov. All rights reserved.
  This file is part of the esp8266 core for Arduino environment.
 
  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
  License as published by the Free Software Foundation; either
  version 2.1 of the License, or (at your option) any later version.

  This library is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  Lesser General Public License for more details.

  You should have received a copy of the GNU Lesser General Public
  License along with this library; if not, write to the Free Software
  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  
  Modified 30 March 2015 by Markus Sattler (esp8266 platform support for UART1)
*/


#include <stdlib.h>
#include <stdio.h>
#include <string.h>
#include <inttypes.h>
#include "Arduino.h"
#include "cbuf.h"

extern "C" {
#include "osapi.h"
#include "ets_sys.h"
#include "mem.h"
#include "uart_register.h"
#include "user_interface.h"
}

#include "HardwareSerial1.h"

HardwareSerial1 Serial1;

uart_t* uart1_init(int baud_rate);
void uart1_set_baudrate(uart_t* uart, int baud_rate);
int  uart1_get_baudrate(uart_t* uart);
void uart1_uninit(uart_t* uart);
void uart1_transmit(uart_t* uart, const char* buf, size_t size);    // may block on TX fifo
void uart1_wait_for_transmit(uart_t* uart);
void uart1_transmit_char(uart_t* uart, char c);  // does not block, but character will be lost if FIFO is full

[...]

PR #31, opened 2015-04-01
rewrite HardwareSerial code for UART0 + UART1 support in ESP8266

Links2004:
see issues #16


igrr:
Great, thanks!

hallard:
Hi guys,
Excellent, I was just looking into this for adding a 2nd UART, you done it; good. But I need a full 2nd UART. So do you think it would be possible to add the same for UART2 to have a full 2nd UART (GPIO13=RXD2 and TXD2=GPIO15) on ESP12 modules ?

hallard:
Forget my message, just saw that UART2 (declared on nodemcu boards) is a "shared" uart with UART0 so need a kind of multiplex soft to have both working at the same time.
Sorry about this, I'm learning this new chip since only few days.

crteensy:
Is there any example code available that demonstrates the use of Serial1? I'm using an ESP-01 module and can't seem to transmit data with it in my arduino sketch. However, during flashing, data seems to be received by the device connected to GPIO1 (Serial1 Tx).

Links2004:
```cpp
Serial1.begin(921600); // or any other Baudrate
Serial1.println("Test123");
```

PR #42, opened 2015-04-03
really remove libc and bugfix for connect issue

Links2004:
really remove libc.
replace all needed functions by own implementation.

use libm to get some functions like fpclassifyd

bugfix for connect issue:
http://www.esp8266.com/viewtopic.php?f=28&t=2251


hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -105,11 +110,14 @@ int ICACHE_FLASH_ATTR  WiFiClient::connect(IPAddress ip, uint16_t port)
             g_localPort = MIN_LOCAL_PORT;
         if (err == ERR_OK)
             break;
-        if (err == ERR_USE)
+        if (err == ERR_USE) {
+        	esp_yield();
```
igrr: Shouldn't we also call esp_schedule before calling esp_yield? Otherwise the sketch task will never be scheduled.

hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -40,10 +40,15 @@ extern "C"
 #include "include/ClientContext.h"
 #include "c_types.h"
 
+
+#define NO_PORT_BIND
```
igrr: What is the effect of NO_PORT_BIND? If we use it always for any reason, then just delete the whole #ifndef NO_PORT_BIND code path.

hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
Links2004: we are here in the sketch task. it is the connect function of the class.
some note the code is disabled currently (the NO_PORT_BIND define is set)

hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
Links2004: i was not sure if you need it for some reasons, so i not delete the code.
i remove it.

Issue #70, opened 2015-04-13
Some functions missing like strcasecmp, sprintf, dtostrf ?


mvdbro:
Using this function call:

strcasecmp

results in compiler error:

error: 'strcasecmp' was not declared in this scope

same issue for these functions:
sprintf
dtostrf
strtol
ltoa

chadouming:
have you imported stdio ? cause sprintf works fine here once stdio is imported.

mvdbro:
#include <stdio.,h> fixed the sprintf
but the others still have issues.
tried stdlib.h but didn't help.

rogerclarkmelbourne:
dtostrf is an AVR specific function.

Some files from the SAM core of the arduino project need to be added to the cores files

See https://github.com/arduino/Arduino/search?utf8=✓&q=dtostrf

Which shows which files need to be copied to the repo and how they get included

I'm not sure why normal c functions are missing e.g ltoa 

Did you try adding your own prototype for these?

Links2004:
the functions are implemented and working in the last git version.
to use the functions you need to so the include like this:
```cpp

extern "C" {

#include <stddef.h>
#include <stdarg.h>
#include <stdio.h>

}

```

rogerclarkmelbourne:
Can't this be added to one of the central includes e.g. Arduino.h ?

luc-github:
strtok neither strtok_r is found even adding 
```
extern "C" {

#include <stddef.h>
#include <stdarg.h>
#include <stdio.h>

}
```
I use the arduino-1.6.1-p1-windows.zip which a part this problem works well so far

mvdbro:
Some stuff is solved and found some workarounds for others.

But still can't use this one: error: 'strcasecmp' was not declared in this scope
strcmp is ok, strcasecmp is not ?
I guess these should be in the same library? Or it's just not implemented?

Links2004:
you shut update to the latest git version
only need to replace the ".\arduino-nightly\hardware\esp8266com\esp8266" directory.
with the directory from git:
https://github.com/esp8266/Arduino/tree/esp8266/hardware/esp8266com/esp8266

strtok_r is implemented:
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/cores/esp8266/libc_replacements.c#L162
and strcasecmp to:
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/cores/esp8266/libc_replacements.c#L204

[...]

PR #77, opened 2015-04-17
some new functions for ESP class, add includes to Arduino.h

Links2004:
may help by issue #70


Links2004:
yes for C++ its the same. 
Its came from the coding rules I used to,
I simply see the "missing" and add it.

PR #80, opened 2015-04-18
Add get and put functions to EEPROM

boybundit:
As available in http://www.arduino.cc/en/Reference/EEPROM
https://github.com/arduino/Arduino/tree/master/hardware/arduino/avr/libraries/EEPROM

EEPROM.get(address, data)
Read any data type or object from the EEPROM.

EEPROM.put(address, data)
Write any data type or object to the EEPROM.

Both functions return a reference to the object passed in.

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/EEPROM/EEPROM.h
+++ b/hardware/esp8266com/esp8266/libraries/EEPROM/EEPROM.h
@@ -37,7 +37,7 @@ class EEPROMClass
 
     template<typename T> T &get(int address, T &t)
     {
-        if (address < 0 || address >= _size)
+        if (address < 0 || address + sizeof(T) > _size)
             return t;
 
         uint8_t *ptr = (uint8_t*) &t;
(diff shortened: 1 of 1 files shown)


hardware/esp8266com/esp8266/libraries/EEPROM/EEPROM.h
```diff
@@ -35,6 +35,27 @@ class EEPROMClass
     void commit();
     void end();
 
+    template<typename T> T &get(int address, T &t)
+    {
+        if (address < 0 || address + sizeof(T) > _size)
+            return t;
+
+        uint8_t *ptr = (uint8_t*) &t;
+        for(int count = 0; count < sizeof(T); ++count) *ptr++ = _data[address + count];
```
igrr: can we just use ```memcpy(ptr, _data + address, count);``` here?

boybundit: Updated accordingly.

PR #83, opened 2015-04-19
Update ESP8266WiFi.cpp

hwwr:
Added SmartConfig func for esp8266, support ESPTOUCH or AIRKISS smartconfig protocol


sticilface:
Is / are there any apps for the iPhone that will work?  my search has come up empty for esptouch or airkiss, only CC3000 Ti one.

igrr:
I doubt Espressif went through the pain of submitting the app into the App Store. They do have a (paid!) app on Google Play: https://play.google.com/store/apps/details?id=com.espressif.iot

PR #85, opened 2015-04-19
Create SmartConfig

hwwr:
Use SmartConfig auto connect to AP, without manually input AP ssid and password.


Imroy:
These three changes should be done as one pull request.

igrr:
I've merged the changes into [smartconfig branch](https://github.com/esp8266/Arduino/tree/smartconfig), but more work needs to be done to fit the code into RAM. Right now (after adding -lsmartconfig) I get
```
section `.text' will not fit in region `iram1_0_seg'
```

Links2004:
may move ```libsmartconfig``` to ```.irom0.text``` like in #106 helps too here

PR #87, opened 2015-04-19
Fixed error: unknown type name 'size_t' and added support for ESPD-12 4 Mbyte flash model

einsteinx2:
Currently trying to adapt https://github.com/cnlohr/ws2812esp8266 for the Arduino IDE and ran into this compile error. Adding the definition to c_types.h fixed it.

Also, I added an option for the new ESPD-12 model to make the full 4 megabytes of flash available. Afaik all EDP-12 boards currently being sold are the new 4 megabyte models unless they're old stock. My supplier confirmed that the 512 kilobyte models are no longer available.

### DIFF
--- a/hardware/esp8266com/esp8266/boards.txt
+++ b/hardware/esp8266com/esp8266/boards.txt
@@ -1,5 +1,7 @@
 menu.UploadSpeed=Upload Speed
 menu.CpuFrequency=CPU Frequency
+
+
 ##############################################################
 generic.name=Generic ESP8266 Module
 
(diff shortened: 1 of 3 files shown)


igrr:
Thanks! Although I would rather include <stddef.h> where necessary, as it
has has size_t defined.

einsteinx2:
I had tried doing that, but at the time thought it was causing my "section `.text' will not fit in region `iram1_0_seg'" errors. If that wasn't the case, I have no problem switching back to using it and removing this line.

In any case, You may still be interested in merging the ESPD-12 board changes.

Links2004:
you  also need adapt the linker file for the new size.

see:
https://github.com/esp8266/Arduino/blob/esp8266/hardware/tools/esp8266/sdk/ld/eagle.app.v6.ld#L8

the iram1_0_seg means the ram segment for code is full but not the flash 
you need to use ```ICACHE_FLASH_ATTR```
like here 
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/cores/esp8266/libc_replacements.c#L282
to move functions from ram to flash

einsteinx2:
Thanks for the info! I'm still new to all this, so still figuring things out. I'll try and get everything working this week.

Links2004:
ok, is the ESPD-12 model identical to the boards sale as ESP-12?
http://www.esp8266.com/wiki/doku.php?id=esp8266-module-family&s[]=esp&s[]=12#esp-12

can you read the marking / naming on the 4MB chip?

einsteinx2:
It appears that all new ESP-12 modules are actually the 4mbyte ESPD-12 modules, but just not advertised as such. I checked one that I got from a random aliexpress seller about a month ago that was advertised as ESP-12 and its 4 mbyte flash. And I recently purchased 20 ESPD-12 advertised modules from a supplier in alibaba, and that supplier indicated that his supplier no longer has 512kbyte models, and that all were now 4mbyte.

I can't see the flash chip as its under the metal cover, but I checked in software and both modules reported 4 mbyte. I can post the code I used when I get home tonight.

einsteinx2:
Adding ICACHE_FLASH_ATTR to all of the functions resolved that build issue

Issue #95, opened 2015-04-22
**SOLVED** Has anyone else experienced "flakey" uploads?  (ESPtool.exe is broken) - read to bottom


tim-in-oakton:
My config is as follows:
- ESP8266 -03 modules
- Regulated 3.3v supply - arduino style battery input/USB regulator thing
- 3.3v FTDI USB:serial converter

On my first module I successfully compiled and uploaded maybe 500 times - occasionally I would need to upload twice if the first upload failed to execute correctly.  This got worse over time, it seemed like restarting the Arduino IDE and rebooting my PC was some help (hard to prove, it was intemittent).  I did try to reflash with a minimal sketch as well, that didn't really seem to help.
The program is a datalogging app for a Beer Keg - maybe 1000 lines of code or so.  Not very big.

Recently it got harder & harder to get a "good" upload (I used the serial monitor & wireless behavior to confirm if it was "working").  This morning I was completely unable to get an upload to work - the upload using the Arduino IDE would complete with no errors, but the uploaded executable image would not execute.

Using the Python ESPTool uploader, I was able to "erase" the flash, but not upload the 512K "blank" file - it would consistenly fail at 32% (hex800000).
I replaced the ESP module with another, the FIRST upload worked fine and executed perfectly, when I made a minor change to the code  however it would upload with no errors but not execute.  Frustrating.

The IDE reports:
Sketch uses 205,912 bytes (39%) of program storage space. Maximum is 524,288 bytes.
warning: unsupported baud rate: 115200, using 115200
Uploading 38112 bytes from C:\Users\timm\AppData\Local\Temp\build7944810091012342573.tmp/Kegerator-v4.3.cpp_00000.bin to flash at 0x00000000
......................................
Uploading 167848 bytes from C:\Users\timm\AppData\Local\Temp\build7944810091012342573.tmp/Kegerator-v4.3.cpp_40000.bin to flash at 0x00040000
....................................................................................................................................................................

I will try to upgrade the power supply to something I trust more than an off-brand Arduino-style regulator, but is this something that others have seen?  Are there any tricks to reliable upload&execute?

I am using the 1.6.3 IDE with a GIT pull from maybe 2 weeks ago - have there been any changes to the upload functionality or to the linker directives?

Any thoughts?  Not sure if I have a bug or a PEBCAK issue- are there any known "gotchas" to be aware of that complicate uploading?
thanks!
tim
##

tim-in-oakton:
[...]

Issue #78, opened 2015-04-17
Add Support for nodemcu


rei-vilo:
The [nodemcu](http://nodemcu.com/index_en.html) is an ESP8266-based 30-pin DIP form-factor board with USB connector and programmer. By default, it comes with Lua but I'd like to use the Wiring / Arduino framework.

On Mac OS X, it requires the installation of the driver [CH341SER_MAC_ZIP](http://www.wch.cn/download/CH341SER_MAC_ZIP.html) and then launch on Terminal `sudo nvram boot-args="kext-dev-mode=1"`.

Thanks!

gerardwr:
I use the Arduino environment from this github on my Mac. My ESP-01 is connected to my Mac via a PL2303 or FTDI board. For the USB-to-to-serial boards you will have to install the Mac OS drivers separately, this is not part of the supplied environment.

Or do I misunderstand?

rei-vilo:
We're tlaking about different products.

I'm using a FTDI board for a plain ESP8266-01 and everything goes fine. 

But the noceMCU board has its own programming chip, CH341, and comes pre-loaded with Lua.

sej7278:
I believe that the Arduino firmware uses/requires the AT firmware from the official SDK and will overwrite the nodemcu firmware.

Drivers are irrelevant, that's down to the user (and aren't even required on Linux).

If the nodemcu board cannot run non-nodemcu firmware then you should report that as a bug to them.

prasertsakd:
i use nodemcu  it different form esp8266 + usb2serial

onboard can auto reset i will ready  experiment on that

only config auto upload key on board.txt form none to ck it work

http://www.ayarafun.com/2015/04/arduino-ide-for-esp8266/

combs:
ch341 is a serial to USB adapter chipset, not a programmer, if that helps explain... no special support needed for it.

AllAboutEE:
@rei-vilo 

Flashing an Arduino sketch to the NodeMcu board does work, I think you are just getting confused because you think that flashing an Arduino sketch is the same thing as uploading a NodeMcu .lua program.

Uploading an Arduino sketch to the ESP8266 is actually equivalent to flashing a different version of the NodeMcu firmware, NOT uploading a .lua program.

So, to upload an Arduino sketch you need to use the "Flash" button on the board not the "User" button. Once you have uploaded the sketch then yes, use the "User" button to run it.

rei-vilo:
Thank you for the answers and pointers. 

[...]

PR #100, opened 2015-04-23
Add pyesptool as a programmer option

unreality:
I found that esptool consistently failed to upload on macos, this adds the python esptool as a programmer option

### DIFF
--- a/hardware/esp8266com/esp8266/platform.txt
+++ b/hardware/esp8266com/esp8266/platform.txt
@@ -93,7 +93,7 @@ tools.esptool.program.params.quiet=
 tools.esptool.program.pattern="{path}/{cmd}" -vvv {program.verbose} -cd {upload.resetmethod} -cb {upload.speed} -cp {serial.port} -ca 0x00000 -cf "{build.path}/{build.project_name}_00000.bin" -ca 0x40000 -cf "{build.path}/{build.project_name}_40000.bin"
 
 tools.pyesptool.cmd=esptool.py
-tools.pyesptool.cmd.windows=esptool.exe
+tools.pyesptool.cmd.windows=esptool.py
 tools.pyesptool.path={runtime.ide.path}/hardware/tools/esp8266
 tools.pyesptool.program.params.verbose=""
 tools.pyesptool.program.params.quiet="" 
--- a/hardware/esp8266com/esp8266/platform.txt
+++ b/hardware/esp8266com/esp8266/platform.txt
@@ -88,8 +88,14 @@ tools.esptool.cmd=esptool
 tools.esptool.cmd.windows=esptool.exe
 tools.esptool.path={runtime.ide.path}/hardware/tools/esp8266
 
-tools.esptool.upload.protocol=esp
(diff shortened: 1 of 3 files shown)


hardware/esp8266com/esp8266/platform.txt
```diff
@@ -88,8 +88,14 @@ tools.esptool.cmd=esptool
...
-tools.esptool.upload.protocol=esp
-tools.esptool.upload.params.verbose=-vv
-tools.esptool.upload.params.quiet=
-tools.esptool.upload.pattern="{path}/{cmd}" {upload.verbose} -cd {upload.resetmethod} -cb {upload.speed} -cp "{serial.port}" -ca 0x00000 -cf "{build.path}/{build.project_name}_00000.bin" -ca 0x40000 -cf "{build.path}/{build.project_name}_40000.bin"
+tools.esptool.program.params.verbose=-vv
+tools.esptool.program.params.quiet= 
+tools.esptool.program.pattern="{path}/{cmd}" -vvv {program.verbose} -cd {upload.resetmethod} -cb {upload.speed} -cp {serial.port} -ca 0x00000 -cf "{build.path}/{build.project_name}_00000.bin" -ca 0x40000 -cf "{build.path}/{build.project_name}_40000.bin"
+
+tools.pyesptool.cmd=esptool.py
+tools.pyesptool.cmd.windows=esptool.exe
```
igrr: windows version of command doesn't look right...

unreality:
Latest commits should work for windows if they have python installed and available in %PATH%, but I have not been able to test it. Should help with #129 , #95 , #138 #3

igrr:
I was looking at the option of packaging esptool.py along with python runtime using [pyinstaller](https://github.com/pyinstaller/pyinstaller). This would remove dependency on Python. Unfortunately, the binaries built using this method don't work on older versions of OS X (~10.6).

igrr:
Esptool.py needs to be updated to work around a bug in the SPIEraseArea ROM function: https://github.com/themadinventor/esptool/issues/33.

Links2004:
i build the fixed version for windows:
https://www.dropbox.com/s/jy9e25kcaceftjl/esptool.zip?dl=0

circuitFlower:
just an fyi, I started fresh with a new esp8266, and got it up and running on my Ubuntu VM. I think the problem was that I was trying to use this pyesptool solution? At one point it said it was erasing the flash, went through to 100%, then hung. I never got that esp working again.

rei-vilo:
From #78: On Mac OS X, esptools.py` requires `libusb` and works only with Python 2.7.

zerog2k:
I'm having a possibly related problem. I'm on ubuntu and esptool-ck will fail to upload, whereas esptool.py will always work fine..

I'm setting up a custom board (10k pullups on ch_pd & reset, 10k pulldown on gpio0).. I hooked up logic analyser, there are some notable differences, but sequence of reset/rts and gpio0/dtr seem to be acceptable for both.

So assuming this is for adding esptool.py for flashing, then +1 for this effort.

PR #107, opened 2015-04-24
missing strn program space wrappers

Makuna:


Makuna:
The whole selection of strn*_p wasn't supported.

PR #111, opened 2015-04-25
ClientContext.h: Remove delay and force output of queued data.

Chris--A:
This replaces a 5 second delay to an LWIP function `tcp_output()`. Rather than waiting for interrupts to handle sending, this will actively send the data.

I proposed this change on the esp8266 forum and it appears to solve a large latency issue: 
http://www.esp8266.com/viewtopic.php?f=28&t=2672

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -170,7 +170,7 @@ class ClientContext {
             _size_sent = will_send;
             DEBUGV(":wr\r\n");
             _send_waiting = true;
-            delay(5000); // max send timeout
+            tcp_output( _pcb );
             _send_waiting = false;
             DEBUGV(":ww\r\n");
             return will_send - _size_sent;


igrr:
The reason this removes the latency is because we are no longer waiting for tcp_sent callback to fire. So essentially with this change we don't wait for confirmation that the packet has actually been delivered.
If this is the desired behavior, it's better to switch to UDP which has lower latency due to not bothering about packet delivery.

igrr:
The execution flow looks like this:
- in the ```send``` method ```_size_sent``` is set to the amount of data that we are attempting to send
- the call to ```delay``` function causes a context switch — flow of execution is transferred to the scheduler
- tcp task gets scheduled and the ```tcp_output``` function is called
- when the packet is delivered, ```ClientContext::_sent``` method is called. It calculates the amount of data that is not yet delivered. If all the data has been delivered, it calls ```esp_schedule```, which causes a context switch back into the sketch (to the ```_send_waiting = false;``` line in ```ClientContext::send```).
- if the packet is not delivered, the 5000-ms timer armed by the ```delay()``` function fires which also causes a context switch back into the sketch.

Chris--A:
Thank you for the explanation. I'll have to do some further research and see where we can save some more time.

igrr:
using tcp_output() **and** delay() is one option that you may try.

Chris--A:
Yeah, I saw your post in the forum and was thinking the same thing.

PR #115, opened 2015-04-25
Fix windows build

Toshik:
Fix Windows build according to official Arduino build process
Now it is required to install JDK and point JAVA_HOME to it's location


Toshik:
Messed up, make new request...

PR #126, opened 2015-04-26
ESP8266WiFi: made connected() return true if data is available

bobvanderlinden:
I found an slight incompatibility with the Arduino documentation for `WiFiClient::connected()`: http://www.arduino.cc/en/Reference/WiFiClientConnected

`WiFiClient::connected()` should return true if the connection is disconnected, but there is still data available.
Before this patch `WiFiClient::connected()` returned true when the connection was still there. I found this more intuitive, but the documentation sadly said otherwise.


bobvanderlinden:
Thanks!

PR #135, opened 2015-04-27
Ability to wake from .deepSleep

linagee:
This was not immediately apparent to me. I don't think it would be apparent to anyone else just playing around with ESP8266/Arduino either. GPIO16 and RST (REST) need to be tied together or you can "never wake from .deepSleep"

It must be GPIO16 and it must be RST, no other GPIO pin will wake it from deepSleep.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ The implementation supports the entire Arduino SPI API including transactions, e
 
 APIs related to deep sleep and watchdog timer are available in the ```ESP``` object.
 
-```ESP.deepSleep(microseconds, mode)``` will put the chip into deep sleep. ```mode``` is one of ```WAKE_DEFAULT```, ```WAKE_RFCAL```, ```WAKE_NO_RFCAL```, ```WAKE_RF_DISABLED```.
+```ESP.deepSleep(microseconds, mode)``` will put the chip into deep sleep. ```mode``` is one of ```WAKE_DEFAULT```, ```WAKE_RFCAL```, ```WAKE_NO_RFCAL```, ```WAKE_RF_DISABLED```. (GPIO16 needs to be tied to RST to wake from deepSleep.)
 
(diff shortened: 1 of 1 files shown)


duncan-a:
Is this waking it by means of an internal timer, receipt of a WiFi command
or an external event from something like an RTC interrupt?

I need to do the RTC method in the near future, the WiFi method would mean
I could have one central RTC and use a broadcast to get data back from
outlying sensors (maybe) - not so sure about internal timers...

Presumably waking from deep sleep isn't possible with an ESP-01 as GPIO16
doesn't appear?Are they connected on the board?

I haven't investigated sleeping ESPs yet - I still have upload issues...

linagee:
ESP-01, get out some really tiny wires, connect right to the chip, and good luck. :-)

I was able to do this successfully with an ESP-12. When you deepSleep, the ESP8266 goes to sleep until an internal timer brings GPIO16 low. Tying this to RST allows it to reboot the ESP8266. (No state is stored, so, see my other request to bring forward the RTC storage, which is stored across a .deepSleep reboot. Another alternative: Store in EEPROM/SPI flash.)

I believe there are other sleeping ways to wake on WiFi/other GPIO interrupt, but they use about 3x as much power. (Also, I don't know what command does this from the ESP8266 SDK, it might not even be easily exposed.) If you want to use the .deepSleep method as a single external interrupt, you could probably .deepSleep (for nearly forever?) then bring RST low to "bring it out of deep sleep" / reboot.

duncan-a:
I'm in my 60s, my hands are no longer steady and I have cataracts - I can
barely solder 0.1"headers on stripboard now - not an option!!!

GPIO16 is brought out to the header pins on the Olimex module - that may be
helpful to someone...

Knocking Reset low to reboot was what I was thinking for an RTC Interrupt
(once I can understand how to eliminate the bootloader 'rubbish' on the
serial lines (I've read the posts, but I'm not there yet) and discovered
how to reduce the connect time.

Wake on WiFi would be good, but the power consumption (I've seen 48mA
quoted somewhere) would be a significant issue.

Thanks

PR #145, opened 2015-04-30
add functions to read out the flash config and flash id

Links2004:
Serial.print(F("Flash Chip Id: "));
Serial.println(ESP.getFlashChipId(), HEX);

Serial.print(F("Flash Chip Size: "));
Serial.println(ESP.getFlashChipSize());

Serial.print(F("Flash Chip Speed: "));
Serial.println(ESP.getFlashChipSpeed());

Serial.print(F("Flash Chip Mode: 0x"));
Serial.println(ESP.getFlashChipMode(), HEX);


igrr:
Ha, you beat me to it, I was just going to add the same :+1:

igrr:
I've removed the size/frequency units from Arduino.h — I have a feeling some people might actually use kB and kHz as variable names in their sketches.
Will add that back as user-defined literals once I switch the compiler to c++14 mode.

Links2004:
i don't known the "user-defined literal" feature nice thing.
it seams part of the c++11, c++14 only at some more predefined things.

its working well with the current compiler settings.

PR #149, opened 2015-04-30
add function to get the flash chip size based of the flash device id

Links2004:
Serial.print(F("Flash Chip Size By Id: "));
Serial.println(ESP.getFlashChipSizeByChipId());

- User-defined literals


Links2004:
move digitalWrite and digitalRead to ram
remove libg
fix 160Mhz mode

PR #173, opened 2015-05-04
add nodemcu board

onkelfunny:


onkelfunny:
the board supports 512k, 1M, 2M, 4M, 8M and 16M (depending on the ESP module)

PR #175, opened 2015-05-05
some readme changes

ficeto:
I did test the clock speeds for I2C at 160MHz clock as well. They are all the same.
Removed OUTPUT_OPEN_DRAIN as it's no longer in, and can please someone explain to me what exactly is an "output open drain"? It makes no sense to me a pin to be output and be open drain at the same time?
I added an explanation for the pinMode(x, SPECIAL) as well.
Also GPIO16 has a pulldown mode, I wander if I'm not missing a register def again for a pullup?


igrr:
Open drain output is an output mode where pin may be only forced low by the driver. It is normally used in combination with external pull-up resistors. I2C SDA should be open-drain, by the way.

ficeto:
:) that is something totally different :)
I2C is that way but that is not OUTPUT OPEN DRAIN.
it just has separate sections that deal with each mode. As soon as you enable OUTPUT (GPOE) on a pin, it becomes output. If you disable OE, it turns into input, and if you had the PULLUP/DOWN confed to that pin, it will pull itself.
Having OUTPUT_OPEN_DRAIN will not work with the rest of the digitalWrite/Read stuff.
what should be done here is:
```
digtalWrite(pin, LOW);
pinMode(pin, INPUT_PULLUP);
pinMode(pin, OUTPUT);
```
pinMode will switch between low and high (input pullup/output low)

ficeto:
that is exactly how the I2C is working now. I have the output set to 0, and the pullup enabled, so I'm just switching GPOE to switch between 0(OUTPUT_LOW) and 1(INPUT_PULLUP)

PR #183, opened 2015-05-06
Add definition for the Olimex MOD-WIFI-ESP8266(-DEV) boards

Imroy:


hardware/esp8266com/esp8266/boards.txt
```diff
@@ -57,6 +58,55 @@ generic.menu.FlashSize.4M=4M
...
+modwifi.upload.maximum_size=2097152
+modwifi.upload.wait_for_upload_port=true
+modwifi.serial.disableDTR=true
+modwifi.serial.disableRTS=true
+
+modwifi.build.mcu=esp8266
+modwifi.build.f_cpu=80000000L
+modwifi.build.board=MOD_WIFI_ESP8266
+modwifi.build.core=esp8266
+modwifi.build.variant=modwifi
```
igrr: Looks like pins_arduino.h for this variant is identical to ```generic```. Do we really need a variant for this module? It doesn't have pin numbers redefined, as far as i can tell.

hardware/esp8266com/esp8266/boards.txt
```diff
@@ -57,6 +58,55 @@ generic.menu.FlashSize.4M=4M
...
+modwifi.menu.UploadSpeed.230400.upload.speed=230400
+modwifi.menu.UploadSpeed.460800.linux=460800
+modwifi.menu.UploadSpeed.460800.macosx=460800
+modwifi.menu.UploadSpeed.460800.upload.speed=460800
+modwifi.menu.UploadSpeed.512000.windows=512000
+modwifi.menu.UploadSpeed.512000.upload.speed=512000
+modwifi.menu.UploadSpeed.921600=921600
+modwifi.menu.UploadSpeed.921600.upload.speed=921600
+
+modwifi.menu.FlashSize.2M=2M
```
igrr: If 2M is the only option for flash size, you can remove this. No need to have a menu with just a single option.

duncan-a: The Olimex modules that I have to hand are fitted with the 25Q16 *16MB*
Flash, straight from the box - maybe they're something of a variable feast
with Olimex?

Of the five Olimex modules that I have, two had clearly been reworked
around the ESP Chip and one was missing the capacitor that feeds the pcb
antenna - so couldn't possibly work.

Attempts to contact Olimex about the missing capacitor have received no
response.

Other than that, they're breadboard-friendly and have all the right pullups
and pulldowns already fitted (with 'solder-pad jumpers'). They don't have
the uFL antenna connector fitted but the pads are on the board - if you can
find the connectors and have the eyesight to solder them on!

hardware/esp8266com/esp8266/boards.txt
Imroy: I suppose not. The role of each pin is printed in the silkscreen, so there's no reason to assign "D0", etc. Just use the GPIO numbers directly.

igrr: Okay, then we can change this to ```modwifi.build.variant=generic``` and delete hardware/esp8266com/esp8266/variants/modwifi

PR #191, opened 2015-05-06
true interrupt enable and disable

Makuna:
This will expose the inherent cpu instruction cycle count, which is handy for bit bang timing.
Further, this fixes the interrupts() and noInterrupts() routines to honor original intent.

Due to the priority interrupt system on the chip, it supports the ability to suppress interrupts at different levels.  The levels are from 0 to 14, where most are on 0, but some of the WiFi is done on 14.

The current implementation of noInterrupts() doesn't block level 14 interrupts, causing routines that rely on not being interrupted to malfunction.  This change will honor the original intent of noInterrupts().


Makuna:
Missing a file!

PR #192, opened 2015-05-06
true noInterrupts() interrupts() support

Makuna:
This will expose the inherent cpu instruction cycle count, which is handy for bit bang timing.
Further, this fixes the interrupts() and noInterrupts() routines to honor original intent.

Due to the priority interrupt system on the chip, it supports the ability to suppress interrupts at different levels. The levels are from 0 to 14, where most are on 0, but some of the WiFi is done on 14.

The current implementation of noInterrupts() doesn't block level 14 interrupts, causing routines that rely on not being interrupted to malfunction. This change will honor the original intent of noInterrupts().

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/Arduino.h
+++ b/hardware/esp8266com/esp8266/cores/esp8266/Arduino.h
@@ -124,11 +124,17 @@ void timer1_write(uint32_t ticks); //maximum ticks 8388607
 void ets_intr_lock();
 void ets_intr_unlock();
 
-void xt_enable_interrupts();
-void xt_disable_interrupts();
+// level (0-15), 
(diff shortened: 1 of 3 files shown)


igrr:
Could it also be handy to suppress interrupts only up to certain level? (I.e. disable timers but keep WiFi running)

Makuna:
yes it would, the rsil requires the "level" to be a constant, and I have not been able to get __stringify to work with the current build to be able to do this.  I have tried __STRING, even pulled over the macros directly and I keep getting compile errors about macro expansion operator not found.

```
#define xt_disable_interrupts(state, level) __asm__ __volatile__("rsil %0,"__STRINGIFY(level)"; esync; isync; dsync":" = a" (state) : "memory")
```

PR #193, opened 2015-05-07
bool redeclaration error and warnings

ficeto:
they are either left over or something else changed and lead to build errors


igrr:
Yep, these were added as part of your latest pull, but not removed in cleanup commit.

ficeto:
I wander how I missed them last night...

Issue #166, opened 2015-05-03
Can't receive udp packets


anesthetized:
I'm working to implement a simple mDNS client (ask who's foobar.local and get response) and i'm stuck as it correctly sends packets, my pc with avahi-daemon responds accordingly (i'm checking with wireshark), but esp doesn't seem to receive any packet!

Wireshark screenshot proof:
![prompt2](https://cloud.githubusercontent.com/assets/10725286/7446590/620411c8-f1dd-11e4-94e5-5b4a1e452bda.png)

First line is esp's mDNS request, my pc answers on both unicast (second line) and multicast (third and fourth). Unicast response has 5353 src port and a random one as dst port. Multicast ports are both 5353.
It is not a unicast-or-multicast problem, since it seems to receive generic packets when i comment out the lookup() function.

Code (with multicast):
```Cpp
#include <ESP8266WiFi.h>
#include <WiFiUdp.h>
#include <Ticker.h>

Ticker nameLookup;

const char* ssid     = "ssid";
const char* password = "passwd";
const char* hostname = "hostname";

const unsigned int loc_port = 5353;
WiFiUDP udp;

void setup() {
  Serial.begin(115200);
  WiFi.begin(ssid, password);
  
  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
  }

  Serial.println("");
  Serial.println(WiFi.localIP());

  udp.begin(5353);
  nameLookup.attach(5, lookup);
}

void loop() {

  int packet = udp.parsePacket();
  if(packet!=0){
    Serial.println("gotcha!");
    delay(100);
  }

}

void lookup(){
    Serial.println("Hostname Lookup sent");
 
    char header[13] = { 0x00, 0x00, // ID
                        0x00, 0x00, // FLAGS
                        0x00, 0x01, // QDCOUNT
                        0x00, 0x00, // ANCOUNT
                        0x00, 0x00, // NSCOUNT
                        0x00, 0x00, // ARCOUNT
                        0x00};      // hostname length

    char trailer[11] = { 0x05, 0x6c, 0x6f, 0x63, 0x61, 0x6c, 0x00, //local + null terminator
                         0x00, 0x01,   // QDTYPE
                         0x00, 0x01};  // QCLASS

    unsigned short int hostname_len = strlen(hostname);
    unsigned short int packet_size = 24 + hostname_len;
    char* packet = new char[packet_size];
    
  // Forge the request packet
[...]

PR #220, opened 2015-05-10
Fix UDP send to not temporarily use connect()

jwatte:
Using connect() temporarily is incorrect, because any packet arriving for the listening port from another host while inside the beginPacket()/endPacket() bracket will be discarded.

Instead, remember the IP and port used in beginPacket(), and provide those as unicast destination when endPacket() calls send().

A better option would be to change beginPacket() to not take the destination, and make endPacket() take the destination, but this would break compatibility with the Arduino WiFi library.

This change also makes the size of each buffer created for buffering data 512 bytes instead of 1024. Most UDP datagrams are < 256 bytes in size, so this is a better trade-off for using on an embedded device.

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiUdp.cpp
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiUdp.cpp
@@ -132,14 +132,13 @@ int WiFiUDP::beginPacket(const char *host, uint16_t port)
 
 int WiFiUDP::beginPacket(IPAddress ip, uint16_t port)
 {
-    ip_addr_t addr;
-    addr.addr = ip;
-
     if (!_ctx) {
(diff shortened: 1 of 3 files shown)


Makuna:
Could you change this to 576 instead of 512.  See https://tools.ietf.org/html/rfc1122

jwatte:
I could but I don't think that's the important part. The implementation supports multiple local buffers for a single outgoing packet. The point of making the buffers smaller is to waste less RAM in the common case that packets are small. The typical UDP payload is often in the 10-100 byte range, so a size of 128 might be optimal for a small, embedded system. I think 512 is a reasonable compromise (I initially had it at 256.)

If you can explain how the buffer size used for outgoing data transport relate to the reassembly MTU in any meaningful way, then I will update the patch. Meanwhile, I think the patch is good the way it is!

anesthetized:
This could solve my #166

PR #234, opened 2015-05-12
make upload callback packets aligned to defined size

ficeto:
having this a multiple of 512 bytes helps writing to SDcard
2048 looks reasonable and fast, but could be lowered if too much


alonewolfx2:
@anakod right !

igrr:
fair enough — even given the fact that this was almost a complete rewrite, "based on" mention would be welcome.

anakod:
> even given the fact that this was almost a complete rewrite, "based on" mention would be welcome.
You can look, it keep many parts of original code, including some things that what I'm searching for a quite big time and by experimental way.

ficeto:
uhmmm.... there is literally nothing left from the old (testing) code that actually did not work and caused the rewrite. So copyrights were to be gone. They were only guidance.
There is a place where I agree they belong and that is spiffs_esp8266.c, and that is because that part worked as opposed to other code that I also tried. Not that this one did not need things done with it. Begs the question how did it even work?

ficeto:
To explain in better way what I did after lots of issues is get the latest git for spiffs, diff it to see any changes (there were some in the garbage cleaner that actually made use of the erase function). So I copied over only the SPIFFS_tail and EOF methods, thought they were removed? The rest I threw away.
from your abstraction layer as you can see i took the api methods, but added the interrupt disable/enable because the esp was unusable with busy fs. Cleaned lots of the rest (diff if you like) and added the new APIs and format methods.
That ended up in a c form of the things. Then went on turning it into C++, thought to keep the _mountInternal() in there since it's hardware specific (BTW _mountInternal was something that I turned your mount method into after deleting 60% of the code)
Should I get into the linker modifications, filesize defines and so on?
What is left from your code besides the aforementioned spiffs_esp8266.c? Now I do not mind to mention that it was created inspired by your work, but to say that it is your work is not correct.

In development we do copy paste each other's work often, long live the open source! 
The fact that you wrote what you wrote is great, but did not work for me.
It did however give me a guidance on how to interface with the library and what are the proper ways to write to the flash. Cudos for that. You are more than welcome to read my code and copy/implement whatever in your project.

ficeto:
https://github.com/ficeto/Arduino/commit/6e6b26b86139222b8561c1466a186df2fdb36863

ficeto:
did any of you take care to read the code? Or do you go just by the particular commit diff? Do you realize that that file's contents have nothing to do with his code and that it was just an old file of his that I have deleted everything from... there is a clear proof about 1000 pixels above this message... just look at the commits... hard evidence... do any of you code actually? What is the GPL covering? Where is his code? Where? It's "funny" how people jump at conclusions based on somebody's word... If @igrr agrees, I'm full on for removing SPIFFS from the IDE. It's 90% my implementation anyway.

PR #243, opened 2015-05-14
Esp8266

ficeto:
proper error and premature connection loss should be implemented to
handle weird cases where we might not get the whole post content


.gitignore
```diff
@@ -58,3 +58,5 @@ avr-toolchain-*.zip
 /hardware/tools/listComPorts.exe
 
 build/macosx/esptool-*-osx.zip
+
+build/macosx/dist/osx-xtensa-lx106-elf.tgz
```
ficeto: should this be here? it showed up as a new file in the repo

PR #245, opened 2015-05-15
fix SPI speed calculation @160Mhz and many other stuff

Links2004:
fix SPI speed calculation @160Mhz Clock
fix a bug by extensive use of os_printf


Links2004:
have add support for selecting the best AP based of RSSI to my branch.
- including a simple example

PR #247, opened 2015-05-15
Update Esp.cpp

tontito:
getVdd33() read voltage add.


chadouming:
I'll push the correct commit later

tontito:
Ok. Your code no problem.

PR #250, opened 2015-05-16
Add interface to read voltage from vcc pin

chadouming:
Use Esp.getvdd33()

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/Esp.h
+++ b/hardware/esp8266com/esp8266/cores/esp8266/Esp.h
@@ -1,8 +1,8 @@
-/* 
+/*
  Esp.h - ESP8266-specific APIs
  Copyright (c) 2015 Ivan Grokhotkov. All rights reserved.
  This file is part of the esp8266 core for Arduino environment.
- 
+
  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


hardware/esp8266com/esp8266/cores/esp8266/Esp.h
```diff
@@ -78,6 +78,7 @@ class EspClass {
         void restart(void);
 
         uint16_t getVcc(void);
+        uint16_t getVdd33(void);
```
igrr: Why do we need another function instead of replacing getVcc implementation?

chadouming: Like i said, getVcc works differently than readvdd33. I'll go through the
appendix documentation and add info about both in the read me.

igrr:
I understand these implementations work differently, the question is — does the user need both?

chadouming:
Probably someone will need one day.

ficeto:
so here are my two cents...
Neither function does what is should. I get constant 0xFFFF from system_get_vdd33()
and I actually get 12 bit analogRead(A0) at readvdd33()
So if any should be used, that is readvdd33() but at analogRead() as it seem a bit more stable.

chadouming:
Use ficeto's workaround. Work just as well for me.

PR #267, opened 2015-05-18
Fix esptool version field to 0.4.4

marvinroger:
A tiny fix and not a big deal, but the esptool `version` doesn't match to the real one.


marvinroger:
Nevermind, fixed by ec3e6d5deb5ee589deed051d1274dddb3ef1ce82.

PR #311, opened 2015-05-23
Added support for NodeMCU dev kit v2

andysmithfal:


tprochazka:
This implementation fail at least on Windows by this error during compilation:
```
c:/users/tomas/appdata/roaming/arduino15/packages/esp8266/tools/xtensa-lx106-elf-gcc/1.20.0-26-gb404fb9/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: cannot open linker script file {build.flash_ld}: No such file or directory
Error compiling.
```

I added  one line with `nodemcuv2.build.flash_ld=eagle.flash.4m.ld` to the boards.txt and then it works correctly.

But why old version of NodeMCU has several more configuraitons like
```
nodemcu.build.spiffs_start=0x100000
nodemcu.build.spiffs_end=0x3FB000
nodemcu.build.spiffs_pagesize=256
nodemcu.build.spiffs_blocksize=8192
```

Which is not included in V2?

And also naming is a little bit confusing, because official version for this new module is v 1.0
https://github.com/nodemcu/nodemcu-devkit-v1.0
And old one https://github.com/nodemcu/nodemcu-devkit is v0.9
But almost on every eshop sellers use 2.0 :-(

PR #325, opened 2015-05-24
add 2 compiler options from last SDK, __attribute__ format to Print::printf,  link to arduinoWebSockets

Links2004:
- 2 compiler options from last SDK
 - __attribute__ format to Print::printf
 - link to arduinoWebSockets


Links2004:
have add the RSSI function from SDK 1.1.0

Links2004:
SDK patch1 fix all the SPI 80Mhz and irq problems for v1.1.0

PR #326, opened 2015-05-25
Expose Chip CCompare0 timer

Makuna:
This timer is handy for lower level support like PWM/Servo work.
I also cleanup the *sync assembly instructions, some where not needed, others were needed in other places.  The *sync should be used exclusively from each other, and in order of inclusiveness they are ISync, RSync, ESync, and then DSync.  ISync includes the others.


Makuna:
Don't merge yet, there is a bug here.

PR #339, opened 2015-05-26
Update NTPClient.ino

moeur:
Do not hardwire the NTP server IP address into the code.  Instead, look-up the IP address from the host name.  This way you will get a different random server from the pool each time you make a request.


chadouming:
Maybe make it in another example, this adds a lot of overhead since it has to call the DNS.

moeur:
How much overhead does it really add?  If an NTP server IP address is going to be hardwired into the code example, at least don't use one of the servers from a pool, especially not one specifically marked "All services busy, not recommended".  Many people will copy and paste this example and use it as-is which could eventually lead to an overloading of the particular server.

On the same note: change the update period to a much longer time for the same reason.

PR #371, opened 2015-06-02
Update readme to include Servo

Makuna:


Makuna:
Did you see that I corrected?  Did you have to fix it?

PR #388, opened 2015-06-05
fix compiler warnings and handle empty SSID paramater

Links2004:


Links2004:
update to SDK 1.1.1_15_06_05

PR #425, opened 2015-06-14
Don't wait for a connection to an AP when we're supposed to be the AP.

peternewman:


Duality4Y:
I also noticed this, and don't use it in my code when my esp is AP mode

PR #426, opened 2015-06-14
  Don't wait for a connection to an AP when we're supposed to be the AP - Original SDK

peternewman:


igrr:
Thanks, cherry-picked this in 59bd30a.

PR #436, opened 2015-06-16
Fix crash on client disconnect

FlyingHacker:
APOLOGIES: This does not fix the issue. Still looking into this.

If a client connects, but then disconnects before the request is handled the server will crash. This seems to fix the issue.

Test this by putting a few second delay in your  loop() before handling the request (via server.handleClient()), connecting with, e.g. curl, and then cancel the request (control-C out of curl) before the request gets handled. Without the check the server crashes.

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -106,6 +106,10 @@ void ESP8266WebServer::handleClient()
     delay(1);
   }
 
+  if (!client) {
+    return;
+  }
+
   if (!_parseRequest(client)) {
     return;
   }


FlyingHacker:
Actually it looks like this patch fixes the issue:

https://github.com/esp8266/Arduino/commit/0bad8376ac1198babc712136a56acc3d3b07f8f3

Thanks, and sorry for the bad solution above.

igrr:
Actually this one is also required: https://github.com/esp8266/Arduino/commit/86ab7b990a0a725bd55004da75a80286cf1dad72

Otherwise when ERR_ABRT comes this leads to an exception because pcb has been destroyed already.

igrr:
Damn, thanks for making me look at this again — there is a bug with _pcb not being zeroed out if tcp_abort is called.

FlyingHacker:
On Tue, 16 Jun 2015, Ivan Grokhotkov wrote:

> Damn, thanks for making me look at this again — there is a bug with _pcb 
> not being zeroed out if tcp_abort is called.


Ah. No worries. Glad I could help by not helping ;-)

Great lib, BTW. Thx.

-- 
Joe Laffey
The Stable
Visual Effects
http://TheStable.tv/?e38069M/

FlyingHacker:
Hmmm. When I applied https://github.com/esp8266/Arduino/commit/86ab7b990a0a725bd55004da75a80286cf1dad72 

I got crashes again. Is that what you were referring to above? (" there is a bug with _pcb not being zeroed out if tcp_abort is called.")

FlyingHacker:
Hey,

Definitely something fishy. Either of the patches actually still cause crashes when I do as described above with curl. Not sure what the issue is. Thx.

Issue #338, opened 2015-05-26
Using ADC 


fchanson:
Hi,

I made some tests on the ADC using a NodeMCU 0.9.5 build 20150318.
My code just reads the value on the pin every 500ms.

Using a Arduino program : 
- pin A0 at GND --> analogRead(A0) = 700
- pin A0 at 3.3V --> analogRead(A0) = 909

I am surprised to have the value 700 when voltage on pin A0 is GND. Should be near 0.

I made the same test with a LUA program using ESPlorer.
The result is the following :
- pin A0 at GND --> adc.read(0) = 4
- pin A0 at 3.3V --> adc.read(0) = 1023

These last results sound good.

Does anyone have the same problem ?

Need something special to use the ADC pin ?

Thanks.

fchanson:
Hi,

I found in the file 'core_esp8266_wiring_analog.c' the code used for the analogRead method.

extern int __analogRead(uint8_t pin) {
  if(pin == 17){
    return readvdd33() >> 2; // readvdd33 is 12 bit
  }
  return digitalRead(pin) * 1023;
}

Does anyone can explain me this part of code ?
Why the method returns the VDD value divided by 4 ?

Thanks a lot.

igrr:
This seems to be related to having or not having esp_init_default_data.bin flashed.
Apparently if you have it (for instance from a previous NodeMcu install), readvdd33 returns the voltage on VDD pin. If you don't, readvdd33 reads the value from the TOUT pin.

fchanson:
Thanks for your reponse.

I don't know how it is possible to "remove" this file during the flash step.

Could you please help me on this point ?

bluewalk:
Having this exact same issue, is there anything we can do to work around this?

czopas:
Hello, I have the same issue like friends before. I am using voltage divider 10k / 33k so on TOUT (ADC pin) I have 850mV. The result of analogRead(A0) is about 20. I have flashed ESP12 with firmware v0.9.5.2 AT Firmware.bin.
As I think, I should reflash my ESP12 with esp_init_default_data.bin and then ADC should work correct, right ? Where I can find this file ?

bluewalk:
`As I think, I should reflash my ESP12 with esp_init_default_data.bin and then ADC should work correct, right ?`
[...]

PR #443, opened 2015-06-18
Update core_esp8266_wiring_analog.c

Testato:
analogRead bugfix discussed here:
https://github.com/esp8266/Arduino/issues/338

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_analog.c
+++ b/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_analog.c
@@ -17,17 +17,18 @@
   You should have received a copy of the GNU Lesser General Public
   License along with this library; if not, write to the Free Software
   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+  
+  
+  18/06/2015 analogRead bugfix by Testato
 */
+
 #include "wiring_private.h"
 #include "pins_arduino.h"
 
-extern uint16_t readvdd33(void);
-
-void analogReference(uint8_t mode) {}
 
 extern int __analogRead(uint8_t pin) {
   if(pin == 17){
-    return readvdd33() >> 2; // readvdd33 is 12 bit
+    return system_adc_read();
   }
   return digitalRead(pin) * 1023;
 }


me-no-dev:
Maybe here is the place to actually use analogReference and make it switch between vdd and tout.
And even if it does nothing, analogReference should stay for compatibility, and good point on the missing arguments!

me-no-dev:
Also to point out, system_get_vdd33 does return a good clean analog reading of the TOUT pin. The fact that you needed to change it needs resolving, because changing it here will break it for me, so needs to e made to work properly in a different way.

Testato:
@me-no-dev  but, like i described befoure the TOUT can be used for only one think, VDD mesaurement OR analogRead. The way for choose this is changing and uploading the init bin file whit the right value on it (33 for analogread and 255 for VddRead).
But this core actually do not support this function.
So if on the ESPboard is uploaded from factory the init bin file whit the AnalogRead function active, we cannot use system_get_vdd33. 

My pullrequest is only for analogRead problem, and it work. Only after the init bin file uploading function added to this core we can have a good readVdd function.

For the analogReference also is there two way that we can choose, whitout the argument it is unusefull, so what you prefer ? I prefer have an error during compilation instead of a good compilation but no clear/working analogReference function.

Testato:
@me-no-dev 
you write: "system_get_vdd33 does return a good clean analog reading of the TOUT pin"

This is a common error, the official SDK documentation is clear, the system_get_vdd33 does NOT return the analog read on TOUT. 
system_get_vdd33 must return the value on power pin 3 and 4, and in this case the TOUT pin 17 must be floating
I posted the official API explanation of system_get_vdd33 and system_adc_read

Links2004:
will this not work for the init bin problem?
https://github.com/esp8266/Arduino/commit/1803f3052afed566c2698110aa5b447e628bc924#diff-3776add87f5c7f96ac02b7fbfef5305aR204

Testato:
This is good, so igrr start to work on probkem :-)
If you look at 107th byte you have what i writed many more. This byte is at 33, so the Tout pin will be used ONLY for analog read.
We need another PHY setting whit 255, and Tout pin floating, for the vdd supply read.

But for now the init bin file is not uploaded, so we must wait fir this to work

igrr:
With the change in 1803f30 it is not necessary to upload the init bin file. Init data is created in code and passed directly to PHY init function. What you have in flash no longer matters.

Testato:
ok this is very good.
So for the analogread vs vddread how we can pass the 33/255 value to vdd33_const (107th byte of PHY init function ?)

PR #446, opened 2015-06-19
add ESP8266WiFiMesh library

jtfell:
I wrote this library for a university project and thought it might be useful for someone else. I tried to put it in the same format as ESP8266WebServer.


rvbcrs:
Wow NICE! Exactly what I wanted! And I think I'm not the only one..
Thanks!

plinioseniore:
Can one ESP acts as router and client at same time?

jtfell:
Yeah that is the idea. It is not 100% reliable, but works pretty well with a reasonable delay between scanning and checking for requests.

PR #448, opened 2015-06-19
DNSServer can now be configured to reply to all requests with one IP address (e.g. for captive portal)

knovoselic:
Added new feature - DNSServer can now be configured to reply to all requests with one IP address (e.g. for captive portal).
Improved initial example and added CaptivePortal example.


iamthemadz:
Awesome!

PR #461, opened 2015-06-21
ESP8266 Blink example for the blue LED on the ESP-01 module

probonopd:


hardware/esp8266com/esp8266/libraries/esp8266/examples/Blink/Blink.ino
```diff
@@ -0,0 +1,25 @@
...
+ This example code is in the public domain
+*/
+
+const int ledPin =  1;         // The blue LED on the ESP-01 module is connected to GPIO1 
+                               // (which is also the TXD pin; so we cannot use 
+                               // Serial.print() at the same time
+
+void setup() {
+  pinMode(ledPin, OUTPUT);     // Initialize the ledPin as an output
+  ESP.wdtDisable();            // Disable the watchdog timer built into the ESP8266
```
igrr: this is not necessary — works fine with WDT enabled.

probonopd:
Changed as you advised, thanks.

igrr:
Another thing is, all Arduino boards have ```BUILTIN_LED``` defined in pins_arduino.h.
For 'generic' variant this is defined like this:
```C++
static const uint8_t BUILTIN_LED = 1;
```
Other variants (NodeMCU, MOD-WIFI, HUZZAH) also have this defined, so the sketch can use this constant instead of fixed pin number, and this will work on all boards. I really wonder why the standard Blink sketch doesn't do this...

probonopd:
Using BUILTIN_LED now.

Contributing to this project is fun because one learns so many things. I think the project can profit much from having great examples coming with it.

liquiddandruff:
My esp12e doesn't seem to listen when I digitalWrite high/low to pin BUILTIN_LED, nothing happens. I'll investigate later. Great comments by the way, didn't know the led is supposed to activate on low.

Protoneer:
Yes... I seem to have the same issue. Strange thing is that the
ESP8266->Blinky seems to work but when I use one of the wifi libs its stops
reacting to digital High and Low.

PR #476, opened 2015-06-27
Added the SweetPea ESP-210 board to the list of supported boards.

Sweet-Peas:
Let me know of any comments, thanks.
/P


Sweet-Peas:
Changed to use max 1044464 bytes. Just our of curiosity what is setting this max limit  ?

igrr:
This limits max size of the uploaded sketch, and allows the IDE to show % of code space used by the sketch.

PR #495, opened 2015-06-30
Initial Upload From IDE

me-no-dev:
For Test ONLY *** Requires Python to be installed and in the PATH!
After uploading the DNS_SD_Arduino_OTA example in the mDNS server examples through the regular method, your ESP should be visible in the ports selection (well give it a few)
When you select the device also make sure you select from the menu "Flash Using OTA".
change the sketch to your liking and give it another go, this time over the air :)

The python script and example are created by @igrr 
The uploader has some issues still, so expect it to not be 100% reliable.

PR #502, opened 2015-07-01
esp8266 development

Links2004:
update SDK


igrr:
Merged the changes here as part of #495. Thanks.

I decided not to add library dumps because they are much less useful than disassembly of the fully linked executable.

PR #508, opened 2015-07-04
Added WPA sketch

momo96:
I added a sketch that connects to a WPA encrypted network and prints IP, MAC address and SSID

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/examples/ConnectWithWPA/ConnectWithWPA.ino
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/examples/ConnectWithWPA/ConnectWithWPA.ino
@@ -2,7 +2,7 @@
  * This example connects to a WPA encrypted Wifi network.
  * Then it prints the  MAC address of the ESP8266,
  * the IP address obtained, and other network details.
- * 
+ *
  * created 13 July 2010
  * by dlf (Metodo2 srl)
  * modified 5 July 2015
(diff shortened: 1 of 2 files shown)


me-no-dev:
How about this instead:
```cpp
#include <ESP8266WiFi.h>

const char *ssid = "yourNetwork";
const char *password = "secretPassword";

void setup() {
  //Initialize serial
  Serial.begin(115200);
  
  // attempt to connect to Wifi network:
  while (true) {
    Serial.print("Attempting to connect to WPA SSID: ");
    Serial.println(ssid);
    
    // Connect to WEP/WPA/WPA2 network:
    WiFi.begin(ssid, password);
    if(WiFi.waitForConnectResult() == WL_CONNECTED){
      break;
    }
    // wait 10 seconds for connection:
    delay(10000);
  }
  
  // you're connected now, so print out the data:
  Serial.print("You're connected to the network ");
  printCurrentNet();
  printWifiData();

}

void loop() {
  // check the network connection once every 10 seconds:
  delay(10000);
  printCurrentNet();
  printWifiData();
}

void printWifiData() {
  // print your ESP8266 IP address:
  Serial.print("IP Address: ");
  
  // print your MAC address:
  Serial.print("MAC address: ");
  Serial.println(WiFi.macAddress());

}

void printCurrentNet() {
  // print the SSID of the network you're attached to:
  Serial.print("SSID: ");
  Serial.println(WiFi.SSID());
  
}
```

igrr:
@momo96 can you please make the changes suggested above:
- use `WiFi.waitForConnectResult` to wait for wifi connection.
- use `Serial.println(WiFi.macAddress())` instead of formatting MAC address manually
- no need to add delay after Serial.begin()
Thanks.

momo96:
@igrr I made all the changes, except for `<Serial.println(WiFi.macAddress())>` which outputs this error: 
> no matching function for call to 'ESP8266WiFiClass::macAddress()'

igrr:
That's weird. Here's this function:
https://github.com/esp8266/Arduino/blob/0cc192525fe4d9f784361408dbb82fff2fb96b39/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/ESP8266WiFi.h#L121

hardware/esp8266com/esp8266/libraries/ESP8266WiFi/examples/ConnectWithWPA/ConnectWithWPA.ino
```diff
@@ -0,0 +1,79 @@
...
+  //Initialize serial and wait for port to open:
+  Serial.begin(115200);
+
+  // attempt to connect to Wifi network:
+  while (true) {
+    Serial.print("Attempting to connect to WPA SSID: ");
+    Serial.println(ssid);
+
+    // Connect to WEP/WPA/WPA2 network:
+    WiFi.begin(ssid, password);
```
mangelajo: Please bear in mind if you use WiFi.begin() alone, it will try any network you previously connected to
in other runs.

That's useful as a sketch because people will get connected to their know wifi first.

PR #530, opened 2015-07-10
digitalWrite cleanup and more compliant with behavior on AVR

h4rm0n1c:
I rewrote digitalWrite because the existing version was breaking
functionality as compared to how it behaves on the AVR.
Specifically, I could not use digitalWrite for a library that works fine on the AVR.

Instead I had to resort to fiddling with GPOC and GPOS and bit masks,
but this rewrite made all of that unnecessary, for whatever reason it
just works better.

This version borrows a little from the AVR library in the sense that the
same logic is applied to determine whether a pin should be high or low
as the AVR version, and yes, it does appear to make a difference, at least as far as my testing on hardware has shown.

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.c
+++ b/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.c
@@ -77,11 +77,13 @@ extern void __pinMode(uint8_t pin, uint8_t mode) {
 }
 
 extern void ICACHE_RAM_ATTR __digitalWrite(uint8_t pin, uint8_t val) {
+  if (pin == 16) {
     if(val == LOW) {
 	  GP16O &= ~1;
 	} else {
 	  GP16O |= 1;
 	}
+  } else if ((pin >= 0) && (pin <= 15)) {
     if(val == LOW) {
       GPOC = digitalPinToBitMask(pin);
 	} else {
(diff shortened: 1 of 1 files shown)


igrr:
The diff appears to miss an `if` related to pin number (==16 or not).

```
/Users/igrokhotkov/projects/esp8266/arduino/build/macosx/work/Arduino.app/Contents/Java/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.c -o /var/folders/15/ybtbgzpj7636vv4wp92l2c700000gn/T/build6991830046946292972.tmp/core_esp8266_wiring_digital.c.o 
/Users/igrokhotkov/projects/esp8266/arduino/build/macosx/work/Arduino.app/Contents/Java/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.c:91:1: error: expected identifier or '(' before '}' token
 }
 ^
Error compiling.
```

igrr:
Could you please point to the library which caused issues? As far as I can tell, the difference between old and new code is that if someone calls digitalWrite with a even non-zero value argument (like 2 or 4), old code would interpret that as 0, hence output would be low, and in new version we will go into else branch and set output high.

Arduino documentation doesn't specify what should happen when digitalWrite is passed value which is not LOW or HIGH, so i think the library you had trouble with makes some assumptions and passes some other integers into digitalWrite.

igrr:
If that is the case, the fix should be as simple as removing this line:
```C++
val &= 0x01;
```

h4rm0n1c:
I'm sorry, I somehow managed to paste a broken function in, I could have sworn it was complete when I copied it from one file to another.

I've corrected it.

igrr:
Could you please also try the original version with this line removed: `val &= 0x01;`?

h4rm0n1c:
I'm happy to report that you're correct, sorry if I got a little overzealous.

PR #531, opened 2015-07-10
Added setTxMode to SPI Library

h4rm0n1c:
setTxMode currently allows for either Duplex or Transmit only style operation.

Essentially, you can Transmit with SPI without losing GPIO12 if you have no need to receive SPI on MISO.


h4rm0n1c:
Never mind.

I'm an idiot.

PR #532, opened 2015-07-10
add synchronous ota

me-no-dev:


kzyapkov:
This works for me. Additionally, this [change](https://github.com/kzyapkov/Arduino/commit/f5653529ecbad1a29631198e2740ed9395e247f3) makes espota.py work with python3 (the other option was to use python2 exeutable which will probably break on many systems).

PR #544, opened 2015-07-12
README: Minor readability improvments

noahwilliamsson:
More consistent formatting and a few fixed typos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,7 @@ $ ant dist
 
 ```pinMode()```, ```digitalRead()```, ```digitalWrite()```, ```analogWrite()``` work as usual.
 
-Pin numbers correspond directly to the esp8266 GPIO pin numbers. To read GPIO2,
-call ```digitalRead(2)```
+Pin numbers correspond directly to the esp8266 GPIO pin numbers. To read GPIO2, call ```digitalRead(2)```.
 
 GPIO0-GPIO15 can be ```INPUT```, ```OUTPUT```, ```INPUT_PULLUP```, and ```INPUT_PULLDOWN```.
 GPIO16 can be ```INPUT```, ```OUTPUT``` or ```INPUT_PULLDOWN```.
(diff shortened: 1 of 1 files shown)


igrr:
Hi, thanks for the improvements. I'm going to merge everything except *"Consistently indicate functions with parentheses"* because while some functions really take zero arguments (like `millis()` and `micros()`), others need arguments (like `pinMode` and `attachInterrupt`), and introducing them without arguments (or some placeholders) is simply not valid. I think the way to go is to make these functions work like hyperlinks pointing to detailed documentation.

noahwilliamsson:
Hi, thanks for the feedback.  

If you're not merging af1b15b you can also skip the dependent commit 68ba4c3.

I agree it would be better show their arguments and also hyperlink them.

If I was to hyperlink the first mention of each function, would it be OK to link to the Arduino reference and use Arduino's syntax style (i.e. leaving out the parameters' data types) where possible?  For instance, like this for the _Basic Wiring Functions_ block:
" [pinMode(pin, mode)](https://www.arduino.cc/en/Reference/pinMode), [digitalRead(pin)](https://www.arduino.cc/en/Reference/digitalRead), [digitalWrite(pin, mode)](https://www.arduino.cc/en/Reference/digitalWrite), [analogRead(pin)](https://www.arduino.cc/en/Reference/analogRead) and [analogWrite(pin, value)](https://www.arduino.cc/en/Reference/analogWrite) work as usual."

(For the board specific methods (```Serial.swap()```, ```Serial.setDebugOutput()```, ```ESP.restart()```, etc) I guess `README.md` will have to do as the authoritative source as there doesn't seem to be any other place where those methods are documented.)

igrr:
Yes, i think keeping the Arduino syntax style and linking to Arduino reference is fine.
I have cherry-picked your commits on to the doc-update branch:
https://github.com/esp8266/Arduino/tree/doc-update
I've started cleaning up and separating the documentation, the goal is to provide "snapshots" for each core version. Right now everyone looks at the most recent version and this leads to some confusion.

igrr:
These commits are now part of https://github.com/esp8266/Arduino/pull/587, closing this one.

Issue #540, opened 2015-07-12
Standard Arduino practices cause WDT to reset esp8266


Makuna:
One standard practice in many examples is to wait for serial input in a tight loop like ...
```
while (!Serial.available());
```
This will cause the WDT to trigger.

Can we insert a yield inside the available() when it would return false so that this common code fragment would not cause a processor reset?

skandragon:
Go for it, but this also causes a WDT to trigger:

```
void setup()
{
  Serial.begin(115200);
  delay(100);
  Serial.println("\nTEST\n");
  delay(1000);
}

void loop() {
  yield();
}
```

I would assume it's broken, except that it works fine with nodeMCU loaded, or at least I can fiddle with pins and type at it.

me-no-dev:
@Makuna doing it is not a bad idea, but it might put the current code to wait when that is not expected.
Any use cases that would have a problem come to mind?
Can't we find a better way? Maybe find out before wdt triggers and feed it? Do you know of any relevant interrupt vector?

skandragon:
Does the standard Arduino even have a WDT enabled by default?  I also see no mention of a WDT in this project's README, other than this little bit:

    ESP-specific APIs

    APIs related to deep sleep and watchdog timer are available in the ESP object,
    only available in Alpha version.

Also, no mention how to find and use this "alpha" version (which I assume is based on master?)

me-no-dev:
this is old text. standard (AVR) arduino does not have watchdog like this one.

Makuna:
The standard Arduino (AVR) does  NOT have the WDT turned on by default. AVR does support a WDT in the processor (I use it) but its uncommon to find it being used.  
The esp8266 does due to the WiFi that runs in the background.  
This is also why I believe most of the available() methods should contain a yield() so that the WiFi will continue to function as expected.  Note that a few of the available() methods on libraries already have a yield() in them, but not all.
I have the changes and will be testing it shortly.

Makuna:
This is what is done in some available() methods, and the proposal is to replicate something similar to all of them.  
```
    static uint32_t lastPollTime = 0;

    if (lastPollTime > esp_micros_at_task_start())
        yield();
```

me-no-dev:
reding the source behind, I come to conclusion that yield will be called every next time, but first, inside the same loop() call. Am i getting this right?
esp_micros_at_task_start is set before calling each loop()

Makuna:
[...]

PR #534, opened 2015-07-11
Issues 475 & 484 fixes

Makuna:
https://github.com/esp8266/Arduino/issues/475
https://github.com/esp8266/Arduino/issues/484

PR #545, opened 2015-07-12
Available yield

Makuna:
https://github.com/esp8266/Arduino/issues/540

NOTE:  There is bleed over in the changes from a previous pull request, don't know how to clean that out other than revert it on my branch; BUT I DO NOT WANT TO REVERT THOSE CHANGES if they get merged in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,8 +213,8 @@ Libraries that don't rely on low-level access to AVR registers should work well.
 - [Blynk](https://github.com/blynkkk/blynk-library) - easy IoT framework for Makers (check out the [Kickstarter page](http://tiny.cc/blynk-kick)).
 - [DallasTemperature](https://github.com/milesburton/Arduino-Temperature-Control-Library.git)
 - [DHT11](https://github.com/adafruit/DHT-sensor-library) - Download latest v1.1.0 library and no changes are necessary.  Older versions should initialize DHT as follows: ```DHT dht(DHTPIN, DHTTYPE, 15);```
-- [NeoPixel](https://github.com/adafruit/Adafruit_NeoPixel) - Adafruit's NeoPixel library, now with support for the ESP8266 (use version 1.0.2 or higher from Arduino's library manager).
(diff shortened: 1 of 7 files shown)


igrr:
Since there isn't a way to handle this yield issue in one single place, i would propose to reduce the amount of code duplication by refactoring this as follows.

- Introduce a global function, e.g. `maybe_yield()`:
```C++
void maybe_yield() {
  if (micros() - esp_micros_at_task_start() > 10000) {
    yield();
  }
}
```
- Update `available` methods to use it instead of `esp_micros_at_task_start` (which was a quick and dirty hack on my side):

```C++
bool isAvailable = ... ; // depending on the class
if (!isAvailable) {
  maybe_yield();
  isAvailable = ...; // perhaps update isAvailable, depending on the class?
}
return isAvailable;
```

igrr:
Also i would like to merge your fixes in 4d3ef20. Would you mind if I rebase that change directly right now and then we go with a clean pull related to isAvailable?

Makuna:
Go ahead and take the other, I will refactor this stuff and put out a new pull

Makuna:
Note that https://github.com/esp8266/Arduino/pull/534 already contains the changes you want, just merge that.

Links2004:
will the ```yield``` brake the code if i call it for example in the Ticker or the IO interrupt handler?

igrr:
That's right, yield is non-reentrant, so calling it from Ticker or GPIO interrupt will cause an exception.
Doing `while(!Serial.available()){}` from an interrupt is not good either, but i think it would be good not to yield on the first invocation of available(). That was the rationale behind the existing code — so that you could call available() at least once from system context.

Links2004:
yes ```while(!Serial.available()){}``` is a bad idea,
but a simple ```int i = Serial.available();``` need to work otherwise we get more problems again.

igrr:
My suggestion here is to only yield if available() was called more than once with zero result.
This way we can still read all available data from both sketch and system contexts, and handle the case of `while(!available());` loop.

Makuna:
this is replaced by https://github.com/esp8266/Arduino/pull/551

PR #551, opened 2015-07-13
optimistic_yield()

Makuna:
this introduces optimistic_yield() used for when standard library
methods are normally used in tight loops waiting for something to
happen, like available().

this replaces https://github.com/esp8266/Arduino/pull/545


Links2004:
I like the ISR check.
one question you add at all files ```extern "C" void optimistic_yield();``` is there a reason for this? 
i think its better to add it to the header file like the ```yield()``` or is there a problem that i don't see?
https://github.com/esp8266/Arduino/blob/ea302aab05480ad36c076b618abb642b1eb1893b/hardware/esp8266com/esp8266/cores/esp8266/Arduino.h#L41

Makuna:
@Links2004 (fixed) I meant to look into that but I was just following what was already done.

igrr:
That's a good addition, thank you.
I also think of moving this check into `yield()` itself, because that would make it more robust. Additionally `esp_yield` needs an additional check to see if it is being executed from sketch context (as opposed to system task).

Issue #535, opened 2015-07-11
TCS34725 RGB sensor with i2c interface fails


bbx10:
The [TCS34725](http://adafru.it/1334) i2c RGB color sensor returns garbage RGB values using board package 1.6.4-673-g8cd3697. Since there are 3 changes to the i2c driver since the 1.6.4 release, I copied core_esp8266_si2c.c from master and overwrote the version installed with 1.6.4. The TCS sensor now works based on limited testing.

Since the fix is on master should I close this issue? I posted just in case someone is working with this sensor.

bbx10:
The sensor is still returning garbage values often enough to make it useless. Sometimes the sensor gets stucks returning garbage until it is reset.

When comparing logic analyzer traces of 16 bit reads, the the last byte is followed by NACK. This is correct according to the i2c spec because the NACK tells the slave-sender to not send another byte.

With this change, the sensor behaves about the same as when it is connected to an Uno.

Issue #526, opened 2015-07-09
i2c to BNO055 fails


skandragon:
I've been trying to connect a BNO055 (from Adafruit) to the Adafruit Huzzah board, but it fails to work.  I have tested the BNO055 against an actual Arduino, and it works.

I've tried using pull-up resistors of various values, but without success.

I've tried setting the Wire.setClock() to 100000, 200000, 300000, and 400000 as well as leaving it unset.  I've tried different pins for SCA/SCL without success.

I don't have access to a logic analyzer or oscilloscope at the moment to dig into this at the protocol level, but I'm hoping someone here has advice to try.

I did see #254, which seems related.

igrr:
#254 is actually fixed, not sure why it's still open.

A logic analyzer or a scope would really be essential to diagnose this issue... Is there any chance that you borrow one to do a test?

skandragon:
I'm currently traveling (who doesn't work on electronics on a plane?) but I'll poke around and see if anyone has one handy.  I'm in the SF Bay area; surely someone has one here!

lrmoreno007:
You don't have a oscilloscope but you have an Arduino... Then you have a Oscilloscope.

skandragon:
I've got little 8 Mhz devices, and so far everyone recommends 16 Mhz to get good results.  The 16 Mhz devices I know of are all 5.0 v, and I'm using 3.3v, but technically that should work well enough, but it might extend high and low times.  I think then 'scope would be better, and I might have access to one tonight.

me-no-dev:
16MHz AVRs run just fine on 3.3V :) even though it's not recommended

skandragon:
I grabbed a 16 Mhz Arduino last night, and tried using it.  I did get some captures, but nothing of note.  Part of the problem is the WDT triggers, and the ESP8266 is basically in a reset - probe loop, so capturing the actual data is harder than it should be.  I'll modify the code to do less in setup(), and put a delay in for the i2c part as well. I may even teach it to cause a particular pin to rise or fall just before the i2c part starts, so I can get a sane trigger point for the scope.

skandragon:
[...]

PR #554, opened 2015-07-15
NACK last byte when read

bbx10:
The TCS34725 RGB color sensor works reliably with this change. See #535 for details.

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
+++ b/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
@@ -1,9 +1,9 @@
-/* 
+/*
   si2c.c - Software I2C library for esp8266
 
   Copyright (c) 2015 Hristo Gochkov. All rights reserved.
   This file is part of the esp8266 core for Arduino environment.
- 
+
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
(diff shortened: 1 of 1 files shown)


me-no-dev:
have you tested this with other i2C devices?

bbx10:
Yes, a [9DOF IMU](http://adafru.it/1714) board which was not working now works with this change and the clock stretch increase (#553). This change might help with the other IMU board (#526).

I will test 4 other i2c devices with this change and #553 tomorrow. 

I am using the following for the i2c protocol reference: 
http://www.nxp.com/documents/user_manual/UM10204.pdf.

Section 3.1.6 talks about 5 conditions for generating NACK. The last one says
> 5. A master-receiver must signal the end of the transfer to the slave transmitter.

This is why the Uno/AVR generates NACK after the last byte read.

igrr:
We used to have this correct long long time ago :)
https://github.com/esp8266/Arduino/blob/7e40b9f11e9035af57248d3d19af50deadeea14c/hardware/esp8266com/esp8266/cores/esp8266/i2c.cpp#L164

I will also test a few more devices myself with this change and #553.

bbx10:
> We used to have this correct long long time ago :)

Thanks, this helps assure me this is a good change. The change affects every i2c read so it has the potential for massive breakage. :(

The 9DOF IMU ran for 3 hours and was still reporting correct values. The i2c chips on board are the L3GD20H and LSM303.

The following other i2c devices work. Only a few minutes testing done per device.
BMP180, HTU21DF, ISL29125, TSL2591, TCS34725

All but the TCS34725 worked even without this change. All testing was done with #553 as well as this change.

igrr:
I have tested this change with MPR121, HMC6352, and one I2C EEPROM (forgot the part name).
All seem to be work fine.

PR #560, opened 2015-07-16
Update README.md

h4rm0n1c:
Corrected missing "PullUp" in the "Minimal hardware Setup for Bootloading and usage" table, now consistent with the "ESP to Serial" schematic just below it.

Corrected case of schematic titles, it just looks nicer.

Added a caveat about the potential issues of trying to take shortcuts with pull-down and pull-up resistors, there's a few people scratching their heads on the forums because of this often overlooked, self inflicted issue, much the same as people were scratching their heads regarding stability issues whilst trying to use their USB to Serial adapters as a power supply.

Rewording of USB to Serial Note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ You need to put ESP8266 into bootloader mode before uploading code.
 For stable use of the ESP8266 a power supply with 3V3 and >= 250mA is required.
 
 * Note
- - using Power from USB to Serial is may unstable, they not deliver enough current.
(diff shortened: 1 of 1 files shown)


igrr:
Could you please re-open this pull request against https://github.com/esp8266/Arduino/tree/doc-update branch instead of this one? I'm rewriting documentation at the moment so these changes are much welcomed there.

igrr:
In particular, core documentation has moved to https://github.com/esp8266/Arduino/blob/doc-update/hardware/esp8266com/esp8266/doc/README.md.
I was also going to split the part related to hardware setup for bare modules into a separate file.

h4rm0n1c:
Okay, no worries, thanks. :)

PR #575, opened 2015-07-20
Re-added lost function memcpy_P

h4rm0n1c:
This was lost in https://github.com/esp8266/Arduino/commit/80a5f29e89ebb48d2414edd75e1847c3f47798e4

I've also changed the type of src to PGM_VOID_P to match the other changes made in @Makuna 's PR.


Makuna:
looks good, thanks for fixing this.

PR #607, opened 2015-07-24
Fix GPIO Interrupt

Makuna:
GPIO interrupt was not restricting the level correctly, the ISR itself
could be interrupted by another ISR.  This will currently make it act
more inline with AVR Arduino and thus stopping ALL other interrupts
from triggering.

@igrr Further, this will fix the WithinISR and the issue with it using int level you found with GPIO interrupts.

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.c
+++ b/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.c
@@ -113,7 +113,7 @@ void interrupt_handler(void *arg) {
   uint32_t status = GPIE;
   GPIEC = status;//clear them interrupts
   if(status == 0 || interrupt_reg == 0) return;
-  ETS_GPIO_INTR_DISABLE();
+  noInterrupts();
   int i = 0;
   uint32_t changedbits = status & interrupt_reg;
   while(changedbits){
@@ -126,7 +126,7 @@ void interrupt_handler(void *arg) {
       handler->fn();
     }
   }
-  ETS_GPIO_INTR_ENABLE();
+  interrupts();
 }
 
 extern void __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int mode) {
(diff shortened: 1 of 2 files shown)


Makuna:
Grrr, again, hit with GitHub problem and even though a merge succeeded, it got lost?

@igrr If you are uncomfortable with having all interrupts suspended, then the following changes will work.

1) fix the Arduino.h by adding the stringify macro above the xt_disable_interrupts macro and replace the original nointerrupts macro.
```
#ifndef __STRINGIFY
#define __STRINGIFY(a) #a
#endif

// level (0-15),
// level 15 will disable ALL interrupts,
// level 0 will disable most software interrupts
//
#define xt_disable_interrupts(state, level) __asm__ __volatile__("rsil %0," __STRINGIFY(level) : "=a" (state))
#define xt_enable_interrupts(state)  __asm__ __volatile__("wsr %0,ps; isync" :: "a" (state) : "memory")

extern uint32_t interruptsState;

#define interrupts() xt_enable_interrupts(interruptsState)
#define noInterrupts() xt_disable_interrupts(interruptsState, 15)
```
2) then use the xt_disable_ xt_enable in the GPIO ISR like
```
void interrupt_handler(void *arg) {
  uint32_t status = GPIE;
  GPIEC = status;//clear them interrupts
  if(status == 0 || interrupt_reg == 0) return;


  // pin interrupts are level 1
  // this will stop other level 1 but not higher level interrupts
  uint32_t intrState;
  xt_disable_interrupts(intrState, 1);

  int i = 0;
  uint32_t changedbits = status & interrupt_reg;
  while(changedbits){
    while(!(changedbits & (1 << i))) i++;
    changedbits &= ~(1 << i);
    interrupt_handler_t *handler = &interrupt_handlers[i];
    if (handler->fn && 
        (handler->mode == CHANGE || 
         (handler->mode & 1) == digitalRead(i))) {
      handler->fn();
    }
  }
  xt_enable_interrupts(intrState);
}
```

igrr:
Okay, I'll look a bit more into this, probably have to deal with #615 at the same time as well.

Makuna:
My comment is wrong, gpio are level 0 interrupts, and setting the level will only block interrupts below the level set.

Makuna:
we talked already a little about this and 615, did you want me to implement and resubmit as this pull should not be applied, or are you going to take it from here?

igrr:
I would appreciate you doing this as you have proposed: make noInterrupts and interrupts work without saving state, and use xt_ macros (or InterruptLock class) where state has to be saved. I won't be able to do this myself for the next few days.

Issue #596, opened 2015-07-23
Crash when sending content from flash to client using the webserver


rnplus:
I get this error using the stable, staging and latest (compiled from source) version of the arduino package.

```
Exception (3):
epc1=0x4000bef4 epc2=0x00000000 epc3=0x00000000 excvaddr=0x40243566 depc=0x00000000
 
ctx: cont
sp: 3ffec940 end: 3ffecbd0 offset: 01a0
 
>>>stack>>>
3ffecae0:  40243566 00000000 00000000 00000030  
3ffecaf0:  00000000 00000000 3ffeaec8 4020244a  
3ffecb00:  3fff7120 00004526 00004526 00000030  
3ffecb10:  3fff65b8 00000000 00000000 4020742e  
3ffecb20:  00000000 00000000 3ffeaec8 40207a7d  
3ffecb30:  00000011 3fff6c08 00000003 00000003  
3ffecb40:  3fff7060 0000001e 0000001e 402052bd  
3ffecb50:  3ffea360 00000000 000003e8 3ffebbb0  
3ffecb60:  00000000 3fff70e8 0101a8c0 3ffecbfc  
3ffecb70:  3fffdc20 000003e9 3ffeaec8 40207b6d  
3ffecb80:  3ffea360 00000000 000003e8 000147a4  
3ffecb90:  3ffeaee0 3fff70e8 00000016 40102185  
3ffecba0:  40201cdd 00000000 00000000 402044f6  
3ffecbb0:  3fffdc20 00000000 3ffecbf4 40201d3a  
3ffecbc0:  00000000 00000000 3ffebbb0 40100378  
<<<stack<<<
 
 ets Jan  8 2013,rst cause:1, boot mode:(1,7)
 
 
 ets Jan  8 2013,rst cause:4, boot mode:(1,7)
 
wdt reset
 
```

``` 
server.on ("/style.css", []() { server.send ( 200, "text/css", DATA_css ); } );
 ```

Data:
```
const char DATA_css[] PROGMEM = R"=====(
body {margin: 0px; padding: 0px; font-family: "Open Sans", sans-serif; font-color: #000; font-size: 14px; background-color: #F2F2F2;}
div.header {position: absolute; left: 0px; right: 0px; top: 0px; background-color: #363636; color: #FFF; margin: 0px; padding: 0px; padding-bottom: 0px; height: 55px;}
div.header h1 {padding: 0px; margin: 0px; padding-top: 5px;}
[...]

PR #611, opened 2015-07-24
send_P, sendContent_P, and FPSTR

Makuna:
This allows for the content of server response to be stored in flash
memory.

see https://github.com/esp8266/Arduino/issues/596#issuecomment-124151609 for some background.


martinayotte:
Thanks, Makuna !

PR #648, opened 2015-08-03
Avrisp

kzyapkov:
A library and example sketch which allow programming AVR targets with avrdude over the network like so:

```
avrdude -c arduino -p m2560 -P net:esp8266-avrisp:2560 -t
```

### DIFF
--- a/build/shared/tools/ESP8266FS/make.sh
+++ b/build/shared/tools/ESP8266FS/make.sh
@@ -15,7 +15,7 @@ fi
 set -e
 
 mkdir -p bin
-javac -target 1.7 -cp "$pde_path:$core_path" \
+javac -target 1.8 -cp "$pde_path:$core_path" \
       -d bin src/ESP8266FS.java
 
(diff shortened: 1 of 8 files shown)


FernandoGarcia:
Hi @kzyapkov!

Firstly happy new year!

Is possible print serial messages from Arduino to a virtual port using this library?

I would like to use it as a bridge to see debug messages from Arduino and programmer.

I found [this tutorial](https://hackaday.io/project/13408-wifi-avr-isp-using-esp8266) but I would like to use without change IDE, mainly because I can't see network port in  tools menu. 

I'm using Tibbo to create a virtual port and AVRDUDESS to burn the hex file.

Best regards.

kzyapkov:
Hi @FernandoGarcia,

Yes, piping the serial port should be possible, but you'd have to disable debugging on the primary UART of the ESP board, run two extra wires for RX and TX and implement a second TCP server for the serial pipe.

PR #678, opened 2015-08-09
Made changes to allow the compiler to compile without warnings.

nkolban:
Changed a few source files to remove compiler warnings when compiled under the Eclipse framework.


far5893:
I have  55  warnings in stable 23 jul under eclipse ,i will try in next stable.
Under arduino ide is quite  simple update from stable releases , but under eclipse
If i change board with new version file remains a lot of old build enviroment variables.

PR #709, opened 2015-08-19
Esp8266 nonblocking connect for WiFiClient

sdrsdr:
The default blocking behaviour of connect is not suitable for all needs :) Here is a patch to introduce a new connectex method with third parameter for behaviour selection


igrr:
I've got a few comments:

- non-blocking network APIs are a much requested feature, so :+1: overall!
- I think it would be better not to mix blocking and non-blocking APIs into one library. Mixing everything into one library we are more likely to create a monster which will be hard to support :) Perhaps we should start working on ESP8266Net library or smth like that.
- usage scenario: i might be mistaken, but there is no option for calling a user callback once connection is established. For me this kinda defeats the purpose of an asynchronous API. I would have to do polling to see if the connection was established. Perhaps I'm missing something. Do you have some example sketch which demonstrates how this should be used?
- method naming: we're not programming with Windows API here, so naming some function variant with "Ex" doesn't make much sense for the user. `connectAsync` might be a better choice. Another option is to add change signature of `connect` function to look like `virtual int connect(IPAddress ip, uint16_t port, bool block = true)`.

If you like the idea of working on a new library for async networking, let's discuss that in the [chat](https://gitter.im/esp8266/Arduino).

me-no-dev:
@igrr this in no way is async library. It just does not yield :)
@sdrsdr lots of the code you wrote is unnecessary. If on _connect you save the type of connection, you can later just add to the appropriate already existing methods and if/else to act appropriate.
```cpp
void WiFiClient::_err(int8_t err)
{
    DEBUGV(":err %d\r\n", err);
    if(!_block) esp_schedule();
   else _asyncerr=true;
}
```

sdrsdr:
@me-no-dev: you are right but this would modify "already working" code which is often frowned upon.  The new code is really small and easier  to track - a separate "code paths" for non blocking event handling  and for the blocking behaviour

me-no-dev:
i looked at all methods and they are the same, with only similar changes to what I've shown above.
Now you have the same code twice.

miguelpom:
Hello, everyone!
I'm currently working on a project with the ESP8266's module "NodeMCU". In this project I am supposed to monitor a matrix of digital sensors from the ESP8266 and, whenever the sensors aren't being "used" (if there is a time window of at least 2 seconds), I must send the accumulated readings over to a Raspberry Pi (working as server) using a WifiClient class.

@sdrsdr : I was hoping to use your non-blocking WifiClient.connect() function (since using the conventional one blocks my program for 5 seconds aprox.).
I tried to install your library implementation, but the Arduino IDE returns a "no valid" message.
Now I see that you deleted the specific branch more than a year ago. 
Is there any way that I can make the library working?
I am new to this module. I tried to copy/paste the "WifiClient.h" and "WifiClient.cpp" files over to the main library imlemented by @igrr . Obviously the branch changed more than just two files and dependencies.
Any help/advice you may give me would be great!

P.S.: The non-blocking feature is quite critical for my project to work. I can give you further description of the project if it helps.

Thanks a lot!!

sdrsdr:
The code I wrote is very old and probably will not compile properly. Non blocking was actually non yielding and generally not implemented well as I remember .. since then I've changed interest. My best advice is to use UDP instead of TCP if there is no NAT between client and server or the server need not respond this is a better and more suitable solution 

Good luck!

me-no-dev:
@miguelpom if you want non-blocking TCP you might want to look at https://github.com/me-no-dev/ESPAsyncTCP

miguelpom:
@me-no-dev , actually I have no problem with the blocking ".write()" function in TCP. My time-critical method inside WifiClient is ".connect()". Whenever my esp8266 wants to stablish a connection with the server, the sketch waits for a maximum of 5 seconds  inside de "connect(...)" function.
I am looking for a non-blocking "connect" function, but reducing that 5-seconds period down to 2 seconds would be equally good for my project.

PR #714, opened 2015-08-21
add toCharArray() to IPAddress class

martinayotte:
I was tired to grab/merge this personal change from one version to the other.

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/IPAddress.cpp
+++ b/hardware/esp8266com/esp8266/cores/esp8266/IPAddress.cpp
@@ -64,3 +64,17 @@ size_t IPAddress::printTo(Print& p) const {
     return n;
 }
 
+char *IPAddress::toCharArray()
+{
+    static char szRet[20];
+    String str = String(_address.bytes[0]);
+    str += ".";
+    str += String(_address.bytes[1]);
+    str += ".";
+    str += String(_address.bytes[2]);
+    str += ".";
+    str += String(_address.bytes[3]);
+    str.toCharArray(szRet, 20);
+    return szRet;
+}
+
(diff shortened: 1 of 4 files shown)


igrr:
Why not add a "toString" method which would return a String? This way you could avoid the hack with a static char array.

sdrsdr:
why not snprintf to the buffer:

 snpintf(szRet,20,"%u.%u.%u.%u",(unsigned int) _address.bytes[0],(unsigned int) _address.bytes[1],(unsigned int) _address.bytes[2],(unsigned int) _address.bytes[3]);

Links2004:
using ```static char szRet[20];``` will restiveness 20 byte even when you never use this function.
i think the string way is the better one. you can easy make a char array out of the String.

i prefer 
```cpp
String IPAddress::toString()
{
     char szRet[16];
     sprintf(szRet,"%u.%u.%u.%u", _address.bytes[0], _address.bytes[1], _address.bytes[2], _address.bytes[3]);
    return String(szRet);
}
```

or for char*
```cpp
char * IPAddress::toString(char * buf)
{
    sprintf(buf, "%u.%u.%u.%u", _address.bytes[0], _address.bytes[1], _address.bytes[2], _address.bytes[3]);
return buf;
}

then its the same way we do it for the mac for example:
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/ESP8266WiFi.cpp#L296

martinayotte:
Ok ! I've pushed the first suggestion from Links2004.
I didn't took second suggestion, since it requires the caller to provide the storage, and that why I used static in the first version. Callers can still do toString().c_str() if "char *" is required.

martinayotte:
The following thread is meaningful for my last commit : 
https://github.com/esp8266/Arduino/issues/596

PR #729, opened 2015-08-29
fix mac 599 for ESP8266HTTPUpdate

Links2004:


hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -300,3 +300,27 @@ void WiFiClient::stopAll()
         }
     }
 }
+
+
+void WiFiClient::stopAllexcepted(WiFiClient * exC) {
+    for (WiFiClient* it = _s_first; it; it = it->_next) {
+        ClientContext* c = it->_client;
+
+        if(c && exC->_client) {
```
igrr: Can it be simply `if (exC->_client == it->_client)`? As far as i can tell, pcb is unique for all WiFiClient instances corresponding to the same connection.

hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -300,3 +300,27 @@ void WiFiClient::stopAll()
         }
     }
 }
+
+
+void WiFiClient::stopAllexcepted(WiFiClient * exC) {
```
igrr: Probably better name would be `stopAllExcept`

hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiClient.cpp
Links2004: yes good point ```_client ``` its an ptr so we can compare the the address, will change it.

PR #743, opened 2015-09-03
Support for multiple I2C buses

paulo-raca:
The ESP8266 doesn't have "actual" I2C hardware, and relies on bit-banging instead.

Since this is a sotware implementation, there is no reason we cannot bit-bang a bunch of I2C ports, right?

My use-case: I need to talk to 2 devices with the same address (Same chip - PN532). AFAIK, the only way is to put them in different buses.

### DIFF
--- a/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
+++ b/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
@@ -22,15 +22,12 @@
 #include "pins_arduino.h"
 #include "wiring_private.h"
 
-unsigned char twi_dcount = 18;
-static unsigned char twi_sda, twi_scl;
-
-#define SDA_LOW()   (GPES = (1 << twi_sda)) //Enable SDA (becomes output and since GPO is 0 for the pin, it will pull the line low)
-#define SDA_HIGH()  (GPEC = (1 << twi_sda)) //Disable SDA (becomes input and since it has pullup it will go high)
-#define SDA_READ()  ((GPI & (1 << twi_sda)) != 0)
-#define SCL_LOW()   (GPES = (1 << twi_scl))
-#define SCL_HIGH()  (GPEC = (1 << twi_scl))
-#define SCL_READ()  ((GPI & (1 << twi_scl)) != 0)
+#define SDA_LOW()   (GPES = (1 << twi->sda)) //Enable SDA (becomes output and since GPO is 0 for the pin, it will pull the line low)
+#define SDA_HIGH()  (GPEC = (1 << twi->sda)) //Disable SDA (becomes input and since it has pullup it will go high)
+#define SDA_READ()  ((GPI & (1 << twi->sda)) != 0)
+#define SCL_LOW()   (GPES = (1 << twi->scl))
+#define SCL_HIGH()  (GPEC = (1 << twi->scl))
+#define SCL_READ()  ((GPI & (1 << twi->scl)) != 0)
 
 #ifndef FCPU80
 #define FCPU80 80000000L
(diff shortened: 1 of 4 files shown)


macman88:
I would be interested in this feature as well - It would be very helpful for a project I am working on which ideally will have two independent I2C buses: one for internal devices and one for interchangeable external sensors.

For what it's worth, there are at least two ways of putting multiple identical i2c devices on a single bus.  If your devices have an address pin, you can usually tie that pin to a GPIO and use it as a chip select to select the device you want to talk to.  It doesn't look like the PN532 has an address pin, though.  There also exist I2C bus multiplexers such as the PCA9544A and PCA9546A which allow you to switch between several i2c devices.  In my case this would not really work as I would like to be able to use the "external" i2c bus pins as GPIOs if there is no I2C sensor attached.

hardware/esp8266com/esp8266/libraries/Wire/Wire.cpp
```diff
@@ -45,9 +45,6 @@ uint8_t TwoWire::transmitting = 0;
 void (*TwoWire::user_onRequest)(void);
 void (*TwoWire::user_onReceive)(int);
 
-static int default_sda_pin = SDA;
-static int default_scl_pin = SCL;
```
igrr: Now that these lines are removed, TwoWire::twi.sda and .scl are never initialized.

hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
```diff
@@ -42,35 +39,35 @@ static unsigned char twi_sda, twi_scl;
...
-  else if(freq <= 400000) twi_dcount = 5;//about 400KHz
-  else if(freq <= 500000) twi_dcount = 3;//about 500KHz
-  else if(freq <= 600000) twi_dcount = 2;//about 600KHz
-  else twi_dcount = 1;//about 700KHz
+  if(freq <= 100000) twi->dcount = 32;//about 100KHz
+  else if(freq <= 200000) twi->dcount = 14;//about 200KHz
+  else if(freq <= 300000) twi->dcount = 8;//about 300KHz
+  else if(freq <= 400000) twi->dcount = 5;//about 400KHz
+  else if(freq <= 500000) twi->dcount = 3;//about 500KHz
+  else if(freq <= 600000) twi->dcount = 2;//about 600KHz
```
igrr: Frequencies will likely change given that static `twi_sda`, `twi_scl` and `twi_dcount` are replaced with a pointer argument which needs to be dereferenced. Could you please check how this change affects the clock frequencies?

hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
```diff
@@ -22,15 +22,12 @@
 #include "pins_arduino.h"
 #include "wiring_private.h"
 
-unsigned char twi_dcount = 18;
```
igrr: `twi->dcount` is also not initialized by default in TwoWire class.

hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
paulo-raca: Good point.
How are these timmings measured/calculated?

hardware/esp8266com/esp8266/libraries/Wire/Wire.cpp
paulo-raca: I'll setup "twi" with default values on the default constructor.

paulo-raca:
Thanks for the review, I'll have the comments fixes.

hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_si2c.c
igrr: Set `dcount` to some value, measure the resulting SCL frequency using a logic analyzer or a scope. Repeat with multiple `dcount` values.

me-no-dev:
couldn't this work with just adding a method to replace the pins or call init with the other pins. It's not gonna take much more time and will keep the library clean. I also think those pointers will mess the timing much and there will be no way to achieve the high speeds again.
```cpp
Wire.init(sda1, scl1);
Wire.doThingsWithFirstBus();
.......
Wire.init(sda2, scl2);
Wire.doThingsWithBus2();
```

Issue #167, opened 2015-05-03
Wire broken


av1024:
1. Wire.pins(sda, scl) removed but SDA/SCL defined as `const uint8_t` in pins_arduino.h so can not be redefined (or I don't know how). These changes at least broke the adafruit_bmp085 module
2. Wire.cpp does not compile in e1a6e3d755 (may be si2c.cpp not included somewhere).

---
    Wire/Wire.cpp.o: In function `TwoWire::write(unsigned char)':
    Wire.cpp:(.text+0x130): undefined reference to `twi_init(unsigned char, unsigned char)'
    Wire.cpp:(.text+0x143): undefined reference to `twi_init(unsigned char, unsigned char)'
    Wire/Wire.cpp.o: In function `TwoWire::begin()':
    Wire.cpp:(.text+0x190): undefined reference to `twi_setClock(unsigned int)'
    Wire/Wire.cpp.o: In function `TwoWire::begin(unsigned char)':
    Wire.cpp:(.text+0x19b): undefined reference to `twi_setClock(unsigned int)'
    Wire/Wire.cpp.o: In function `TwoWire::begin(int)':
    Wire.cpp:(.text+0x1a8): undefined reference to `twi_readFrom(unsigned char, unsigned char*, unsigned int, unsigned char)'
    Wire/Wire.cpp.o: In function `TwoWire::setClock(unsigned int)':
    Wire.cpp:(.text+0x1ca): undefined reference to `twi_readFrom(unsigned char, unsigned char*, unsigned int, unsigned char)'
    Wire/Wire.cpp.o: In function `TwoWire::requestFrom(int, int, int)':
    Wire.cpp:(.text+0x29c): undefined reference to `twi_writeTo(unsigned char, unsigned char*, unsigned int, unsigned char)'
    Wire/Wire.cpp.o: In function `TwoWire::beginTransmission(unsigned char)':
    Wire.cpp:(.text+0x2ba): undefined reference to `twi_writeTo(unsigned char, unsigned char*, unsigned int, unsigned char)'
    collect2: error: ld returned 1 exit status
    Error compiling.

igrr:
Should be fixed in 2e3e98b

av1024:
Compilation success but TWI not worked. The test code in BMP085.begin() fails (read8 returns 0xFF instead of 0x55)

1. deprecated `pins()` realization is not valid because does not store passed values and next `begin()` uses pre-defined SDA/SCL constants, not passed to `pins()`
[...]

PR #753, opened 2015-09-06
add ESP8266WebServer::sendContent_P and ESP8266WebServer::send_P with contentLength

martinayotte:
- add ESP8266WebServer::sendContent_P and ESP8266WebServer::send_P with size_t contentLength argument to allow binary content to be send, such as images.
- fix in WiFiClient::write_P to use memcpy instead of memccpy allow binary content to be send, such as images.

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -221,6 +221,35 @@ void ESP8266WebServer::sendContent_P(PGM_P content) {
     }
 }
 
+void ESP8266WebServer::sendContent_P(PGM_P content, size_t size) {
+    char contentUnit[HTTP_DOWNLOAD_UNIT_SIZE + 1];
+
+    contentUnit[HTTP_DOWNLOAD_UNIT_SIZE] = '\0';
+
+    while (content != NULL) {
+        size_t contentUnitLen;
+        PGM_P contentNext;
+
+        // due to the memcpy signature, lots of casts are needed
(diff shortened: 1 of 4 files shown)


hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -169,6 +169,13 @@ void ESP8266WebServer::send_P(int code, PGM_P content_type, PGM_P content) {
     sendContent_P(content);
 }
 
+void ESP8266WebServer::send_P(int code, PGM_P content_type, PGM_P content, size_t contentLength) {
+    String header;
+    _prepareHeader(header, code, String(FPSTR(content_type)).c_str(), contentLength);
```
igrr: Ouch, you are calling `c_str()` method of a temporary String instance here. Better move String declaration into a separate line.

martinayotte:
Hi IGRR, 
I've simply copied the code from Makuna, line #167, so if it need to be changed, it need to be done on both.

igrr:
Oh okay, somehow i missed that. I'll merge everything then and do the necessary changes.

martinayotte:
Hi again Ivan,
I hate putting my hands/arms into a wringer.
during my previous tests, I forgot to add PSTR("image/png") in my calls.
Adding that, I got some crashes, even in Makuna original code.
So, I conclude that original code isn't working : 

prepareHeader(header, code, String(FPSTR(content_type)).c_str(), contentLength);

Changing that to what you suggest me doesn't work either : 

String type = String(FPSTR(content_type));
_prepareHeader(header, code, type.c_str(), contentLength);

So, I've think the proper way is again to use memccpy_P(), but I feel it is a bit ugly : 

char type[64];
memccpy_P((void*)type, (PGM_VOID_P)content_type, 0, sizeof(type));
_prepareHeader(header, code, (const char* )type, contentLength);

I will commit that for now, but I think it needs to be revisited...
I will appreciate you help here !

PR #759, opened 2015-09-07
optimized board view, robustness of NTPclient enhanced

holgerlembke:
- visualization of INPUT_PULLUP and INPUT_PULLDOWN_16 in esp12 board
view
- added handling of failing hostname resolving to NTPclient
- added handling of non working udp receiver to NTPclient


hardware/esp8266com/esp8266/libraries/ESP8266WiFi/examples/NTPClient/NTPClient.ino
```diff
@@ -67,20 +67,39 @@ void setup()
...
   // wait to see if a reply is available
+  udperrorcounter++;
   delay(1000);
-  
+
   int cb = udp.parsePacket();
   if (!cb) {
     Serial.println("no packet yet");
+
+    if (udperrorcounter > 3) { // 3 is just randomly choosen. in case of lot packet drops...
```
Links2004: you can simply use ```if(udp)``` to check if the socket is still active and is not closed by WiFi or any other event.
using ```udp.stopAll()``` may also is ab bad idea in case of having more then one function using UDP like.
NTP + Network update.

```cpp
if(!udp) {
    udp.begin(localPort);
}
```
shut do the job just fine.

holgerlembke: I don't get it.

Why again stopall is a bad idea, when udp has stopped working?

Links2004: based on the error it can be the case that only the UDP socket for NTP is closed, but other sockets can still be alive. so with ```udp.stopAll()``` you will kill them all.

the bool compare is overloaded for the WiFiUdp class:
https://github.com/esp8266/Arduino/blob/6422688cc61115b20cdd7fb32be97cb296f977e2/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/WiFiUdp.h#L42
and it will check if exactly this socket it alive or not.

holgerlembke:
I see all this things, too. And they make the code so much easier.

But for me the question remains on the end user side: is all that trustable? Or perhaps add another level of diagnostic. My suggestion will detect any packets lost/not arriving/whatever for any reason. 

The "if (!udp)" only will check for problems on the esp8266 side and then only if the implementation from Espressif and all following is right.

So replacing stopall() with a stop() is a good idea. For the remains I'm sceptical.

igrr:
`if (!udp)` doesn't call into Espressif code. This is simply a check the there is a UDP pcb assigned to the WiFiUDP instance. 

Unrealted nitpick: would you mind uploading the modified .svg file as well as .png?

holgerlembke:
So igrr, what are the implication of your answer? Sounds to me like don't use if (!udp) and do a stop()/begin() anyway? 

For the png: I worked with the pixels in the png, so svg is unchanged. It was only intended for my local use in company training.... but I'll look into moving the changes to that file.

igrr:
That's fine to use `if (!udp)` to detect that the socket was killed due to WiFi disconnect. I was just explaining what it does.
I would like to see the following improvements:

- Simplify error handling (this is a sample, don't make it more complicated than it needs to be). Use single `if (!udp)` at the beginning of the loop to check if you need to call `udp.begin` again. Remove `udperrorcounter` since it doesn't really improve readability.
- Move `WiFi.hostByName` call into `setup`.
- Fix indentation (line 98).
- If you plan to change the picture, do the changes in svg and export to png.

holgerlembke:
As being a sample it should teach usage. 

And without  udperrorcounter   there is no way to see whether it works or not. No way to detect possible errors in communication. I think it is important to have.

I just moved the png changes to the svg.

igrr:
I think that would make a great case for another sample, like FailsafeUDP or RobustUDPCommunication. 
I don't think mixing different concepts in a single example sketch is what we need to make this environment more accessible to beginners. Most of our example sketches are really complicated, let's try to decouple things into separate examples.
The same applies to WebServer examples, these really deserve some refactoring...

holgerlembke:
sounds like a good plan to follow.

PR #762, opened 2015-09-08
for project

prshah25:
taking for project


Links2004:
I think here is something going totally wrong, from my point of view it make no sense to merge this to https://github.com/esp8266/Arduino.
what is the goal of this merge ?

igrr:
wrong branch apparently.

PR #771, opened 2015-09-11
fix bug: wrong file path from request uri

pgollor:
We don't need the extra `path` parameter for `ESP8266WebServer::serveStatic` anymore, do we?


pgollor:
No longer needed since commit:
https://github.com/esp8266/Arduino/commit/1f8c14d8c21818a2cee76a3de3954ebd551cebd6

PR #781, opened 2015-09-14
Changes needed for upload with esptool in Eclipse

lrmoreno007:


pgollor:
I use the same workaround. The `-vv` option fails on eclipse.
But there is no need to add this for each board. You can change this line:
https://github.com/lrmoreno007/Arduino/blob/esp8266/hardware/esp8266com/esp8266/platform.txt#L93

igrr:
Why exactly does `-vv` option fail on eclipse? Is there any error message? Does `-v -v` work?

pgollor:
Yes `-v` works very well. But i can't remember the error message. It fails to upload the scetch if you use the arduino plugin within an existing eclipse installation. I tested it on Ubuntu 14.04 LTS.
The pre build eclipse arduino version works better.
So i am not fully sure where the problem is. But i think `-v` works also good enough.

igrr:
So it would be okay to change `-vv` to `-v -v` in platform.txt?

lrmoreno007:
With  -vv Eclipse call to esptool...
C:\Program Files (x86)\arduino\hardware\esp8266com\esp8266/tools/esptool.exe false -cd ck -cb 921600 -cp COM2 -ca 0x00000 -cf C:\Eclipse\workspace\xxx/Release/xxx.bin

The problem is Eclipse don't use -vv and write "false" and then esptool fail. I don't know because do it.

-v run OK.

lrmoreno007:
I'm going to try it in platform.txt

lrmoreno007:
I think I found the problem, in platform.txt the line:
tools.esptool.upload.pattern="{path}/{cmd}" {upload.verbose} -cd {upload.resetmethod} -cb {upload.speed} -cp "{serial.port}" -ca 0x00000 -cf "{build.path}/{build.project_name}.bin"

Must be:
tools.esptool.upload.pattern="{path}/{cmd}" {upload.params.verbose} -cd {upload.resetmethod} -cb {upload.speed} -cp "{serial.port}" -ca 0x00000 -cf "{build.path}/{build.project_name}.bin"

"upload.verbose" is not defined, and then Eclipse put "false"

With this change work perfectly with -vv and -v.

lrmoreno007:
Sorry, changing to "upload.params.verbose" everithing run OK, but have no one parameter in the command line.

We must found what parameter is set to false.

igrr:
Did you try changing tools.esptool.upload.params.verbose as follows?
```
tools.esptool.upload.params.verbose=-v -v
```

lrmoreno007:
yes, false again

lrmoreno007:
Ok, Eclipse is getting the upload.verbose parameter from preferences.txt in C:\Users\<User>\AppData\Roaming\Arduino15

I change it to true, reload entirely the project in preferences of the project and....
image2

I don't know how to solve it. But with my commit is "solved".

lrmoreno007:
Uffff now don't work in Arduino native...

image1

No upload

I don't know how to solve it.

lrmoreno007:
Adding in platform.txt only:
upload.verbose=-v

Are working in both platforms.

igrr:
But with one `-v` there will be less debug output than with `-v -v` or `-v`...

lrmoreno007:
Run OK with -vv and with -v and with -v -v

I like with -v only but you decide.

Thanks

lrmoreno007:
Changed to -vv.

Please merge it.

igrr:
Since you're changing `tools.esptool.upload.verbose` to `upload.verbose`, can this affect other upload tools?

jantje:
Jantje here
The problem is that "verbose" parameters should be optional. From the discussion it seems to me the verbose part is not optional.
If -vv is not optional for the upload I advise to add -vv to the upload pattern.

From the plugin plugin point of view verbose is not (yet?) supported. Which explains why it is not picked up.

lrmoreno007:
Now I make one menu for every board, and set in everyone the verbose option.

This run OK and it's particularised the verbose -vv for esptool only.

lrmoreno007:
Thanks for merge

PR #794, opened 2015-09-16
Adding getMode() function to WiFi library so I can easily get the current mode of the wifi.

paulmand3l:


Links2004:
you need to do ```return (WiFiMode )wifi_get_opmode();```

PR #802, opened 2015-09-18
Flash SPIFFS over the air (OTA)

pgollor:
I updated the `espota.py`, the `UpdaterClass` and added an example sketch: `OTA-MDNS-SPIFFS`

I did some tests with Ubuntu 14.04 (python version 2.7) and a generic ESP8266-01 module with 1M64k:
- Flashing sketch over console with `espota.py`
- Flashing sketch with the current ESP8622 Arduino version.
- Flashing SPIFFS image over console with `espota.py`

All test work successfully but it has to be tested with another OS and someone has to be update the esp8266fs.jar to upgrade the SPIFFS image over the air if this options is active.

`espota.py` changes:
- Add option parser.
- Add logging.
- Send command to controller to differ between flash image and SPIFFS image.

Please feel free to post your test results here. Tanks!


igrr:
May I suggest to update the example to acknowledge the fact that SDK will automatically reconnect to the last used WiFi network? We only need to call WiFi.begin() if there was a change to SSID/password. Otherwise this leads to rewriting SDK config sectors on every boot.

pgollor:
Your idea is great but i had to add two functions to the ESP8266WiFi class:
- `begin()`: To start connecting without new parameters.
- `psk()` : To read the pre shared key from sdk configuration.

I tried it multiple times without calling `WiFi.begin(...)`. But `WiFi.status()` returns always `WL_IDLE_STATUS` (0) and the esp8266 didn't connect to the AP are stored in the sdk configuration. I checked the sdk config with `WiFi.printDiag(Serial);` for right parameters.

So I hope that you are able to understand my explanation in english ;) and you are to agree with my solution?

igrr:
Adding `begin()` and `psk()` is a good idea. This has been requested in other tickets.
I think we need to look at the overall logic of WiFi class, i.e. how switching between STA/AP modes should work. But it is a separate task...
I will merge this pull request now, and let's update the sample once these new functions are added.

igrr:
It might be worth changing `psk` (and SSID()) function to return a `String` rather than `char*`, because I think users might make the following error:

```c++
const char* ssid = "adasdasd";
const char* psk = "23801380123";
...
if (WiFi.ssid() != ssid || WiFi.psk() != psk) { ... }
```
I.e. this will result in pointer comparison rather than string comparison.

chrisfraser:
I agree. Needing to do a `strcmp` is going to confuse some users.
    if (strcmp(WiFi.ssid(), ssid) + strcmp(WiFi. psk(), psk) == 0){ ... }

How about internally doing this check in `begin(ssid, psk)` before rewriting the sector?

pgollor:
i will add string return functions in a few days.

Issue #830, opened 2015-09-28
Client closing an HTTP form upload crashes the server.


mangelajo:
I have found this while testing the web update example, if you cancel the upload of the new binary from server, and then you try to upload a new one, you get the connection open and accepted, but then it crashes on:

LmacRxBlk:1
LmacRxBlk:1
LmacRxBlk:1
LmacRxBlk:1
LmacRxBlk:1
LmacRxBlk:1
LmacRxBlk:1
LmacRxBlk:1

I suspect the loop on ESP8266WebServer::_parseForm is not handling client disconnection, I tried to add a if (!client.connected()) return false inside the while(1) loop without success...

mangelajo:
Hmm, ok, the reason is that https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing.cpp#L277  uploadReadByte will lock forever in the while loop.

We should check when client.connected() goes false, and then send a UPLOAD_FILE_ABORTED or similar to the handler, so the app would be able to cleanup.

Links2004:
can you try: 
```cpp
uint8_t ESP8266WebServer::_uploadReadByte(WiFiClient& client){
  int res = client.read();
  if(res == -1){
    while(!client.available()) {
      yield();
      if(!client.connected()) {
          break;
      }
    }
    res = client.read();
  }
  return (uint8_t)res;
}
```

mangelajo:
I will send you a patch, that works, but also you have to make the loop on top of this (_parseForm) check client.connected() and exit when necessary.

mangelajo:
I sent a pull request with the patch. It's been manually tested.

PR #833, opened 2015-09-28
Handle form upload abortion in ESP8266WebServer

mangelajo:
If the client connection gets closed during a form upload,
the fileUploadHandler is notified with the new
UPLOAD_FILE_ABORTED status, and the loop is ended
gracefully.

Fixes issue #830


mangelajo:
Ouch, I got it added to the same PR, I didn't want that. Sorry.

PR #837, opened 2015-09-29
ESP8266HTTPUpdateServer extracted from WebUpdate example.

mangelajo:
It depends on issue #833 fix.


hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
```diff
@@ -0,0 +1,88 @@
+#include <Arduino.h>
+#include <WifiClient.h>
+#include <WifiServer.h>
+#include <ESP8266WebServer.h>
+#include <WifiUDP.h>
+#include "ESP8266HTTPUpdateServer.h"
+
+
+const char* ESP8266HTTPUpdateServer::_serverIndex = "<html><body><form method='POST' action='/update' enctype='multipart/form-data'>"
```
igrr: I would suggest to store this HTML as a [raw string literal](http://en.cppreference.com/w/cpp/language/string_literal)

hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+  _serial_output = serial_debug;
+  _server = NULL;
+  _port = port;
+}
+
+ESP8266HTTPUpdateServer::~ESP8266HTTPUpdateServer()
+{
+  if (_server)
+    delete _server;
```
igrr: This doesn't look right in case when ESP8266WebServer instance was allocated statically outside this class. Consider the following code:
```c++
void runServer() {
  ESP8266WebServer httpServer(80);
  ESP8266HTTPUpdateServer updateServer;
  updateServer.setup(&httpServer);
  while (something) {
    httpServer.handleClient();
  }
}
```
Upon exit from this function, ESP8266WebServer will be deleted twice.

igrr:
Another thought is that it would probably make sense to pass ESP8266WebServer instance into web server callbacks — this way you would not need to store a _server pointer in your class. This would allow for pluggable "middleware" classes like in node.js. I'll take a look if we can achieve this without breaking a lot of existing sketches.

hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
mangelajo: Good catch, also like this, we make the dependency explicit, and make sure the user will import the ESP8266WebServer from the main .ino, otherwise, it won't get included/linked since arduino libs can't set dependencies AFAIK.

mangelajo:
Thanks Ivan!, will do.

Yeah, the passing of server itself via the events would be nice, it's generally a good thing to provide the source of the event. I wonder if there's some way to do it without breaking existing sketches.

May be by having ::on overloaded with the different possible callbacks.

hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
igrr: Btw, nightly builds of IDE already have library dependency tracking!

mangelajo:
Hold on, the example is gone away

hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
mangelajo: Hmm, sorry, I missed that comment. It seems that My C++ is outdated :) , sounds good to me, will do.

hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
mangelajo: Ohh, nice :)

mangelajo:
Any idea of why could it be failing like this: 

/home/travis/build/esp8266/Arduino/build/linux/work/hardware/esp8266com/esp8266/libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp:2:24: fatal error: WifiClient.h: No such file or directory
 #include <WifiClient.h>

?

It does compile locally

mangelajo:
What do you think about renaming the lib to ESP8266HTTPUpdateApp or ESP8266HTTPUpdateHandler,
since now, we make it clear that we register to an existing webserver.

mangelajo:
I don't see the gain of the raw string literal exactly:

I tried: 

  0 const char* ESP8266HTTPUpdateServer::_serverIndex =
  1         R"(<html><body><form method='POST' action='/update' enctype='multipart/form-data'>
  2                                           <input type='file' name='update'>
  3                                           <input type='submit' value='Update'>
  4                                       </form>
  5                                 </body></html>)";
  6

mangelajo:
@igrr , any idea of why the travis WebUpdater.ino compilation could be failing?, It executed it locally the same way, and it does provide the -I path to WifiClient :-?

igrr:
In ESP8266HTTPUpdateServer.cpp, you need to observe the case in include names, i.e. WiFiClient instead of WifiClient, and others as well. Travis CI runs on Linux, so file names are case sensitive.

mangelajo:
arghhh!, damn case insensitive OSX default, thanks ...

mangelajo:
@iggr I could also do something like this for the DNS_SD_Arduino_OTA if you think that's valuable, and I'd be willing in testing it for robustness, recovery, and all kind of evil stuff we could want to do to it.

igrr:
Encapsulating all the update logic into a nice ArduinoOTA library would be an improvement I guess. I don't think anyone is working on it, so if you can do it, that's great.

igrr:
@mangelajo As discussed, i have made a change to ESP8266WebServer to allow writing more self-sufficient handlers.
Here's the sample: https://gist.github.com/igrr/0da0c4adc7588d9bd911
I'm still not certain on the ownership transfer. Right now user creates a handler with `new` and gives it to the server, which will delete the instance when required. Making the handler owned by user is also possible, but i'm not sure how to distinguish these cases better.

PR #842, opened 2015-09-30
[FIX]Disconnecting STA when turning off SOFTAP 

MrWolvetech:
This patch fixes the issue that when disconnecting/disabling SOFTAP it also disconnects the STATION
Already tested and worked.


hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/ESP8266WiFi.h
```diff
@@ -113,7 +113,7 @@ class ESP8266WiFiClass
       *
       * return: one value of wl_status_t enum
       */
-    int softAPdisconnect(bool wifioff = false);
```
igrr: Changing this default value will likely change behaviour of existing sketches. Is this change really required?

PR #846, opened 2015-10-01
Handle form upload abortion in ESP8266WebServer to fix Issue #833

mangelajo:
If the client connection gets closed during a form upload,
the fileUploadHandler is notified with the new
UPLOAD_FILE_ABORTED status, and the loop is ended
gracefully.


mangelajo:
I re-created this PR from a different branch that esp8266, as I was pushing -f to it, and breaking my PR.

PR #853, opened 2015-10-04
avoid to dereference a NULL pointer if update is called before begin

mpromonet:
Hi,
This pull request is about ESP8266mDNS, it add in the `update` method a check that _conn is not NULL before using it.
It allows to have a loop that always call it even when ESP8266mDNS is not still initialized.
Regards,
Michel.


igrr:
Actually `update` should not be called from the loop any more. I think all the library examples reflect that.

PR #873, opened 2015-10-09
fix possible free of null in String

Links2004:


igrr:
The free() function deallocates the memory allocation pointed to by ptr. If ptr is a NULL pointer, no
     operation is performed.

Links2004:
it will generate an error message when ```setDebugOutput``` is active.

pgollor:
Sorry for the question. But why we are using malloc and free and not new and delete. Malloc is for c and new is the better way in CPP to allocate memory.
Or am i wrong?

Links2004:
and ```new``` and ```delete``` are implemented with ```malloc``` and ```free``` ;)
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/cores/esp8266/abi.cpp#L28

pgollor:
thanks @Links2004 :)

mangelajo:
+1, let's avoid unnecessary error messages in debug if possible.

Issue #795, opened 2015-09-16
+full HEADER in ESP8266WebServer


xbary:
Hi

I would ask to be added to the official version of such an allowance:

-----------------------------------------------------------------------------------------------------
in file ESP8266WebServer.h (on comment // Add by XBary)
...
  WiFiClient client() { return _currentClient; }
  HTTPUpload& upload() { return _currentUpload; }
  String HEADER() { return _currentHEADER; } // Add by XBary

  String arg(const char* name);   // get request argument value by name
...
and
...
HTTPMethod  _currentMethod;
  String      _currentUri;
  String      _currentHEADER; // Add by XBary

  size_t           _currentArgCount;
  RequestArgument* _currentArgs;
...
-----------------------------------------------------------------------------------------------------
in file Parsing.cpp
...
  String req = client.readStringUntil('\r');
  client.readStringUntil('\n');
  _currentHEADER = req + "\r\n"; // Add by XBary

  // First line of HTTP request looks like "GET /path HTTP/1.1"
  // Retrieve the "/path" part by finding the spaces
...
and
...
      headerName = req.substring(0, headerDiv);
      headerValue = req.substring(headerDiv + 2);
	_currentHEADER += (req + "\r\n"); // Add by XBary
	  
	  #ifdef DEBUG
	  DEBUG_OUTPUT.print("headerName: ");
...
and
...
      if (headerDiv == -1){
        break;
      }
      headerName = req.substring(0, headerDiv);
      headerValue = req.substring(headerDiv + 2);
	 _currentHEADER += (req + "\r\n"); // Add XBary
	  
	  #ifdef DEBUG
	  DEBUG_OUTPUT.print("headerName: ");
	  DEBUG_OUTPUT.println(headerName);
	  DEBUG_OUTPUT.print("headerValue: ");
...
-----------------------------------------------------------------------------------------------------

Example use:
...
void handle_mainmenu()
{
	WiFiClient client = server.client();
	String hostheader = server.HEADER();
	int sesID = GetSession(client, hostheader);
...

The application mainmenu now have access to the entire HTTP header. I read User-agent: or cookies, etc. 

Regards XBary

brianensor:
My vote (if I get one) is to make headers accessible in the same way as arguments. Then one could do something like server.header("Cookie") to get the value. I've coded this up reusing RequestArgument and it's working great. I'll submit a pull request later today or tomorrow. Thanks for the great project!

xbary:
[...]

PR #886, opened 2015-10-13
Add header access using same method as arguments

brianensor:
Here's my fix for issue #795 which adds access to request headers by the same method that arguments are accessed. I needed to access the Cookie header and now I can do that like this.

```C
if (server.hasHeader("Cookie")) {
    String cookieValue = server.header("Cookie");
}
```

### DIFF
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -41,6 +41,10 @@ ESP8266WebServer::ESP8266WebServer(int port)
 }
 
 ESP8266WebServer::~ESP8266WebServer() {
+  if (_headerKeys) {
+    free(_headerKeys);
+    _headerKeys = 0;
+  }
   if (!_firstHandler)
     return;
   RequestHandler* handler = _firstHandler;
@@ -289,6 +293,14 @@ String ESP8266WebServer::header(const char* name) {
       return _currentHeaders[i].value;
   }
   return String();
+}
+
+void ESP8266WebServer::collectHeaders(const char* headerKeys[], const size_t headerKeysCount) {
+  _headerKeysCount = headerKeysCount;
+  _headerKeys = (const char**)malloc(_headerKeysCount*sizeof(char*));
+  for (int i = 0; i < _headerKeysCount; i++){
+    _headerKeys[i] = headerKeys[i];
+  }
 }
 
 String ESP8266WebServer::header(int i) {
--- a/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -283,6 +283,38 @@ bool ESP8266WebServer::hasArg(const char* name) {
   return false;
 }
 
+String ESP8266WebServer::header(const char* name) {
+  for (int i = 0; i < _currentHeaderCount; ++i) {
+    if (_currentHeaders[i].key == name)
+      return _currentHeaders[i].value;
+  }
+  return String();
+}
+
+String ESP8266WebServer::header(int i) {
+  if (i < _currentHeaderCount)
+    return _currentHeaders[i].value;
+  return String();
+}
+
+String ESP8266WebServer::headerName(int i) {
+  if (i < _currentHeaderCount)
+    return _currentHeaders[i].key;
+  return String();
+}
+
+int ESP8266WebServer::headers() {
+  return _currentHeaderCount;
+}
+
+bool ESP8266WebServer::hasHeader(const char* name) {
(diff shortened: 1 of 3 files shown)


drewfish:
I worry about the memory and CPU overhead, specifically if I never use this new API. Ideally `hasHeader()` and `header()` would only create that internal object if/when these functions are called, that way those folks who don't use them don't have to pay the cost.

brianensor:
@drewfish, that's definitely a valid concern and I think I have a good solution to it. I'll update the code later today once I get everything working.

brianensor:
I've updated the method of collecting request headers so that it is off by default in response to @drewfish's resource concerns. Now if I want to have access to the User-Agent and Cookie headers for example...

```C
const size_t headerKeysCount = 2;
const char* headerKeys[headerKeysCount] = {"User-Agent", "Cookie"};
server.collectHeaders(headerKeys, headerKeysCount);
```

The methods to access the collected headers remain the same.

hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -41,6 +41,10 @@ ESP8266WebServer::ESP8266WebServer(int port)
 }
 
 ESP8266WebServer::~ESP8266WebServer() {
+  if (_headerKeys) {
+    free(_headerKeys);
+    _headerKeys = 0;
```
drewfish: Do we also maybe want to do `_headerKeysCount = 0` just to be safe?

brianensor: I doubt it because the object is being destroyed anyway but why not, can't be too safe. Does the rest of the PR work for you?

hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -83,6 +83,13 @@ class ESP8266WebServer
   String argName(int i);          // get request argument name by number
   int args();                     // get arguments count
   bool hasArg(const char* name);  // check if argument exists
+  
+  void collectHeaders(const char* headerKeys[], const size_t headerKeysCount); // set the request headers to collect
+  String header(const char* name);   // get request header value by name
```
drewfish: Will this end up copying the returned String (or will the compiler optimize)? Is there a "best practice" (examples in other well-used libraries) on how to return a String efficiently?

(Am I focusing _too_ much on optimizations :smile: :disappointed: ?)

drewfish: For example, would `const String&` be a better return type? (Will be tricky to handle the "not found" case, but we could make a single empty string to use for all "not found" return values.)

drewfish:
I like the approach. I'm uncertain about fine-tuning the performance, since I'm not that experienced with microprocessor (and specifically esp8266) programming.

hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.h
brianensor: If you look just above (ESP8266WebServer.h Line 81), you'll see that the existing argument access method is the same. There may be a better way but I decided to follow the conventions of the existing code.

drewfish: OK, cool, good enough for me.

drewfish:
+1 from me

igrr:
Looks mostly good to me. Few comments from my side:
- adding a usage example would be great
- who is supposed to own _headerKeys? The calling code or the ESP8266WebServer?
- let's use new/delete[] instead of malloc/free in new code

luc-github:
Hi looks very nice and I am really willing to use it ( I need a cookie :smile: )
any time frame for modification/merging ?

luc-github:
I would be happy to help doing modification - I just do not want overlap 
also it won't be better instead of free / allocate the header array at each page to allocate one for all ?
So we could directly copy the requested keys in _currentHeaders[i].key at the initialization and just set  _currentHeaders[i].value="" for each reset, instead of delete all array and allocate it again.
because _collectHeader(headerName.c_str()) already look if it requested - so instead of boolean answer just give the position in _currentHeaders[] to put the value or -1 if not requested

so no need to allocate the Keys array and no need to delete/allocate the header array at each page

luc-github:
@brianensor @igrr I have done modifications I mentioned and which match @igrr request, and did a sample using header for simple authentication using cookie (login page / disconnection/etc...) - if no comment against, I will do a PR tonight, I hope it will be fine, if not I will adjust it

igrr:
@luc-github please do, this should be merged before Friday.

luc-github:
Ok just did https://github.com/esp8266/Arduino/pull/933
If anything need to be changed, I will do

PR #899, opened 2015-10-18
update SPIFFS to version 0.3.3

pgollor:
In a new staging version of this, we have to use the latest mkspiffs version with SPIFFS 0.3.3 which includes the pull request https://github.com/igrr/mkspiffs/pull/10


igrr:
Thanks! Will do some tests and merge.
It's nice that all c-string arguments are now `const`, so I can remove this stuff from the wrapper:
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/cores/esp8266/spiffs_api.cpp#L66-L69 and elsewhere...

pgollor:
okay. Thanks for your great work.

PR #911, opened 2015-10-21
Add _buffer initialization and some sanity check

luc-github:
This seems fix the Exception (28)/ (9) that occur on Captive portal with
Soft AP


igrr:
Could you please fix line indentation in lines 120-173? Or even better, replace if(buffer != NULL) {  whole function body } with `if (buffer == NULL) { return; }`.

luc-github:
Yes you right will be better - will do

PR #933, opened 2015-10-28
Add header access using same method as arguments 2

luc-github:
based on @brianensor PR
+sample and some sanity check


luc-github:
if any change is needed - let me know - I will adjust

igrr:
Thank you

xbary:
Thanks for that functionality :) I like

Issue #859, opened 2015-10-06
Support for exporting sketches.


igrr:
Using `Sketch > Export compiled binary` menu causes an error:

    Warning: This core does not support exporting sketches. 
    Please consider upgrading it or contacting its author

PR #934, opened 2015-10-28
Support Sketch > Export compiled binary

cpacey:


cpacey:
Should resolve  #859.

I'll admit I don't love the output file name, but it's consistent with the Arduino boards.

PR #952, opened 2015-11-01
Add SVG as file type and default to binary stream for static handler

KaloNK:


grahamehorner:
would in not be better to have this changed to determine from a array of key value pairs; rather than a list of if statements; the key value pairs could also be populated via user code or an external data (json) file. I feel that the hardcoding of file extensions/meta types isn't in code isn't good.

igrr:
@grahamehorner If you want to customize this list, you can basically write your own version of [StaticRequestHandler](https://github.com/esp8266/Arduino/blob/20de6b8a6a3c322737b2c79b2886191adc18b108/libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h#L32-L93) and pass it to ESP8266WebServer::addHandler method.

grahamehorner:
@igrr I can foresee a lot of custom StaticRequestHandler implementions not everyone will need or want these items in the list; it would be better to have perhaps to create a callback which implements the getContentType and have the ability to replace the default with developer code?

igrr:
Yes it's possible to add such a callback as an optional argument to ESP8266WebServer::serveStatic method.

grahamehorner:
@igrr ok, missed that one

PR #961, opened 2015-11-03
Update ArduinoOTA.cpp

sticilface:
Restarts the UDP port if it becomes disconnected. Works when for example the wifi is disconnected and reconnected

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -51,6 +51,11 @@ void ArduinoOTA::setup() {
 }
 
 void ArduinoOTA::handle() {
+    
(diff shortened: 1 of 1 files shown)


libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -51,6 +51,11 @@ void ArduinoOTA::setup() {
 }
 
 void ArduinoOTA::handle() {
+    
+  if (!*_udp_ota) {
+    _udp_ota->begin(_port); 
+    Serial.println(F("OTA restarted"));
```
Links2004: please add ```if (_serial_debug)```.
serial may be used for other thinks then debug ;)

PR #976, opened 2015-11-06
Fix FAILD typo

marvinroger:
@Links2004 :)


Links2004:
thanks ;)

pgollor:
Thanks. The fix should be complete.
`grep -ri "faild" Arduino/*` returns nothing.

But should we leave the "wrong" typos for backward compatibility with older sketches?

igrr:
I think that would be a relatively minor issue.

PR #980, opened 2015-11-07
IDE OTA Upload changes

me-no-dev:
* added md5.h and MD5Builder with access to the ROM md5 functions
* added DIGESTish_MD5 authentication to espota.py and ArduinoOTA
* added MD5 check of the received image in Updater
* added extra fields to mDNS to enable IDE OTA for when/if our changes are accepted by the Arduino team
* add/changed recipe for network upload through the IDE for when/if our changes are accepted by the Arduino team
* big overhaul to ArduinoOTA to facilitate all necessary authentication and sanity checks
* fixed relevant example sketches

my Arduino IDE branch with IDE Upload included: https://github.com/me-no-dev/Arduino-1/tree/esp8266-ota


eliabieri:
did this fix the error, where the arduino IDE always wanted a password?

-- 

Elia Bieri
whateveryouwant@eliabieri.com

me-no-dev:
no. This fixes everything else and adds security
when/if the IDE changes are accepted then upload will work as it is supposed to and ask for password only if you have secured OTA running.
The reason you get asked for password is because by default only Yuns use the network and they are SSH secured

sticilface:
The new OTA, defines 'extern ArduinoOTAClass ArduinoOTA;' 
Im working on a settings manager that allowed you to start and stop the OTA with the flick of a switch.  say you have lots of ESPs... This used new and delete to create the object dynamically.  my knowledge of c++ is not good enough to know if 1) you can override that or  2) if i used  
```
       ota_server = new ArduinoOTAClass; 
```

would that effectively mean i have two objects ota_server & ArduinoOTA in memory?  i.e. a waste or resources?

igrr:
Today I was actually thinking of reverting this change — I don't think we
should have ArduinoOTA instance allocated by default.

The pattern of statically allocated instances works well when these
instances map to hardware resources (e.g Serial and SPI). For things such
as network services this pattern is less applicable, and this becomes
obvious when application complexity passes the level of "I only initialize
things once".

sticilface:
Ah ok.  Good to know. thanks

For future reference having two objects, even if one is never used double the memory usage?

me-no-dev:
but looked at it differently, you can have only one ArduinoOTA and you allocate it when you include the header

me-no-dev:
So even though it's not directly "hardware" related, it still can exist only once. Any other situation will lead to unknown results. You can not have more than one UDP listener on a particular port and even if you could have more, you could not know which one will get the packet and so on. Therefore I strongly believe that in this case a global object should be used (I think I even made it so it does not init twice).
Surely a method to stop the OTA can be added, but what will you actually stop? MDNS, the UDP or?
This is the only reason I made it global: it can have only one instance running at any given time.

sticilface:
I agree, only one instance running at a time.  My two cents is that i've now got ~ 9 of these in the home, more if u count the ones not running the arduino code... so it means i've got a lot of OTA stuff going on, in that menu.  My settings manager allowed me to turn off OTA by deleting a dynamically created OTA object. (  i also see your point looking at the MDNS stuff that there is no destructor , or remove service options, once it is added).  This meant I can just have the unstable / projects I'm working on show up.  means I'm less likely to go and flash the wrong device.   I have to go enable it, then flash it.  

so +1 for stop method.  New lib is ace btw.  much better!

PR #995, opened 2015-11-09
Updates for time.c

Juppit:
Some fixes for time.c
declare timezone as long
call sntp_setservername() with server id


igrr:
Btw, on esp8266 sizeof(int) == sizeof(long) :)

Issue #480, opened 2015-06-28
Different circuit neeeded for automatic upload using CP210x


probonopd:
[Comment shortened in order not to confuse people -- see https://github.com/esp8266/Arduino/issues/480#issuecomment-122993574 for the recommended circuit]

I can reliably upload using baud rate 921600.

It would be good if others could test this setup too and if it works for everyone, it could be changed in the README.

Links2004:
the problem is the Arduino IDE or you Serial terminal.

is the boot loading still working with the 10k?
i don't think its good for the stability of the uploading.

the problem shut be fixed in software (Arduino IDE) so that DTR and RTS is not activate in our case.
i use "real term" for serial debugging there i can control DTR and RTS manually.

probonopd:
Yes 921600 is working fine for me.

cimba007:
I thing I have the same problem here. What are your values for "pull up and pull down"? I am using 10k and only get ~2,2V on my gpio0, resulting from 3,3V high on DTR as it is not in use and a 10k pulldown but ofc. this is not enough if DTR got some decent amps I guess.

probonopd:
10K

cimba007:
What are u doing with GPIO0 and RESET when the CP210x  is not connected? What values are you using in this case?

probonopd:
Good point. In my case the CP210x is always connected and is also used to provide the 3.3V (which appears to work in my case).

Links2004:
https://github.com/esp8266/Arduino/blob/esp8266-sdk-1.0/README.md#minimal-hardware-setup-for-running-only

probonopd:
Following advice from igrr, I am now using the circuitry from the NodeMCU 1.0 for my ESP-12Es. See schematics of the NodeMCU 1.0.

Two 2N2222 npn transistors (10K) and 3 resistors, thereof 1 pullup for CH_PD. Working well.This circuit can be used for flashing and running, it does auto-reset and the serial monitor in the IDE can be used.

Possibly the documentation in the README should be updated to reflect this.

JeroenBeemster:
Thanks probonopd, This solved my issue aswell.For everyone who find it hard to read electronic schematics. I created a simplified example using a breadboard explanation.
http://www.arduinesp.com/blink

probonopd:
Thanks JeroenBeemster, I hope you are not offended if I attach your example for quick reference here.

[...]

PR #1006, opened 2015-11-12
add menu "Reset Method"

volca:
I added the "Reset Method" for some custom esp8266 modules "nodemcu like" but has only 512K or 1M flash.


asetyde:
Reset ? There are a ESP.reset();  method yet ?

volca:
I think "Auto Reset Method" is better.  Please see the issue #480

esp8266

PR #1008, opened 2015-11-12
Protect transactions with flag

KaloNK:
With soft CS and when called from interrupt, to protect the transaction it is not enough to just wait for the transfer to end.
Added a flag and allow other tasks to run while we wait to begin transaction

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -37,6 +37,7 @@ SPIClass SPI;
 
 SPIClass::SPIClass() {
     useHwCs = false;
+    inTransaction = false;
 }
 
 void SPIClass::begin() {
@@ -74,13 +75,17 @@ void SPIClass::setHwCs(bool use) {
 }
 
 void SPIClass::beginTransaction(SPISettings settings) {
-    while(SPI1CMD & SPIBUSY) {}
+    while(inTransaction || SPI1CMD & SPIBUSY) {
+        yield();
+    }
+    inTransaction = true;
     setFrequency(settings._clock);
     setBitOrder(settings._bitOrder);
     setDataMode(settings._dataMode);
 }
 
 void SPIClass::endTransaction() {
+    inTransaction = false;
 }
 
 void SPIClass::setDataMode(uint8_t dataMode) {
--- a/libraries/SPI/SPI.h
+++ b/libraries/SPI/SPI.h
@@ -74,6 +74,7 @@ public:
   void endTransaction(void);
 private:
   bool useHwCs;
+  bool inTransaction;
   void writeBytes_(uint8_t * data, uint8_t size);
   void writePattern_(uint8_t * data, uint8_t size, uint8_t repeat);
   void transferBytes_(uint8_t * out, uint8_t * in, uint8_t size);


libraries/SPI/SPI.cpp
```diff
@@ -74,13 +75,17 @@ void SPIClass::setHwCs(bool use) {
 }
 
 void SPIClass::beginTransaction(SPISettings settings) {
-    while(SPI1CMD & SPIBUSY) {}
+    while(inTransaction || SPI1CMD & SPIBUSY) {
```
igrr: I don't think this is correct. 
`inTransaction` may only be set to `false` by a call to `endTransaction`. But how will `endTransaction` be called if you are spinning inside this loop?
Could you please share a simple sketch which illustrates the issue fixed by this change?

KaloNK: I want to read sensors data from an interrupt or ticker, but as the interrupt may come while writing to the LCD or even while reading another sensor, I was looking for a way to protect the transaction.
The idea is that the last called interrupt handler will yield() to allow the previous one to complete it's SPI communication.
Why yield is not interrupt safe? Were can I read more about it?

libraries/SPI/SPI.cpp
```diff
@@ -74,13 +75,17 @@ void SPIClass::setHwCs(bool use) {
...
+        yield();
```
Links2004: the SPI itself is interrupt safe, but you add a ```yield``` and that is not.
all function that can trigger a task switch like ```yield``` and ```delay``` are designed to be only on the arduino task (setup or loop), if you call them in a interrupt the system will crash.
https://github.com/esp8266/Arduino/blob/master/doc/reference.md#timing-and-delays

KaloNK: > the SPI itself is interrupt safe

unfortunately it is not when software CS and/or multiple devices are used.
Consider interrupt is coming while inside `loop` there is a write (of several lines one by one) to LCD and it is right in the middle of that block (properly protected with `beginTransaction` `endTransaction`):
 * The CS for the LCD is held down and write is started
 * The interrupt routine is called and it successfully starts new transaction after the write is complete
 * The interrupt routine is taking the sensor CS low, but the LCD CS is still low
 * writing garbage to LCD or worse - damaging MISO of the devices if they both try send data
 
The only other way (I can think of) to solve the problem is to disable the interrupts during the SPI transaction. Should I make that change instead?

igrr: Can you set a flag inside the interrupt and then do the actual transaction
later inside loop, if the flag is set?
Also, which interrupt are you talking about? SPI or GPIO?

KaloNK: I mean GPIO interrupt and SPI transaction should happen inside the interrupt handler.
I want to read the data from MCP3914 immediately when it is ready (attaching GPIO interrupt to the DR assigned GPIO pin) and probably several times during a single `loop`, even during WiFi, TCP/IP or other tasks (if/when possible) that may take too long to complete.
Other sensors I will be using, do allow the use of a flag, but not this one, so i will probably just disable the interrupt while working with SPI inside `loop` and re-enable it again.

igrr: Okay, in this case disabling GPIO interrupt is a good solution.

KaloNK: Should I modify the patch to do this or it should not be part of the SPI code?

igrr: I think this behaviour should be handled in user code. Some people might want GPIO interrupts during SPI transactions, if GPIO interrupts serve other purposes than starting SPI transactions.

Issue #1011, opened 2015-11-12
Disconnected postmortem capability


cpacey:
The [built-in postmortem](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_postmortem.c#L54) logging has been very useful.  However, the data is always output to UART0, which functionally means that I need to be physically connected to the ESP to get that data.  More and more, I'm wanting to get this information from a device in the field which has nothing attached to UART0.

I'm not entirely sure the best way to do this yet - among the options I've considered:
1. Copy the crash information (`rst_info`, `sp`, the stack contents, etc) to flash before restarting, and then process as desired at next start up.  My biggest worry about this is that we might do too many flash writes to a single sector, especially if the device gets into a tight crash loop.
2. Send the info somewhere else, e.g. a quick UDP transmit.  I've not tried this yet, so I don't know if it's feasible.

Regardless of the specific implementation, I could do it myself if I could register my own callback for when a crash occurs.

Links2004:
the reset reason is allready stored you can use
```cpp
Serial.println(ESP.getResetInfo());
```
https://github.com/esp8266/Arduino/blob/master/cores/esp8266/Esp.cpp#L291

to get basic info, no full stack dump but epc1, epc2, epc3, excvaddr and depc
which deliver a good start for a debugging.

cpacey:
We're already using ESP.getResetInfo(), but aside from the missing stack trace (which has been very useful for me), ESP.getResetInfo() doesn't give us anything for Exceptions or Soft WDT resets, at least for the Huzzah boards we're using.  Essentially, I can only use ESP.getResetInfo() right now to find out if there was a hard wdt reset.

PR #1012, opened 2015-11-12
Add the ability to be called back when the device is about to reset

cpacey:
Attempt to address #1011.  Feedback is desired, because it's not the cleanest code.

Caveats:
- Be very careful in your callback.
- Will only actually get called for Exceptions and Soft WDT resets.

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.c
+++ b/cores/esp8266/core_esp8266_postmortem.c
@@ -37,6 +37,11 @@ static void uart1_write_char_d(char c);
 static void print_stack(uint32_t start, uint32_t end);
 //static void print_pcs(uint32_t start, uint32_t end);
 
+extern void __custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) {
+}
+
+extern void custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) __attribute__ ((weak, alias("__custom_crash_callback")));
+
 void __wrap_system_restart_local() {
     register uint32_t sp asm("a1");
 
@@ -92,6 +97,9 @@ void __wrap_system_restart_local() {
 
     // print_pcs(sp + offset, stack_end);
     print_stack(sp + offset, stack_end);
+
+    custom_crash_callback( &rst_info, sp + offset, stack_end );
+
     delayMicroseconds(10000);
     __real_system_restart_local();
 }


cores/esp8266/Esp.h
```diff
@@ -131,6 +133,13 @@ class EspClass {
         bool eraseConfig();
 
         inline uint32_t getCycleCount();
+
+        /**
+         * Be very careful in your callback.  Don't call anything that could
+         *  yield (so pretty much don't call Arduino code).  And god help you if
+         *  it crashes.
+         */
+        void installCustomCrashCallback( CrashCallback );
```
cpacey: This was the best place I could find to allow installing the callback.

cores/esp8266/Esp.h
```diff
@@ -81,6 +81,8 @@ typedef enum {
      FM_UNKNOWN = 0xff
 } FlashMode_t;
 
+typedef void (*CrashCallback)( struct rst_info *, uint32_t, uint32_t );
```
cpacey: I couldn't find a suitable header to put this typedef in to avoid duplicated the sig in other files, and I was hesitant to create a new header just for this one thing.

cores/esp8266/Esp.h
igrr: My suggestion is to install custom crash handler by overriding some weak
function, which defaults to no-op.

cpacey: I didn't know about that feature, that's way cleaner.  I'll make that change, thanks!

cores/esp8266/core_esp8266_postmortem.c
```diff
@@ -92,6 +97,9 @@ void __wrap_system_restart_local() {
 
     // print_pcs(sp + offset, stack_end);
     print_stack(sp + offset, stack_end);
+
+    __custom_crash_callback( &rst_info, sp + offset, stack_end );
```
cpacey: Well that's embarrassing.

cores/esp8266/core_esp8266_postmortem.c
```diff
@@ -37,6 +37,11 @@ static void uart1_write_char_d(char c);
 static void print_stack(uint32_t start, uint32_t end);
 //static void print_pcs(uint32_t start, uint32_t end);
 
+extern void __custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) {
+}
+
+extern void custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) __attribute__ ((weak, alias("__custom_crash_callback")));
```
cpacey: I've aped the style [used elsewhere](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_wiring_digital.c#L177) [in this project](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_main.cpp#L96), though I must admit I don't know the reason for the alias (my guess is that it makes for less confusion when looking at stack dumps).

cpacey:
Apologies for all the churn.  I'll try to keep that down in the future.

PR #1056, opened 2015-11-19
Async ota

me-no-dev:
still up to the user to call ArduinoOTA.handle() to start the upload


libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -70,7 +84,12 @@ void ArduinoOTAClass::begin() {
     _port = 8266;
   }
 
-  _udp_ota.begin(_port);
+  _udp_ota = new UdpContext;
```
igrr: You have to check that `_udp_ota` is nullptr before creating another one. If it isn't, `unref` it first. This could happen if someone calls begin() twice for some reason.

Also destructor is missing... Add it along with `if (_udp_ota) _udp_ota->unref();`. I know that ArduinoOTA is a static object now, but it's omission makes any c++ programmer weep :)

me-no-dev: there is a destructor :) it's empty though

igrr: Oh right. Lost it among other one-liners.

PR #1085, opened 2015-11-26
Fix handling of / to load index.htm

sticilface:
Recent changes broke the handling of / leading to index.htm.

Im not sure if changing the requestUri is the best / or valid way to do this.  but it works.  It will just change it on first call to the handle.  
If you prefer, I can make a local copy of the requestUri, append the index.htm if required, then use that throughout the rest of the function?

### DIFF
--- a/libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
@@ -81,7 +81,7 @@ public:
 
         String path(_path); 
 
-        if(path.endsWith("/")) path += "index.htm";
+        if(requestUri.endsWith("/")) requestUri += "index.htm";
 
         if (!_isFile) { 
             // Base URI doesn't point to a file. Append whatever follows this


pgollor:
Thanks. I will check it at the weekend okay?

sticilface:
No problem.  

I think it comes from the fact that the path does not have a leading / when it is a directory.  when it is the request that should really interpreted.  My concern is that it might be better to cache the uri within the handler rather than modify the uri.  or if the uri is to be modified by appending index.htm then it could be done earlier before the handles are called. (if that makes sense)

pgollor:
Sorry for the late replay. I did some tests and i think your usage of the `serveStaic` function is wrong.
In my opinion it have to be `server.serveStatic("/index.htm", SPIFFS, "/test/index.htm", "max-age=86400");`

This works without changing the source code.

sticilface:
I was under the impression that servestatic allows you to serve directories?  What is the purpose of this part?
```
        if (!_isFile) { 
            // Base URI doesn't point to a file. Append whatever follows this
            // URI in request to get the file path.
            path += requestUri.substring(_baseUriLength); 
        }
```

This is a good feature as if you have 10 files in a directory you do not need a request handler for each separate file.  Each handler uses a not insignificant amount of memory.  It is standard web server behaviour to look for an index.htm file in a directory.   Plus this used to work, before the refactoring a few commits back. That is just my two cents.

sticilface:
I think the important thing to emphasise is that this used to work, just fine.

igrr:
Looks almost good to me.

Serving whole directories was actually what I had in mind when I added `serveStatic`. However due to a bug it could also be used to serve individual files. So both cases should work fine.

I moved the `if(requestUri.endsWith("/"))` clause inside `if (!_isFile)`, because another valid use would be:
```c++
server.serveStatic("/info/", SPIFFS, "/info.html");
```
I.e. we should not try adding `index.htm` to the URI if target path is a file already.

I will close this pull, but thanks for pointing out the fix.

Issue #1088, opened 2015-11-27
SPIFFS filename length > 30, causes unexpected behaviour


sticilface:
Creating files of length greater than 31 causes extra characters to be appended to filename, which cannot then be deleted.  (really it is 31, but the first / counts) shown below, is an example.  
```
     FS File: /29aaaaaaaaaaaaaaaaaaaaaaaaaa, size: 31B
     FS File: /30aaaaaaaaaaaaaaaaaaaaaaaaaaa, size: 32B
     FS File: /31aaaaaaaaaaaaaaaaaaaaaaaaaaaa, size: 33B
     FS File: /32aaaaaaaaaaaaaaaaaaaaaaaaaaaaa. ", size: 34B
     FS File: /33aaaaaaaaaaaaaaaaaaaaaaaaaaaaa. #, size: 35B
```
Here is a gist that demonstrates the problem.  
https://gist.github.com/sticilface/20c9973da4ccbe993c97
Warning it deletes all files on the device.

run the sketch, then run it again and you will see that the files 32 + 33, are not deleted.  

proposal  , adding length check to file.open()
```
File FS::open(const char* path, const char* mode) {
    if (!_impl) {
        return File();
    }

    OpenMode om;
    AccessMode am;
    if (!sflags(mode, om, am)) {
        DEBUGV("FS::open: invalid mode `%s`\r\n", mode);
        return File();
    }

    if (strlen(path) > 31) return File(); 
    
    return File(_impl->open(path, om, am));
}
```

I do realise that 32 is the max length, I just wanted to demonstrate, but it appears 31 is really the practical max length.  If you accidentally make this it should fail and not create an un-deletable file.

igrr:
e7024fb

PR #1089, opened 2015-11-27
Proposed filename length check for SPIFFS

sticilface:
as mentioned https://github.com/esp8266/Arduino/issues/1088

### DIFF
--- a/cores/esp8266/FS.cpp
+++ b/cores/esp8266/FS.cpp
@@ -196,6 +196,7 @@ File FS::open(const char* path, const char* mode) {
         DEBUGV("FS::open: invalid mode `%s`\r\n", mode);
         return File();
     }
+    if (strlen(path) > 31) return File(); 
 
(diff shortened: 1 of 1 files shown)


igrr:
- I think this check should be done not it FS abstraction layer, but in [SPIFFSImpl](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/spiffs_api.cpp#L49). Reason is, we may have another FS implementation (like SD) which don't share this limitation.
- This check should also be done in all other functions which take file names. Perhaps it's worth adding a function like `bool SPIFFSImpl::_isValidFilename(const char*)` which would be called from other places.

sticilface:
This sounds very reasonable

sticilface:
Looking at https://github.com/esp8266/Arduino/blob/master/cores/esp8266/spiffs_api.cpp#L49 it might be better for you to implement it.  I don't understand the code that well!

PR #1144, opened 2015-12-04
add a simple TCP example

Links2004:


Juppit:
Is it realy a good name, WiFiClientMin?
What about WiFiClientBasic?

Am 04.12.2015 um 18:21 schrieb Markus:

Links2004:
yes this is a better name ;)

PR #1148, opened 2015-12-05
added ESPino to supported boards

Rodmg:
Added support for the ESPino board (http://www.espino.io/en) in boards.txt and variants/espino.


boards.txt
```diff
@@ -691,3 +691,88 @@ d1_mini.menu.FlashSize.4M1M.build.spiffs_start=0x300000
...
+espino.menu.CpuFrequency.80.build.f_cpu=80000000L
+espino.menu.CpuFrequency.160=160 MHz
+espino.menu.CpuFrequency.160.build.f_cpu=160000000L
+
+espino.menu.FlashFreq.40=40MHz
+espino.menu.FlashFreq.40.build.flash_freq=40
+espino.menu.FlashFreq.80=80MHz
+espino.menu.FlashFreq.80.build.flash_freq=80
+
+espino.menu.FlashMode.dio=DIO
```
igrr: Since you probably know the exact flash model used by this board, DIO/QIO and flash frequency selection is probably unnecessary?

igrr:
Could you please add this board to doc/boards.md and to package/package_esp8266com_index.template.json as well?

Rodmg:
@igrr, Thank you for your remarks, I added the requested changes. About the flash frequency and mode, I removed the frecuency selection as it is always 40, but I think keeping flash mode selection could be handy as future versions of this board will come with the ESP-12E module. I'm not completely sure but I've read somewhere that changing flash mode frees some of the extra pins of that module?

PR #1154, opened 2015-12-05
fix #1151 + eclipse makefile example

### LINKED ISSUE #1151: SPI definition correction
Hi,

Correction SPI len Adrr in esp8266_peri.h : 
#define SPILADDR	26 // not 16

from source Espressif SDK spi_register.h 
#define SPI_USR_ADDR_BITLEN_S 26

(Sorry, i don't know how to submit from my clone :-/ )

Links2004:


asetyde:
this correction is on staging ?

Links2004:
its only in git right now

asetyde:
ok , if my sensor uses a pin gpio10 or nodemcu SD3 that is also used for SPI , it can be a cause that my device with same code on 1.6.5 stable go but on 2.0.0 not go

PR #1160, opened 2015-12-06
replace delay() with a while loop in WiFiClient.ino

martinayotte:
Since many users on the forum complained that the example was not running due to the fact that delay(10) wasn't enough for some slow servers, I'm proposing this change.


Links2004:
for http client better use:
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266HTTPClient/examples/BasicHttpClient/BasicHttpClient.ino 
may wie shut change the WiFiClient example to a simple client (non http)

PR #1190, opened 2015-12-10
update ota docs

Links2004:


krzychb:
Nice :smile: :+1:
People keep asking how it works.

PR #1213, opened 2015-12-14
Interrupt safety fixes for HardwareSerial

alltheblinkythings:


cores/esp8266/cbuf.h
```diff
@@ -113,9 +113,9 @@ class cbuf {
             return (ptr == _bufend) ? _buf : ptr;
         }
 
-        size_t _size;
+        const size_t _size;
         char* _buf;
-        char* _bufend;
+        const char* const _bufend;
```
Links2004: two time const?

igrr: That's a const pointer to chars which are const, so `_bufend` is const and `*_bufend` is const as well.

PR #1232, opened 2015-12-16
SoftwareSerial and SoftwareSerialRX based on timer 1 

me-no-dev:
stable speeds up to 115200bps, some issues may be seen on first few
received bytes at 230400bps
Speeds above that are not possible
PWM can not be used when one of those libraries are in use


libraries/ESPSoftRx/ESPSoftRx.h
```diff
@@ -0,0 +1,43 @@
...
+    int peek(void) override;
+    int read(void) override;
+    size_t read(char*, size_t);
+    void flush(void) override;
+    size_t write(uint8_t data) override {
+      return -1;
+    }
+};
+
+extern ESPSoftRxClass SerialRx;
```
igrr: Instead of creating a global object inside the library, how about leaving this to the user?
This would allow user to pass buffer size to the constructor.
I know that you can only have one instance, but a note "You can only have one software serial instance in your sketch" in documentation is good enough.
Same applies to the second software serial class.

libraries/ESPSoftRx/ESPSoftRx.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+#include "Arduino.h"
+#include "ESPSoftRx.h"
+#include <cbuf.h>
+
+extern "C" uint32_t interrupt_reg;
```
igrr: Since you have put interrupt_reg into Arduino.h, no need to re-declare it here. Same for the other class.

libraries/ESPSoftRx/ESPSoftRx.h
igrr: One more thing: if you avoid creating global objects inside the library and move all your static variables (_srx_rx_pin, _srx_rx_buf, etc) into the class, then you can combine both classes (bidirectional and rx-only) into a single library. That would count as a big improvement. Plus you should be able to reuse some code, because bidirectional one is basically rx-only + more code for tx.

reaper7:
:+1: 
one common library is a good solution

libraries/ESPSoftRx/ESPSoftRx.cpp
me-no-dev: will revert that and go exclusive on the pin interrupt. If the user has a handler on a lower pin and those trigger at the same time or one prevents the other, then the pulse will be missed.

libraries/ESPSoftRx/ESPSoftRx.h
me-no-dev: I'm really going for all interrupts in RAM here ;) if you pull something away from RAM, things go slow and stuff stops working. Not sure how you want me to tackle the global object. What's the point of giving option to create more than one if no more than one can exist?

igrr: You can keep ICACHE_RAM_ATTR, i was not suggesting to remove it.
The point of not declaring a global object inside library is *not* to allow multiple instances. It's to achieve the following:
- allow user to pass buffer size to the constructor
- combine two classes into a single library and increase code reuse (through inheritance or composition)

me-no-dev: combining is not an issue at all, I have them separate on purpose rather :) Honestly I was thinking of not even publishing the RX/TX lib and have just the RX for Serial1. I get the point for the buffer though and had that in mind.
will storm a bit more later and think where this lib should go. Needs inverting, parity, stopbits and more implemented as well.

Issue #1111, opened 2015-11-30
HttpUpdate not booting properly


adrionics:
I managed to get the httpUpdate working except that it was crashing soon after writing out one line to the serial port.
My httpUpdate call is as follows:
  t_httpUpdate_return ret = ESPhttpUpdate.update(host, port, url, currentver);

After comparing the process to the POST update example, I found that the _startAddress value in Updater.cpp was different between the WebUpdate example and the httpUpdate method.
In the POST method, the _startAddress was basically the existing sketch size and in the httpUpdate it was the remaining free space less the new sketch size.

After changing the Updater.cpp file as follows, it started working:
Line 74:
    updateStartAddress = currentSketchSize ;//(WAS) updateEndAddress - roundedSize;

Why is the updateStartAddress different between the httpUpdate method and the WebUpdate POST example?
What would the correct fix be?

What a wonderful feature this OTA update is!!

me-no-dev:
the correct way is to start at least after the current sketch rounded up to a block.
I see no reason for them not to work.
Main difference is that the POST example does not know the size of the incoming sketch so it reserves all of the available space, minus the EEPROM.

Links2004:
can you provide a serial log with:
```cpp
Serial.setDebugOutput(true);
```
and enable the debugs for the update:
https://github.com/esp8266/Arduino/blob/master/cores/esp8266/Updater.cpp#L6
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.h#L35

adrionics:
I found the cause. I had an incorrect flash memory size (larger than the chip) and so the startaddress was calculated incorrectly.

The httpupdateserver example works due to the startaddress being just after the existing sketch, where the httpupdate library does not since it somehow uses the size of flash.

The exact situation was setting the flash memory to 4M (Adafruit Huzzah) and writing the sketch to a 1M  ESP-07.

Curiously, EEPROM works. I do not use SPIFFS and suspect it may have problems in this situation.

Would it be an idea for the update library to use the chip reported size rather than the compiler reported - if that is what's happening?

Links2004:
if the chip size is wrong selected, this will make many problems,
simply using the real one in not that easy, since the memory regions are different and they are configured during compile time (linker files).
but its possible cancel the update when the current settings are wrong, by this its easier to see that there is a problem.
[...]

PR #1235, opened 2015-12-17
HTTPupdate add flash size check + better error output

Links2004:
see #1111


cores/esp8266/Updater.cpp
```diff
@@ -57,6 +57,14 @@ bool UpdaterClass::begin(size_t size, int command) {
     return false;
   }
 
+  if(ESP.getFlashChipRealSize() != ESP.getFlashChipSize()) {
```
igrr: I think we only need to return an error if `ESP.getFlashChipRealSize() < ESP.getFlashChipSize()`.

PR #1242, opened 2015-12-17
mDNS fixes and enhancements

edog1973:
I've made some fairly significant changes to the mDNS lib.  I had posted a question to the esp8266 forum a week ago asking if anyone else was working on these changes and received no response, so I hope this is OK.

I've fixed a couple of bugs that caused invalid packets to be returned when requesting  SRV or TXT records.  In order to fix these I changed _reply to a much more linear flow, but this also removed the compression that was used.  IMHO the greatly reduced code complexity justifies the negligible decrease in efficiency.

There was a bug that caused multiple services to lost.  This has been fixed as well.

I added methods to allow TXT records to be added.  Each service/proto combination added has it's own list of TXT records. I also moved arduinoEnabled to use the new format.  

Lastly I added a setInstanceName method.  This allow you to set a friendly human readable name to be  used with the PTR, TXT and SRV records.


me-no-dev:
good job :) I could have probably answered to you, but I do not visit the forum :)

igrr:
Thank you for the fixes and refactoring!
Could you please do two more things?
- replace `char _hostName[63]`, `char _instanceName[63]`, and `char _txt[128]` with `String`s. Calling `enableArduino` allocates four TXT records, and there is quite an overhead of storing "tcp_check=no" inside an 128-byte buffer. Plus you gain flexibility by supporting TXT records larger than 128 bytes. Host and instance names should also be less than 63 characters most of the time.
- move definitions of `struct MDNSService` and `struct MDNSTxt` to .cpp file, and use forward declarations in the header file.

PR #1259, opened 2015-12-20
add magic byte check for Update class

Links2004:
Note: only check of first 0xE9 possible.

### DIFF
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -172,6 +172,14 @@ bool UpdaterClass::end(bool evenIfRemaining){
 #endif
   }
 
+  if(!_verifyEnd()) {
+#ifdef DEBUG_UPDATER
+    printError(DEBUG_UPDATER);
+#endif
+    _reset();
+    return false;
+  }
+
   if (_command == U_FLASH) {
     eboot_command ebcmd;
     ebcmd.action = ACTION_COPY_RAW;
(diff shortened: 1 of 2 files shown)


igrr:
Is this going to break the case when `_command == U_SPIFFS`?

Links2004:
good point will change.

Links2004:
has the SPIFFS also a magic byte at the start?

igrr:
It has, but i think hard-coding this into Updater.cpp is not really good. It might be better to move that logic into some "Verifier" (i.e. MD5Verifier, SketchVerifier, SPIFFSVerifier, etc.)

igrr:
For SPIFFS the magic is not located at the start of the image:
https://github.com/esp8266/Arduino/blob/342c4ae6fb847bfc787f80b89a2bb888d942dc32/cores/esp8266/spiffs/spiffs_nucleus.h#L201

Links2004:
a verify at the beginning is limited to the first byte, (peek problem).
I solved it for network but make it for Stream class is no good idea since it then brake every lib that uses Streams.

for the sketch a verify at the end is possible since we not have override the working one there,
for the SPIFFS this is not possible.

igrr:
Ok, then I think for SPIFFS we will just use MD5 verification at the end of update.

Links2004:
will do some refactoring, start test for flash first byte,
end test the flash size test like i do in the http update.
moved to two new functions

Links2004:
updated

igrr:
LGTM

PR #1267, opened 2015-12-21
optimize pwm interrupt handler for better precision

me-no-dev:
next timer should be set before writing to GPIO and GPIO should not be written to if no pins are attached there.


igrr:
Thanks, I've pulled the PWM optimization fix.
- Regarding tcp status redefinition, this change breaks the following sketch (because `CLOSED` is not defined anymore).
    ```c++
    #include <ESP8266WiFi.h>

    void setup() {
      WiFiClient client;
      if (client.status() != CLOSED) {
        panic();
      }
    }

    void loop() {}
    ```

- Additional alignment for `pvPortMalloc` and `pvPortRealloc` should not be necessary because the first thing `pvPortMalloc` does is a call to `xPortWantedSizeAlign`, which aligns the size to `portBYTE_ALIGNMENT`.

me-no-dev:
that CLOSED needs to be defined only once. I can't begin to explain what a pain it is to write lwip tcp backed library outside of ESP8266WiFi.
It's a real problem and needs a resolution really. Maybe lwip haders should go to the core?
Regarding alignment. That is a big issue with SDK 1.5 and definitely gets fixed with overloading the functions. But SDK 1.5 is a whole topic on it's own... I have a branch that is still on 1.3 so I can actually do some networking with the ESP.
Maybe come by gitter for a conversation?

tablatronix:
Who wants to commit this ?

me-no-dev:
gimme a sec and will do

me-no-dev:
https://github.com/esp8266/Arduino/pull/1391

PR #1277, opened 2015-12-22
Fix RAM corruption caused by our hook of register_chipv6_phy(init_data*)

alltheblinkythings:
"init_data", when non-NULL, is on the heap, and the register_chipv6_phy call
sometimes modifies data in (at least) the offset range [128:249], suggesting
that it is a buffer larger than 128 bytes in size (the size of our
"phy_init_data" buffer).  When we use our static buffer (prior to this
change), the call could would overwrite the .rodata section and lead to
undefined behaviour.

To address this, just patch the heap-allocated buffer with our data.

Move phy_init_data to flash as it's now readonly and never modified.

### DIFF
--- a/cores/esp8266/core_esp8266_phy.c
+++ b/cores/esp8266/core_esp8266_phy.c
@@ -24,7 +24,9 @@
  #include <stddef.h>
  #include <stdbool.h>
 
-static uint8_t phy_init_data[128] =
+#include "c_types.h"
+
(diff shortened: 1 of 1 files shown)


alltheblinkythings:
Retry of https://github.com/esp8266/Arduino/pull/1276 .  The issue was a null ptr dereference as only the first call to register_chipv6_phy passes a non-null init_data (and more are made after some runtime).

PR #1289, opened 2015-12-23
handle short writes

gtalusan:
Calling HTTPClient->getString() would call writeStream and this report back a short write on success because StreamString returns 0 on success.

### DIFF
--- a/cores/esp8266/StreamString.cpp
+++ b/cores/esp8266/StreamString.cpp
@@ -25,14 +25,10 @@
 
 size_t StreamString::write(const uint8_t *buffer, size_t size) {
     if(reserve(length() + size + 1)) {
-        for(size_t i = 0; i < size; i++) {
-            if(write(*buffer)) {
-                buffer++;
-            } else {
-                return i;
-            }
-        }
-
+        const uint8_t *s = buffer;
+        const uint8_t *end = buffer + size;
+        while(write(*s++) && s < end);
+        return s - buffer;
     }
     return 0;
 }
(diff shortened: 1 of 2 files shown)


alltheblinkythings:
Won't StringStream::write return one too many when a call to write() fails,
after this change?

Easy case to consider: it returns "1", even if the first call to write()
fails.

Preventing this is likely the reason why the previous code had a
conditional increment of *buffer; it was just missing a "return size;" in
the success case.

I think it will also attempt a write if size is 0.

If you want to keep it as a 1 line loop, consider reworking as:

while(s < end && write(*s)) s++;

alltheblinkythings:
I think we'll also return a bad value for bytesWritten, too, because a
partial write isn't accounted in the error case.

(Sorry if this is wrong, am on mobile and only glabced at surrounding code.)

Links2004:
i think this shut be a faster one:
```cpp
size_t StreamString::write(const uint8_t *data, size_t size) {
    if(size && data) {
        if(reserve(length() + size + 1)) {
            memcpy((void *)(buffer + len), (const void *)data, size);
            len += size;
            *(buffer + len) = 0x00; // add null for string end
            return size;
        }
    }
    return 0;
}
```
we can access the string buffer direct.

gtalusan:
@Links2004 yes, memcpy works much better.  Thanks!

gtalusan:
@alltheblinkythings yes, you're correct.  A short write was a short write in my calling code no matter how many bytes were written, but I can see the need for better accounting.  #1294 hopes to address this.

PR #1304, opened 2015-12-26
Add espduino boards

SmartArduino:
ESPDuino boards from doit.am (arduino uno look like board for esp8266)


package/package_esp8266com_index.template.json
```diff
@@ -55,6 +55,9 @@
             {
               "name": "WifInfo"
             }
```
me-no-dev: missing comma after the bracket above and an extra one at the end of "ESPDuino" brackets

PR #1312, opened 2015-12-27
Allow setting TCP timeout

denvera:
Added method to set TCP timeout for connect on HTTP requests. Not sure if its worth allowing this to be set at runtime though, but it was useful in one of my own projects.

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -50,6 +50,7 @@ HTTPClient::HTTPClient() {
     _returnCode = 0;
     _size = -1;
     _canReuse = false;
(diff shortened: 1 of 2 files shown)


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -253,6 +254,14 @@ void HTTPClient::setAuthorization(const char * auth) {
 }
 
 /**
+ * set the timeout for the TCP connection
+  * @param timeout unsigned int
+ */
+void HTTPClient::setTimeout(uint16_t timeout) {
+    _tcpTimeout = timeout;
```
Links2004: may add:
```cpp
if(connected()) {
 _tcp->setTimeout(timeout);
}
```
to allow changing timeout when the connection is already there

denvera: Great idea, thank you for that. I've added in your check.

Issue #1324, opened 2015-12-29
HTTPClient does not support chunked HTTP responses


furyfire:
Chunked HTTP responses are not parsed in the ESP8266HTTPClient lib leading to confusing return data from a http.getString() call.

Sample URL with chunked data. 
http://api.openweathermap.org/data/2.5/weather?q=London,uk&appid=2de143494c0b295cca9337e1e96b00e0

ESP8266HTTPClient should abstract this in the implementation

https://en.wikipedia.org/wiki/Chunked_transfer_encoding

Links2004:
will implement it, after I finish my WiFi code rework.

igrr:
Released in 2.1.0

PR #1333, opened 2015-12-30
HTTPclient - add CHUNKED encoding support

Links2004:
#1324


sticilface:
Seems to have made things a bit unstable..
I get sometimes a wdt reset... and always on the final file a stack dump
```
[3] /configjava.js -> http://sticilface.github.io/espman/examples/ESPmanager-example/data/configjava.js
13.15KB downloaded
Download FAILED 13.15KB downloaded (4.00GB required)
[4] /ajax-loader.gif -> http://sticilface.github.io/espman/examples/ESPmanager-example/data/ajax-loader.gif
6.10KB downloaded
Download Successful
[5] /config.htm -> http://sticilface.github.io/espman/examples/ESPmanager-example/data/config.htm

Exception (9):
epc1=0x40100c16 epc2=0x00000000 epc3=0x00000000 excvaddr=0x2274696e depc=0x00000000

ctx: cont 
sp: 3fff2830 end: 3fff31d0 offset: 01a0

```

sticilface:
Still size -1 though.  this will not fix that though correct?  only the insertion of the length sent into the content

Links2004:
(4.00GB required) sounds wrong to me.
if the server not report any size you only can check the written bytes count returned from 
writeToStream function.
can you use objdump and show me the function / line here it hangs?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -468,61 +470,36 @@ int HTTPClient::writeToStream(Stream * stream) {
...
+    } else if(_transferEncoding == HTTPC_TE_CHUNKED) {
+        while(1) {
+            String chunkHeader = _tcp->readStringUntil('\n');
+            chunkHeader.trim(); // remove \r
+
+            // read size of chunk
+            len = (uint32_t) strtol((const char *) chunkHeader.c_str(), NULL, 16);
+            DEBUG_HTTPCLIENT("[HTTP-Client] read chunk len: %d\n", len);
+            if(len > 0) {
+                int r = writeToStreamDataBlock(stream, len);
```
gtalusan: Should this check if r != len to handle short writes?  If there's a short write then it looks like this'll continue to iterate over the response.

sticilface:
the 4GB is because i was using size_t as value for http.getsize(), not realising that it can be negative.  So  -1 was obviously falling back to some really large number... not a biggy

Links2004:
@gtalusan short write now have its own error (HTTPC_ERROR_STREAM_WRITE) no need to detect it in the application any more.
have add a retry too.

gtalusan:
Awesome!

Spongman:
i don't understand why this is necessary. it's always sending the `Connection: close` header. The only reason to use chunking is if you're using HTTP/1.1 and you _omit_ both this header _and_ `Content-Length`.

Pablo2048:
This is HTTP client so when connecting to NOT ESP8266WebServer there has to be chunking implemented...

Spongman:
oh, sorry, my bad.

PR #1336, opened 2015-12-31
Fixed crash bug with mDNS

BuzzBurrowes:
The character buffer protoName, created on the stack in the function MDNSResponder::_parsePacket(), is uninitialized. It can remain uninitialized all the way to the bottom of that function, where it is passed to the function MDNSResponder::_reply. 

In MDNSResponder::_reply, the pointer passed in is named 'proto'.  You will will find these lines of code near the top of that function...

    //build proto name with _
    char protoName[5];
    os_strcpy(protoName,underscore);
    os_strcat(protoName, proto);
    size_t protoNameLen = 4; 

You can see there that if proto points to an uninitialized character buffer on the stack, the strcat call can blow way the stack.

I was getting random crashes when a client tried to connect to my esp8266 server using the 'server.local' host name. Those crashes have disappeared since I made this change.

Oh... one more thing... I changed a few memcpy to memmove calls where the source and destination overlap. Not sure that was causing a problem, but I believe memcpy does not guarantee correct results if the source and destination overlap.

### DIFF
--- a/libraries/ESP8266mDNS/ESP8266mDNS.cpp
+++ b/libraries/ESP8266mDNS/ESP8266mDNS.cpp
@@ -495,7 +495,7 @@ void MDNSResponder::_reply(uint8_t replyMask, char * service, char *proto, uint1
   size_t hostNameLen = hostName.length();
 
   char underscore[]  = "_";
-
+ 
   // build service name with _
   char serviceName[os_strlen(service)+2];
   os_strcpy(serviceName,underscore);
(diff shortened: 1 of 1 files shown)


BuzzBurrowes:
This continuous build failure looks unrelated to my change.

Links2004:
yes build server has failed, make a empty commit or merge again from master if possible.

BuzzBurrowes:
Done

PR #1338, opened 2015-12-31
Add ability to calculate MD5 of Stream

sticilface:
Add ability to calculate MD5 of Stream


cores/esp8266/MD5Builder.cpp
```diff
@@ -23,6 +23,47 @@ void MD5Builder::addHexString(const char * data){
   free(tmp);
 }
 
+bool MD5Builder::addStream(Stream & stream, const int total_len) {
+    const int buf_size = 512;
+    int bytesleft = total_len;
+    int bytestoread;
```
gtalusan: bytestoread looks unused here.

cores/esp8266/MD5Builder.cpp
```diff
@@ -23,6 +23,47 @@ void MD5Builder::addHexString(const char * data){
...
+                }
+
+                // not read more the buffer can handle
+                if(readBytes > buf_size) {
+                    readBytes = buf_size;
+                }
+
+                // read data  
+                int bytesread = stream.readBytes(buf, readBytes);
+                bytesleft -= bytesread;
```
gtalusan: Also, here, I think bytesread should be subtracted after the check.. and handle the case where the read is short.

Links2004: I think there can not be a short read since sizeAvailable is what we already have in the buffer to be used.

gtalusan: That's assuming the implementation details of Stream, though.  If you imagine something like a GZIPStream that decompresses on read then you might have issues if there's a problem decompressing a block of the underlying data.

gtalusan: But probably not a big deal if you're only interested in whether or not the MD5 matches.

Links2004: yes in this case you are right, 
zip or gzip support on the ESP where nice, 
but haven't looked in to it yet.

sticilface:
mmm is there a better way to do this, that doesn't involve me committing and uncommiting, and making a bit of a mess here.  I guess i can delete the pull request and just start afresh - that would loose comments though (i guess)... i do have a love hate with git.. any advice welcome?

Links2004:
you can combine commits an make a forced push to your git.

asetyde:
Hi , it isn’t possible do a staging version ? at now stable is unique

sticilface:
mah.... i combined them... but it won't let me push as I'm technically behind!  ugh... hate git sometimes...

sticilface:
Couldn't do it..  my git is not very hot.
see
https://github.com/esp8266/Arduino/pull/1344

PR #1344, opened 2016-01-01
Improvements for MD5Builder with Stream

sticilface:
Addition of addStream(Stream, length) to MD5 builder
Changes to FS to allow super fast readBytes.  
These changes drop ReadBytes of a 20K file from 1400ms to 63ms, so worth having. 

Sorry for the new pull request, couldn't squash commits i'd pushed... tried...


sticilface:
ahhhhhhhhhhh!!!

PR #1350, opened 2016-01-02
allow control of enabling debug and debug level from IDE

Links2004:
allow control of  debug and debug level from IDE

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -183,14 +183,24 @@ generic.menu.DebugLevel.Core=Core
 generic.menu.DebugLevel.Core.build.debug_level=-DDEBUG_ESP_CORE
 generic.menu.DebugLevel.SSL=Core + SSL
 generic.menu.DebugLevel.SSL.build.debug_level=-DDEBUG_ESP_CORE -DDEBUG_ESP_SSL
+generic.menu.DebugLevel.WiFic=Core + WiFi
+generic.menu.DebugLevel.WiFic.build.debug_level=-DDEBUG_ESP_CORE -DDEBUG_ESP_WIFI
+generic.menu.DebugLevel.WiFi=WiFi
+generic.menu.DebugLevel.WiFi.build.debug_level=-DDEBUG_ESP_WIFI
 generic.menu.DebugLevel.HTTPClient=HTTPClient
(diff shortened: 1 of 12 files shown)


asetyde:
About esp Gdb or other ??

Links2004:
its manage the serial out debug messages.
its independent of GDBStub, its run-time debugging. 

often we need more info when we get a new issue,
the we most ask to change this line and that to activate debug output.
whit this changes it simple in the IDE menu and its way easier to use.

asetyde:
Ok because I need a use Gdb debug becomes from old serial debug too few information .

I've see that my board it freeze and. I don't know why .. Rebuild and rewrite code many many times can  do problems ..at now with new code board never it freeze ...

Use esp.reset / restart / eraseConfig and new flash can clean board ? Because I think it missed some part ..

Links2004:
esp.reset / restart are only like shutdown and reset button on the computer.
eraseConfig will erase the SDK config area in flash.

for GDBstub take a look here:
https://github.com/esp8266/Arduino/tree/master/libraries/GDBStub

asetyde:
Many Thanks

krzychb:
Hi Markus,

This is great idea to bring this up to menu level for key libraries :+1:

*Generic ESP8266 Module* is getting really rich of features.
Who will buy other types of modules? :wink:



Krzysztof

hallard:
That's very interesting feature, if there any tips on how can we get this into our own sketches ?
I mean enable debug for any sketch and also use the serial defined in IDE ? Certainly with some #ifdef but I don't know the values I need to check ?

Links2004:
it work with every sketch that enables the Serial interface that is selected as debug port.
you still can use the Serail interface for your own output,
the debug output is additional and will not disable any interface from usage in the sketch.

all used defines are listed here:
https://github.com/esp8266/Arduino/blob/master/boards.txt#L180-L207

the port has the define ```DEBUG_ESP_PORT``` possible value:
Disabled: define not existing
Serial: Serial 
Serial1: Serial1

hallard:
Exactly what I needed, thanks, and I wish an happy new year to all team members ;-)

PR #1357, opened 2016-01-03
Added default debug Mode to WifInfo boards

hallard:


Links2004:
if you set debug_port=Serial1 the sketch must include a Serial1.begin() or it may crash.
the DEBUG_ESP_WIFINFO define will not have any effect to the code what is the goal of it?

hallard:
@Links2004 
 WifInfo sketch already include a Serial1.begin(), but good to know it's mandatory I will include it with #ifdef of the define.
the DEBUG_ESP_WIFINFO will be used in Wifinfo sktech like this user will be able to enable/disable debug from IDE, I will check this define in the sketch. But if there is a rule for define name or even not the good way to do this, I can change it, it's not fixed, just let me know.

PR #1362, opened 2016-01-03
A content length of zero should also be sent

shmuelzon:
This is needed since when the content-length header is not sent the clients will
wait for data anyways. Sending a content length of zero will tell the client not
to expect any content and it will close the connection immediately.

Since size_t is unsigned, checking if the content-length is >=0 seems redundant.

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -213,11 +213,10 @@ void ESP8266WebServer::_prepareHeader(String& response, int code, const char* co
         content_type = "text/html";
 
     sendHeader("Content-Type", content_type, true);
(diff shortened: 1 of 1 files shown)


EUA:
I think this patch could fix https://github.com/esp8266/Arduino/issues/1365

igrr:
For the case when `_contentLength == CONTENT_LENGTH_UNKNOWN`, this will call
```c++
sendHeader("Content-Length", String(contentLength));
```
where `contentLength` may be zero. So this will send `Content-Length: 0` header even though content length was declared to be unknown.

shmuelzon:
I've rearranged the conditions a bit so the code (to me) is a bit more readable and also answers the issue Ivan raised.

Links2004:
the build/test server has failed, try to merge the latest version to trigger a retest.
@igrr may the get.py script need some retry function for download errors.

shmuelzon:
Done

PR #1364, opened 2016-01-04
add support for manually selecting ip and port for host side

mogorman:
my firewall wouldnt allow for the esp to talk to my pc.  I added support for maually selecting host ip and port for espota.py code.

### DIFF
--- a/tools/espota.py
+++ b/tools/espota.py
@@ -5,11 +5,12 @@
 #
 # Modified since 2015-09-18 from Pascal Gollor (https://github.com/pgollor)
 # Modified since 2015-11-09 from Hristo Gochkov (https://github.com/me-no-dev)
+# Modified since 2016-01-03 from Matthew O'Gorman (https://githumb.com/mogorman)
 #
 # This script will push an OTA update to the ESP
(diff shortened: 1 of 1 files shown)


tools/espota.py
```diff
@@ -7,9 +7,9 @@
 # Modified since 2015-11-09 from Hristo Gochkov (https://github.com/me-no-dev)
 #
 # This script will push an OTA update to the ESP
-# use it like: python espota.py -i <ESP_IP_address> -p <ESP_port> [-a password] -f <sketch.bin>
+# use it like: python espota.py -i <ESP_IP_address> -I <Host_interface> -p <ESP_port> -P <Host_port> [-a password] -f <sketch.bin>
 # Or to upload SPIFFS image:
-# python espota.py -i <ESP_IP_address> -p <ESP_port> [-a password] -s -f <spiffs.bin>
+# python espota.py -i <ESP_IP_address> -I <Host_interface> -p <ESP_port> -P <HOST_port> [-a password] -s -f <spiffs.bin>
```
igrr: This should probably be <host IP>, because <host interface> might imply something like eth0 or wlan0.

mogorman: yup you are right. fixed.

atrent:
See also my proposal on #2480

PR #1395, opened 2016-01-09
fix bad (recently changed) function signatures in lwip

me-no-dev:


libraries/ESP8266WiFi/src/lwip/mem_manager.h
```diff
@@ -69,9 +69,9 @@ static const unsigned short heapSTRUCT_SIZE	= ( sizeof( xBlockLink ) + portBYTE_
 
 void vApplicationMallocFailedHook( void ) ;//ICACHE_FLASH_ATTR;
 
-void *pvPortMalloc( size_t xWantedSize ) ;//ICACHE_FLASH_ATTR;
+void *pvPortMalloc( size_t xWantedSize, const char* file, int line ) ;//ICACHE_FLASH_ATTR;
```
Links2004: void *pvPortMalloc(size_t xWantedSize, const char* file, int line) __attribute__((malloc, alloc_size(1)));
is better for the compiler optimization

me-no-dev: sure thing

PR #1397, opened 2016-01-09
Use latest version of SoftwareSerial

plerup:


igrr:
I would actually prefer to have the version updated explicitly. If we just use the head of master branch, then we don't know if we have to add something like "Updated espsoftwareserial library to 1.1.0" to the change log.
If you plan to update the library frequently without having to go through a pull request every time, then another option to consider is to add this library to the library manager. See this [FAQ](https://github.com/arduino/Arduino/wiki/Library-Manager-FAQ) for details. This way the library will be updated automatically and you can do this as often as you like.

plerup:
Ok, point taken. I will make a new pull request with a specific and properly tag'ed commit.

PR #1401, opened 2016-01-10
WebServer: handle initial read/close timeouts asynchronously

lp0:
If there is a delay before data becomes available on a new connection
then the handler function blocks for up to 1 second, and there's another
wait on connection close for up to 2 seconds.

This doesn't make the whole server process asynchronous (there is another
delay between headers/data) but it helps when there are other events
that need attention while handling web requests.

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -80,6 +80,7 @@ ESP8266WebServer::~ESP8266WebServer() {
 }
 
 void ESP8266WebServer::begin() {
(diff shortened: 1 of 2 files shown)


igrr:
I've done a few tests and this seems to work fine. However I would ask more people to test this change and report feedback here, as there might be some potential edge cases depending on client behaviour.
Just do a `git clone https://github.com/lp0/esp8266-Arduino.git esp8266 && git checkout webserver-async1` in your `Arduino/hardware/esp8266com` directory.

FWeinb:
I tested it and can confirm that the web server runs stable for me.

andig:
Can't say if its faster but no (additional) problems observed.

PR #1416, opened 2016-01-12
added a way to disable SSL

ettisan:
Hi,

I was trying to slim down the binary size of one of my projects and saw that a lot of stuff is included in the binary even if it is never used (empty sketch needs >50% of my ESP module). This is particularly annoying since I want to use HTTP-OTA. 

This PR allows to disable SSL using a preprocessor definition (currently not configureable in the UI).

For me this saves ~14% sketch size (since axtls is ommited).

I have not yet figured how to inspect the contents of the binary image but i guess there are a lot of things I do not require in all of my sketches (e.g. WPS, ...). I'm happy about any pointers.

Wolfgang

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -142,11 +142,13 @@ void HTTPClient::begin(String url, String httpsFingerprint) {
             if(!hasPort) {
                 _port = 80;
             }
+#ifndef DISABLE_SSL
         } else if(protocol.equalsIgnoreCase("https")) {
             _https = true;
             if(!hasPort) {
                 _port = 443;
             }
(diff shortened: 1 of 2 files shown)


igrr:
I would rather rely on the linker to remove unused stuff (through -gc-sections option). Need to investigate why something gets pulled from axTLS even when you don't use wificlientsecure.

Do you use httpclient library in your sketch?

ettisan:
Maybe a little background:
I have a device running where the sketch takes about ~52% of the space. I want to build an update stub that only connects to the wifi and calls ESPhttpUpdate.update(). 
This way i can update the device to this stub (afterwards very little space is used). And this stub then updates to the actual new version (which may take a lot of space).

igrr:
#1857 changes HTTPClient so that WiFiClientSecure is not pulled into the final program, if it is not required.

PR #1423, opened 2016-01-14
Allow setting alternate TX for UART 0, so GPIO1 is available as SPI_CS1

KaloNK:


KaloNK:
Commit 2022fa93a9b3327837d3c8953c9133eb639e0fc0 only. The rest i have messed up my local master branch and had to revert them

KaloNK:
Closing this one and will fix the mess first

igrr:
Could you please change the `bool alternate_tx` parameter to an explicit pin number? As it is, it's pretty hard to tell which one is 'alternate'. 
My suggestion would be to introduce `HardwareSerial::pins(int tx, int rx)` which would swap/remap pins as necessary.
Also could you please squash all the commits?

PR #1424, opened 2016-01-14
Allow setting alternate TX for UART 0, so GPIO1 is available as SPI_CS1

KaloNK:
Second attempt after fixing my local branch mess

### DIFF
--- a/cores/esp8266/HardwareSerial.cpp
+++ b/cores/esp8266/HardwareSerial.cpp
@@ -101,10 +101,11 @@ void uart_disarm_tx_interrupt(uart_t* uart);
 void uart_set_baudrate(uart_t* uart, int baud_rate);
 int uart_get_baudrate(uart_t* uart);
 
-uart_t* uart_start_init(int uart_nr, int baudrate, byte config);
+uart_t* uart_start_init(int uart_nr, int baudrate, byte config, bool alternate_tx);
 void uart_finish_init(uart_t* uart);
 void uart_uninit(uart_t* uart);
-void uart_swap(uart_t* uart);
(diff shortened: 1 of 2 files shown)


igrr:
Copying my comment from the other pull request:

Could you please change the bool alternate_tx parameter to an explicit pin number? As it is, it's pretty hard to tell which one is 'alternate'. 
My suggestion would be to introduce HardwareSerial::pins(int tx, int rx) which would swap/remap pins as necessary.

KaloNK:
Updated as suggested

hallard:
Hi guys,
this fix seems very interesting for my project, may I have more information on question going to my head. What I understood so far was :
- UART0 (TXD0/RX0) use GPIO1/GPIO3 and can be swapped to GPIO15/GPIO13 (but you swap RX and TX and not only one of them)
- UART1 (TXD1/GPIO2) has only TX out

Now, if all off this is correct, now it seems that we can affect UART0 TX to another pin which is GPIO2 but of course, if we do that, we can't use UART1 TXD1 because it's also GPIO2, correct ?

Now my most interesting one, looking the changes, it seems may be, we can just also disable UART0 TX (or affect it's pin func to "nothing"), which means that I can use GPIO1 as I/O and GPIO2 as TXD1 and GPIO3 as UART0 RX. Since on UART0 for my projects, I don't need TX it will be really a good news because I will get new I/O I'm missing.

Do you think it's doable ?

KaloNK:
Hi,
 you can do this even now without this patch - just set the mode for UART 0 to SERIAL_RX_ONLY on init

igrr:
Thanks!
Would be wonderful if you could add these new methods to reference.md.

PR #1451, opened 2016-01-16
advertise all hosted services

probonopd:
Make the OTA port appear in the IDE (without having to restart it) and make it show up in `avahi-browse -a`. Thanks to @me-no-dev for writing this patch.


probonopd:
Thanks to @me-no-dev for writing this patch.

igrr:
This new method should probably be protected/private, because it's only called internally.

probonopd:
Thanks, updated the PR.

me-no-dev:
I think it should stay public :) could be needed if services change maybe readvertisement... Etc
Move to private if proven unnecessary

PR #1454, opened 2016-01-17
Download tools without cd'ing in tools directory first

probonopd:
To install esp8266/Arduino from git, you currently have to run something like

```
git clone https://github.com/esp8266/Arduino.git ~/Arduino/hardware/esp8266/git/ 
cd ~/Arduino/hardware/esp8266/git/tools/ 
python ./get.py
cd -
```

With this patch, it boils down to

```
git clone https://github.com/esp8266/Arduino.git ~/Arduino/hardware/esp8266/git/ 
python ~/Arduino/hardware/esp8266/git/tools/get.py
```

If this is applied, we can adjust the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ Documentation: [http://esp8266.github.io/Arduino/versions/2.1.0-rc1/](http://esp
 - Clone this repository into hardware/esp8266com/esp8266 directory and download binary tools (you need Python 2.7)
 - Note: This has been tested on Linux; on other systems you might have to adjust the path to the `hardware` directory
 ```
-git clone https://github.com/esp8266/Arduino.git ~/Arduino/hardware/esp8266/git/ 
-python ~/Arduino/hardware/esp8266/git/tools/get.py
+git clone https://github.com/esp8266/Arduino.git ~/Arduino/hardware/esp8266/esp8266/ 
+python ~/Arduino/hardware/esp8266/esp8266/tools/get.py
 ```
 - Restart Arduino
 
--- a/README.md
+++ b/README.md
@@ -43,18 +43,11 @@ Documentation: [http://esp8266.github.io/Arduino/versions/2.1.0-rc1/](http://esp
 ### Using git version [![Linux build status](https://travis-ci.org/esp8266/Arduino.svg)](https://travis-ci.org/esp8266/Arduino)
 
 - Install Arduino 1.6.7
-- Go to Arduino directory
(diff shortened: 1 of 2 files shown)


Links2004:
please update the readme/docs too.

probonopd:
Updated docs

README.md
```diff
@@ -43,18 +43,11 @@ Documentation: [http://esp8266.github.io/Arduino/versions/2.1.0-rc1/](http://esp
...
-cd esp8266com
-git clone https://github.com/esp8266/Arduino.git esp8266
+- Clone this repository into hardware/esp8266com/esp8266 directory and download binary tools (you need Python 2.7)
+- Note: This has been tested on Linux; on other systems you might have to adjust the path to the `hardware` directory
 ```
-- Download binary tools (you need Python 2.7)
-```bash
-cd esp8266/tools
-python get.py
+git clone https://github.com/esp8266/Arduino.git ~/Arduino/hardware/esp8266/git/
```
me-no-dev: I think it should be 
```
git clone https://github.com/esp8266/Arduino.git ~/Arduino/hardware/esp8266/esp8266/ 
```
because the name of the folder actually makes ARDUINO_ARCH_[capitalized folder name]

me-no-dev: which means that we can have a main folder espressif in hardware and esp8266 and esp32 subfolders to differentiate the ARCH

probonopd:
I am afraid I don't fully understand the previous comment. Feel free to adjust the patch as needed.

me-no-dev:
I'm not saying it will not work for you :) but if a library uses ARDUINO_ARCH_ESP8266 to filter the architecture, it will fail if your ESP core is in folder named "git", because you will actually have defined architecture ARDUINO_ARCH_GIT. You can look at the build process debug and see -DARDUINO_ARCH_[folder] in every command

probonopd:
Thanks for the clarification, changed the patch accordingly.

krzychb:
I like reduction and simplification  :+1:

Unfortunately “~” (tilde) does not work for Windows.

For that reason I would keep current description of https://github.com/esp8266/Arduino#using-git-version- not to confuse Windows users or find a solution that works and is tested for all platforms (Linux, Windows and Mac OS).


Krzysztof

igrr:
Don't you need to change `dist_dir` as well? It still uses a relative path, AFAICT.

probonopd:
Argh. Windows, you are right, I didn't consider this. Sorry. Then probably it can't be simplified much in a cross-platform way.

PR #1495, opened 2016-01-24
Added function to return reset reason in human readable format

hallard:
Added getResetReason()


asetyde:
Hardware reset WDT ??

hallard:
That's what the SDK says, but I'm surprised too and I'd like to know also ;-)

PR #1503, opened 2016-01-26
Servo value read and write fixes

Makuna:
Values that were written and then read would be different due to using
map.
If user defined a different min/max pulse width would cause the
read/write to incorrectly calculate the pulse width.


Makuna:
fix for issue#1370

PR #1547, opened 2016-02-01
millis() - wraparound fixes

WereCatf:
A lot of code in the libraries do not take wraparound of millis() into account, meaning that after 49 days of uptime they'd suddenly fail to work. I am using the signed/unsigned trick here to fix it and as far as I can tell everything works, but some more experienced code-wizard should double-check.

### DIFF
--- a/cores/esp8266/Stream.cpp
+++ b/cores/esp8266/Stream.cpp
@@ -37,7 +37,7 @@ int ICACHE_FLASH_ATTR Stream::timedRead() {
         if(c >= 0)
             return c;
         yield();
-    } while(millis() - _startMillis < _timeout);
+    } while(((long)millis() - (long)_startMillis) < _timeout);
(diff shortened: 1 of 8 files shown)


drmpf:
the wrap around on an unsigned long actually works with subtraction, try some examples

igrr:
What kind of testing did these changes get? Do you have some test cases which were failing and are fixed with these changes?

WereCatf:
To be honest, I don't know anymore. I woke up again in the middle of the night briefly and this seemed like such a good idea then, but now I don't remember why. I just tested and the comparisons seem to work fine even as they are. I'll just close this and try not to make an even worse fool of myself.

drmpf:
Don't worry about it. I only 'know' they work by convincing myself with testing.
+1 to igrr for his encouraging comment

PR #1570, opened 2016-02-04
WiFiClient::write refactoring

igrr:
- Moved write behaviour (splitting payload into chunks, waiting for ACKs to arrive) from ClientContext into strategies.
- Implemented `BufferStrategy` for payloads already present in RAM, and `ChunkedStrategy` for payloads which need to be copied to RAM on the fly.
- WiFiClient::write(Stream&) will now send as much data as possible, as soon as possible. Previously it would send 1460 byte chunks, and would wait for each chunk to be ACKed before sending new one. Depending on network latency, this gives transfer speed increase within 25% — 100%.
- WiFiClient will now get write timeout from `_timeout` member (1 second by default). Previously timeout was hard-coded to 5 seconds
- ESP8266WebServer sets client timeout of 5 seconds before writing data. This is done to preserve old behaviour of WebServer.
- `WiFiClientSecure` still has 5 seconds timeout hardcoded, some refactoring is necessary to expose `_timeout` member at the point when write function is called by axTLS.
- ESP8266WebServer sendContent (_P) will now rely on WiFiClient::write (_P) to deliver the data

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -241,6 +241,9 @@ void ESP8266WebServer::sendHeader(const String& name, const String& value, bool 
   }
 }
 
+void ESP8266WebServer::setContentLength(size_t contentLength) {
+    _contentLength = contentLength;
+}
 
 void ESP8266WebServer::_prepareHeader(String& response, int code, const char* content_type, size_t contentLength) {
     response = "HTTP/1.1 ";
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -307,67 +307,15 @@ void ESP8266WebServer::send(int code, const String& content_type, const String& 
 }
 
 void ESP8266WebServer::sendContent(const String& content) {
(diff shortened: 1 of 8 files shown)


madsci1016:
Having WiFiClient::write(buf, length) require a stream object breaks compatibility with the Generic Arduino WiFiClient function and is not noted in the ESP Arduino documentation. (Caused me to loose some time today tracking this issue down). It also means Client.write() will behave differently than other stream-like function including serial.write(). 

I was casting some structures as a byte array and passing the pointer to .write functions. Since they aren't objects with available and source members, complier throws an issue.

madsci1016:
Nevermind, please ignore. Not sure why It was defaulting to the stream function when I was passing in a byte array the exact same way I passed it into Serial.write() but I correctly cast it and the compiler took it anyway.

igrr:
WiFiClient has the `write (const uint8_t*, size_t)` overloaded member function, both with and without these changes: https://github.com/esp8266/Arduino/blob/wificlient-strategies/libraries/ESP8266WiFi/src/WiFiClient.h#L50

Before these changes, we had an additional overloaded template member function `template <typename T> write(T& file, size_t unitSize)`. It was a bad design decision. When someone intended to call `write(const uint8_t*, size_t)` and passed a `const char*` for the first argument, compiler selected this template overload, which obviously failed due to lack of `.size()` member function for a POD type.

madsci1016:
Yep, that's what's biting my butt right now. Thanks for maintaining this codebase.

Issue #900, opened 2015-10-18
Tone not working


questuk:
Hi

Have spent an hour trying to solve this and looked on the internet.

When i add this #include <Tone.h> i get a fatal error.

'libraries\Tone\Tone.cpp:24:27: fatal error: avr/interrupt.h: No such file or directory #include <avr/interrupt.h>'

But I am not sure if the Tone library has been implemented on the ESP8266 ?

Regards

Gary

Links2004:
Tone is not implemented for ESP
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/cores/esp8266/Tone.cpp#L46

gorec2005:
I test "tone" work ok only if use pwm for it...

dirty func:
void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) {
pwmFreq(frequency);
analogwrite(_pin,500);
delay(duration);
analogwrite(_pin,0);
}

igrr:
Same as #582?

alx-1:
gorec2005, I get this error when compiling with your tone function : 'pwmFreq' was not declared in this scope'.  Do you have a separate function for it?

gorec2005:
don't mention it... i think somebody implement it to the "kernel" of arduino-esp IDE :-) i don't have enought knowlege for this :-(

Links2004:
closed will be tracked at #582

PR #1581, opened 2016-02-05
Implementation of Tone library using timer1

bjpirt:
I needed to use tones on the ESP so I've made an implementation of this base library. Some notes:
 - it uses timer 1
 - it only currently supports one pin, though it would be quite simple to support two by using timer 0 but I haven't looked at that yet. (the standard library only supports one pin)
 - I've tested it with a logic analyser and the frequency looks good

Let me know if it looks OK to you.
Cheers,
Ben

### DIFF
--- a/cores/esp8266/Tone.cpp
+++ b/cores/esp8266/Tone.cpp
@@ -1,51 +1,125 @@
-/* Tone.cpp
+/*
+  Tone.cpp
 
- A Tone Generator Library
+  A Tone Generator Library for the ESP8266
 
- Written by Brett Hagman
+  Copyright (c) 2016 Ben Pirt. All rights reserved.
+  This file is part of the esp8266 core for Arduino environment.
 
- This library is free software; you can redistribute it and/or
- modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


pieman64:
Is tone now included in 2.10 release?

pieman64:
After changing to 2.1.0 release I still don't have any sound from my 16 ohm passive buzzer.

Tried:

tone(13, 2000, 150); on WeMos D1 Mini but nothing.

The solution at https://github.com/esp8266/Arduino/issues/900 provides sound to the buzzer but not regular tone. Any ideas?

pieman64:
OK I have tome working, in a fashion, on WeMos and regular ESP's.

tone(13, 2000, 1000) will send a tone to GPIO 13 (D7) at 2000Hz for 1 second.
Probably due to the way tone uses interrupts (just one at present I believe) I can't call tone for a second time or call noTone(13) followed by tone(13, 2000, 1000) for a second time.

Is there a work around for this so that the buzzer doesn't just play one fairly short tone but rather tone on, tone off, tone on etc?

PR #1586, opened 2016-02-05
Static initialization guards implementation

igrr:
This pull request adds implementation of libsupc++ functions which are used by gcc to protect local static initializers.

Interrupts are disabled for the duration of static initialization. This is not ideal, but i don't see a way to implement this without disabling interrupts. Imagine the following scenario:

```c++
int first_millis() {
  static int x = millis();
  return x; 
}

void isr_handler() {
  first_millis();
}

void setup() {
  Serial.begin(115200);
  attachInterrupt(0, &isr_handler, CHANGE);
  delay(10);
  Serial.println(millis() - first_millis());
}

void loop() {
}
```

Let's consider the case when `first_millis` is called from `setup` first, and GPIO interrupt happens at the point when `millis` is being executed. Execution context is switched to `isr_handler`, which calls `first_millis` again. C++ standard requires static initialization to be finished before execution is allowed to proceed further, but since we are now inside an interrupt, there is no way of doing that. ISR has higher priority and `millis` will proceed only when ISR is finished.

Therefore I think we have to live with interrupts disabled for the duration of static initialization.

### DIFF
--- a/cores/esp8266/abi.cpp
+++ b/cores/esp8266/abi.cpp
@@ -25,7 +25,10 @@ extern "C" {
 #include "osapi.h"
 #include "mem.h"
 }
+#include <Arduino.h>
+#include <cxxabi.h>
 
+using __cxxabiv1::__guard;
 
 void *operator new(size_t size) {
     size = ((size + 3) & ~((size_t)0x3));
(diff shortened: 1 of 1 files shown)


loopj:
Looks good to me, just tested this and works great in my (previously broken) app.

tuxedo0801:
Looks like the issue is gone with this fix. Thanks a lot.

hallard:
Guys,

I clearly understand this problem, so does this mean now best practices are not using static declaration ? sure we can use them but we're not able to initialize them.

so avoid this
```arduino
void loop() {
  static unsigned long oldms = millis();
}
```
And do that ?
```arduino
unsigned long oldms = millis();
void loop() {
  static oldms = millis();
}
```
is that correct ?

Links2004:
this code change brakes some of my projects,
they not booting any more, only:
```
 
load 0x4010f000, len 1264, room 16  
tail 0 
chksum 0x42 
csum 0x42 
~ld
```

the reason is the malloc change, the umm makes problems.

brankos:
Latest malloc changes cause Telnet example crash and reboot in loop

Links2004:
check #1630

igrr:
Yep, I got interrupt locks wrong, testing a fix here.

richardwilliamson:
Hello, I don't fully understand how git up works- can you advise whether this is now sorted? How can I go about updating the Arduino IDE to use the new code?

Issue #1587, opened 2016-02-05
Pin labels are incorrect


EnochHwang:
In the esp12.png file, the pins on the right-hand side are labeled
D1 (TX0), D3 (RX0), D5, D4, D0, D2, D15, GND

However, in the ESP_min.png and ESP_improved_stability.png files, the pins are labeled
D1 (TX0), D3 (RX0), D4, D5, D0, D2, D15, GND

Notice D4 and D5 are switched.

I think the correct labels should be
D1 (TX0), D3 (RX0), D5, D4, D0, D2, D15, GND

Please verify and correct in pictures

PR #1588, opened 2016-02-06
Schematic pin reordering fix

h4rm0n1c:
Reorders Schematic Pins GPIO 4 and 5 so they reflect the module pinout.

This brings the schematics in line with the module pinouts depicted
elsewhere in the documentation.


h4rm0n1c:
Created in response to
https://github.com/esp8266/Arduino/issues/1587

hallard:
@h4rm0n1c 
thanks for this fix, may be having eagle cad .sch files in the repo also may help other to improve/change and have starting point of quick schematic drawing, just my 2 cents ;-)

duncan-a:
Eagle Components would be very useful if someone has created them...

Rhat brings the total to 4 cents :)

h4rm0n1c:
Most of the libraries I've found have this issue.

However, this is esp8266modules.lbr that I am using for my projects, I only use ESP-12 modules so that's the only part I can verify as being good.

I have used this to make boards, so it's a known good library.

I have modified it to place the GPIOS in the correct order, I also think this is better than newer versions of the library that don't have the off-center solder pads for easy hand soldering.

[esp8266modules.zip](https://github.com/esp8266/Arduino/files/124147/esp8266modules.zip)

PR #1605, opened 2016-02-08
Add ESPert's ESPresso Lite to boards.txt & variants

NAzT:


igrr:
Could you please add the board to doc/boards.md and package/package_esp8266com_index.template.json as well? 
Thanks.

igrr:
Also BUILTIN_LED is kinda non-standard, but we keep it for compatibility with old sketches. Could you please add LED_BUILTIN definition as well?

NAzT:
@igrr  Sure, fixed.

Thanks :D

PR #1629, opened 2016-02-13
Avoid redefining printf

andig:
Needed to avoid userland error when including `umm_malloc.h` for heap diagnosis.

### DIFF
--- a/cores/esp8266/umm_malloc/umm_malloc.c
+++ b/cores/esp8266/umm_malloc/umm_malloc.c
@@ -493,6 +493,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include "ets_sys.h"
 
 #include "umm_malloc.h"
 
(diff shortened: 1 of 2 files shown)


igrr:
Thanks for a reminder, I have opted for less invasive changes (1fc8918).

andig:
I see. How do you expect users to get diagnostic output- do we have to re-implement the define inside the c file?

igrr:
I have simply replaced all ets_printf calls with printf, which works much the same way (both of them call ets_vprintf). As far as output goes, there should be no difference. Enabling debug macros defined in umm_malloc is up to the user, I suppose.

andig:
Unfortunately, it does not link for me:

	C:\Users\xx\AppData\Local\Temp\build585916e0a596b3409ffab1e615c9d964.tmp/arduino.ar(umm_malloc.c.o):(.text.check_poison$part$0$constprop$4+0x8): undefined reference to `putchar'

	C:\Users\xx\AppData\Local\Temp\build585916e0a596b3409ffab1e615c9d964.tmp/arduino.ar(umm_malloc.c.o): In function `dump_mem':

	C:\andi\arduino\hardware\esp8266com\esp8266\cores\esp8266\umm_malloc/umm_malloc.c:786: undefined reference to `putchar'

	exit status 1
	Error compiling.

jeffryr:
Some sketches work, some fail with the undefined reference to `putchar'.  A quick unscientific survey:

Blink example works
AdvancedWebServer fails

I suspect any sketch that invokes Serial.begin() will fail to link.  Here is a the simplest  example that fails.

    void setup() {
      Serial.begin(115200);
    }

    void loop() {
      Serial.println("Hello, World!");
      delay(1000);
    }

Undo the patch and this links correctly.

Links2004:
can confirm:
```umm_malloc.c:1708: undefined reference to `putchar'```

PR #1630, opened 2016-02-14
fix umm memory managment

Links2004:
wrap all SDK memory functions in all libs, to get umm to work
fix umm_malloc.c:1708: undefined reference to `putchar'

### DIFF
--- a/cores/esp8266/memory_wrap.c
+++ b/cores/esp8266/memory_wrap.c
@@ -52,12 +52,24 @@ void * ICACHE_RAM_ATTR __wrap_mem_malloc(size_t size) {
     return malloc(size);
 }
 
+void * ICACHE_RAM_ATTR __wrap_mem_calloc(size_t num, size_t size) {
+    return calloc(num, size);
+}
+
 void * ICACHE_RAM_ATTR __wrap_mem_realloc(void *ptr, size_t size) {
     return realloc(ptr, size);
 }
 
 void * ICACHE_RAM_ATTR __wrap_mem_zalloc(size_t size) {
     return calloc(1, size);
+}
+
+void ICACHE_RAM_ATTR __wrap_mem_init(void) {
+    umm_init();
+}
+
+void * ICACHE_RAM_ATTR __wrap_mem_trim(void *ptr, size_t size) {
+    return ptr;
 }
 
 size_t ICACHE_RAM_ATTR __wrap_xPortGetFreeHeapSize(void) {
--- a/cores/esp8266/memory_wrap.c
+++ b/cores/esp8266/memory_wrap.c
@@ -1,0 +1,65 @@
+/**
+ memory_wrap.c
+
+ Copyright (c) 2015 Markus Sattler. All rights reserved.
+ This file is part of the esp8266 core for Arduino environment.
+
+ This library is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


Links2004:
will fix #1629 and problems from #1586

igrr:
pvPortMalloc and friends are already defined here: https://github.com/esp8266/Arduino/blob/master/cores/esp8266/heap.c, and corresponding functions from libmain.a are marked as weak, so is additional wrapping required?

igrr:
Regarding printf: sorry for breaking the build, but let's fix the real issue (printf not compiling) instead of replacing it with ets_printf.

me-no-dev:
I use umm without any wrapping at all. Have just the Port stuff weakened and overwritten. All has been working nice since some time before you even introduced it in the main repo. Have additionally added alignment check and put all in RAM and that's about it. :) 2 bytes poison and integrity check are on

Links2004:
yes the wrapping is needed some of my projects only boot when the wrapping is here.
and all mem_* stuff is normally in the rom of the chip, for this we need wrapping too.

Links2004:
to get printf work it is normally not more then this:
```cpp
int putchar_cpp(int c) {
    if(Serial.write(c)){
        return c;
    }
    return -1;
}

extern "C" {
int putchar(int c) {
    return putchar_cpp(int c);
}
}
```

but i think debug an error messages are better in os_printf since it easy to disable the output when serial is needed for something else, and we have already many debug out at os_printf.
and it is better when the user can select where printf is printing.

igrr:
Okay, figured that out. The issue has actually been triggered by 737f6c2, in a very interesting manner.

TL;DR: Depending on the libraries used by the sketch, weakening pvPortMalloc/vPortFree may be not sufficient.

Here is a simple sketch which will fail to run (stripped down version of the one provided by @Links2004):

```c++
#include <Arduino.h>
#include <ESP8266WiFi.h>

void setup() {
    Serial.begin(115200);
    Serial.setDebugOutput(true);
    Serial.println();
    Serial.println();
    Serial.println();
    for(uint8_t t = 4; t > 0; t--) {
        Serial.printf("[SETUP] BOOT WAIT %d...\n", t);
        delay(1000);
    }
    Serial.printf("[SETUP] HEAP: %d\n", ESP.getFreeHeap());
    WiFi.mode(WIFI_AP);
    WiFi.softAP("ESP_SLDT", "yesSecureHere");
}

void loop() { }
```

At the same time, more complex sketches (e.g. FSBrowser sample) seem to work fine.
The reason this happens has to do with linker search order. If none of the object files before `libmain.a` reference `pvPortMalloc` symbol, it will not be pulled from `arduino.ar` (archive file containing the core library). Instead, linker will loop inside libmain.a until `pvPortMalloc`/`vPortFree` are found (that's the rule for dealing with archive files: loop until no additional symbols are resolved). If the sketch uses a library which calls `pvPortMalloc` (usually this happens via `os_malloc` and `os_free` defines), for instance ESP8266mDNS, linker will process it first, and will resolve the reference to `pvPortMalloc` via arduino.ar, pulling the "right" definition. From that moment on, the right definition will be used elsewhere.

I'm going to remove mem_manager.o from libmain.a completely, so that the "right" symbols are always used.

PR #1665, opened 2016-02-22
add ISSUE_TEMPLATE.md

Links2004:
hopefully we get better Issues from this.


igrr:
Wow, @Links2004 you are reading my thoughts today. Or maybe vice versa. :)

PR #1668, opened 2016-02-23
Fix a crash in String::changeBuffer()

raheelh:
Calling `String::reserve(0)` causes a crash if String object is in invalidated state. Per the comment on the method's declaration in WString.h, `reserve(0)` would validate the string object but in fact it just causes a crash instead of recovering from invalidated strings.

This change fixes the edge case bug in `String::changeBuffer()` which is the root cause of the crash exposed from `String::reserve()`.

Following test code was used to reproduce the problem and also to validate the fix:

```c++
String result;
while(true){ // suppose this is a loop which reads data from a web page.
  char c = 'A'; // suppose this is a byte read from client.
  result += c; // the loop will eventually cause malloc() to fail here
  if (result.c_str()==0) // even if we detect this
  {
    Serial.println("String INVALIDATED!!!!!");
    // attempt to validate string object so we can use it.
    result.reserve(0);   // before fix, this would crash.
    // if reserve(0) worked, the object is validated, i.e. available for use again.
    Serial.println("Trying to empty...."); // try to make use of object to see.
    result=""; 
    Serial.println("Emptied!!!!"); // works!
    break;
  } 
}
```

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -156,9 +156,11 @@ unsigned char ICACHE_FLASH_ATTR String::changeBuffer(unsigned int maxStrLen) {
     char *newbuffer = (char *) malloc(newSize);
     if(newbuffer) {
         memset(newbuffer, 0, newSize);
-        memcpy(newbuffer, buffer, len);
         if (buffer)
+        {
+            memcpy(newbuffer, buffer, len);
             free(buffer);
+        }
         capacity = newSize - 1;
         buffer = newbuffer;
         return 1;


raheelh:
The method String::changeBuffer() uses double the amount of memory it should use. I wish if we could use memory more optimally we'd be able to do more. For e.g. we could concat bigger json web data etc.

I was trying to replace malloc() & strcpy() by realloc() but my tests find that I don't end up saving any runtime memory. That's weird.

char *newbuffer = (char *) realloc(buffer, newSize);
if(newbuffer) {
  size_t oldSize = capacity + 1; // include NULL.
  memset(newbuffer + oldSize, 0, newSize - oldSize);
  capacity = newSize - 1;
  buffer = newbuffer;
  return 1;
}
buffer = newbuffer;
return 0;

Is realloc() not a real realloc()? If it did actually extended the existing memory block, I should have concatinated a longer string which I could not.

Possible explanations I can think of are 1) realloc() is internally faking by doing a malloc() & strcpy(), or 2) I did not have enough memory available when I performed the experiment. It's hard to believe the latter. It seems likely that realloc() is not behaving the way I would expect it to.

Any ideas?

igrr:
umm_realloc will attempt to extend the buffer if it is possible, i.e. the adjacent block is not allocated.
You can check umm_malloc source code (it is included in this repository).

raheelh:
Makes sense. I just created a pull request with a change to leverage realloc() in this method.

PR #1686, opened 2016-02-26
Added support for printing uint64_t numbers

mykeul:


igrr:
The issue with this implementation of printing 64-bit numbers is that every other number (8, 16, 32-bit) will be converted to 64 bit before printing. This is a pretty big overhead, especially since most users will not be actually printing 64 bit numbers.
It has been suggested on Arduino developers mailing list [1] that the right way to implement this feature is to write separate routines for 64-bit case, so that smaller sized integers still go through the 32-bit `printNumber` function. Besides, we are much less constrained on code size than AVR based Arduinos, so having an extra function for the 64-bit case *if* it is actually used is not an issue.

[1] https://groups.google.com/a/arduino.cc/d/topic/developers/Royw6V4ictU/discussion

mykeul:
Hi Ivan,

I updated my branch. printNumber() is now a template and produces two versions of the function : printNumber\<unsigned long\>() and printNumber\<uint64_t\>(). Shorter (16 or 8 bits) versions may be used too, but I didn't moved that far as it may use more flash on some projects.

MyKeul

igrr:
Waiting for https://travis-ci.org/esp8266/Arduino/builds/122758554 to check code size difference.

PR #1735, opened 2016-03-08
fix umm_malloc null ptr.

Links2004:
(endless reboot) like:
```
Fatal exception (28):
epc1=0x402097e8, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000001, depc=0x00000000
Fatal exception (28):
epc1=0x402097e8, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000001, depc=0x00000000
Fatal exception (28):
epc1=0x402097e8, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000001, depc=0x00000000
```

```
0x402097e8: umm_info at umm_malloc/umm_malloc.c line 1000
```

### DIFF
--- a/cores/esp8266/umm_malloc/umm_malloc.c
+++ b/cores/esp8266/umm_malloc/umm_malloc.c
@@ -971,6 +971,10 @@ void ICACHE_FLASH_ATTR *umm_info( void *ptr, int force ) {
 
   unsigned short int blockNo = 0;
 
+  if (umm_heap == NULL) {
+      umm_init();
+  }
+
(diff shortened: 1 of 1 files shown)


igrr:
Probably need to send that upstream as well...

Links2004:
yes, the SDK config can come in a state where it trigger this endless, 
the sketch code then never will be called.

igrr:
I think we can release a 2.1.1 as quite a few issues have been fixed already.

PR #1741, opened 2016-03-09
Extended mDNS-SD support (sending queries and receiving answers)

larsenglund:
Extended mDNS-SD support that allows users to send queries for services and receive the answers.


probonopd:
Thanks 👍

PR #1760, opened 2016-03-12
Allow indefinite duration for `tone()`

urish:
Bugfix: the `duration` parameter should be optional, per [`tone()`](https://www.arduino.cc/en/Reference/Tone) docs.


urish:
Now this PR also fixes #1761

Issue #1149, opened 2015-12-05
pulseIn cause Nodemcu to reset with some pins.


Leen15:
Hi, i was trying to make working a SRF04 with my nodemcu and i notice something strange that reboot the board. So i tried to minimize my code for find the problem, and appear that pulseIn is causing the board to reset with some pins. (i don't know how).

With this code, and nothing connected to the board:

```
const int ECHO_PIN = D3; // WORKING WITH: D1,D2,D8 -  NOT WORKING WITH: D3,D4,D5,D6,D7,D9,D10       
void setup() {
   // initialize serial communication:
    Serial.begin(115200);       
    pinMode(ECHO_PIN,INPUT);  
}

void loop() {
  long duration;
  duration = pulseIn(ECHO_PIN,HIGH);
  Serial.println(duration);  
  delay(100);
}
```

I read in serial values of type 1 or 2 and the board keep running with ECHO_PIN set to D1,D2, or D8, but with pins D3,D4,D5,D6,D7,D9,D10 I receive instead something like:

```
1000001
1000001
1000000

ctx: cont 
sp: 3ffe9aa0 end: 3ffe9c90 offset: 01b0

>>>stack>>>
3ffe9c50:  00000000 00000000 00000016 3ffe9cbc  
3ffe9c60:  3fffdc20 00000000 3ffe9cb4 40201d90  
3ffe9c70:  00000000 00000000 00000000 402018ea  
3ffe9c80:  00000000 00000000 3ffe8c70 40100118  
<<<stack<<<

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 1264, room 16 
tail 0
chksum 0x42
csum 0x42
~ld
```

and it reset every 3-4 prints.

Did i miss something in the sketch? Why only 3 pins work?

PS: I tried with 3 different nodemcu boards, ESP12E.

Thanks
##

framb00s:
Tnx for the info.

I get strange results with dustPin set to D3
void loop() {
  duration = pulseIn(dustPin, LOW);
  if(duration!=0){                                                    // number positive readings
    i++;
  }

when I run this program on a arduino (nano or uno) it works OK
but on the nodeMCU I get false readings.
I'll try with D1,3 or 8 and see if this works.

DiApostola:
[...]

Issue #1072, opened 2015-11-22
pulseIn does not return 0 on timeout (it returns 1 or timeout value depending on pin state)


darkskiez:
In core_esp8266_wiring_pulse.c, it seems to return duration=timeout rather than duration=0 on a timeout, or, because it resets the start time after a timeout, it can also return duration=1.
##

darkskiez:
Whilst you're in that file, you could remove the pinMode setting, I may have set PULLUP etc.

framb00s:
I have the same problem
const int dustPin= D2;
pinMode(dustPin,INPUT_PULLUP);
int i=0;
void loop() {
  duration = pulseIn(dustPin, LOW);
  if(duration!=0){                                                    // number positive readings
    i++;
  }
  lowpulseoccupancy = lowpulseoccupancy+duration; 
.....
give with nothing attached after 30 seconds 30 reads (i==30)
and a total pulse-low_time between 30 and 35

devyte:
@igrr the referenced PR was merged quite some time ago. Can this be closed, or is there something pending?

devyte:
@me-no-dev I just realized this is assigned to you, maybe you're the right person to ask?

me-no-dev:
Yeah I have no clue how and when I assigned this to myself. I guess it should be closed, unless the current implementation also does not work

devyte:
Alright, closing. If there is still a problem, let's handle it in a new issue.

PR #1768, opened 2016-03-13
Fix pulseIn

igrr:
#1072 
#1149

- fix return value on timeout
- add optimistic_yield to allow WiFi/network tasks to run
- use ccount instead of micros (which causes an integer division on every call)


mangelajo:
looks good to me.

PR #1771, opened 2016-03-14
Update index.html

gmag11:
Fix context menu due to wrong variable name "event" --> "e"


gmag11:
There is an error in file. Event is stored in function parameter "e" but it is referenced as "event" in two lines. This makes that context menu does not appear.

PR #1781, opened 2016-03-16
add CPOL setting

reiyawea:
Add CPOL setting thus supporting SPI MODE 2&3.

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -102,7 +102,10 @@ void SPIClass::setDataMode(uint8_t dataMode) {
     }
 
     if(CPOL) {
-        //todo How set CPOL???
+        SPI1P |= 1<<29;
+    } else {
+        SPI1P &= ~(1<<29);
+        //todo test whether it is correct to set CPOL like this.
     }
 
 }


0xPIT:
I've tested just now with my logic analyzer. Unfortunately, it does not work.
I've also tried all modes (0,1,2,3) and SCK always idles in low state, whereas with `CPOL=1` (in modes 2 & 3) it should idle at high.

Where did you find the info about bit 29 in `SPI1P` is setting `CPOL`?

reiyawea:
Sorry for replying so late. It's difficult to access gmail here.
First, a search on bing gives that setting bit 29 in SPI_PIN register will
make CPOL=1. (See: http://bbs.espressif.com/viewtopic.php?t=1570)
It seems to be a different register naming system. Then I reviewed the
codes of other ESP8266 projects and find that one of them uses the same
naming system. It's called NodeMCU.
By matching the absolute address I found that SPI1P here is the same
register as SPI_PIN there.
I use the modified code to light up a VFD which uses SPI MODE2 and
succeeded.
Hope it helps.

2016-03-27 5:42 GMT+08:00 0xPIT <notifications@github.com>:

reiyawea:
I probed the SCK pin of SPI MODE2 and found that it was truly idle at high.

​

2016-04-03 23:00 GMT+08:00 高轶凡 <yfgao1009@gmail.com>:

Issue #1790, opened 2016-03-17
ESP8266WebServer::authenticate() makes the sketch crash.


dancol90:
#### Hardware
Hardware:			ESP-12E (NodeMCU)
Core Version:      	2.1.0

### Description

When `ESP8266WebServer::authenticate()` is called, the program crashes. By analyzing the stack, it seems related to sprintf and delete[] commands inside the web server library code.

The sketch loads fine, it connects to wifi, but after connecting with a browser and entering auth info (right or wrong, doesn't matter), the program crashes. A stack dump is provided below.

Looking at `ESP8266WebServer::authenticate()` code, and after some experiments, it seems that the program breaks when calling `delete[] toencode`. Removing the `sprintf` call inside the function avoids the crash (obviously the function will return always false, but whatever).

Also, rewriting the function using Arduino's `String` objects instead of char buffer formatting makes the crash disappear.

Sometimes the program crashes before the browser asks for a username and a password.

### Settings in IDE

Same problem is observed using PlatformIO.

Module:  			Generic ESP8266 Module
Flash Size: 		4MB/1MB
CPU Frequency:		80Mhz
Flash Mode:			qio
Flash Frequency:	40Mhz
Upload Using:		SERIAL
Reset Method:		nodemcu

### Sketch

Every time HTTP authentication is used as is intended to be used. The simplest example is the HttpBasicAuth example, provided with the ESPWebServer lib.

```cpp
server.on("/", [](){
  if(!server.authenticate(www_username, www_password))
    return server.requestAuthentication();
    server.send(200, "text/plain", "Login OK");
});
```

### Stack dump

```
0x40101daa: pp_post at ?? line ?
0x40101daa: pp_post at ?? line ?
0x40107080: printf at /Users/daniele/Library/Arduino15/packages/esp8266/hardware/esp8266/2.1.0/cores/esp8266/libc_replacements.c line 206
0x40102c4c: trc_NeedRTS at ?? line ?
0x40227be2: ets_vsnprintf at ?? line ?
0x402093f6: String::copy(char const*, unsigned int) at /Users/daniele/Library/Arduino15/packages/esp8266/hardware/esp8266/2.1.0/cores/esp8266/WString.cpp line 177
0x4010031d: check_poison at /Users/daniele/Library/Arduino15/packages/esp8266/hardware/esp8266/2.1.0/cores/esp8266/umm_malloc/umm_malloc.c line 817
[...]

PR #1794, opened 2016-03-18
Fix crashes caused by http authentication

kzyapkov:
I have a sketch which consistently crashes when correct HTTP credentials are provided in the `Authorization` header. Experiments showed that the crashes are caused by the `delete[]` statements for `toencode` and `encoded`. I wasn't able to decipher the core dumps.

Trial-and-error lead to the solution here. I have no explanation why it works, but definitely fixes the world for me.

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -22,7 +22,7 @@
 
 
 #include <Arduino.h>
-#include <libb64/cencode.h>
+#include "base64.h"
 #include "WiFiServer.h"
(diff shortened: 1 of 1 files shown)


igrr:
I'd advise against use of VLA. Since we don't have a way to reliably catch stack overflow issues, there is no way to make sure variable-length stack allocation succeeds (and crash if it fails). So it's better to use either static fixed-length arrays or dynamically allocated arrays. I think to fix the issue it is sufficient to change `+1` with `+2` in `toencodelen` calculation.

igrr:
https://github.com/esp8266/Arduino/issues/1790#issuecomment-198337941

kzyapkov:
Okay!

PR #1817, opened 2016-03-25
Add a start callback for HTTP updater

baruch:
The callback is called when the upgrade actually starts rather than just
the initial query so that the user can know that it will not take longer
and can also prepare for the upgrade by shutting down other works.

### DIFF
--- a/libraries/ESP8266httpUpdate/examples/httpUpdate/httpUpdate.ino
+++ b/libraries/ESP8266httpUpdate/examples/httpUpdate/httpUpdate.ino
@@ -17,6 +17,10 @@
 
 ESP8266WiFiMulti WiFiMulti;
 
+void update_starting(void) {
+    Serial.println("Update starting");
+}
+
 void setup() {
 
     USE_SERIAL.begin(115200);
@@ -40,6 +44,7 @@ void loop() {
     // wait for WiFi connection
     if((WiFiMulti.run() == WL_CONNECTED)) {
 
+        ESPhttpUpdate.onStart(update_starting);
         t_httpUpdate_return ret = ESPhttpUpdate.update("http://server/file.bin");
         //t_httpUpdate_return  ret = ESPhttpUpdate.update("https://server/file.bin");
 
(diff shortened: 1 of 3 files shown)


igrr:
This almost looks good to me, but I would propose to replace plain C function pointer with an `std::function<void(void)>`, similar to what we do in ESP8266WebServer library. 
Obvious benefit is that you can bind a member function to a pointer/reference using `std::bind`, and pass the result into std::function. With plain C function pointer there is no way to do that.

mangelajo:
If you had time to add progress and end callbacks like in ArduinoOTA, that'd be great.

Cheers!,

baruch:
The problem with progress and end is that at that time there is no TCP/UDP active as they are cancelled by the upgrade process.

As for the method calls, I can change to that, they are not that common in Arduino code though and all other callbacks throughout the system is normal function calls too.

mangelajo:
@baruch , but you don't need any other open TCP/UDP connection to show progress, the use case is device itself showing on a display/leds. I use it in all my projects so the end user can have some indication of the upgrade progress.

I'd be glad to see an standard interface for those callbacks for consistency within esp8266-arduino at least.

krzychb:
@esp8266-arduino-bot,
You are not a human, are you?
Great avatar :smile: :+1:

igrr:
Sorry for the noise. My regression testing setup is misbehaving.

igrr:
@baruch just to clarify regarding callback functions: if an argument is an `std::function<void(void)>`, you *can* pass plain C function pointer and it will be implicitly converted. This will behave just like other Arduino functions, `attachInterrupt` for example. However, this also allows the user to pass in any other function object (lambda function, anything that std::bind returns, or even something custom). So in general, replacing a C function pointer argument with a corresponding std::function won't hurt.

karlp:
After just running into this lack in the http update server, I've started work on this based on the ArduinoOTA style.

earlephilhower:
The core and libraries have changed enough that this PR now has a merge conflict.

earlephilhower:
Closing as abandoned for #6796

Issue #1818, opened 2016-03-25
Fix to SPIFFS openDir() enumeration ..


seclorum:
There is a bug in SPIFFS.

This simple tweak to SPIFFS:

EDIT: PR here https://github.com/esp8266/Arduino/pull/1819

````
DirImplPtr SPIFFSImpl::openDir(const char* path) {
    if (strlen(path) > 0) {
        if (!isSpiffsFilenameValid(path)) {
            DEBUGV("SPIFFSImpl::openDir: invalid path=`%s` \r\n", path);
            return DirImplPtr();
        }
    }
````

.. means that we can do this:

````
  POVDir = SPIFFS.openDir("");
````

.. and get the 'root' - i.e. all filenames in the SPIFFS

Note: without this code, there is no way to get files you put into SPIFFS that are *not* named with the prefix '/', for example .. which is not enforced anywhere else in SPIFFS, so you can in fact put files on SPIFFS that you can't retrieve, at the moment ..

PR #1819, opened 2016-03-25
Update spiffs_api.cpp

seclorum:
Fixes a bug where un-prefixed files are irretrievable with openDir(""). Described: https://github.com/esp8266/Arduino/issues/1818.

### DIFF
--- a/cores/esp8266/spiffs_api.cpp
+++ b/cores/esp8266/spiffs_api.cpp
@@ -63,12 +63,14 @@ bool SPIFFSImpl::exists(const char* path)
     return rc == SPIFFS_OK;
 }
 
-DirImplPtr SPIFFSImpl::openDir(const char* path)
-{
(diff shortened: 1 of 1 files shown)


igrr:
You may need to update the test as well. Old behaviour (`openDir("")` should return empty list) was checked here: https://github.com/esp8266/Arduino/blob/master/tests/host/fs/test_fs.cpp#L163-L164

hallard:
Just my 2 cents, why not change the original  `isSpiffsFilenameValid` where is the problem, checking len >0 by adding `*name == 0` in the 1st if, like this ?
```
bool isSpiffsFilenameValid(const char* name)
{
    if (name == nullptr || *name == 0) {
        return false;
    }
    auto len = strlen(name);
    return len > 0 && len < SPIFFS_OBJ_NAME_LEN;
}
```

igrr:
Reopened as #2143, with test cases fixed.

PR #1827, opened 2016-03-28
Prevent exception by detaching Timer0 interrupt when updating

0xPIT:
I use Timer0 for my encoder. When the timer is active and being reloaded after firing, updates fail with an exception. Detaching the timer fixes this problem. (Verified on bare ESP-12)


igrr:
Is your ISR handler marked with ICACHE_RAM_ATTR?

This change doesn't look good to me, because 1) you probably need to detach all other interrupts as well, and 2) this should also be done for other update methods (ArduinoOTA for example).

0xPIT:
Of course you are right, I've missed out the ICACHE_RAM_ATTR.
Thanks.

PR #1829, opened 2016-03-29
WifiClient example - Fix issue where not all lines are read from slow servers

rogerclarkmelbourne:
The while loop which reads the response from the server can finish before all data has been sent by the server, if the server returns the data slower than the code is reading it from the buffer.

As in the example, the server is requested to close the connection after the data has been sent, the problem can be resolved by checking if the client is still connected to the server as well as checking if there is data in the client buffer using client.available()

### DIFF
--- a/libraries/ESP8266WiFi/examples/WiFiClient/WiFiClient.ino
+++ b/libraries/ESP8266WiFi/examples/WiFiClient/WiFiClient.ino
@@ -81,7 +81,7 @@ void loop() {
   }
   
   // Read all the lines of the reply from server and print them to Serial
-  while(client.available()){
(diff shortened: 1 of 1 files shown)


devyte:
This PR looks correct to me, and the relevant code is still present in the example in latest git. 
In theory, a webserver could send data line by line. If after sending one line it takes too long to send the next, and the relevant while loop in the example empties the client buffer, on next iteration it will stop looping and be done, even though the reply is incomplete. Given that the headers specify Connection: close, it's ok to continue looping until connection is closed by the server.
@rogerclarkmelbourne the relevant branch is out of date with base branch. Could you please rebase?

PR #1836, opened 2016-03-30
Added support for RTC user memory in ESP-specific APIs.

MacroYau:


igrr:
It would be really cool if we could remove the limitation of `system_rtc_mem_read`/`write` that pointers have to be 4-byte aligned. We are doing something similar in SPIFFS HAL, where we split reads and writes to make sure they are aligned: https://github.com/esp8266/Arduino/blob/master/cores/esp8266/spiffs_hal.cpp#L31-L88.
That, and also would be great to add `offset` parameter so that users could write something not only at the beginning of RTC memory, but at some offset as well.

PR #1839, opened 2016-03-31
Update BasicOTA.ino

Testato:
- removed unnecessary #include
- changed \n with .println because \n not work for me on Arduino Ide 1.6.8


madleech:
Note that on platform.io, the #includes are required; using this sketch it will fail to compile.

igrr:
These includes are required when compiling with platformio and with Arduino IDE < 1.6.6.

PR #1842, opened 2016-04-01
Added extra method for reading current 'passphrase'

kehribar:
Hi all,

Firstly, thanks for all the hard work you put into this library.

I added an extra method for reading currently stored SSID passphrase. The placement may not be the optimal but in the end it is functional. :)

This is useful for reading the password obtained from esp-touch and then storing it to the internal filesystem. SSID name reading has been already implemented. 

Someone else might find this useful too, so I'm sending this pull request.

Best,
ihsan.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFi.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFi.cpp
@@ -93,8 +93,8 @@ String ESP8266WiFiClass::passphrase()
 {
     struct station_config conf;
     wifi_station_get_config(&conf);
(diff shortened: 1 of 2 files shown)


igrr:
Thanks for the pull request, but we already have one :)
https://github.com/kehribar/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h#L74

(not documented anywhere and not used in any example)

kehribar:
Even though I searched inside the source files, I couldn't manage to find that one! Thanks for pointing out :) I'll discard the change that I made and use this one instead. 

Is there any particular reason why it is not documented? Other can find it useful too.

PR #1850, opened 2016-04-03
Allow PSK instead of passphrase in WiFiSTA::begin

4m1g0:
In WPA protocol, the maximum length of the passphrases is 64 characters in order to distinguish them from the actual PSK which is 64 ASCII characters long, so in most systems if a 64 chars string is passed, it is assumed to be a PSK, otherwise is treated as a passphrase and is used to compute the PSK.


4m1g0:
I've tested nodemcu firmware and it allows this too. I think it's standard.

PR #1860, opened 2016-04-05
Changed timeout logic in wificlient example

hemalchevli:
Changed timeout to unsigned long. Using addition with millis() is not recommended. 
Source: http://www.gammon.com.au/millis


chaeplin:
I think this has problem.
```
 if (millis() - timeout < 5000) {
```

should be ?
```
 if (millis() - timeout > 5000) {
```

hemalchevli:
You are right @chaeplin. I made a mistake. thanks for correcting.

PR #1917, opened 2016-04-16
remove malloc / realloc Parsing.cpp

gpepe:


libraries/ESP8266WebServer/src/Parsing.cpp
```diff
@@ -138,40 +138,33 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
...
+        // some clients send headers first and data after (like we do)
+        // give them a chance
+        int tries = 1000;  // 1000ms max wait
         size_t newLen;
-        while( !(newLen = client.available()) && tries--) delay(1);
+        while (!(newLen = client.available()) && tries--) delay(1);
         if (!newLen) break;
-        plainBuf = (plainBuf == nullptr) ?  (char *) malloc(newLen + 1) : (char *) realloc(plainBuf, plainLen + newLen + 1);
-        client.readBytes(&plainBuf[plainLen], newLen);
+        char plainBuf[newLen + 1];
```
igrr: This can potentially overflow the stack, and you have no way of checking whether this happens.
Quoting `man alloca`:

    The alloca() function returns a pointer to the beginning of the
    allocated space.  If the allocation causes stack overflow, program
    behavior is undefined.

VLAs basically use the same alloca mechanism internally.

PR #1922, opened 2016-04-17
Rename WiFiUdp.h to WiFiUDP.h

fenderrex:
UDP stands for 'User Datagram Protocol' don't  camelcase abbreviations it makes a mess further more it pissed me off for an hour untill i found it was just slightly renamed.


WereCatf:
@fenderrex You can't just rename the file and not change all the references to it in the existing code, like e.g. sketches and other libraries -- you just broke all the example sketches that use WiFiUdp.h, for example.

igrr:
Not to mention any existing sketches that users have on their computers — these will be broken as well.

WereCatf:
Fixing something like this after the mistake has been made and lingered on for a while already is just not that easy anymore once there's code out there that use the mistaken way.

@igrr How would you feel about a dummy WiFiUDP.h instead that just includes the WiFiUdp.h? It wouldn't break anything, but would allow more generic Arduino-sketches and libs to function without changes if they use the standard WiFiUDP.h.

I don't even know if this is a problem worth fixing, but a dummy like that certainly doesn't require much work on anyone's part.

igrr:
The trouble with dummy file is that when you select ESP8266WiFi from "add library" menu, it will add both header files to the sketch.

Another thing is, with such change we break compatibility with WiFi shield library:
https://github.com/arduino/Arduino/blob/master/libraries/WiFi/src/WiFiUdp.h

WereCatf:
Oh, right. Scratch that idea, then.

Issue #1969, opened 2016-04-24
Debug tracing menu items not shown for NodeMCU board


innesm:
If NodeMCU board is selected in the IDE, no debug menus (for configuring debug serial output) are shown.
#### Hardware

Hardware:           NodeMCP v1.0
Core Version:       latest esp8266 from github?
### Description

The required menu defs are just missing from the NodeMCU section of boards.txt. It was possible to get these added with some copy and pasting.
### Settings in IDE

Module:             NodeMCU v1.0
##

tkerby:
Likewise for the Wemos boards

davisonja:
This just needs the entries copied in boards.txt;
I've wondered, given the repetition if it might be worth having a script to
generate the file, which would avoid having to edit all the (other than key
name) identical bits for any and every change.
I've duplicated it for the NodeMCU (which I've been using) so it is just as
simple as copying and pasting the lines (updating the key names).

igrr:
Generating boards.txt during build phase (i.e. building boards manager package or bootstrapping a version cloned from git) is the way to go, I think. This also dovetails with your suggestion to keep all flash geometry info in boards.txt. Maybe we can keep it in the linker scripts, and generate boards.txt automatically? Or have another file with all the flash sizes data and generate boards.txt and linker scripts from it.

davisonja:
I had wondered about that too, instead of complicating the command lines a
package-build step which would automate the lot including the various menu
options. (Though ultimately a more hierarchical config system for the IDE
would be nice)
If that were the way forward, what do we write such a script in? Are there
any existing similar scripts? I'd rather not add too many arbitrary
dependencies :)

davisonja:
I'd vote for Python (would have been my first choice :) so shall I organise
such a script in that in preference to adjusting linker command lines?

igrr:
This script would allow us to handle both the linker scripts issue and this one, so I would vote for writing the script. In the past, we had some bugs due to incomplete/inconsistent updates to boards.txt, and such a script could help as well. It will also make adding new options and boards easier.
Input data for this script can also be given directly in Python source code, to avoid defining another format, parsing it, and so on.

devyte:
The Menus are undergoing a full rework, which is tracked in #3582 . Closing.

PR #1970, opened 2016-04-24
Add debug menus to NodeMCU board.

innesm:


innesm:
https://github.com/esp8266/Arduino/issues/1969

igrr:
Will be handled by the boards.txt generator.

PR #1987, opened 2016-04-29
Add ESP8266WifiScanClass::getScanCount

mettke:
Adding getScanCount to be able to retrieve _scanCount after using scanNetworks


igrr:
I think this function has to return some error code if the scan is still in progress. But if you add that, it will be an exact duplicate of `scanComplete` function.

mettke:
Sorry, I completely overlooked the scanComplete function. 
You're right the scanComplete function does the same functionality even though its naming is a bit misleading.
Thanks for clearing up

PR #1996, opened 2016-05-05
Reworked Wifi connection examples

hallard:
Revamped all samples Wifi connection process to test sketch default ssid and password vars has been
filled with user SSID/PSK. If not (both vars start with *) Wifi try to
connect with latest known SDK Config so if ESP8266 was already conected
to your WiFi, compiling sketch "as is" will connect to user AP


WereCatf:
In NTPClient.ino you are referencing password, but the variable is named pass. In multiple sketches you keep calling WiFi.begin() over and over while waiting for connection, but it's redundant, it won't speed connection-attempts up. Also, if the user hasn't specified a custom username or password you don't actually call WiFi.begin() at all!

hallard:
@WereCatf thanks pointing this out, I modified a bunch of files I missed that one!

By the way according to what you wrote, the following code should be okay?

```arduino
  // Put your own Wifi Router SSID/KEY here
  const char* ssid = "**********";
  const char* password = "**********";

  // If sketch as no default SSID and PSK (both start with *)
  // it should try to connect to SDK saved one (if any)
  if (*ssid!='*' && *password!='*')
    WiFi.begin(ssid, password);

  while (WiFi.waitForConnectResult() != WL_CONNECTED){
    delay(1000);
    Serial.println("Retrying connection...");
  }
```

 If so I will change my modification to be consistent with this one on all samples I already changed.

Makes sense for you?

WereCatf:
Well, I don't really know why you even changed the while-loop in the first place. If we were to change it I'd use a bit more verbose code myself, like e.g. the following:

```
Serial.print(F("Waiting for WiFi-connection."));
while (WiFi.waitForConnectResult() != WL_CONNECTED){
  delay(500);
  Serial.print(F("."));
}
Serial.println(F(" Connected."));
```

I just don't know if all this work is worth the effort.

hallard:
Yeah sounds good too, at the beginning, I changed to have the examples to be consistent with each other but I'm often trying examples on already working ESP8266 target and each time, to have them working I need to:
- To load them in IDE
- Uncheck in IDE 'use external editor'
- Change SSID/PASS
- Compile and upload 
- Check back 'use external editor'
(and repeat also next time I need to test same sample code). Of course I could also edit source file but it's far on folder tree (I'm on Windows) and if I save the file my next `git pull` will fail because of modified file so I though changing source was the easy way...

The changes I made are just to check that if psk/pass are unchanged in sample source code (******) then just try to connect with SDK saved one (if any of course) and not with *****/******* which for sure will not connect ;-)
now just to do  
- To load them in IDE
- compile and upload 

Not sure it worth the effort, I agree but once done, it's done ;-)

WereCatf:
Well, you won't see me complaining: if you wish to improve the stuff that's already there, even a very minor improvement, then I say go for it! :)

igrr:
How about setting default SSID and password combination to invalid values?
I.e. have password shorter than 8 characters. Then if you don't change
default values, WiFi.begin will fail to set new values, and any previous
setting will be used instead. This way you don't have to add any "magic"
code to the examples.

hallard:
@igrr 
Love this one, I'll set SSID/PASS to 7 stars and give you the 8th ;-)

hallard:
@igrr 
I've tried just in case, SDK contains my old settings that are working when I call `WiFi.begin()` but if I call with `WiFi.begin("*******","*******")` it won't connect even if values are invalid (<8 char) and  values are stored in config ;-(

Here the sketch I'm using below results
```arduino
void setup()
{
  Serial.begin(115200);
  Serial.println("");

  // Show SDK stored values
  Serial.print("SDK SSID : "); Serial.println(WiFi.SSID());
  Serial.print("SDK PSK  : "); Serial.println(WiFi.psk());

  WiFi.begin(ssid, password);
  while (WiFi.waitForConnectResult() != WL_CONNECTED){
    delay(500);
    Serial.printf("%5ld : ...\n", millis());
  }
```
Here are the results on serial
- First run ssid and password contains my real one and connect fine
- now set ssid and password to 6 stars each (so wrong) and flashed back sketch (but SDK contains correct ones)
```
SDK SSID : CH2I-HOTSPOT
SDK PSK  : My@Password
12445 : ...
12945 : ...
13445 : ...
13945 : ...
```
Won't connect even if ssid/pass are wrong (<8 char)?

So I just reset the board (no upload)
```
SDK SSID : ******
SDK PSK  : ******
12445 : ...
12945 : ...
13445 : ...
13945 : ...
```
seems params have been saved? Does I'm missing something ?

hallard:
Ok guys after some testing seems this code works fine every time

```arduino
  // If sketch as no default SSID and bad PSK (len <8)
  // it should try to connect to SDK saved one (if any)
  Serial.print(F("Connecting with SSID of "));
  if ( strlen(password)<8 ) {
    Serial.printf( "SDK '%s'\r\n", WiFi.SSID().c_str() );
    // If autoconnect is disabled force connection
    if (!WiFi.getAutoConnect()) 
      WiFi.begin();
  } else {
    Serial.printf( "Sketch '%s'\r\n", ssid ); 
    WiFi.begin(ssid, password);
  }
  // Will be blocked in this while loop until connected
  while (WiFi.waitForConnectResult() != WL_CONNECTED){
    delay(1000);
    Serial.printf("%4ld sec : ...\n", millis()/1000);
  }
```

I've update all needed examples with this one
may be adding new method `connect(...)` to ESP8266WiFiSTA.cpp encapsulating all this stuff and adding more feature such as time out worth it.

igrr:
Don't get me wrong, i can see that it is convenient not to have edit SSID and pass for every sample you upload. But adding 15 lines of code to each sample for that purpose is not the way to go, i think.
This should probably be fixed in WiFi.connect implementation, so that if an invalid (empty?) ssid/password are provided, the thing still connects to previously stored config, if any.

hallard:
@igrr I',m not also fan of cuting and pasting so adding new method sounds good to me. But this mean that all samples need to call this new method right? Also I need to update the documentation to add this method.
If it's okay for you I'll do it.

by the way most of the samples was containing the while loop waiting for connection, and some debug print so basically in term of coding; I just replaced (and for sure added comments)
```arduino
    WiFi.begin(ssid, password);
```
by 
```arduino
  if ( strlen(password)<8 ) {
    if (!WiFi.getAutoConnect()) 
      WiFi.begin();
  } else {
    WiFi.begin(ssid, password);
  }
```
;)

hallard:
Oh, sorry, forgot to mention, no real link between, but this does not explain why providing bad ssid/pass (<8 char) is still saved by SDK ?

hallard:
@igrr 
I was Thinking about this, I'm not sure creating connect method is the best idea, I think it will just confuse end user than helping them. May be some little change in begin() could be a better option.
Let's says something if checking SSID/Password are both set to 7 stars then connect with SDK saved credentials, which could do in ESP8266WiFiSTA.cpp begin something like that
```arduino
    // If sketch as uninitialized SSID and PSK "*******" 7 stars
    if(passphrase && ssid && strlen(passphrase)==7 && strlen(ssid)==7 ) {
      if (strncmp(ssid,"*******",7)==0 && strncmp(passphrase,"*******",7)==0)
        // it should try to connect to SDK saved one (if any)
        return begin();
    }

```
Using this, nothing to change in sketch examples except setting ssid and pass to 7 stars
Does this looks better for you ?

PR #1999, opened 2016-05-06
SSDP Response broken

b-pass:
The SSDP response template and sprintf call are out of sync, and so the response that is sent back is partly garbled.


b-pass:
Sorry, duplicate of #1976

PR #2008, opened 2016-05-09
UdpNtpClient rewritten in a clearer, more pedantic fashion.

mgaman:
I have modified the classic UdpNtpCLient sketch by adding the structure of the Ntp data packet (as described in RFC1305) and referring to fields within the structure.
My objective is to encourage the writing of code easier to maintain, especially by someone other than the original author.

### DIFF
--- a/libraries/Ethernet/examples/udpntpclient_pedantic/RFC1305.h
+++ b/libraries/Ethernet/examples/udpntpclient_pedantic/RFC1305.h
@@ -1,3 +1,5 @@
+#ifndef RFC1305_H
+#define RFC1305_H
 /*
  * see https://www.eecis.udel.edu/~mills/database/rfc/rfc1305/rfc1305c.pdf
  * https://tools.ietf.org/html/rfc1305
@@ -44,4 +46,5 @@ struct sRFC1305 {
 
 #define ENDIAN_SWAP_32(l) ((l>>24) |((l>>16)<<8)&0xff00 | ((l>>8)<<16)&0xff0000 | (l << 24)) 
 #define ENDIAN_SWAP_16(l) ((l>>8) | (l << 8)) 
+#endif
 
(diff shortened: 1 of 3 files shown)


libraries/Ethernet/examples/udpntpclient_pedantic/RFC1305.h
```diff
@@ -0,0 +1,47 @@
+/*
+ * see https://www.eecis.udel.edu/~mills/database/rfc/rfc1305/rfc1305c.pdf
+ * https://tools.ietf.org/html/rfc1305
+ */
+
```
igrr: It is considered common good practice to start every header file with include guards.

libraries/Ethernet/examples/udpntpclient_pedantic/UdpNtpClient_Pedantic.ino
```diff
@@ -0,0 +1,160 @@
...
+  // Open serial communications and wait for port to open:
+  Serial.begin(9600);
+  while (!Serial) {
+    ; // wait for serial port to connect. Needed for Leonardo only
+  }
+  //Serial.println(NTP_PACKET_SIZE); // just for debugging
+  //Serial.println(ENDIAN_SWAP_32(0x11223344),HEX);
+  //Serial.println(ENDIAN_SWAP_16(0xAABB),HEX);
+  // start Ethernet and UDP
+  if (Ethernet.begin(mac) == 0) {
```
igrr: Ethernet? I feel adding an example which uses WiFi would be more relevant to the scope of this project.

mgaman: The original client was in the Ethernet library and then copied to the Wifi library. Just like I updated the Ethernet version I could just go ahead and apply the same to the Wifi version.

PR #2016, opened 2016-05-11
Fix #2015 ESP8266mDNS doesn't accept queryService responses from avahi

### LINKED ISSUE #2015: ESP8266mDNS doesn't accept queryService responses from avahi-daemon
Avahi-daemon includes the IPv6 of the server in its responses, and ESP8266mDNS can't parse them.

unaiur:
Ignore unknown records in the query response; this way we can extract
the IPv4 address and connect to the server.

### DIFF
--- a/libraries/ESP8266mDNS/ESP8266mDNS.cpp
+++ b/libraries/ESP8266mDNS/ESP8266mDNS.cpp
@@ -470,8 +470,8 @@ void MDNSResponder::_parsePacket(){
     }
 
     int numAnswers = packetHeader[3];
-    // Assume that the PTR answer always comes first and that it is always accompanied by a TXT, SRV and A answer in the same packet.
-    if (numAnswers != 4) {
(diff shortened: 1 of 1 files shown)


libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -589,12 +589,19 @@ void MDNSResponder::_parsePacket(){
...
-      if (answerType == MDNS_TYPE_A) {
+      else if (answerType == MDNS_TYPE_A) {
         partsCollected |= 0x08;
         for (int i = 0; i < 4; i++) {
           answerIp[i] = _conn_read8();
         }
       }
+      else {
+#ifdef MDNS_DEBUG_RX
+          Serial.printf("Ignoring unsupported type\n", tmp8);
```
igrr: You are passing a parameter to printf, but there isn no corresponding format specifier in format string.

unaiur:
Sorry, I mistakenly deleted my master branch.

PR #2023, opened 2016-05-13
Fix #2015 ESP8266mDNS doesn't accept queryService responses from avahi

### LINKED ISSUE #2015: ESP8266mDNS doesn't accept queryService responses from avahi-daemon

unaiur:
Ignore unknown records (AAAA) in query responses sent by avahi; this way we can extract
the IPv4 address and connect to the server.


unaiur:
Resynchronized with master branch.

PR #2054, opened 2016-05-24
Real board name available in Sketch/MDNS/OTA

hallard:
- Changed WifInfo defaults settings
- Changed WeMos defined board name not to be ESP8266_NODEMCU but WEMOS_D1_MINI and WEMOS_D1
- Added board name defined symbol at compilation stage to have it available in sketch, MDNS and OTA using ARDUINO_BOARD (like in [ESP8266mDNS.h](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266mDNS/ESP8266mDNS.h#L51))
This avoid board name to be "generic" in case of known board

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -944,7 +944,7 @@ d1_mini.serial.disableRTS=true
 
 d1_mini.build.mcu=esp8266
 d1_mini.build.f_cpu=80000000L
-d1_mini.build.board=WEMOS_D1_MINI
+d1_mini.build.board=ESP8266_WEMOS_D1MINI
 d1_mini.build.core=esp8266
 d1_mini.build.variant=d1_mini
 d1_mini.build.flash_mode=dio
@@ -1009,7 +1009,7 @@ d1.serial.disableRTS=true
 
 d1.build.mcu=esp8266
 d1.build.f_cpu=80000000L
-d1.build.board=WEMOS_D1
+d1.build.board=ESP8266_WEMOS_D1MINI
 d1.build.core=esp8266
 d1.build.variant=d1
 d1.build.flash_mode=dio
--- a/boards.txt
+++ b/boards.txt
@@ -944,7 +944,7 @@ d1_mini.serial.disableRTS=true
 
 d1_mini.build.mcu=esp8266
 d1_mini.build.f_cpu=80000000L
-d1_mini.build.board=ESP8266_NODEMCU
+d1_mini.build.board=WEMOS_D1_MINI
 d1_mini.build.core=esp8266
 d1_mini.build.variant=d1_mini
 d1_mini.build.flash_mode=dio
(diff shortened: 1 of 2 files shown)


igrr:
`build.board` variable shouldn't be unique for each board. It indicates the pinout which the given board uses. Some libraries need to change their behaviour depending on board pinout, so they can use `#ifdef ESP8266_NODEMCU`, `#ifdef ESP8266_ESP01`, etc to figure out which board the sketch will run on. Making this define different for similar boards doesn't help library writers.

On the second change, what is the use case for having `ARDUINO_BOARD` as a string literal?

hallard:
@igrr 
I thought boards pinout was defined in `variant/board_name/pins_arduino.h`, and the 2 `{build.board}` I changed are WeMos D1 and WeMos D1 Mini (which are not NODEMCU boards) so pinout is correct and has not changed in pin definition of the following files:
[NodeMCU](https://github.com/esp8266/Arduino/blob/master/variants/nodemcu/pins_arduino.h#L25)
[D1_Mini](https://github.com/esp8266/Arduino/blob/master/variants/d1_mini/pins_arduino.h#L25)
[D1](https://github.com/esp8266/Arduino/blob/master/variants/d1/pins_arduino.h#L25)

As the boards are declared different on `boards.txt` file (NODEMCU, WeMos D1 and WeMos D1 Mini) I thought it was making sense to be consistent on "Name" `build.board` view.

For the 2nd point, my goal was to 
- have correct board name when declared in 
[MDNS](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266mDNS/ESP8266mDNS.h#L51) instead of "generic"
- have real board name in Arduino IDE network port menu when using OTA instead of Generic ESP8266 board
- Ability to print board name in my sketch

I don't think this change will have any side effect.

igrr:
Let's at least keep the naming convention for build.board variable then — `ESP8266_D1MINI` or `ESP8266_WEMOS_D1MINI`.

hallard:
Makes sensen thanks, 
It's done

jcRBern:
i use WEMOS D1 Mini and i compile it on  Visual  studio 2013 with  Visual micro plugin.
It's the same problem , i can not compile without error on  macro ARDUINO_BOARD.
if i change platform.txt as described on #2211, compilation is OK on Visual studio  , but not OK if i compile directly with ARDUINO IDE 1.6.9.
****I don't think this change will have any side effect**.** from  hallard  --> is not  correct!!!
For large projects I prefer to use Visual Studio, for tests Arduino is sufficient, it would be nice if the two environments work.

jjssoftware:
Yes it looks like introducing escaping double quotes into platform.txt for "{build.board}" works for the eclipse arduino plugin too but as you say it breaks compile for the arduino IDE. So escaping quotes in platform.txt doesn't look like a viable solution.

I've been working with jantje in https://github.com/jantje/arduino-eclipse-plugin/issues/493 to try and come up with a fix in the eclipse arduino plugin for this problem and at the moment it is proving to be quite a difficult problem to solve for all operating systems. It's even more problematic if the -D defined string literal symbol contains one or more spaces.

PR #2060, opened 2016-05-26
Support HTTP header without white space.

shrhdk:
ESP8266HTTPClient can't process HTTP header which has no white space after `:`.

A header, `Content-Length:1234` makes `234`.

Some implementation of HTTP server returns HTTP header which has no white space.


shrhdk:
Thanks 👍

PR #2074, opened 2016-05-31
Fix init array order

unaiur:
GCC for Xtensa LX106 generates constructor sections names in inverse order. 

The link script in tools/sdk/ld/eagle.app.v6.common.ld is also designed to create an __init_array in inverse order.

So we must traverse that array starting from the end.

### DIFF
--- a/cores/esp8266/core_esp8266_main.cpp
+++ b/cores/esp8266/core_esp8266_main.cpp
@@ -119,9 +119,9 @@ static void loop_task(os_event_t *events) {
 }
 
 static void do_global_ctors(void) {
-    void (**p)(void);
-    for(p = &__init_array_start; p != &__init_array_end; ++p)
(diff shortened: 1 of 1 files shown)


igrr:
What exactly do you mean by "inverse order"? Afaik the standard doesn't define the order of initialization for global objects. What do you use as reference for "correct order"?

unaiur:
The correct order is the order described in the init_priority attribute of
gcc:

https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html

Best regards

unaiur:
I've written an small testcase here http://pastebin.com/HpcXqXsm that shows the issue.

igrr:
Ah, right, priorities are not handled correctly this way. Thanks for the fix.

PR #2079, opened 2016-06-01
Added Serial.baudRate() to get current baud rate

hallard:


igrr:
Might be worth noting that this is not a standard HardwareSerial method, and will not work on Arduino boards.

hallard:
Yeah, that's done, I also indicated I've done same PR on official Software Serial library.

PR #2084, opened 2016-06-03
Attempt at fixing min and max

igrr:
Replace macros with template functions in a hope that not we don't break too much.


devyte:
@igrr this PR shows all checks have failed, and the relevant branch is also out of date.
The code changes look correct to me, my (complex) app builds without issues, and behavior of the min/max seems correct (as expected from std versions). In addition, it allowed removing all the infernal kludges #ifdef min #undef min #endif and so on.
Unless there is some issue that I'm not encountering, I suggest merging.

igrr:
CI build has failed because this change broke some examples, e.g. CaptivePortalAdvanced.
Need to look into this again, but I have probably left this without attention because I didn't have further ideas how to fix remaining issues at that time.

devyte:
@igrr I tried building CaptivePortalAdvanced locally over my manually modified Arduino.h, and it builds without issues.
I looked at the CI logs here, and it seems to me that the errors aren't related to min/max at all, so I went ahead and clicked on Update branch... and the CI checks kicked off again automatically, so now the logs for the failed run aren't available anymore. Sorry, I'm still very new to this whole github-CI thing.
This thing has been a thorn in my side for almost a year, and not just to me, but other users as well. I didn't know this PR existed already, so I'd like to pursue it, if that's ok. Please let me know if there is anything else you would like me to test.

devyte:
@igrr Servo had an issue and didn't build. I pushed a fix, and from the logs, it looks like CI checks pass now (still running for pio).
Is there anything else I can help with?

igrr:
Thanks @devyte, much appreciated. Should be good to merge now.

cimba007:
This change breaks existing implementations ..

https://github.com/ZinggJM/GxEPD/issues/20

PR #2104, opened 2016-06-06
make HTTP Update Server more secure

me-no-dev:
* added option for authentication
* added option to change the url for upload


libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.h
```diff
@@ -9,9 +9,12 @@ class ESP8266HTTPUpdateServer
     bool _serial_output;
     ESP8266WebServer *_server;
     static const char *_serverIndex;
+    char * _username;
+    char * _password;
+    bool _authenticated;
   public:
     ESP8266HTTPUpdateServer(bool serial_debug=false);
-    void setup(ESP8266WebServer *server=NULL);
+    void setup(ESP8266WebServer *server, const char * path="/update", const char * username=NULL, const char * password=NULL);
```
igrr: Let's use a pair of overloaded functions instead of a function with bunch of default arguments here.
I.e.
```c++
void setup(ESP8266WebServer *server); // used to keep code backward compatible, calls setup(server, "/update", NULL, NULL)
void setup(ESP8266WebServer *server, const char * path); // optional, add this if you want... not really necessary IMO. calls setup(server, path, NULL, NULL);
void setup(ESP8266WebServer *server, const char * path, const char* username, const char* password); // new overload, all arguments are required
```
Reason for such a design is that it's easier to refactor/extend/deprecate an interface based on overloaded functions compared to an interface which uses function with many default arguments. This was very obvious with ESP8266HTTPClient and ESP8266HTTPUpdate, which used default arguments heavily. It was really painful to separate functions which needed TLS from the ones which didn't...

me-no-dev:
better?

libraries/ESP8266HTTPUpdateServer/examples/SecureWebUpdater/SecureWebUpdater.ino
```diff
@@ -0,0 +1,46 @@
...
+  httpUpdater.setup(&httpServer, update_path, update_username, update_password);
+  httpServer.begin();
+
+  MDNS.addService("http", "tcp", 80);
+  Serial.printf("HTTPUpdateServer ready! Open http://%s.local%s in your browser and login with username '%s' and password '%s'\n", host, update_path, update_username, update_password);
+}
+
+void loop(void){
+  httpServer.handleClient();
+  delay(1);
```
igrr: Delay not needed here

me-no-dev: yeah... was in the other example, so I left it there thinking might be needed. I'll remove it from both

me-no-dev:
let me add one more thing. message to the browser is a bit cryptic :)

libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
```diff
@@ -7,34 +7,41 @@
...
     // handler for the /update form POST (once file upload finishes)
-    _server->on("/update", HTTP_POST, [&](){
-      _server->sendHeader("Connection", "close");
-      _server->sendHeader("Access-Control-Allow-Origin", "*");
-      _server->send(200, "text/html", (Update.hasError())?"FAIL":"<META http-equiv=\"refresh\" content=\"15;URL=/update\">OK");
+    _server->on(path, HTTP_POST, [&](){
+      if(!_authenticated)
+        return _server->requestAuthentication();
+      _server->send(200, "text/html", Update.hasError() ? _failedResponse : _successResponse);
       ESP.restart();
```
me-no-dev: @igrr you think maybe we should not reboot if Update failed?

igrr: I'll merge this as is, let's address this in a separate pull request. I think a good idea would be to add update callbacks similar to ESP8266HTTPUpdate (client). Actually maybe we need to review APIs of HTTP update client and HTTP update server and make them more uniform. Let's discuss further on gitter.

PR #2126, opened 2016-06-10
MD5Builder::addStream: fixed falsy calculated hash for len > filelength

comino:
We found an issue with the MD5Builder::addStream method. 
If len > filelength handled a false hash was calculated. Refactored for better readability.  

Tested with a filestream resulted in calculating the correct hash.


igrr:
Thanks! Would you mind adding your test case into [host tests folder](https://github.com/esp8266/Arduino/tree/master/tests/host/core) as well?

comino:
oh.. of course! :)

Issue #2133, opened 2016-06-12
Add a hook to run code from user_init


igrr:
As seen [in #2111](https://github.com/esp8266/Arduino/issues/2111#issuecomment-224251391), would be useful to disable WiFi early, and maybe for some other cases as well.
##

devyte:
@igrr I suggest pushing this one back to 2.5.0 (unless it's trivial to implement?)

igrr:
Sure. At this point i am not planning on adding more features to 2.4.0, only bug fixes are considered. I guess we can mass-reassign all feature requests from 2.4.0 to 2.5.0.

d-a-v:
Closed by #5395

Issue #2111, opened 2016-06-07
Problems with ESP12f without WiFi on 50mA power source


tuxedo0801:
#### Hardware
Hardware:			ESP-12F
Core Version:      	2.2.0

### Description

I want to drive my ESP12F module from a 50mA@3.3V source. For this application I don't need WiFi at all. So I disabled it.
If I measure the current consumption with my multimeter, I get about 14mA@3.3V. That would be okay for my power source.
But as soon as I connect the ESP to the meant power source, the 3.3V line drops to 1.8V and I can measure a current consumption of about 30..35mA. The blue LED on the ESP12F starts to flicker continuously.
My current theory is:
The ESP8266 has a too high starup current consumption my power source cannot handle. I added an 100µF capacitor, but this does not solve the issue.
I don't have a scope to measure the exact startup current. 

Question is:
* Is my code okay (wrt. disabling WiFi completely)? Of course the consumtion I can measure is just about 14mA, but maybe there is a "faster way" of disabling it to reduce startup current consumption?!
* Does anybody know about the startup power consumption without wifi or can do a measurement with a scope?!
* Are there known issues with this scenario (no wifi, low current power source)?

### Settings in IDE

Module:  			Generic ESP8266 Module
Flash Size: 		4MB
CPU Frequency:		80Mhz
Flash Mode:			qio
Flash Frequency:	40Mhz
Upload Using:		SERIAL
Reset Method:		ck

### Sketch

```cpp

#include <ESP8266WiFi.h>

void setup() {
    // turn off wifi completely   
    WiFi.disconnect(); 
    WiFi.mode(WIFI_OFF);
    WiFi.forceSleepBegin();
    delay(10);
    
    Serial.begin(19200);  

}

void loop() {
   
  Serial.println("HIGH");
  
  delay(3000);
  
  Serial.println("LOW");

  delay(3000);
  
} 
```

igrr:
Can you try adding `RF_MODE(RF_DISABLED);` line between `include` and `void setup`?
After that calling WiFi.xxx functions should not be needed.
Note that this requires a pretty recent version, like staging 2.3.0-rc1 or git version.

tuxedo0801:
Thanks.I will test it this evening. 

But I'm still interested in a scope-screenshot showing the current consumption on startup. So If someone can help ...

igrr:
Bummer, `RF_MODE(RF_DISABLED);` works only after deep sleep wakeup, not after normal power-up.
[...]

PR #2136, opened 2016-06-12
#2111 Added a hook to user_init in core_esp8266_main.c to be able to do ver…

tuxedo0801:
…y early things, like disabling wifi

see: 
- https://github.com/esp8266/Arduino/issues/2111#issuecomment-224251391
- https://github.com/esp8266/Arduino/issues/2133


tuxedo0801:
No idea why this fails in ci check... works great on my 2.2.0 installation here.

igrr:
You call `early_init` but it's not defined anywhere, right? So linker legitimately complains. It works in your installation because you actually have `early_init` function in your sketch, I suppose. Now try opening and compiling one of the examples...

tuxedo0801:
indeed, but `early_init` should be optional and not mandatory :-(

How can this be done? (sorry, I'm more an java than an c expert)

tuxedo0801:
This should be easy for an C expert I guess?!

I understand why this error happens on travis-ci, and I understand why it works in my sketch.

But I have no idea how I can the compiler that there "might" be a function called "early_init".

In my lib I use function-pointers to have this kind of feature. But I have to give the function pointer first to my lib so that it can execute the function.

d-a-v:
@tuxedo0801 (two years after)

> But I have no idea how I can tell the compiler that there "might" be a function called "early_init".

We can do that with the "weak " gcc extension in function declaration.
#5395 took care of it.

Issue #2098, opened 2016-06-05
Serial.begin(,,SERIAL_TX_ONLY) doesn't works


stephjj:
Hello,

If I use Serial.begin(,,SERIAL_TX_ONLY) to use GPIO3 as a LED output on ESP-01, I got no output on TX.
But GPIO3 is usable with digitalWrite.

In uart.c (line 174), the tx_pin is set to 255 if rx_enable is false.
If I set it to 1, every thing is working fine.

Thanks.

looxonline:
I was actually about to post about this myself. Thanks for your hard work Igrr!

igrr:
I have attempted to fix this in #2141.

PR #2141, opened 2016-06-13
Fix UART pins setting (#2098)

igrr:
`rx` and `tx` were mixed up in a few places.


looxonline:
Just made the same adjustments in my 2.3.0 RC2 uart.c core file and I can confirm that they work well.

PR #2154, opened 2016-06-15
return true sketch size using ESP.getSketchSize() 

sticilface:
return true sketch size using ESP.getSketchSize() 
https://github.com/esp8266/Arduino/issues/2153

add MD5 for current binary ESP.getSketchMD5()


sticilface:
just realised that my md5 method is pants!  it does work but remainder bit needs fiddling

themindfactory:
do you release the 512 byte buffer?

PR #2167, opened 2016-06-18
added macro for max open SPIFFS files, setting it to 1 saves about 1k heap

tijnkooijmans:
Since heap is so precious on this chip, I though this might be of use. Most applications I write don't need more then 1 file open simultaneously.


kaeferfreund:
like this one 👍

reaper7:
I can define that value inside sketch?
#define SPIFFS_MAX_OPEN_FILES 2

igrr:
@reaper7 you can, but it will not have any effect.
This needs to be defined when compiling core cpp files.
I suspect @tijnkooijmans is not using Arduino IDE to build the code, so he can add this define to compiler flags.

tijnkooijmans:
indeed, I use platformIO in which you can set build flags in platform.ini

reaper7:
@igrr & @tijnkooijmans - thx for clarification

PR #2172, opened 2016-06-19
Prevent returning a pointer in a previous stack

baruch:
This fixes a possible of data corruption and crashes due to using an overwritten pointer.


baruch:
Never mind, I noticed now that the pointer is from sntp_asctime which hopefully has it in a static buffer itself. I blame this on not enough coffee.

PR #2220, opened 2016-07-04
Fix duplicated keys UploadSpeed.230400.macosx in phoenix & espresso lite.

NAzT:


igrr:
Are you trying to make both `256000` and options `230400` available on Windows?
Are both options actually supported by USB-serial adapter driver (on Windows)?
If it is, there a point in making two similar baud rates available in the menu? I.e. if 256000 is supported and it works, what's the point in having 230400?

On a related note, for boards with an integrated USB-UART I would suggest keeping just two baud rates: 115200 (which works reliably) and highest baud rate known to work (>90% success rate) on each platform. For example, for NodeMCU 1.0 this is around 921600.

PR #2234, opened 2016-07-05
Initial SPI Slave implementation and examples

me-no-dev:


ThijsB1983:
Hi,

This library is exactly what i'm looking for. I want to use my ESP12 (mounted on a wemos D1 board) as a SPI slave.

I implemented both the master and slave examples (SPISlave_Master.ino and SPISlave_Test.ino) on two different wemos boards. The master is sending "Are you alive?" -> the slave responds with "Alive for x seconds!" 

The master reads the data from the slave, but something strange is happening, I have to shift the byte 1 position to the right. Like this:

``` arduino
void readData(uint8_t * data)
{
    digitalWrite(_ss_pin, LOW);
    SPI.transfer(0x03);
    SPI.transfer(0x00);
    for(uint8_t i=0; i<32; i++) {
        data[i] = SPI.transfer(0);
        data[i] = data[i] >> 1;       // bit shift to right 
    }
    digitalWrite(_ss_pin, HIGH);
}

Is it me, or is this a little bug in the library?

**Comment by me-no-dev** [comment]
there isn't much that can be bug :) the lib reads from the hardware directly. I had it running for a couple of days here, no miss. What speeds are you using for SPI? Did you try the SafeMaster sketch?

**Comment by me-no-dev** [comment]
BTW I tested with Arduino UNO and ESP. Will try with two ESPs later (expecting a module in the mail today)

**Comment by ThijsB1983** [comment]
I did also try it with an Arduino UNO and and an ESP. That works well.... (master = arduino, slave = ESP).
The problem occurs when I use two ESP's, also with the SafeMaster sketch.
I'm new in the world of SPI, so I don't know what kind of speed I did use. In the SPIClass::begin() the frequency is:
`setFrequency(1000000); ///< 1MHz`

Is it possible that it has to do with the CPU frequency? Maybe I did setup one EPS with 80Mhz and one with 160Mhz. But I think this is not the problem.

**Comment by me-no-dev** [comment]
yes default speed is 1MHz, so that is way lower than the 8MHz that I usually run the sketches at... 
CPU freq should not matter at all :)
Will see what the result will be today with two ESPs attached to each other

**Comment by ThijsB1983** [comment]
any news?

**Comment by cacerrillos** [comment]
I can confirm that you need to shift the bits by 1 in the example code when using two ESPs.

//Bump

**Comment by vicnevicne** [comment]
Probably linked to the clock polarity. See [this post](http://www.esp8266.com/viewtopic.php?f=32&t=10579&start=12) (and followings)...

File: libraries/SPISlave/src/hspi_slave.c [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+            _hspi_slave_tx_status_cb(arg);
+        }
+        if((status & SPISWSIS) != 0 && (_hspi_slave_rx_status_cb)) {
+            uint32_t s = SPI1WS;
+            _hspi_slave_rx_status_cb(arg, s);
+        }
+        if((status & SPISWBIS) != 0 && (_hspi_slave_rx_data_cb)) {
+            uint8_t i;
+            uint32_t data;
+            uint8_t buffer[33];
```
sheinz: It seems like this buffer is not used at all.

beegee-tokyo:
Same here between an ESP32 (SPI master) and an ESP8266 (SPI slave). Need to shift the received data by 1 bit to receive the correct data on the ESP32. The ESP8266 receives data correct.
SPI is set to MSBFIRST and SPI_MODE1.
Looking on my osci shots, it looks like the ESP8266 Slave expects the command to be only 7 bits instead of 8 bits, because he starts to push the MISO line already at the 8th bit of the command!
Top is CLK, bottom is MISO

vdeconinck:
Could you switch to another SPI mode on the ESP8266 (only) ? Others have observed weird (reversed) behaviours. See https://github.com/esp8266/Arduino/issues/2416

beegee-tokyo:
@vdeconinck there is no possibility to change the SPI mode in the ESP8266 slave. Only public functions are    
in SPISlave.h: 
```cpp
public:
    SPISlaveClass()
        : _data_cb(NULL)
        , _status_cb(NULL)
        , _data_sent_cb(NULL)
        , _status_sent_cb(NULL)
    {}
    ~SPISlaveClass() {}
    void begin();
    void setData(uint8_t * data, size_t len);
    void setData(const char * data)
    {
        setData((uint8_t *)data, strlen(data));
    }
    void setStatus(uint32_t status);
    void onData(SpiSlaveDataHandler cb);
    void onDataSent(SpiSlaveSentHandler cb);
    void onStatus(SpiSlaveStatusHandler cb);
    void onStatusSent(SpiSlaveSentHandler cb);
};
```
and in hspi_slave.h:
```cpp
//Start SPI SLave
void hspi_slave_begin(uint8_t status_len, void * arg);

//set the status register so the master can read it
void hspi_slave_setStatus(uint32_t status);

//set the data registers (max 32 bytes at a time)
void hspi_slave_setData(uint8_t *data, uint8_t len);

//set the callbacks
void hspi_slave_onData(void (*rxd_cb)(void *, uint8_t *, uint8_t));
void hspi_slave_onDataSent(void (*txd_cb)(void *));
void hspi_slave_onStatus(void (*rxs_cb)(void *, uint32_t));
void hspi_slave_onStatusSent(void (*txs_cb)(void *));
```
The rest is hidden in the non-public code of Espressif.

vdeconinck:
Mmmh, I once used SPI slave code by @me-no-dev that included the possibility to select the mode, but it didn't seem to make it to the "official" SPI slave lib.
Could I suggest that you try modes 0, 2 and 3 on the ESP32 (master) ? Just to know if I'm on the right track ?

beegee-tokyo:
I tried all four modes, only mode1 works, but with this strange bit shift.
I might test tomorrow between 2 ESP8266 to see the difference.

beegee-tokyo:
Ok did a ESP8266 SPI Master to ESP8266 SPI Slave test. Getting the same result. And same picture on the oscilloscope:    

esp8266_slave_problem_2

Wondering how this ever was tagged as working. I have no level shifters laying around so I cannot test against an Arduino Uno as SPI Master, but I doubt that there is any difference.

Maybe between writing the SPI Slave library and today was some changes in the SDK, but I am not sure how to check this.

vdeconinck:
Code I used was from http://www.esp8266.com/viewtopic.php?f=32&t=10579&hilit=Spi+slave#p50011 but it predates "official" spi slave lib in the core. I think that thread is worth a read.
Note: using a Uno as a reference is a good idea. just use 1K / 1K5 resistor dividers on clk and mosi for level shifting.

beegee-tokyo:
In the [ESP8266 Technical Reference]((https://espressif.com/sites/default/files/documentation/esp8266-technical_reference_en.pdf) I found the following:
> 4.3.4    Slave SPI Communication Format Supported by Current API
>            The API function of ESP8266 SPI has a slave initialization mode which is compatible with
>            most of the devices in bytes. Set the slave communication format of 7 bits command+8
>            bits read/write data so that other master SPI devices could read and write bytes of slave
>           SPI via the 16 bits communication (or two times 8 bits with low lever CS). Please refer to
>          Section 4.4.2 for detailed information.    

But the waveform shown in the same documents shows that the data starts after the 8th bit of command:

waveform-graph-2

I could not find anything to change the slave command length, so I am wondering what is correct. Guess I have to check now how to set the ESP32's SPI master command length to 7 bits.

PR #2259, opened 2016-07-11
Add OTA type to onStart callback

andig:
Allows to decide if unmounting SPIFFS and potentially writing config files to SPIFFS is necessary.

Mentioned this on gitter- right now none of the examples show use of `SPIFFS.end()` nor is it possible to decide if thats even needed. Idea of this PR is to tell the handler what type of spiffs update is actually doing on. 

If updating the handler is undesired I could also a `getOTAType` function or similar that would expose the `_cmd` value.

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -180,7 +180,7 @@ void ArduinoOTAClass::_onRx(){
       nonce_md5.add(String(micros()));
       nonce_md5.calculate();
       _nonce = nonce_md5.toString();
-      
+
       char auth_req[38];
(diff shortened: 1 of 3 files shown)


me-no-dev:
maybe instead of changing the current callback API, you can just add another method to ArduinoOTA like so:
```cpp
int getCommand(){ return _cmd; }
```
and then check for that in the regular callback

andig:
> maybe instead of changing the current callback API, you can just add another method to ArduinoOTA like so:

Had this as an after thought, too and like it better. I'm unsure however if we need any precaution for when it's called- seems to be meaningful only after `onStart` has been called. Just add this to the docblock?

me-no-dev:
sounds good to me :)

PR #2266, opened 2016-07-11
Add support to Print::printf for printing from flash

me-no-dev:


me-no-dev:
@igrr is travis borked?

cores/esp8266/Print.cpp
```diff
@@ -63,6 +63,53 @@ size_t Print::printf(const char *format, ...) {
     return len;
 }
 
+size_t Print::printf(const __FlashStringHelper * format, ...) {
```
middelink: How about introducing Print:vprintf(PGM_P format, va_list argp) and make the other 2 use it?
```C++
size_t Print::printf(const __FlashStringHelper * format, ...) {
	va_list arg;
	va_start(arg, format);
	PGM_P fmt = reinterpret_cast<PGM_P>(format);
	size_t len = vprintf(fmt, arg);
	va_end(arg);
	return len;
```

me-no-dev: removed that printf all together. Check the latest commit

middelink: Good enough.

hallard:
@me-no-dev 
thanks, I was waiting for this one for so long....

Issue #2224, opened 2016-07-04
Pull upstream improvements to String class


igrr:
Arduino String class has recently got updated with support for C++11 range-based for loops (i.e. `String::begin` and `String::end` methods). Need to cherry-pick these changes here.

PR #2267, opened 2016-07-12
Updated String library to use C++11 iterators.

middelink:
This will allow using the String library in a ranged for loop:

```C++
String s("Hi, this is a test");

for (const char& ch : s) {
  Serial.print(ch);
}
```

and even modify

```C++
String s("Hi, this is another test");

for (char& ch : s) {
  ch++;
}
Serial.println(s);
```


middelink:
It's a "fix" for #2224

PR #2275, opened 2016-07-13
Implement Async NBNS (NetBIOS) name resolution for Windows

me-no-dev:
Source: http://www.xpablo.cz/?p=751#more-751


igrr:
What's the license for this code? Can we add appropriate copyrights?

me-no-dev:
There were none. I have not changed anything but to make it async, so all that was there is still there. Code came in a zip. Url is in the properties file

igrr:
That's what I mean, let's get in touch with the author so that he can help adding this.

me-no-dev:
Sure thing. You or me?

PR #2287, opened 2016-07-16
Adding onComplete event handler

anisimovsergey:
Adding onComplete event handler support for the asynchronous network scanning.


anisimovsergey:
Thanks for merging.

PR #2299, opened 2016-07-19
Fix analogWrite

me-no-dev:
Move to NMI interrupt
Rework the whole ISR to take less and more constant cycles
Current defaults run 10bits/1KHz@160MHz and 9bit/1KHz@80MHz (soft
mapped)


Diaoul:
Can you make a release with this? At least a beta one?

PR #2310, opened 2016-07-21
Added PUT Request to HTTP Client

fdesoye:
Added HTTP Put in order to use some HTTP APIs.

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -339,6 +339,20 @@ int HTTPClient::POST(String payload)
 }
 
 /**
+ * sends a put request to the server
+ * @param payload uint8_t *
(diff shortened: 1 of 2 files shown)


villaocho:
@fdesoye Could you please add an example of the PUT command.
I'm having issues replicating this command :
`curl --header "Content-Type: text/plain" --request PUT --data "10" http://192.168.1.200:8080/rest/items/HA_Temperature_CuartoPpal/state`

With :

`HTTPClient http;
http.begin("http://192.168.1.200:8080/rest/items/HA_Temperature_CuartoPpal/state");
http.addHeader("Content-Type", "text/plain");
int httpCode = http.PUT("10");
http.end();`

Brings error 405.
Any thoughts?

fdesoye:
@villaocho, sorry for not answering earlier. The put method had been pretty much copied from the post. Do you have a way to see on server side what has been sent? Like server logs or a tcpdump? That's how I would debug it.

Issue #2304, opened 2016-07-20
SDK 2.0 has been released


marvelousNinja:
Hi there! The new SDK version has been published on Espressif forum.

http://bbs.espressif.com/viewtopic.php?f=46&t=2451

**ESP8266_NONOS_SDK_V2.0.0_16_07_19 Release Notes**:
1. Updated libphy.a to 1055, fixed an issue of large current in Light-sleep.
2. Updated AT+ to 1.3.0:
   1. Added Light-sleep wakeup command, AT+WAKEUPGPIO;
   2. Fixed abnormal AT+CWDHCPS IP allocation;
   3. Added at_sdio demo under example directory.
3. Fixed probable system default when calling cur and def interfaces.
4. Fixed the issue of high current in Deep-sleep mode without disabling SPI first.
5. Fixed an issue where the SDK would crash when switching from AP to Station mode by calling wifi_set_phy_mode.
6. Updated secondary boot loader to v1.6:
   1. Supports AT + based on SDIO;
   2. Supports entering testing mode through GPIO.
7. Added support for MXIC Flash QIO mode.
8. Fixed exception caused during TCP handshake and retransmission.
9. Fixed issues in ESP-NOW.
10. Added ESP-PAIR (Simple-Pair) feature, for APIs please refer to ESP8266 SDK API Guide.
    wpa2-enterprise time function derivation and time check disable can be set by users.
11. Support for PEAP/TTLS in wpa2-enterprise, for APIs please refer to ESP8266 SDK API Guide.
12. Added mqtt demos under examples directory.
13. Other issue fixes.
14. Resolved an issue that RF_Init_data sector may be broken in stress test. Provided a function user_rf_cal_sector_set which has to be added in application by software developer. More details about user_rf_cal_sector_set refer to documentation "2C-ESP8266__SDK__API Guide". (Resolved in ESP8266_NONOS_SDK_V1.5.4.1)
15. Fix a potential risk that will cause rf not work after external reset. (Resolved in ESP8266_NONOS_SDK_V1.5.4.1)
16. Add SDIO AT support. (Resolved in ESP8266_NONOS_SDK_V1.5.4.1)
    Fix a potential bug in espconn. (Resolved in ESP8266_NONOS_SDK_V1.5.4.1)

**ESP8266_NONOS_SDK Notes:**
1. esp_init_data.bin has to be downloaded into flash at least once.
[...]

Issue #2263, opened 2016-07-11
Compilation error with Ardino/master, lack of newlibc maybe?


middelink:
#### Hardware
Hardware:		ESP-12E
Core Version:      	2.3.0

### Description
Compilation failure after removal of various function due to use of newlibc. Sadly, newlibc is not included so when fetching HEAD, compilation fails miserable under platformio. And likely other projects too

Problem description
I would like to be able to compile projects with esp8266/Arduino :)

### Settings in IDE
platformio actually.

Module:  			NodeMCU ESP12E
Flash Size: 		4MiB
CPU Frequency:	80Mhz
Flash Mode:		qio
Flash Frequency:	40Mhz
Upload Using:		SERIAL
Reset Method:		nodemcu

### Sketch

*Anything which includes some serious use of ESP8266Webserver, mDNS, PubSubClient and/or NTP and likely many other libraries.*

### Debug Messages
*None, as it does not run*

### Messages
```bash
xtensa-lx106-elf-g++ -o .pioenvs/espressif_esp12_board/firmware.elf -Os -nostdlib -Wl,--no-check-sections -u call_user_start -Wl,-static -Wl,--gc-sections -Wl,-wrap,system_restart_local -Wl,-wrap,register_chipv6_phy -Wl,-T"esp8266.flash.4m.ld" .pioenvs/espressif_esp12_board/src/tmp_ino_to.o -L/home/middelin/.platformio/packages/framework-arduinoespressif/tools/sdk/lib -L/home/middelin/.platformio/packages/ldscripts -L.pioenvs/espressif_esp12_board -Wl,--start-group .pioenvs/espressif_esp12_board/libFrameworkArduinoVariant.a .pioenvs/espressif_esp12_board/libFrameworkArduino.a -lmesh -lwpa2 -lsmartconfig -lpp -lmain -lwpa -llwip -lnet80211 -lwps -lcrypto -lphy -lhal -laxtls -lgcc -lm -lstdc++ .pioenvs/espressif_esp12_board/libESP8266WiFi.a .pioenvs/espressif_esp12_board/libESP8266WebServer.a .pioenvs/espressif_esp12_board/libESP8266mDNS.a .pioenvs/espressif_esp12_board/libSPI.a .pioenvs/espressif_esp12_board/libPubSubClient_ID89.a .pioenvs/espressif_esp12_board/libMD_MAX72XX_ID430.a -Wl,--end-group
[...]

PR #2317, opened 2016-07-23
[WIP] Update SDK to 2.0.0 (#2304)

marvelousNinja:
For those who will struggle to run this with PlatformIO (I have version 2.11.1), here is a guide:
1) Copy and replace files from this repo to `~/.platformio/packages/framework-arduinoespressif/`.
2) Add some missing build flags to your `platformio.ini` (bluntly stolen from #2263):

```
build_flags = -I$PLATFORMFW_DIR/tools/sdk/libc/xtensa-lx106-elf/include -L$PLATFORMFW_DIR/tools/sdk/libc/xtensa-lx106-elf/lib -lc
```

3) Make some manual modifications to linker scripts. Simple copy & replace won't do since target file has some PlatformIO specific changes. You need to modify this file:
`~/.platformio/.platformio/packages/ldscripts/esp8266.flash.common.ld`.

It should comply with changes from files in this repo:
- `Arduino/tools/sdk/ld/eagle.app.v6.common.ld`
- `Arduino/tools/sdk/ld/eagle.rom.addr.v6.ld`

In my case, I had to apply the following changes:

```
...

*.pioenvs/*/src/*.o(EXCLUDE_FILE (umm_malloc.o) .literal*, \
EXCLUDE_FILE (umm_malloc.o) .text*)
*libc.a:(.literal .text .literal.* .text.*)          /* <======= add this line */
*libm.a:(.literal .text .literal.* .text.*)
*libgcc.a:_umoddi3.o(.literal .text)
*libgcc.a:_udivdi3.o(.literal .text)
*libsmartconfig.a:(.literal .text .literal.* .text.*)
*libmbedtls.a:(.literal .text .literal.* .text.*)    /* <======= add this line */
*(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom.text .irom.text.*)
...

PROVIDE ( SPI_write_enable = 0x4000443c );
PROVIDE ( Wait_SPI_Idle = 0x4000448c );
PROVIDE ( Enable_QMode = 0x400044c0 );               /* <======= add this line */
PROVIDE ( SPIEraseArea = 0x40004b44 );
PROVIDE ( SPIEraseBlock = 0x400049b4 );
...

```


marvelousNinja:
@igrr, what are the next steps? Here's something from the top of my head:
1) Update 'Upgrading SDK' doc with a couple of details. It might speed up SDK upgrades in the future.
2) Update SDK version file, release notes & stuff.
3) There are some new static library files in the diff (and some of them were in the previous SDK releases too). They probably should be deleted: `libdriver.a`, `libgcc.a` and `liblwip_536.a` (this one is new).
4) Flash layout issue you've mentioned. To be honest, I'm not sure where to begin. I'm not a C++ expert by any means, but I find this **segfaultfest** a bit refreshing.

igrr:
Flash layout needs to be fixed:
- [ ] reduce irom0_0_seg length by 4096 bytes for all flash layouts, except 4m1m (see `tools/sdk/ld/eagle.flash.*.ld`). For 4m1m layout, there is a 2 megabyte hole between the end of irom0_0_seg and SPIFFS, so lowering irom0_0_seg size is not needed.
- [ ] add `PROVIDE( _RFCAL_start = 0x40xxxxxx );` with the address of this new 4k sector. This sector will be at the end of irom0_0_seg, 4096 bytes lower than `_SPIFFS_start`). Adding something like `PROVIDE( _RFCAL_start = _SPIFFS_start - 0x1000 );` is also a good option to reduce the number of hardcoded constants.
- [ ] add `PROVIDE( _OTA_end = _RFCAL_start );` to make the end of code segment more explicit.
- [ ] `cores/esp8266/Updater.cpp` uses `_SPIFFS_start` as the end address of code segment. Replace that with `_RFCAL_start`. Same probably happens in `cores/esp8266/Esp.cpp`. Also grep other source files in cores/ and libraries/ folder and check if `_SPIFFS_start` is used outside of the actual SPIFFS implementation and fix if necessary.
- [ ] use `&_RFCAL_start` in `user_rf_cal_sector_set` (calculate sector number by subtracting 0x40200000 and dividing by 0x1000).
- [ ] test if OTA and SPIFFS functions work with each flash layout after these changes (probably the hardest part).

Other things to do:
- [ ] update LwIP source code (tools/sdk/lwip) and do `make install` in `tools/sdk/lwip/src`.
- [ ] revert changes to c_types.h (it should work okay without redefining stuff from stdint.h)

igrr:
Opened #2440 which addresses points outlined above.

PR #2319, opened 2016-07-23
allow F() to be used for uri parameter

everslick:
currently we can only pass const char * ("foo") as uri parameter to the on() method of the webserver. in more complex sketches with many uri handlers those strings use significant amount of RAM (in my case 250+ bytes). changing the uri parameter from const char * to const String &, allows strings to be moved to flash by wrapping them in the F() macro. example:

webserver->on(F("/files"), HTTP_GET,  handle_file_page_cb);

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -128,15 +128,15 @@ void ESP8266WebServer::requestAuthentication(){
   send(401);
 }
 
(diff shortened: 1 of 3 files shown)


andig:
Wouldn't it make sense to update serveStatic as well?

everslick:
from my point of few, yes it would. but i wanted the PR to be as small as
possible to increase the chance of being merged soon.

in the long run the whole libraries would need a big overhaul to unify the
API, but i guess this will never happen as it would break backwards
compatibility.

but at least one could review all APIs and replace 'const char *' by 'const
String &' everywhere. i think is is safe regarding code, that is out there,
still compiling.

PR #2326, opened 2016-07-25
set max_connection from code (when less than 4 is needed)

Stefano-Orsolini:
not very common but some might require to set the maximum number of clients, from code, to less than 4.
in my case i must allow only one client at a time.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
@@ -67,7 +67,7 @@ static bool softap_config_equal(const softap_config& lhs, const softap_config& r
         return false;
     }
     if(lhs.max_connection != rhs.max_connection) {
-    return false;
+        return false;
     }
     return true;
 }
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -66,6 +66,9 @@ static bool softap_config_equal(const softap_config& lhs, const softap_config& r
     if(lhs.ssid_hidden != rhs.ssid_hidden) {
         return false;
     }
+    if(lhs.max_connection != rhs.max_connection) {
+    return false;
```
igrr: Please check indentation of this line

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -81,7 +84,7 @@ static bool softap_config_equal(const softap_config& lhs, const softap_config& r
  * @param channel       WiFi channel number, 1 - 13.
  * @param ssid_hidden   Network cloaking (0 = broadcast SSID, 1 = hide SSID)
  */
```
igrr: function comment block needs to be updated to include new argument

Stefano-Orsolini: included

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
Stefano-Orsolini: corrected

PR #2344, opened 2016-07-29
allow global object instances be switch off with defines

everslick:
creating global objects can be switched off by the following defines:

globally:

NO_GLOBAL_INSTANCES

locally:

NO_GLOBAL_SPIFFS
NO_GLOBAL_NETBIOS
NO_GLOBAL_SD
NO_GLOBAL_SPI
NO_GLOBAL_TWOWIRE
NO_GLOBAL_EEPROM
NO_GLOBAL_ETHERNET
NO_GLOBAL_ARDUINOOTA
NO_GLOBAL_MDNS
NO_GLOBAL_TFT
NO_GLOBAL_SPISLAVE
NO_GLOBAL_HTTPUPDATE
NO_GLOBAL_SSDP

i tried to find the most logical name for the define derived from the class name. creation of the global objects have been unified on the go.

EEPROM class got a new constructor, to allow 'external' instantiation of the class the same way es the global one (without knowing the location in flash).

### DIFF
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -139,6 +139,8 @@ using fs::SeekEnd;
 using fs::FSInfo;
 #endif //FS_NO_GLOBALS
 
+#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SPIFFS)
 extern fs::FS SPIFFS;
+#endif
 
 #endif //FS_H
--- a/cores/esp8266/spiffs_api.cpp
+++ b/cores/esp8266/spiffs_api.cpp
@@ -124,11 +124,13 @@ extern "C" uint32_t _SPIFFS_block;
 #define SPIFFS_MAX_OPEN_FILES 5
 #endif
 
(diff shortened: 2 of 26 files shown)


everslick:
I want to add, that the only global object, that is not affected by this PR is WiFi. The object is referenced from different places from within the ESP8266Wifi library and simply removing it, would not work. We'd need to refactor the WiFi class and pass around the instance pointer or reference to the parts that need it. Or we just leave it that way for the time being under the assumption that ESP needs WiFi, end of story.

hallard:
This one seems very interesting ;-)

I'm not sure it's the best way, but did you think about adding this one (or other name) NO_GLOBAL_WIFI_ENABLE that could early disable WiFi hardware so save power at startup.
As seen in #2111 this one could add @igrr modifications on `user_init()` in [core_esp8266_main.cpp](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_main.cpp#L156-L173) to disable WiFi hardware

```arduino
// previous existing code
#if defined(NO_GLOBAL_WIFI_ENABLE )
  WiFi.persistent(false);
  WiFi.mode(WIFI_OFF);
  WiFi.forceSleepBegin();
#endif
```

Issue #2233, opened 2016-07-05
lwip\mdns.h and lwip\dns.h mutual exclusive


pgScorpio:
#### Hardware

Hardware:           ESP-12E
Core Version:       2.3.0
### Description

lwip\include\lwip\mdns.h uses **LWIP_DNS_H** as #include guard.

This is the same guard as dns.h uses so you cannot include both headers from the same file.

me-no-dev:
fixed :)

PR #2396, opened 2016-08-15
fix mdns.h guard defines

me-no-dev:


me-no-dev:
should fix https://github.com/esp8266/Arduino/issues/2233

PR #2398, opened 2016-08-15
 change some 'char *' to 'const char *' to fix bad recursion in SD methods

martinayotte:
as explained in gitter : 
change some 'char *' to 'const char *' to fix bad recursion in SD methods


igrr:
cf. https://github.com/arduino/Arduino/commit/e3fae38c4167aff5237a6b5ec0820bedd542ddd7

martinayotte:
Thanks !

PR #2413, opened 2016-08-19
Added PATCH

felloe:
Added PATCH requests


igrr:
Seems to be available in latest git. Closing, thanks for the PR anyway!

PR #2417, opened 2016-08-21
ESP8266WebServer port for ESP8266 Ethernet library

burner-:


mkeyno:
@burner-  where can I find merged library?

burner-:
@mkeyno What you mean? Check that commit and you see. It is not generic wifi/ethernet library. It is new library based to ESP8266WebServer library. If you mean that "where I get generic merged wifi/ethernet library" then answer is: it will be great if we all will find it from your next pull request ;)

libraries/ESP8266EthernetWebServer/library.properties
```diff
@@ -0,0 +1,9 @@
+name=ESP8266EthernetWebServer
+version=1.0
+author=Ivan Grokhotkov
+maintainer=Ivan Grokhtkov <ivan@esp8266.com>
```
igrr: @burner- What's your plan for maintaining this library? Do you plan to update it with all the changes/fixes which the original ESP8266WebServer library receives?

If not, may I suggest that we take a look at the original library and add a bit of generics so that it can work both with ESP8266WiFi and Ethernet libs? I feel this may be a more maintainable approach for the longer term.
In any case, thank you for the contribution and would love to hear what you think of this.

burner-: I like your idea. Reason why I just fork that library instead doing generics was that I am not so good with c++. Because I had to get ethernet support I just fork it. If you can make support to that original library it would be even better. What I hope is that you also take this pin documentation etc from sample sketch and add it to wiki documentation. I use total 7 full days in last 6 months to hit my head to wall because of undocumented Ethernet library. I finally get it work and I hope that no one else need to feel my pain and do so much try & errors if A) pins are documented B) know that it can take almost minute that tcp server wake up even it reply already to ping.

mkeyno:
@burner- I thought it is merged in [ESP libraries](https://github.com/esp8266/Arduino/tree/master/libraries) any way it is nice work, I remember how painful it was when trying parse url and header  from incoming request  by Arduino Ethernet shield, and it is so relief when using Ivan webserver function without care anythings behind them , I'm gonna test your lib in my ESP-Ethernet bidirectional module , tnx

burner-:
@mkeyno I must say that I am not so good with c++/c that I can do generics. But if Igrr can do it would be great. I am also interest to help if there is need for tests. ota update via ethernet would be also really cool feature.

devyte:
Ethernet is being rewritten with lwip integration. There is already a prototype that works. As a result, the Ethernet server/client classes will essentially be aliases of the wifi client/server classes, and so the proposed webserver port won't be needed.
Closing.

Manchicor:
> Ethernet is being rewritten with lwip integration. There is already a prototype that works. As a result, the Ethernet server/client classes will essentially be aliases of the wifi client/server classes, and so the proposed webserver port won't be needed.
> Closing.

Is there a working ethernet library/lwip version that works with the same ESP8266WebServer yet? I´ve built a multi-protocol pixel node that works perfectly with either wifi or ethernet connections, but the webadmin page that allows for configuration, scene recording or playing can only be accessed via wifi connection so far. Thank you!

jandrassy:
the port exists now https://github.com/khoih-prog/EthernetWebServer

Manchicor:
> the port exists now https://github.com/khoih-prog/EthernetWebServer

Gonna try it today! Thank you!

Manchicor:
> the port exists now https://github.com/khoih-prog/EthernetWebServer

I tried it and I have two different questions about it: 1) Does it only work with the Ethernet library bundled with the ESP8266 package? I couldn´t get it to work. I´m using the Ethernet 2.0.0 library from Paul Stoffregen, found here https://platformio.org/lib/show/872/Ethernet, specially because of the "Ethernet.hardwareStatus()" feature. 2) Since functions keep the same name as those from ESP8266WebServer you can use only one or the other, there´s no way to have the ESP work with both wifi and ethernet, choosing one or the other from within the code, is it?

jandrassy:
> > the port exists now https://github.com/khoih-prog/EthernetWebServer
> 
> I tried it and I have two different questions about it: 1) Does it only work with the Ethernet library bundled with the ESP8266 package? I couldn´t get it to work. I´m using the Ethernet 2.0.0 library from Paul Stoffregen, found here https://platformio.org/lib/show/872/Ethernet, specially because of the "Ethernet.hardwareStatus()" feature. 2) Since functions keep the same name as those from ESP8266WebServer you can use only one or the other, there´s no way to have the ESP work with both wifi and ethernet, choosing one or the other from within the code, is it?

ask there, in the EthernetWebServer repository

d-a-v:
@Manchicor You can also have a look to #6680 .

PR #2424, opened 2016-08-23
Add missing '\r's and SSDP message formatting

opticron:
This fixes missing carriage returns that were causing odd output when
debugging was enabled. This also adds the capability to define a custom
SSDP message formatter for devices that need more control over the
actual wire format of messages. For devices emulating Hue bridges, the
order of message elements is important as well as the ability to add
non-standard message elements.


earlephilhower:

earlephilhower:
After looking at this I think it's not worth fixing with a new PR.

The \n to \r\n in debug messaged is the first big change, but nowhere else in the core do we do \r\n for debugging, so that's out as far as I'm concerned.

The other part involves adding an optional callback with 12(!) parameters before a SSDP message is formatted and no examples using this CB.

This *might* be worth a debug printf() (just dump the formatted message), but there's not been a clamoring for such a thing in the 3+ years this PR has been here.

Closing.

PR #2432, opened 2016-08-24
Update to axTLS 2.0.0

me-no-dev:


krzychb:
Every commit having a block with brief/param like below is worth a gold :smile: :+1:  

image

PR #2445, opened 2016-08-27
ArduinoOTA optimizations

me-no-dev:
- Added option to control if the ESP should be rebooted on success
- Added delay before ESP.restart() is called
- Added some comments to the header


igrr:
Looks good to me

me-no-dev:
I have the feeling that not delaying before the reboot is the root cause of many OTA errors received.

igrr:
True. Just noticed i had `delay(100);` added before `ESP.restart` in my local tree :)

PR #2449, opened 2016-08-28
KmpDinoWiFiESP

kmpelectronics:
Hello,
We are company producing esp8266 copatibile a device. Please, add our library and examples for this device. Information about our device you can see here: http://kmpelectronics.eu/en-us/products/futureproducts.aspx

Thank you, 
Plamen
KMP Electronics LTD


WereCatf:
`libraries/DiNoWiFi/src/KMPDinoWiFiESP.h:17:22: fatal error: arduino.h: No such file or directory`
Should be Arduino.h, not arduino.h -- the case matters.

igrr:
Hey guys,
I see this pull request contains two parts: new board and a new library. Let me comment on each of them separately.

New board: looks mostly good to me, two notes:
- please add a short section about this board into doc/boards.md and add board name into package/package_esp8266com_index.template.json.
- you are probably using some specific flash chip; please double-check boards.txt and leave only flash size option relevant to your chip. Currently i see both 2M and 4M options in there.

Regarding DiNoWiFi library:
- Please take another moment to think whether you actually want your library to be part of this repository. Currently our release cycle is pretty long (>= 2 months), and it will not get shorter in the future. This means that if you want to make some changes to your library, you will need to a) make a PR and wait for it to be reviewed and merged, and b) wait for the new version of Arduino core to be released. There is another route though: you can distribute the library via Arduino Library Manager. This way you can release new versions as often as you like by just tagging new version with git tag. The library is also easily installable via Arduino IDE. You are welcome to add link to this library in the list of supported libraries and also in description of your board (doc/boards.md file).
- Licensing examples under GPL is an interesting approach, is there any specific reason to do that? In Arduino world it is more common to release examples into public domain or at least to distribute them under a very permissive license (if releasing into public domain is not possible in your jurisdiction).
- Please check code style of your Arduino examples. You can find correct astyle options in arduino repository: https://github.com/arduino/Arduino/blob/master/format.every.sketch.sh

Last two notes may be useful to you regardless of your choice about the way to distribute your library.

kmpelectronics:
Thank you Ivan. Your information us helped a lot.

Issue #2225, opened 2016-07-04
ESP8266WebServer::collectHeaders can not be used from libraries


igrr:
`collectHeaders` methods expects the headers list to be supplied in a single call. This effectively prevents other libraries from using `collectHeaders`. I.e. ESP8266HTTPUpdateServer should be able to request Content-Length header independently from other headers which may be requested from the sketch.

Add `collectHeader` method to collect just one header. Replace C array (`_currentHeaders`) with `std::map<String, String>` and `insert` into the map. Make `collectHeaders` call `collectHeader` for each array item.

me-no-dev:
how about we use [this](https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/StringArray.h)

me-no-dev:
Ivan, please check this PR and tell me what you think

PR #2475, opened 2016-09-02
Allow ESP8266WebServer::collectHeader to be used from libraries

me-no-dev:
Move to std::map<String, String>
Connected to: https://github.com/esp8266/Arduino/issues/2225

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -49,8 +49,6 @@ ESP8266WebServer::ESP8266WebServer(IPAddress addr, int port)
 , _lastHandler(0)
 , _currentArgCount(0)
 , _currentArgs(0)
-, _headerKeysCount(0)
-, _currentHeaders(0)
 , _contentLength(0)
 , _chunked(false)
 {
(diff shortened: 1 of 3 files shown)


libraries/ESP8266WebServer/src/Parsing.cpp
```diff
@@ -68,9 +68,9 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
   String req = client.readStringUntil('\r');
   client.readStringUntil('\n');
   //reset header value
-  for (int i = 0; i < _headerKeysCount; ++i) {
-    _currentHeaders[i].value =String();
-   }
+  for (auto& x: _collectedHeaders) {
```
igrr: this can be replaced with `_collectedHeaders.clear();`

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -396,46 +388,70 @@ bool ESP8266WebServer::hasArg(String  name) {
...
+  if(_collectedHeaders.find(name) == _collectedHeaders.end()){
+    _collectedHeaders[name] = "";
   }
 }
 
 String ESP8266WebServer::header(int i) {
-  if (i < _headerKeysCount)
-    return _currentHeaders[i].value;
+  if(i >= 0 && (size_t)i < _collectedHeaders.size()){
+    int n = 0;
```
igrr: this is more concisely expressed as 
```c++
auto it = std::begin(_collectedHeaders);
std::advance(it, i);
return it.second;
```

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -396,46 +388,70 @@ bool ESP8266WebServer::hasArg(String  name) {
 
 
 String ESP8266WebServer::header(String name) {
-  for (int i = 0; i < _headerKeysCount; ++i) {
-    if (_currentHeaders[i].key.equalsIgnoreCase(name))
-      return _currentHeaders[i].value;
+  name.toLowerCase();
```
igrr: instead of modifying `name`, i would suggest to specialize `map` with a custom comparison operator (which would do `compareIgnoreCase`)

me-no-dev:
wow :D this is so far back I will need a minute to rewind

earlephilhower:

PR #2487, opened 2016-09-04
EEPROM Library:

ittacco:
Improved put function, compare data and only in case are different set _dirty flag, copy the data

### DIFF
--- a/libraries/EEPROM/EEPROM.h
+++ b/libraries/EEPROM/EEPROM.h
@@ -53,8 +53,11 @@ public:
     if (address < 0 || address + sizeof(T) > _size)
       return t;
 
-    memcpy(_data + address, (const uint8_t*) &t, sizeof(T));
-    _dirty = true;
+    _dirty = (memcmp(_data + address, (const uint8_t*) &t, sizeof(T)) !=0);
+
+    if (_dirty) 
+	  memcpy(_data + address, (const uint8_t*) &t, sizeof(T));
+
     return t;
   }
 
--- a/libraries/EEPROM/EEPROM.h
+++ b/libraries/EEPROM/EEPROM.h
@@ -52,11 +52,10 @@ public:
   const T &put(int address, const T &t) {
     if (address < 0 || address + sizeof(T) > _size)
       return t;
-
-    _dirty = (memcmp(_data + address, (const uint8_t*) &t, sizeof(T)) !=0);
-
-    if (_dirty) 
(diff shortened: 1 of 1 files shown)


libraries/EEPROM/EEPROM.h
```diff
@@ -53,8 +53,11 @@ class EEPROMClass {
     if (address < 0 || address + sizeof(T) > _size)
       return t;
 
-    memcpy(_data + address, (const uint8_t*) &t, sizeof(T));
-    _dirty = true;
+    _dirty = (memcmp(_data + address, (const uint8_t*) &t, sizeof(T)) !=0);
```
mfalkvidd: Is it really OK to overwrite _dirty? Consider this scenario:
1. EEPROM.write is called. _dirty is set to true
2. EEPROM.put is called with a struct that is identical to what is already on the flash. _dirty is set to false
3. EEPROM.commit is called. Since _dirty is false, commit returns true without actually writing anything.

I'm not an expert c/c++ programmer, so I might have missed something. Sorry for bothering you if that is the case.

me-no-dev: what is actually happening is that he checks the RAM copy of eeprom against what he is about to write and if different, then dirty is set to true and data is written to the RAM copy, ready for writing to flash.

mfalkvidd: Ok. So that means that the eeprom.write in step 1 won't be lost in step 3?

me-no-dev: You are correct. If in step 2 there is no change, first change will be gone

lorenzotacconi: I could modify the code in this way:
    if (memcmp(_data + address, (const uint8_t*) &t, sizeof(T)) !=0) {
         _dirty = true;
	  memcpy(_data + address, (const uint8_t*) &t, sizeof(T));
   }
It will grant that the _dirty flag is reset only at EEPROM.commit() and no changes are lost

mfalkvidd: Shorter **and** better. I like it :)

ppostuma: Nice stuff, but I was confused for a bit when I didn't see get/put in .cpp and thought maybe the only change was the one noted above. Any reason they're in .h?

ittacco: I just changed the lines above. 
put and get are function template and are part of class definition.

PR #2500, opened 2016-09-10
Do not drop query string arguments for zero-payload POST requests

kzyapkov:
When posting from JS code with URL-encoded parameters, but with Content-Lenght=0 and no Content-Type, the query string arguments are not available to the handler.


earlephilhower:

kzyapkov:
Sorry, I haven't played with this project for years, this pull request was long forgotten. Feel free to close it or let someone else adapt the fix.

earlephilhower:
Thanks @kzyapkov .  I've checked the current code base and someone moved the _parseArguments call out of the if() blocks already, rendering this change unneeded.  Closing.

PR #2528, opened 2016-09-16
Allow for POSTs that are larger than a few 100 bytes.

dirkx:
When sending POST responses of well over a K - _write() may not sent it all. Make sure we do -- but cap the individual writes - as somehow large 2-3k blobs seem to cause instability (on my 12F units).


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -388,8 +388,23 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
...
+            //
+            if (towrite > HTTP_TCP_BUFFER_SIZE)
+                towrite = HTTP_TCP_BUFFER_SIZE;
+            written  = _tcp->write(&p[0], towrite);
+            if (written < 0)
+                 return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
+            else 
+            if (written == 0)
+                  return returnError(HTTPC_ERROR_CONNECTION_LOST);
+            byteswritten += size;
```
stickbreaker: Line 406 is incorrect.  It should be:
```c++
byteswritten += written;
```
Chuck

earlephilhower:

dirkx:
Thats a shame - as this code probably goes wrong in a second way - on short writes too.

d-a-v:
I'm sorry I broke it while trying to merge by hand. @dirkx I'm going to PR on your fork to fix my mistake.

dirkx:
Ok - so basically all we need to do is check the return of the write()s and

	bytes_written = 0;
        while(bytes_writen < len) {
         	r = write(buff + bytes_written, len - bytes_written);
		if (r == 0)
			return error that connection is lost
		bytes_written += r;
	}

And as far as I know there is no r = -1 with various ERRNO sort of things that need to be retried (EAGAIN, EINTR, BSD style). As ‘r’ seems to be a sssize_t.

Dw.

earlephilhower:
Superseded by #6800 which is now merged.  Closing.

PR #2577, opened 2016-10-05
Added authorization type/scheme (default BASIC)

c-mysec:
ESP8266HttpClient uses fixed authorization scheme (BASIC). Add the possibility to use other schemes such as Bearer or Digest with setAutorizationType(String);

nevermind.....i just saw i  can use setheader (had misunderstand the "if" preventing from adding an "Authorization" header.


devyte:
Closing per added comment.

Issue #2584, opened 2016-10-07
mDNS library does not have support for querying host IP addresses


pilotniq:
### Description

The mDNS library is only coded to support lookup of services, and not host IP addresses.

It would be useful to be able to resolve the IP addresses of other hosts on the network which have mDNS names, such as other ESP modules or Raspberry Pis

devyte:
Closing in view of #5442 with a full rewrite of mdns.

PR #2585, opened 2016-10-07
Added support for querying host IP numbers with the new method

pilotniq:
Fix for issue #2584 

Introduces new method:

`IPAddress queryHost( const chat *host );`

Please review the consequences for service queries, I've made some changes to the parsePacket code.

### DIFF
--- a/libraries/ESP8266mDNS/ESP8266mDNS.cpp
+++ b/libraries/ESP8266mDNS/ESP8266mDNS.cpp
@@ -38,7 +38,7 @@ License (MIT license):
 #define LWIP_OPEN_SRC
 #endif
 
-#include "ESP8266mDNShost.h"
+#include "ESP8266mDNS.h"
 #include <functional>
 
 #include "debug.h"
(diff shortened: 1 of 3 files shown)


me-no-dev:
I actually have a reworked parser ready for wider testing that follows MDNS compression and extracts better the information from the answers. That includes names, texts and addresses from host or service searches.
Will post it soon as I expect most of the code to be portable to ESP32 also.
Maybe you will want to contribute to that if I have missed something.

pilotniq:
me-no-dev: Sounds good. My commit also has a new query function, as the old one could only query for services. Have you redone the queries as well?

me-no-dev:
here is the current status: https://gist.github.com/me-no-dev/e13b507a6e850af6cef733b1b35d9923
it's based on AsyncUDP but that is not an issue in this case

earlephilhower:
Thanks for your PR, but MDNS has been completely rewritten with this release, so this PR can't apply anymore without massive rework.

I'll leave open for two weeks and come back after Feb 10 and see if there's interest from the original submitter.  If not, will close then.

PR #2614, opened 2016-10-18
fallback onto index.html if index.htm fails

gtalusan:
Check SPIFFS to see if index.htm and if not then try index.html.


earlephilhower:
Simple and small.  Should either be merges 2.6 or closed as won't add.

I like it, could be useful as people move from old SD apps (8.3) to new ones (LFN).

kolos:
So THIS broke my *.htm.gz requests..

File uploaded as: dashboard.htm.gz
Http request: http://server/dashboard.htm
Actually served
- Before: dashboard.htm OR dashboard.htm.gz => file found
- After:  dashboard.htm OR dashboard.html OR dashboard.html.gz => file not found

Not a big issue, had to rename the files.
Just a FYI if someone else searches for this.

earlephilhower:
Can you open a bug on that, @kolos.  We should be checking if the HTML file actually exists before silently upgrading the request.

kolos:
Sure: #6792

PR #2639, opened 2016-10-26
Set SPI_HAS_TRANSACTION to 1 and implement buffer transfer

Lauszus:
When implementing support for the ESP8266 in the USB Host library: https://github.com/felis/USB_Host_Shield_2.0/pull/263 I noticed some differences compared to the standard Arduino core. This commit fixes that.

Note the data pointer is decreased by the data size in the buffer transfer, as transferBytes increments the pointer every time data is read: https://github.com/esp8266/Arduino/blob/master/libraries/SPI/SPI.cpp#L481.

Since I do not have a ESP8266 at hand I haven't tested these changes.

Also see: https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/libraries/SPI/src/SPI.h#L21, https://www.arduino.cc/en/Reference/SPITransfer, and https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/libraries/SPI/src/SPI.h#L244-L257 for reference.


xxxajk:
Just a small comment.
UHS30 uses SPI from within an interrupt.
This needs to be ISR safe to be complete (SPI_ATOMIC_VERSION 1) then it could work with UHS30.
ATOMIC access also allows > 1 SPI device to be accessed without collisions when one or more targets on the SPI bus use an ISR.

earlephilhower:

earlephilhower:

xxxajk:
Yeah, It's very simple.

xxxajk:
briefly, the patch is a one-liner.
https://github.com/Lauszus/Arduino/blob/SPI/libraries/SPI/SPI.h#L27
That's all that needs to be added.

xxxajk:
and this macro too.
https://github.com/Lauszus/Arduino/blob/SPI/libraries/SPI/SPI.h#L66

Lauszus:
This issue was partly fixed in https://github.com/esp8266/Arduino/commit/3cc12b1e08968aa6f35395d6a3bb265e197e91d5.

Please see the new PR: https://github.com/esp8266/Arduino/pull/6591

Issue #2578, opened 2016-10-05
softAP Subnet Mask unable to be set for DHCP


mtongnz:
When running in softAP mode, there is no way to set what subnet mask is sent to DHCP clients.  WiFi.softAPConfig allows you to set the subnet for the ESP8266 but this isn't passed to the DHCP server - it always assigns 255.255.255.0

It appears to me that this is set in [Arduino/tools/sdk/lwip/src/app/dhcpserver.c on lines 135-150](https://github.com/esp8266/Arduino/blob/18f66e99693020a758b8f5ac92e042cbdecb7760/tools/sdk/lwip/src/app/dhcpserver.c#L135).  The subnet mask is hard coded in.

Is it possible to link [WiFi.softAPConfig](https://github.com/esp8266/Arduino/blob/4897e0006b5b0123a2fa31f67b14a3fff65ce561/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp#L183) to set these values or provide another function to achieve this?

There seem to be [a few other issues logged](https://github.com/esp8266/Arduino/issues/1756) which incorrectly state the subnet isn't set at all.  The subnet appears to be set correctly on the ESP8266, just not given out by the DHCP server.

Thanks

devyte:
@mtongnz how does PR #3637 work for you?

mtongnz:
This issue was resolved with #2648.  Sorry I forgot to close this :)

PR #2648, opened 2016-10-29
Set DHCP subnet correctly for softAP

mtongnz:
Removes hard coded subnet of 255.255.255.0 and fetches the correct subnet from wifi_get_ip_info()


mtongnz:
Fixes issue https://github.com/esp8266/Arduino/issues/2578

PR #2651, opened 2016-10-29
Adding onBefore()

holgerlembke:
ith adding a onBefore() it is easy to block/grant access, log requests, verify data etc. for any on()-webrequest.

bool AllowWebServerAccess() {
 HTTPMethod m = webserver->method();

if (m == HTTP_POST) {
 if (webserver->client().remoteIP().toString().startsWith("192.168.")) {
 return true;
 } else {
 Serial.print("Access blocked to ");
 return false;
 }
 } else {
 return true;
 }

// paranoid.
 return false;
 }


earlephilhower:

PR #2652, opened 2016-10-30
Updated the Arduino version and simplified the README instructions

bogdanr:


jrobeson:
maybe it'd be better to just list the minimum version instead of keeping it updated every release.

Also, I wonder if the board manager setup should be used instead of cloning from the git repository, but perhaps that's best left to another PR.

bogdanr:
Board manager works just fine and GIT also works well.
I had a bad experience with OTA updates with older Arduino versions but with 1.6.12 OTA updates worked more reliably.

devyte:
This PR no longer applies.

PR #2683, opened 2016-11-12
ESPDuino boards change hardware

SmartArduino:
Update espduino.upload.resetmethod from ck to nodemcu, as we update the ESPDuino boards hardware.


WereCatf:
Doing it like this would break support for the ESPduinos that are already out there. I would either list these new ones that use Nodemcu-reset as either a, separate revision or at least add a way of letting people select the reset-method:
```

generic.menu.ResetMethod.ck=ck
generic.menu.ResetMethod.ck.upload.resetmethod=ck
generic.menu.ResetMethod.nodemcu=nodemcu
generic.menu.ResetMethod.nodemcu.upload.resetmethod=nodemcu
```

d-a-v:
@SmartArduino your ESPduino's [changes](https://github.com/SmartArduino/SZDOITWiKi/wiki/How-to-Use) are beeing incorporated into boards.txt.

But as @WereCatf had commented, why is the V1 using ```nocemcu``` as reset method, like V2 is (so there is no difference between the two) since it was using ```ck``` before ?

Is there a mistake in your ```boards.txt``` used as a reference for your board ?

I plan to revert V1 back to ```ck``` reset method for backward compatibility with previous ```boards.txt```, and put V2 ```nodemcu``` by default in menu.

PR #2694, opened 2016-11-16
HTTPUpdateServer callbacks for progress/errors

karlp:
Incomplete: I've not updated any documentation yet.  If this style looks good, I'll happily go and update the documentation (likewise for the examples)

One thing I've noticed, I'm using NeoPixelBus library as well, and when the update starts, I can't use the pixels anymore.  I was hoping to animate some leds for the progress handler, but it's as if something is killing off internal timers or something?  Any ideas?  Serial printf works just fine in the progress handlers.

This is an alternative implementation of https://github.com/esp8266/Arduino/pull/1817

Given that there was no existing style guide, I did restyle this library to match some other libraries and be at least self consistent.  Those can be dropped of course if necessary.


karlp:
if there's some docs on how coverage testiing works, I'd be happy to work on that too

earlephilhower:

karlp:
and wait another three years?  I'll consider it if I'm working on another arduino esp8266 project again, but there's none on the table right now.

earlephilhower:
Thanks, @karlp.  Closing as #6796 replaces it with the new code org/etc.

PR #2695, opened 2016-11-17
Timer1 define typo fix DIV265 to DIV256

egertl:
Timer1 define typo fix DIV265 to DIV256,
80MHz / 256 = 312.5Khz


wavexx:
This seems like a no-brainer. Are we worried about source compatibility?

igrr:
> Are we worried about source compatibility?

Yes, very much so. I think having both defines is okay though.

devyte:
Closing in favor of #4318

Issue #2678, opened 2016-11-09
HTTP POST problem


GolPesar132:
I have have a problem to send something use POST method and then read the page. my code is below:

    
        #include <Arduino.h>
        #include <ESP8266WiFi.h>
        #include <ESP8266WiFiMulti.h>
        #include <ESP8266HTTPClient.h>
        #define USE_SERIAL Serial
       
        ESP8266WiFiMulti WiFiMulti;
        HTTPClient http;
     void setup() {
      USE_SERIAL.begin(115200);
     // USE_SERIAL.setDebugOutput(true);
     USE_SERIAL.println();
     USE_SERIAL.println();
     USE_SERIAL.println();

    for(uint8_t t = 4; t > 0; t--) {
        USE_SERIAL.printf("[SETUP] WAIT %d...\n", t);
        USE_SERIAL.flush();
        delay(1000);
    }
    WiFiMulti.addAP("SSID", "PASSWORD");
     }

          void loop() {
         // wait for WiFi connection

        if((WiFiMulti.run() == WL_CONNECTED)) {
        HTTPClient http;
        USE_SERIAL.print("[HTTP] begin...\n");
        // configure traged server and url
        http.begin("http://server/test.php"); //HTTP

        http.addHeader("Content-Type", "application/x-www-form-urlencoded");
         http.POST("id=s1&api=123456789");
         http.writeToStream(&Serial);
         http.end();

        USE_SERIAL.print("[HTTP] GET...\n");
        // start connection and send HTTP header
        int httpCode = http.GET();

        // httpCode will be negative on error
        if(httpCode > 0) {
            // HTTP header has been send and Server response header has been handled
            USE_SERIAL.printf("[HTTP] GET... code: %d\n", httpCode);

            // file found at server
            if(httpCode == HTTP_CODE_OK) {
                String payload = http.getString();
                USE_SERIAL.println(payload);
            }
        } else {
            USE_SERIAL.printf("[HTTP] GET... failed, error: %s\n", http.errorToString(httpCode).c_str());
        }

        http.end();
    }

    delay(10000);}
when i run it i get the 411 error code. whats wrong here? how can i solve it?

maurotrochez:
Hi @GolPesar132 . Do you test this request in [POSTMAN?](https://www.getpostman.com/) is for discard that the service in php is good or you missing any parameters.

GolPesar132:
[...]

PR #2704, opened 2016-11-22
POST http client example

jamesmarlowe:
I thought an example using POST might be useful as per https://github.com/esp8266/Arduino/issues/2678


josevill:
Thanks for this example, i was looking for a working ino file. +1 👍

earlephilhower:
Simple example, but needs to be updated to use the new HTTPClient interface, not the deprecated one.  I think it can be in for 2.6.0.  If not, just drop it completely.

devyte:
This needs to be looked at and tested. Pushing back,

Issue #878, opened 2015-10-11
How to access raw POST request content


Diaoul:
Sorry if this is a dumb question I couldn't find any mention of this in the docs or examples.
Using ESP8266WebServer, I want to process the raw content of an incoming POST request.

Is that possible?
##

Duality4Y:
How do you meen the raw content of the POST request ?

Diaoul:
Well, the data POST, after the headers. I'm not using `application/x-www-form-urlencoded` but json.

Diaoul:
After digging in the source code of the library I found the answer, raw content is considered as an argument named "plain".

I leave the issue open because this is quite unusual and should be documented somewhere.

me-no-dev:
what do you think is the proper way to do it?

Diaoul:
RFC defines the content of a request as the "message-body" (see https://tools.ietf.org/html/rfc7230#section-3)
I think "content" is also a good name and is used in the popular python library [requests](http://docs.python-requests.org/en/latest/user/quickstart/#response-content)

me-no-dev:
I vote for "messageBody" :) or just "body"

marvinroger:
I also vote for `body`, the message part is implicit. It is also a name used by popular libraries, like the `request` module in Node.js

dave-prosee:
Another vote for getting this documented. Thanks for posting the issue and your findings; it helped me.

tinxx:
This issue can be closed. The documentation is available in the [README](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WebServer/README.rst).

> ``arg`` - get request argument value, use ``arg("plain")`` to get POST body

PR #2705, opened 2016-11-22
POST web server example

jamesmarlowe:
Should help with the confusion around the `plain` keyword https://github.com/esp8266/Arduino/issues/878


earlephilhower:
Simple example.  I think it can be in for 2.6.0.  If not, let's just drop it completely.

PR #2708, opened 2016-11-22
Disconnect only if the ESP has connected before

CloudyPadmal:
I've tried the scanning function in ESP many times with many ESPs and every time they stop functioning and crashed. So I checked line by line and found out that this _disconnect()_ upon no connection is the one causing it. So I added the control block to see if there is any connection and now they're working fine!


earlephilhower:
@devyte, does this still make sense?  Looks to me to be a race condition (there's some small chance WiFi IRQ will get connection between `if (!connected())` and the disconnect.

Also, I thought this kind of thing was fixed w/the IRQ stack fixes recently implemented.

devyte:
Maybe you're right, but then calling disconnect when there is no prior connection should be tested. if thst doesn't crash, just close this.

devyte:
I just tested, and calling WiFi.disconnect() without a prior connection doesn't crash with latest git.
That means this PR is not needed.
Closing.

PR #2713, opened 2016-11-24
fix send error header sometime and add collectAllHeaders() function

kingxf1995:
Fix send error header when url is like "http://github.com" (not end with '/').
Add collectAllHeaders() to support collecting all response headers. Old collectHeaders() can not handle two header with the same name.


kingxf1995:
If use this code
```
httpclient.begin("http://xxx.xxx.xxx");
code = httpclient.GET();
```
The httpclient will send
> GET  HTTP/1.1
> Host: xxx.xxx.xxx
> User-Agent: ESP8266HTTPClient
> Connection: close
> Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0
But it is a bad request, after fix httpclient will send
> GET / HTTP/1.1
> Host: xxx.xxx.xxx
> User-Agent: ESP8266HTTPClient
> Connection: close
> Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0

earlephilhower:

earlephilhower:

PR #2715, opened 2016-11-27
WIP: Add ESP8266MQTTClient Library

tuanpmt:
Hello,
This is Lightweight MQTTClient port from MQTT library for Contiki and used for ESP8266 NON-OS SDK as `esp_mqtt`.
I tried to simplify api and easier to use. I post here for anyone more familiar in SSL and WebSocket can add support mqtt over tcps & mqtt over ws(s)
This library has tested with qos0, 1, 2 - I designed outbox to support the QoS > 0 packet retransmission, but not its implement

**MQTT URI Scheme** - currently support MQTT over TCP, MQTTS & MQTT over Websocket
- `mqtt://[username][:password@]hostname[:port][#clientId]`
    + `mqtt` for MQTT over TCP 
    + `mqtts` for MQTT over SSL 
    + `ws` for MQTT over Websocket

Quick api & usage:
```cpp
#include <ESP8266MQTTClient.h>
#include <ESP8266WiFi.h>
MQTTClient mqtt;

void setup() {
  Serial.begin(115200);
//  WiFi.begin("ssid", "pass");

    delay(500);
    Serial.print(".");
  }

  //topic, data, data is continuing
  mqtt.onData([](String topic, String data, bool cont) {
    Serial.printf("Data received, topic: %s, data: %s\r\n", topic.c_str(), data.c_str());
  });

  mqtt.onSubscribe([](int sub_id) {
    Serial.printf("Subscribe topic id: %d ok\r\n", sub_id);
    mqtt.publish("/qos0", "qos0", 0, 0);
  });
  mqtt.onConnect([]() {
    Serial.printf("MQTT: Connected\r\n");
    Serial.printf("Subscribe id: %d\r\n", mqtt.subscribe("/qos0", 0));
//    mqtt.subscribe("/qos1", 1);
//    mqtt.subscribe("/qos2", 2);
  });

  mqtt.begin("mqtt://test.mosquitto.org:1883");
// mqtt.begin("ws://test.mosquitto.org:1883");
//  mqtt.begin("mqtt://test.mosquitto.org:1883", {.lwtTopic = "hello", .lwtMsg = "offline", .lwtQos = 0, .lwtRetain = 0});
//  mqtt.begin("mqtt://user:pass@mosquito.org:1883");
//  mqtt.begin("mqtt://user:pass@mosquito.org:1883#clientId");

}

void loop() {
  mqtt.handle();
}
```
Thank @igrr for awesome work


me-no-dev:
hey Tuan :) nice to see you here :)
By the way, in Arduino you do not need ```ICACHE_FLASH_ATTR```. It goes there by default

tuanpmt:
hehe, I have used and feel the power of the Arduino for ESP8266
@me-no-dev help me take a look on the code to make it better

Thank for your awesome work too

libraries/ESP8266MQTTClient/src/ESP8266MQTTClient.cpp
```diff
@@ -0,0 +1,522 @@
...
+    _mqtt_state.pending_msg_type = mqtt_get_type(_mqtt_state.outbound_message->data);
+    _mqtt_state.pending_msg_id = mqtt_get_id(_mqtt_state.outbound_message->data,
+                                 _mqtt_state.outbound_message->length);
+    LOG("Sending MQTT CONNECT message, type: %d, id: %04X\r\n",
+        _mqtt_state.pending_msg_type,
+        _mqtt_state.pending_msg_id);
+    write_len = _tcp->write(_mqtt_state.outbound_message->data,
+                            _mqtt_state.outbound_message->length);
+    LOG("Reading MQTT CONNECT response message\r\n");
+    while (_tcp->available() == 0);
```
me-no-dev: check if _tcp->connected() while waiting for data, because it will wait forever if the server closes the connection.

libraries/ESP8266MQTTClient/src/ESP8266MQTTClient.cpp
```diff
@@ -0,0 +1,522 @@
...
+    _publish_cb = fn;
+}
+void MQTTClient::onData(THandlerFunction_Data fn)
+{
+    _data_cb = fn;
+}
+void MQTTClient::handle(void)
+{
+    mqtt_outbox *ob;
+    while(!connected()) {
```
me-no-dev: this will block the whole sketch if connects fail. Maybe you want to return on connect fail and do not try to connect for ```MQTT_RECONNECT_TIMEOUT```, but let the rest of the sketch run.

libraries/ESP8266MQTTClient/src/ESP8266MQTTClient.h
```diff
@@ -0,0 +1,152 @@
...
+	int auto_reconnect;
+	mqtt_connect_info_t* connect_info;
+	uint8_t* in_buffer;
+	uint8_t* out_buffer;
+	int in_buffer_length;
+	int out_buffer_length;
+	uint16_t message_length;
+	uint16_t message_length_read;
+	mqtt_message_t* outbound_message;
+	mqtt_connection_t mqtt_connection;
```
me-no-dev: this could be named just connection :)

libraries/ESP8266MQTTClient/src/ESP8266MQTTClient.h
```diff
@@ -0,0 +1,152 @@
...
+protected:
+	std::unique_ptr<WiFiClient> _tcp;
+	MQTTTransportTraitsPtr _transportTraits;
+	bool connected();
+	int process_read();
+	void queue(int remove_on_sent);
+	void send_ping();
+	bool deliver_publish(uint8_t *message);
+
+	mqtt_state_t  _mqtt_state;
```
me-no-dev: and also here it could be named just _state.

libraries/ESP8266MQTTClient/src/URIParser.h
```diff
@@ -0,0 +1,28 @@
+#ifndef URIParser_h_
```
me-no-dev: I think URIParser can be it's own thing and be used by other libs as well :) it's kinda great

libraries/ESP8266MQTTClient/src/URIParser.h
```diff
@@ -0,0 +1,28 @@
...
+typedef struct{
+	char *scheme;               /* mandatory */
+	char *host;                 /* mandatory */
+	char *port;                 /* optional */
+	char *path;                 /* optional */
+	char *query;                /* optional */
+	char *fragment;             /* optional */
+	char *username;             /* optional */
+	char *password;             /* optional */
+} parsed_url;
```
me-no-dev: this should end on _t and maybe there could be a better name? what is the result? URI or URL actually (i kinda get confused with both).

libraries/ESP8266MQTTClient/src/URIParser.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /*
+     * Declaration of function prototypes
+     */
+    parsed_url *parse_uri(const char *);
+    void http_parsed_url_free(parsed_url *);
```
me-no-dev: maybe ```free_parsed_url``` sounds a bit better? no need to get "http" into the mix since it's not mentioned anywhere and the url can be of other scheme (```file://test.txt```). Putting what you are freeing after the word "free" makes it sound more english :)

libraries/ESP8266MQTTClient/src/URIParser.cpp
```diff
@@ -0,0 +1,316 @@
+/*_
```
me-no-dev: shouldn't this be C file? It can be and you do use extern "C" in the header

me-no-dev:
I really like that you are using queues. I think this can fit really nicely in an Async scenario, though there already is an Async MQTT.

tuanpmt:
Thank @me-no-dev. Tested ok before I commit

me-no-dev:
Nice! Now only one thing left to do. Move uri_parser into the core, so it could be used in any lib. I'll snatch it for ESP32 also ;)
But before you move it, I would like to hear @igrr what he thinks.

tuanpmt:
about uri_parser, i've found here: http://draft.scyphus.co.jp/lang/c/url_parser.html
don't see any license to use. But i think he publish for free usage 😄

me-no-dev:
guess it's free :)

ttytyper:
About url_parser, do be careful with assuming that it's free. It says pretty clearly in the source:
"Copyright 2010-2011 Scyphus Solutions Co. Ltd.  **All rights reserved**."
Unfortunately I'm pretty sure that prevents you from freely using/copying/adapting/redistributing that code in any way, unless you have made an agreement with the author.

me-no-dev:
cool :) let's write our own then. It's not like it's a big deal. Let the chaps keep their "copyright"

tuanpmt:
Rewrite, I lost 1/2 of contributions line code 😄 of uri_parser file.
only test with this:
```
  print_uri(parse_uri("mqtt://test.mosquitto.org"));
  print_uri(parse_uri("mqtt://test.mosquitto.org:1883"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883/path"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883?query"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883#fragment"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883/path?query"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883/path#fragment"));
  print_uri(parse_uri("mqtt://user:pass@test.mosquitto.org:1883/path?query#fragment"));
```

me-no-dev:
Good one Tuan! I talked to @igrr and we can move the parser to the core (and I'll steal it over to esp32 also) and let it be used by other libs (I have a few in mind)

tuanpmt:
@me-no-dev I've updated support websocket, fulltest for uri-parser, using fragment as clientId
Please taka look at the code. About websocket, may need to improve more

tuanpmt:
Update MQTTS
Need this callback for verifying fingerprint
```cpp
mqtt.onSecure([](WiFiClientSecure & client, String host) {
  Serial.printf("Verify: %s\r\n", host.c_str());
  return client.verify(fingerprint.c_str(), host.c_str());
});
```

korg1000:
Just to know: this library is not released yet, i'm correct?

tuanpmt:
Yes @korg1000 
I need to finish wss and do more test before

tuanpmt:
Need to re-abstract the transport layer to easily integrate with ws and wss

PR #2726, opened 2016-12-02
http redirection (basic wifi range extender)

quertenmont:
Adding a special case for http request redirection to another server
"/fwd/" address are used to trigger this mode

http://SERVERIP/fwd/REDIRECTIP --> very useful in meshnetwork architecture
 http://192.168.0.7/fwd/192.168.4.2 --> for instance, this allows to connect the chip with IP 192.168.4.2 which is connected to the softAP of node 192.168.0.7  
Both GET and POST requests are supported, which is very convenient in order to use the HTTPUpdater in multi-hop network


libraries/ESP8266WebServer/src/Parsing.cpp
```diff
@@ -111,6 +201,19 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
   }
   _currentMethod = method;
 
+  //allows http redirection with URL of the form:
+  //http://SERVERIP/fwd/REDIRECTIP --> very useful in meshnetwork architecture
+  //http://192.168.0.7/fwd/192.168.4.2 --> for instance, this allows to connect the chip with IP 192.168.4.2 which is connected to the softAP of node 192.168.0.7  
+  // (POST requests are also supported)
+  if(url.indexOf("/fwd/") == 0){
```
igrr: This certainly has to be an optional feature. You can implement it as a library which provides a class derived from a RequestHandler, instead of plugging directly into the server class.

quertenmont:
Nop, Actually it can't be implemented the way you suggest, because one need to have access to all the request before they get preprocess.  (Headers typically).

earlephilhower:
This is quite old and needs large refactoring to address @igrr's comments.  The new PR #5214 looks like it can do things like this, for those so inclined, so closing this to clean up the PR list.

PR #2742, opened 2016-12-07
Memory issue in WiFiClient class

KelrobGithub:
Hi all,
I encountered stability problems with my html server running on ESP8266 if there are multiple connections from clients running web browsers that (intentionally or not) do not close the connection while I'm calling client.stop() after each web page transmission to limit memory usage.

If I call client.stop() when the client does not close the connection from its side the memory is freed after a couple of minutes so there is no memory leak what I was thinking a first glance, but within this timeout for every new connection the heap is reduced by a few hundred bytes.

Even if I reload a web page from my mobile manually I could see that heap decreases by tens of kilobytes until the timeout kicks in. If this is done much faster on the client's side surely the server will run into an out-of-memory condition.

Therefore I added an abort function to WiFiClient class which I call now if the client does not disconnect within a given amount of time before calling client.stop() to immediately free the memory used by the connection which I don't need any more.

Please fell free to try the sample sketch provided below. The heap usage is printed on serial console, the behavior can be changed by commenting the #define in first line.
If #define is disabled and you connect with a browser of your choice and reload the provided web page in a fast way you can see how the heap is reduced over time. With #define enabled the heap usage stays the same which solved my issues.

Best regards,
KelrobGithub

#define CLOSE_CONNECTION_CHECK 500  // timeout in ms to check for gracefull disconnection of client side before aborting from server side

extern "C" {
  #include "user_interface.h"
}
#include <ESP8266WiFi.h>

WiFiServer server(80);
WiFiClient client;

void setup()
{
  Serial.begin(74880); // bootloader UART rate is 74880
  delay(1);
  Serial.println();

  WiFi.setAutoConnect(false);
  WiFi.setAutoReconnect(false);
  WiFi.persistent(false);

  WiFi.mode(WIFI_AP);
  WiFi.softAPConfig( IPAddress(192, 168, 5, 1), IPAddress(0, 0, 0, 0), IPAddress(255, 255, 255, 0) );
  WiFi.softAP("TEST-AP");
  
  server.begin();
  server.setNoDelay(true);
  Serial.print("Server running, use this URL to connect: HTTP://");
}

uint32_t maxHeap = 0;
uint32_t minHeap = 0;
uint32_t currentHeap = 0;
uint32_t currentHeapDisplayCounter = 0;
int timeoutClientCounter = 0;

void loop()
{
  delay(100);
  if (currentHeapDisplayCounter++ >= 20) // update serial console every 2 seconds
  {
    currentHeapDisplayCounter = 0;
    currentHeap = system_get_free_heap_size();
    if ((maxHeap == 0) && (minHeap == 0))
    {
      maxHeap = currentHeap;
      minHeap = currentHeap;
    }
    else
    {
      if (currentHeap > maxHeap) maxHeap = currentHeap;
      if (currentHeap < minHeap) minHeap = currentHeap;
    }
    Serial.printf("Free heap size: %d bytes (min: %d, max: %d)\r\n", currentHeap, minHeap, maxHeap);  
  }

  client = server.available();
  if (!client) return;
  
  Serial.println("Client detected");
  client.readStringUntil('\n'); // Assuming Webbrowser GET request to simplify test sketch

  // "Connection: close" & "Content-Lenght: ???" are both needed in the HTML header to close the connection in a fast way to not cause large memory usage on ESP8266.
  const String samplePageBody = "<!DOCTYPE HTML>\r\n <html> <head> <title> LED testing </title> </head> <body> TEST </body> </html>\r\n";
  // Correct HTML header requesting the client to close the connection
  // String samplePage = "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\nConnection: close\r\nContent-Length: "; 
  // Debug HTML header requesting the client to NOT close the connection so our abort implementation kicks in
  String samplePage = "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\nConnection: keep-alive\r\nContent-Length: "; 
  samplePage += String(samplePageBody.length());
  samplePage += "\r\n\r\n";
  samplePage += samplePageBody;
  
  client.print(samplePage);
  client.flush();

#ifdef CLOSE_CONNECTION_CHECK
  timeoutClientCounter = 0;
  while ((client.connected()) && (timeoutClientCounter < CLOSE_CONNECTION_CHECK))
  { // Check for disconnection from client side as required by HTML header send above
    timeoutClientCounter += 10;
    delay(10);
    client.flush();
  }
  if (client.connected())
  { // Abort connection from server side in case the client is still connected and does not close the connection on its own.
    // The following line needs to be added to WiFiClient.cpp to allow abort of client connection to free memory faster than normal preventing the system from memory issues:
    //   void WiFiClient::abort() {if (!_client) return; _client->abort();}  
    // The following line needs to be added to WiFiClient.h for the same purpose:
    //   virtual void abort();
    // Path to both files (if using Win7) is "C:\Users\<YOUR_USER_NAME>\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\2.3.0\libraries\ESP8266WiFi\src\"
    client.abort();
    Serial.println("Client is still connected, aborting connection...");
  }
#endif

  // Call client.stop only AFTER making sure that the client has already disconnected and no data is waiting,
  // else the context is destroyed only after a long timeout of ~2 minutes causing memory issues in case of multiple fast client requests...
  client.stop(); 
}

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClient.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClient.cpp
@@ -269,6 +269,13 @@ void WiFiClient::stop()
     _client = 0;
 }
 
+void WiFiClient::abort()
+{
+    if (!_client)
+        return;
+    _client->abort();
+}
+
 uint8_t WiFiClient::connected()
 {
     if (!_client)
--- a/libraries/ESP8266WiFi/src/WiFiClient.h
+++ b/libraries/ESP8266WiFi/src/WiFiClient.h
@@ -63,6 +63,7 @@ public:
     return peekBytes((uint8_t *) buffer, length);
   }
   virtual void flush();
+  virtual void abort();
   virtual void stop();
   virtual uint8_t connected();
   virtual operator bool();
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -217,14 +217,16 @@ void SPIClass::setClockDivider(uint32_t clockDiv) {
     SPI1CLK = clockDiv;
 }
 
-inline void SPIClass::setDataBits(uint16_t bits) {
+inline void SPIClass::setDataBits(uint32_t bits) {
     const uint32_t mask = ~((SPIMMOSI << SPILMOSI) | (SPIMMISO << SPILMISO));
     bits--;
     SPI1U1 = ((SPI1U1 & mask) | ((bits << SPILMOSI) | (bits << SPILMISO)));
 }
 
 uint8_t SPIClass::transfer(uint8_t data) {
+#ifndef SPI_SKIP_INITIAL_BUSY_CHECK
     while(SPI1CMD & SPIBUSY) {}
+#endif
     // reset to 8Bit mode
     setDataBits(8);
     SPI1W0 = data;
@@ -256,7 +258,9 @@ uint16_t SPIClass::transfer16(uint16_t data) {
 }
 
 void SPIClass::write(uint8_t data) {
+#ifndef SPI_SKIP_INITIAL_BUSY_CHECK
     while(SPI1CMD & SPIBUSY) {}
+#endif
     // reset to 8Bit mode
     setDataBits(8);
     SPI1W0 = data;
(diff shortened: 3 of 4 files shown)


devyte:
@d-a-v does this make sense in relation to the tcp_abort example?

d-a-v:
@KelrobGithub you seized the issue.
There is a more simple way very freshly described in the faq ([link](https://arduino-esp8266.readthedocs.io/en/latest/faq/readme.html#how-to-clear-tcp-pcbs-in-time-wait-state)).

earlephilhower:
This has been taken care of as described in the FAQ.  Closing.

PR #2745, opened 2016-12-08
Functional Interrupts

hreintke:
Imlementation of the possibility to use std::functions for Interrupt callback.

Example usage : 
```#include <Arduino.h>

class InterruptClass
{
public:
	InterruptClass(){};
	void classFunctionInterrupt(void)
	{
		Serial.printf("Class function interrupt\r\n");
	}
};

bool blinker;
void blink(int blinkPin)
{
	if (blinker)
	{
		digitalWrite(blinkPin,HIGH);
	}
	else
	{
		digitalWrite(blinkPin, LOW);

	}
	blinker = !blinker;
	Serial.printf("Blink on pin %d, value %d\r\n", blinkPin,blinker);
}

void globalInterrupt()
{
	Serial.printf("Global function interrupt\r\n");
}

InterruptClass demoClass;

void setup() {

	Serial.begin(74880);

	pinMode(D1, INPUT);
	pinMode(D2, INPUT);
	pinMode(D5, INPUT);
	pinMode(D4, OUTPUT);

	attachInterrupt(D1, std::bind(blink,D4), CHANGE);

	attachInterrupt(D2, globalInterrupt,CHANGE);

	attachInterrupt(D5, std::bind(&InterruptClass::classFunctionInterrupt,&demoClass), CHANGE );
}


void loop() {

}
```


igrr:
This looks pretty good!
I am pondering if we can somehow remove the need for the `#include <FunctionalInterrupt.h>`... Maybe this can be added to wiring.h under a `#ifdef __cplusplus` conditional? Would that break anything?

hreintke:
Hi,
I tried but with no positive result.
Have not much experience with c/c++ combinations and the low-level setup of the library.
Will need some help to get an alternative solution.

devyte:
Hello,

I find this to be far too useful to not merge asap, given that it basically allows passing a context to an ISR. 

@igrr Is it possible to merge as-is (taking into account a branch update, of course), and open a new issue to cover removing the need of the header include?

philbowles:
I LOVE you guys! This is *exactly* the code I was looking into writing for my current project...and so of course I agree with devyte that it should be in as soon as possible. I have already done something similar by extending the Ticker class, allowing functional calls on timers. My overall project attempts to serialise all external events into a central task queue to avoid WDTs, resource clashes etc. It's pretty much done and dusted and working apart from the final piece in the jigsaw...which is THIS!

You have saved me weeks of hair-pulling!  I now just need to get this into my own dev code and have a single interrupt global handler with pin Arg which pushes the appropriate function into the task queue...job done!

I'm a bit of a newbie with github and "branches" and "pulls" so I'd appreciate some advice please, on how to get the latest version into my arduino ide without breaking anything!

devyte:
Read using git version of the core in readthedocs. You have to delete (or
move out) the ide version of the core to make sure it's picked up.

I'm already using this, and it works very well, so thank you @igrr for
merging so quickly.

cores/esp8266/FunctionalInterrupt.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+// Helper functions for Functional interrupt routines
+extern "C" void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void*fp , int mode);
+
+// Structure for communication
+struct ArgStructure {
+	std::function<void(void)> reqFunction;
+};
+
+void interruptFunctional(void* arg)
```
devyte: @igrr it's a bit late, but shouldn't this have the ICACHE_RAM_ATTR tag?

cores/esp8266/FunctionalInterrupt.cpp
```diff
@@ -0,0 +1,24 @@
...
+{
+	((ArgStructure*)arg)->reqFunction();
+}
+
+void attachInterrupt(uint8_t pin, std::function<void(void)> intRoutine, int mode)
```
devyte: @igrr what about here? should this be std::function<void ICACHE_RAM_ATTR (void)> ?

cores/esp8266/FunctionalInterrupt.cpp
igrr: Yes, you're right, under the current model it should have ICACHE_RAM_ATTR.

cores/esp8266/FunctionalInterrupt.cpp
igrr: This is more tricky. You can't add attributes to function prototypes like that, and furthermore the invoke method of std::function (which gets called when you use operator() on the function object) is not placed into IRAM either (it's a part of libstdc++).

I think the only way to work around this issue is to implement interrupt masking for the duration of SPI flash operation (#3579), then IRAM_ATTRs will not be needed at all.

hostmit:
Can this be impletemented for ESP32?

hreintke:
@hostmit See https://github.com/espressif/arduino-esp32/pull/1728

nixmeer:
Hello :)

I would like a constructor to attach an interrupt to a given pin and the callback to be a member function of that object that is being constructed.

`class MyClass {
public:
  MyClass(int _pin);
private:
  void callback();
  int counter;
}

MyClass::MyClass(int _pin){
  attachInterrupt(_pin, std::bind(&MyClass::callback, this), CHANGE);
}

void MyClass::callback(){
  counter++;
}`

The error is: 
`/Users/jt/Documents/Arduino/libraries/ESP_SmartHomeDevice/ESP_SHD_MotionSensor.cpp: In constructor 'ShdMotionSensor::ShdMotionSensor(uint8_t)':
/Users/jt/Documents/Arduino/libraries/ESP_SmartHomeDevice/ESP_SHD_MotionSensor.cpp:7:75: error: cannot convert 'std::_Bind_helper<false, void (ShdMotionSensor::*)(), ShdMotionSensor* const>::type {aka std::_Bind<std::_Mem_fn<void (ShdMotionSensor::*)()>(ShdMotionSensor*)>}' to 'void (*)()' for argument '2' to 'void attachInterrupt(uint8_t, void (*)(), int)'
   attachInterrupt(pin, std::bind(&ShdMotionSensor::pinChange,this), CHANGE);`

MyClass = ShdMotionSensor
callback = pinChange

What am I doing wrong?

nixmeer:
just for the record: 
#include <FunctionalInterrupt.h> is still necessary

Issue #1923, opened 2016-04-17
memory leak 184 bytes at a time


yomasa:
### Basic Infos
looks like 184 bytes get lost every time you post to the page ie :'http://192.168.0.6/gpio/1'
#### Hardware
Hardware:		ESP-12, wemos mini, esp-7
Core Version:      	2.2.0-rc1

### memory leak?

using the example sketch WifiWebServer added one line of code at the end of sketch

 Serial.println( ESP.getFreeHeap(),DEC);

### Settings in IDE

Module:  			wemos mini
Flash Size: 		?4MB/1MB?
CPU Frequency:		?80Mhz?
Flash Mode:			?qio?
Flash Frequency:	?40Mhz?
Upload Using:		?OTA / SERIAL?
Reset Method:		?ck / nodemcu?

### Sketch

/*
 *  This sketch demonstrates how to set up a simple HTTP-like server.
 *  The server will set a GPIO pin depending on the request
 *    http://server_ip/gpio/0 will set the GPIO2 low,
 *    http://server_ip/gpio/1 will set the GPIO2 high
 *  server_ip is the IP address of the ESP8266 module, will be 
 *  printed to Serial when the module is connected.
 */

#include <ESP8266WiFi.h>

const char* ssid = "your-ssid";
const char* password = "your-password";

// Create an instance of the server
// specify the port to listen on as an argument
WiFiServer server(80);

void setup() {
  Serial.begin(115200);
  delay(10);

  // prepare GPIO2
  pinMode(2, OUTPUT);
  digitalWrite(2, 0);
  
  // Connect to WiFi network
  Serial.println();
  Serial.println();
  Serial.print("Connecting to ");
  Serial.println(ssid);
  
  WiFi.begin(ssid, password);
  
  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
  }
  Serial.println("");
  Serial.println("WiFi connected");
  
  // Start the server
  server.begin();
  Serial.println("Server started");

  // Print the IP address
  Serial.println(WiFi.localIP());
}

void loop() {
  // Check if a client has connected
  WiFiClient client = server.available();
  if (!client) {
    return;
  }
  
  // Wait until the client sends some data
  Serial.println("new client");
  while(!client.available()){
    delay(1);
  }
  
  // Read the first line of the request
  String req = client.readStringUntil('\r');
  Serial.println(req);
  client.flush();
  
  // Match the request
  int val;
[...]

Issue #1070, opened 2015-11-22
memory leak by using WiFiClient (40 Byte per usage)


Links2004:
test code:
```cpp
void loop() {
    USE_SERIAL.printf("RAM: %d\n", ESP.getFreeHeap());
    // wait for WiFi connection
    if((WiFi.status() == WL_CONNECTED)) {
        WiFiClient * _tcp = new WiFiClient();
        _tcp->connect("192.168.1.12", 80);
        _tcp->stop();
        _tcp->~WiFiClient(); //does not help but shut abort any connection and free all
    }
    delay(100);
}
```

return:
```
RAM: 41312
RAM: 41312
RAM: 41048
RAM: 40816
RAM: 40584
RAM: 40352
RAM: 40104
RAM: 39872
RAM: 39640
RAM: 39408
RAM: 39176
RAM: 38944
RAM: 38712
RAM: 38480
RAM: 38248
RAM: 38016
RAM: 37784
```

igrr:
https://github.com/esp8266/Arduino/issues/230#issuecomment-101371864

igrr:
You are doing active close. Send Connection: close header and wait for server to do active close.

Links2004:
can confirm this but the active close takes very long (60sec).
but i still miss 376Byte.

start: 41296
avter close timeout: 40920 

````cpp
void loop() {
    static int i = 10;
    static uint32_t last_ram = 0;
    uint32_t ram = ESP.getFreeHeap();
    USE_SERIAL.printf("RAM: %d (%d)\n", ram, (ram - last_ram));
    last_ram = ram;
    // wait for WiFi connection
    if((WiFiMulti.run() == WL_CONNECTED) && ram > 2000 && --i > 0) {

        WiFiClient * _tcp = new WiFiClient();
        _tcp->connect("192.168.1.12", 80);
        _tcp->stop();
        _tcp->~WiFiClient(); //does not help but shut abort any connection and free all
        delay(100);
    } else {
        delay(1000);
    }
}
```

````
[SETUP] WAIT 2...
[SETUP] WAIT 1...
RAM: 41296
RAM: 41296 (41296)
RAM: 41032 (-264)
RAM: 40800 (-232)
RAM: 40568 (-232)
RAM: 40336 (-232)
RAM: 40088 (-248)
RAM: 39856 (-232)
RAM: 39624 (-232)
RAM: 39392 (-232)
RAM: 39160 (-232)
RAM: 39160 (0)
[...]

Issue #230, opened 2015-05-12
Memory leak ESP8266WebServer


robertgregor:
Hello, I have a sketch, which is using the webserver. Seems there is memory leak, thus using timer, I am printing each 2 seconds:
 Serial.print("Free heap:");
 Serial.println(ESP.getFreeHeap(),DEC); 

I did the regular access to the Webserver port - and it was in debug mode and see the output:

 Free heap:6288
Free heap:6288
Free heap:6288
Free heap:6288
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:6056
Free heap:6056
Free heap:6056
Free heap:6056
Free heap:6056
New client
method: GET url: /cwd search: 
args: 
Request: /cwd
 Arguments: 
Free heap:5864
Free heap:5864
Free heap:5864
Free heap:5864
New client
method: GET url: /cww search: 
args: 
Request: /cww
 Arguments: 
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
Free heap:5672
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:5480
Free heap:5480
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:5288
Free heap:5288
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:5096
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:4904
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:4712
Free heap:4712
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:4520
Free heap:4520
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:4328
Free heap:4528
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:4336
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:4144
Free heap:4144
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:3952
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:3952
New client
method: GET url: / search: 
args: 
Request: /
 Arguments: 
Free heap:3760
New client
method: GET url: / search: 
args: 
[...]

PR #2767, opened 2016-12-15
Added close_abort() function to WiFiClient

pfabri:
This branch adds a close_abort() method to the WiFiClient class.

### How it works, what it does:
Calling `close_abort()` will close and abort the client connection it
is invoked on and, as a result, free its resources (i.e. memory).

**WARNING:** aborting connections without a good reason violates the
TCP protocol, because a closed connection would normally need to
spend some time in `TIME_WAIT` state before its resources are freed.

### Usage example:
    WiFiClient client;		// set up a client

    { /* do things with your client */ }

    client.stop_abort()		// when you're done,abort the
				// connection if you must

### Why it's useful:
1. Give programmers a way to shut down connections immediately if
   need be. The underlying `tcp.c` file has an abort function, but
   this has not been directly accessible via the `WiFiClient`
   class until now.

2. There are a number of reported issues for the repository
   addressing the heap corruption that can result from trying to
   retain too many connections in `TIME_WAIT` state (most notably:
   #230, #1070, #1923). Although the warning above holds, there may be
   circumstances where this isn't very important. For example an ESP8266
   running in AP mode hosting a page, which requests a new
   connection every second via an AJAX script to  monitor
   a sensor/button/etc. continusously.

### Currently existing alternative approach:
When building a project, defining the
`-D MEMP_NUM_TCP_PCB_TIME_WAIT=5`compiler directive will limit the
maximum number of clients allowed to stay in TIME_WAIT state. `5` is
the default, lower it as necessary. See reference
[here](https://github.com/esp8266/Arduino/blob/master/tools/sdk/lwip/include/lwipopts.h#L263)

### Thanks:
Thank you to @me-no-dev, @everslick and @Palatis for bringing the `
MEMP_NUM_TCP_PCB_TIME_WAIT` option to my attention.


psy0rz:
We at the ESPEasy project need this as well: Sometimes we need to make a lot of new connections in a short time, causing memory problems.

Cant seem to find a workaround that does not involve patching Wificlient itself.

kevinkk525:
Thanks a lot! This solved my issue after hours of searching for the memory leak in my server application.. lost 1.6kB of heap within 10 messages with different clients.
Why is this commit still not merged into the main branch?
I somehow doubt that this problem is so minor that it can be ignored. or is my server software doing something wrong by not closing the connection or similar?

psy0rz:
a cleaner workaround that doesnt require changes to core libraries: https://github.com/esp8266/Arduino/issues/1923#issuecomment-298431891

kevinkk525:
Thanks for your answer! Funny thing is, I also found that, but it does not make any difference at all. I put this code before any includes except Arduino.h as I get an error if I put it before Arduino.h or use my code without Arduino.h.
It just does not make any difference for me. If it did, it would be the better workaround of course.

everslick:
Did you also call tcpCleanup() in loop() ?

kevinkk525:
Thanks a lot! This does work quite well! Forgive me, I somehow forgot to call that function..
Thank you for all the help!

psy0rz:
yeah you have to call it at a point you think there is some cleaning up to do. or just in loop(). no harm in calling it too often i think.

earlephilhower:
@d-a-v, I think this is no longer needed with your TIME_WAIT fixes.  If so, can you please close?

earlephilhower:

devyte:
This has a conflict, pushing back.

d-a-v (approved):

d-a-v:
> I think this is no longer needed with your TIME_WAIT fixes. If so, can you please close?

The primary goal of this PR has indeed been fixed by @me-no-dev's patch that is ported to lwip2.
If such an API is not needed for other purpose we can close this PR.

d-a-v:
Without feedback and due to old age, also with another fix addressing the original issue that is already in the core, this PR should be closed.
It can be reopened after discussion, if any.

Issue #2491, opened 2016-09-07
Tone causes crash with frequency = 0


cranphin:
It's not uncommon to use tone with frequency = 0 to indicate 'silence/no note'.
Example: https://github.com/arduino/Arduino/blob/master/build/shared/examples/02.Digital/toneMelody/toneMelody.ino
This crashes right now, I expect because of a division in zero in: https://github.com/esp8266/Arduino/blob/master/cores/esp8266/Tone.cpp

>  timer1_write((clockCyclesPerMicrosecond() \* 500000) / frequency);
> Admittedly the behavior on Arduino for frequency = 0 doesn't seem to be documented, and I expect on some other cores it would also crash or do weird things..

Razpudding:
I second this issue. I worked around it by checking if frequency is '0' and if so setting it to '1'. If someone has an actual solution I'd love to hear it.

pfeerick:
Over on the Oak we've also just experienced this issue as well, as it is essentially based on the ESP8266 core.  In the issue [I opened over there](https://github.com/digistump/OakCore/issues/75) to document that it's a known problem, I proposed the following solution to work around the issue, but only at the user sketch level... it doesn't really fix the problem as it results from the implmentation, but it at least prevents the sketch from crashing...

Basically, Instead of

```tone(3, melody[thisNote], noteDuration);```

to use

```if (melody[thisNote] == 0)
{
  noTone(3);
  delay(noteDuration);
}
else
{
  tone(3, melody[thisNote], noteDuration);
}```

devyte:
Code is already merged, closing.

PR #2780, opened 2016-12-20
Prevent divide by zero error causing tone() to crash

pfeerick:
As per the issue at #2491, there is a divide by error issue resulting from the specification of 0 as the frequency. This does not appear to affect the AVR implementation, but it crashes on ESP8266s. I have merely removed the division if the frequency is zero, which appears to be giving the expected results (no tone), without any code crashes. 

To test, simply load the toneMelody sketch included with the Arduino IDE (Examples -> 02. Digital -> toneMelody) and change the piezo to something else if you need to. On the Witty module used to test this, I could also tell by the wifi led blinking every time the code crashed as the ESP8266 immediately rebooted.

### DIFF
--- a/cores/esp8266/Tone.cpp
+++ b/cores/esp8266/Tone.cpp
@@ -82,7 +82,12 @@ void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) {
         timer1_isr_init();
         timer1_attachInterrupt(t1IntHandler);
         timer1_enable(TIM_DIV1, TIM_EDGE, TIM_LOOP);
-        timer1_write((clockCyclesPerMicrosecond() * 500000) / frequency);
+        if (frequency == 0) {
+          timer1_write((clockCyclesPerMicrosecond() * 500000));
+        }
+        else {
+          timer1_write((clockCyclesPerMicrosecond() * 500000) / frequency);
+        }
         break;
     }
   }
--- a/cores/esp8266/Tone.cpp
+++ b/cores/esp8266/Tone.cpp
@@ -64,6 +64,13 @@ void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) {
     // Set the pinMode as OUTPUT
     pinMode(_pin, OUTPUT);
 
+	  // Alternate handling of zero freqency to avoid divide by zero errors
+	  if (frequency == 0)
+    {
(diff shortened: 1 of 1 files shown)


igrr:
Wouldn't this cause a 1Hz signal to be generated instead?
I would suggest to move the check for zero frequency to the top of the function, and do a `delay(duration)` followed by a `return` in that case.

pfeerick:
Possibly - I didn't check on the 'scope to see what the output from that pin was with the changes - just that the piezo stopped making (audible?) noise, and the code not crashing. I don't think using a delay() is a viable option though, as tone() is supposed to be a non-blocking function - it merely sets the timer and the ISR does the work of maintaining the PWM signal and disabling it if necessary. 

So where does that leave things? Setting a flag in tone if the frequency is 0 and using it on the ISR to leave the pin low until the timer runs out?

igrr:
Ah, you are right. I've missed the non-blocking nature of `tone`.

I don't know what is the AVR Arduino behavior for tone with zero frequency — does it generate positive pulse with length equal to `duration`, or it just leaves the output low? Can anyone check this? If it is the former, then indeed we need to use a timer (and check for possible timer overflows...), if it is the latter, then just bail out of `tone` without doing anything if `frequency` is zero.

pfeerick:
On Arduino (AVR anyway) tone with zero frequency runs with no audible sound, for the specified period. I'll check in a few days what the actual output is on a 'scope. Is used in the toneMelody example, as it allows for a pause without having to do anything fancy, as the melody is stored in an array, and it is just playing each element in sequence, with a second array specifying note lengths.

pfeerick:
Actually, on second thoughts... and after some more checks against a scope with an Arduino Nano + reading some more documentation on tone, it can't go below 31 Hz anyway, and simply outputs nothing if you try. So probably just a check for a 0 frequency and an redirect to noTone instead as that will disable the timer again, and set the pin low / off.

pfeerick:
It would appear the ESP8266 is better at tone() than the Arduino in some instances... the arduino implementation states it is limited to a minimum of 31-41 hertz (depending on board), whereas the ESP8266 can go down as far as at least 5hz. The latest commit appears to produce the expected results, and  a frequency of 0 does indeed result in no crashes and no output.

PR #2784, opened 2016-12-21
platform.txt: don't set compiler.warning_flags to -w

silbe:
makeEspArduino doesn't have any menu logic, so it uses
compiler.warning_flags rather than one of the compiler.warning_flags.*
properties. The GCC option "-w" isn't overridable, so there's no way
for the user to enable warnings.

Set compiler.warning_flags to the same value as
compiler.warning_flags.default, i.e. empty, to fix this.


earlephilhower:
This is simple enough for us to either merge in 2.6.0 or close as a won't merge, to declutter the PR list.

Issue #2499, opened 2016-09-10
OTA (HTTP client mode) running out of memory and crashing when using TLS (https)


silbe:
#### Hardware
Hardware:			ESP-12
Core Version:      	git master (commit 4897e000)

### Description

When modifying the ESP8266httpUpdate example "httpUpdate" to use an https URI, the ESP8266 crashes with a null pointer dereference (Exception 29 = StoreProhibitedCause,  excvaddr=0). Debugging suggests we're running out of DRAM, even though the sketch is not doing anything other than OTA. The server uses a 2048-bit RSA key with a certificate issued by letsencrypt.org.

The null pointer dereference is already tracked as #2201; this ticket is about running out of memory in the first place. The failure is actually very similar, but in #2201 the server used a 4096-bit key (which **could** be considered excessive), whereas we're using the absolute minimum currently permitted by most CAs and major browsers. If it doesn't even work with a 2048-bit key and an otherwise empty sketch, it's pretty much useless for practical purposes.

### Build system

Build system:		makeEspArduino (commit 18d85bc5)
Module:  			Generic ESP8266 Module (BOARD=generic)
Flash Size: 		4MB/1MB (FLASH_DEF=4M1M)
CPU Frequency:		unmodified
Flash Mode:			unmodified
Flash Frequency:	unmodified
Upload Using:		SERIAL
Reset Method:		manual (power switch)

### Sketch

```cpp

/**
 * httpUpdate.ino
 *
 *  Created on: 27.11.2015
 *
 */

#include <Arduino.h>

#include <ESP8266WiFi.h>
#include <ESP8266WiFiMulti.h>

#include <ESP8266HTTPClient.h>
#include <ESP8266httpUpdate.h>

#define USE_SERIAL Serial

ESP8266WiFiMulti WiFiMulti;

void setup() {

    // USE_SERIAL.begin(115200);
    USE_SERIAL.begin(57600);
    // USE_SERIAL.setDebugOutput(true);

    USE_SERIAL.println();
    USE_SERIAL.println();
    USE_SERIAL.println();

    for(uint8_t t = 4; t > 0; t--) {
        USE_SERIAL.printf("[SETUP] WAIT %d...\n", t);
        USE_SERIAL.flush();
        delay(1000);
    }

//    WiFiMulti.addAP("SSID", "PASSWORD");
    WiFiMulti.addAP("bbox", "PASSWORD");

}

void loop() {
    // wait for WiFi connection
[...]

PR #2785, opened 2016-12-21
Updater: enable final checks from outside

silbe:
As mentioned in #2499, I'm using a custom OTA implementation that fetches only a cryptographically secure checksum via HTTPS and performs the actual update over HTTP (to avoid around running out of memory during the update). To perform the checksum verification, it needs to know where in flash the update has been downloaded to (i.e. _startAddress). To prevent the update from getting applied if the checksum doesn't match but still allow future update attempts, it needs a way to abort an update operation.

These changes are useful not just for performing verification of non-MD5 checksum, but for any kind of post-download verification (e.g. embedded signatures, rollback protection).

### DIFF
--- a/cores/esp8266/Updater.h
+++ b/cores/esp8266/Updater.h
@@ -69,6 +69,8 @@ class UpdaterClass {
     */
     bool end(bool evenIfRemaining = false);
 
+    void abort() { _reset(); clearError(); }
+
     /*
       Prints the last error to an output stream
     */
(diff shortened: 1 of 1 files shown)


earlephilhower:
I think Updater.cpp has been refactored a lot since this PR came in.  Right now the PR adds an abort() function and a simple member accessor, which is reasonable, but there is no way to stop the actual updater from applying the update if you don't use it.

Since we've redone the Updater, and added automatic crypto-signed verification, and it can't actually affect things with current code, I'm closing this.

PR #2786, opened 2016-12-22
[i2s] use static allocated dma buffer

Palatis:
`i2s_slc_begin()` does `malloc()` the DMA buffers, but `i2s_slc_end()`
doesn't `free()` them.
so multiple calls to `i2s_begin()` will just cause memory leak.

using static allocated buffer because
1. they're always available
2. doesn't leak...

### DIFF
--- a/cores/esp8266/core_esp8266_i2s.c
+++ b/cores/esp8266/core_esp8266_i2s.c
@@ -1,11 +1,11 @@
-/* 
+/*
   i2s.c - Software I2S library for esp8266
-  
+
   Code taken and reworked from espessif's I2S example
-  
+
   Copyright (c) 2015 Hristo Gochkov. All rights reserved.
   This file is part of the esp8266 core for Arduino environment.
- 
+
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
@@ -52,7 +52,7 @@ struct slc_queue_item {
 
 static uint32_t i2s_slc_queue[SLC_BUF_CNT-1];
 static uint8_t i2s_slc_queue_len;
-static uint32_t *i2s_slc_buf_pntr[SLC_BUF_CNT]; //Pointer to the I2S DMA buffer data
+static uint32_t i2s_slc_buf_pntr[SLC_BUF_CNT][SLC_BUF_LEN]; //I2S DMA buffer data
 static struct slc_queue_item i2s_slc_items[SLC_BUF_CNT]; //I2S DMA buffer descriptors
 static uint32_t *i2s_curr_slc_buf=NULL;//current buffer for writing
 static int i2s_curr_slc_buf_pos=0; //position in the current buffer
@@ -95,9 +95,8 @@ void ICACHE_FLASH_ATTR i2s_slc_isr(void) {
 void ICACHE_FLASH_ATTR i2s_slc_begin(){
   i2s_slc_queue_len = 0;
   int x, y;
-  
+
   for (x=0; x<SLC_BUF_CNT; x++) {
-    i2s_slc_buf_pntr[x] = malloc(SLC_BUF_LEN*4);
     for (y=0; y<SLC_BUF_LEN; y++) i2s_slc_buf_pntr[x][y] = 0;
 
     i2s_slc_items[x].unused = 0;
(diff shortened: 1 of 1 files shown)


everslick:
wouldn't it be better to just free() the buffer in i2s_slc_end()? additionally i2s_slc_begin() could check if the buffer is already allocated and not malloc() it. having a static buffer limits options unnecessarily. (people might enable/disable code at runtime)

Palatis:
hmm... the situation over here is that, i'm actually tried dynamically malloc() buffers, but axTLS want continues big chunk of memory.
if i enable/disable during `i2s_slc_begin()` the memory gets fragmented, and axTLS throws fatal exception.

and currently calling `i2s_begin()` multiple the memory's actually leaked, that might have to be fixed.

gicho:
Some friendly advices that make it easy to decide such questions:
- avoid dynamic memory as much as possible
- obey "Single responsibility" principle - don't add "second" job to the
i2s function (memory management).
Moving "allocation" out of the i2s allows any kind of implementation in the
user code (global, stack, heap, ,,,), fixing it to malloc inside the i2s is
limiting (always heap).

2016-12-23 18:30 GMT+02:00 Victor Tseng <notifications@github.com>:

Palatis:
so @gicho you mean that we should expose something like `i2s_init_buffer(uint32_t * buffers, size_t count, size_t len)` to the user so they can manage the buffer by themselves?

like
```
size_t count = 8, length = 64;
uint32_t * buf = (uint32_t *)malloc(I2S_DATA_SIZE * count * length);
i2s_init_buffer(buf, count, length);
i2s_begin();
/* ... play your music! */
i2s_end();
free(buf);
```

Palatis:
the stack is probably not suitable for i2s buffer anyway.

sample rate @32000 means you have to spit out 32000 samples per second.
if you want to have ~10ms buffer that's 320 samples which is 1.28kb.
that's easily exceeded when doing network tasks.

and you normally have got only 4kb stack (if you do not modify CONT_STACKSIZE).

everslick:
i temporarily allocate as much as 1500 bytes from the stack for dynamically
generating html pages. but fir DNA buffer? nay.

gicho:
Yes, stack is not sutable - in this case. It was just an example to show
the flexibility of the "single responsibility" (functional) design.
In embedded microcontrollers there might be various "memories" - external
RAM, internal RAM, tightly-coupled memories/caches, non-volatile areas.
Even if you prepare separate "heaps" in all of them and keep alloc in the
i2s you still have to supply application-specific flag to i2s -
"USE_FAST_HEAP" or "USE_NVRAM_HEAP". This is platform-specific enum and
exact value is application defined so not good to hardcode.
I would say that is not part of i2s driver to decide (hardcode) which one
to be used.Especially if data in those buffer are used exchange data with
higher level (application).
My proposal - leave buffer management (alloc/free, static(global)) to the
upper layer - the application of higher layer component.


2016-12-24 1:38 GMT+02:00 Victor Tseng <notifications@github.com>:

Palatis:
getting i2s out of the business might just break existing codes, which is bad IMHO.

this one is more like a "hot patch" to address the memory leak issue.

me-no-dev:
Guys I have a totally different concept running on ESP32 that I really actually like. It allocates very little data for the bus to run, and then it uses as much as you give it.

Workflow is very easy:
- Have 2 DMA structs allocated
- Have 1 64 byte buffer allocated and zeroed (silence)
- Attach the zero buffer to both DMA structs and link them with one another
- Configure I2S and give it the structs to work with
- Create Queue to hold the buffers that you want to send to I2S
- Attach to OUT_EOF interrupt and in that interrupt check if there is a buffer waiting and replace the zero buffer to the one in the queue or if the zero buffer is not already attached, attach it so there is silence on the output

igrr:
also relevant discussion about I2S in Arduino in general: https://groups.google.com/a/arduino.cc/forum/#!topic/developers/pER4iiz7WZM
https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/I2S/src/I2S.h

Palatis:
@me-no-dev can't we avoid the 64byte zeroed buffer and just give it addresses those always reads zero?
like `0x3FF10000` or `0x3FF30000` or `0x3FFE0000`?

https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map

me-no-dev:
@Palatis sure :)

devyte:
Closing in favor of #3790 .

PR #2804, opened 2016-12-28
move stdc++ to flash

hreintke:


igrr:
Aside from the indentation mismatch, looks good to me.

PR #2806, opened 2016-12-28
[SSDP] add `schema(Print &) const`

Palatis:
useful when using AsyncWebServer.


earlephilhower:

earlephilhower:
Closing as superceded.

PR #2807, opened 2016-12-28
allow disabling global Serial and Serial1 object

Palatis:
allow the user to disable specific global `Serial` objects to save
memory.

that's 0x1c bytes per object.


igrr:
I think the "arduino" way of doing this is to move each object's definition into a separate cpp file. Then due to -fdata-sections -Wl,gc-sections the objects which aren't used anywhere will not be linked into the output file. I don't particularly mind the approach taken in this MR though...

Palatis:
I took this approach because others does, too.

Such as SSDP and MDNS...

romansavrulin:
@igrr @Palatis Where should I put `NO_GLOBAL_INSTANCES` define get benefit from that? Main sketch file?

Palatis:
in platformio.ini, `build_flags  = -DNO_GLOBAL_INSTANCES`

romansavrulin:
@Palatis thank you, that way is expected. I thought there's some "Arduino Way of doing this"

Palatis:
"the Arduino way of doing this" is leave them there and hope they get stripped by the linker...

d-a-v:
It can't be stripped since it is a global variable with a constructor.
You can put this in a new file ```platform.local.txt```:
```
compiler.c.extra_flags=-DNO_GLOBAL_INSTANCES
```
(next to ```platform.txt```)

PR #2814, opened 2016-12-30
Use correct cast for downcasting reference.

awong-dev:
For downcasting, static_cast<> is the appropriate cast. Using reinterpret_cast<> will NOT correctly adjust the `this` pointer and dereferencing such a value is undefined by spec. See [expr.reinterpret.cast]p7 for the relevant passage. The only legal use of this pointer is in another set of reinterpret_cast expressions that either land it into a numeric value, or back to the original type.


igrr:
Very good catch, thanks for that!

PR #2821, opened 2017-01-02
Add method to ignore TLS verification errors

dshepherd:
Added HTTPClient::setIgnoreTLSVerifyFailure() so that it is possible to ignore TLS fingerprint and domain mistmatch errors, which are sometimes unavoidable during development.

It is set to false by default so existing code should be unaffected.


dshepherd:
Will this PR be considered for inclusion or is there something I'm not doing right?

igrr:
The change itself looks okay to me, however I'm not sure if the case when certificate verification can not be performed is very common.

dshepherd:
Part of my testing server is built on demand and I'm using a self-signed certificate. This means the key's fingerprint changes frequently.

Even though it's a fairly trivial change, I thought it may help someone else somewhere down the line.

igrr:
Ah, now I get the actual problem — HTTPClient library still can not validate the certificate chain. Will adding certificate chain validation help in your case?

gertvantslot:
I will try to explain, why adding certificate chain validation will not work (in my case).

I am using Google Cloud Messaging to send messages from the Esp8266 to a mobile phone. In order to do so I need to make https requests. 

The server is owned by Google. And Google changes it's certificates very frequently (about every 1 - 2 weeks). I do not have an exact schedule of this change. 

Using chain validation, will make the frequency of the update's a bit less. But the intermediate certificate of Google expires on 1/1/2018. I do not know when they will change this certificate. But I do know that when they do, the communication will fail immediately.  

I have a schedule in mind, relaying the certificate verification to remote servers, and receiving an updated hash etc. For most users this is overly complex, and not feasible for the hobby and prototyping projects they are working on. 

**tl;dr** Unless there is a way to do a (simple) complete verification of the certificate, I do not see how a partial verification will not fail frequently. 

Ignoring the verification will help.

dragondaud:
Just FYI, I am using your changes because of trouble validating with Google API services. Even with the correct fingerprint, the verification always fails. I'm only collecting timezone data, and dont care about certificate validity. Using git version of ESP8266 libs, with lwip2 enabled. Works great! THANKS!

[HTTP-Client][begin] url: https://maps.googleapis.com/maps/api/timezone/json?location=38.3818152,-104.696089&timestamp=1510025092&key=<KEY>
[HTTP-Client][begin] host: maps.googleapis.com port: 443 url: /maps/api/timezone/json?location=38.3818152,-104.696089&timestamp=1510025092&key=<KEY>
[HTTP-Client][begin] httpsFingerprint: ‎CE:64:34:A2:49:B6:BF:91:9B:40:50:6F:9A:D4:9E:C4:DD:58:55:78
[HTTP-Client] connected to maps.googleapis.com:443
[HTTP-Client] transport level verify failed but continuing anyway

torntrousers:
@guino 👍

Maddosaurus:
I'm using these changes in a way @dragondaud does and can confirm that they absolutely work as intended.  
As Google and other big companies are pushing HTTPS-only, there is more and more services that only support HTTPS communication. Most of the time, I'm not really interested in checking if the cert chain / fingerprint is real, I'm just forced to use HTTPS to communicate with the Server.   
For the sake of usability (especially for non-technical folks), this PR could simplify HTTPS for many people.
Please merge this PR into the main codebase so that manual installation and patching of the source isn't necessary anymore.  
Thanks!

earlephilhower:
The repo for this pull was deleted, so it can't ever be merged.  PR #4980 will allow for a saner way of doing this than explicitly setting a fingerprint then saying "oh, don't bother if it doesn't match the one I told you to look for" by allowing you to send in a axTLS::WiFiClientSecure or a BearSSL::WiFiClientSecure with the `setInsecure()` validation setting.

Closing.

Issue #2858, opened 2017-01-14
SPIFFS filename length limit?


JamesGKent:
#### Hardware
Hardware:	Wemos D1 R2
Core Version:      git master

### Description
I'm using the latest git version of this library through the arduino 1.6.9 IDE and I've noticed that one of the files  i loaded into the SPIFFS filesystem was not being served when it was requested. when i tried to debug this i found that SPIFFS.exists(filename) was returning false for this particular file even though it was definitely there. however another file with a similar name (3 chars shorter) was being returned just fine, so i renamed the file to shorten the filename and re uploaded it, at which point SPIFFS.exists succeeds.
the filename in question was this:  
iframeResizer.contentWindow.min.js  
and when shortened:  
iframeResizer.content.min.js  

however another file:
iframeResizer.contentWindow.map  
was being returned just fine.

does the SPIFFS filesystem have some inherent limit on the length of filenames?
if so does this include the whole directory (if it was in a subdirectory) or just the filename?
and if so, what is the limit and why?

vicnevicne:
Yes, as indicated in the [readme included with SPIFFS](https://github.com/esp8266/Arduino/blob/4897e0006b5b0123a2fa31f67b14a3fff65ce561/cores/esp8266/spiffs/README.md), it is designed for small systems with sparse RAM (which is the limited factor on ESP8266). Consequently, it has several simplifications and limitations.
Basically, SPIFFS does not support directories. It's just a list of files. 
But contrary to traditional filesystems, the slash (/) is allowed in filenames, so the code mimics listing files in directory "abc" by listing all filenames that begin with "abc/".
Plus, as you have discovered, there is a total limit of 32 chars for filenames (full path, including directories). I even think (tbc) that the 32 include a mandatory '\0' termination char, so that leaves you with 31 usable characters. 
Consequently, it is advised to keep filenames short and not use deep nested directories.
So basically that's not a bug but a known limitation of SPIFFS. I learned it the hard way too and I think it should be added to the [Filesystem documentation](https://github.com/esp8266/Arduino/blob/master/doc/filesystem.md) ...

JamesGKent:
i figured that would probably be the reason, but having a quick look in the docs and the code and not seeing anything jump out at me i figured i would ask the question...
having it added to the documentation would make a lot of sense.

vicnevicne:
[...]

PR #2860, opened 2017-01-14
Added a section on filesystem limitations due to issue #2858

vicnevicne:
Many users (me included) had to learn the hard way that SPIFFS has severe limitations on filename length. Although they can be easily overcome, one has to known their existence, and there wasn't a single line about it in the filesystem documentation, so thought I could help :-)


JamesGKent:
i know its a slightly different issue, but would it make sense to update the upload tool to flag a warning/error if some of the files being uploaded would be too long?

Issue #2871, opened 2017-01-16
ClientConnection uses too much heap when streaming files


joelucid:
It should be possible to stream arbitrary large files to a ClientContext. However this is done via a BufferedStreamDataSource which allocates a buffer in RAM to keep the whole memory chunk requested in get_buffer. These requests can ask for > 2800 bytes at a time.

To preserve memory the size of the chunk requested should be limited. The following patch achieves that:

diff --git a/libraries/ESP8266WiFi/src/include/ClientContext.h b/libraries/ESP8266WiFi/src/include/ClientContext.h
index d2ee090..df12760 100644
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -346,6 +346,7 @@ protected:
             can_send = 0;
         }
         size_t will_send = (can_send < left) ? can_send : left;
+		if( will_send > 256 ) will_send = 256;
         if (will_send) {
             const uint8_t* buf = _datasource->get_buffer(will_send);
             err_t err = tcp_write(_pcb, buf, will_send, TCP_WRITE_FLAG_COPY);

igrr:
Actually it will not pull the *whole* file into memory. It should never allocate more than 2 * TCP_MSS bytes at a time (that's 2920 bytes).

joelucid:
I see. Still that's quite a bit of memory for an embedded system. It caused my application to break. There's very little benefit to using a larger block size (see https://github.com/pellepl/spiffs/wiki/Performance-and-Optimizing) for the SPIFFS case.

igrr:
This optimization is mainly for TCP throughput, not for SPIFFS performance. I'll see if there is a nice way to configure TCP window size individually for each WifiClient connection, so that you can dial down buffer size of the application requires that.

joelucid:
Is that really a necessary optimization for TCP throughput? The tcp_write documentation has this to say:

>  * Write data for sending (but does not send it immediately).
> 
>  * It waits in the expectation of more data being sent soon (as
>  * it can send them more efficiently by combining them together).
>  * To prompt the system to send data now, call tcp_output() after
>  * calling tcp_write().

So shouldn't one just write in smaller chunks until MSS is reached and then call tcp_output to flush?

> I'll see if there is a nice way to configure TCP window size individually for each WifiClient connection, so that you can dial down buffer size of the application requires that.

[...]

PR #2874, opened 2017-01-17
ClientConnection uses too much heap when streaming files #2871

joelucid:


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -345,16 +346,21 @@ class ClientContext
         if (_pcb->snd_queuelen >= TCP_SND_QUEUELEN) {
             can_send = 0;
         }
-        size_t will_send = (can_send < left) ? can_send : left;
-        if (will_send) {
-            const uint8_t* buf = _datasource->get_buffer(will_send);
-            err_t err = tcp_write(_pcb, buf, will_send, TCP_WRITE_FLAG_COPY);
-            _datasource->release_buffer(buf, will_send);
+		size_t will_send = (can_send < left) ? can_send : left;
```
igrr: These lines seem to use TABs for indentation, but the original file uses spaces. Could you please reformat to match indentation of the surrounding code?

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -333,6 +333,7 @@ class ClientContext
         return _written;
     }
 
+#define TCP_WRITE_CHUNK_SIZE 256
```
igrr: Can we change this to a ClientContext member variable? Then we can later add an API to WiFiClient to configure the chunk size at runtime. This variable can be default-initalized to 256 for now.

joelucid:
Both requested changes done.

PR #2877, opened 2017-01-17
enable loading of root-CAs, certificates and keys from byte array...

sauttefk:
instead of SPIFFS-stream

Something like this is now possible:
```c
const size_t DST_Root_CA_X3_len = 846;
const uint8_t DST_Root_CA_X3[] = {0x30, ....., 0xfd}; 
client.loadCACert(DST_Root_CA_X3, DST_Root_CA_X3_len);
```


igrr:
Sorry, isn't this what `setCertificate` et al currently do? I am probably missing something, how are these new functions different?

sauttefk:
@igrr `setCertificate` sets the certificate the client presents to the server => server can authenticate the client.

`loadCACert` allows the client (esp8266) to validate the keychain presented by the server, but this is only possible by using a SPIFFS-stream. This little patch allows the usage of byte arrays for the certificates.

```c
    bool loadCACert(Stream& stream, size_t size);
 +  bool loadCACert(const uint8_t* pk, size_t size);
```

the `loadCertificate` and `loadPrivateKey` functions with byte arrays are just for completeness of the API.

igrr:
My point was that the new functuon which sets CACert from byte array should be called `setCACert`, to match the existing `setCertificate` and `setPrivateKey`.
Adding `loadCertificate` and `loadPrivateKey` isn't necessary because that's what `setCertificate` and `setPrivateKey` already do. `set*` functions take byte arrays as arguments, `load*` functions take streams as arguments.

sauttefk:
OK, got you point.
I will change the pull request.

sauttefk:
New Pull Request

PR #2880, opened 2017-01-19
Add an LLMNR responder implementation

swarren:
This is a simple implementation of an LLMNR responder for the ESP8266
Arduino package. Only support for advertizing a single hostname is
currently implemented.

LLMNR is a very similar protocol to MDNS. The primary practical
difference is that Windows systems (at least Windows 7 and later;
perhaps earlier) support the protocol out-of-the-box, whereas additional
software is required to support MDNS. However, Linux support is
currently more complex, and MacOS X support appears non-existent.

### DIFF
--- a/libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
+++ b/libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
@@ -1,0 +1,281 @@
+/*
+ * ESP8266 LLMNR responder
+ * Copyright (C) 2017 Stephen Warren <swarren@wwwdotorg.org>
+ *
+ * Based on:
+ * ESP8266 Multicast DNS (port of CC3000 Multicast DNS library)
+ * Version 1.1
+ * Copyright (c) 2013 Tony DiCola (tony@tonydicola.com)
+ * ESP8266 port (c) 2015 Ivan Grokhotkov (ivan@esp8266.com)
+ * MDNS-SD Suport 2015 Hristo Gochkov
+ * Extended MDNS-SD support 2016 Lars Englund (lars.englund@gmail.com)
+ *
+ * License (MIT license):
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


igrr:
Thanks for contributing this library!
I would like to know whether you plan to maintain it in the long run. Also, what is your expectations about the rate of changes to this library in the future. If you plan to make a lot of changes, it may be easier for you to distribute this library via Arduino Library Manager, which allows updating library independent of the boards manager package. As you may have noticed, our release cadence has been somewhat lacking.

swarren:
I don't expect to make many more changes to the code; it seems to work as-is. Of course, there are some violations of the RFC (unimplemented features), but they shouldn't cause problems in practice. I don't know if anyone else will implement them. I'd guess that this library would see roughly the same rate of changes as the existing MDNS code, since it does roughly the same thing.

About the only changes I might guess at are:
- Add an end() API, or an API to change the hostname. If I did this, I'd make the same change to MDNS too, so splitting this out into a separately repo/package wouldn't help your workload.
- Implement LLMNR queries; the current code only supports responding to queries. I don't need this feature though, so I doubt I'd implement it.

I see an advantage of keeping this library with all the other ESP8266 libraries. It's easier for people to find and install just one extra thing in the Arduino IDE, rather than having to track down each library separately. It also makes it easier to test all the libraries and core ESP8266/SDK code together since everything is part of the same release, so there's no mixing/matching versions.

Honestly, I'm not really looking to pick up maintenance of anything much. As I'm sure you're aware, that rather takes time away from doing other stuff. I'm happy to answer questions or fix issues if they come up though, to keep the code working.

libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
```diff
@@ -0,0 +1,284 @@
...
+bool LLMNRResponder::begin(const char* hostname) {
+    // Max length for a single label in DNS
+    if (strlen(hostname) > 63) {
+        return false;
+    }
+
+    _hostname = hostname;
+    _hostname.toLowerCase();
+
+    WiFi.onStationModeGotIP([this](const WiFiEventStationModeGotIP& event){
```
igrr: You need to save the return value as a member variable of the class. The callback lifetime is controlled by the returned `EventHandler` value.
Same thing a few lines below.

libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
```diff
@@ -0,0 +1,284 @@
...
+    ip_addr_t multicast_addr;
+    multicast_addr.addr = (uint32_t)LLMNR_MULTICAST_ADDR;
+
+    if (igmp_joingroup(IP_ADDR_ANY, &multicast_addr) != ERR_OK)
+        return false;
+
+    _conn = new UdpContext;
+    _conn->ref();
+
+    if (!_conn->listen(*IP_ADDR_ANY, LLMNR_PORT))
```
igrr: It's a minor thing, but in some other places you put braces around one-line if bodies. Suggest doing the same thing everywhere.

swarren:
New version uploaded to address the review comments, thanks.

swarren:
Great, thanks!

PR #2883, opened 2017-01-19
Fixed libb64 decoder.

rodionkvashnin:
Libb64 decoder works when "char" type is signed. In esp8266 xtensa gcc "char" is unsigned, so libb64 decoder does not work. In the implementation file I replaced "char" to "int8_t" and created function wrappers in order to keep existing interface.

I  tested it, it works!

### DIFF
--- a/cores/esp8266/libb64/cdecode.c
+++ b/cores/esp8266/libb64/cdecode.c
@@ -86,17 +86,14 @@ static int base64_decode_chars_signed(const int8_t* code_in, const int length_in
   return len;
 }
 
-int base64_decode_value(char value_in)
-{
-	return base64_decode_value_signed(*((int8_t *) &value_in));
+int base64_decode_value(char value_in){
+  return base64_decode_value_signed(*((int8_t *) &value_in));
 }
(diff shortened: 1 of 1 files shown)


cores/esp8266/libb64/cdecode.c
```diff
@@ -77,10 +78,25 @@ int base64_decode_block(const char* code_in, const int length_in, char* plaintex
...
   base64_init_decodestate(&_state);
-  int len = base64_decode_block(code_in, length_in, plaintext_out, &_state);
+  int len = base64_decode_block_signed(code_in, length_in, plaintext_out, &_state);
   if(len > 0) plaintext_out[len] = 0;
   return len;
 }
+
+int base64_decode_value(char value_in)
+{
+	return base64_decode_value_signed(*((int8_t *) &value_in));
```
igrr: the rest of the file doesn't seem to use tabs for indentation

rodionkvashnin: Oops, sorry for this.

PR #2892, opened 2017-01-22
Update keywords.txt for coloring syntax

wosk:
Add ESP keywords and update ESP8266Wifi library


igrr:
This is fantastic! I think Arduino IDE has a feature that it also recognizes keywords.txt for the core directory. If this indeed works, we can move the keywords file from esp8266 library into the core directory; but at any rate this way is also good.

Thinking about this further, would it be possible to keep these files up to date automatically, e.g. using tools like Sphinx?

wosk:
I not found how keywords.txt must placed at core directory. That's why I use library directory. 
I'm not familiar with the Spinx tool. But using it is good idea.

PR #2904, opened 2017-01-28
Update filesystem.md

seanspotatobusiness:
I added a part about placing the ESP8266 module in bootload mode which is something worth including for confused noobs like me.

### DIFF
--- a/doc/filesystem.md
+++ b/doc/filesystem.md
@@ -82,13 +82,14 @@ For more details on the internals of SPIFFS implementation, see the [SPIFFS read
 *ESP8266FS* is a tool which integrates into the Arduino IDE. It adds a menu item to *Tools* menu for uploading the contents of sketch data directory into ESP8266 flash file system.
 
 - Download the tool: https://github.com/esp8266/arduino-esp8266fs-plugin/releases/download/0.2.0/ESP8266FS-0.2.0.zip.
-- In your Arduino sketchbook directory, create `tools` directory if it doesn't exist yet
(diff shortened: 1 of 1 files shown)


igrr:
I'm not sure if this extra sentence actually clarifies something. If anything, i didn't get any instructions for placing the board into bootloader mode from my supplier. Also the concept of placing the board into bootloader mode may not be known to people who use boards with a circuit to enter download mode automatically.
How about changing this along the lines of "do whatever you normally do when uploading the sketch", which may include manual placing of the board into download mode, for some boards?

seanspotatobusiness:
There are boards that enter download mode automatically? I only have an Adafruit HUZZAH ESP8266 and I think you have to do it manually be pressing buttons. I think it would be good to rephrase the instructions in the way you suggested so it makes it clear to people who have to manually start the bootload mode without confusing people whose devices do it automatically.

davisonja:
Depends on what you consider automatic... my NodeMCU's don't require any
physical interaction, it's all done by the IDE (or esptool).

igrr:
Huzzah should be capable of entering download mode automatically, iirc. I don't have one myself, but if there are other huzzah users reading this, please share your experiences.

seanspotatobusiness:
The code I need to use was made for the Arduino so maybe it's a difference between the Arduino and NodeMCU code/IDE?

vicnevicne:
I would suggest : 

> - If your board requires you to press a button (or other) when flashing a sketch, do the same operation now

Just my 2 cents

igrr:
Yep, the sentence by @vicnevicne sounds OK.

Issue #2910, opened 2017-01-30
WifiEvent Handling


tkerby:
### Description

Is there a way to use the WiFi Event Handling to recover the RSSI and Mac Address of received probe requests?

Specifically I would like to setup an AP using SoftAP and monitor for any probes it receives, logging the time, MAC and RSSI.

I've taken a good look at the WifiEvent example for a client but it doesn't cover how to access the data in the structures associated with the events. In particular I'm looking to read from this data in ESP8266WiFiType.h

+struct WiFiEventSoftAPModeProbeRequestReceived
+{
+    int rssi;
+    uint8 mac[6];
+};

Ideallu I'd like to expand this to both run as a dummy AP, but connect to an access point as a client to get the time from NTP and forward Probe requests to an MQTT server

tkerby:
@igrr I think you may be working on this. Has there been any progress?

igrr:
No, I'm not working on this issue. I don't know of a way to monitor probe requests in the non-OS SDK in SoftAP mode. So no clue how this can be implemented.

tkerby:
@igrr Apologies - I saw your comment here: https://github.com/esp8266/Arduino/issues/2100#issuecomment-224143871 and thought you were referring to passing the structures associated with events into the event handler so this data could be read.

I know it's possible in LUA and by using the SDK directly but I'd like to find a way to do this in Arduino

igrr:
Sorry — I haven't understood your question correctly last time. Opened #2917 which implements the method you need. If you're using the git version of ESP8266 Arduino core, you can give it a go by doing `git fetch origin && checkout feature/wifi_probereq_event`.

tkerby:
That works perfectly! Thanks very much for your help with this @igrr

igrr:
Thanks @devyte, indeed this has been implemented and is available in master. Applied the right labels.

d-a-v:
@tkerby's issue seems to be solved.
@devyte any feedback on your side ?

PR #2917, opened 2017-02-02
wifi: add SoftAPModeProbeRequestReceived event handler

igrr:
- add probe request event handler (#2910)
- update WiFi events handling example to use new handler

Pro tip: replace blinking LED with ‘analogWrite’ and connect the pin to
a loudspeaker (or use a servo to hit a bell). Get notified when someone
with a smartphone wanders around your country house (yourself
included).


reaper7:
👍

PR #2954, opened 2017-02-12
add missing loadCACert template

john7doe:


igrr:
This function seems to be already present in latest git version. Thanks for the PR anyway!

PR #2956, opened 2017-02-13
Fix: urlDecode done before parsing args

DottoreTozzi:
Currently, urlDecode() is being called before parsing arguments by '&'.
This is causing problems when passing a string containing special characters, strong passwords for example (WPA2-PSK in my particular scenario).
This tiny patch fixed it for me, while otherwise I'd have to use enctype="multipart/form-data" as a workaround.
I hope this won't break anything else (I've tested it thoroughly, however).

### DIFF
--- a/libraries/ESP8266WebServer/src/Parsing.cpp
+++ b/libraries/ESP8266WebServer/src/Parsing.cpp
@@ -91,7 +91,7 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
   String searchStr = "";
   int hasSearch = url.indexOf('?');
   if (hasSearch != -1){
-    searchStr = urlDecode(url.substring(hasSearch + 1));
+    searchStr = url.substring(hasSearch + 1);
     url = url.substring(0, hasSearch);
   }
   _currentUri = url;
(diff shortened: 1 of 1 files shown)


universam1:
@igrr Any reason why this is not getting merged? Severe bug imho. Thanks!

igrr:
Paging @me-no-dev to review this.

Ideally, a test case for this in https://github.com/esp8266/Arduino/blob/master/tests/device/test_http_server/test_http_server.ino would be nice to have, but I'm not going to ask @DottoreTozzi to figure out how the testing framework works :)

DottoreTozzi:
I actually might, once I'm back from holidays (mid August).

me-no-dev:
I approve this merge :) had stumbled upon similar issue on the Async front recently ;)

universam1:
thanks @me-no-dev @igrr !!

pieman64:
@igrr will this commit also cover https://github.com/esp8266/Arduino/pull/3398 or are they separate issues?

igrr:
@pieman64 Not sure. @me-no-dev and @teejaydub please check and advise!

teejaydub:
No, it's a similar area of the code, but two distinct issues.

PR #3001, opened 2017-02-28
Add WiFiServerSecure and ESP8266WebServerSecure, SSL enabled servers for HTTPS/SSH/etc.

earlephilhower:
Howdy,

First of all, thanks for the amazing work in making the ESP8266 toolchain very approachable.  I'd like to give back with a change that I did this weekend to support a SSL server (useful for SSH or HTTPS) mode of operations.

Example code is included with a self-signed certificate generation script to get users up to speed quickly.


earlephilhower:
Latest commit adds in a new HTTPS server class, ESP8266WebServerSecure and example HelloServerSecure.  Both use the WiFiSecureServer in the previous commits.  I learned more than I wanted to about C++ object slicing and hidden class variables, thanks to the existing object hierarchy, but it's run a loop of WGET's against the  HelloServerSecure overnight on my system.

earlephilhower:
Added and tested a SSL-encrypted SecureWebUpdater example.  Since the SecureWebServer class is passed by pointer, there's only 2-lines of change (+key/cert addition) to enable secure, SSL encrypted, password-protected ESPP updates.

I'll squash everything to a single commit closer to the 2.5.0 release for easier merging, assuming it's accepted.

igrr:
Thanks, this is a really awesome feature. I intend to review and merge this (along with other changes, such as @d-a-v 's LwIP PR) once 2.4.0 is released, which hopefully can be done soon — the release candidate has been out for a few weeks and no major issues (aside from memory use) have been found.

earlephilhower:
Happy to help out. Hopefully this will make it easier for everyone to make more secure IoT things!

jmseight:
Hi,
I would like to use the code that you created.  I downloaded the Arduino master but cannot find the code for WiFiServerSecure or ESP8266WebServerSecure, and cannot find them using the search function in GitHub.  I do have  WiFiClientSecure.  Would you please tell me how to get the code.
Thanks,
James

earlephilhower:
@jmseight , You'll need to use GIT and get the head of the main branch then use the GitHub instructions to get this pull (#3001) merged into your local copy.  See https://help.github.com/articles/checking-out-pull-requests-locally/ for more info.

jmseight:
Sorry for not understanding how to get the items.  I tried "git fetch origin pull/#3001/esp8266:master" and got "fatal: Not a git repository (or any of the parent directories): .git", for "Modifying an inactive pull request locally".

Also, I cannot find "Near the bottom of the pull request, in the merge box, click command line instructions. Follow the sequence of steps to bring down the proposed pull request." as instructed for "Modifying an active pull request locally".

earlephilhower:
@jmseight, I think you just didn't change into the cloned Arduino directory.  Here's my exact command output:

earle@server:/tmp$ git clone **https://github.com/esp8266/Arduino.git**
Cloning into 'Arduino'...
remote: Counting objects: 14632, done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 14632 (delta 11), reused 19 (delta 7), pack-reused 14597
Receiving objects: 100% (14632/14632), 36.66 MiB | 8.23 MiB/s, done.
Resolving deltas: 100% (8477/8477), done.
Checking connectivity... done.
earle@server:/tmp$ **cd Arduino/**
earle@server:/tmp/Arduino$ **git fetch origin pull/3001/head:pull3001**
From https://github.com/esp8266/Arduino
  [new ref]         refs/pull/3001/head -> pull3001
earle@server:/tmp/Arduino$ **git checkout pull3001**
Switched to branch 'pull3001'
earle@server:/tmp/Arduino$ **head libraries/ESP8266WiFi/src/WiFiServerSecure.cpp** 
/*
  WiFiServerSecure.cpp - SSL  server for esp8266, mostly compatible
                   with Arduino WiFi shield library
...

jmseight:
Hi,

Thank you for the instructions.  I was able to download the items.  

Sorry for not understanding this well.  However, when I tried to Include Libraries as ZIP file and chose the folder, but I get errors with enum of WEP (in user_interface.h) and errors with udp.h.  

I tried to replace user_interface.h and udp.h with the ones that I got from the download, but I get many other errors with the library.

Currently I am using V2.3.0 of the ESP8266 board SW.  I think the version that you are using have is 3.2.3, correct?  I can only get up to 2.3.0 form the Board manger in Arduino IDE on mac.  Is this why there is a conflict?

Thanks
James
---
James H-J Yeh, Ph.D.
Assistant Professor, Engineering and Computer Science
Azusa Pacific University

On Aug 16, 2017, at 9:21 PM, Earle F. Philhower, III <notifications@github.com> wrote:

@jmseight <https://github.com/jmseight>, I think you just didn't change into the cloned Arduino directory. Here's my exact command output:

earle@server:/tmp$ git clone https://github.com/esp8266/Arduino.git <https://github.com/esp8266/Arduino.git>
Cloning into 'Arduino'...
remote: Counting objects: 14632, done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 14632 (delta 11), reused 19 (delta 7), pack-reused 14597
Receiving objects: 100% (14632/14632), 36.66 MiB | 8.23 MiB/s, done.
Resolving deltas: 100% (8477/8477), done.
Checking connectivity... done.
earle@server:/tmp$ cd Arduino/
earle@server:/tmp/Arduino$ git fetch origin pull/3001/head:pull3001
From https://github.com/esp8266/Arduino <https://github.com/esp8266/Arduino>
[new ref] refs#3001/head -> pull3001
earle@server:/tmp/Arduino$ git checkout pull3001
Switched to branch 'pull3001'
earle@server:/tmp/Arduino$ head libraries/ESP8266WiFi/src/WiFiServerSecure.cpp
/*
WiFiServerSecure.cpp - SSL server for esp8266, mostly compatible
with Arduino WiFi shield library
...

chaitanyanettem:
@igrr Hi, Any update on when this might be merged?

igrr:
@chaitanyanettem As soon as 2.4.0 is released. Can't convert that to date/time, sorry.

cube8875:
Hi, 

I did everything in the earlephilhower manual and get this message:

In file included from sketch/ESP8266WebServerSecure.cpp:28:0:
sketch/ESP8266WebServerSecure.h:28:30: fatal error: WiFiServerSecure.h: No such file or directory
 #include <WiFiServerSecure.h>
                              ^
compilation terminated.
exit status 1

earlephilhower:
@cube8875 This is a pull request and not yet merged.  You'll need to follow github procedures for checking out a pull over the GIT clone of the project, and ensure that your Arduino is using the git clone for its ESP8266 toolchain and not the release one from the JSON file.

But, since it seems there was a change in WCS where a variable was renamed, you'll have to wait, anyway, until I can resolve the merge conflicts.  I'll try and find time this weekend to look at it.

earlephilhower:
Failure looks due to VM load on the CI system:
> The job exceeded the maximum time limit for jobs, and has been terminated.

Actual run doesn't show any errors.  I can't find a way to resubmit, so I'll wait for some more origin/master changes and then merge them to cause a new rebuild later.

bluemurder:
Thank you very much for your great work! I appreciate it very much.

I'm interested in secure IoT systems, and I'm investigating the possibility to use elliptic curve-based certificates with your code.
Currently you worked on scripts to generate rsa-512 key pairs, but I noticed the possibility to use x509 certificates to build a WifiServerSecure object. In your opinion a certificate generated in a similar way to https://www.guyrutenberg.com/2013/12/28/creating-self-signed-ecdsa-ssl-certificate-using-openssl/ could work?
The main reason of it is that using an ecdsa 256-bit certificate the security level is similar to the one obtained with rsa 3072-bit keys (see http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r4.pdf, page 53).
Currently I have no idea if this is currently supported by your code and if the memory resources of esp8266 are enough. My interest will be to carry on some benchmarks by varying ras key size and monitoring https server speed/resources usage, and compare them with the ones obtained when using ecsda signed certificates.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -68,15 +69,17 @@ class FS;
 
 class ESP8266WebServer
 {
+  friend class ESP8266WebServerSecure; // HTTPS server needs access to privates
```
stackfull: Wouldn't it be better to make the privates protected instead of using friend class?

earlephilhower:
@bluemurder Thanks for the idea.  Unfortunately, after looking at the sources, it seems like AXTLS does not support EC cryptography.  So that seems like it's not going to happen.  There is, however, talk of using BearSSL instead of AXTLS at some point in the future (see https://github.com/esp8266/Arduino/issues/3490 ).  When/if that happens, it'll "just work."

As for the WiFiClientSecure API changes to be more uniform, I'd rather not change them as I'm sure there are tons of apps that depend on the current behavior.  Such a change would be a completely different PR than this one, and would probably encompass a lot more changes.  I've no plans, but if you want then you may want to try it on your own and submit your own PR.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
earlephilhower: @stackfull At the time of writing, I wanted to minimize any impact it would have on the parent class like changing the semantics of existing member variables/functions.

Practically, though, you've got a point here.  I'll have to do some testing first but I'll try and update this PR later in the week.

earlephilhower:
Looks like build failed due to disk space (quota) issues.  Ouch.  Happy New Year, time to rm -rf something. :(

Will retry on a later merge.

christophejourdan:
Hi,

In the example HelloServerSecure.in, how did you generated the certificate and key? Is there a way to generate one dynamically which would use the mDNS domain and other informations that would be variables set at the setup from a configuration file?

earlephilhower:
@christophejourdan I don't think that you can generate the certificate on the ESP itself, if that's what you need.  I don't think we have enough RAM or entropy on the device, and AFAIK there is no AXTLS code to actually generate a new cert.

That said, what I did was generate the certificate on my own desktop/server.  There's a script, make-cert.sh, in the patch which generates a new one with the openssl tools.  You can look at that and it should be obvious where your own domain/etc. would go.

If you already have a cert you want to use, you can just use that as well.  The make-cert.sh script bottom half takes the binary files and generates a C-parsable #include, you can take that bit for your own use to make the files for your pre-generated cert.

christophejourdan:
@earlephilhower I see, thank you for the fast answer!

d-a-v:
2.4.0 is released :)

earlephilhower:
Thanks, @d-a-v.  But if it's not disk out of space errors or job timeout errors, then it's pypi connectivity issues.  :(  I'll throw another rebuild next time I see a change in master.

On a side note, I *think* you can squash the commit into a single one in a merge, right?  Or should I now do my own squash?  There's probably too much sausage-making in the commit history...

d-a-v:
I'm new to this. Usually I squash myself my own PR but it appears that github does the job for the merger. 
And yes connectivity failed. That's unfortunate, all the previous merges succeeded.

earlephilhower:
@d-a-v  I just did a manual squash and there a few merge conflicts that need to be fixed as well as one new member function that was added that accesses a renamed/semantically changed variable.  Nothing bad, surprisingly.

Unfortunately, it seems like the change that @igrr did to fix issue #4024 breaks SSL server mode.  I'm going to if() around it in the case of _isServer and do some testing before push --force'ing the patch.

earlephilhower:
Did a complete squash, and a few fixes to make to 100% w/2.4.0.  It should be easier to merge now.

I built and ran through all the new SSL examples and a couple of the existing ones (to make sure I didn't break anything) and it seems good now.  Able to run plain SSL server, WebServerSecure, and do a HTTPS enabled Update() without incident under a variety of Chrome, Links, and Firefox instances.

libraries/ESP8266HTTPUpdateServer/examples/SecureHTTPSUpdater/SecureHTTPSUpdater.ino
```diff
@@ -0,0 +1,144 @@
...
+const char* update_username = "admin";
+const char* update_password = "admin";
+const char* ssid = "........";
+const char* password = "........";
+
+ESP8266WebServerSecure httpServer(443);
+ESP8266HTTPUpdateServer httpUpdater;
+
+// The certificate is stored in PMEM 
+static const uint8_t x509[] ICACHE_RODATA_ATTR = {
```
igrr: perhaps `PROGMEM` could be used here, similar to the other sketch?

libraries/ESP8266HTTPUpdateServer/examples/SecureHTTPSUpdater/SecureHTTPSUpdater.ino
```diff
@@ -0,0 +1,144 @@
+/*
+  To upload through terminal you can use: curl -u admin:admin -F "image=@firmware.bin" esp8266-webupdate.local/firmware
+*/
```
igrr: Please add a public domain/CC0 licence header to this file.
E.g.

https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiEvents/WiFiEvents.ino#L12-L13

libraries/ESP8266HTTPUpdateServer/examples/SecureHTTPSUpdater/SecureHTTPSUpdater.ino
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  MDNS.begin(host);
+
+  httpServer.setServerKeyAndCert_P(rsakey, sizeof(rsakey), x509, sizeof(x509));
+  httpUpdater.setup(&httpServer, update_path, update_username, update_password);
+  httpServer.begin();
+
+  MDNS.addService("https", "tcp", 443);
+  Serial.printf("HTTPSUpdateServer ready! Open https://%s.local%s in your browser and login with username '%s' and password '%s'\n", host, update_path, update_username, update_password);
```
igrr: Could this string be broken into multiple lines for better readability in the sketch?

libraries/ESP8266WebServer/examples/HelloServerSecure/HelloServerSecure.ino
```diff
@@ -0,0 +1,136 @@
+#include <ESP8266WiFi.h>
```
igrr: Likewise, need a license header, a brief description what this sketch does, and instructions for generating certificate and key arrays.

Also, a note that browsers will not (by default) trust a self-signed certificate might be useful.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -73,10 +73,10 @@ class ESP8266WebServer
   ESP8266WebServer(int port = 80);
   ~ESP8266WebServer();
```
igrr: AFAIK the destructor also needs to be virtual.

libraries/ESP8266WebServer/src/ESP8266WebServerSecure.cpp
```diff
@@ -0,0 +1,149 @@
...
+  if (_currentHeaders)
+    delete[]_currentHeaders;
+  _headerKeysCount = 0;
+  RequestHandler* handler = _firstHandler;
+  while (handler) {
+    RequestHandler* next = handler->next();
+    delete handler;
+    handler = next;
+  }
+  close();
```
igrr: if the destructor is made virtual, ESP8266WebServerSecure will not need to do parts of the cleanup which are done by ESP8266WebServer.

libraries/ESP8266WebServer/src/ESP8266WebServerSecure.h
```diff
@@ -0,0 +1,55 @@
...
+  void handleClient() override;
+  void close() override;
+
+private:
+  size_t _currentClientWrite (const char *bytes, size_t len) override { return _currentClientSecure.write((const uint8_t *)bytes, len); }
+  size_t _currentClientWrite_P (PGM_P bytes, size_t len) override { return _currentClientSecure.write_P(bytes, len); }
+
+protected:
+  WiFiServerSecure _serverSecure;
+  WiFiClientSecure _currentClientSecure;
```
igrr: Hmm, so ESP8266WebServerSecure will still carry _server and _currentClient members in addition to these new ones. Perhaps we can refactor this by turning ESP8266WebServer into a template, like `ESP8266WebServerImpl<typename TServer, typename TClient>` and then doing a `typedef ESP8266WebServerImpl<WiFiServer, WiFiClient> ESP8266WebServer` and `typedef ESP8266WebServerImpl<WiFiServerSecure, WiFiClientSecure> ESP8266WebServerSecure`.
Small part of the implementation, of course, will need to be moved into the header file.

(I'm okay with doing this refactoring in a subsequent PR)

libraries/ESP8266WiFi/examples/WiFiHTTPSServer/WiFiHTTPSServer.ino
```diff
@@ -0,0 +1,173 @@
+/*
+ *  This sketch demonstrates how to set up a simple HTTP-like server using HTTPS encryption.
+ *  A sample self-signed certificate is included.  For your own application, please be sure
+ *  to GENERATE YOUR OWN AND REPLACE.
```
igrr: Adding some instructions on how to do that would be nice.

Also, same thing with public domain / CC0 license.

libraries/ESP8266WiFi/examples/WiFiHTTPSServer/WiFiHTTPSServer.ino
```diff
@@ -0,0 +1,173 @@
...
+  // Read the first line of the request
+  String req = client.readStringUntil('\r');
+  Serial.println(req);
+  client.flush();
+  
+  // Match the request
+  int val;
+  if (req.indexOf("/gpio/0") != -1)
+    val = 0;
+  else if (req.indexOf("/gpio/1") != -1)
```
igrr: I regret the idea of demonstrating WiFiServer by writing an HTTP server. Many users are adapting that example instead of using ESP8266WebServer and are having trouble because of that. I'm okay merging this, but perhaps we should move towards more basic WiFi{Client,Server}[Secure] examples.

libraries/ESP8266WiFi/src/WiFiClientSecure.h
```diff
@@ -34,6 +34,8 @@ class WiFiClientSecure : public WiFiClient {
   ~WiFiClientSecure() override;
   WiFiClientSecure(const WiFiClientSecure&);
   WiFiClientSecure& operator=(const WiFiClientSecure&);
+  // Only called by WiFiServerSecure
+  WiFiClientSecure(ClientContext* client, bool usePMEM, const uint8_t *rsakey, int rsakeyLen, const uint8_t *cert, int certLen);
```
igrr: Perhaps can be made `protected` and then WiFiServerSecure can be made `friend`?

libraries/ESP8266HTTPUpdateServer/examples/SecureHTTPSUpdater/SecureHTTPSUpdater.ino
earlephilhower: Done.  New header on the 3 examples includes public domain release, info on using and making SSL certs, browser interactions, etc.

libraries/ESP8266HTTPUpdateServer/examples/SecureHTTPSUpdater/SecureHTTPSUpdater.ino
earlephilhower: Done.  No idea why I had that instead of PROGMEM...

libraries/ESP8266HTTPUpdateServer/examples/SecureHTTPSUpdater/SecureHTTPSUpdater.ino
earlephilhower: Done

libraries/ESP8266WebServer/examples/HelloServerSecure/HelloServerSecure.ino
earlephilhower: Done (see above).

libraries/ESP8266WebServer/src/ESP8266WebServer.h
earlephilhower: Done.

Good point.  GCC is normally good about catching non-virtual destructors in this usage, weird that it didn't flag it.  I suppose it depends on where the destruction is done.  In my code and examples I believe the app caller is doing the destruction and knows which one to call, but virtual takes care of this and the following one.

libraries/ESP8266WebServer/src/ESP8266WebServerSecure.cpp
earlephilhower: Done.  Did not know that, thought the vtable only had the final destructor.  Learn something new every day!

libraries/ESP8266WiFi/examples/WiFiHTTPSServer/WiFiHTTPSServer.ino
earlephilhower: Done (see above)

libraries/ESP8266WiFi/examples/WiFiHTTPSServer/WiFiHTTPSServer.ino
earlephilhower: No other way to test it for most people, unfortunately.  Ubuntu and RedHat don't ship with telnet any more by default, and neither does Windows 7+.  Netcat would do, but there's no standard Windows equivalent so you're kind of stuck.

libraries/ESP8266WiFi/src/WiFiClientSecure.h
earlephilhower: Done.

I had a friend class originally, but a reviewer suggested that, as per the C++ FAQ, that "friend class" is very much frowned upon by the C++ community.  Being a hardware person, I took the FAQ at its word.

libraries/ESP8266WiFi/examples/WiFiHTTPSServer/WiFiHTTPSServer.ino
Juppit: To enable telnet in [Windows7](https://social.technet.microsoft.com/wiki/contents/articles/910.windows-7-enabling-telnet-client.aspx) or in  [Windows10](https://social.technet.microsoft.com/wiki/contents/articles/38433.windows-10-enabling-telnet-client.aspx)

libraries/ESP8266WebServer/src/ESP8266WebServerSecure.h
earlephilhower: You're right, it is crufty.  Can we do this on a later PR?  I've used templates, of course, but never had to write my own.  It's probably not as bad as I imagine, but it's something I'd need to spend a bit reading up on.

Another alternative might be to convert to pass-by-reference instead of -by-value.  Then it may "just work" without worrying if it's a WiFiServer or WiFiServerSecure in the methods that don't care about SSL.  That's probably a larger change, though, and adds dynamic memory allocations (but String is used everywhere so I'd say that worry is not a big deal).

devyte: Every time I see inheritance and virtual etc in an embedded project I cry a little. Inheritance is really ill suited for systems that are really memory constrained. There are better ways.
Having said that, I don't mind handling that change separately. I've been sort of thinking about an overhaul of some classes into a non-inheritance implementation anyways, but that would be a Big Change, and right now I'd like to stay focused on stability and cleanup.

libraries/ESP8266WebServer/src/ESP8266WebServerSecure.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+    _serverSecure.setServerKeyAndCert(key, keyLen, cert, certLen);
+}
+
+ESP8266WebServerSecure::~ESP8266WebServerSecure() {
+  // Nothing to do here.
+  // Base class's destructor will be called to clean up itself
+}
+
+// We need to basically cut-n-paste these from WebServer because of the problem
```
devyte: This makes no sense to me. Where is there an assignment such as ESP8266Webserver = ESP8266WebServerSecure? If there is anywhere where an assignment is done, then there is likely a bug there.

earlephilhower: The problem is not the webserver, but the WiFiServer.available() model...
````
...
    WiFiClient(Secure) client = _server(Secure).available();
...
````
Because WiFiServer and it's children return a copy of a object, not a pointer to a new()'d one, the caller needs to know how big it will be/etc.  So the WebServerSecure class can't fall back on the WebServer::handle() method and let the vtable work its magic.  Object slicing, if I remember the term properly.

If there was an method like "virtual WiFiClient *WiFiServer.available();" it would be easier to reuse the WebServer code.  The Secure HTTPS Updater example shows this use.

Comment (devyte) [reply libraries/ESP8266WebServer/src/ESP8266WebServerSecure.cpp]: I know about object slicing. When you encounter it, it's usually a sign of a bug or of bad design (not always).
I should investigate this. Sigh, so much to do!
In any case, not a fault in your implementation, which is why I only commented and approved.

File: libraries/ESP8266WebServer/src/ESP8266WebServerSecure.h (hunk 8 above)
Comment (igrr) [reply libraries/ESP8266WebServer/src/ESP8266WebServerSecure.h]: Yes, I'm okay doing this in a later PR. Just thought I'd mention this here. Passing things by reference will indeed help with code duplication.

File: libraries/ESP8266WebServer/src/ESP8266WebServerSecure.cpp (hunk 12 above)
Comment (igrr) [reply libraries/ESP8266WebServer/src/ESP8266WebServerSecure.cpp]: This can also be solved with the template-based implementation and passing by reference, to some extent. Plus a few helper methods to clear the current client and get a new one from the WebServer. Will handle in another PR.

**Comment by pieman64** [comment]
@earlephilhower thanks for this PR and @igrr for the commit.
I notice you state "that it use SHA256 and 1024 or 512 bits" but I thought 2056 bits was OK.
Is there an up to date doc which states max key size and a list of acceptable ciphers?

Any chance of seeing an example for remote OTA updates rather than local OTA updates?

**Comment by igrr** [comment]
@pieman64 here's the list of cipher suites supported by axTLS:
https://github.com/igrr/axtls-8266/blob/master/README.md
Regarding "remote OTA updates", do you mean an example of pulling an update binary from remote server?

**Comment by pieman64** [comment]
@igrr thanks for the link to the available cipher suites and the info on key sizes.

Yes I think bin updates from remote servers with SSL are probably more popular than local updates because generally you can trust your own local network.

**Comment by igrr** [comment]
Ok, essentially that's `ESPhttpUpdate.update("https://server/file", version, cert_fingerprint);`

There's some work in progress to make it possible to verify server certificate using root certificate rather than using a fingerprint. Once done, i will add an example about ESPhttpUpdate with root cert verification.

**Comment by pieman64** [comment]
@igrr we are already using the fingerprint system and it's working OK with a basic sketch and bin files hosted on GitHub. The problem is once you start building a meaningful sketch it runs out of memory.

**Comment by igrr** [comment]
The only real solution to that is to allocate most of the things dynamically in the sketch. Then you can go into "update" mode or "normal" mode and enable features separately.

There will be some improvement in memory usage down the road once we fix the support for fragment length negotiation extension. That will, however, require having a server which supports that extension.

**Comment by d-a-v** [comment]
Thanks for this !

**Comment by earlephilhower** [comment]
Thanks all for the input and feedback to make it a better patch.

@pieman64 while it does support 2048 bits, you may run into out of RAM issues during SSL negotiations.  It has worked before (you can change the # of bits in the make-cert.sh script and try w/your own self-signed ones), but practically it puts you in a very tight RAM situation...

@igrr 's comment about dynamically allocating things is exactly what I'm doing in my own self-contained ESP8266 power plug firmware.  To get enough RAM for upgrade w/SSL I have to stop and free pretty much everything when a toggle is set, then after a timeout force a system reset to get back to normal mode in case the upgrade is not done.

**Comment by stefaandesmet2003** [comment]
@earlephilhower 

Hi, i'm unable to get the HelloServerSecure example working correctly, and would greatly appreciate your help. 
I'm testing the https server from chrome on Win10, and always get NET::ERR_CERT_INVALID. (on the info page chrome adds : ... You cannot visit 192.168.1.22 right now because the website sent scrambled credentials that Google Chrome cannot process. ...) "Hello from esp8266 over HTTPS!" is never displayed in the browser.

I modified the example to use a AWS IOT client certificate, and then after a couple of resets and an occasional watchdog, I get the "Hello from esp8266 over HTTPS!", Chome only complaining about net::ERR_CERT_AUTHORITY_INVALID, which is normal for self-signed certificates, right?
I have no clue where to start digging. I added serial debugging and included 2 serial logs below, one from the original example, the other from the same code with my AWS client certificate.

Don't know if the "Error : ..." statement has anything to do with my problem; it seems to come from the axtls library, but i'm using the libaxtls.a from the repo. 

.
Connected to TP-LINK
IP address: 192.168.1.22
MDNS responder started
HTTPS server started
=== CERTIFICATE ISSUED TO ===
Common Name (CN):		127.0.0.1
Organization (O):		ESP8266
=== CERTIFICATE ISSUED BY ===
Common Name (CN):		127.0.0.1
Organization (O):		ESP8266
Not Before:			Sat Mar 18 14:49:18 2017
Not After:			Mon Nov 25 14:49:18 2030
RSA bitsize:			512
Sig Type:			SHA256
State:	receiving Client Hello (1)
State:	sending Server Hello (2)
State:	sending Certificate (11)
State:	sending Server Hello Done (14)
State:	receiving Client Key Exchange (16)
State:	receiving Finished (16)
State:	sending Finished (16)
New secure client
**Error: maximum number of certs added (1) - change of compile-time configuration required**
State:	receiving Client Hello (1)
State:	sending Server Hello (2)
State:	sending Certificate (11)
State:	sending Server Hello Done (14)
State:	receiving Client Key Exchange (16)
State:	receiving Finished (16)
State:	sending Finished (16)
New secure client
pm open,type:2 0
************************************
I modified the example with a AWS IOT client certificate, and then after a couple of resets and an occasional watchdog, I get the "Hello from esp8266 over HTTPS!", with the following serial output :

.
Connected to TP-LINK
IP address: 192.168.1.22
MDNS responder started
Success to open cert file
certificate size = 861
Success to open private cert file
private key size = 1193
HTTPS server started
=== CERTIFICATE ISSUED TO ===
Common Name (CN):		AWS IoT Certificate
Organization (O):		<Not Part Of Certificate>
Basic Constraints:		critical, CA:FALSE, pathlen:10000
Key Usage:			critical, Digital Signature
=== CERTIFICATE ISSUED BY ===
Common Name (CN):		<Not Part Of Certificate>
Organization (O):		<Not Part Of Certificate>
Organizational Unit (OU):	Amazon Web Services O=Amazon.com Inc. L=Seattle ST=Washington C=US
Not Before:			Thu Nov 23 21:56:42 2017
Not After:			Tue Nov 25 17:31:43 1913
RSA bitsize:			2048
Sig Type:			SHA256
State:	receiving Client Hello (1)
State:	sending Server Hello (2)
State:	sending Certificate (11)
State:	sending Server Hello Done (14)
State:	receiving Client Key Exchange (16)
pm open,type:2 0
New secure client
State:	receiving Finished (16)
State:	sending Finished (16)
Error: maximum number of certs added (1) - change of compile-time configuration required
State:	receiving Client Hello (1)
State:	sending Server Hello (2)
State:	sending Certificate (11)
State:	sending Server Hello Done (14)
State:	receiving Client Key Exchange (16)
New secure client
State:	receiving Finished (16)
State:	sending Finished (16)
method: GET url: / search: 
headerName: Host
headerValue: 192.168.1.22
headerName: Connection
headerValue: keep-alive
headerName: Cache-Control
headerValue: max-age=0
headerValue: 1
headerName: User-Agent
headerValue: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36
headerName: Accept
headerValue: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
headerName: Accept-Encoding
headerValue: gzip, deflate, br
headerName: Accept-Language
headerValue: en-GB,en;q=0.9,en-US;q=0.8,nl;q=0.7
args: 
Request: /
 Arguments: 
Error: maximum number of certs added (1) - change of compile-time configuration required
Fatal exception 29(StoreProhibitedCause):
epc1=0x4000df0e, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000005, depc=0x00000000

Exception (29):
epc1=0x4000df0e epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000005 depc=0x00000000

ctx: cont 
sp: 3fff09c0 end: 3fff0d20 offset: 01a0

>>>stack>>>
3fff0b60:  6f727245 6d203a72 6d697861 6e206d75  
3fff0b70:  40208872 666f2072 72656320 61207374  
3fff0b80:  00000005 64252820 202d2029 6e616863  
3fff0b90:  6f206567 6f632066 6c69706d 00ae0001  
3fff0ba0:  00000005 00000000 00000005 40208cb9  
3fff0bb0:  00000320 00000000 3fff533c 4022142b  
3fff0bc0:  00000200 3fff7d4c 3ffefca0 3fff0c40  
3fff0bd0:  3fff8334 3fff220c 00000200 401007ac  
3fff0be0:  3fff8334 3fff220c 3fff533c 40220098  
3fff0bf0:  3fff0c40 0000035d 3fff220c 402220f9  
3fff0c00:  3fff220c 0000035d 3fff37c4 01000000  
3fff0c10:  3fff0c40 000050e7 3fff533c 40221900  
3fff0c20:  40100a2e 00000030 3fff0ca0 3fff0c40  
3fff0c30:  00001387 000050e7 3fff8334 40208c49  
3fff0c40:  3fffdab0 00000000 3fffd9d0 3ffefcec  
3fff0c50:  3fff8fdc 00000000 3fff30d4 000004a9  
3fff0c60:  3fff2d64 3fffdad0 3ffefcec 00000030  
3fff0c70:  00000000 3fff0cd0 3fff368c 3ffefcec  
3fff0c80:  3fffdad0 3fff0cd0 3ffef9f0 40209437  
3fff0c90:  0000035d 00000000 3ffefd00 402080f7  
3fff0ca0:  40107360 00000000 00001388 402086d4  
3fff0cb0:  3ffefa24 3fff8fdc 3fff8334 40208f1a  
3fff0cc0:  00000001 00000000 3ffef910 4020a65e  
3fff0cd0:  40107308 00000000 00003a98 feefeffe  
3fff0ce0:  00000000 00000000 00000000 feefeffe  
3fff0cf0:  3fffdad0 00000000 3ffefce4 402073fc  
3fff0d00:  3fffdad0 00000000 3ffefce4 4020e664  
3fff0d10:  feefeffe feefeffe 3ffefd00 40100a2c  
<<<stack<<<

tail 8
chksum 0x2d
csum 0x2d
v00000000
~ld
scandone

scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 1
cnt 

connected with TP-LINK, channel 6
dhcp client start...
ip:192.168.1.22,mask:255.255.255.0,gw:192.168.1.1
.
Connected to TP-LINK

**Comment by earlephilhower** [comment]
That error's actually quite good compared to the normal "randomly crash when something's wrong" we seem to have everywhere!

Can you put your code up (obfuscate your client cert, of course!) and decode the exception?

You're only allowed a single cert per SSL connection, so on an server.available() it removes the old SSL context + cert and stuffs in its host + pricate key.  It dos its work and then the server's wificlientsecure goes away.  At that point, the orig client will have its connection dropped and reconnect after manually reloading its client certs.  Or at least that's the theory.

**Comment by stefaandesmet2003** [comment]
Hi,
thanks for your fast reply. Do you mean the https server library doesn't work with you either?
To be more precise, I tried 2 versions of the HelloServerSecure example, one the unmodified example code, and a second using a AWS client certificate & private key loaded from spiffs. So i'm only using 1 certificate at a time. But only with the amazon certificate, the browser showed the "Hello from esp8266 over HTTPS!" line, after a couple resets only ..
Here is the code and a trace file for each version. My hardware is a Wemos D1 mini 4MB flash. As for library version : I cloned esp8266-arduino yesterday and copied the whole repository in the 2.4.0 directory. I added #define DEBUG_OUTPUT Serial in ESP8266WebServerSecure.cpp , to make the code work with debug enabled.
Initially i thought the resets with the AWS certificate had to do with the longer key used, but from the stack trace, it looks the cause of the resets is the same for both versions.
Hope you can shed a light on this.
[HelloServerSecure.zip](https://github.com/esp8266/Arduino/files/1693202/HelloServerSecure.zip)

**Comment by earlephilhower** [comment]
Howdy @stefaandesmet2003 .

The HelloServerSecure runs fine for me, as does a SSL MQTT client in parallel not using client SSL certs (I do that w/psychoplug).  Due to axtls SSL limitations, when an incoming web SSL request comes in, the outgoign SSL MQTT link gets dropped (and reconnected next pass thru the loop).

Are you saying your attached ZIP, with the code as-commented, also gives you a failure?  It looks like it's effectively the same as the base example, no?

If you're really brave, in #4273 I've replaced axtls w/BearSSL which allows for multiple SSL contexts and user-defined buffer sizes so you really can have parallel incoming and outgoing connections.  I've run the mqtt_esp8266 example w/client certs on it (to my own server) for several hours w/o incident.

**Comment by stefaandesmet2003** [comment]
Hi @earlephilhower , 
yes, the original example code with the hardcoded certificate/key raises the exception 29 and NET::ERR_CERT_INVALID in Chrome. I have a working SSL MQTT client sketch using AWS client certificate (PubSubClient+WiFiClientSecure); the issue seems to be in the **ServerSecure libraries.
I'm not brave enough right now for your last suggestion :( ; i don't understand enough yet of the tls internals.

**Comment by stefaandesmet2003** [comment]
Hi @earlephilhower ,
I did some more testing. The WiFiHTTPSServer example works consistently without resets. But every first request from the browser is not correctly decrypted : the 'req' string is empty, serial prints 'invalid request', and client.print (http 404) never arrives at the browser. Every second request (probably a retry  from the browser) is correctly decrypted : req : GET /gpio/1 HTTP/1.1. In the output log i added a ESP.getFreeHeap() printout every 5 seconds. Heap is ok here.

The HelloServerSecure example now works without reset for 1 browser request, and then esp8266 resets systematically on the 2nd request. From the same ESP.getFreeHeap() printout every 5s, you see initial heap size of 35k is reduced to 5k after the first browser request, and is not restored. Then the 2nd request (reload page) fails with a reset. 
Main difference between the 2 examples is the ESP8266WebServerSecure library.
Could you please check if you have the same behavior on your hardware?
Cheers,

**Comment by earlephilhower** [comment]
Great, thanks @stefaandesmet2003!

Well, not really *great* but now I can reproduce what you're seeing.  There is obviously a memory leak somewhere in here when used with the WebServerSecure.  There's some ugly refcount stuff going on in all the SSL code (to share the single SSL_CTX memory), and I believe some logic in the Server/Client is borked when the refcnt is higher than expected that I've flubbed.

Since this pull is closed and merged, would you like to open a new issue so it's tracked properly?  Memory leak in WebServerSecure on repeated connections.

**Comment by earlephilhower** [comment]
Anyone running this in their own apps, please check #4305 as I've done significant cleanup and it's now way more stable and usable, allowing parallel client and server connections, each with their own certs, and no interference between the two.

**Comment by stefaandesmet2003** [comment]
thanks for your big effort. Will give it go very soon!

**Comment by MNGregO** [comment]
Thanks to all who have worked on this! I was able to get the HTTPS example running.
Are there any plans for an SSH server example? I've tried to tweak the HelloServerSecure example to become an SSH server (instead of HTTPS) and then tried connecting with Tera Term, but I haven't been able to get it to work. I'm really new to this, so I'm sure it's really simple to switch from HTTPS to SSH, but I can't figure it out.

**Comment by d-a-v** [comment]
Like @pornin [explains](https://security.stackexchange.com/questions/1599/what-is-the-difference-between-ssl-vs-ssh-which-is-more-secure?answertab=votes#tab-top), it is unlikely we can have ssh on the esp8266.
Still, we have have telnet over ssl and this is great.

**Comment by d-a-v** [comment]
Use WiFiClientSecure, not the webserver.

**Comment by MNGregO** [comment]
Thanks again, @d-a-v . I think I have it working now. Next I just have to find a telnet client that is capable of using SSL so I can test it.
If anyone is interested in the final code I came up with, let me know - it looks like we can't upload .ino files here, but I could post it here using "insert code."

**Comment by d-a-v** [comment]
You can try with
```
openssl s_client -connect $ip:$port
```

PR #3007, opened 2017-03-01
Added HEAD method

hawwwran:
Added method HEAD, which allows to get just headers in the response from the server. So the server now accepts HEAD method and strips content ig the request method was HEAD.

I added error code 418 I'm a teapot, which is in RFC and there is no need to ignore it.


earlephilhower:

earlephilhower:
On inspection, we already have a HEAD method in ESP8266WebServer.

https://github.com/esp8266/Arduino/blob/01e9d948c781af0ca22c119be825e0fe622a4b65/libraries/ESP8266WebServer/src/ESP8266WebServer.h#L33

https://github.com/esp8266/Arduino/blob/01e9d948c781af0ca22c119be825e0fe622a4b65/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h#L483

Closing.

PR #3018, opened 2017-03-03
WiFiClientSecure half-duplex contract.

diegopx:
The internal axTLS library is half duplex, but WiFiClientSecure acts as if it was full duplex. This results in subtle bugs due to erroneous memory access. This branch fixes these bugs and establishes the half-duplex contract in the WiFiClientSecure interface.


diegopx:
Cancelling to add the updated libaxtls.a library.
Note: it doesn't seem wise to upload binary compiled libraries to the repo...

Issue #3002, opened 2017-02-28
WiFiClientSecure invalid read buffer


diegopx:
#### Hardware
Hardware:			ESP-8266 in a Sonoff Switch
Core Version:      	2.3.0

### Description

WiFiClientSecure may read from an invalid buffer, causing any number of undefined behaviours.

To compute available(), the internal SSLContext in WiFiClientSecure calls axTLS's [ssl_read(SSL* ssl, uint8_t** in_data)](http://axtls.sourceforge.net/dox/group__c__api.html#gaf369882880d2c7c87fe1fb4f4949798c), which gives access to a temporary internal buffer through in_data (See notes which indicate to use in_data before performing any other ssl_* operations). SSLContext then proceeds to save a reference to the internal buffer in its _read_ptr member field, as a cache for subsequent calls to SSLContext.read().

However, unless the read() calls are performed immediately after calling available(), the reference saved in _read_ptr may not be valid anymore. In particular, if the context is used for writing in between, the reference will be obsolete and calls to WiFiClientSecure.read() will return garbage (or access invalid memory).
The problem is aggravated by the fact that many functions call available(), including connected().

In particular, I'm using a modified PubSubClient to implement a MQTT client (modified to use an internal WiFiClientSecure as its underlying channel). Using a simple loop as shown below halts the program, since by the time the program actually reads the buffer, the MQTT packet's "message length" is an arbitrary number (typically larger than the real message length, which is very small) and the readPacket() function tries to read beyond what the broker has sent, blocking forever (or until the socket times out).

```cpp
void loop() {
  // ...
  // The broker has sent a message to the device (waiting on the ssl socket)
  if (mqtt.connected()) {
    // connected() -> available() -> ssl_read()
    // WiFiClientSecure.read() would return 'device, I love you'
    
    if (!credentials) {
      // ask for credentials
      mqtt.publish(...) // -> WiFiClientSecure.write() -> ssl_write()
    }
    
    // The reference is invalid, WiFiClientSecure.read() may return the complete
    // works of Shakespeare (modulo FLASH_MEMORY_SIZE);
    
    mqtt.loop();  // -> readPacket() -> WiFiClientSecure.read()
    // this line may never be reached, because loop() halted
  }
  else {
    // connect mqtt
  }
}
```

If the order of the internal operations is reversed, everything works fine.

```cpp
void loop() {
  // ...
  // The broker has sent a message to the device (waiting on the ssl socket)
  if (mqtt.connected()) {
[...]

PR #3019, opened 2017-03-03
WiFiClientSecure half-duplex contract.

diegopx:
The internal axTLS library is half duplex, but WiFiClientSecure acts as if it was full duplex. This results in subtle bugs due to erroneous memory access. This branch fixes these bugs and establishes the half-duplex contract in the WiFiClientSecure interface.


thirsch:
+1

Also fixes the call to ssl_client_new and passes the hostname in the correct way. You can't connect to a host with multiple certificates using TLS extension SNI otherwise.

MumblesNZ:
Thanks Diegopx, I was having issues with getting kicked periodically from my MQTT connection (using pubsubclient), and this + your pubsubclient changes seemed to have fixed it.

vshymanskyy:
@igrr could you please take a look at this?
It should fix the major issue we experience using Blynk with SSL.
Thank you!

igrr:
@vshymanskyy could you add a reference to the issue you are facing?

vshymanskyy:
@igrr The issue is well explained here: https://github.com/esp8266/Arduino/issues/3002
We're facing the same behavior as Blynk protocol is full-duplex like MQTT

tripish:
@igrr Do you have any sense of when this might be resolved? I'm encountering the same issue. I'd like to pull in the changes in the interim, but the associated axtls binaries are pretty old compared and I'm not sure I want to dive down the rabbit hole to check the changes and re-compile if it might be resolved soon.

vshymanskyy:
@tripish @igrr https://github.com/esp8266/Arduino/pull/3019 helps a lot. in fact, I couldn't get it to crash in my scenario again.
It would be great to solve the problem for the upcoming release...

igrr:
Closing in favor of #4024

Issue #3004, opened 2017-02-28
Arduino UNO WiFi's ESP8266 is uncompatible with this framework.


gvJaime:
#### Hardware
Hardware:			Arduino UNO WiFi's esp8266
Core Version:      	2.3.0

#### Description of the board

The [Arduino UNO WiFi](http://www.arduino.org/products/boards/arduino-uno-wifi) is an Arduino UNO with an ESP module integrated on the board. (That is, ESP8266, SPI flash memory, everything).

The board is provided with [a firmware](https://github.com/arduino-org/Esp-Link) that allow certain WiFi functionalities to be called from the ATMEGA, but this requires the I2C pins to be used, despite the serial pins being connected between the ESP8266 and the ATMEGA. It makes you miss some funcionalities, so I tried to make a custom firmware using this framework.

This board could be a quite decent integrated solution for having an ESP8266 as main processor of a Thing, and the ATMEGA as a microcontroller, (instead of having an ESP module hanging from an Arduino UNO, with the logical converter and everything) but it seems uncompatible with anything done with this framework.

#### Description of the issue

While loading a custom firmware into the ESP8266 integrated in this board is possible, that ESP8266 won't do anything when loaded with a firmware made with this framework. No wifi, no serial, no GPIO response. Not even debug messages, or stack messages. The ESP8266 won't even heat.

#### What does work

It will however, boot with the [Arduino original firmware](https://github.com/arduino-org/Esp-Link) and the [original Jeelabs firmware](https://github.com/jeelabs/esp-link) this is based on.

Original source-code-examples will boot, but serial communication is difficult due to [the wiring of the UART0 to the USB port](http://download.arduino.org/products/UNOWIFI/Arduino-UNO-WiFi-V4_AF1.pdf).

#### What does not work

Anything made with this framework.

Using eboot to boot the original esp-link firmware won't work. Also, trying to boot Arduino Core made firmware with espressif's bootloader won't work either (I don't know if doing that even makes sense).

#### Hardware particularities

[...]

PR #3025, opened 2017-03-05
Added support for 40MHz crystal frequency

jeremypoulter:
This is needed for some ESP8266 modules that, for example the module in [this](http://www.banggood.com/Smart-Home-WiFi-Socket-Switch-UK-Plug-Wireless-Timer-Remote-Control-p-1022471.html?p=7U30132019453201507N) SmartPlug, that have a 40Mhz crystal rather than the 'normal' 26MHz.


torntrousers:
Hi Jeremy, that SmartPlug looks interesting but I can't find much on it, do you have any links to more about it - photos of inside or info about re-programing it etc?

jeremypoulter:
Some info on the module can be found here; https://detail.1688.com/offer/45107082796.html and there is some discussion on it in this forum; https://www.letscontrolit.com/forum/viewtopic.php?t=2159 however I am still working on the full documentation of my efforts with this plug.

gvJaime:
I've had issues #3004 and #3094 with the Arduino UNO WiFi, because of this little 40MHz crystal detail.

It would be great that this pull request gets merged because right now I have to modify the framework by hand to make it work. It would be easier with this simple flag.

jeremypoulter:
Somewhat frustratingly this change was added (not quite as cleanly IMHO) by #3121 so closing this

PR #3053, opened 2017-03-16
Digest Authentication in Webserver Library

Lan-Hekary:
Added and Example called HTTPAdvancedAuth to show the user how to use the auth method
the Normal Basic Auth wasn't affected as it's the default mode in the function call 
the user can change the auth relm too
also added a feature to change the auth failure html content


Lan-Hekary:
Hello @igrr , will this feature be added to the Master Branch ??
if it's added please close the Pull request :D

Lan-Hekary:
@igrr any news ??
I've been waiting for a long time now ..
tell me if there is are problem or modification ..

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -121,15 +127,97 @@ bool ESP8266WebServer::authenticate(const char * username, const char * password
...
+void ESP8266WebServer::requestAuthentication(HTTPAuthMethod mode, const char* realm, const String& AuthFailMsg){
+  if(realm==NULL){
+    _srealm = "Login Required";
+  }else{
+    _srealm = String(realm);
+  }
+  if(mode==BASIC_AUTH){
+    sendHeader("WWW-Authenticate", "Basic realm=\"" + _srealm + "\"");
+  }else{
+    _snonce=String(RANDOM_REG32,HEX)+String(RANDOM_REG32,HEX)+String(RANDOM_REG32,HEX)+String(RANDOM_REG32,HEX);
```
igrr: Does String(val, HEX) automatically pad the string to 8 hexadecimal places, or the string can end up being shorter?

igrr: Also, it might be a bit nicer to put this into a separate function, a la `String getRandomHexString(size_t length)`.

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -121,15 +127,97 @@ bool ESP8266WebServer::authenticate(const char * username, const char * password
...
     }
     authReq = String();
   }
   return false;
 }
 
-void ESP8266WebServer::requestAuthentication(){
-  sendHeader("WWW-Authenticate", "Basic realm=\"Login Required\"");
-  send(401);
+void ESP8266WebServer::requestAuthentication(HTTPAuthMethod mode, const char* realm, const String& AuthFailMsg){
```
igrr: suggest `authFailMsg` instead of `AuthFailMsg`.

libraries/ESP8266WebServer/examples/HttpAdvancedAuth/HttpAdvancedAuth.ino
```diff
@@ -0,0 +1,45 @@
+#include <ESP8266WiFi.h>
```
igrr: Please add a public domain / CC0 license at the top of the sketch

libraries/ESP8266WebServer/examples/HttpAdvancedAuth/HttpAdvancedAuth.ino
```diff
@@ -0,0 +1,45 @@
...
+#include <ESP8266WebServer.h>
+
+const char* ssid = "........";
+const char* password = "........";
+
+ESP8266WebServer server(80);
+
+const char* www_username = "admin";
+const char* www_password = "esp8266";
+const char* www_realm = "Custom Realm";             // allows you to set the realm of authentication Default:"Login Required"
```
igrr: These long comments will not fit into default Arduino IDE window.

libraries/ESP8266WebServer/examples/HttpAdvancedAuth/HttpAdvancedAuth.ino
```diff
@@ -0,0 +1,45 @@
...
+String AuthFailResponse = "Authentication Failed";    // the Content of the HTML response in case of Unautherized Access Default:empty
```
igrr: same here, variable names in Arduino usually start with lowercase letter

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
Lan-Hekary: String(val, HEX) in the original Arduino depends on itoa which does not support padding
But it's not a problem really
the nonce and opaque values are arbitrary in length anyway

but the idea of a function is much better as I need to call it twice ..
I will modify it right away ..

libraries/ESP8266WebServer/examples/HttpAdvancedAuth/HttpAdvancedAuth.ino
Lan-Hekary: I copied this sketch from the HttpBasicAuth folder and it does not have a license ..
I am not sure of what should I put here ..
can you give an Example ??

Lan-Hekary:
thanks for the fast reply @igrr :dagger:

Concerning this :
> One major request is to make sure new functionality does not get linked into the application if Basic auth is used. My suggestion is to add two functions, requestBasicAuthentication and requestDigestAuthentication, keep requestAuthentication for compatibility (making it call requestBasicAuthentication). This way the extra code only gets added if Digest auth is actually used in the sketch.

The function prototype in the header file is the following
`void requestAuthentication(HTTPAuthMethod mode = BASIC_AUTH, const char* realm = NULL, const String& AuthFailMsg = String("") );`

it defaults to use the Basic Auth if no parameters is passed to the function call
so the user won't feel the difference.

this code for example exactly works the same as if no modifications were added 
`if(!server.authenticate(www_username, www_password))
    return server.requestAuthentication();`

if you are worried about the extra code added .. the Bulk of the Code resides in the function `authenticate` and I have no Idea how to reduce it if the user prefers a Basic Auth ..
I guess it won't make a big difference in code size .. but it would simplify things a bit ..

please advise on how to proceed I will modify it as you say it if you say so :D 

I am working on the other modifications in the meanwhile waiting for a reply to push the last modification

libraries/ESP8266WebServer/examples/HttpAdvancedAuth/HttpAdvancedAuth.ino
igrr: E.g. https://github.com/esp8266/Arduino/blob/master/libraries/esp8266/examples/Blink/Blink.ino#L1-L4

igrr:
Ok, i understand. It's okay to keep the function as is (`requestAuthentication` with a parameter).

Lan-Hekary:
I've made the modifications you asked for
I wrote a Function to get random hex but I made it a fixed length and it's padded now
`String getRandomHexString();`
to avoid the hassle of dynamic allocation and to avoid any additional overhead

libraries/ESP8266WebServer/examples/HttpAdvancedAuth/HttpAdvancedAuth.ino
igrr: Sorry, i mean public domain license, as in Blink example. It has been mentioned before that using LGPL license on examples is too restrictive. We already have a few such examples, and it would be better to migrate all the examples to CC0/public domain license.

Lan-Hekary: sorry I just copied the license from the next sketch in the folder .
fixed it now

Lan-Hekary:
any news ?? @igrr 
I think you forgot me again ..

PR #3076, opened 2017-03-23
[FIX] MDNS queryService emtpy buffer before query service

kdsoo:
Fix previous data left to in buffer for queryService()
In case of queried service doesn't exist on the net, there was a problem of queryService returning previously queried results.

Signed-off-by: Dongsoo Kim <dongsoo.kim@gmail.com>


devyte:
Closing in favor of #5384 .

PR #3123, opened 2017-04-07
Added a disconnect before connect

terryjmyers:
as per section 3.5.13 in the "ESP8266 Non-OS SDK API Reference" (https://espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf) the function wifi_station_connect states:
"If the ESP8266 is already connected to a router, wifi_station_disconnect
must be called first, before calling wifi_station_connect. "
I've always had to call WiFi.Disconnect() first which calls wifi_station_disconnect(), but if we add a disconnect always before a standard connect, it will make programming on the fly network changes eaiser.


igrr:
Thanks for the PR!

Could you please clarify the issue which is being solved? I have always thought that calling `WiFi.disconnect()` before `WiFi.connect` with different credentials is intentional, because it makes the purpose of the code clear — `disconnect` from the current network and then connect to the new one.

pfeerick:
I'd suggest that this was proposed merely for convenience... i.e. being able to just call WiFi.connect() again without having to worry about the disconnect if already connected to a network. Handy, but if you really want that, I think a macro in that sketch would be better, as I agree with iggr, it would be better to have it explicit so the intent is clear in the code. Maybe we can let this one go?

earlephilhower:
This is a simple 1-liner and we should be able to agree or drop it with the next release.

If calling `wifi_station_disconnect()` when disconnected is legal (and required before `wifi_station_connect()`, I'm not seeing any issue with it.  If it is illegal to do so, this should be closed.

devyte:
I am not convinced about this. My own wifi setup sequence doesn't (sdk-) disconnect before connecting, and it works even when already connected. My suspicion is that the sdk already does this internally, likely implemented in a later release than when this PR was proposed.
Closing.

Issue #3075, opened 2017-03-23
servestatic (SPIFFS) method will not send the whole file (tested with 30kB file) 


maichaell:
#### Hardware
Hardware:			ESP-12E NodeMcu 1.0
Core Version:      	latest GIT
### Description

Problems is that when using **servestatic** method esp8266 server does not send all content and browser stays at 'Connecting....'

I have uploaded files :

index.html
jquery-3.2.0.min.js.gz

into flash using ESP8266FS tool from Arduino IDE.

It seems index.html is sent okay but when ESP8266 server tries to send jquery-3.2.0.min.js.gz file it only sends part of it and not everything. Browser will stay 'connecting'

I this known error ? Can I enable some debugging to see better what is going on ? 

### Settings in IDE

Module:  			NodeMCU 1.0
Flash Size: 		4MB/1MB
CPU Frequency:		80Mhz
Flash Mode:			qio
Flash Frequency:	40Mhz
Upload Using:		SERIAL
Reset Method:		nodemcu

### Sketch

```cpp

#include <Arduino.h>

void setup(void){
  pinMode(led, OUTPUT);
  digitalWrite(led, 1);
  Serial.begin(115200);
  SPIFFS.begin();
  
  Serial.println();
  
    Dir dir = SPIFFS.openDir("/");
    while (dir.next()) {
      String fileName = dir.fileName();
      size_t fileSize = dir.fileSize();
      Serial.printf("FS File: %s, size: %s\n", fileName.c_str(), String(fileSize).c_str());
   
   Serial.println();
}
  
  WiFi.mode(WIFI_AP);

  WiFi.softAP(ssid);
  delay(200);

  IPAddress myIP = WiFi.softAPIP();

  Serial.println("AP IP address: ");
  Serial.println(myIP.toString()); 
        
  server.on("/post", HTTP_POST, handleRootPOST);
  
  server.on("/inline", [](){
    server.send(200, "text/plain", "this works as well");
  });  
  
  server.serveStatic("/index.html", SPIFFS, "/index.html");
  server.serveStatic("/jquery-3.2.0.min.js.gz", SPIFFS, "/jquery-3.2.0.min.js.gz");  
  server.serveStatic("/", SPIFFS, "/index.html");

  server.begin();
  Serial.println("HTTP server started");
}

void loop() {
  server.handleClient();
}

maichaell:
Capture from WireShart using 'Follow TCP Stream'

[...]

PR #3129, opened 2017-04-10
prepare lwip2

d-a-v:
This PR brings minimum changes for libraries to compile with lwip2 (https://github.com/d-a-v/esp8266-phy).
#3075 seems to have been solved with it.

It is harmless otherwise.


d-a-v:
Someone explains me how is it that codecov was complaining the first time (about unrelated files),
and says nothing the second time (where I just synced with master, unrelated files too) :)

PR #3142, opened 2017-04-16
Put SSDP constants in progmem

sidoh:
Saves a few hundred bytes of RAM. Tested that discovery still worked from a Windows machine.


libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -240,7 +243,9 @@ void SSDPClass::_send(ssdp_method_t method){
 
 void SSDPClass::schema(WiFiClient client){
   uint32_t ip = WiFi.localIP();
-  client.printf(_ssdp_schema_template,
+  char buffer[strlen(_ssdp_schema_template)+1];
```
igrr: shouldn't this be `strlen_P`?

libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -201,9 +201,12 @@ void SSDPClass::_send(ssdp_method_t method){
   char buffer[1460];
   uint32_t ip = WiFi.localIP();
 
-  int len = snprintf(buffer, sizeof(buffer),
+  char valueBuffer[strlen(_ssdp_notify_template)+1];
```
igrr: likewise

sidoh:
Ah, yep. Fixed. Thanks for catching that :)

I'm pretty perplexed as to how this works without using `strlen_P`.

igrr:
My guess is that it used to work because the compiler can evaluate `strlen` at compile time, if the argument is a constant string.

Issue #3101, opened 2017-03-31
ESP8266mDNS:  Replies are not reliably sent on the correct network interface


ccooper21:
#### Hardware
Hardware: NodeMCU (ESP-12E)
Core Version: Latest (commit a01638f from Mar 13, 2017)

### Description

I spent the day experimenting with the [mDNS-SD_Extended](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266mDNS/examples/mDNS-SD_Extended/) example and the related [ESP8266mDNS](https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266mDNS) library.  I did not have any success getting my ESP to respond to mDNS queries for most the day.  In order to investigate, I used both `WireShark` on my laptop and `tcpdump` on my router to monitor the relevant network traffic.  There just weren't any mDNS/UDP packets getting sent in response to queries from my other devices.  Ultimately, I discovered that the responses were getting sent, but through the `SOFTAP_IF` interface instead of the `STATION_IF` interface that was connected to my local network.  As a workaround, disabling the `SOFTAP_IF` interface was a step in the right direction, but it was not sufficient.  I noticed that in the `MDNSResponder::advertiseServices(...)` method that all traffic gets sent to the `SOFTAP_IF` interface.  It appears that this is just a simple coding mistake.  These problems appear to have been caused as a result of commit a546d64.  Aside from the simple coding mistake that I mentioned, unfortunately, I'm not familiar enough with all of the code involved or mDNS to understand exactly what needs to be fixed to get mDNS working reliably with both interfaces enabled as was the intent of the referenced commit.

### Settings in IDE

Module: NodeMCU 1.0 (ESP-12E)
Flash Size: 4MB
CPU Frequency: 80Mhz
Flash Mode: Unknown
Flash Frequency: Unknown
Upload Using: Serial over USB
Reset Method: NodeMCU

### Sketch

See [mDNS-SD_Extended](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266mDNS/examples/mDNS-SD_Extended/) example

### Debug Messages

```
Entering MDNSResponder::queryService(...)...
queryService esp tcp
Sending mDNS query...
Entering send(...)...
Calling udp_sendto(,, 224.0.0.251, 5353)...
Payload:
000000000001000000000000045f657370045f746370056c6f63616c00000c0001
 . . . . . . . . . . . . . _ e s p . _ t c p . l o c a l . . . . .
udp_send
[...]

PR #3144, opened 2017-04-17
Mdns fixes

swarren:
Fixes for issues described in issue #3101 
- Services advertized on SOFTAP interface only.
- Query code missing multi-interface support.


igrr:
Thank you for the fixes, merged as part of #3170.

Issue #3140, opened 2017-04-15
pgm_read_byte() does not work well without delay()


milkpirate:
I have the following code:

```
PROGMEM const uint8_t table[2][3] = {
  {0x7B,0x60,0x5D},
  {0x7D,0x48,0x3E}
};

void setup() {
  Serial.begin(115200);
}

void loop() {
  uint8_t data = pgm_read_byte(&table[1][1]);
  Serial.println(data, HEX);
  delay(250);
}
```
which results in:
```
Exception (3):
epc1=0x40201c0e epc2=0x00000000 epc3=0x00000000 excvaddr=0x4022e71c depc=0x00000000

ctx: cont
sp: 3ffef200 end: 3ffef3c0 offset: 01a0

>>>stack>>>
3ffef3a0:  feefeffe feefeffe feefeffe 40202098
3ffef3b0:  feefeffe feefeffe 3ffee3a0 40100114
<<<stack<<<

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 1384, room 16
tail 8
chksum 0x2d
csum 0x2d
v09f0c112
```
But swapping `Serial.println(data, HEX);` and `delay(250);` then works. Why?

**EDIT:**
Shorter delays make it work as well even a "not existing" one:
```
PROGMEM const uint8_t table[2][3] = {
  {0x7B,0x60,0x5D},
  {0x7D,0x48,0x3E}
};

void setup() {
  Serial.begin(115200);
}

void loop() {
  uint8_t data = pgm_read_byte(&table[1][1]);
  delayMicroseconds(0);
  Serial.println(data, HEX);
  delay(250);
}
```

igrr:
Could you please upload the elf file for the code which demonstrates the crash? That would really help.

milkpirate:
Elf file (1.8.0): http://www21.zippyshare.com/v/Vhm4iyHc/file.html (sorry could not find the option to attach files).

milkpirate:
Elf file (1.8.0): http://www21.zippyshare.com/v/Vhm4iyHc/file.html (sorry could not find the option to attach files).

igrr:
[...]

PR #3164, opened 2017-04-22
Fix Progmem read over optimization

Makuna:
The compiler is overoptimizing a 32bit aligned read to the requested
size (8 or 16 bit) but due to the flash requirement that reads must
always be 32bits in size this will cause an exception when the code is
run.
This change hints to the compiler that it can't optimize the read size.


Makuna:
https://github.com/esp8266/Arduino/issues/3140

igrr:
Decided to go for inline assembly approach, but thanks for the PR and looking into this @Makuna!

PR #3176, opened 2017-04-28
HttpClient: Adding SSL Certification Verification

salqadri:
This PR adds a new API called "setRootCA" in which the user can pass in the root CA with which to validate an SSL connection.

WiFiClientSecure enables users to add multiple certificates for validation. But the underlying axTLS library actually only cares for the deepest certificate for verification. As such, I decided to have a much simpler API on HttpClient of setting a single certificate to prevent users from getting confused on how many certs they need to add. I call it Root to indicate that this must be the deepest certificate.

Unfortunately, the existing implementation of HttpClient was not conducive to what I wanted to do. I wanted the following API experience:
`
  http.begin("https://www.google.com", 443);
  http.setRootCA(Equifax_cer, Equifax_cer_len);
  int result = http.GET();
`
But the problem was that both the connect call and the actual http request were both done in the path for http.GET. axTLS Certificate validation _requires_ a TCP connection to first exist, and THEN you can load a certificate (I spent several hours figuring this and other issues like the fact that you HAVE to initialize sntp before validation can happen out by digging deep into the axTLS codebase).

So given this issue, I tried to solve this problem without causing API changes. The result was that I moved the connect phase into the begin command, and added a new state variable called _verified to determine if verification has already been done to avoid doing it again on subsequent connections.

I've also put in an example ino test file that anyone should be able to run to see how this works. One key thing to note is that you HAVE to initialize sntp before attempting validation because the certificate is validated with respect to time.


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -178,12 +212,31 @@ bool HTTPClient::beginInternal(String url, const char* expectedProtocol)
 
 bool HTTPClient::begin(String host, uint16_t port, String uri)
 {
+    DEBUG_HTTPCLIENT("[HTTP-Client][begin] host: %s port: %d uri: %s\n", host.c_str(), port, uri.c_str());
     clear();
+
+    // Remove protocol if any
```
igrr: This method takes `host` as an argument, i.e. just the DNS name or IP address. It should not expect to see a protocol prefix.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -304,6 +364,32 @@ void HTTPClient::setTimeout(uint16_t timeout)
     }
 }
 
+bool HTTPClient::setRootCA(uint8_t * cert, size_t size)
+{
+    bool result = false;
+    if(_tcp && (_protocol == "https" || _port == 443)) {
+        auto client = static_cast<WiFiClientSecure*>(_tcp.get());
```
igrr: If you follow the existing `TransportTraits` pattern, you can create a new TransportTraits class which performs root certificate verification instead of fingerprint verification. In `setRootCA` method, you can set `_transportTraits` to an instance of, say, `new TLSRootCertTraits(cert, size)`, which will then perform verification using the existing interface.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -121,11 +141,25 @@ bool HTTPClient::begin(String url, String httpsFingerprint)
...
+        _port = 443;
+        _transportTraits = TransportTraitsPtr(new TLSTraits());
+    }
+    else {
+        _port = 80;
+        _transportTraits = TransportTraitsPtr(new TransportTraits());
+    }
+    _tcp = _transportTraits->create();
+
+    if(!_tcp->connect(_host.c_str(), _port)) {
```
igrr: This is a fairly big behavioural change, and it's not hard to come up with a bunch of cases when it will break existing sketches. If you need to defer setting the CA until the time connection is established, save the CA information in the appropriate Traits structure (probably a new one, dedicated to verification using a root CA cert). Then use this information when the verification has to take place. If HTTPClient needs to interact with Traits class at some other point in the process, add a new interface method and implement it as needed. It doesn't look like it is worth changing the behaviour of user-facing class, when the problem can be solved inside the implementation.

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
```diff
@@ -0,0 +1,172 @@
+/**
+ * SSLCertificateVerification.ino
+ *
+ * Created on: 04.28.2017
+ *
+ * Copyright (c) 2017 Syed Salman Qadri. All rights reserved.
+ * This file is part of the ESP8266HTTPClient for Arduino.
+ *
+ * This library is free software; you can redistribute it and/or
```
igrr: examples are normally licensed as public domain or CC0

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
```diff
@@ -0,0 +1,172 @@
...
+#include <ESP8266HTTPClient.h>
+#include <time.h>
+
+//WiFi
+const char wifissid[] = "WiFi";
+const char wifipsk[] = "Password";
+
+// This is a binary dump of the root CA.
+// You can also choose to load the cert via a file stream.
+// Here are the steps I did to dump the CA on my Mac:
```
igrr: Suggest writing this in a slightly more impersonal way :)

igrr: Also, what steps would users on other platforms take? Windows is probably the biggest question. Instructions for Linux will likely be the same, except for Mac-specific names ("Keychain Access", "Terminal").

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
```diff
@@ -0,0 +1,172 @@
...
+//   1) openssl s_client -connect google.com:443
+//        The output has a "Certificate chain" section. Your root CA is the bottom line.
+//        In the case of google.com, this line is
+//          "i:/C=US/O=Equifax/OU=Equifax Secure Certificate Authority"
+//        This means I need to validate with the Equifax Secure CA
+//   2) Open Keychain Access. In the Search bar type "Equifax" and press Enter.
+//   3) Right-click the cert and click Export and save it as a .cer.
+//   4) In the terminal, type xxd -i <path_to_cer_file>
+unsigned char Equifax_cer[] = {
```
igrr: Can this be put in PROGMEM? We also need to support setRootCA_P in this case.

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
```diff
@@ -0,0 +1,172 @@
...
+  0x50, 0x95, 0xd1, 0x9a, 0x6f, 0xe4, 0x11, 0xde, 0x63, 0x85, 0x6e, 0x98,
+  0xee, 0xa8, 0xff, 0x5a, 0xc8, 0xd3, 0x55, 0xb2, 0x66, 0x71, 0x57, 0xde,
+  0xc0, 0x21, 0xeb, 0x3d, 0x2a, 0xa7, 0x23, 0x49, 0x01, 0x04, 0x86, 0x42,
+  0x7b, 0xfc, 0xee, 0x7f, 0xa2, 0x16, 0x52, 0xb5, 0x67, 0x67, 0xd3, 0x40,
+  0xdb, 0x3b, 0x26, 0x58, 0xb2, 0x28, 0x77, 0x3d, 0xae, 0x14, 0x77, 0x61,
+  0xd6, 0xfa, 0x2a, 0x66, 0x27, 0xa0, 0x0d, 0xfa, 0xa7, 0x73, 0x5c, 0xea,
+  0x70, 0xf1, 0x94, 0x21, 0x65, 0x44, 0x5f, 0xfa, 0xfc, 0xef, 0x29, 0x68,
+  0xa9, 0xa2, 0x87, 0x79, 0xef, 0x79, 0xef, 0x4f, 0xac, 0x07, 0x77, 0x38
+};
+unsigned int Equifax_cer_len = 804;
```
igrr: suggest `sizeof(Equifax_cer) / sizeof(Equifax_cer[0])` here.

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
Defozo: The commands works on Windows too.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
salqadri: No saving the CA is a bad idea as it would consume a lot of RAM. I've been trying to explicitly avoid storing the CA.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
salqadri: The approach you suggest would be more expensive as it requires storing the cert, which can be quite large.

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
salqadri: Good to know. I can update that it works on Mac and Windows.

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
salqadri: These two lines are generated by the previous command, so I'm trying to show that you can just copy-paste verbatim.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
salqadri: Makes the function more resilient for me. It is also more consistent with the other overload, making it more easy to swap out the API a developer may want to use.

salqadri:
@igrr  I hadn't noticed that you guys provided feedback on my PR until now :). Thanks! I'll happily make the various fixes and redo it, but only if we don't make it so that we have to store the cert in the traits. That approach consumes too much memory, and I'd rather give the developer more control on how they wish the cert to be stored. Whether the developer is storing it in a file system, or in memory, either way the traits approach leads to duplication which causes crashes.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
igrr: OK, that is a fair observation.
Note that if the certificate is loaded from file, we need to have a copy of it in RAM at some point anyway to pass this copy into axTLS. Same for PROGMEM. So in these two cases, there is no optimization possible (which wouldn't involve changing axTLS to be capable of reading from streams).
In the case when the certificate is already in RAM (such as in your example) we can store the pointer to the certificate and pass it on to axTLS without making a copy. This however is likely the least useful option, because certificate places into .rodata will take up precious RAM during the entire application life. On the other hand, the certificate in FS or PROGMEM will be in RAM only when it is actually needed there — during the TLS connection phase.

libraries/ESP8266HTTPClient/examples/SSLCertificateVerification/SSLCertificateVerification.ino
salqadri: Great idea. Maybe we can do that in a subsequent PR.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
salqadri: Agreed. What I'm trying to do is make it a temporary cost; you only need it in RAM during the verification phase, and then you can unload it from RAM. Something like this:
`
    SPIFFS.begin();
    File ca = SPIFFS.open(F("/Equifax.cer"), "r");
    if (ca) {
      if (!firestream->setRootCA(ca, ca.size())) {
        Serial.println(F("Root CA failed to load"));
      }
      ca.close();
    }
    SPIFFS.end();
`

salqadri:
@igrr My branch seems to be broken after I merged from master. SSL cert verification is no longer working for me. Sometime after c5c138ec5abd734dad44bdb3a69a20c2eda98211 and between d6f1f0d5d77508ed2b3bb6051a6fba3d69ab98ca something you did has caused my fork to now fail. It seems that there were a lot of changes to various SSL aspects, but I'm having a hard time understanding what you were trying to do. Could you possibly explain the changes to things like WiFiClientSecure etc. so that I can figure out what broken?

salqadri:
I have now fixed this branch by getting everything from master except that axtls library update. So I'm using the version of axtls from prior to the May 9 2017 update. Would love it if @igrr you can fix axtls so that I don't have to do this workaround.

Defozo:
When it'll be merged?

cimba007:
Any update so far?

cimba007:
@salqadri: Any chance to fix this?

earlephilhower:
I believe this PR is superseded by #4980 which lets you stick in a custom configured `WiFiClient*` to `HTTPClient`.

Issue #1062, opened 2015-11-20
Support for SPI0 in SPI library


Shibonja:
Hi guys,

First of all Thanks for the good work, you made development much easier with esp integration with Arduino IDE. 
I have one question if someone can answer on them. 
1) are the extra SPI pins on esp-12e supported in Arduino IDE and if yes are they using bit-banging.

ESP8226-12E https://www.adafruit.com/products/2491

Thanks in advance, Haris.
##

igrr:
SPI library uses hardware HSPI (a.k.a. SPI1) peripheral (pins 12-15).
Extra pins on ESP-12E are connected to SPI0 peripheral, which is not supported by SPI0 library at the moment.

Shibonja:
Hi @igrr ,
Thanks for quick answer.
Are you planning to add support for them in future releases?

igrr:
Actually I did implement this in a hacky manner about a month ago, need to revisit these changes and clean up the code.

Shibonja:
Hi @igrr ,
I ordered a few ESP8226-12E, so if u need any help, maybe I can be at your service. I am not a C/C++ guru but I can find my way around.

Links2004:
is SPI0 not shared with the flash?
so all the SPI code need to be in ram,
may gives problems with other interrupts too?

igrr:
Actually this is done as follows: you are still using SPI1 peripheral (i.e. registers are the same), but its pins are mapped to the same pins as SPI0. There is a hardware arbiter which can demultiplex two SPI peripherals to the same physical set of pins, controlling two CS lines automatically. SPI0 has higher priority, so flash requests are still serviced, and you can use cached code. Access times via SPI1 become a bit unpredictable in this mode, but you can have an SPI peripheral and keep pins 12-15 free at the same time.

reaper7:
@igrr - very interesting...is there any chance for tests this solution?

steffenmauch:
@igrr Is there any possiblity to test your solution?

KaloNK:
+1 for this feature.

Will it be possible to switch SPI1 between hardware and pins 12-15 at runtime?

I was looking for something similar to https://github.com/espressif/esp8266_mp3_decoder implementation for SPI SRAM (or [FRAM](https://github.com/adafruit/Adafruit_FRAM_SPI) ) chip access.
[...]

PR #3189, opened 2017-05-02
Implement support for HSPI overlap mode

riataman:
Hey this implement Issue #1062. Allows activation of HSPI overlap mode. I tested this with a 23LC1024 memory and it worked fine. But couldn't make it work with an SD card. The SD card was getting confused by the sharing of the SPI bus.

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -22,6 +22,12 @@
 #include "SPI.h"
 #include "HardwareSerial.h"
 
+#define SPI_PINS_HSPI			0 // Normal HSPI mode (MISO = GPIO12, MOSI = GPIO13, SCLK = GPIO14);
+#define SPI_PINS_HSPI_OVERLAP	1 // HSPI Overllaped in spi0 pins (MISO = SD0, MOSI = SDD1, SCLK = CLK);
+
+#define SPI_OVERLAP_SS 0
+
+
 typedef union {
         uint32_t regValue;
         struct {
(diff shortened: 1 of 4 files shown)


libraries/SPI/SPI.cpp
```diff
@@ -38,37 +38,88 @@ SPIClass::SPIClass() {
 }
 
 void SPIClass::begin() {
-    pinMode(SCK, SPECIAL);  ///< GPIO14
-    pinMode(MISO, SPECIAL); ///< GPIO12
-    pinMode(MOSI, SPECIAL); ///< GPIO13
+	begin(SPI_PINS_HSPI);
```
igrr: The file used to be formatted with 4 spaces, could you please adjust indentation in your changes?

libraries/SPI/SPI.h
```diff
@@ -54,6 +60,7 @@ class SPIClass {
 public:
   SPIClass();
   void begin();
+  void begin(uint8_t pinSet, uint8_t miso = 0, uint8_t mosi = 0, uint8_t sck = 0, uint8_t ss = 0);
```
igrr: I would rather suggest keeping just miso/miso/sck/ss arguments. You can determine _pinSet value internally.

Another point, do Arduino libraries typically call SPI.begin internally? If so, maybe adding an independent .pins method would be a more portable choice. That way, one could first call SPI.pins in the sketch and then call .begin method of another library which would do SPI.begin internally.

igrr: If going with begin method, suggest at least making it take the same arguments as it does in the ESP32 library:
https://github.com/espressif/arduino-esp32/blob/master/libraries/SPI/src/SPI.h

riataman:
I think I submitted the updated changes (I suck at git).

I like your idea of adding an interface to setup the SPI before the call to SPI.being, as you say most libraries just call it internally.

I'm kind of wary of this SPI::pins interface as, in absence of SoftwareSPI there's just two magical set of pins that work. And even if we implement SoftSPI all three modes are different enough that I'd rather the caller set explicitly where they want normal HSPI, overlapped HSPI or SoftwareSPI.

There are a lot of caveats in overlap mode that I'm  in doubt that any library designed for Arduino(avr) is going to work out of the box. First and foremost you can't use any random GPIO for the SS line in overlap mode, you *MUST* use GPI0 as Hardware CS, otherwise you would likely be asserting your GPIO CC when the flash is using the pins.

Now since you are using hardware CS it will deassert your device after every write/transfer. Since most arduino libraries are only doing single byte read/transfers the devices will be constantly asserted/deasserted, but at least the 23LC1024 I used to test this expects CS to be LOW for the entire transfer of a whole address/write so you need to convert existing libraries to use SPI.writeBytes/SPI.transferBytes otherwise they will not work.

Let me know what you think.

riataman:
Also I didn't updated the documentation my in latest changes. I would do that once we agree on an interface.

igrr:
Okay, this version with `pins` method looks good to me. Please re-indent again, some lines still have tabs.

riataman:
Done, I removed the missing tabs, updated the doc and squashed all the commits back into one.

PR #3191, opened 2017-05-02
Free memory we allocated and actually stop i2s.

riataman:
i2s_end was leaking memory and wasn't actually stopping i2s.

### DIFF
--- a/cores/esp8266/core_esp8266_i2s.c
+++ b/cores/esp8266/core_esp8266_i2s.c
@@ -146,6 +146,10 @@ void ICACHE_FLASH_ATTR i2s_slc_end(){
   SLCIE = 0;
   SLCTXL &= ~(SLCTXLAM << SLCTXLA); // clear TX descriptor address
   SLCRXL &= ~(SLCRXLAM << SLCRXLA); // clear RX descriptor address
+
+  for (int x = 0; x<SLC_BUF_CNT; x++) {
+	  free(i2s_slc_buf_pntr[x]);
+  }
 }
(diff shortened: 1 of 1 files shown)


riataman:
I put in this patch, but at some point the i2s support should probably change a lot. There are actually for modes that can be made to work:

i2s output master mode (this is what the current code supports only).
i2s output slave mode.
i2s input master mode.
i2s input slave mode (I have this working on my machine with a teensy audio board (SGTL500)).

- I2s input and output can be working at the same time.
- You can set the clock rate in master mode, but it doesn't make sense in slave mode.
- It would be very nice to be able to set the size of the DNA buffers and the amount of buffers depending on your usage, for example if you are reading/saving to SD cards 512 bytes makes sense as the buffer size. But if you are pushing to the network you may want bigger packages (about MTU size I guess would maximize throughput).

The fact that ESP8266 can't output a master clock that is a multiple of the sample rate is unfortunate, that's why I had to put the SGTL5000 in master mode (ESP9266 in slave mode) so that it can generate its own clocks (I'm feeding the 26Mhz XTAL_CLOCK from pin0 to it just to make it go).

I haven't looked into much detail into the ESP32 API, but if it already supports all of this, it would be something nice to emulate.

Anyway, I should probably open an issue with all of this, instead of piggy backing this into this simple fix.

igrr:
Thanks! Suggest opening a new issue for discussion of further I2S enhancements. CC @me-no-dev there regarding the ESP32 library.

PR #3206, opened 2017-05-06
open source lwip2

d-a-v:
PLEASE check #3362 out instead

----------------------------------------------

* builder downloads original sources
* precompiled version included
* optionally selectable in arduino IDE menus
* legacy lwip selected by default


igrr:
Thanks for the PR! I haven't yet had time to review and test it, however I have one question: why do we want to keep old lwIP version around? I.e. why did you choose to add it side by side with the old one instead of replacing?

d-a-v:
My humble opinion states that it is early for total and definitive lwip replacement :)
Further testing is needed because so many things rely on lwip (I alone did not cover all libraries using udp/tcp/...).

The current PR sets current-lwip-v1.4 as the default version. What could be done is to set v2 as the default version, leaving v1.4 still available just in case, at least for a little while.

d-a-v:
Also, for lwip-v2 to compile, a great deal of lwip-v1.4-include-files.h are needed (because of blobary-structures bindings).

d-a-v:
now works with #3001 (sntp was missing)

devyte:
hi, what's the status of this?

d-a-v:
Hi,

@igrr (and I) prefer [waiting](https://github.com/esp8266/Arduino/pull/3001#issuecomment-309630622) for release 2.4.

I would like to mention here that I made a new pull request (#3362) for open source lwIP. Its code is updated but above all the integration is different: it uses submodule. Discussion is needed before I close here.

d-a-v:
closing in favour of #3783

Issue #2915, opened 2017-02-02
SD Web Server example fails with latest master code


torntrousers:
As also mentioned in the Issues 1853 comment [here](https://github.com/esp8266/Arduino/issues/1853#issuecomment-248483979) SD Web Server example fails to serve the complete pins.png image with Serial console message "sent less than expected!". Looks like anything larger than 32k gets truncated. Works ok going back to the 2.3.0 release.

igrr:
Should be fixed by #3209.

PR #3209, opened 2017-05-06
Remove SD file available size saturation

korzo89:
A problem with file streaming arose during some experiments with the SDWebServer example. It seems that the current version of the software has a problem with sending large files stored on the SD card to connected clients.

After trying to find the cause of this problem, I have found that large files are always streamed up to the first 32767 bytes, regardless of the value returned by the size() method of the File class. This can be easily seen in browser network diagnostic tools:

The issue is caused by the SD library's File class available() method, which causes its result to be saturated to 0x7FFF (32767), making the file streaming function behave incorrectly. After removing the value clamping code, the streaming works perfectly for files with sizes larger than 32kB:

I have also tested the fix with very large files (~40MB) and can confirm that it works (albeit quite slowly).

I suspect that the value saturation code was borrowed from the original Arduino SD libraries, which usually use 8-bit AVR compilers, where the int type (returned by the available() method) is a 16-bit integer - hence the clamping.

### DIFF
--- a/libraries/SD/src/File.cpp
+++ b/libraries/SD/src/File.cpp
@@ -98,9 +98,7 @@ int File::read(void *buf, uint16_t nbyte) {
 int File::available() {
   if (! _file) return 0;
 
-  uint32_t n = size() - position();
-
-  return n > 0X7FFF ? 0X7FFF : n;
+  return size() - position();
 }
 
 void File::flush() {


igrr:
Based on the description it looks like this fixes https://github.com/esp8266/Arduino/issues/2915.

PR #3214, opened 2017-05-08
ClientContext: don’t read DataSource in TCP callback

### LINKED ISSUE #2399: Yield context issue
I am using Git version and ESP12F CPU 160Mz 4MB/1MB SPIFFS and SD card library https://github.com/greiman/SdFat/tree/master/SdFat

when using :

```
File datafile = SD.open((char *)path.c_str());
                  if (!datafile){
                       LOG("Cannot open file!\n");
                       web_interface->WebServer.sendContent_P(NOT_AUTH_NF);
                       page_not_found = true;
                    }
                if (web_interface->WebServer.streamFile(datafile, contentType) != datafile.size()) {
                    LOG("Sent less data than expected!\n");
                    }
                else LOG("Stream done\n");
```

I have a crash when downloading files bigger than few Kb but no issue with small files 
I have opened an issue on sdfat https://github.com/greiman/SdFat/issues/39 and conclusion it is yield issue on esp core
Discussion with @me-no-dev and @martinayotte  on gitter start here : https://gitter.im/esp8266/Arduino?at=57b1da0323f0cacc34e7e119
and end here: https://gitter.im/esp8266/Arduino?at=57b1e7dabe8025f16943d246

igrr:
Previously, `_write_some` function would be called each time TCP stack
notifies the application that some data was delivered (via the `sent`
callback). In turn, `_write_some` would obtain more data to be sent from
the DataSource. In case of a DataSource backed by a Stream, this would
read from a stream. Some libraries (such as SD) may call `yield` and
other blocking operations from Stream read function, which can not be
used in TCP stack callbacks.

This change moves the data sending loop back into the Arduino task, with
a negligible loss of performance. TCP callback now wakes the main task
via `esp_schedule`, which performs stream read and provides more data
to the TCP stack.

Possible future optimization would be to buffer Stream data ahead of
time. That way, buffered data could be sent immediately from the TCP
callback. On the other hand, this optimization would need extra TCP_MSS
of temporary storage, per connection.

Fixes #2399.

### DIFF
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -317,9 +317,7 @@ protected:
 
     void _cancel_write()
     {
-        if (_datasource) {
-            delete _datasource;
-            _datasource = nullptr;
+        if (_send_waiting) {
             esp_schedule();
         }
     }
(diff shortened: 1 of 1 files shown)


igrr:
@luc-github I know it's been a long time since you have reported #2399, but if you can verify that this change fixes the issue, that would be great!

luc-github:
Hi sure I will check and report back soon -  thank you

luc-github:
@igrr I have deeply tested and so far so good : I have reverting the workaround in SDFat and no crash with big files (14Mb), I have did also big stream and deal with long run tests and I am not able to reproduce the issue anymore

Thank you ^_^

PR #3223, opened 2017-05-09
WIP: Add a SHA256Builder

comino:
A secure hashing algorithm is a crucial part of a secure signature. 
At the moment only MD5 is part of the core, thus I propose to add the SHA256Builder. 

Can then be implemented with:  https://github.com/esp8266/Arduino/pull/3105/files

ToDo
- [x] Implement a "HashBuilder" base class
- [x] Implement a seperate unit test for HashBuilder
- [ ] Check performance impact of new implementation
- [ ] (opt) Implement RSA/AES builder


igrr:
As discussed in #3105, i think the way forward is to add axTLS source as a submodule (or a subtree) to the Arduino core. This way we can use hashing, HMAC, x509 function without having to duplicate them.

igrr:
The other note about this: it seems to me that a lot of code in MD5Builder and SHA256Builder is very similar, with difference coming only in actual hashing function calls, hash context structures, and hash sizes. I do like the addition of SHA256Builder, but please consider folding implementation of these two hashes into one, both to reduce code size overhead (if both are used) and to reduce source code duplication.

comino:
Thanks for that constructive feedback! 

To not introduce breaking changes I can do a "hashLib" baseclass and inherit it in MD5Builder and SHA256Builder? Or drop this and think of a serious axTLS submodule?

igrr:
I think if you make a base class for hash builders and temporary introduce (duplicate) SHA256 code this will be okay. Once someone figures out how to add axTLS as a submodule, this SHA256 code can be removed.

comino:
The Crypto Submodule you mentioned - is there a concrete plan somewhere for this? 
I have my own implementation for a "CryptoBuilder" very similar to my proposed HashBuilder. 
If wanted I could implement it into the core (if wanted) like: 

/core
/core/crypto/md5.h
/core/crypto/sha256.h
/core/crypto/aes.h
/core/crypto/rsa.h
... etc

/core/CryptoBuilder.h  (like the current HashBuilder)
/core/RSABuilder.h /*.cpp
/core/MD5Builder.h /*.cpp
/core/SHA256Builder.h /*.cpp
/core/AESBuilder.h /*.cpp

The drawBack is, that the core will be filled with more classes.

Humancell:
Does this include HMAC SHA256?

comino:
Not yet, but I have a dirty implementation for signature checks based with SHA256 if we could agree on a submodule structure.

cores/esp8266/HashBuilder.cpp
```diff
@@ -0,0 +1,94 @@
...
+    return true;
+}
+
+// RESULTS
+String HashBuilder::_byteVecToString( std::vector<uint8_t>& res){
+  String resString;
+  resString.reserve(res.size()*2);
+  static constexpr char hex[] = "0123456789abcdef";
+  for (uint8_t & c : res){
+      resString += String(hex[c / 16]);
```
igrr: I guess I can't be upset with the indentation given that we don't have specific rules in this project (which is a shame)... but would be great if indentation was at least consistent at file scope.

cores/esp8266/HashBuilder.cpp
```diff
@@ -0,0 +1,94 @@
...
+    auto res = _result();
+    for(uint8_t i = 0; i < res.size(); i++){
+        output[i] =  res[i];
+    }
+}
+
+void HashBuilder::getChars(char * output){
+  auto res = _result();
+  for(uint8_t i = 0; i < res.size(); i++){
+    sprintf(output + (i * 2), "%02x", res[i]);
```
igrr: This does nearly the same thing as _byteVecToString but uses sprintf instead of the `hex` array. Would it make sense to use same method here?

cores/esp8266/HashBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+    void add(uint8_t * data, uint16_t len);
+    void calculate(void);
+
+    void addHexString(const char * data);
+    bool addStream(Stream & stream, const size_t maxLen);
+    String toString(void);
+
+    // deprecated
+    void getBytes(uint8_t * output );
+    void getChars(char * output);
```
igrr: suggest expressing deprecation with `__attribute__((deprecated))` Why are these deprecated, by the way?

cores/esp8266/HashBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+    // INLINE
+    void add(char * data){ add((const char*)data); }
+    void add(String data){ add(data.c_str()); }
+    void addHexString(char * data){ addHexString((const char*)data); }
+    void addHexString(String data){ addHexString(data.c_str()); }
+    void add(const char * data){ add((uint8_t*)data, strlen(data)); }
+
+  protected:
+    virtual void _init(void) = 0;
+    virtual void _update(uint8_t * data, uint16_t len) = 0;
```
igrr: I know this is carried over from the original code, but since you are improving things, would you mind replacing this uint16_t with size_t? It doesn't serve any purpose and only adds extra `extui` instructions to truncate 32-bit int to a 16-bit one.

cores/esp8266/HashBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+    virtual void _final() = 0;
+    virtual std::vector<uint8_t> &_result() = 0;
+
+    uint8_t _hexCharToByte(uint8_t c);
+    String _byteVecToString( std::vector<uint8_t>& v);
+};
+
+// might be renamed to cryptoBuilder, cause same baseclass
+// might be reused in CrpytoSubmodule
+typedef HashBuilder CryptoBuilder;
```
igrr: I don't think that the current HashBuilder interface is a good fit for AES/RSA. There may be common operations such as conversions from/to byte/hex arrays, but this doesn't mean that these features should share a common interface, given how different the use cases are. If we happen to need the byte/hex transforms for crypto classes, my suggestion is to move these transforms into helper functions/classes shared between HashBuilder and the future crypto classes.

cores/esp8266/MD5Builder.cpp
```diff
@@ -1,101 +1,19 @@
 #include <Arduino.h>
```
igrr: Can't comment on them directly, but there is a bunch of binary files with .gch extensions added in this PR.

cores/esp8266/MD5Builder.cpp
```diff
@@ -1,101 +1,19 @@
...
-{
-    memset(_buf, 0x00, 16);
-    MD5Init(&_ctx);
-}
-
-void MD5Builder::add(uint8_t * data, uint16_t len)
-{
+void MD5Builder::_update(uint8_t * data, uint16_t len){
     MD5Update(&_ctx, data, len);
+    return;
```
igrr: no need for explicit return here

cores/esp8266/MD5Builder.cpp
```diff
@@ -1,101 +1,19 @@
...
 
-uint8_t hex_char_to_byte(uint8_t c)
-{
-    return  (c >= 'a' && c <= 'f') ? (c - ((uint8_t)'a' - 0xa)) :
-            (c >= 'A' && c <= 'F') ? (c - ((uint8_t)'A' - 0xA)) :
-            (c >= '0' &&  c<= '9') ? (c - (uint8_t)'0') : 0;
+void MD5Builder::_init(void){
+  _resultVector = {};
+  MD5Init(&_ctx);
+  return;
```
igrr: likewise

cores/esp8266/MD5Builder.cpp
```diff
@@ -1,101 +1,19 @@
...
-
-String MD5Builder::toString(void)
-{
-    char out[33];
-    getChars(out);
-    return String(out);
+void MD5Builder::_final(void){
+    _resultVector.resize(HEX_STR_LEN);
+    MD5Final(_resultVector.data(), &_ctx);
+    return;
```
igrr: likewise

cores/esp8266/SHA256Builder.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+void SHA256Builder::_update(uint8_t * data, uint16_t len){
+    SHA256_Update(&_ctx, data, len);
+    return;
+}
+
+void SHA256Builder::_final(void){
+    _resultVector.resize(HEX_STR_LEN);
+    SHA256_Final(_resultVector.data(), &_ctx);
+    return;
```
igrr: returns, indentation

cores/esp8266/SHA256Builder.h
```diff
@@ -0,0 +1,43 @@
+/*
+  MD5Builder.h - exposed md5 ROM functions for esp8266
```
igrr: the ever present comment line :)

igrr:
Regarding importing axTLS code — let me give it a try...

comino:
Thanks a lot  @igrr for taking the time for constructive feedback! 
Will digest them and polish asap!

devyte:
CC @earlephilhower 
How does this fit in with #4273 ? I'd love to have a standalone crypto module with all the crypto algorithms usable from anywhere, and completely separate from axtls or bearssl.

earlephilhower:
Without a hardware module to accelerate things, either axtls or bearssl would be doable here.  But so would any standard C/Arduino library, and it may end up importing less code than either one.  There are hundreds of implementations out there, most optimized for 32 or 64 bit processors, so if this is really important I'd suggest looking there and not using either SSL lib as a base.

The way BearSSL is implemented the hash functions are written as pseudo-C++ classes in straight C.  It's got MD5, SHA 1, 224, 256, 384, and 512 implemented IIRC.  It's not rocket science pulling them out and doing a wrapper, but I have not done any benchmarking to compare their performance.

earlephilhower:
Given this is a WIP with 2 years w/o update, intended for use on signed updates (which are already in there using something similar but using BearSSL), I'm closing.  If someone wants to do an abstract wrapper for hashes around the BSSL implementation (honestly, they're simple enough to use directly w/o a wrapper), a new PR would be the best way forward.

PR #2701, opened 2016-11-21
HTTP update server can handle flash and spiffs files (spiffs size lim…

PA4WD:
HTTP update server can handle flash and spiffs files (spiffs size limited by free space)

PR #3234, opened 2017-05-11
Allow SPIFFS update via ESP8266HTTPUpdateServer

sandeen:
This adds capability to update the SPIFFS image via
the same mechanism as firmware in the 
ESP8266HTTPUpdateServer.

It does not provide any dependency or linkage between
firmware and spiffs image updating; they are each taken
on their own, each followed by a reboot.

(I wrote this before seeing the other PR #2701 for similar
functionality; I like this a bit better, because it uses
the available SPIFFS size, and does not hide magic numbers
(U_SPIFFS) in the html...)

(It also cleans up a stray \n from commit ace0622)

We'd really like to see this functionality in the library for the
[OpenEVSE wifi](https://github.com/OpenEVSE/ESP8266_WiFi_v2.x) project

This is my first try at this, mostly cribbing from the ESP8266httpUpdate version of the updater - happy to fix up anything as needed.

Thanks,
-Eric


libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
```diff
@@ -71,9 +80,16 @@ void ESP8266HTTPUpdateServer::setup(ESP8266WebServer *server, const char * path,
         WiFiUDP::stopAll();
         if (_serial_output)
           Serial.printf("Update: %s\n", upload.filename.c_str());
-        uint32_t maxSketchSpace = (ESP.getFreeSketchSpace() - 0x1000) & 0xFFFFF000;
-        if(!Update.begin(maxSketchSpace)){//start with max available size
-          _setUpdaterError();
+        if (upload.name == "spiffs") {
+          size_t spiffsSize = ((size_t) &_SPIFFS_end - (size_t) &_SPIFFS_start);
+          if (!Update.begin(spiffsSize, U_SPIFFS)){//start with max available size
```
sandeen: Oh, interesting.  The README does say _Use this method before updating SPIFFS using OTA._
I didn't see anything like that in ./libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp either - should it get added to both?

igrr: I think `SPIFFS.end` was only added recently, precisely for this use case. The libraries haven't been updated to take advantage of it, however. So updating the behavior of OTA libraries would indeed be welcome.

devyte:
Any progress with this? This is a very nice feature.

caverna:
@sandeen great piece of software!!!
I was facing this issue: I have a ESP-01 (black), but I was limited to write only 512KB and 64KB of SPIFF.
But now I can unleash the whole 1MB!
Thank you a lot!
Do you need any help to prepare/adjust this PR to be able to merge to core?

libraries/ESP8266HTTPUpdateServer/src/ESP8266HTTPUpdateServer.cpp
```diff
@@ -6,13 +6,22 @@
...
+  R"(<html><body>
+     <form method='POST' action='' enctype='multipart/form-data'>
+     Firmware:<br>
+                  <input type='file' name='firmware'>
+                  <input type='submit' value='Update Firmware'>
                </form>
-         </body></html>\n)";
+     <form method='POST' action='' enctype='multipart/form-data'>
+     Spiffs:<br>
+                  <input type='file' name='spiffs'>
```
caverna: What about to include ```accept='.bin'``` in both fields to filter a little?

caverna:
Hi @sandeen !
Would you like some help with the conflict?
It will be great to have this feature in next release...

sandeen:
I'm not currently working on this project and unfortunately don't have time to pick it up again right now, so I can't do proper testing a the moment.  If anyone wants to take my patch as a base and get it done, go right ahead.

caverna:
OK, I'll do it!

earlephilhower:
Seems like a reasonable, minor addition and looks like it's already in a state to use (modulo whenever LittleFS gets in and SPIFFS_end turns into FS_end).  Let's consider this for 2.6.

d-a-v:
Closing, replaced by #3732

Issue #3247, opened 2017-05-16
can't customize the WifiClient timeout for connect


Zonglin-Li6565:
#### Hardware
Hardware:		NodeMCUv2, ESP-12E
Core Version:      	?

### Description
The WifiClient connect function hang the program
### Settings in IDE
Doesn't matter

### Sketch

```cpp

#include <Arduino.h>

void setup() {

}

void loop() {
        WiFiClient client;
        IPAddress ipAddr;
        ipAddr.fromString(IPADDR);
        Serial.println("Trying to log");
        if (client.connect(ipAddr, PORT)) {
            client.print(msg);
            Serial.println("Logged");
        }
        // Have some other time critical stuff here. The connect takes too long if the ipaddress doesn't exist. 
}

igrr:
This issue is tracked in https://github.com/esp8266/Arduino/issues/2120.

igrr:
Also https://github.com/esp8266/Arduino/issues/1420.

Issue #2120, opened 2016-06-08
Timeout set using Client::setTimeout is not taken into account in WiFiClient::connect


hallard:
#### Hardware

Hardware:       nodemcu ESP12
Core Version:       latest git
### Description

I think I misunderstood something, with HTTPClient library, I'm calling `http.setTimeout(1000);` it  works once connected to server and when doing a `http.get()`. No pb at all
But whatever value I use if the server is down or not responding `http.begin()` seems to default to 5 seconds
### Settings in IDE

Module:             NodeMCU V1.0
Flash Size:         4MB/1MB
CPU Frequency:  160Mhz
Flash Mode:     qio
Flash Frequency:    40Mhz
Upload Using:       OTA 
Reset Method:       nodemcu
### Sketch

Here my function 

``` arduino
bool WS_post(char * host, uint16_t port, char * url)
{
  HTTPClient http;
  bool ret = false;

  // Just in case
  http.setTimeout(2000);

  Serial.printf("[HTTP] %s:%d/%s", host, port, url);

  // configure target server and url
  if ( http.begin(String(host), port, String(url)) )
  {
    // start connection and send HTTP header
    int httpCode = http.GET();

    // httpCode will be negative on error
    if (httpCode > 0) {
      // HTTP header has been send and Server response header has been handled
      Serial.printf("[HTTP] GET... code: %d\r\n", httpCode);

      // file found at server
      if (httpCode == HTTP_CODE_OK) {
        String payload = http.getString();

        Serial.printf("'%s' (%d ms)\r\n",payload.c_str(), millis()-start);
        ret = true;
      }
    } else {
      Serial.printf("[HTTP] GET... failed, error: %s\n", http.errorToString(httpCode).c_str());
    }
  } else {
    Serial.printf("[HTTP] Begin... error\r\n");
  }

  http.end();
  return ret;
}
```
### Debug Messages

```
[HTTP] 80.11.xxx.xxx:82//?blablabla
after 5 seconds and not 2
[HTTP] GET... failed, error: connection refused
```
##

josep112:
Hi,Hallard 
You can solve this? I have the same problem

Humancell:
I'm curious about the status of this?  Particularly, I'm wanting to know what the default timeout is when using the WiFiClient client; / client.connect() call?  And if I know the default, can I change this?

Likewise (slightly off topic) what is the default timeout on ESPhttpUpdate.update()?  And is this also settable?

[...]

PR #3257, opened 2017-05-18
Configurable timeout for WiFiClient connect and write operations

### LINKED ISSUE #1420: client.connect() takes too long if server is unavailable...
In case the server is unavailable, the client.connect() takes too long to return false and come out of the block. This takes 1.5 secs approx. which is too long for my project. Is there a way to solve this problem? Can there be a timeout set in case the client cannot connect within N milli secs?
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29836488-client-connect-takes-too-long-if-server-is-unavailable?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

igrr:
WiFiClient inherits `Stream::setTimeout` method. The value passed to `setTimeout` is used by `Stream` methods like `readBytes`, `readStringUntil` and so on. Other operations with WiFiClient such as `connect` and `write` previously didn't have configurable timeouts. `connect` would time out after 5 seconds, and `write` would return only if LwIP reported an error.

This change implements timeout during `connect` and `write`. For `connect`, same timeout value is used for name resolution and the actual connection, so it is possible that worst case timeout will be twice the timeout requested. For writes, `write` function will return the amount of data sent by the time the timeout has happened.

Fixes #1420, #3247, #2120

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
@@ -426,6 +426,8 @@ void wifi_dns_found_callback(const char *name, ip_addr_t *ipaddr, void *callback
 void wifi_dns_found_callback(const char *name, const ip_addr_t *ipaddr, void *callback_arg);
 #endif
 
+static bool _dns_lookup_pending = false;
+
 /**
  * Resolve the given hostname to an IP address.
  * @param aHostname     Name to be resolved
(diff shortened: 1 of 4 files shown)


Jeroen88:
Hi Ivan, how can I set the connect timeout for the ESP8266HTTPClient?

aminjoharinia:
@Jeroen88 Did you found a solution?

Jeroen88:
@aminjoharinia Hi Amin, I think you can use void setTimeout(uint16_t timeout), this passes the timeout to the WiFiClient.

en00x:
Hi everyone!
Seems like the timeout isn't considered for write operations in my case.
When i run the following code and enter an area with bad wifi connection the write method returns after 300-400 ms but the timeout was set to 10 ms. It's blocking the rest of the program too long.
```
WiFiClient* client = new WiFiClient();
client->setTimeout(10);
...
unsigned long ptime = millis();
uint16_t rc = client->write(buf+(MQTT_MAX_HEADER_SIZE-hlen),length+hlen);
Serial.print("writetime: ");
Serial.println(millis()-ptime);
```

Issue #3192, opened 2017-05-03
Latest UART buffers break Serial.swap()


riataman:
I have a NodeMCU, I was using Serial.swap() just fine with 2.3.0. After upgrading to master Serial.swap doesn't work anymore. 

If I backup the serial buffer changes it works again.

igrr:
By "doesn't work", you mean that the output is still sent to the original pins after `swap`? Or some other issue?

igrr:
@riataman  Can you confirm that the issue is that you can not `swap` back to the original pins?

riataman:
Sorry, I haven't been able to get back to test this, but as I remember I was just calling ´Serial.swap´ once and the alternative pins didn't worked.

I'll try to get back to test this again soon.

devyte:
@riataman ping

d-a-v:
Closing due to age and because `Serial.swap()` is working.
Please open a new issue with more details if your tests are not satisfactory.

PR #3258, opened 2017-05-18
HardwareSerial: fix issue with swapping pins twice

igrr:
Previously calling HardwareSerial::swap twice would set the pins back
to the original setting (1, 3). This feature was lost when swap function
got an optional argument for the TX pin. This change reverts that,
restoring ability to swap pins back.
To use TX pin 2, HardwareSerial::pins or HardwareSerial::set_tx methods
can be used.


igrr:
Ref https://github.com/esp8266/Arduino/issues/3192

devyte:
@igrr what is the status of this? Assuming it's ok'd in #3192 by the OP, can it just be merged, or is there anything pending?

igrr:
This was long time ago... I don't remember how thoroughly this was tested. Will revisit this and try to test all possible pin options.

igrr:
I've done some testing on this and there are still some issues. Will consider #3192 a known issue for 2.4.0 and work on this fix again after the release.

igrr:
#hwtest

igrr:
#hwtest

devyte:
@igrr what is needed to move forward on this?

devyte:
This doesn't fit into 2.5.0. Pushing milestone back.

earlephilhower:

devyte:
This needs work, pushing back.

dok-net:
@devyte Regarding the issue mentioned in the title - if there's anything else to this PR beyond that, I have not looked for it and am disregarding it - it's not present in master, and this PR could be closed. I have successfully `swap()`ed multiple times, and used the `pins()` function as well, no unexpected behavior occurs.

PR #3285, opened 2017-05-24
Add a new board

sabiha321:


sabiha321:
We have designed a development board using ESP8266 chip. What should I done
to get ESP8266 library support(as other boards ex:Sparkfun thing,
adafruit boards).

vicnevicne:
Err, as igrr wrote above, please add some description (name of the product, specifications) as well as a link to a site to buy it from.
If that dev board is for internal use only (or not available yet), then it does not make sense to include it in the list that will pop-up for everyone.

sabiha321:
we have designed the board and it will be available as a product.I will
make some changes(Writing proper name for the board) and also I will add
adescription. Wil you add it to ESP8266 library so that it will be in the
list of the ESP8266boards as other boards.

PR #3121, opened 2017-04-06
Arduino boards with ESP8266

sergiotomasello:
This PR introduces a new board called `Arduino` with three models:

- `Star OTTO`
- `Primo`
- `UNO WiFi`

PR #3288, opened 2017-05-24
Add Crystal Frequency menu item for Generic module

jwrw:
The Generic ESP8266 Module provides the basis to compile for anything not on the current list.  Although most current boards use the 26MHz crystal, there are many using the 40MHz.  This change permits these to be readily accommodated as generic ESP8266 boards.


jwrw:
This builds on an earlier change to support Arduino boards with ESP8266 that have 40MHz crystals (#3121)

PR #3305, opened 2017-05-28
UDP updates

sticilface:
1)  Allows an Rx callback to be attached to WiFiUDP
2) Initial proof of concept idea to get MDNS to use WiFiUDP instance and not UDPContext. 

not expecting this to get merged straight away.  But kind of the direction I'm suggesting.  
Having an Rx callback makes WiFiUDP more attractive.

comments please. 

The callback is still compatible with the existing parsePacket() and available methods, they just never get called as the onRx function gets called first.

### DIFF
--- a/libraries/ESP8266WiFi/examples/WiFiUDPUsingCallback/WiFiUDPUsingCallback.ino
+++ b/libraries/ESP8266WiFi/examples/WiFiUDPUsingCallback/WiFiUDPUsingCallback.ino
@@ -1,6 +1,6 @@
 
 #include <ESP8266WiFi.h>
-#include <WiFiUDP.h>
+#include <WiFiUdp.h>
 #include <ArduinoOTA.h>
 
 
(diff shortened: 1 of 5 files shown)


sticilface:
Happy to chime in once this gets looked at.  I don't suppose the underlying problem has changed even with lwip2?

tonton81:
how can you have a callback when the packet going to it may be lost, or the callback comming back getting lost? for a guarenteed callback, you need a two-way acknowledgement system (which i already wrote in one of my projects) which guarentees both the send and return acknowledgements, without data loss.

sticilface:
quite easily.  The callback is for received UDP packets, which is a totally separate thing to sending packets.  UDP has no response to a sent packet.  The underlying implementation of UDP on esp8266 uses callbacks.  These changes just expose them for the user so you are not required to call parsepacket() every loop to check if there are any incoming packets.

tonton81:
nor is the callback guarenteed to deliver either :)

tonton81:
would be nice to have it return the payload and length instead of having it read off the internal buffer, async udp by no-me-dev does this

earlephilhower:
Thanks for your PR, but MDNS has been completely rewritten with this release, so this PR can't apply anymore without massive rework.

I'll leave open for two weeks and come back after Feb 10 and see if there's interest from the original submitter.  If not, will close then.

earlephilhower:

PR #3307, opened 2017-05-30
Ethernet: Fix out of bounds memset on reset DHCP lease

scop:


scop:
See discussion in https://github.com/arduino/Arduino/pull/6337

devyte:
Shouldn't that be memset(blah, 0, sizeof(blah))? I mean use sizeof()
instead of a magic number like "4".

scop:
Well kind of, I suppose. But then again, the 20 and the out of bounds memset is apparently intentional here, see arduino/Arduino#6337 (i.e. the 20 means `sizeof(_dhcpSubnetMask) + sizeof(_dhcpGatewayIp) + sizeof(_dhcpLocalIp) + sizeof(_dhcpDhcpServerIp) + sizeof(_dhcpDnsServerIp)` IIUC)

devyte:
I understand the meaning of the magic number, I'm just asking about
robustness and good coding practices. Also, being explicit with sizeof()
tells anyone who reads the code exactly what is being memset().

PR #3309, opened 2017-05-30
Python 3 fixes

scop:
Probably not quite enough to get the code running with Python 3, but it's a start and doesn't hurt with Python 2.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -159,7 +159,7 @@ linkcheck_anchors_ignore = ["/#!"]
 #
 # on_rtd is whether we are on readthedocs.org
 env_readthedocs = os.environ.get('READTHEDOCS', None)
-print env_readthedocs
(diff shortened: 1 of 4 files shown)


igrr:
Don't you need to add `from __future__ import print_function` to be compatible with Python 2?

scop:
On Sun, Oct 15, 2017 at 10:08 AM, Ivan Grokhotkov
<notifications@github.com> wrote:
>
> Don't you need to add from __future__ import print_function to be compatible with Python 2?

Not in simple cases like these.

devyte:
@scop @igrr Is this PR still valid? Is anything pending?

scop:
Still valid, yes, nothing pending from me. Just rebased against current master.

scop:
I think the CI failures are unrelated to these changes.

PR #3313, opened 2017-05-31
Fix URL parameter decoding in web server

scop:
The parameters string needs to be first split on `&` and `=`, and URL decoding on parts done _after_ that. Otherwise URL encoded `&` and `=` within parameter names and values cause incorrect splitting.

I haven't been able to actually test this, except that it compiles. Running the test gives me a "begin timeout" for some reason before the test run starts. Compilation and upload appear to work fine.

```
$ make TEST_LIST=test_http_server/test_http_server.ino V=1
[...]
Running tests
../../tools/esptool/esptool -v -cp /dev/ttyUSB0 -cd nodemcu -cr
esptool v0.4.9 - (c) 2014 Ch. Klippel <ck@atelier-klippel.de>
opening port /dev/ttyUSB0 at 115200
begin timeout
```


igrr:
I think this has been fixed in https://github.com/esp8266/Arduino/pull/2956, please reopen if I'm wrong.

scop:
@igrr The issue might be fixed in #2956 (haven't checked), but the test suite improvements here would still be very good to have merged. I don't think there's any way I can reopen this PR -- could you do it for me, thanks? (Or even better, just merge the two first commits in this issue... :))

igrr:
Ok, rebased the test-related commits on top of master, and the tests fail. Haven't looked into the "why" yet, but i guess the fix which was merged was not comprehensive enough.

libraries/ESP8266WebServer/src/Parsing.cpp
```diff
@@ -317,8 +313,8 @@ void ESP8266WebServer::_parseArguments(String data) {
       continue;
     }
     RequestArgument& arg = _currentArgs[iarg];
-    arg.key = data.substring(pos, equal_sign_index);
-	arg.value = data.substring(equal_sign_index + 1, next_arg_index);
+    arg.key = urlDecode(data.substring(pos, equal_sign_index));
```
innodron: Why decode the key as well, any use case ?

scop: Any case that uses non-URL-safe characters in key names is a use case for this. They need to be decoded as well as values.

devyte:
@igrr there has been corroboration of this fix by one user. Just waiting on feedback from #4013 before merging, so I added it to the 2.4.0 milestone.

PR #3319, opened 2017-06-01
Make pgm_read_byte() and pgm_read_word() again backwards compatible c…

wkraft-fablabka:
…ompatible to the Arduino implementation.

In result to issue #3140 they were converted from preprocessoer macros to functions.  This however breaks other libraries, which are checking the existence of those functions and otherwise redefining them.


wkraft-fablabka:
closed, as I will create another one

PR #3325, opened 2017-06-02
Define pgm_read_{byte,word} as macros for compatibility

wkraft-fablabka:
Ref #3140

pgm_read_byte() and pgm_read_word() need to be macros, as some third party libraries and sketches testing macro existence.


wkraft-fablabka:
As discussed before

PR #3328, opened 2017-06-03
Add getUserPasswordHash() and auth using hash

Sn4k3r:
Add getUserPasswordHash() which returns base64 hash.
Now you can use Username&Password or stored base64 hash for
authentication


earlephilhower:

earlephilhower:
After reviewing this, I don't see any use case.  Since the hash itself is just the b64 encoding of the raw user/pass, it's no more secure than putting them in plaintext in the app.

Seeing as it's also ancient, abandoned, and has merge-conflicts I'm closing.

PR #3329, opened 2017-06-03
SPI Write16, Write32, WriteBytes optimizations

rsaxvc:
A few patches to decrease idle time while refilling FIFO, as well as decrease codespace.


rsaxvc:
@vicnevicne , you're correct. I've updated the 16-bit version to take this into account. I've also expanded the commit messages.

PR #3398, opened 2017-07-03
Fix to parse parameters in the URL of a POST with empty content.

### LINKED ISSUE #3397: Parameters in a POST URL that has empty content are not parsed
#### Hardware
Hardware:			ESP-8266
Core Version:      	2.4.0-rc1

### Description

When an `ESP8266WebServer` receives a `POST` request that just consists of a URL with parameters, but no content, its URL is not parsed for parameters.

This is reproducible using [this example from the Blynk project](https://github.com/blynkkk/blynk-library/tree/master/examples/Export_Demo/Template_ESP8266) but it affects any project that uses the WebServer class and requires simple POST requests like:

```
POST http://example.com/?param=value
```
with an empty body.

teejaydub:
Looking at the change at 0d996ab75e5e368d4a9de5b87dfecc91c86a4c83, it looks like it was intended to move the call to `_parseArguments()` into each branch above it, so you could decode characters first in the special case of #2241. But it needs an additional call in the branch where it's not a form and there's no content - otherwise a POST with parameters in the URL breaks.

Fixes #3397.


pieman64:
@teejaydub thanks for the commit, confirming Blynk provisioning now works again.

PR #3423, opened 2017-07-15
Subtree axtls

madpilot:
I was about to get cracking again on #3105 but based on the feedback from @igrr and the comments on #3223 I thought it worth tackling bringing AxTLS in to the build tree.

Using a submodule would work, but they are error prone (forgetting to sync up the submodules is a pain), so I thought about using a subtree, but as I was refreshing my memory on how to set them up, I came across this subrepo project: https://github.com/ingydotnet/git-subrepo#readme. Have a read through, though the TL;DR is there is no extra software or things to worry about for developer just working on the top level repo (ie almost everbody), and there is just an additional script forr those working with the axtls repo.

I've added in https://github.com/igrr/axtls-8266 (in the tools/sdk/axtls directory) and replaced the sha1 implementation in Hash.cpp to use the one from the library instead of the standalone implementation that is currently used.

I've also hacked the library so it would compile when using WifiClientSecure, however it's become clear that some code from the axtls has already been manually copied over, and there has been a few other modifications to the library that may be better located in the core tree.

IMO it would makes sense to have the axtls library as close to the sourceforce version, to make updating easier, so any ESP8266 specific stuff should be in core (where possible).

So my questions:

1) is the intention of igrr/axtls-8266 to bring axtls to this project, or is it intended for non-arduino projects too?
2) How do we feel about removing the copied code, and using the version from  igrr/axtls-8266?
3) Has there been heaps of modifications to the existing axtls library?

I wouldn't mind having a bit of a discussion before committing time to cleaning this all up.


madpilot:
Whoops. Wrong branch. Disregard!

PR #3438, opened 2017-07-20
Update ESP8266HTTPClient.cpp

michaelchien1972:
1. bypass httpsFingerPrint check for Firebase
2. add Type, Event, Pending switch

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -63,6 +63,7 @@ public:
 
     bool verify(WiFiClient& client, const char* host) override
     {
+	    //make sure you change WiFiClientSecure.cpp to by pass httpsFingerPrint check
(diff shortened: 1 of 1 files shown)


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -234,6 +235,36 @@ void HTTPClient::end(void)
...
+/* support FirebaseStream - Type, Event */
+String HTTPClient::getType(void)
+{
+	String type = _tcp->readStringUntil('\n').substring(7); //put
+	return type;
+}
+
+String HTTPClient::getEvent(void)
+{
+	String event = _tcp->readStringUntil('\n').substring(6); //{"path":"/","data":"416.28"}
```
igrr: This is something specific to Firebase, and does not make much sense in the context of plain HTTP protocol. I suggest making a class derived from HTTPClient which would have these Firebase-specific changes.

PR #3439, opened 2017-07-20
Update ESP8266HTTPClient.h

michaelchien1972:
add Type, Event, and pending switch for Firebase


igrr:
This seems to be part of #3438. Closing, please see my comment there.

PR #3440, opened 2017-07-20
Update WiFiClientSecure.cpp

michaelchien1972:
skip httpsFingerPrint check for Google Firebase


copercini:
But this will skip the verification of all sites, turning the library unsecure...

In your case, just don't call `.verify();` in your code

igrr:
Agree with @copercini, that's not a good approach.

PR #3478, opened 2017-07-29
Move UMM debug messages into PMEM

earlephilhower:
Store the UMM debug format strings in PMEM instead of RODATA.  Frees around 300 bytes of RAM with debugging disabled, much more when debugging is enabled.


cores/esp8266/umm_malloc/umm_malloc.c
```diff
@@ -550,44 +551,45 @@
 #undef DBG_LOG_FORCE
 
 /* ------------------------------------------------------------------------- */
+#define printf_P(fmt, ...)  do { static const char fstr[] PROGMEM = fmt; char rstr[sizeof(fmt)]; for (size_t i=0; i<sizeof(rstr); i++) rstr[i] = fstr[i]; printf(rstr, ##__VA_ARGS__); } while (0)
```
igrr: Since we're doing this, can we also `#define printf(...) printf_P(__VA_ARGS__)`? It will make it much easier to upgrade umm to upstream version, if we ever want to do that.

igrr:
(p.s. Thank you for all the contributions — I have also added you as collaborator to the repository!)

cores/esp8266/umm_malloc/umm_malloc.c
earlephilhower: Thanks for the collaborator status.  I'm happy to help out.

I'll refactor the patch this weekend and put in a new one with printf() macro'd instead of the changes inline.

earlephilhower:
Here's the new, 2-liner.

PR #3487, opened 2017-08-01
Case-insensitive deviceType compare and skip NOTIFY processing

snosrap:
* ifndef'd LWIP_OPEN_SRC to prevent a redefined warning
* ABORT on NOTIFY to prevent responding to NOTIFY messages (we should only respond to M-SEARCH messages)
* case-insensitive compare of _deviceType to enable response to all-lowercase Alexa queries (robustness principle)


igrr:
@me-no-dev, can you have a look at this SSDP change?

me-no-dev:
LGTM! I should actually get on this and clean up the code a bit.

PR #3492, opened 2017-08-03
ESP8266WebServer minor public interface addition - content from const char*

timw1971:
Added public functions to allow content to be uploaded using const char*. For some cases, this can remove the need for content to be copied into a String, and thus can be considerably more space-efficient.


libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -114,13 +114,15 @@ class ESP8266WebServer
...
   void send(int code, char* content_type, const String& content);
+  void send(int code, const char* content_type, const char* content, size_t contentLength);
   void send(int code, const String& content_type, const String& content);
   void send_P(int code, PGM_P content_type, PGM_P content);
   void send_P(int code, PGM_P content_type, PGM_P content, size_t contentLength);
 
   void setContentLength(size_t contentLength);
   void sendHeader(const String& name, const String& value, bool first = false);
   void sendContent(const String& content);
+  void sendContent(const char* content, size_t contentLength);
```
igrr: For many people programming in C, `const char*` implies a zero-terminated string. Of course, it can also be used to pass arbitrary binary data, but most APIs i have seen use either `const void*` or `const uint8_t*` for this purpose. I would suggest changing the type here to hint that a null-terminated string is not expected.

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -341,6 +349,20 @@ void ESP8266WebServer::sendContent(const String& content) {
   }
 }
 
+void ESP8266WebServer::sendContent(const char* content, size_t contentLength) {
+  const size_t unit_size = HTTP_DOWNLOAD_UNIT_SIZE;
+
+  while (contentLength) {
```
igrr: This implementation does not match the one for sendContent(String&), which uses chunked encoding and leaves fragmentation to the lower level code (WiFiClient). It would be nice if this sendContent(data, size) and sendContent(String) would behave the same way.

earlephilhower:

earlephilhower:
Replaced with #6797

PR #3497, opened 2017-08-05
ESP8266WebServer form file upload boundary tag fix

scottwday:
When performing a file upload from a form post, C#'s HttpClient MultipartFormDataContent places the boundary string in quotes. 
This change strips the quotes from the boundary string (if detected)


igrr:
Can you please check if https://github.com/esp8266/Arduino/pull/3455 does the same thing?

scottwday:
Oops- missed that one- yes, this is a duplicate of #3455

PR #3503, opened 2017-08-08
Fix localPort for udp connection

tackelua:
In before version: localPort = 2390;
-> example NTPClient.ino get no udp packet after sends NTP packer


igrr:
Reading https://tools.ietf.org/html/rfc4330#section-4, i don't see why the source port needs to be set to 123. Specifically,

> The UDP port number assigned by the
   IANA to NTP is 123.  The SNTP client should use this value in the UDP
   Destination Port field for client request messages.  *The Source Port
   field of these messages can be any nonzero value* chosen for
   identification or multiplexing purposes.  The server interchanges
   these fields for the corresponding reply messages.
   *This differs from the RFC 2030 specifications, which required both
      the source and destination ports to be 123.*

herrold:
@igrr 
> i don't see why the source port needs to be set to 123

This is convention rather than RFC -- many low knowledge graphical firewall management tools may only ( or only easily) configured to have identical USP Source and Dest ports ... Stupid, but the world is full of such, and using a privileged local Source port is harmless to our purposes here

Put the port to use in a configuration variable perhaps as a more general solution, and default set to to '0', meaning: use a random high Source port, perhaps.  That way people who need it may over-ride it and use a matching port.  Would you like a diff?

d-a-v:
> example NTPClient.ino get no udp packet after sends NTP packer

After a quick search including one NTP RFC, I couldn't find any pointer describing this convention. Feel free to post any pointer about that. Such a change should be accompanied with doc/ref.

The example is working as-is (tried with 2390 and 123 without notable difference), the issue is old, lwIP offers automatic and configurable NTP support by default, and libc(newlib) allows to set timezone per continent and city with DST support (https://github.com/esp8266/Arduino/pull/6373#issuecomment-536235781).

PR #3515, opened 2017-08-13
add DigiStump Oak pins & board

boneskull:
The [DigiStump Oak](http://digistump.com/category/22) is an ESP8266-based board, which is no longer focused on Particle integration.  AFAIK, the only way these boards are now useful is if they are running firmware built with community-supported SDKs, like this or NodeMCU.

This PR adds the Oak to the list of boards.  Here's a pinout:

image


ivankravets:
@igrr This PR looks fine. Could we merge it? We also will accept this PR https://github.com/platformio/platform-espressif8266/pull/47

igrr:
Package template json file also needs to be updated to include the new board.

boneskull:
@igrr I'm not sure what file you mean.  I have a `package.json`, the contents of which are:

```json
{
    "name": "framework-arduinoespressif8266",
    "description": "Arduino Wiring-based Framework (ESP8266 Core)",
    "url": "https://github.com/esp8266/Arduino",
    "version": "2.4.0-rc.1"
}
```

igrr:
Sorry I wasn't clear. I mean this file: https://github.com/esp8266/Arduino/blob/master/package/package_esp8266com_index.template.json

Another thing is that you also need to add a section to the documentation: https://github.com/esp8266/Arduino/blob/master/doc/boards.rst. As far as I understand, this board does not have a built-in usb-uart adapter, so you may want to add some information on connecting the board to an external usb-uart to achieve automatic entry into download mode.

boneskull:
Will do.

boneskull:
@igrr I've made the requested changes.

PR #3572, opened 2017-09-04
Update ESP8266HTTPClient.cpp to resolve malformed HTTP request

mateuszdrab:
This fixes an issue which was causing HTTP requests to get sent with wrong HTTP headers (e.g. HTTP 1.F or HTTP 1.A) which resulted in IIS rejecting the requests with 400 errors and failing the firmware upgrade.


devyte:
Hi, I'm confused as to how the F macro would fix the described issue,
considering that the relevant 1-char string is appended to a String object?

mateuszdrab:
I believe this could be because I injected the code needed to make this work into my existing sketch which was doing a lot of string manipulation, there could have been some kind of buffer overflow. Anyway, after taking all the string code away and converting into char arrays, it still persisted. I then saw that F macros are being used in some parts of the code and not others and changed it, once I did that the issue got resolved immediately and the HTTP requests succeed. I guess the nobrainer here is that if the string is stored in flash, it can't get overwritten easily...

igrr:
@mateuszdrab  There are a few other `header += something` lines where the right hand side is not in PROGMEM, would you mind updating those lines as well as part of the PR?

mateuszdrab:
Hi @igrr

I've made a few more changes to lines 'header += something', please have a look and let me know what you think.

Regards

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -889,7 +889,7 @@ bool HTTPClient::sendHeader(const char * type)
         _base64Authorization.replace("\n", "");
         header += F("Authorization: Basic ");
         header += _base64Authorization;
-        header += "\r\n";
+        header += F("\r\n");
     }
 
     header += _headers + "\r\n";
```
igrr: At least one more here

igrr: Since \r\n is used in a bunch of places, it might make sense to define it once as `static` variable at the top of the file (e.g. `crlf`) and reference it from all these places.

earlephilhower:

earlephilhower:
Closing because the HTTPClient has been refactored massively in the interim and the referenced Strings are now in flash already or converted to chars.

PR #3576, opened 2017-09-05
Increased I2C buffer size to match Espressif's ESP32 Arduino core

carlsa:
I have increased the I2C buffer from 32 to 128 bytes to match the buffer length that Espressif uses in their ESP32 Arduino core so that code will be compatible.

https://github.com/espressif/arduino-esp32/blob/master/libraries/Wire/src/Wire.h


michapr:
somebody could merge it ?
There are problems as sample with SPS30 (and maybe other devices. Best would be configurable buffer of course ;)
(see https://github.com/Sensirion/embedded-sps/issues/14 )

PR #3577, opened 2017-09-06
Avoid memory leak on multiple i2s_begin()s

earlephilhower:
i2s_begin allocated sample buffers every time it is called, but those
buffers were never freed in the _end() call.  Calling i2s_begin more
than a single time ever would therefore result in the leak of the old
buffers.

Free the buffers on i2s_end(), and ensure that no i2s is running in
the i2s_start() call to avoid the leak and any potential race conditions.


earlephilhower:
My bad, need to merge updated version before the PR.  Closing, ignore please.

PR #3578, opened 2017-09-06
Update spiffs and mkspiffs

me-no-dev:


igrr:
LGTM. @me-no-dev, do you have a Raspberry Pi around to build mkspiffs for linux? Can't find how I cross-compiled mkspiffs for linux-arm last time...

me-no-dev:
Yup will do a bit later :)

igrr:
Added armhf build mkspiffs, and merged in 6ac4812, thanks!

Lan-Hekary:
I've noticed that when I change the release form 2.3.0 to 2.4.0 or the opposite 
I experience a delay in the start of the code and then the Exciting File System gets Formatted !!!
after a bit of digging into the cause I noticed that `SPIFFS.begin()` Doesn't return a false value if the mounting fails but instead formats the file system ..

and the mounting obviously fails because the change in SPIFFS Version ..
so ,is there any way to maintain Compatibility between Different Versions of SPIFFS ??

and if there isn't a way .. at least `SPIFFS.begin()` Should not Format but return a false and the user should decide to format or roll back and get a backup of his files first :D 
I lost a Huge log file due to this unexpected behavior ..

Issue #3570, opened 2017-09-04
Wire.h cannot be initialized from a class constructor in a library


willz1200:
### Basic Infos

Wire.h cannot be initialized from a class constructor in a library.

#### Hardware
Hardware:		ESP-12E
Core Version:      	2.3.0

### Description

When trying to use Wire.begin(4, 5) from a class constructor within a library the esp crashes Exception (28). Because the class is initialized within the library, Wire.begin will be ran before the setup loop. I'm using software I2c which uses the twi.h esp8266 core.

By the way I've also tried doing this with SPI.begin() and it worked perfectly, so its just an issue with a feature of Wire.h

### Settings in IDE

Module:  			NodeMCU 1.0 (ESP-12E Module)
Flash Size: 		4M (3M SPIFFS)
CPU Frequency:	80Mhz
Upload Using:		SERIAL
Reset Method:		NodeMCU

### LibraryUsingWire.h

```cpp
#include <Arduino.h>
#include <Wire.h>

class classUsingWire{
	public:
		classUsingWire(); //The Constructor
};

extern classUsingWire Something;
```

### LibraryUsingWire.cpp

```cpp
#include <LibraryUsingWire.h>

classUsingWire::classUsingWire(){
	Serial.begin(115200);
	Serial.println("Serial Online");
	Wire.begin(4, 5);
}

classUsingWire Something; //Initialize class
```

### Sketch

```cpp
#include <Arduino.h>
#include <Wire.h>
#include <LibraryUsingWire.h>

void setup() {
  
}

void loop() {

}
```

### Debug Messages

```
Serial Online

Exception (28):
epc1=0x40201d9e epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000014 depc=0x00000000

ctx: sys 
sp: 3ffff900 end: 3fffffb0 offset: 01a0

>>>stack>>>
3ffffaa0:  4022fd50 00000001 3ffee2e8 40201d9c  
3ffffab0:  3ffe8d9c 60000600 3ffee3ec 40201e05  
3ffffac0:  4020439a 00000000 3ffe8918 40201e20  
3ffffad0:  402050ca 00000000 60000600 402022b1  
3ffffae0:  00000000 40206ef9 00000000 40206f05  
3ffffaf0:  40101b70 000003fd 00000010 60000200  
3ffffb00:  3fffff10 d7be081a 9f9205fd c357c8b0  
[...]

PR #3586, opened 2017-09-08
#3570 - Software i2c crashing ESP

willz1200:
If software i2c is initialised before the arduino setup function (e.g. In a class constructor) it crashes due to Exception (28). flush() is conflicting with something external to Wire.h and causing the crash. Its unnecessary because the buffers are initialised to 0 anyway.

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -58,7 +58,6 @@ void TwoWire::begin(int sda, int scl){
   default_sda_pin = sda;
   default_scl_pin = scl;
   twi_init(sda, scl);
-  flush();
 }
 
 void TwoWire::pins(int sda, int scl){


igrr:
Interesting, i thought that class member would take precedence over global (which would have to be called via `::flush`). Can you check if replacing `flush();` with `this->flush();` solves the issue? The reason why i think we shouldn't remove it is that `begin` can, in theory, be called multiple times.

willz1200:
I tried replacing `flush()` with `this->flush()` but the esp is still crashing. Any other thoughts?

devyte:
Access of a global object during construction of a different global object declared in a different TU is undefined behavior. The reason is that there is no guarantee in the order of construction or initialization, because the order of linking of the TUs is non deterministic.

The only case where the order is assured is in construction of globals within the same TU.

In Arduino world, global singletons is a common programming pattern. In such cases, where one object may need to init another, the init must be deferred, and can't be done on construction. To be clear, calling begin() of a global object from the constructor of a different object in a different TU may crash. In fact, calling most methods of a global object from the constructor of another object may crash.  
This is the main reason behind the begin() methods, which are supposed to be called from setup(). At that time, all globals are already constructed, so it's ok to init them in whatever order required.
In other words, your global class should also have a begin() or start() method that is called from setup(), and that method should call Wire.begin().

Closing due to user error.

Issue #2916, opened 2017-02-02
Race in interrupts??


andbjo:
#### Hardware
Hardware:			ESP-12E
Core Version:      	?2.1.0-rc2?

### Description

Problem description
In file: esp8266/cores/esp8266/core_esp8266_wiring_digital.c
In function: __attachInterrupt
the interrupt handler is set before cleaning interrupts for the receive pin witch make the receive sensitive to transitions high to low or low to high. 
When this happends the device gets stuck in boot phase. 

My proposal is to move 
the line `interrupt_handler_t *handler = &interrupt_handlers[pin];`  to after cleaning interrupt.
As:

extern void ICACHE_RAM_ATTR __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int mode) {
  if(pin < 16) {
//    interrupt_handler_t *handler = &interrupt_handlers[pin];
    handler->mode = mode;
    handler->fn = userFunc;
    interrupt_reg |= (1 << pin);
    GPC(pin) &= ~(0xF << GPCI);//INT mode disabled
    GPIEC = (1 << pin); //Clear Interrupt for this pin
// Put tis line here:
interrupt_handler_t *handler = &interrupt_handlers[pin];
    GPC(pin) |= ((mode & 0xF) << GPCI);//INT mode "mode"
  }
}

### Settings in IDE

Module:  			NodeMCU
Flash Size: 		4MB
CPU Frequency:		80Mhz
Flash Mode:			?qio?
Flash Frequency:	40Mhz
Upload Using:		SERIAL/USB
Reset Method:		nodemcu

/Anders

devyte:
@igrr if this is valid, I think we should give it priority, as it can cause a soft-brick.

igrr:
Moving the declaration of `handler` would not work like that, but thank you for pointing out the issue @andbjo. #3598 has been created to fix this. Please take a look (and try, if you can).

andbjo:
I know it was not the right way do do it, but it was a quick and dirty fix for me.

d-a-v:
@andbjo this issue seems to be solved by #3598, can it be closed ?

andbjo:
onsdag 13 december 2017 11.17.30 skrev  david gauchard:
> @andbjo this issue seems to be solved by #3598, can it be closed ?

I have not had time to test the new code yet. 
When I look at the code, it looks good.
So i think it can be closed.

-- 
My memory is like the Bermuda Triangle. Information sent there, can completely 
[...]

PR #3598, opened 2017-09-12
core: mask GPIO interrupts while attaching/detaching ISR handler

igrr:
Fixes https://github.com/esp8266/Arduino/issues/2916.

### DIFF
--- a/cores/esp8266/core_esp8266_wiring_digital.c
+++ b/cores/esp8266/core_esp8266_wiring_digital.c
@@ -152,6 +152,7 @@ void ICACHE_RAM_ATTR interrupt_handler(void *arg) {
 
 extern void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void *arg, int mode) {
   if(pin < 16) {
+    ETS_GPIO_INTR_DISABLE();
     interrupt_handler_t *handler = &interrupt_handlers[pin];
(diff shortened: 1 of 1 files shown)


igrr:
@ivankravets it looks like `pip install -U https://github.com/platformio/platformio/archive/develop.zip` fails in Travis CI, could you please take a look?

https://travis-ci.org/esp8266/Arduino/builds/274115561
https://travis-ci.org/esp8266/Arduino/builds/274563672
https://travis-ci.org/esp8266/Arduino/builds/274563832

ivankravets:
@igrr I don't know how did it work before 😱 

Please merge this PR: https://github.com/esp8266/Arduino/pull/3599

igrr:
@ivankravets thanks, I see the same line in arduino-esp32 travis-ci script, should it be updated there as well?

ivankravets:
@igrr arduino-esp32 is OK because it uses [language: python](https://github.com/espressif/arduino-esp32/blob/29a253ac9823fdb8189cfd72745a28143fe3c37d/.travis.yml#L3). So, Travis.CI will create Python virtual environment automatically before a build start. In case with `language: bash`, `pip install` will install packages to system "storage" where we need `sudo` permissions.

teprom:
Just giving my feedback on that: I have installed the last release with the above fix and now the system is not being reset anymore. Seems like this solved the problem. Thanks...

PR #3635, opened 2017-09-22
Fix erase size in ESP.eraseConfig

igrr:
SDK uses final 4 sectors of flash for configuration data.
ESP.eraseConfig would only erase 2 sectors, so in some cases of
corrupted data ("system param error"), users could not fix the issue
using ESP.eraseConfig, and had to use esptool instead.

Thanks @HugoML for reporting this.


tablatronix:
What is stored in first 2 sectors? Because I cannot see any effect from eraseConfig at all in 1.5.3, or at least I see no change in opmode_default or STA SSID, password.

However, This patch seems to work for 1.5.3, not sure how one would do semantic version comparisons to polyfill it, hmm.

PR #3654, opened 2017-09-26
Add close() method for SD

mohamadkhani:
The problem was we cannot close spi connection of sd card and release the port for another usages.
I add a method that do that and works for me.


devyte:
Closing in favor of #5402

PR #3712, opened 2017-10-12
Invoke callback with NULL ip address on error or name not found

fernandolopez:
The line that jumps to the piece of code invoking the callback on errors is commented.

If there is an error or an inexistent name is used with dns_gethostbyname() the callback is never invoked leading to the program getting stuck.


igrr:
In the current version, `hostByName` method of WiFi class has a timeout parameter, so the application shouldn't get stuck indefinitely. If you are seeing this issue in latest git version, could you please explain how to reproduce it? Thanks.

fernandolopez:
I was calling dns_gethostbyname() directly in a similar fashion as hostByName invoked it in version 2.3.0, with the caller invoking esp_yield() and the callback invoking esp_schedule().

Now I see the adventages of using a timeout to avoid any problems.

Issue #2536, opened 2016-09-18
Serial.flush() not flushing correctly at 160Mhz?


themindfactory:
### Hardware
Hardware:		ESP-12E
Core Version:      	2.3.0

### Description

When I run a sketch at 80Mhz all is OK 160Mhz brings in new issues.

I send out about 3000 bytes of data to serial with:

```
ptr = s;
while (*ptr) {
   Serial.write(*ptr++);
   // just in case its very long
   yield();
}

```

then I do with no delay

```
Serial.flush();
//delay(10);
Serial.swap();

```

I end up with an odd character on the serial port 0xEA when I expected a 0x0A (I guess we call it the default serial port on pins 1 and 3)

if I add a delay(10) after the flush all is OK.... its like the flush does not get to complete.

RichardS

### Settings in IDE

Module:  			Generic ESP8266 Module
Flash Size: 		4MB/1MB
CPU Frequency:	160Mhz
Flash Mode:		dio
Flash Frequency:	40Mhz
Upload Using:		SERIAL
Reset Method:		nodemcu

WereCatf:
This is something I've seen a whole bunch of people complaining about lately and I tried to look through the sources, but I can't find the problem there. I don't know if I'm just missing it, or if it's an actual H/W-side problem? Just out of curiosity, have you tried with a smaller value to delay(), and if so, what's the smallest delay you can use that still ends in the correct results? Does delay(1) work, for example, or does it then spit wrong data out?

stickbreaker:
The flush() call should wait until the uart issues the TXComplete interrupt.  On the AVR ATMega328p this is the TXCn bit of the UCSRnA register. I looked at the code in the HardwareSerial library, but I was not able to understand/decode how to find the equivalent status bit for the ESP.  Is there any documentation available for the ESP UART?
Chuck.

themindfactory:
Very interesting....

So I tried:
delay(1) and it worked
delayMicroseconds(10) and it worked
delayMicroseconds(9) and it worked
delayMicroseconds(8) and it failed

That's almost exactly one bit time at 115200! == 8.7uS

RichardS

WereCatf:
@stickbreaker https://github.com/esp8266/Arduino/blob/master/cores/esp8266/uart.c -- uart_wait_tx_empty does wait until the corresponding register changes status, but apparently when you run the ESP8266 at 160MHz the register changes status too early, leaving one more bit to be transferred.

[...]

Issue #2502, opened 2016-09-11
Serial.flush() have a problem


jyzhkj:
void loop()
{
digitalWrite(12, 1);
Serial.print("abcdffew");
Serial.flush();
digitalWrite(12, 0);
delay(100); // wait for a second
}

see the 12 pin when down, the last byte is not Send End.

so I try to change the HardwareSerial.cpp to:

void HardwareSerial::flush()
{
if(!_uart || !uart_tx_enabled(_uart)) {
return;
}

uart_wait_tx_empty(_uart);
long tt =uart_get_baudrate(_uart);
tt=12000000/tt;

delayMicroseconds(tt);
}

plinioseniore:
Hi @jyzhkj from your code looks like you are using an USART to RS485, I've found the same behavior and the last byte is not send as the flush doesn't empty the send queue before moving to the next instruction.

fishersolutions:
I have also found the same issue.
Is there a fix other than adding in your own delay?

d-a-v:
@jyzhkj This 2 years old issue seems to have been solved a year ago by #3714.
Is it OK, can you close it ?

PR #3714, opened 2017-10-13
Add a workaround-delay in Serial.flush()

WereCatf:
In relation to #2536 and #2502

Tested at 80MHz and 160MHz with flash-frequency at both 40MHz and 80MHz, the bug mentioned in the above issues manifests in all cases. The proposed workaround seems to work fine, I tested at 2400bps, 9600bps, 115200bps, 230400bps and 2Mbps and didn't see anomalous output.

### DIFF
--- a/cores/esp8266/HardwareSerial.cpp
+++ b/cores/esp8266/HardwareSerial.cpp
@@ -154,7 +154,7 @@ void HardwareSerial::flush()
     uart_wait_tx_empty(_uart);
     //Workaround for a bug in serial not actually being finished yet
     //Wait for 8 data bits, 1 parity and 2 stop bits, just in case
-    delayMicroseconds(1100000 / uart_get_baudrate(_uart) + 1);
+    delayMicroseconds(11000000 / uart_get_baudrate(_uart) + 1);
 }
 
 size_t HardwareSerial::write(uint8_t c)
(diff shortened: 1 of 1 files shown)


cores/esp8266/HardwareSerial.cpp
```diff
@@ -152,6 +152,9 @@ void HardwareSerial::flush()
     }
 
     uart_wait_tx_empty(_uart);
+    //Workaround for a bug in serial not actually being finished yet
+    //Wait for 8 data bits, 1 parity and 2 stop bits, just in case
+    delayMicroseconds(1100000 / uart_get_baudrate(_uart) + 1);
```
devyte: Your original comment in the issue said this worked:
delayMicroseconds((int)(1000000.0 / uart_get_baudrate(_uart) * 11 + 1))

There is a difference of a factor of 10. Which is correct?

WereCatf: 1000000 * 11 == 11000000. Nuff said.

devyte: Yes, but 1100000 != 11000000
I think your PR is missing a zero.

WereCatf: Yeah, good catch. I don't know how that did happen.

PR #3717, opened 2017-10-13
fix gettimeofday()

Juppit:
tp->tv_usec = micros() not micros() * 1000


igrr:
Cherry-picked in 1c8bcf6, thanks!

PR #3722, opened 2017-10-15
boards.txt generator

d-a-v:
To quickly try it in a separate branch:
```git fetch origin pull/3722/head:pr-3722 && git checkout pr-3722; git branch```
```tools/boards.txt.py --help```
```tools/boards.txt.py > boards.txt```
and restart arduino.

[latest](https://raw.githubusercontent.com/d-a-v/Arduino/boards/boards.txt)  boards.txt - without debug option NULL nor generic boards' LED_BUILTIN_LED
[2.4.0rc2](https://raw.githubusercontent.com/esp8266/Arduino/2.4.0-rc2/boards.txt) boards.txt before the generator

TODO: #3982

### CODE REVIEW

**Comment by davisonja** [comment]
Why are entries in 'opts' in a different format (key/value) to the entries in 'macros' (list)?

**Comment by d-a-v** [comment]
Python's dicts do not preserve order (like map:: in c++).
I had dicts everywhere but I switched later to lists were I needed to preserve order.
Should I convert 'opts' format to lists ?

**Comment by davisonja** [comment]
OrderedDicts preserve order.
I prefer the key value approach of dicts  (and arrange my structure
processing to account for the orderlessness - hence the mix of  lists in
dicts of my examples) so would subsequently prefer them.

However, I would say yes, consistency is better, particularly when the
different bits are actually to be used in the same place in the output.


On 16/10/2017 8:29 pm, "david gauchard" <notifications@github.com> wrote:

Python's dicts do not preserve order (like map:: in c++).
I had dicts everywhere but I switched later to lists were I needed to
preserve order.
Should I convert 'opts' format to lists ?

—

<https://github.com/esp8266/Arduino/pull/3722#issuecomment-336802774>, or mute
the thread

**Comment by d-a-v** [comment]
By 'preserve order' I was meaning the order coded in the source code which reflects the order in arduino gui's menu. I just learned that python>=3.6 preserves the source order. Or we could use python's OrderedDict (the syntax changes). 
I was wrong saying that we could easily use list everywhere, because I use 'opts' dict to make a search (line 656)  to allows 'opts' to override a macro entry.
```
if [...] (keyval[0] in board['opts'])
```
This leaves:
 - use ```from collections import OrderedDict```
 - force the use of python 3.6 :-)
 - stick with mix of list and dict
 - something else I'm not aware of

If you need the search facility of dicts, I guess I should try and use OrderedDict ?

**Comment by d-a-v** [comment]
There are dicts instead of lists now.

File: tools/boards.txt.py [hunk 1]
```diff
@@ -652,9 +643,9 @@ def all_flash_size ():
 		macrolist += board['macro']
 	macrolist += [ 'debug_menu' ]
 	for block in macrolist:
-		for keyval in macros[block]:
-			if not ('opts' in board) or not (keyval[0] in board['opts']):
-				print short + keyval[0] + '=' + keyval[1]
+		for key in macros[block]:
+			if not ('opts' in board) or not (str(key) in board['opts']):
```
davisonja: I also think the dict makes this line clearer with 'key' instead of 'keyval[0]' :)

davisonja:
It occurred to me a little later that lists would prevent the lookup in line 656 too :)

Using an OrderedDict (as you've done) is, I think, the simplest.

The other relatively easy approach (though with the layers here I wouldn't recommend it!) is to use `dict`s to store the info, and lists to index with order. So, for determining the order of the boards, instead of
````python
boards = dict(...)
for board in boards:
    process(board)
````
you instead have
````python
boards = dict(...)
menu_order = ['board1', 'board2', 'board3', ...]
for menu_entry in menu_order:
    process(boards[menu_entry])
````

**Comment by igrr** [comment]
Not intending to start a holy war here, but since this is Python, can we switch indentation to spaces?

**Comment by d-a-v** [comment]
I hope I understood everything well with the last commit.

**Comment by davisonja** [comment]
My question is about the 'helper function 'flash_size' - I'm wondering what advantage this provides over specifying the flash info in the same way as for the other board info (in `dict()`s)?
I like the approach of having all the data in data structures, and processing code that deals with it. In addition to clearly separating the (changeable) data from the code, it lends itself to having an inverse script that can generate the data structures from an existing boards.txt (and you can then plug in the data structures).
I'm not suggesting that the existing script is hard to follow, but it feels more involved and intertwined than is really necessary for the fairly simple source data we're dealing with - which leads me to my original question as I wonder what I'm missing :)

I'm also not a fan of positional arguments (particularly when mixed with keyword arguments) and would vote for using keyword arguments across the board.

**Comment by igrr** [comment]
I vaguely remember writing something about 'helper function flash_size' some time ago, but can't even find my own comment anymore. Anyway, current changeset looks good to me!

**Comment by d-a-v** [comment]
This question has [raised](https://github.com/esp8266/Arduino/commit/32f6826dcff160e1815f24b065da0f8517b55a9e) about the default serial upload speed. In the script I put 115200 by default for all boards, because
- I thought it was like that in the old boards.txt
- It is the most compatible value (I experienced problems with 921k on macOS and wemos, 460k was ok)

But after checking, it appears that at least the ```Wemos D1 R2/mini``` had 921k by default.
So should we use 921k by default for all boards ?

**Comment by davisonja** [comment]
I like reliability so I'd vote 115200 (or other well known minimum).
Changing things like boards.txt should be for improvement (faster) rather
than functionality (works at all)

**Comment by igrr** [comment]
Originally I thought it would be up to the board author to pick the highest
baud rate which worked reliably for a given board as the default. In my
experience, both nodeMCU 1.0 and D1 mini/ mini pro work reliably at 921600.
I think setting that as a default is okay as long as the user has an option
to lower the baud rate.

**Comment by davisonja** [comment]
It's potentially install dependant (board authority is best source,
hopefully) which leads me to the other thought about local overrides. Is it
worth having a .local with settings the script will use in preference to
the distribution options. Particularly with baud rates it might be handy..?

**Comment by igrr** [comment]
You mean boards.txt.local? I think the IDE actually supports some kind of a
local override file, but can't find a reference to it now, need to look at
the source code.

**Comment by davisonja** [comment]
Exactly, though I also wasn't sure if the ide had support (unlike
platform.txt.local, I think it is). We can certainly offer it in the
generation script regardless (so I started there).

**Comment by d-a-v** [comment]
and what about writing by default a "secure 115200" config, with command line options for users willing to regenerate with other defaults ?

**Comment by davisonja** [comment]
I like that idea.
I've also been wondering if a console-text-based question/answer wizard
might be useful. It could also offer this sort of thing (the other obvious
inclusion would be turning entire boards on or off, and the like)

**Comment by davisonja** [comment]
Just checked, it is "boards.local.txt", from revisions.txt:
> ARDUINO 1.6.6 - 2015.11.03
> [...]
> * Added boards.local.txt support: like platform.local.txt, allows to enrich a boards.txt definition without modifying the original file. Thanks @Wackerbarth

**Comment by devyte** [comment]
@davisonja @d-a-v what's the status on this?

**Comment by d-a-v** [comment]
Current output is in master's board.txt with so far no issues.
Improvements will be done about defaults values (as asked [here](https://github.com/esp8266/Arduino/pull/3722#issuecomment-342013817) and [here](https://github.com/esp8266/Arduino/pull/3722#issuecomment-342024334))
@davisonja may be working on it for autogeneration of ldscripts.
I can try and automate ```all_flash_size()```.
@igrr what are the priorities ?

**Comment by davisonja** [comment]
I am in theory, but have been a bit pressed for time this week. Will get onto it this weekend properly.
What automation are you thinking of for all_flash_size()?

I think should have an additional function for generating the ldscripts (possibly refactor so we have several functions that output to the file system, one for ldscripts and one for boards.txt itself?) to keep the flow clean.

**Comment by d-a-v** [comment]
all_flash_size() could call flash_size() from within a clever loop. They way it is currently done is subject to typos. I rekon it's icing on cake :)

**Comment by igrr** [comment]
Sorry, i lost track of this. I think in terms of priorities, we should fix anything that is seriously broken in 2.4.0-rc2 and prepare the final release. In that respect, board generating can probably wait a bit (since we have the new generated output already in master).

I'll try to devote some time this weekend to triaging of the bugs reported since 2.4.0-rc2 was released.

File: boards.txt [hunk 2]
```diff
@@ -178,13 +178,13 @@ generic.menu.FlashSize.16M15M.build.spiffs_blocksize=8192
 generic.menu.LwIPVariant.open=v2
```
igrr: It seems that placing lwip2 before v1.4 variants has a side effect of it being used by default.

I don't necessarily mind that (gets tested more, right?), just asking whether that was intended.

d-a-v: Yes this is true. In this board.txt and in current master, lwip2 is the default.
And yes lwip2 gets tested more this way. There is an option in the script to change that (```--lwip 1```).
A crash has since been fixed, and the lack of lwip's ping api (declared in espressif SDK) has been noticed but without further issue.

Ldscripts autogeneration is coming in the next few days.

whyameye:
I can confirm that this fixes issue #3850

d-a-v:
@igrr @davisonja @devyte  Please review.
It's working and generates/replaces boards.txt and ldscripts, it makes backups before that too.
I tried to keep it simple but I'm not a python expert.

davisonja:
Looks like a solid foundation to build on, and it shouldn't be too much to add in what I need for the ld script generation.

Personally I'd prefer some classes with properties/methods rather than functions with side-effects based on globals, but that's a potential for refinement, rather than a bug :)

d-a-v:
boards.txt generator is now able to patch/regenerate the boards part of ```package/package_esp8266com_index.template.json``` (with backup of the original file)

d-a-v:
@davisonja @igrr  Unless you think it is unwise,  I plan to produce ```doc/boards.rst``` with the generator.

devyte:
@d-a-v I was actually going to suggest the same thing. Not generating makes for doc maintenance, and additional issues.

Pablo2048:
I'm trying to implement 4M/2M SPIFFS variant an I ended with attached linker script and following modification to boards.txt:

generic.menu.FlashSize.4M2M=4M (2M SPIFFS)
generic.menu.FlashSize.4M2M.build.flash_size=4M
generic.menu.FlashSize.4M2M.build.flash_ld=eagle.flash.4m2m.ld
generic.menu.FlashSize.4M2M.build.spiffs_start=0x200000
generic.menu.FlashSize.4M2M.build.spiffs_end=0x3FB000
generic.menu.FlashSize.4M2M.build.spiffs_blocksize=8192
generic.menu.FlashSize.4M2M.build.spiffs_pagesize=256
generic.menu.FlashSize.4M2M.upload.maximum_size=1044464 

So I'm asking to guide how to add this into the generator?

[eagle.flash.4m2m.zip](https://github.com/esp8266/Arduino/files/1558952/eagle.flash.4m2m.zip)

d-a-v:
@Pablo2048 
A new line [there](https://github.com/d-a-v/Arduino/blob/cb6ba44d3c7ef0c5d537b40fba967248050678d9/tools/boards.txt.py#L751) is needed.
Size is 0x3FB000-0x200000
The ldscript would autogenerated with proper option

Pablo2048:
@d-a-v thank You. And how to handle the linker script?

Pablo2048:
@d-a-v looks like this line did the trick
f4m.update( flash_size(  '4M', '4M2M',    'eagle.flash.4m2m.ld',      '2M SPIFFS', 1044464, 0x200000,  0x1FB000, 8192))
so I'll do heavy testing on this and report back...

d-a-v:
with option ```---ld``` or ```--ldgen```

```
$ ./tools/boards.txt.py 

boards.txt generator for esp8266/Arduino

usage: ./tools/boards.txt.py [options]

        -h, --help
        --lwip                  - preferred default lwIP version (default 2)
        --led                   - preferred default builtin led for generic boards (default 2)
        --board b               - board to modify:
                --speed s       - change default serial speed
        --premerge              - no NULL debug option, no led menu
        --customspeed s         - new serial speed for all boards

        mandatory option (at least one):

        --boards                - show boards.txt
        --boardsgen             - replace boards.txt
        --ld                    - show ldscripts
        --ldgen                 - replace ldscripts
        --package               - show package
        --packagegen            - replace board:[] in package

available serial speed options (kbps): 9 57 115 230 256 460 512 921 
available board names: generic(115k) esp8285(115k) espduino(115k) huzzah(115k) espresso_lite_v1(115k) espresso_lite_v2(115k) phoenix_v1(115k) phoenix_v2(115k) nodemcu(115k) nodemcuv2(115k) modwifi(115k) thing(115k) thingdev(115k) esp210(57k) d1_mini(921k) d1_mini_pro(921k) d1_mini_lite(921k) d1(921k) espino(115k) espinotee(115k) wifinfo(115k) arduino-esp8266(115k) gen4iod(115k) oak(921k) 
```

d-a-v:
All commit are squashed into a single one.

commit log:
```
boards.txt generator
+ generates boards.rst
+ replace boards in package.json
+ generate ldscripts
+ new debug option: OOM
+ led menu for generic board
```
Only boards.txt is updated (OOM debug option, led menu for the generic board and I included #4083)
I hope you like the usefulness of OOM debug option as much as I do :)

This needs review (the script itself only if you have too much time).
The OOM debug option spreads accross multiple files but is harmless when not enabled.

cores/esp8266/heap.c
```diff
@@ -32,18 +32,101 @@ void* _calloc_r(struct _reent* unused, size_t count, size_t size)
...
+
+#undef malloc
+#undef calloc
+#undef realloc
+
+void* malloc (size_t s)
+{
+    void* ret = umm_malloc(s);
+    if (!ret)
+        os_printf(":oom(%d)@?\n", (int)s);
```
devyte: Can the strings like this one ":oom()blah" be put in flash?

d-a-v:
all these strings are now moved to flash
```
:oom(46200)@/my/path/to/realloc.ino:31
:oom(5)@user_interface.c:111
```

devyte:
I think this has been reviewed enough. If there are further things to change, let's discuss it in a new PR.

PR #3730, opened 2017-10-17
Add time modification support to SPIFFS

luc-github:
Impact mkspiffs


igrr:
Thanks for the PR @luc-github. Do you know if filesystem created with SPIFFS_OBJ_META_LEN set to 0 will be compatible with SPIFFS code built with SPIFFS_OBJ_META_LEN set to 4? In other words, will this PR create an issue for those who already have SPIFFS created using an earlier version of the core?

luc-github:
Hi @igrr yes this PR will break backward compatiblity, need a reformat  and it affect mkspiffs - defined by SPIFFS owner and confirmed by my tests
  
My proposal was originaly for ESP32 (https://github.com/espressif/esp-idf/issues/1061) as no backward yet but as is affect mkspiffs and currently it is used for ESP8266 and ESP32 without discrimination but version ? (https://github.com/igrr/mkspiffs/issues/32)  

So after discussing with @me-no-dev, he suggested me to push all PR( ESP8266 / IDF / esp32-arduino, but not mkspiffs) to be a base of further discussion -  I think for ESP8266 this PR must come with a 3.0 release not in 2.X if it is even applied  but then if we want to apply to ESP32 need to add a compilation flag and a visual result in version to discriminat which mkspiffs is usable with ESP-  and which is for ESP32, as some people may use mkspiffs as standalone tool outside of core.
 
Current PR give visibility of needed changes in ESP8266.
I know you have a lot in your plate - this change is minor in code and break backward support but it also bring nice feature IMHO - sorry to bring more questions than it should

luc-github:
Any chance this could be implemented before 3.0 ?
This is one year old already 2.5.0 is not yet released and I see 2.6.0 is already planned, so 3.0 projected time would be in 2020 if no intermediate release is added.

Do not get me wrong, the maintainers do a great job but when I suggested 3.0, I was not expecting it would happen 2 years or more after pushing this PR .

earlephilhower:
Sorry 'bout that.  Since it changes the on-flash layout, how do you see this actually being integrated/swapped between?  Having people update their Arduino rev but lose all their files isn't a very good experience (and might mean this would have to be at 3.0 since it would be a breaking change).  Seems like SPIFFS and other embedded FSes traded saving 100 bytes of header info describing the on-flash format for allowing this kind of flexibility. :(

luc-github:
I know but I do have to ask, if 3.0 in 2020, 3.0 in 2020 will be (if SPIFFS is still used )
I was waiting this PR in ESP8266 to add this feature in my projects to keep ESP8266/ESP32 aligned in identical features, now I have confirmation it won't happen, before very long time - I will drop the feature for ESP8266
Thank you for your great work

d-a-v:
`SPIFFS_OBJ_META_LEN` could be switched to a variable, and with autodetection on mount.

luc-github:
Yes I adding a menu /option to set SPIFFS_OBJ_META_LEN would be great as keeping the compatibility I am not sure the impact on mount as SPIFFS_OBJ_META_LEN is already set
But in that case mkspiffs is actually the one that would need a big rewrite adding a flexible  SPIFFS_OBJ_META_LEN as parameter, if no parameter = 0

d-a-v:
Replaced by #6544

PR #3731, opened 2017-10-17
Put WiFi into station mode in examples

jes:
I was surprised that many of the examples implicitly create an
unsecured access point with complete access to my application, which
was based off the WiFiWebServer example. I'm sure I'm not the only one
surprised by this behaviour.

Putting the WiFi into station mode prevents the open access point from being
created.


jes:
I've added more commits by pushing to the master branch on jes/Arduino but they're not showing up in this pull request...

jes:
Thanks!

PR #3738, opened 2017-10-19
Bootloader OTA update

lugiann:
Is there a way to update the bootloader over the air? 
Every SDK release from espressif include a new version of the bootloader and in a production system that work with ota updates there are some circumstances that requires an update of the bootloader. 
A specific example is with the latest RC. You have updated the SDK to the 2.1.0 that include the v1.7 of the bootloader. We don't have a way that allow to update the bootloder over the air. In this cases the OTA is unusable because to update ALL the flash parts we need a cable connection to the ESP.


devyte:
Why is this a pull request? I would think this should be an issue with the feature request.

igrr:
Also, note that the bootloader from the SDK is not used in this project.

lugiann:
Yes...sorry for the pull...it was a mistake... 
Is true that the bootloader is not used in this project, but why not including the bootloader update as same as the firmware/spiffs...
Obvisously if the SDK have the API for this...i have tried to search information about but without success...if you can provide some info about this task i can try to add this feature to the currently update environment...

davisonja:
The flash (as opposed to rom) bootloader for Arduino (eboot) is already included in each compiled sketch. Each update therefore already updates the bootloader.

PR #3742, opened 2017-10-21
Increase version for PlatformIO

julian-w:
Just a small change in order to stay in sync with  the version numbering on PlatformIO


julian-w:
PlatformIO will check the git for a release/tag containing the version specified in the package.json and will download this version. 

So it would be nice to change this to the name of the current release so that PlatformIO will update to the current version and not download the old rc1-version.

PR #3750, opened 2017-10-23
Add WeMos D1 mini Lite in boards.txt

maxpromer:
from https://forum.wemos.cc/topic/274/how-to-flash-the-new-d1-mini-lite-with-arduino

Thanks !


igrr:
Thanks for the PR. D1 mini is already included, so closing this.
https://github.com/esp8266/Arduino/blob/master/boards.txt#L2096

PR #3769, opened 2017-10-28
Compressed dns

jonatanolofsson:
According to [(RFC 1035)](https://www.ietf.org/rfc/rfc1035.txt), clients must understand compressed dns messages. These commits aim to add the support of this to the ESP8266mDNS library.


libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -539,15 +539,24 @@ void MDNSResponder::_parsePacket(){
     while (numAnswers--) {
       // Read name
       stringsRead = 0;
+      size_t here = 0;
```
devyte: Please rename this into something more meaningful

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -632,31 +649,36 @@ void MDNSResponder::_parsePacket(){
         uint16_t answerPrio = _conn_read16(); // Read priority
         uint16_t answerWeight = _conn_read16(); // Read weight
         answerPort = _conn_read16(); // Read port
+        size_t here = 0;
```
devyte: Doesn't this declaration of here shadow the declaration on line 542?

jonatanolofsson:
Thank you for the feedback, I updated the code accordingly.

devyte:
Help needed with testing this!

jonatanolofsson:
The way I discovered this was missing was that I got errors with mDNS when connecting to a mac (which is compressing its data). Thus, to test this, set up a local domain with dns-sd with and without this patch and monitor the esp8266 errors on the serial port as it tries to connect (compile with DEBUG_ESP_MDNS_RX)

devyte:
@jonatanolofsson my wife has a mac. I hate the thing, but I could use it to test this, if you provide detailed instructions for the setup on it.

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -539,15 +539,24 @@ void MDNSResponder::_parsePacket(){
...
         tmp8 = _conn_read8();
-        if (tmp8 & 0xC0) { // Compressed pointer (not supported)
-          tmp8 = _conn_read8();
-          break;
-        }
         if (tmp8 == 0x00) { // End of name
           break;
         }
+        if (tmp8 & 0xC0) { // Compressed pointer
+          uint16_t offset = ((((uint16_t)tmp8) & 0xC0) << 8) | _conn_read8();
```
d-a-v: 1.
Would the resulting offset's boundaries be checked here ?

2.
I don't understand something. If the first byte is 0xca and the second 0xaa, then the offset would be 0xcaaa not 0xaaa 
 like with (((uint16_t)(tmp8 & ~0xC0)) << 8) | _conn_read8() ?

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -632,31 +649,45 @@ void MDNSResponder::_parsePacket(){
...
+        last_bufferpos = 0;
 
         (void) answerPrio;
         (void) answerWeight;
 
         // Read hostname
         tmp8 = _conn_read8();
-        if (tmp8 & 0xC0) { // Compressed pointer (not supported)
+        if (tmp8 & 0xC0) { // Compressed pointer
+          uint16_t offset = ((((uint16_t)tmp8) & 0xC0) << 8) | _conn_read8();
```
d-a-v: Is it &~0xc0 too here ?

d-a-v: Do you think the calculated offset can be checked beeing out of bound to prevent crashing with malformed or crafted dns replies ?

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -143,6 +144,21 @@ class UdpContext
...
+    }
+
+    void seek(const size_t pos)
+    {
+        assert(isValidOffset(pos));
+        _rx_buf_offset = pos;
+    }
+
+    bool isValidOffset(const size_t pos) const {
+        return (pos <= _rx_buf->len);
```
d-a-v: Thanks.
Since you are going to use this offset, would'nt it be ```pos < _rx_buf->len``` ?
Sorry I don't mean to be pushy :]

jonatanolofsson: Not at all, I appreciate your comments, I just havn't had much time to do more than fix them during breaks at work today =D
The check on line 222 in the 'read' function checks if the offset is at least equal to the limit, which should give that to set the buffer as "fully read", the offset should be set just past the limit, i.e. equal to the length. I figured "finished reading" was a valid use-case and thus a valid offset.

d-a-v: Thanks for the explanation

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -539,15 +539,34 @@ void MDNSResponder::_parsePacket(){
...
+          if (_conn->isValidOffset(offset)) {
+              last_bufferpos  = _conn->tell();
+#ifdef DEBUG_ESP_MDNS_RX
+              DEBUG_ESP_PORT.print("Compressed pointer, jumping from ");
+              DEBUG_ESP_PORT.print(last_bufferpos);
+              DEBUG_ESP_PORT.print(" to ");
+              DEBUG_ESP_PORT.println(offset);
+#endif
+              _conn->seek(offset);
+              tmp8 = _conn_read8();
```
d-a-v: Are you reading tmp8 twice (here and in L568) ?

d-a-v:
@jonatanolofsson can you tell us more about the steps to test your PR ? 
Wife also has a mac. Wives...

libraries/ESP8266mDNS/ESP8266mDNS.cpp
jonatanolofsson: Indeed, I forgot to remove the second one. Fixed.

jonatanolofsson:
I'll try to find time to make a test description. Not going to happen this week unfortunately. My computer is a mac, but I'm only ever booting into linux anymore :)

d-a-v:
Linux is no problem if you know how to update avahi's conffile so.

jonatanolofsson:
Right, I wasn't sure if avahi compressed the payload

devyte:
@jonatanolofsson I now understand your code changes, and I think they are correct. I am currently rewriting the mdns code based on the idf functionality, and was struggling a bit with compression (it's very different in idf due to fundamental differences in how things work at low level). This helps quite a bit. I'll merge this, and integrate it into my current work.

PR #3770, opened 2017-10-29
Remove forced OTA DNS announce

jonatanolofsson:
OTA should not force a DNS publication. This is not the job of the OTA server, and it interferes with previously configured DNS announcements.

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -127,12 +127,14 @@ void ArduinoOTAClass::begin() {
   if(!_udp_ota->listen(*IP_ADDR_ANY, _port))
     return;
   _udp_ota->onRx(std::bind(&ArduinoOTAClass::_onRx, this));
-  MDNS.begin(_hostname.c_str());
+  if(!MDNS.started()) {
+    MDNS.begin(_hostname.c_str());
 
-  if (_password.length()) {
-    MDNS.enableArduino(_port, true);
-  } else {
-    MDNS.enableArduino(_port);
+    if (_password.length()) {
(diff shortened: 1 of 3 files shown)


igrr:
I agree that we should avoid overriding previously configured mDNS. But as it is, the change will break existing sketches which did just ArduinoOTA.begin. From the perspective of the user who just wants to use OTA, but doesn't care about mDNS, ArduinoOTA should take care of mDNS registration.

How about adding a function to mDNS class to check if mDNS is configured, and then using it in ArduinoOTA? If not configured, then ArduinoOTA must call mDNS.begin, otherwise OTA will not work (Arduino IDE will not see the device on the network).

Also suggest keeping 'setHostname'/'getHostname' function for compatibility with existing sketches.

jonatanolofsson:
I rewrote the commit per your suggestion and added a check if the mDNS had already been started

jonatanolofsson:
@igrr I just rebased this. Should I do something more before it is merged?

igrr:
Thanks, the changes look good. This will be merged once 2.4.1 is released, which should happen very soon.

CZEMacLeod:
Would be better if we could have OTA enabled without mDNS at all.
Perhaps an optional `#define ArduinoOTA_No_mDNS` and some `#ifndef ArduinoOTA_No_mDNS`'s around the include and calls?
I have problems with mDNS crashing and causing other problems if it is started at all, so would prefer (for now at least) to leave it out altogether.
But this helps and is backwards compatible which is always good.

jonatanolofsson:
@igrr I just rebased this. Should I do something more before it is merged?

d-a-v:
duplicate #5494 which is already merged

PR #3783, opened 2017-11-01
lwIP-v2 ready to use

d-a-v:
boards.txt comes from #3722
lwIP sources are cloned from their origin, currently v2.0.3stable


devyte:
I'm now very confused with all the lwips flying around. Is there any way to extract the version info from the lib? Something like a #define, or version string, or git commit, or whatever, that can get printed out? I don't trust what I'm building or linking against anymore.

davisonja:
Worst case md5 is simple

d-a-v:
```
#ifndef LWIP_HASH_STR
#define LWIP_HASH_STR "esp1.4"
#endif
printf("lwIP-%d.%d.%d(%s)",
        LWIP_VERSION_MAJOR, 
        LWIP_VERSION_MINOR,
        LWIP_VERSION_REVISION,
        LWIP_HASH_STR);
```

```LWIP_HASH_STR``` is not defined with lwip1.4

With lwip2, it says: ```lwIP-2.0.3(92f23d6(tag:STABLE-2_0_3_RELEASE))```.
The hash number is the lwip commit ref, with the tag if there's one.

The two previous PR are now closed.

d-a-v:
@devyte @davisonja ```LWIP_VERSION_MAJOR``` is accessible in sketch with ```#include "lwip/init.h"```.
It was already the case with lwip1.4. What do you think should be further done about that ?
@igrr @devyte @davisonja are you OK with the included ```boards.txt``` ?
I have only tested Generic/nodemcu(ESP01), wemos D1R1, D1-mini and D1-mini-lite(with esp8285).

igrr:
boards.txt looks good, but i'll check with nodemcu. Will report back.

devyte:
Be aware that the ESPAsync libs don't seem to build with lwip2.
@me-no-dev at first glance, the changes needed seem minimal, but I didn't go through them in detail.

d-a-v:
@me-no-dev changes probably will be similar as #3129.

me-no-dev:
on it :)

devyte:
@me-no-dev ESPAsyncTCP builds now. Will you also update ESPAsyncUDP? That would allow me to build my application and start testing :)

devyte:
ESP8266LLMNR lib doesn't seem to build due to differences in UdpContext signatures

reaper7:
@devyte - ESPAsyncUDP also updated, works for me with LWIPv2

PR #3790, opened 2017-11-03
Allow specifying I2S buffer sizes and counts

earlephilhower:
I've been doing a lot of audio work but am finding that sometimes the ESP8266 internal WiFi code can go off and do something for quite a long while (relative to audio delays, that is) causing the internal DMA buffers to empty and hiccups in the output to occur.

The simplest way to avoid this is to increase the size and number of RAM buffers we can pre-fill to cover these WiFi stack times.  Granted, with only ~40K available you're not going to add seconds worth of buffering, but every little bit can help.

Here's a backwards compatible patch to allow an app to specify a larger buffer size/count if necessary, while preserving the original behavior (and saving a few bytes in everyone else's RAM).

---snip---
Add a new i2s_begin_custom() signature that takes a buffer count and buffer
size (in samples).  This allows for specifying a larger or smaller amount
of DMA data which can help cover periods when the ESP's internal WiFi
processing takes longer than the few milliseconds of data currently
allocated (i.e at 44.1KHZ, the original buffers would last for 512 samples
which works out to only 11.6 milliseconds).

This also frees up ~76 bytes of RAM in all applications that do not use
I2S since it dynamically allocates pointer/counter arrays and only has a
single static pointer defined.


cores/esp8266/core_esp8266_i2s.c
```diff
@@ -92,10 +95,16 @@ void ICACHE_FLASH_ATTR i2s_slc_isr(void) {
...
 
-void ICACHE_FLASH_ATTR i2s_slc_begin(){
+void ICACHE_FLASH_ATTR i2s_slc_begin(uint16_t buffers, uint16_t buffer_samples){
+  SLC_BUF_CNT = buffers;
+  SLC_BUF_LEN = buffer_samples;
+
   i2s_slc_queue_len = 0;
   int x, y;
   
+  i2s_slc_queue = (uint32_t*)malloc((SLC_BUF_CNT-1) * sizeof(uint32_t));
```
devyte: Is it correct that this malloc is 1 item smaller than the others? I realize this is how it is currently implemented, but I have to wonder.

devyte: Before the malloc is done, the pointer should be checked against null, and free()d if not.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -50,10 +53,10 @@ struct slc_queue_item {
   uint32  next_link_ptr;
 };
 
-static uint32_t i2s_slc_queue[SLC_BUF_CNT-1];
+static uint32_t *i2s_slc_queue;
```
devyte: These ptrs should be inited to null.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -148,8 +157,14 @@ void ICACHE_FLASH_ATTR i2s_slc_end(){
   SLCRXL &= ~(SLCRXLAM << SLCRXLA); // clear RX descriptor address
 
   for (int x = 0; x<SLC_BUF_CNT; x++) {
-	  free(i2s_slc_buf_pntr[x]);
+    free(i2s_slc_buf_pntr[x]);
```
devyte: The ptrs here should be nulled after free.

cores/esp8266/core_esp8266_i2s.c
earlephilhower: They are, actually.  In ANSI-C, static variables are set to all 0 and live in BSS, even if you don't explicitly list it.  I'm happy to add in the assignment, of course, to make it absolutely clear, of course.

cores/esp8266/core_esp8266_i2s.c
earlephilhower: SIze is -1 by design (see below, the original code only alloc'd buf_cnt-1).  Good catch on 2nd comment.  It's safe to free(NULL), so I'll always free before allocing.

cores/esp8266/core_esp8266_i2s.c
earlephilhower: Actually, scratch that.  i2s_slc_buf_pntr[] is a dynamically allocated array of pointers.  I free each pointer i2s_slc_buf_pntr[0...n], then completely free and null i2s_slc_buf_pntr.  So there is no way to get to those old pointers, ever, so no need to NULL.

cores/esp8266/core_esp8266_i2s.c
devyte: You probably don't want to free and malloc again if the call is done with the same size arg.

cores/esp8266/core_esp8266_i2s.c
devyte: What I am guarding against is a second call to begin() without a call to end(), I.e. make sure no leaks. This seems to be common usage in ArduinoLand.

earlephilhower: Arduinoland is a dangerous place to live...

It'll still (randomly, making it even more fun) crash w/o the changes I mentioned on the i2s_begin() call due to ISR and DMA working w/the old addresses.

earlephilhower:
I'm going to punt on this as it looks like I2S may be undergoing a larger change related to supporting input.  It's small enough to regenerate in the future if there is interest.

PR #3808, opened 2017-11-07
lwip2: stability and time management

### LINKED ISSUE #1679: time() does not work without active NTP server
#### Hardware

Hardware:           ESP-12E
Core Version:       2.1.0-rc2
### Description

"Internal" time.c functions always return 0 /1970-01-01/ if not synced via NTP.

I want to initialize internal clock via constant or from RTC for log timestamping (so I don't need precise time on  startup, but valid measured intervals)
### Settings in IDE

Flash Size:         4MB
CPU Frequency:      80Mhz
Flash Mode:         qio
Flash Frequency:    80Mhz
Upload Using:        SERIAL
Reset Method:       ck
### Sketch

``` cpp

#include <Arduino.h>
#include <time.h>

void setup() {
 // WiFi.begin(); // wifi disabled for example
 // configTime(3*3600, 0, "not.avail.ntp"); 
 Serial.begin(115200);
}

void loop() {
  time_t now = time(nullptr);
 Serial.println(ctime(&now));
 delay(1000);
}
```
### Debug Messages

```
Thu Jan 01 00:00:00 1970

Thu Jan 01 00:00:00 1970

...
```
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31264128-time-does-not-work-without-active-ntp-server?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2505: configTime and time seem to ignore DST
### Basic Infos

configTime and time seem to ignore DST
#### Hardware

Hardware:           ESP-12
Core Version:       latest master
### Description

doing a config time, with DST offset does not seem to be actually doing anything to returned time. Time zone offset applies alone.
### Sketch

``` cpp

#include <Arduino.h>

void setup() {
  configTime(2 * 3600, 1 * 3600, "pool.ntp.org", "time.nist.gov");
  delay(2000);
  time_t now = time(nullptr);
  Serial.println(ctime(&now));
}

void loop() {

}
```
### Debug Messages

Expected GMT + 3 to be printed. It actually prints GMT + 2

Cheers

d-a-v:
sync with upstream lwip2:
expose sntp_set_system_time() for external RTC without NTP server
fix #1679 (time() works from start)
fix #2505 (honour DST in configTime())
fix lwip2 crashing with ```WiFi.softAPConfig(ip,ip,ip)```


d-a-v:
I found nowhere in the docs where ```configTime()``` is mentionned.
I would be worth mentionning ```sntp_set_system_time()``` which allows to use an external RTC.
This function has always been existing but was not exposed.

Current version of lwipopts (v2) allows only one ntp server. Should it allows 3 servers (as per ```configTime()```) ?

Also, timezone and DST are only taken into account when ```sntp_set_system_time()``` is called, which can happen:
- by user
- by configTime() (but takes effect once an IP address is setup)
- by dhcp client (if dhcp server provides ntp server)

Currently the API does not let the user decide to refuse the ntp server provided by dhcp (ntp was not fed by dhcp with lwip1.4).

Juppit:
[configTime()](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/time.c) filles only the structures in [sntp.h](https://github.com/esp8266/Arduino/blob/master/tools/sdk/include/sntp.h) 
Note that sntp_set_daylight() is still not in [sntp.h](https://github.com/esp8266/Arduino/blob/master/tools/sdk/include/sntp.h) where it supposed to be.

d-a-v:
You are perfectly right about ```sntp_set_daylight()``` missing declaration. I just realized that we have also ```Wno-implicit-function-declaration``` in gcc options. We would then have to declare ```millis()```, ```micros()``` too. I think this should go to another PR.
This PR is all about lwip2 currently only in master. ```configTime()``` calls ```sntp_set_daylight()``` which is implemented [there](https://github.com/d-a-v/esp82xx-nonos-linklayer/blob/master/glue-lwip/esp-sntp.c#L369).

PR #3818, opened 2017-11-10
Added micros64 and used to fix and improve gettimeofday.

### LINKED ISSUE #3814: gettimeofday incorrect value for tv_usec
The current implementation _gettimeofday_r is incorrectly setting the tv_usec field.

https://github.com/esp8266/Arduino/blob/master/cores/esp8266/time.c#L104

According to http://www.gnu.org/software/libc/manual/html_node/Elapsed-Time.html

tv_usec should be "This is the rest of the elapsed time (a fraction of a second), represented as the number of microseconds. It is always less than one million."

The current code is simply doing `tp->tv_usec = micros();`.  This leads to the wall clock time advancing forward too quickly.  I am certainly aware of the caveats of tracking wallclock (ie CLOCK_REALTIME) in ESP8266, but this is just making it far worse.

I have some ideas on how to square this away once and for all (crossing my fingers).  I will try to follow-up with additional info or possibly a pull request.

BrandonLWhite:
This resolves #3814 and adds a `micros64()` global function that is very useful.

There is certainly more that can be done to improve wallclock time-keeping, but this is a step in the right direction I feel.  This allows my application to have a reasonably decent current timestamp.

This now works well enough for me, when it previously was wrong:
```
auto now_ms = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
```

I purposely did not touch `clock_gettime()` or `time()` because I want to focus on achieving a functional `gettimeofday` first.  If this PR is accepted, I'm willing to work on those functions too.


BrandonLWhite:
Whoops, this might be bugged.  I need to do some more testing, then I'll be back!

PR #3821, opened 2017-11-12
Fix ESP8266LLMNR for build with lwip2

devyte:
Original fix by @d-a-v  (thanks!)


devyte:
@igrr I think this is ok to merge, assuming you have no comments.
BTW, am _I_ allowed to merge? I am presented with the option, but I haven't wanted to do anything so far, mostly because I've never done it :)

PR #3835, opened 2017-11-15
lwip2 fixes and time/ntp management

d-a-v:
#1679: sntp is automatically started 
#2505: honor DST in configTime()
fix clock_gettime() with micros64()
add settimeofday()
new example esp8266/NTP-TZ-DST.ino
--- lwip2 (mandatory for the above):
the above fixes
remove old and unwanted definitions of millis() and gettimeofday()
fix crashing with WiFi.softAPConfig(ip,ip,ip)


cores/esp8266/time.c
```diff
@@ -73,17 +72,35 @@ void configTime(int timezone, int daylightOffset_sec, const char* server1, const
     setServer(1, server2);
     setServer(2, server3);
 
-    s_timezone_sec = timezone;
-    s_daylightOffset_sec = daylightOffset_sec;
+    //s_timezone_sec = timezone;
+    //s_daylightOffset_sec = daylightOffset_sec;
```
igrr: Since the variables have been removed, can these lines be removed as well?

libraries/esp8266/examples/NTP-TZ-DST/NTP-TZ-DST.ino
```diff
@@ -0,0 +1,96 @@
+
```
igrr: Need a header here explaining what the example does, and the if/endif choices below. Also should include a public domain or cc0 license (like the original Arduino examples).

igrr: The file also needs to be formatted same way as other sketches. Simple way to do this is to load it into arduino, then use Tools > Format sketch.

d-a-v:
Correct: localtime() and gmtime() return both local time with isdst=0 (hence both wrong).

So there are currently three bugs:
* localtime() should set isdst to 1
* gmtime() should show UTC
* gettimeofday() is wrong when called the first time before ntp date is set

The example is updated with header, explanation, format and localtime/gmtime tests.

devyte:
@d-a-v I updated the branch and restarted travis, but it still fails. The reason seems to be a linker error about undefined reference settimeofday.

d-a-v:
It's because travis links against lwip1.4 where ```settimeofday()``` is not defined.
Should we add a wrapper in lwip1.4 for the sake of coherency ?

d-a-v:
I'm afraid we can't update the internal timestamp *without* modifying lwip1.4 (every useful stuff is static in ```src/core/sntp.c```, timestamp can only be setup by ntp).

We can end up with:
- no ```settimeofday()``` (travis should be teached about lwip2)
- ```settimeofday()``` doing nothing
- ```settimeofday()``` yelling

everslick:
I have an ``` `undefined reference to `ping_start'` ``` when linking against LWIP2

d-a-v:
Right. ping is in esp-lwip-v1.4's apps but not in current lwip's apps.
It is however in lwip's [contrib](http://git.savannah.gnu.org/cgit/lwip/lwip-contrib.git/tree/apps/ping/).
It's two files to compile.
@igrr @devyte should it be in lwip2 ?

d-a-v:
@igrr PR 3835 is squashed into a single commit

everslick:
Interestingly ping_option and ping_resp are declared correctly.

devyte:
@d-a-v my only feedback about the ping files: try to keep the lwip directory as clean as possible, so that we have fewer headaches in the future if we ever need to do another migration. In other words, all files not distributed with lwip should probably be put elsewhere.

d-a-v:
@devyte ping may have been distributed in lwip-v1, now it is in their lwip-contrib.
Still, the fonctions are declared in espressif's SDK (```tools/sdk/include/ping.h```).
@everslick please make a new issue for this so further discussion happen there.

PR #3836, opened 2017-11-15
Added constant time string comparison to avoid possible time-based attacks.

### LINKED ISSUE #1127: Timing attack in OTA authentication
Since the comparison function of the secret for OTA authentication is not constant-time, it is possible for an attacker to discover the OTA secret over the network and flash whatever firmware they like.

The following article has an explanation and fix:

http://blog.ircmaxell.com/2014/11/its-all-about-time.html
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28762390-timing-attack-in-ota-authentication?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bluemurder:
This should resolve #1127.


libraries/ArduinoOTA/ArduinoOTA.h
```diff
@@ -93,6 +93,7 @@ class ArduinoOTAClass
     void _onRx(void);
     int parseInt(void);
     String readStringUntil(char end);
+	bool constantTimeEquals(const String &string1, const String &string2);
```
igrr: Suggestion: can this method be moved into String class, so reuse in other places is possible?

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -358,6 +358,27 @@ int ArduinoOTAClass::getCommand() {
...
 }
 
+// To avoid possible time-based attacks present function compares given 
+// strings in a constant time.
+bool ArduinoOTAClass::constantTimeEquals(const String & string1, 
+  const String & string2) {
+
+  // Preliminary check  
+  if(string1.length() != string2.length()) {
+	  return false;
```
igrr: Indentation

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -358,6 +358,27 @@ int ArduinoOTAClass::getCommand() {
...
+  }
+  
+  // Evaluates every character
+  bool equals = true;
+  unsigned int len = string1.length();
+  for(unsigned int i = 0; i < len; i++){
+    equals &= (string1.charAt(i) == string2.charAt(i));
```
igrr: It would be interesting to see the generated code for this loop at -O2 optimization level... I wonder whether the compiler is smart enough to convert this 'and' into 'branch-if-zero'.

bluemurder: I agree. Maybe better to write not so easily optimizable code. Ill think about that

libraries/ArduinoOTA/ArduinoOTA.cpp
bluemurder: fixed

libraries/ArduinoOTA/ArduinoOTA.h
bluemurder: Ok I'll do that

bluemurder:
@igrr tried to met all your requests

devyte:
Except for my two nitpick cleanups, this looks good to me.

devyte:
This should still be tested more thoroughly, but after careful inspection, the code looks good, so merging.

Issue #2188, opened 2016-06-24
ESP8266WiFiMulti.cpp code inspection: Unexecutable code in ESP8266WiFiMulti::run


SteveToulouse:
While looking through the code to understand what's being done and see how issue #2117 could be resolved, [lines 51 to 63](https://github.com/esp8266/Arduino/blob/633e48f3aec5f1c3c11d4498fc90d378d49e6e9f/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp#L51-L63) caught my eye.

The conditions of the "if"s at lines L51 `if (scanResult > 0)` and L61 `if (scanResult <= 0)` are mutually exclusive. So the code at line 62 can never be executed. Unless I've missed something?

Not a bug as such (it doesn't break anything) but it adds complexity when reading the code. The enormous "else" clause, from 63 onwards to nearly the end of `::run`, could be reduced one indentation level.

EDIT This was introduced by the commit https://github.com/esp8266/Arduino/commit/9a1ff7f70dab5b948e4cd48f90ddc0fba5a17be8

devyte:
Code reference is still present in latest git. Most likely this
`if (scanResult <= 0)
`
should be
`if (scanResult == 0)`

devyte:
@SteveToulouse please take a look at the referenced PR.

devyte:
I merged the PR after some sanity checks. Any new problems will be handled in a different issue.
Closing.

PR #3847, opened 2017-11-19
Refactor of WiFiMulti::run to reduce complexity and remove dead code (#2188)

devyte:
@igrr I'm starting to look at the issues labeled "easy" with milestone 2.4.0, and this is my first one. In this case, the refactoring was not so minor, so I think a review is in order. Who could review?


devyte:
PR is for #2188 .

devyte:
I did some sanity tests, new code seems to work as expected. Merging. Any issues will be handled separately.

PR #3849, opened 2017-11-19
EEPROM: dirty flag, reallocate only if necessary, add getConstDataPtr method (#2217)

### LINKED ISSUE #2217: EEPROMClass _dirty
#### Hardware

Hardware:       n.a.
Core Version:       2.2.0
### Description

1:
There is no way to undo/revert any changes after a begin.
Currently the only way would be to call begin again, but... See 2:

I suggest adding:

```
bool EEPROMClass::revert() {
  if (_data && _size)
  {
    noInterrupts();
    spi_flash_read(_sector * SPI_FLASH_SEC_SIZE, reinterpret_cast<uint32_t*>(_data), _size);
    _dirty = false;
    interrupts();
    return true;
  }

  return false;
}

```

2: 
`void EEPROMClass::begin(...)` Does read _data from eeprom but does not clear the _dirty flag.
Also it would be handy if `EEPROMClass::begin(...)` would return the _data pointer.

I suggest:

```
uint8_t * EEPROMClass::begin(size_t size /* =  SPI_FLASH_SEC_SIZE */) {
  if (_data) {
    delete[] _data;
    _data = NULL;
  }
  _size = 0;
  _dirty = false;

  if ( (size > 0) && (size <= SPI_FLASH_SEC_SIZE) ) {
     size = (size + 3) & (~3); 
    _data = new uint8_t[size];
    if ( _data)  {
      _size = size;
      noInterrupts();
      spi_flash_read(_sector * SPI_FLASH_SEC_SIZE, reinterpret_cast<uint32_t*>(_data), _size);
      interrupts();
    }
  }
  return &_data[0];
}

```

3:
`EEPROMClass::getDataPtr()` Does always set _dirty. This is not wanted if we are using the pointer only for reading....

I suggest: 

```
uint8_t * EEPROMClass::getDataPtr(bool forWrite /* = true */) {
  if (forWrite) _dirty = true;
  return &_data[0];
}
```

with the default forWrite = true `getDataPtr()` would still be compatible with the current version but when just reading `getDataPtr(false)` would not set _dirty to true

devyte:
Fixes #2217 .

### DIFF
--- a/libraries/EEPROM/EEPROM.cpp
+++ b/libraries/EEPROM/EEPROM.cpp
@@ -56,16 +56,21 @@ void EEPROMClass::begin(size_t size) {
 
   size = (size + 3) & (~3);
 
-  if (_data) {
+  //In case begin() is called a 2nd+ time, don't reallocate if size is the same
+  if(_data && size != _size) {
     delete[] _data;
+    _data = new uint8_t[size];
+  } else if(!_data) {
+    _data = new uint8_t[size];
   }
 
-  _data = new uint8_t[size];
   _size = size;
 
   noInterrupts();
   spi_flash_read(_sector * SPI_FLASH_SEC_SIZE, reinterpret_cast<uint32_t*>(_data), _size);
   interrupts();
+
+  _dirty = false; //make sure dirty is cleared in case begin() is called 2nd+ time
 }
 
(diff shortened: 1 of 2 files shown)


devyte:
@igrr who could review this one?

PR #3851, opened 2017-11-19
Fix too wide 'word' type -- it should be 16-bit instead of 32-bit

Mystic-Mirage:
The _word_ type is an alias for _unsigned int_. It came directly from original Arduino SDK because ATmega-based boards has 16-bit int. But int for esp8266 board stores 32-bit value so _word_ in meaning _combination of two bytes_ is broken here.

For esp8266 it should be defined 16-bit long explicitly.


devyte:
I don't see this really being used anywhere in the code hosted in this repo, and the fix is correct.

Issue #2183, opened 2016-06-23
Recent update (2.3.0) gives some warnings


dandandandandandandan:
#### Hardware

Hardware:           NodeMCU 0.9
Core Version:       2.3.0
### Description

During verbose compilation in the Arduino IDE the new 2.3.0 version gives me some warnings
 ### Settings in IDE

Module:             NodeMCU 0.9
Flash Size:         4MB/3MB
CPU Frequency:      80Mhz
Upload Using:   SERIAL
Reset Method:       nodemcu
### Debug Messages

esp8266\hardware\esp8266\2.3.0\libraries\ESP8266WiFi\src\WiFiClientSecure.cpp: In function 'int ax_port_read(int, uint8_t*, size_t)':

hardware\esp8266\2.3.0\libraries\ESP8266WiFi\src\WiFiClientSecure.cpp:539:53: warning: suggest parentheses around '&&' within '||' [-Wparentheses]

```
 if (!_client || _client->state() != ESTABLISHED && !_client->getSize()) {

                                                 ^
```

\packages\esp8266\hardware\esp8266\2.3.0\libraries\ESP8266HTTPClient\src\ESP8266HTTPClient.cpp: In member function 'bool HTTPClient::beginInternal(String, const char*)':

\esp8266\hardware\esp8266\2.3.0\libraries\ESP8266HTTPClient\src\ESP8266HTTPClient.cpp:135:10: warning: unused variable 'hasPort' [-Wunused-variable]

```
 bool hasPort = false;

      ^
```

\Arduino15\packages\esp8266\hardware\esp8266\2.3.0\cores\esp8266\Schedule.cpp:17:13: warning: 'void init_lists()' defined but not used [-Wunused-function]

 static void init_lists()

```
         ^
```

juansta:
In addition to the above I am also seeing additional warnings on compilation

```
.arduino15/packages/esp8266/hardware/esp8266/2.3.0/libraries/ESP8266mDNS/ESP8266mDNS.cpp: In member function 'void MDNSResponder::_parsePacket()':
.arduino15/packages/esp8266/hardware/esp8266/2.3.0/libraries/ESP8266mDNS/ESP8266mDNS.cpp:590:18: warning: unused variable 'answerPrio' [-Wunused-variable]
uint16_t answerPrio = _conn_read16(); // Read priority
^
[...]

PR #3854, opened 2017-11-20
Removed unused init_lists() function that serves no purpose

devyte:
Removed init_lists(). It was already marked as unused, so it doesn't generate a warning, but in truth it serves no purpose. The function pre-inits to 4 empty function slots, but the linked-list algorithms work without doing that init, and pre-initing that way has no performance advantage either.
If the function is ever needed in the future, for whatever reason, it can be retrieved from the file history.


devyte:
Finalizes #2183 .

PR #3856, opened 2017-11-21
Removed unneeded undef min|max, they no longer clash with vector et al

devyte:


devyte:
@igrr I need a bit of help with this one. I wanted to remove two remaining #undef min|max left over from before the Arduino.h fix submitted some days ago (see top commit above), but Travis failed when building tests. To my surprise, I found another Arduino.h in there, which has fallen behind vs. the core one. I tried bringing it somewhat to par with the current Arduino.h, specifically with the min|max fix, but the build still fails, this time complaining about isnan etc already defined.
Oddly, if I run make under tests/host locally in my Ubuntu machine, it builds fine. 
At this point, I suspect an issue with the version of gcc. Travis builds with 4.8.4, while my linux box has 5.4.0.
How can I remove the two #undef min|max without breaking the build?

devyte:
@igrr ping
Not critical for 2.4.0, but I'd love to get this cleanup in.

devyte:
Alright, it looks like there is no point to pursue this under gcc 4.x. From what I can gather, the whole isnan/isinf thing was known under gcc >= 4.6, and a patch was applied at some point later (I couldn't figure out the details). Probably gcc 5.x has it.
So there isn't much point in pursuing this one until after we migrate to a newer gcc.

earlephilhower:
To 3.0.0 as we'll probably be on GCC 9 for that release and will be able to use this.

Issue #1168, opened 2015-12-07
WiFi.config signature differs from Arduino docs


FWeinb:
The [official Arduino docs](https://www.arduino.cc/en/Reference/WiFiConfig) state that the WiFi.config signature should look like this:

```
WiFi.config(ip); 
WiFi.config(ip, dns); 
WiFi.config(ip, dns, gateway); 
WiFi.config(ip, dns, gateway, subnet); 
```

But for the esp8266 lib it is implemented like:

```
WiFi.config(ip);
WiFi.config(ip, gateway, subnet, dns);
```

See [ESP8266WiFi.h](https://github.com/esp8266/Arduino/blob/40da463ee631133a16fc8ddde7c82d507060068b/libraries/ESP8266WiFi/src/ESP8266WiFi.h#L95-L104)

This is a little confusing.
##

Links2004:
dns support where implemented later for ESP.
since every one already using 
```WiFi.config(ip, gateway, subnet);```
at this point we add it to the end, or all the existing code where broken.

but yes its can be confusing.

FWeinb:
Maybe in an major update this could be done. Just to smooth out some edges. Great work so far, really delighted with this amazing project.

marvinroger:
Btw, is the project following semver?

ohosrry:
I also found this, this troubled me for a long time.can anybody fix it?

Links2004:
there is no real "fix" since if we change now, we break all existing code on every ESP.
one of two cases will be always broken.

ohosrry:
how about a "version macro" ? like:
`#ifdefine _V1_
    //arg1,arg2...
#else
    //arg2,arg1...
#endif`

Links2004:
if you define something in the ino file the define not exists in any other cpp file.
so its not that easy.
the only way to use defines for all files are setting in the plattform.txt or over the menu (like for debug)
but that not a good way to do it for this problem.

ohosrry:
whatever,you will change the function's declaration sooner or later.
you must make it identical with arduino docs,because users will be more and more.
btw,everyone comes here is for convenience,but this problem is very inconvenient. especially to new guy.
if code dose not work,we don't know what's wrong,because there is no diagnostic trace about gateway and netmask.

Links2004:
i think that is something we can change when we go to version 3.0.0,
 since with this version will have many changes in the core and will not be full compatible in all cases anyway.

igrr:
[...]

PR #3860, opened 2017-11-22
Add compatibility check for ::config() method with arduino arg order (#1168)

devyte:
Issue #1168  .


devyte:
@igrr is this what you had in mind?

PR #3861, opened 2017-11-22
Updated doc with Serial.setRxBufferSize method description

bluemurder:


bluemurder:
I close because Github included too olc unrelated commits

PR #3870, opened 2017-11-23
Collect all headers of incoming web server request (and don't just drop them)

OlexOlex:
Fixed missing functionality to collect headers. Headers would only be updated if the name property was already known in an entry of the according array.
Now the array of headers gets replaced by a larger array when a previously unknown header name is passed to _collectHeader. In this case the function returns false as before, but the header is saved to the array.
If the passed header name is the same as an already existing entry, the value of the existing entry still gets overwritten and the function returns true, nothing changed here.

This fix is simple, short and tested with the "productive" branch used by the Arduino IDE (but it seems like since this version nothing changed relating to this part of the code)


igrr:
@me-no-dev please review

PR #3872, opened 2017-11-24
Create non-blocking socket with WiFiClientSecure

zipiju:
SSL_READ_BLOCKING was causing the code to get stuck on functions using
SSL socket (available(), connected() etc.) when connection to the server
suddenly broke (in a manner that ESP still was associated to the AP and had an IP).
As this was not really a stuck CPU WDT never fired.


marcelowajnsztok:
This correction didnt solved the problem, the code still working, but cant connect to the server.

marcelowajnsztok:
It is a good work around.

earlephilhower:
Gonna close this due to axtls being deprecated now.  If there is a similar issue w/BearSSL, please open a new issue and we can look at ways to make it happen.  Thx!

PR #3877, opened 2017-11-24
Added Chip ID into HTTP header in ESP8266HTTPUpdate lib

c0degeekPL:
Added a chip ID into HTTP header.


earlephilhower:
Seems trivial and non-breaking, but no reason described why you'd want to use this change.

We should either merge for 2.6 or close as won't add.

PR #3892, opened 2017-11-29
Fixed verifyCert

timkoers:
This is compiling properly now, sorry for the previous pull requests.


igrr:
I think verifyCert should not be exposed as a user-facing API, as it doesn't validate the domain name.

PR #3914, opened 2017-12-05
Add board wifi slot

acosinwork:
Added new board WiFi Slot with esp12-f.
Description: http://wiki.amperka.ru/%D0%BF%D1%80%D0%BE%D0%B4%D1%83%D0%BA%D1%82%D1%8B:wifi-slot


d-a-v:
Check #3722. New boards in boards.txt will be integrated through a script. In order to add your board, and what is not clear enough in your files, 
* is the flashing method, is it dout/qio/dio or qout ?  (or should we leave all of them with dout as default)
* is the flash frequency 40 or 80Mhz ?
* serial speed default is 115k, but would you prefer 921k as default ?

PR #3916, opened 2017-12-05
add board WiFi Slot

acosinwork:
Added new board WiFi Slot with esp12-f.
Description: http://wiki.amperka.ru/%D0%BF%D1%80%D0%BE%D0%B4%D1%83%D0%BA%D1%82%D1%8B:wifi-slot


d-a-v:
Check #3722. New boards in boards.txt will be integrated through a script. In order to add your board, and what is not clear enough in your files, 
* is the flashing method, is it dout/qio/dio or qout ?  (or should we leave all of them with dout as default)
* is the flash frequency 40 or 80Mhz ?
* serial speed default is 115k, but would you prefer 921k as default ?

acosinwork:
Hi, thank you.
Only now I found out that our module easily supports these modes.

- we can set qio as as default. But we must leave all of them accessible.
- 80 Mhz as default will be great. But I would leave the opportunity to choose 40 Mhz
- 921k as default - ok. And leave 115200 possible for a choice 

Sorry, I did not really understand. Do I have to change the boards.txt now?

d-a-v:
Boards.txt will be generated by a script. Boards are [entries](https://github.com/d-a-v/Arduino/blob/34e6738fb84f0d8b1c75b7e6e006c340a7d1bedd/tools/boards.txt.py#L53) in an array
If you leave all options as user-definable, then what is the difference with the generic board ?

acosinwork:
For this board is important that reset method is nodemcu, custom .build.board (wifi_slot in variants) and fash size is more than 1M. I really liked your suggestions for the default settings. But I can not be sure that all the family esp12 will be stable in the future

acosinwork:
Should I change the script in your branch?

d-a-v:
Your new board brings more changes than a simple new board declaration.
So declaring it in the script in it will make it half integrated and not usable.
I apologize I came too fast to you and did not look at your other changes.
So if your changes are integrated before the generator, then I will update it.
If the generator is integrated before, then you'll have to adapt your pull request.

variants/wifi_slot/pins_arduino.h
```diff
@@ -0,0 +1,125 @@
...
+static const uint8_t A0 = PIN_A0;
+static const uint8_t A1 = PIN_A1;
+static const uint8_t A2 = PIN_A2;
+static const uint8_t A3 = PIN_A3;
+static const uint8_t A4 = PIN_A4;
+static const uint8_t A5 = PIN_A5;
+static const uint8_t A6 = PIN_A6;
+static const uint8_t A7 = PIN_A7;
+
+static const uint8_t LED_BUILTIN = 2;
```
d-a-v: please use a `#define LED_BUILTIN` like in the other variants and get rid of `BUILTIN_LED`, it is defined in common.h that you should use to avoid duplicate code.

Is that fair ?

acosinwork: Thanks for your feedback.
I'll fix the `#define LED_BUILTIN`. Unfortunately, I can not just connect `common.h`, because the pinout of our board conflicts with this:
https://github.com/esp8266/Arduino/blob/1a9403df1b8aa06430ac7480c0a1f9729eaa987c/variants/generic/common.h#L50-L52

I could change the `common.h` in this part to this code:
```
#ifndef PIN_A0
#define PIN_A0 (17)
#endif

static const uint8_t A0 = PIN_A0;
```
but will you accept this change?

d-a-v: I don't see any problem with this proposal,
as long as the API remains compatible with the Arduino use of `analogRead(A0)`
which seems to be the case in your PR and your new analog pins and I guess a analog multiplexer.

acosinwork:
@d-a-v travis build faild, but i do not understand why. It feels like it's not in my code. Could you help me?

acosinwork:
@d-a-v , it seems that the code has magically passed the tests. Could you look at this please?

PR #3919, opened 2017-12-05
Update NTPClient.ino

Danara5:
When testing NTPClient on many devices on a home gateway, the router would only forward NTP responses to the first device to call NTP using that port.   By giving each device a random port, the NTP responses are forwarded to the correct internal IP.  The UDP port conflict may not be evident on all NAT routers, and could depend on the timing between NTP calls and how long the router keeps the  entry in the port forward table.  Randomizing the port used minimizes the chance of a conflict with other devices.


themindfactory:
Good man!
:-)

devyte:
I think this code change is not viable, because the result of the random call could conflict with a whole range of local services, e.g.: mdns on port 5353. 
In addition, if it's true that having multiple ESPs with the same localport can cause the router NAT to mess up, this won't fix the issue, it would only mitigate it, as multiple ESPs could still end up with the same port. It would be unlikely, but still possible, and if it should happen, the failure would be spurious.

earlephilhower:
Closing as @devyte makes a valid point about this stomping on local services (and being 2 yrs+ old).  If another maintainer feels otherwise, feel free to reopen.

Issue #3212, opened 2017-05-07
ESP8266mDNS - Add support for overwriting/removing TXT records


WeeJeWel:
I advertise a device's name on mDNS using `MDNS.addServiceTxt("myservice", "tcp", "name", "Foo");`, but this name can change over time by user input. When calling the method again, another TXT entry is added, but I expected the previous one to be replaced.

igrr:
@swarren and @ccooper21 have contributed to mDNS library recently. Maybe they would be willing to implement this?

Alternatively, you can get MDNS object into a clean state like this:
```c++
MDNS = MDNSResponder();
```
which should delete the old one, freeing all resources, and create a new one. Now if this doesn't work for some reason, feel free to open a new issue :)

mkillewald:
I realize this is probably low priority, but I would like to see support for being able to change or remove TXT records as well.  I could not get the suggestion of `MDNS = MDNSResponder()` to work. It clears the MDNS object but then I can't seem to re-initialize the object with either `MDNS.begin()`, `MDNS.setInstanceName()`, or `MDNS.addService()`.

romuye123:
@mkillewald 
Thanks for the UPDATE. But when I keep monitoring the updates on my Android app, the updates are not realtime / immediate. 

```
  MDNS.clearServiceTxt("http", "tcp");
  MDNS.addServiceTxt("http", "tcp", "name",myname); 
```
Please suggest if I am wrong or let me know how can I instantly update my client mobile app.

mkillewald:
@romuye123 
Due to the nature of mDNS the updates are not going to occur immediately. As a work around for this in my mobile clients, I've made a refresh action that drops all discovered nodes and re-discovers them causing the changes to be shown. Hope that helps!

devyte:
Closing in view of #5442 with a full rewrite of mdns.

PR #3936, opened 2017-12-08
ESP8266mDNS - Add support for removing TXT records. Issue #3212

mkillewald:
I have added a method to clear the TXT record as a whole. This will clear all key, value pairs and allow the entire TXT record to be rebuilt as needed.

### DIFF
--- a/libraries/ESP8266mDNS/ESP8266mDNS.cpp
+++ b/libraries/ESP8266mDNS/ESP8266mDNS.cpp
@@ -268,8 +268,20 @@ void MDNSResponder::clearServiceTxt(const char *name, const char *proto){
     //Checking Service names
     if(strcmp(servicePtr->_name, name) == 0 && strcmp(servicePtr->_proto, proto) == 0){
       //found a service name match
-      servicePtr->_txts = 0;
-      servicePtr->_txtLen = 0;
+      if (servicePtr->_txts != 0) {
+      	MDNSTxt * txtPtr = servicePtr->_txts;
+      	MDNSTxt * nextPtr = txtPtr->_next;
+      	delete txtPtr;
+      	txtPtr = 0;
+      	while (nextPtr !=0) {
+      		txtPtr = nextPtr;
+      		nextPtr = txtPtr->_next;
+      		delete txtPtr;
+      		txtPtr = 0;
+      	} 
+      	servicePtr->_txts = 0;
+      	servicePtr->_txtLen = 0;
+      }
     }
   }
 }
--- a/libraries/ESP8266mDNS/ESP8266mDNS.cpp
+++ b/libraries/ESP8266mDNS/ESP8266mDNS.cpp
@@ -261,7 +261,7 @@ bool MDNSResponder::addServiceTxt(char *name, char *proto, char *key, char *valu
   return false;
 }
 
-void MDNSResponder::clearServiceTxt(char *name, char *proto){
+void MDNSResponder::clearServiceTxt(const char *name, const char *proto){
 	MDNSService* servicePtr;
 	//Find the service
   for (servicePtr = _services; servicePtr; servicePtr = servicePtr->_next) {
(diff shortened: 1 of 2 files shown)


libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -261,6 +261,19 @@ bool MDNSResponder::addServiceTxt(char *name, char *proto, char *key, char *valu
   return false;
 }
 
+void MDNSResponder::clearServiceTxt(char *name, char *proto){
```
d-a-v: why don't you use "const char*" here ?

mkillewald: Good question. Mainly because I'm a newb and I was basing it off the structure of addServiceTxt.  I will change it to const char*

mkillewald:
Not sure what I do about the errored check. Is this because I merged master back into my branch?

d-a-v:
Travis CI is exhausted:
> The job exceeded the maximum time limit for jobs, and has been terminated.

check ```Details```, not your fault :)

devyte:
In any case, you should fix the mem leak before CI gets restarted.

mkillewald:
@devyte please excuse my newb status, but please help me understand what is leaking and possibly how I should fix it.  Is the leak from where I set the _txts pointer to 0?

devyte:
in MDNSResponder::addServiceTxt(...) , this:
servicePtr->_txts = new ....
So there is a mem allocation with new, and the pointer isn saved in _txts. If you then do this in your code:
servicePtr->_txts = 0;
What happens to the allocated mem? You lose the pointer, so it can no longer be recovered via a delete, hence it leaks.
You seem to be new to C/C++. I suggest studying mem allocation/deallocation. In C it's usually done with malloc()/free(). In C++ those work as well, but in general special keywords new/delete are used, because they include handling for object constructor/destructors.
Keep in mind that in C/C++ there is no garbage collection, so you can't just lose pointers to allocated mem and expect that mem to magically be recovered. You have to handle the mem yourself.

mkillewald:
Definitely new to C and C++ but always eager to learn. Thank you for the quick lesson. I will take a stab at fixing the leak.

mkillewald:
Hopefully, this fixes the leak.  Perhaps there is a better way?

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -261,6 +261,31 @@ bool MDNSResponder::addServiceTxt(char *name, char *proto, char *key, char *valu
   return false;
 }
 
+void MDNSResponder::clearServiceTxt(const char *name, const char *proto){
+	MDNSService* servicePtr;
+	//Find the service
+  for (servicePtr = _services; servicePtr; servicePtr = servicePtr->_next) {
```
devyte: declare servicePtr inside the for loop
for (MDNSService * servicePtr = _services; ...)

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -261,6 +261,31 @@ bool MDNSResponder::addServiceTxt(char *name, char *proto, char *key, char *valu
...
+    //Checking Service names
+    if(strcmp(servicePtr->_name, name) == 0 && strcmp(servicePtr->_proto, proto) == 0){
+      //found a service name match
+      if (servicePtr->_txts != 0) {
+      	MDNSTxt * txtPtr = servicePtr->_txts;
+      	MDNSTxt * nextPtr = txtPtr->_next;
+      	delete txtPtr;
+      	txtPtr = 0;
+      	while (nextPtr !=0) {
```
devyte: why is the service clear done with a for loop and the txt clear done with a while loop? Suggest changing this to a for loop similar as service case

devyte:
@mkillewald I think that was correct, but can be improved for consistency/readability.
I would request one additional thing: put the loop that clears all texts input its own function, then call that function from inside the service for loop.  Better structure, and that function could be reused e.g.: from another function to clear all txts of all services or whatever.

devyte:
The ability to remove TXTs is already implemented in the new MDNS responder written by @LaborEtArs .
Closing.

Issue #2750, opened 2016-12-09
Incorrect Assumption There Is A TCP Connection When There May Not Be


kentaylor:
### Basic Infos
Where there is an unreliable network connection the ESP8266 using the Arduino Core can get stuck in an endless loop waiting for a message that will never arrive. 
#### Hardware
Hardware:			ESP-12E
Core Version:      	All

### Description

The effect of the problem is described in [another thread](https://github.com/squix78/esp8266-weather-station/issues/8#issuecomment-230306417) that links to several instance reports. It happens often for me when the ESP is connected to a WiFi AP that uses a mobile phone network for its internet connectivity.

The failure is a consequence of an incorrect assumption that the ESP8266 knows whether it has a TCP connection.  The assumption is correct when the TCP connection is created and destroyed in accordance with this [state](http://tcpipguide.com/free/t_TCPOperationalOverviewandtheTCPFiniteStateMachineF-2.htm) diagram  ![TCP Connection flow diagram](http://tcpipguide.com/free/diagrams/tcpfsm.png) 
but can be invalid if there is packet loss once the TCP connection gets to the ESTABLISHED state shown in green in the diagram. 

When lost packets are recovered in the TCP layer the state diagram is followed and when packets sent from the ESP are not acknowledged the ESP should know the TCP connection is broken but when the remote socket is closed incorrectly, perhaps through a power outage or packets have been sent from a remote socket but not received by the ESP it can not know. If packets are sent from the remote socket but not received, the remote machine will retry and eventually close its socket. If the ESP is waiting for incoming packets it will still think the TCP connection exists when it doesn't.

[...]

PR #3937, opened 2017-12-08
keepalive(2h,75s,9) API + status helper + change/disable methods (not enabled by default)

d-a-v:
(*not yed* automatic) tcp-keep-alive for all new connections with standard default values
#2750

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClient.h
+++ b/libraries/ESP8266WiFi/src/WiFiClient.h
@@ -30,6 +30,10 @@
 
 #define WIFICLIENT_MAX_PACKET_SIZE 1460
 
+#define TCP_DEFAULT_KEEPALIVE_IDLE_SEC          7200 // 2 hours
+#define TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC      75   // 75 sec
+#define TCP_DEFAULT_KEEPALIVE_COUNT             9    // fault after 9 failures
+
 class ClientContext;
 class WiFiServer;
 
@@ -84,6 +88,13 @@ public:
 
   static void stopAll();
   static void stopAllExcept(WiFiClient * c);
+
+  void     keepAlive (uint16_t idle_sec = TCP_DEFAULT_KEEPALIVE_IDLE_SEC, uint16_t intv_sec = TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC, uint8_t count = TCP_DEFAULT_KEEPALIVE_COUNT);
+  bool     isKeepAliveEnabled () const;
+  uint16_t getKeepAliveIdle () const;
+  uint16_t getKeepAliveInterval () const;
+  uint8_t  getKeepAliveCount () const;
+  void     disableKeepAlive () { keepAlive(0, 0, 0); }
 
 protected:
 
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -43,6 +43,9 @@ public:
         tcp_sent(pcb, &_s_sent);
         tcp_err(pcb, &_s_error);
         tcp_poll(pcb, &_s_poll, 1);
+
+        // not enabled by default for 2.4.0
+        //keepAlive();
     }
 
     err_t abort()
@@ -339,6 +342,38 @@ public:
         }
         ProgmemStream stream(buf, size);
         return _write_from_source(new BufferedStreamDataSource<ProgmemStream>(stream, size));
+    }
+
+    void keepAlive (uint16_t idle_sec = TCP_DEFAULT_KEEPALIVE_IDLE_SEC, uint16_t intv_sec = TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC, uint8_t count = TCP_DEFAULT_KEEPALIVE_COUNT)
+    {
+        if (idle_sec && intv_sec && count) {
+            _pcb->so_options |= SOF_KEEPALIVE;
+            _pcb->keep_idle = (uint32_t)1000 * idle_sec;
+            _pcb->keep_intvl = (uint32_t)1000 * intv_sec;
+            _pcb->keep_cnt = count;
+        }
+        else
+            _pcb->so_options &= ~SOF_KEEPALIVE;
+    }
+
+    bool isKeepAliveEnabled () const
+    {
+        return !!(_pcb->so_options & SOF_KEEPALIVE);
+    }
+
+    uint16_t getKeepAliveIdle () const
+    {
+        return isKeepAliveEnabled()? (_pcb->keep_idle + 500) / 1000: 0;
+    }
+
+    uint16_t getKeepAliveInterval () const
+    {
+        return isKeepAliveEnabled()? (_pcb->keep_intvl + 500) / 1000: 0;
+    }
+
+    uint8_t getKeepAliveCount () const
+    {
+        return isKeepAliveEnabled()? _pcb->keep_cnt: 0;
     }
 
 protected:
--- a/libraries/ESP8266WiFi/src/WiFiClient.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClient.cpp
@@ -344,3 +344,29 @@ void WiFiClient::stopAllExcept(WiFiClient* except)
         }
     }
 }
+
+
+void WiFiClient::keepAlive (uint16_t idle_sec, uint16_t intv_sec, uint8_t count)
+{
+    _client->keepAlive(idle_sec, intv_sec, count);
+}
+
+bool WiFiClient::isKeepAliveEnabled () const
+{
+    return _client->isKeepAliveEnabled();
+}
+
+uint16_t WiFiClient::getKeepAliveIdle () const
+{
+    return _client->getKeepAliveIdle();
+}
+
+uint16_t WiFiClient::getKeepAliveInterval () const
+{
+    return _client->getKeepAliveInterval();
+}
+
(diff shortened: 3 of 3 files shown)


libraries/ESP8266WiFi/src/WiFiClient.h
```diff
@@ -84,6 +84,13 @@ class WiFiClient : public Client, public SList<WiFiClient> {
   static void stopAll();
   static void stopAllExcept(WiFiClient * c);
 
+  void keepAlive (uint16_t idle_sec, uint16_t intv_sec, uint8_t count);
+  uint16_t getIdleKeepAlive ();
+  void disableKeepAlive () { keepAlive(0, 0, 0); }
+
+  static void setKeepAlive (void* pcb, uint16_t idle_sec, uint16_t intv_sec, uint8_t count);
+  static void getKeepAlive (void* pcb, uint16_t* idle_sec, uint16_t* intv_sec, uint8_t* count);
```
igrr: How about moving these two into ClienContext? Exposing details such as pcb does not look right...
If you move it into ClienContext, then the first argument can also be changed from void* to tcp_pcb*.

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -31,6 +31,10 @@ extern "C" void esp_schedule();
 
 #include "DataSource.h"
 
+#define TCP_DEFAULT_KEEPALIVE_IDLE_SEC          7200 // 2 hours
+#define TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC      75   // 75 sec
+#define TCP_DEFAULT_KEEPALIVE_COUNT             9    // fault after 9 failures
```
igrr: Do you think that releasing 2.4.0 with keepalives enabled by default, after two release candidates didn't have them, may cause some issues? How about adding methods to enable keepalives first, and then enable them by default after the release?

libraries/ESP8266WiFi/src/WiFiClient.h
d-a-v: WiFiClient.cpp has <lwip/*.h>
ClientContext has private _pcb
You make me realize that since ClientContext knows tcp_pcb, it has <lwip/*.h> too.
will fix :)

libraries/ESP8266WiFi/src/include/ClientContext.h
d-a-v: I set up an example with 5s idle-keep-alive (default value is 2 hours), logs are:
```
16:41:47.308767 IP 10.43.1.163.23 > 10.43.1.254.51006: Flags [.], ack 28, win 2117, length 0
16:41:47.308797 IP 10.43.1.254.51006 > 10.43.1.163.23: Flags [.], ack 1, win 29200, length 0

16:41:52.808481 IP 10.43.1.163.23 > 10.43.1.254.51006: Flags [.], ack 28, win 2117, length 0
16:41:52.808512 IP 10.43.1.254.51006 > 10.43.1.163.23: Flags [.], ack 1, win 29200, length 0
```
.163 is esp with tcpserial example. We can see that tcp sends empty packets which are answered by peer. ```WiFiClient::connected()``` will answer false once ```keep-alive-count``` packets are not answered. Libraries and Sketchs already take care of ```::connected()``` result, so yes I think there will be no harm. 
I understand and agree that we can never predict side effects. I made this PR so that #2750 can see and test, it can wait 2.5, or I can disable it by default for 2.4.

d-a-v:
keepalive PR is squash-updated and still enabled by default.

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -325,6 +328,26 @@ class ClientContext
...
+            _pcb->so_options |= SOF_KEEPALIVE;
+            _pcb->keep_idle = (uint32_t)1000 * idle_sec;
+            _pcb->keep_intvl = (uint32_t)1000 * intv_sec;
+            _pcb->keep_cnt = count;
+        }
+        else
+            _pcb->so_options &= ~SOF_KEEPALIVE;
+    }
+
+    void getKeepAlive (uint16_t* idle_sec, uint16_t* intv_sec, uint8_t* count)
```
devyte: Maybe use of references instead of pointers would be better here?

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -325,6 +328,26 @@ class ClientContext
         return _write_from_source(new BufferedStreamDataSource<ProgmemStream>(stream, size));
     }
 
+    void keepAlive (uint16_t idle_sec = TCP_DEFAULT_KEEPALIVE_IDLE_SEC, uint16_t intv_sec = TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC, uint8_t count = TCP_DEFAULT_KEEPALIVE_COUNT)
+    {
+        if (idle_sec && intv_sec && count) {
```
devyte: What happens if any 2 of the 3 values are non-zero? Are all combinations allowed?

devyte: (mostly thinking of sanity checks, if necessary)

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -325,6 +328,26 @@ class ClientContext
...
+            _pcb->keep_cnt = count;
+        }
+        else
+            _pcb->so_options &= ~SOF_KEEPALIVE;
+    }
+
+    void getKeepAlive (uint16_t* idle_sec, uint16_t* intv_sec, uint8_t* count)
+    {
+        bool enabled = _pcb->so_options & SOF_KEEPALIVE;
+        if (idle_sec) *idle_sec = enabled? (_pcb->keep_idle + 500) / 1000: 0;
```
devyte: Indentation (body of the if statement on a new line)

libraries/ESP8266WiFi/src/WiFiClient.h
```diff
@@ -30,6 +30,10 @@
 
 #define WIFICLIENT_MAX_PACKET_SIZE 1460
 
+#define TCP_DEFAULT_KEEPALIVE_IDLE_SEC          7200 // 2 hours
```
devyte: Do #defines count against available heap, or is it better to make these static const and put them in flash?
(I'm thinking of the free heap project, every byte counts)

d-a-v: unlike ```static const```, ```#define``` count for nothing in ram nor flash.

libraries/ESP8266WiFi/src/include/ClientContext.h
d-a-v: any 0 is an invalid value for keep-alive api, so any 0 disables keep-alive.

libraries/ESP8266WiFi/src/include/ClientContext.h
d-a-v: This way with pointers let the user choose which value to get, like with ```gettimeofday(&tv, NULL)``` we never want the second one. This can't be done with references
I put this ```getKeepAlive()``` method mainly for the user to be informed if keep-alive is activated through ```WiFiClient::``` since ```ClientContext``` is not accessible.
I'm open to any other method than this one.

libraries/ESP8266WiFi/src/include/ClientContext.h
d-a-v: will fix

libraries/ESP8266WiFi/src/include/ClientContext.h
devyte: I missed your original intent. In that case, I think I'd prefer three inline methods, one for each value, each returning the requested value, plus a 4th inline method that just returns the enabled flag as a bool, something along the lines of:
getKeepAliveCount()
getKeepAliveInterval()
getKeepAliveIdle()
isKeepAliveEnabled()

devyte: Body could be like this:
return isKeepAliveEnabled() ? _pcb-> blah : 0;

d-a-v: keepalive-api is updated, PR squashed-rebased

devyte:
@d-a-v if you add the getter, I think this can be merged

d-a-v:
The getters are already present (I push-squashed to reduce the number of commits).
But now we are at pre-2.5, should'nt we enable keepalive by default ?
The default idle value is 2h, It could be 10s (it sends empty packets).
Regarding #2330 what good value could we use for a good start ?

devyte:
@d-a-v isn't WiFiClient missing a getter isKeepAliveEnabled()?

d-a-v:
right, this is fixed

devyte:
Still pending: verify if the defaults are ok, but that can be a different issue.

d-a-v:
> Why isn't keepAlive enabled by default?

Because we did not check all side effects yet.
But you can enable it as soon as you have a valid WiFiClient.

> And does WiFiServer support keepAlive?

keepAlive is for TCP stream (WiFiClient here).
This is irrelevant for WiFiServer: WiFiServer is awaiting for those streams.

Androbin:
I'm asking because `ESP8266WebServer` is backed by `WiFiServer`.
And `ESP8266WebServer` is responding with `Connection: close`.
(I'm sending `Connection: keep-alive` in the request btw.)
I couldn't figure out how to keep the connection open. Is that currently possible?

d-a-v:
This is unrelated. This pull-request is about keep-alive TCP's feature, not connection type in HTTP header.
Please open a new issue and fill the issue template.

PR #3950, opened 2017-12-12
Update a02-my-esp-crashes.rst

### LINKED ISSUE #1388: Document requirement for ISR handlers to be in IRAM
Hello *,

I modified the UART based NeoPixelBus to use I²S and DMA for CPU-less transfer. (it is based on https://github.com/cnlohr/esp8266ws2812i2s)

First i tried to use Interrupts to poll for kinda vsync implementation.
But while doing that when my code wanted to save some config, SPIFFS always locked up. (no reaction, reset after a couple of minutes)
Sometimes i got an exception with PC right at the interrupt handler address which made me curious.

In the end i found out that _it seems_ that the ROM call for flash writing does not disable (at least) SLC interrupts.
This is a no-go of course if you are rewriting flash content since the flash state machine will get corrupted on every flash access and instruction fetches return crap.

Now when i am polling SLC_RX_EOF_DES_ADDR instead of using an interrupt, everything works like a charm.

Can you confirm this behavior or am i misinterpreting things?
Wouldnt it be a good idea to disable interrupts before calling the ROM functions for flash rewriting?

BR,
Georg
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29715734-document-requirement-for-isr-handlers-to-be-in-iram?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

devyte:
Added section Other causes for crashes with ISR, Async callbacks, memory guidelines, stack.
Fixes #1388 .


devyte:
@igrr @d-a-v any feedback here?

devyte:
@igrr @d-a-v ping! BTW, the relevant issue is targeted for 2.4.0.

doc/faq/a02-my-esp-crashes.rst
```diff
@@ -229,6 +230,77 @@ dropped. The same procedure applies to crashes caused by exceptions.
...
+   case there is no access to the underlying code.
+
+*Some techniques for reducing memory usage*
+
+   * Don't use const char * with literals. Instead, use const char[] PROGMEM. This is particularly true if you intend to, e.g.: embed html strings.
+   * Don't use global static arrays, such as uint8_t buffer[1024]. Instead, allocate dynamically. This forces you to think about the size of the array, and its scope (lifetime), so that it gets released when it's no longer needed. If you are not certain about dynamic allocation, use std libs (e.g.: std:vector, std::string), or smart pointers. They are slightly less memory efficient than dynamically allocating yourself, but the provided memory safety is well worth it.
+   * If you use std libs like std::vector, make sure to call its ::reserve() method before filling it. This allows allocating only once, which reduces mem fragmentation, and makes sure that there are no empty unused slots left over in the container at the end.
+
+Stack
+   The amount of stack in the ESP is tiny at only 5KB. For normal developement in large systems, it
```
igrr: Isn't it 4kB? (`#define CONT_STACKSIZE 4096`)

igrr:
Looks good otherwise!

Issue #3853, opened 2017-11-20
Travis CI is hitting the 50min limit too often


devyte:
#### Hardware
Core Version:      	latest git

### Description

Travis CI is hitting the 50min limit more often than not, even when only a single build is running. As a result, builds need to be restarted often, sometimes several times in a row, before it runs to completion.

devyte:
@igrr I'm currently stuck, because travis CI keeps failing due to the timelimit. It has consistently failed several times in a row now.
I know exactly nothing about travis ci, and after an hour or so of googling, I still don't really have a concrete clue as to how to proceed. 
Help?

igrr:
After a brief look, i can't find why it started failing all of a sudden.
Will work on splitting the build into separate jobs (at least two: one for Arduino IDE builds, one for platformio).

devyte:
@igrr From what I can see, a normal build run is taking 42-45 minutes. I suppose it's possible there is too much load on the host cluster these days?
From my googling I gathered that the build runs on a 2-core system. Does make run with -j 2? If not, that might give another speed boost.

devyte:
Not the right place to ask this, but what do I do with issues after I have merged a fix? Do I just add label:staged-for-release, or is there something else I should do? Like I said elsewhere, I'm looking at issues with label:easy that are marked with milestone 2.4.

igrr:
When building using Arduino IDE, `make` is not used, so we can't easily enable parallel builds there. Platformio, as far as i know, does parallel builds by default.
Splitting the whole build into multiple jobs looks like the easiest option.

Regarding tagging, for issues which have been present in previous release (or pre-release), i suggest tagging with "staged-for-release" and keeping open, so that the issue still pops up in search in case someone else encounters it. For regressions introduced in master after the last release (or pre-release), it is okay to just close the issue.

devyte:
@igrr this is happening frequently again...

PR #3959, opened 2017-12-13
ci: split into multiple jobs

igrr:
Fixes https://github.com/esp8266/Arduino/issues/3853

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,18 @@ language: bash
 os: linux
 dist: trusty
 
-env:
-  - BUILD_TYPE=build
+matrix:
+  include:
+    - env:
+      - BUILD_TYPE=build
+    - env:
+      - BUILD_TYPE=platformio
+    - env:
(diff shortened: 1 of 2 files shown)


igrr:
Seems to be happily running, somewhat in parallel: https://travis-ci.org/esp8266/Arduino/builds/315902513

devyte:
Looks like a drop in total runtime of 13minutes.

igrr:
Fortunately, total time for the build is not limited, only the time for a single job is. So as long as the longest job stays below 50 minutes, it should be okay.

Will try to split the longest jobs into two parts as well, to get into the ballpark of <20 minute build.

igrr:
Splitting sketch build into two parts looks pretty hairy, mainly because of the size log calculation. Will revisit this later.

devyte:
I think this is enough for now. The biggest grievance was having the build fail often due to the time limit, and I think that's unlikely now.
Thank you!!

PR #3967, opened 2017-12-14
fix #1002 ::Flush() wait for empty send buffer

### LINKED ISSUE #1002: Fix Flush implementations to match corrected code in latest Arduino libraries
See the Arduino issue for details and code changes
https://github.com/arduino/Arduino/issues/4029
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28169596-fix-flush-implementations-to-match-corrected-code-in-latest-arduino-libraries?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

d-a-v:
ClientContext::arduinoFlush() needs review

### DIFF
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -107,7 +107,7 @@ public:
         if(this != 0) {
             DEBUGV(":ur %d\r\n", _refcnt);
             if(--_refcnt == 0) {
-                flush();
+                discard_received();
                 close();
                 if(_discard_cb) {
                     _discard_cb(_discard_cb_arg, this);
@@ -277,7 +277,7 @@ public:
         return copy_size;
     }
 
-    void flush()
+    void discard_received()
     {
         if(!_rx_buf) {
             return;
@@ -290,15 +290,20 @@ public:
         _rx_buf_offset = 0;
     }
 
-    void arduinoFlush()
+    void wait_until_sent()
     {
-#if 1
-	// this is useless since _write_from_source() always exits with _datasource==NULL
-        while (state() == ESTABLISHED && _datasource && _datasource->available()) {
+        // fix option 1 in
+        // https://github.com/esp8266/Arduino/pull/3967#pullrequestreview-83451496
+        // TODO: option 2
+
+        #define WAIT_TRIES_MS 10	// at most 10ms
+
+        int tries = 1+ WAIT_TRIES_MS;
+
+        while (state() == ESTABLISHED && tcp_sndbuf(_pcb) != TCP_SND_BUF && --tries) {
             _write_some();
             delay(1); // esp_ schedule+yield
         }
-#endif
     }
 
     uint8_t state() const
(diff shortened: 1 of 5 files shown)


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -290,6 +290,17 @@ class ClientContext
         _rx_buf_offset = 0;
     }
 
+    void arduinoFlush()
+    {
+#if 1
+	// this is useless since _write_from_source() always exits with _datasource==NULL
+        while (state() == ESTABLISHED && _datasource && _datasource->available()) {
```
igrr: Okay, assuming we are calling this function from Arduino loop context (not from system callback), _datasource is NULL, and the remaining data, if any, has been passed to LwIP. FWIW, i don't think that the official WiFi101 library actually guarantees that the data has been delivered (only that it was passed to the WiFi modem). But let's assume we want to do better than that.

I see two ways of implementing this:

1. Wait until TCP window size is reset by LwIP. That involves checking if `tcp_sndbuf(_pcb) == TCP_WND_MAX(_pcb)`, and delaying, say, for 1 ms while it is not. Normal write timeout needs to apply, i think.

2. Introduce a flag for "some data not acked", and set it to 1 when writing the last chunk of data from the datasource. Set it to 0 from `_sent` callback, if `_datasource == NULL`. When entering flush method, check if the flag is set. If it is set, set `_send_waiting` to 1, and call `esp_yield` (or `delay(timeout)`). When `_sent` callback is called by LwIP, it will call `_write_some_from_cb`, which will `esp_schedule`, resuming the main task. At this point we know that the data is written.

Option 2 looks more "correct" to me, because it doesn't involve delaying and semi-busy-waiting with an arbitrary interval, but it adds another state flag which can potentially introduce bugs more severe than the one we want fixed here. So given that rc-2 is already behind, i would personally go for option 1 and add a ticket to implement option 2 later.

igrr: Also, how about renaming the existing `flush` to `discard_received` and this new function to something like `wait_until_sent` or similar?

d-a-v: I did my homework

d-a-v: notes:
It appears that ```TCP_WND_MAX()``` is for the receive windows. Max value of snd_buf is a defined ```TCP_SND_BUF``` constant.
I raised the max delay to 10ms after measuring it with a heavily loaded tcp echo tester (both side on wifi)

devyte:
At a glance, code looks good to me (untested).
Can this be merged, or is there anything pending for 2.4.0? (I know there is something pending for 2.5.0)

d-a-v:
I guess it can be merged (author can't approve their own PR :-) )

jandrassy:
and moving flush() from Stream to Print?

PR #4001, opened 2017-12-20
handle tv.tv_usec in settimeofday()

### LINKED ISSUE #1679: time() does not work without active NTP server

d-a-v:
optional settimeofday()'s callback
fix #1679

### DIFF
--- a/cores/esp8266/coredecls.h
+++ b/cores/esp8266/coredecls.h
@@ -2,8 +2,19 @@
 #ifndef __COREDECLS_H
 #define __COREDECLS_H
 
-extern bool s_bootTimeSet;
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 // TODO: put declarations here, get rid of -Wno-implicit-function-declaration
+
+extern bool timeshift64_is_set;
+
+void tune_timeshift64 (uint64_t now_us);
+void settimeofday_cb (void (*cb)(void));
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif // __COREDECLS_H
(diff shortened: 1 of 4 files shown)


BrandonLWhite:
The changes pertaining to tune_timeshift64 look good to me!  Thanks @d-a-v !

PR #4011, opened 2017-12-21
Added heltec wifi kit 8

9a4gl:
This one goes together with https://github.com/platformio/platform-espressif8266/pull/67

### DIFF
--- a/variants/heltec_wifi_kit_8/pins_arduino.h
+++ b/variants/heltec_wifi_kit_8/pins_arduino.h
@@ -1,0 +1,49 @@
+/*
+  pins_arduino.h - Pin definition functions for Arduino
+  Part of Arduino - http://www.arduino.cc/
+
+  Copyright (c) 2007 David A. Mellis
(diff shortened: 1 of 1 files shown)


d-a-v:
As this would be certainly useful for platformio, this PR lacks consistency in the core repo, as this is a variant with no matching board.
Would you provide more information on this ?

d-a-v:
@ivankravets @9a4gl Is this core issue still relevant ?

9a4gl:
> @ivankravets @9a4gl Is this core issue still relevant ?

I have no idea what core has and should be. I just provided what was missing to be able to select that board. It was almost 1 year ago, in meanwhile I do not use that board anymore.

ivankravets:
We have this board in our registry and it uses ["variant": "d1_mini"](https://github.com/platformio/platform-espressif8266/blob/develop/boards/heltec_wifi_kit_8.json#L10).

If this board is not compatible with `d1_mini` variant, the new variant should be provided.

Issue #4002, opened 2017-12-20
Boundary condition in cbuf


boblemaire:
Basic Infos

Hardware

Hardware:	NodeMCU 1 (ESP-12E)
Core Version: 2.4.0-rc2

Description

Problem description
While trying to figure out how data is being lost in webserver responses, I went through the code for cbuf.  I discovered a boundary condition in the resize function.  If you were to issue the function:
`cbuf:resize(cbuf.available());`
to reduce the size of a buffer to it's current contents, the new cbuf will be the exact size of the current contents, but cbuf logic requires that the buffer always be at least one byte larger than it's current contents. 

Settings in IDE

platformio
Module: NodeMCU 1 (ESP-12E)
Flash Size: 4MM
CPU Frequency:	80Mhz
Upload Using:	USB

SKETCH
 Here is a sequence of calls to cbuf that illustrates the problem with a cbuf(10):
```
cbuf buf(10);
Serial.printf("\r\nsize=%d, available=%d, room=%d, full=%s\r\n", buf.size(), buf.available(), buf.room(), buf.full() ? "true" : "false");
for (int i=0; i<8; i++){
    Serial.printf("Write one byte, returned write count=%d, cbuf:",buf.write("X",1));
    Serial.printf("size=%d, available=%d, room=%d, full=%s\r\n", buf.size(), buf.available(), buf.room(), buf.full() ? "true" : "false");
}
Serial.printf("Resize=%d, returned size=%d, cbuf:",buf.available(),buf.resize(buf.available()));
    
Serial.printf("size=%d, available=%d, room=%d, full=%s\r\n", buf.size(), buf.available(), buf.room(), buf.full() ? "true" : "false");
for (int i=0; i<8; i++){
    Serial.printf("Write one byte, returned write count=%d, cbuf:",buf.write("X",1));
    Serial.printf("size=%d, available=%d, room=%d, full=%s\r\n", buf.size(), buf.available(), buf.room(), buf.full() ? "true" : "false");
}
```
Serial OUTPUT
```
size=10, available=0, room=9, full=false
Write one byte, returned write count=1, cbuf:size=10, available=1, room=8, full=false
Write one byte, returned write count=1, cbuf:size=10, available=2, room=7, full=false
Write one byte, returned write count=1, cbuf:size=10, available=3, room=6, full=false
Write one byte, returned write count=1, cbuf:size=10, available=4, room=5, full=false
[...]

PR #4016, opened 2017-12-22
Fix boundary condition on resize

boblemaire:
cbuf.size() must be at least one byte larger than cbuf.available() for
logic to work.  reject request to resize = available.  Issue #4002.


devyte:
Thanks for finding this. It took me a while to understand the underlying issue. For the record, you are also correct in that cbuf allocates 1byte more than it really uses. Therefore, you can't allow downsizing to the exact number of elements contained, and must allow to the number of elements + 1 and above.
It may be worth considering replacing this code with a wrapper around an STL-style container, or e.g.: boost's ring buffer, but for now, your fix is enough.

PR #4019, opened 2017-12-24
added public cleanAPlist() function.

NayanKaran:
cleaning AP list without calling destructor.


earlephilhower:
This conflicts and looks like it's missing bits to actually do anything.  Closing.

Issue #2256, opened 2016-07-10
Secure MQTT getting exception 29 with two-way traffic


mtnbrit:
#### Hardware

Hardware:          ESP-12F
Core Version:       2.3.0
### Description

running secure mqtt on port 8883, im using the lmroy PubSubClient mqtt lib connected to mosquitto 1.4.8 broker, esp is listening (subscribed) to several topics on broker, and also periodically publishing its own messages to different topics. I can trigger exception 29 by publishing something to one of the subscribed topics. I spent yesterday debugging the PubSubClient and MQTT libs to find that if the esp is publishing something while also receiving messages on topics its subscribed to, the received stream gets corrupted (random types and zero length or huge messages lengths) and crashes the PubSubClient new Publish() function called from client.loop(), _recv_message() and readPacket() . 

It feels like the tls receive buffer is getting overwritten, I remember one of the memory optimizations igrr made was to use the same memory for both the rx and tx buffers, or something akin?

If I switch to non-tls mqtt on port 1883, it doesn't crash at all under the same conditions.

Maybe its not possible to send and receive at the same time over secure mqtt with the current state of axTLS buffers?

I haven tried the knolleary pubsubclient lib yet, its a lot of work to switch over, but I would if this is known to work.
### Settings in IDE

Module:             Generic ESP8266 Module
Flash Size:         4MB/1MB
CPU Frequency:      80Mhz
Flash Mode:         dio
Flash Frequency:    40Mhz
Upload Using:       OTA
Reset Method:       ck
### Sketch

I haven't had time to boil my sketch down to just the parts that crash, hopefully there is enough general info in the problem description to confirm or decode what may be going on.

...normal loop execution...
ESP free heap: 18240
(waiting between loops)
                                    <---- I publish a message to one of the subscribed topics here
Exception (29):
epc1=0x4000df64 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

ctx: cont 
sp: 3fff2d60 end: 3fff3120 offset: 01a0

> > > stack>>>
> > > 3fff2f00:  3fff6904 00000000 3fff69e4 4020f404  
> > > 3fff2f10:  00000000 00000000 00000000 00003034  
> > > 3fff2f20:  00000000 3fff6a08 3fff6904 401004d8  
> > > 3fff2f30:  00000002 00000000 3fff6904 0000002d  
> > > 3fff2f40:  3fff69e4 00000032 3fff6904 4020f944  
[...]

PR #4024, opened 2017-12-25
WiFiClientSecure: don't trash unread decrypted data when writing

igrr:
When application requests to write data, check if there is any unread decrypted data left. If there is, don't write immediately, but save the data to be written. When all decrypted data has been consumed by the application, send out the saved outgoing data.

Also don't use `available()` in the test for `connected()` to reduce the chance that plaintext buffer gets filled with incoming data when the application does not actually need it, and does not need to know its exact size.

Fixes https://github.com/esp8266/Arduino/issues/2256.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
@@ -193,6 +193,12 @@ public:
         return cb;
     }
 
+    // similar to availble, but doesn't return exact size
+    bool hasData()
+    {
+        return _available > 0 || (s_io_ctx && s_io_ctx->getSize() > 0);
+    }
+
     bool loadObject(int type, Stream& stream, size_t size)
     {
         std::unique_ptr<uint8_t[]> buf(new uint8_t[size]);
@@ -458,7 +464,7 @@ err     x       N           N
 uint8_t WiFiClientSecure::connected()
 {
     if (_ssl) {
-        if (_ssl->available()) {
+        if (_ssl->hasData()) {
             return true;
         }
         if (_client && _client->state() == ESTABLISHED && _ssl->connected()) {
--- a/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
@@ -27,6 +27,7 @@ extern "C"
 #include "osapi.h"
 #include "ets_sys.h"
 }
+#include <list>
 #include <errno.h>
 #include "debug.h"
 #include "ESP8266WiFi.h"
@@ -49,6 +50,26 @@ extern "C"
 #else
 #define SSL_DEBUG_OPTS 0
 #endif
+
+
+typedef struct BufferItem
+{
+    BufferItem(const uint8_t* data_, size_t size_)
+    : size(size_), data(new uint8_t[size])
+    {
+        if (data.get() != nullptr) {
+            memcpy(data.get(), data_, size);
+        } else {
+            DEBUGV(":wcs alloc %d failed\r\n", size_);
+            size = 0;
+        }
+    }
+
+    size_t size;
+    std::unique_ptr<uint8_t[]> data;
+} BufferItem;
+
+typedef std::list<BufferItem> BufferList;
 
 class SSLContext
 {
@@ -139,6 +160,10 @@ public:
         _available -= will_copy;
         if (_available == 0) {
             _read_ptr = nullptr;
+            /* Send pending outgoing data, if any */
+            if (_hasWriteBuffers()) {
+                _writeBuffersSend();
+            }
         }
         return will_copy;
     }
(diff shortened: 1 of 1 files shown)


igrr:
Vector doesn't fit: when you try to append a new block of data, it will try to allocate contiguous region to fit all existing data and new data. With fragmented heap, this might not be possible. List-of-buffers implemented here does not have this limitation.
List from STL could be used, but it has higher overhead: two allocations (one for list entry, one for the data itself) instead of one, and two pointers instead of one to maintain the list. Pulling in std::list also increases code size by a few kB, compared to the proposed solution.

(Not to say I'm totally happy with the existing code — which is the reason this is marked as WIP).

igrr:
std::list might be okay though, if I can make the code size hit less noticable.

earlephilhower:
Does this mean that BearSSL replacement is no longer on the table?  It's separate RX and TX buffers (of separately configurable sizes) would avoid contortions like this, no?

igrr:
@earlephilhower It doesn't meant that. This workaround is intended for axTLS only. I assume that if/when BearSSL is adopted, WiFiClientSecure will need a more or less full rewrite.

igrr:
@devyte Please take a look, i have replaced all C-style buffer handling with std::list + std::unique_ptr. Turns out I have already pulled in std::list from ESP8266WiFiGeneric.cpp, so code size overhead from this change is just a couple hundred bytes, compared to STAILQ.

devyte:
I was doing just that. Looks much better! I'll add specific comments to code change lines.

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
```diff
@@ -282,12 +333,59 @@ class SSLContext
...
+        if (_writeBuffers.back().data.get() == nullptr) {
+            _writeBuffers.pop_back();
+            return 0;
+        }
+        return size;
+    }
+
+    int _writeBuffersSend()
+    {
+        while (!_writeBuffers.empty()) {
```
devyte: Is this a lengthy operation? Are there timing concerns due to the exection time of this loop?

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
```diff
@@ -282,12 +333,59 @@ class SSLContext
...
+            auto& first = _writeBuffers.front();
+            int rc = _write(first.data.get(), first.size);
+            _writeBuffers.pop_front();
+            if (rc < 0) {
```
devyte: Maybe a DEBUGV msg here for the failure and for the buffers getting dropped (data drop)?

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
```diff
@@ -155,10 +176,34 @@ class SSLContext
...
+            if (_hasWriteBuffers()) {
+                _writeBuffersSend();
+            }
         }
         return result;
     }
 
+    int write(const uint8_t* src, size_t size)
+    {
+        if (!_available) {
```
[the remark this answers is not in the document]
igrr: No need to check for hasData, as far as i can tell — we only need to make sure that the fragment buffer is empty.

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
```diff
@@ -155,10 +176,34 @@ class SSLContext
...
+           the fragment buffer, corrupting the received data.
+           Save a copy of the outgoing data, and call ssl_write when all
+           recevied data has been consumed by the application.
+        */
+        return _writeBufferAdd(src, size);
+    }
+
     int peek()
     {
         if (!_available) {
```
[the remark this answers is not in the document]
igrr: I don't think so. The logic is:

1. Is there some incoming plaintext data? If not, we'll try to get any received data from the TCP stack, and decrypt it.
2. If after step 1 there is some plaintext data, return the first byte of it.

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
igrr: Depends on the amount of data queued for transmission. In most cases, this will be zero. For protocols like MQTT this may send a few packets. There will be `yield`s in ClientContext, so we don't need to worry about starving the WiFi stack here.

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
igrr: Added DEBUGV for ssl_write failure into _write and a message about buffers being dropped here.

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
```diff
@@ -282,12 +333,59 @@ class SSLContext
...
+
+    int _writeBufferAdd(const uint8_t* data, size_t size)
+    {
+        if (!_ssl) {
+            return 0;
+        }
+
+        _writeBuffers.emplace_back(data, size);
+        if (_writeBuffers.back().data.get() == nullptr) {
+            _writeBuffers.pop_back();
```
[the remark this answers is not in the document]
igrr: Added a DEBUGV into BufferItem constructor, where the memory allocation failure is first detected.

vshymanskyy:
@igrr Thank you. Looks stable in our scenario

Issue #3944, opened 2017-12-11
Configurable TLS timeout for mutual authentication


ztittle:
https://github.com/esp8266/Arduino/blob/7df28582858e5a935c721d131064e6e3082edfe8/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp#L356

I've been trying to connect to the AWS IOT MQTT endpoint, but it appears that the ESP 8266 chip takes longer than the hard-coded 5-second timeout to perform the TLS handshake. After increasing this value to a generous 30 seconds, I was able to successfully connect to AWS IOT. 

For reference, I was using the example code listed on https://github.com/copercini/esp8266-aws_iot/blob/master/examples/MQTT_x509_bin/MQTT_x509_bin.ino, but loading the certs from PROGMEM so I would not run out of memory.

Could the timeout be made configurable through a #DEFINE or an overload on connect()? This would greatly help adoption when integrating to a secure MQTT service such as AWS IoT. 

`Debug output`
```
.........wifi evt: 7
....................wifi evt: 3
.192.168.1.252Attempting MQTT connection...[hostByName] Host: 54.68.144.180 is a IP!
:ref 1
please start sntp first !
State:  sending Client Hello (1)
:wr 100 100 0
:wrc 100 100 0
:sent 100
:rn 1460
:rd 5, 1460, 0
:rdi 1460, 5
:rd 1455, 1460, 5
:rdi 1455, 1455
:c0 1455, 1460
:rn 1460
:rd 1460, 1460, 0
:rdi 1460, 1460
:c0 1460, 1460
:rn 118
:rd 118, 118, 0
:rdi 118, 118
:c0 118, 118
State:  receiving Server Hello (2)
State:  receiving Certificate (11)
=== CERTIFICATE ISSUED TO ===
Common Name (CN):               *.iot.us-west-2.amazonaws.com
Organization (O):               Amazon.com, Inc.
Location (L):                   Seattle
Country (C):                    US
State (ST):                     Washington
Basic Constraints:              CA:FALSE, pathlen:10000
Key Usage:                      critical, Digital Signature, Key Encipherment
Subject Alt Name:               iot.us-west-2.amazonaws.com *.iot.us-west-2.amazonaws.com
=== CERTIFICATE ISSUED BY ===
Common Name (CN):               Symantec Class 3 Secure Server CA - G4
Organization (O):               Symantec Corporation
Organizational Unit (OU):       Symantec Trust Network
Country (C):                    US
[...]

PR #4027, opened 2017-12-26
Configurable WiFiClientSecure connect timeout, better default value

igrr:
- Use Client::_timeout in WiFiClientSecure::connect instead of a hardcoded value. This can be modified via Client::setTimeout before connecting.
- Set default _timeout for WiFiClientSecure to 15 seconds.

Ref. https://github.com/esp8266/Arduino/issues/3944


ztittle:
This change looks good to me. Thanks !

PR #4029, opened 2017-12-26
follow arduino API: pure virtual Stream::flush() moved to empty virtual Print::flush()

### LINKED ISSUE #4018: flush() is a TX related function
WiFiClient.flush should send all data from tx buffer. it is not true that in Arduino world it is a rx buffer thing. the Client class is inherited from Print class and the [flush method](https://www.arduino.cc/reference/en/language/functions/communication/stream/streamflush/)  is from Print. And in Print implementations the flush is a tx thing. And the reference for the Client class implementation is [Ethernet library](https://www.arduino.cc/en/Reference/ClientFlush).

d-a-v:
re fix #4018
::flush() declaration are now exactly the same as arduino's
[flush added](https://github.com/arduino/Arduino/blob/cd9a15ebd1647336200b9328bc0912495e3a7195/hardware/arduino/avr/cores/arduino/Print.h#L90)
[flush removed](https://github.com/arduino/Arduino/blob/cd9a15ebd1647336200b9328bc0912495e3a7195/hardware/arduino/avr/cores/arduino/Stream.h#L62)


igrr:
Looks like Print is also missing the virtual destructor... wonder why the compiler isn't warning about this. Change looks good though.

d-a-v:
I guess that's because all arduino classes are generally global variables not dynamically allocated.
I'll try to make the compiler complain and eventually add the virtual destructor.

PR #4044, opened 2017-12-29
Extern C blocks (#1352)

devyte:


devyte:
@igrr I added the C++ guards to the files in tools/sdk/include where it made sense (skipped those with just types and defines, and json/*), and updated some core and library files. Please review, I hope I didn't miss anything.

PR #4054, opened 2017-12-31
ota: fix potential network error by checking return values

d-a-v:
This needs further testing and is not needed for 2.4.0.


devyte:
I don't know how to check this. I don't really understand the involved code.

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -307,7 +307,7 @@ void ArduinoOTAClass::_runUpdate() {
     int waited = 1000;
     while (!client.available() && waited--)
```
earlephilhower, line 308: This should be a PolledTimeout, now that they exist...

cores/esp8266/Updater.h
```diff
@@ -139,38 +139,24 @@ class UpdaterClass {
...
-          }
-        }
-        if(remaining() == 0)
-          return written;
-        delay(1);
-        available = data.available();
+      // load exactly one _bufSize before flashing it
+      // (the last one may be smaller)
+      size_t wantedBufSize = remaining() < _bufferSize?
+        remaining():
```
earlephilhower, line 148: Kind of odd multiline trinary.  Style related only.  Logic seems fine.

cores/esp8266/Updater.h
```diff
@@ -139,38 +139,24 @@ class UpdaterClass {
...
+        _bufferSize;
+      size_t readThisTime = 0;
+
+      while (data.available() && _bufferLen < wantedBufSize) {
+        size_t got = data.read(_buffer + _bufferLen, wantedBufSize - _bufferLen);
```
earlephilhower, line 153: Do we need to check for `if (!got) return fail;` due to something weird on the connection side?

d-a-v: Semantic ensures `data.read()>0` because `data.available()` is true, but adding a test just in case.
There is no constraint on number of bytes read, nor duration / timeout.

d-a-v:
superseded by #6369

PR #4062, opened 2018-01-01
Add Functional and Scheduled callback option to Ticker

hreintke:
Add Functional and Scheduled callback option to Ticker

### DIFF
--- a/libraries/Ticker/Ticker.h
+++ b/libraries/Ticker/Ticker.h
@@ -37,21 +37,30 @@ public:
 	Ticker();
 	~Ticker();
 
+	typedef void (*callback_t)(void);
 	typedef void (*callback_with_arg_t)(void*);
-	typedef std::function<void(void)> TickerFunction;
+	typedef std::function<void(void)> callback_function_t;
 
-	void attach(float seconds, TickerFunction tf, bool reqSchedule = false)
+	void attach_scheduled(float seconds, callback_function_t callback)
 	{
-		internalTicker = tf;
-		scheduleTicker = reqSchedule;
-		attach(seconds, internalCallback, (void*)this);
+		attach(seconds,std::bind(schedule_function, callback));
 	}
 
-	void attach_ms(uint32_t milliseconds, TickerFunction tf, bool reqSchedule = false)
+	void attach(float seconds, callback_function_t callback)
 	{
-		internalTicker = tf;
-		scheduleTicker = reqSchedule;
-		attach_ms(milliseconds, internalCallback, (void*)this);
+		_callback_function = callback;
+		attach(seconds, _static_callback, (void*)this);
+	}
+
+	void attach_ms_scheduled(uint32_t milliseconds, callback_function_t callback)
+	{
+		attach_ms(milliseconds, std::bind(schedule_function, callback));
+	}
+
+	void attach_ms(uint32_t milliseconds, callback_function_t callback)
+	{
+		_callback_function = callback;
+		attach_ms(milliseconds, _static_callback, (void*)this);
 	}
 
 	template<typename TArg>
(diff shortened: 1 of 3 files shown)


libraries/Ticker/Ticker.h
```diff
@@ -34,17 +36,22 @@ class Ticker
 public:
 	Ticker();
 	~Ticker();
-	typedef void (*callback_t)(void);
```
igrr: It might be better to keep this definition for compatibility. I ran grep on my sketchbook and found a few uses of `Ticker::callback_t`. Perhaps others could be using it as well.

libraries/Ticker/Ticker.h
```diff
@@ -34,17 +36,22 @@ class Ticker
...
+
 	typedef void (*callback_with_arg_t)(void*);
+	typedef std::function<void(void)> TickerFunction;
 
-	void attach(float seconds, callback_t callback)
+	void attach(float seconds, TickerFunction tf, bool reqSchedule = false)
```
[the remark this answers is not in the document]
igrr: Also, why did `void attach(float seconds, void (*callback)(TArg), TArg arg)` overload not get the version with "scheduled" option?

I would expect to see the product of {attach, once} x {sec, msec} x {no arg, arg} x {normal, scheduled}, i.e. 16 functions.

libraries/Ticker/Ticker.h
```diff
@@ -24,6 +24,8 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include <functional>
+#include <Schedule.h>
```
igrr: This header file is probably only needed in Ticker.cpp

libraries/Ticker/Ticker.h
```diff
@@ -96,10 +107,14 @@ class Ticker
 
 protected:	
 	void _attach_ms(uint32_t milliseconds, bool repeat, callback_with_arg_t callback, uint32_t arg);
+	static void internalCallback (void* arg);
 
 
 protected:
 	ETSTimer* _timer;
+	TickerFunction internalTicker = nullptr;
+	bool scheduleTicker = false;
```
igrr: Instead of introducing a flag, can we do something like this for the "scheduled" case?
```c++
internalTicker = std::bind(schedule_function, tf);
```
and
```c++
internalTicker = tf;
```
for the normal (not scheduled)?

libraries/Ticker/Ticker.h
```diff
@@ -96,10 +107,14 @@ class Ticker
 
 protected:	
 	void _attach_ms(uint32_t milliseconds, bool repeat, callback_with_arg_t callback, uint32_t arg);
+	static void internalCallback (void* arg);
 
 
 protected:
 	ETSTimer* _timer;
+	TickerFunction internalTicker = nullptr;
```
igrr: For consistency with the rest of this file, please use snake_case for member and function names, and prefix the member names with an underscore. Also, perhaps `_callback` would be a better name for this member? `internalTicker` makes it look like there is an instance of `Ticker` here.

hreintke:
This is the update for making it conistent with your remarks except 

- Schedule.h include -> now needed in Ticker.h removed from Ticker.cpp
-  void attach(float seconds, void (*callback)(TArg), TArg arg) overload

I did not overload these because they are now just a specific case of the nomal functions

for `ticker.attach(10, callback, 0)` being scheduled you can use `ticker.attach_scheduled (10,std::bind(callback,0));`

The first is kept in to have backward compatibility. 
If someone needs the scheduled version there need to be a source update anyhow.

If you want the overload still included just let me know

igrr:
@hreintke Yes, please rebase on the current master. Thanks.

hreintke:
Have some trouble getting it done but will solve it.

libraries/Ticker/Ticker.h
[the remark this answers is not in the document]
bertmelis: There's no point in repeatedly `attach` the same callback as the callback will run after the loop and there no way (yet?) to stop that. Or do I miss something?

Just asking because I'm using the esp8266's interface in my esp32's version of the Ticker lib.

hreintke:
Could only rebase by creating a new PR : #4209

devyte:
Closing in favor of #4209 .

Issue #4063, opened 2018-01-01
twi_status clockCount  wrong usage ??


miky2k:
while (SDA_READ()==0 && clockCount>0){ 

should be

while (SDA_READ()==0 && clockCount-->0){ 

i feel that clockCount should be 10 not 20 (eight bit + ack + stop)

devyte:
@miky2k When you opened the issue, you were presented with an issue template requesting specific information. You have completely ignored that. Please edit your post and add the required info.
Also, can you please explain better what you mean?

miky2k:
clockCount is initialized , tested , but never changed (so why test it?),
I'm sure that, if this is true , the board,the compiler and other stuff are irrelevant, it's only c/c++ programming issue.

twi_status is helpful to unlock i2c after random reset if stolen by slave for what i understand.

5chufti:
maybe devs would be more willing to looking into this if you'd provide more reference like e.g. filename and line number ?

devyte:
After investigating the universe, I managed to descipher that:
- this issue refers to I2C [here](https://github.com/esp8266/Arduino/blob/4897e0006b5b0123a2fa31f67b14a3fff65ce561/cores/esp8266/core_esp8266_si2c.c#L205) 
- the comment does seem to make sense: clockCount seems to be some kind of timeout, and in that case decrementing it is likely correct
- There is a relevant PR: #4070 

The question that remains is whether clock_count should start at 20 like now (the PR doesn't change this value), or at 10 as proposed here.

PR #4070, opened 2018-01-02
Update core_esp8266_si2c.c

mrwgx3:
### Add 'clockCount' decrement, while-loop, twi_status()

While researching solutions for a stalled I2C bus, I found the following issue and pull-request:

I2C bus reset with status info to user, re issue 1025 #2058  
method for recovering I2C bus #1025  

The software provide by both @drmpf and @dave-prosee both worked fine, but the while-loop inside Dave's ['twi_status()'][twi_status] seems to be missing a 'clockCount' decrement, as noted in pull-request #2058 by both @Frida854 and @vlast3k. This pull-request adds the missing decrement.

[twi_status]: https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_si2c.c#L201-L208 "I2C Bus Reset Code"

The following code was used for testing:
```
uint8_t  I2CIO::busInit ( void )
{
// Set I2C bus pins to INPUT_PULLUP
   Wire.begin( SDA, SCL );            // Default I2C bus pins used here

// Change bus clockrate and stretch timeout here
   Wire.setClock( 100000 );           // 'Wire.begin()' defaults
   Wire.setClockStretchLimit( 230 );  //  shown here

// Reset the I2C bus
   uint8_t  status = Wire.status();   // Wrapper for 'twi_status()'

// If I2C bus reset went OK, re-initialize bus
   if ( status == I2C_OK )
      Wire.begin();

   return( status );
   
} //busInit

### CODE REVIEW

**Comment by mrwgx3** [comment]
Add 'clockCount' decrement, while-loop, twi_status()

**Comment by devyte** [comment]
@mrwgx3 In #4063 it is proposed to reduce the clockCount init from 20 down to 10. Does that make sense to you?

**Comment by mrwgx3** [comment]
```
@mrwgx3 In #4063 it is proposed to reduce the clockCount init from 20 down
to 10. Does that make sense to you?
```

Hi @devyte and @miky2k

I don't know the max. clock cycles required to reset an I2C bus, but I suspect that it's device dependent. I've seen 9 or 10 clock cycles given in some data-sheets, and the DS3231 RTC data-sheet (the device giving me problems) just tells you to clock until you clear the bus:

```
I2C Interface
The I2C interface is accessible whenever either VCC or VBAT is at a valid level.
If a microcontroller connected to the DS3231 resets because of a loss of VCC or other
event, it is possible that the microcontroller and DS3231 I2C communications could
become unsynchronized, e.g., the microcontroller resets while reading data from
the DS3231. When the microcontroller resets, the DS3231 I2C interface may be placed
into a known state by toggling SCL until SDA is observed to be at a high level.
At that point the microcontroller should pull SDA low while SCL is high,
generating a START condition.
```

I believe the source of the '20' clock cycles can be found here:
http://www.forward.com.au/pfod/ArduinoProgramming/I2C_ClearBus/index.html

The author, @drmpf, however, doesn't given the reason why he picked 20 as a maximum (you might ask him).

As for solving my own problem, I ultimately chose to power-cycle all the devices on the I2C bus, at which point I stopped researching the issue. Unless I could find the required number in a definitive I2C spec., I would leave things alone (maybe @miky2k has seen the spec and can provide a link) Sorry I can't give you a better answer.

Issue #2247, opened 2016-07-07
WifiMulti.addAP code inspection : Incoherent test for max length of passphrase


SteveToulouse:
### Description

WifiMulti.addAP limits the passphrase to 63 characters by the test at [L175](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp#L174).

On the other hand, Wifi.begin recognises a length of 64 chars as a PSK instead of a passphrase, see tests at [L109](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L109) and [L118](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L118).

This seems to be incoherent (wifimulti.run calls wifi.begin) unless we aren't allowed to pass PSK to wifimulti by design? 

NOTE : SSID is handled the same in both places (31 char max). _EDIT: the standard says 32 characters max for an SSID so there may be rare cases where an AP has a full 32 char, non zero-terminated SSID and in that case ESP8266 won't be able to connect to it_.

zenmanenergy:
According to [wl_definitions.h](https://github.com/esp8266/Arduino/blob/4897e0006b5b0123a2fa31f67b14a3fff65ce561/libraries/ESP8266WiFi/src/include/wl_definitions.h#L32) it looks like it should be 63. Does that mean the WifiMulti.addAP is correct with 63 and the ESP8266WiFiSTA.cpp is incorrect with 64?

devyte:
Per Espressif API reference v2.1.2, section 7.2.1:
char	password[64]

The above includes the null terminator, though.

devyte:
It took me a while to fully understand what the correct behavior is supposed to be, and what is right/wrong in the code. At irst I thought this was just a boundary condition that needed fixing. As it turns out, there are two cases for passphrases: 63 char printable ascii with null terminator, and 64 byte hex with/without null terminator.
WiFi implements both cases, with certain assumptions for the 64byte hex case, while WiFiMulti implements only the 63 ascii case.
The Espressif SDK field for the passphrase is 64bytes in size, so the hex case must be handled in a special manner, and not as a null terminated string.
[...]

PR #4076, opened 2018-01-03
fixed support for psk

devyte:
fixed in WiFiSTA, added support for psk to WiFiMulti, added count accessor, minor code cleanups


devyte:
Fixes #2247 .

igrr:
Changes look good. Just a question: have you tested that connection using psk actually works?

devyte:
@igrr no, and I don't have a setup for testing it either, so I need verification from someone else.
On the plus side, the relevant code shouldn't affect non-psk behavior, so in theory we could merge without affecting anything else.

devyte:
It's been over a week. Given that the fix shouldn't affect current behavior, that this has been looked at by @igrr, and in the interest of moving forward, I'm merging. If there is still a problem, I'll handle it in a separate issue.

PR #4086, opened 2018-01-04
ArduinoOTA: don't crash on unrecognized packets

yoursunny:


devyte:
Nice job figuring this out!

lrmoreno007:
@Juppit Yes, that solve the problem.

And an observation, in the function readStringUntil, now value is an integer but later it is compared with two char ('\ 0' and end).

From my little knowledge, I think this is better to leave it as it was.

yoursunny:
> With the for loop index is now incremented double

Oops, I didn’t notice that. I’m not used to `while` loops.
I wish there’s a way to have automated tests...

> in the function readStringUntil, now value is an integer but later it is compared with two char ('\ 0' and end).

I believe these are fine. `char` is implicitly convertable to `int` without losing precision, and they are compared as `int`.

devyte:
I missed that as well. If someone makes a PR, I'll look at it and merge it asap.

PR #4090, opened 2018-01-04
Update ArduinoOTA.cpp to solve #4086

lrmoreno007:
Update ArduinoOTA.cpp to solve #4086

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -146,7 +146,7 @@ int ArduinoOTAClass::parseInt(){
   uint8_t index;
   char value;
   while(_udp_ota->peek() == ' ') _udp_ota->read();
-  for(index = 0; index < sizeof(data); index++){
+  for(index = 0; index < sizeof(data); ++index){
     value = _udp_ota->peek();
     if(value < '0' || value > '9'){
       data[index++] = '\0';
@@ -159,13 +159,13 @@ int ArduinoOTAClass::parseInt(){
 
 String ArduinoOTAClass::readStringUntil(char end){
   String res = "";
-  char value;
+  int value;
   while(true){
     value = _udp_ota->read();
-    if(value < '0' || value == '\0' || value == end){
+    if(value < 0 || value == '\0' || value == end){
       return res;
     }
-    res += value;
+    res += static_cast<char>(value);
   }
   return res;
 }
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
   uint8_t index;
   char value;
   while(_udp_ota->peek() == ' ') _udp_ota->read();
-  for(index = 0; index < sizeof(data); ++index){
+  for(index = 0; index < sizeof(data); index++){
     value = _udp_ota->peek();
     if(value < '0' || value > '9'){
       data[index++] = '\0';
       return atoi(data);
     }
-    data[index++] = _udp_ota->read();
+    data[index] = _udp_ota->read();
   }
   return 0;
 }
 
 String ArduinoOTAClass::readStringUntil(char end){
   String res = "";
-  int value;
+  char value;
   while(true){
     value = _udp_ota->read();
-    if(value < 0 || value == '\0' || value == end){
+    if(value < '0' || value == '\0' || value == end){
       return res;
     }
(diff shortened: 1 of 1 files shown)


libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
...
-    data[index++] = _udp_ota->read();
+    data[index] = _udp_ota->read();
   }
   return 0;
 }
 
 String ArduinoOTAClass::readStringUntil(char end){
   String res = "";
-  int value;
+  char value;
```
igrr: What is the motivation for this change?

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
   uint8_t index;
   char value;
   while(_udp_ota->peek() == ' ') _udp_ota->read();
-  for(index = 0; index < sizeof(data); ++index){
+  for(index = 0; index < sizeof(data); index++){
     value = _udp_ota->peek();
     if(value < '0' || value > '9'){
       data[index++] = '\0';
```
Juppit: For me, it would be even clearer to not incrementieren the index here, something like this:
`data[index+1] = '\0';`

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
   uint8_t index;
   char value;
   while(_udp_ota->peek() == ' ') _udp_ota->read();
-  for(index = 0; index < sizeof(data); ++index){
+  for(index = 0; index < sizeof(data); index++){
```
devyte: This change is undesired (++index is better style in C++ in a for loop)

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
   uint8_t index;
   char value;
```
devyte: value should be declared int, because peek() returns an int that can be -1 on error

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
...
   while(_udp_ota->peek() == ' ') _udp_ota->read();
-  for(index = 0; index < sizeof(data); ++index){
+  for(index = 0; index < sizeof(data); index++){
     value = _udp_ota->peek();
     if(value < '0' || value > '9'){
```
devyte: This if statement needs a check against peek error return: if(value < 0 || everythingelse)

libraries/ArduinoOTA/ArduinoOTA.cpp
devyte: value should be declared int, because read() returns an int that can be -1 on error

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
...
 }
 
 String ArduinoOTAClass::readStringUntil(char end){
   String res = "";
-  int value;
+  char value;
   while(true){
     value = _udp_ota->read();
-    if(value < 0 || value == '\0' || value == end){
+    if(value < '0' || value == '\0' || value == end){
```
devyte: This change needs to be undon. The check value < 0 was meant to cover an error return from peek() or read(). Undo.

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -146,26 +146,26 @@ int ArduinoOTAClass::parseInt(){
...
       return res;
     }
-    res += static_cast<char>(value);
+    res += value;
```
devyte: This change needs to be undone, given that value needs to be an int.

libraries/ArduinoOTA/ArduinoOTA.cpp
devyte: This should be data[index] = '\0';

libraries/ArduinoOTA/ArduinoOTA.cpp
lrmoreno007: The motivation is back to original before yoursunny PR (it was char) and homogenize data types ( '/0' and end are char)

PR #4097, opened 2018-01-05
Add gc() and gc_quick()

aplikatika:
Missing functions SPIFFS_gc() and SPIFFS_gc_quick() added to FS.

### DIFF
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -125,6 +125,9 @@ public:
     bool rename(const char* pathFrom, const char* pathTo);
     bool rename(const String& pathFrom, const String& pathTo);
 
+    bool gc(uint32_t size);
+    bool gc_quick();
+
 protected:
     FSImplPtr _impl;
 };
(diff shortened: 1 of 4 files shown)


cores/esp8266/FS.h
```diff
@@ -125,6 +125,9 @@ class FS
     bool rename(const char* pathFrom, const char* pathTo);
     bool rename(const String& pathFrom, const String& pathTo);
 
+    bool gc(uint32_t size);
+    bool gc_quick();
```
igrr: These are SPIFFS specific functions, I don't think they should be present in the generic FS interface.

aplikatika: Agree. Other solution would be making FSImplPtr FS::_impl public or global static instance of SPIFFSImpl. Currently we are unable to call gc() because SPIFFSImpl* _fs is unreachable. What would you recommend?

aplikatika: Ok, since NO_GLOBAL_SPIFFS is available option, I will create my spiffs instance locally. Removing this pull request.

devyte:
@igrr did you mean don't expose them, or did you mean don't expose them in this manner (i.e.: improve naming/interface)? I seem to remember an issue by someone else where this was requested.

PR #4110, opened 2018-01-07
Docs and minor improvements for host/device tests

igrr:
- add tests/README.md with instructions on using host/device tests
- add generation of HTML test report for device tests
- update requirements.txt (remove packages which are installed as dependencies, use latest versions)
- minor makefile fixes and tweaks


igrr:
Reviewing the README might be easier here: https://github.com/esp8266/Arduino/blob/tests_improvements_and_docs/tests/README.md

PR #4112, opened 2018-01-07
Adds digest authentication example.

1995parham:


libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
+#include <ESP8266WiFi.h>
```
igrr: Please add a public domain/CC0 licence header to this file.

https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiEvents/WiFiEvents.ino#L12-L13

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+#include <ESP8266WiFiMulti.h>
+
+#include <ESP8266HTTPClient.h>
+
+#define USE_SERIAL Serial
+
+ESP8266WiFiMulti WiFiMulti;
```
igrr: Do we really need to use `ESP8266WiFiMulti` in this file? I know that some of the existing examples do that, but it would be great if the examples could be made orthogonal.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
+#include <ESP8266WiFi.h>
+#include <ESP8266WiFiMulti.h>
+
+#include <ESP8266HTTPClient.h>
+
+#define USE_SERIAL Serial
```
igrr: Suggest using `Serial` directly

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+
+ESP8266WiFiMulti WiFiMulti;
+
+String _exractParam(String& authReq, const String& param, const char delimit){
+	int _begin = authReq.indexOf(param);
```
igrr: Please load this sketch into Arduino IDE and use Tools > Auto Format feature to fix the indentation.

igrr: Also, it is not a common practice to prefix function and field names with an underscore, when not implementing class members, at least in Arduino.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+	USE_SERIAL.println();
+
+	for(uint8_t t = 4; t > 0; t--) {
+		USE_SERIAL.printf("[SETUP] WAIT %d...\n", t);
+		USE_SERIAL.flush();
+		delay(1000);
+	}
+
+	WiFi.mode(WIFI_STA);
+	WiFiMulti.addAP("Niligo-Prism-yyr9uma", "");
```
igrr: You might want to replace this with something generic, e.g. `"wifi-ssid", "wifi-password"`, and move these settings to the top of the sketch, after includes.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+	if((WiFiMulti.run() == WL_CONNECTED)) {
+
+		HTTPClient http;
+
+		USE_SERIAL.print("[HTTP] begin...\n");
+		// configure traged server and url
+
+
+		// http.begin("http://admin:admin@192.168.100.1/api/state?power=1");
+		http.begin("http://192.168.100.1/api/state?power=0");
```
igrr: Perhaps a URL of an existing service can be used?

For example, http://httpbin.org/ has a test endpoint which accept Digest Auth (`/digest-auth/:qop/:user/:passwd/:algorithm`).

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+		/*
+		// or
+		// or
+		http.begin("http://192.168.1.12/test.html");
+		http.setAuthorization("dXNlcjpwYXN3b3Jk");
```
igrr: Perhaps a comment could be added explaining the `setAuthorization` line? (E.g. how to generate the string)

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+		if (httpCode > 0) {
+			String authReq = http.header("WWW-Authenticate");
+			USE_SERIAL.println(authReq);
+
+			// file found at server
+			String payload = http.getString();
+			USE_SERIAL.println(payload);
+
+			String username = "admin";
+			String password = "admin";
```
igrr: These might also need to be moved to the top of the sketch, together with wifi SSID/password.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+			
+			// extracting required parameters for RFC 2069 simpler Digest
+			// String _realm    = _exractParam(authReq, "realm=\"", '"');
+			String _realm = "Protected";
+			String _nonce    = _exractParam(authReq, "nonce=\"", '"');
+			// String _nonce = "2cadff7e2f160969";
```
igrr: is this commented out line still needed? if not, please remove it.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+			// file found at server
+			String payload = http.getString();
+			USE_SERIAL.println(payload);
+
+			String username = "admin";
+			String password = "admin";
+			
+			// extracting required parameters for RFC 2069 simpler Digest
+			// String _realm    = _exractParam(authReq, "realm=\"", '"');
+			String _realm = "Protected";
```
igrr: Can you explain why it says "extracting required parameters" and then the `extractParam` call is commented out?

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+			String _H2 = md5.toString();
+
+			md5.begin();
+			md5.add(_H1 + ":" + _nonce + ":" + "00000001" + ":" + "gDBuFY4s" + ":" + "auth" + ":" + _H2);
+			md5.calculate();
+			String _response = md5.toString();
+
+			http.begin("http://192.168.100.1/api/state?power=0");
+			USE_SERIAL.println(_nonce);
+			String Author = "Digest username=\"admin\", realm=\"Protected\", nonce=\"" + _nonce + "\", uri=\"/api/state?power=0\", algorithm=\"MD5\", qop=auth, nc=00000001, cnonce=\"gDBuFY4s\", response=\"" + _response + "\"\r\n";
```
igrr: Suggest wrapping this part of the code (starting from extractParam calls) into a function which users can copy into their own sketch.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+			md5.calculate();
+			String _H1 = md5.toString();
+
+			md5.begin();
+			md5.add(String("GET:") + String("/api/state?power=0"));
+			md5.calculate();
+			String _H2 = md5.toString();
+
+			md5.begin();
+			md5.add(_H1 + ":" + _nonce + ":" + "00000001" + ":" + "gDBuFY4s" + ":" + "auth" + ":" + _H2);
```
igrr: What are the `00000001` and `gDBuFY4s` magic values?

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+
+			// parameters for the RFC 2617 newer Digest
+			MD5Builder md5;
+			md5.begin();
+			md5.add(username + ":" + _realm + ":" + password);  // md5 of the user:realm:user
+			md5.calculate();
+			String _H1 = md5.toString();
+
+			md5.begin();
+			md5.add(String("GET:") + String("/api/state?power=0"));
```
igrr: It might be worth storing the URI part in a separate variable and using it both when making HTTP call and here, instead of duplicating.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+			md5.calculate();
+			String _H2 = md5.toString();
+
+			md5.begin();
+			md5.add(_H1 + ":" + _nonce + ":" + "00000001" + ":" + "gDBuFY4s" + ":" + "auth" + ":" + _H2);
+			md5.calculate();
+			String _response = md5.toString();
+
+			http.begin("http://192.168.100.1/api/state?power=0");
```
igrr: Likewise, use a variable for URI to avoid duplication.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
igrr: username and URI are duplicated here again, along with nonces...

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,117 @@
...
+			USE_SERIAL.println(_nonce);
+			String Author = "Digest username=\"admin\", realm=\"Protected\", nonce=\"" + _nonce + "\", uri=\"/api/state?power=0\", algorithm=\"MD5\", qop=auth, nc=00000001, cnonce=\"gDBuFY4s\", response=\"" + _response + "\"\r\n";
+			USE_SERIAL.println(Author);
+			http.addHeader("Authorization", Author);
```
igrr: Please use lowercase names for local variables. Also, the name `author` is a bit confusing, should it be `authorization`?

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: That's awesome thanks, I will use it. 🙌

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: I must remove them, we do not need them anymore.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: You are right, I removed them all.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: In digest authentication client must generate nonce count and client nonce we set them here manually but they can set by creating a random string.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: Yes I created separate variables for them

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,105 @@
...
+
+		md5.begin();
+		md5.add(_H1 + ":" + _nonce + ":" + "00000001" + ":" + "gDBuFY4s" + ":" + "auth" + ":" + _H2);
+		md5.calculate();
+		String _response = md5.toString();
+
+		http.end();
+		http.begin(String(server) + String(uri));
+
+		String authorization = "Digest username=\"admin\", realm=\"" + _realm + "\", nonce=\"" + _nonce + "\", uri=\"" + uri + "\", algorithm=\"MD5\", qop=auth, nc=00000001, cnonce=\"gDBuFY4s\", response=\"" + _response + "\"";
```
igrr: let's move the part starting from `// extracting required parameters for RFC 2069 simpler Diges` to calculation of `authorization` into a separate function, e.g. `getDigestAuth`.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,105 @@
...
+	int httpCode = http.GET();
+
+	if (httpCode > 0) {
+		String authReq = http.header("WWW-Authenticate");
+		Serial.println(authReq);
+
+		// extracting required parameters for RFC 2069 simpler Digest
+		String _realm = exractParam(authReq, "realm=\"", '"');
+		String _nonce = exractParam(authReq, "nonce=\"", '"');
+		String _opaque = exractParam(authReq, "opaque=\"", '"');
```
igrr: As mentioned previously, it is not a common practice in Arduino sketches to start local variable names with underscores.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,105 @@
...
+		md5.calculate();
+		String _H1 = md5.toString();
+
+		md5.begin();
+		md5.add(String("GET:") + String(uri));
+		md5.calculate();
+		String _H2 = md5.toString();
+
+		md5.begin();
+		md5.add(_H1 + ":" + _nonce + ":" + "00000001" + ":" + "gDBuFY4s" + ":" + "auth" + ":" + _H2);
```
igrr: Still some hard-coded strings here.
Nonce `gDBuFY4s` should be randomly generated.
And as far as i understand the RFC, `00000001` should be incremented on each request, which is not the case here.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
igrr: Please break this line, Arduino IDE window is just ~60 characters wide, most of the line will not be readable.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,105 @@
...
+const char* ssidPassword = "........";
+
+const char *username = "admin";
+const char *password = "admin";
+
+const char *server = "http://httpbin.org";
+const char *uri = "/digest-auth/auth/admin/admin/MD5";
+
+String exractParam(String& authReq, const String& param, const char delimit){
+	int _begin = authReq.indexOf(param);
```
igrr: The comment about reformatting the sketch using Arduino IDE is still valid, by the way. Arduino sketches use 2 spaces for indentation, this file uses tabs.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: Yes, I am sorry about them.

1995parham:
@igrr What do you think about digest authentication into ESP8266HTTPClient?

igrr:
Looks pretty good @1995parham! I think moving this feature into ESP8266HTTPClient class could be useful. It's up to you whether to do it or not, adding it in the form of an example is also okay with me.

I have only one remaining comment about the request counter field, "00000001". I think you should pass the counter as an argument into `getDigestAuth` function, and increment it each time the loop runs. (Perhaps the web servers also allow sending the same counter, that's just my reading of the RFC).

1995parham:
@igrr Yes, I add this feature to the `getDigestAuth` function.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,133 @@
...
+}
+
+String getDigestAuth(String& authReq, const String& username, const String& password, const String& uri, int counter) {
+  // extracting required parameters for RFC 2069 simpler Digest
+  String realm = exractParam(authReq, "realm=\"", '"');
+  String nonce = exractParam(authReq, "nonce=\"", '"');
+  String cNonce = getCNonce(8);
+
+  char nc[8];
+  sprintf(nc, "%08d", counter);
```
igrr: If that's "%08d", the size of `nc` should be 9 bytes to accommodate the zero terminator.

devyte: Counter is an int, so nc should have size 12. Then, the format specifier should be %11d.
Counter should actually be an unsigned int imo, it doesn't make sense for negative values. In that case, nc should have size 11.

igrr: Ok, let's read the [RFC](https://tools.ietf.org/html/rfc2617):

     The nc-value is the hexadecimal
     count of the number of requests (including the current request)
     that the client has sent with the nonce value in this request.  For
     example, in the first request sent in response to a given nonce
     value, the client sends "nc=00000001".

and

     nc-value         = 8LHEX

So it must be "%08x".

1995parham: yes, you are right, it must be in hex format so I convert it into `%08x`.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
```diff
@@ -0,0 +1,133 @@
...
+    return s;
+}
+
+String getDigestAuth(String& authReq, const String& username, const String& password, const String& uri, int counter) {
+  // extracting required parameters for RFC 2069 simpler Digest
+  String realm = exractParam(authReq, "realm=\"", '"');
+  String nonce = exractParam(authReq, "nonce=\"", '"');
+  String cNonce = getCNonce(8);
+
+  char nc[8];
```
igrr: Sorry, nc is still an 8 byte buffer, so the zero terminator will be written out of array bounds. Could you please change the array size to 9, and use `snprintf(nc, sizeof(nc), "%08x", counter);`?

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
devyte: Agreed. But what stresses that counter should be unsigned :P

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: Yes sorry for these mistakes.

libraries/ESP8266HTTPClient/examples/DigestAuthorization/DigestAuthorization.ino
1995parham: @devyte I corrects that too :)

1995parham:
@igrr Thanks a lot

PR #4119, opened 2018-01-08
working code for sigma delta generator

stefaandesmet2003:
provide a simple interface to attach/detach pins to the sigma-delta generator, and get/set the 2 parameters prescaler & target


devyte:
@stefaandesmet2003 this is for #4119 , yes? I see you figured it out :)

stefaandesmet2003:
Yes, indeed

cores/esp8266/sigma_delta.h
```diff
@@ -18,15 +18,112 @@
...
+2. sigma_delta_attachPin(pin), any pin 0..15, TBC if gpio16 supports sigma-delta source
+     This will set the pin to NORMAL output mode (pinMode(pin,OUTPUT))
+3. sigma_delta_setPrescaler(uint8_t) : reduce the output frequencies
+4. sigma_delta_setTarget(uint8_t) : set the output signal duty cycle, duty cycle = target/256
+
+5. sigma_delta_detachPin(pin), this will revert the pin to NORMAL output mode & GPIO source. 
+The sigma delta source remains on until :
+6. sigma_delta_disable()
+
+copy this code as example :
```
devyte: This example should be made into a working example that builds with the rest of the core, so that at least build issues can be tracked.

stefaandesmet2003: Good idea. The example would then be under libraries\esp8266\examples ?

libraries/esp8266/examples/SigmaDeltaDemo/SigmaDeltaDemo.ino
```diff
@@ -0,0 +1,84 @@
+/*
+  Sigma Delta Generator example
+  
+  This example demonstrates the use of the ESP8266 internal hardware sigma delta source.
+  Each GPIO pin or multiple pins can be connected to the one sigma delta source.
+  THE sigma delta output frequency is defined by the ESP8266 hardware as:
+
+  FREQ = 80,000,000/prescaler * target /256  HZ,          0<target<128
+  FREQ = 80,000,000/prescaler * (256-target) /256  HZ,    128<target<256
```
earlephilhower: Are these constant even in 2x mode where CPU_F = 160,000,000?

libraries/esp8266/examples/SigmaDeltaDemo/SigmaDeltaDemo.ino
```diff
@@ -0,0 +1,84 @@
...
+  - the sigma delta source produces much higher frequencies. This allows the output to be easily smoothed by a RC-filter and realise a decent 8-bit DAC
+  - the arduino analogWrite pwm has a higher resolution, default 12-bit, and up to 16-bit resolution. The sigma delta source has only 8-bit resolution
+    
+  The example doesn't require additional hardware.
+  
+  With an RC-filter (for example R=17k & C=100nF), a 'DC' output signal can be obtained with voltage between 0 and 3.3V
+  For higher value capacitors, it might be necessary to add a high-speed buffer between the esp8266 output and RC-filter to limit the current ac load on the 
+  pin output driver.
+
+*/
```
earlephilhower: Could you please add attribution and license (CC0 or PD) to the example header?

sigma_delta.h
```diff
@@ -0,0 +1,72 @@
...
+void sigma_delta_enable(void);
+void sigma_delta_disable(void);
+void sigma_delta_attachPin(uint8_t pin);
+void sigma_delta_detachPin(uint8_t pin);
+bool sigma_delta_isPinAttached(uint8_t pin);
+uint8_t sigma_delta_getTarget(void);
+void sigma_delta_setTarget(uint8_t target);
+uint8_t sigma_delta_getPrescaler(void);
+void sigma_delta_setPrescaler(uint8_t prescaler);
+
```
earlephilhower: This is good, but I think to be most useful to end users, who may not be as familiar w/ clock dividers and sigma-delta operation a simpler API might be helpful.

I'd recommend renaming "setTarget/GetTarget" to simply, "write" or "read" as effectively they're trying to write an analog value and "target," while the name used by the hardware designers in the HDK, doesn't really mean much if you don't understand the error-diffusion behind things.

I'd also suggest a simpler "setFrequency(long f) setter method that would do the math described in the header and figure out the closest prescaler given the F_CPU.  That'd make end-user code simpler to understand.

libraries/esp8266/examples/SigmaDeltaDemo/SigmaDeltaDemo.ino
earlephilhower: Could you please add attribution and license (CC0 or PD) to the example header?

libraries/esp8266/examples/SigmaDeltaDemo/SigmaDeltaDemo.ino
stefaandesmet2003: Good point, i never tested 160MHz. Will verify and correct accordingly!

libraries/esp8266/examples/SigmaDeltaDemo/SigmaDeltaDemo.ino
stefaandesmet2003: PD i guess?

sigma_delta.h
stefaandesmet2003: Not sure if I fully understand the setFrequency idea. Every call to 'write' (setTarget) would need to adapt the prescaler as well to fit this chosen frequency? And remove the setPrescaler/getPrescaler functions altogether?
There was some sample code like this in an earlier post to this issue.  In many cases the effective output frequency is way off the chosen frequency f. Suppose you chose 1 MHz. With 50% duty cycle the closest you get is 312.5kHz. So you give the user the impression that the generator can be used as a fixed frequency source, but in fact it cannot.

earlephilhower: Good point, I withdraw this silly request. :)  I was thinking this would be useful for someone to, say, drive a 4-pin PWM fan where they'd need to set frequency = 25khz, duty = %age, but it seems it's not really precise enough for such a use, anyway.

For the examples, please do add the Public Domain attribution.  I think the team is trying to get these examples cleaned up and that's an easy thing to do.

libraries/esp8266/examples/SigmaDeltaDemo/SigmaDeltaDemo.ino
stefaandesmet2003: Checked : CPU frequency has no influence on the SD frequencies

sigma_delta.h
stefaandesmet2003: Apparently ESP32 has an implementation similar to what you suggested. I modified the code to be compatible. ESP32 has apparently 8 independent channels, esp8266 has only 1. Apart from that, the behavior is the same (eg. in terms of output frequencies.) So it makes sense to align the programming interfaces.
Arduino-esp32 also provides a working sigma delta example; there it's not necessary to explicitly include the sigma_delta.h file in the example code; it would be convenient if this was the same in arduino-esp8266, but my example currently doesn't work without the explicit include

cores/esp8266/core_esp8266_sigma_delta.c
```diff
@@ -104,51 +76,77 @@ bool ICACHE_FLASH_ATTR sigma_delta_isPinAttached(uint8_t pin)
...
+  
 }
-
 /******************************************************************************
- * FunctionName : sigma_delta_getPrescaler
- * Description  : get the prescaler value from the GPIO_SIGMA_DELTA register
- * Parameters   : none
- * Returns      : uint8 prescaler, CLK_DIV , 0-255
+ * FunctionName : sigmaDeltaRead
+ * Description  : set the duty cycle for the sigma-delta source
```
earlephilhower: Typo here...this "get the duty..."

cores/esp8266/core_esp8266_sigma_delta.c
```diff
@@ -0,0 +1,152 @@
+#include "Arduino.h" // using pinMode
```
earlephilhower: Needs standard licence header, please.

cores/esp8266/core_esp8266_sigma_delta.c
```diff
@@ -0,0 +1,152 @@
...
+ * Parameters   : none
+ * Returns      : none
+*******************************************************************************/
+void ICACHE_FLASH_ATTR sigmaDeltaDisable()
+{
+  GPSD = (0 << GPSDT) | (0 << GPSDP) | (0 << GPSDE); //SIGMA_DELTA_TARGET(0) | SIGMA_DELTA_PRESCALER(0) | SIGMA_DELTA_ENABLE(DISABLED)
+}
+
+/******************************************************************************
+ * FunctionName : sigma_delta_attachPin
```
earlephilhower: Some FunctionName comments got updated to camelCase, but some like this did not.

sigma_delta.h
```diff
@@ -0,0 +1,72 @@
...
+5. sigma_delta_detachPin(pin), this will revert the pin to NORMAL output mode & GPIO source. 
+The sigma delta source remains on until :
+6. sigma_delta_disable()
+
+*******************************************************************************/
+
+#ifndef SIGMA_DELTA_H
+#define SIGMA_DELTA_H
+
+#ifdef __cplusplus
```
earlephilhower: The function prototypes here are "extern "C""'d, but the actual function definitions aren't.  Any reason not to drop the "extern C" here?

stefaandesmet2003: linking stage doesn't work without this. Maybe you know why?
/SigmaDeltaDemo.ino:16: undefined reference to `sigmaDeltaGetPrescaler()'
and the same for all other sigmaDelta functions

earlephilhower: I just noticed you've got the file named "sigma_delta.**c**" and not "sigma_delta.**cpp**".  Can you "git mv" the file, that'll remove the need for this construction, and allow the better checking of the CPP compiler to be used, too.

earlephilhower: Ignore my previous comment.  The other "core_esp8266_*" files are all .**c**.  So ignore the request to remove extern C, leave it there.  Sorry for the confusion!

cores/esp8266/core_esp8266_sigma_delta.c
stefaandesmet2003: ok done

sigma_delta.h
stefaandesmet2003: Hi @earlephilhower ,
I don't manage to get the style_check right on the travis build. Could you have a look what the issue is, please?
The other builds are ok now.

earlephilhower: Will do tonight.  I think it's just a matter of getting the examples in Arduino IDE format (i.e. 2 space indent, brackets in the right spot, etc.).

earlephilhower: Looks like your problem is DOS linefeeds.  Can you save it with Unix ones, that seems to be what IGRR's checker requires to work properly.  I can't do it for you as I don't have write permission to your source repo (nor should I, of course!).

earlephilhower:
I can't upload to your repo, but I can give you the file to upload yourself.  Rename to .ino and replace your existing sample and I think you'll be all set:

[SigmaDeltaDemo.ino.txt](https://github.com/esp8266/Arduino/files/1799007/SigmaDeltaDemo.ino.txt)

stefaandesmet2003:
thanks for your help.
after 2 more attempts it still fails.
I quit,sorry.

earlephilhower:
@stefaandesmet2003 Sorry for the run around!  The Windows/Unix line ending difference is just breaking things for you.

I'll try and pull your branch into my Arduino fork, fix it, and open a linked bug just to get this done.

earlephilhower:
Just committed your changes, closing this PR.

PR #4131, opened 2018-01-10
Update readme.rst

jonhp:
fix broken link to file system flash layout


Rotzbua:
the link is correct, no change required

jonhp:
If I go to: https://arduino-esp8266.readthedocs.io/en/latest/ota_updates/readme.html
and then click on the "flash layout" link in  "Keep in mind that the File system and EEPROM for example needs space too (one time) see flash layout." which is https://arduino-esp8266.readthedocs.io/en/latest/filesystem.rst#flash-layout
then I get a page that says 
` \          SORRY            /
         \                         /
          \    This page does     /
           ]   not exist yet.    [    ,'|`
That is not the result that I expect. 
If I manually change the ".rst" to ".html" then the page displays as I would expect.
Maybe my proposed fix was incorrect, but I have now fully demonstrated the problem which is repeatable.

Rotzbua:
You observed correctly that `rst` != `html`, but you look at the rendered page. The source are `.rst` files which are rendered by a compiler and are served as static `html` to an server.

So the source files are correct with `rst`, so it seems there is something wrong with the compiler or settings. Well some month ago @igrr changed all files from `.md` to `.rst` and I guess he forgot to change the configuration of readthedocs.io .

pfeerick:
The RST file you modified should have ``` :ref:`Flash layout` ``` instead of ``` `flash layout <../filesystem.html#flash-layout>`_ ``` and then it _should_ be fine, as it should then find the matching label and serve it properly. (Link to an external RST label).

jonhp:
I took a shot at fixing this again and an on-page link. Not sure if I am any closer and not sure how to test these fixes either (hints anyone?).

pfeerick:
AFAIK, that looks right. If you really want to validate it without waiting to see if someone can check it, it may be possible to follow the instructions [here](https://docs.readthedocs.io/en/latest/getting_started.html#in-rst) to get a test environment running, but going by the [explicit external links reference](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#explicit-links), it seems like it should work ('cheat sheet' that I refer to whenever I need to learn more about reStructuredText formatting and markup).

devyte:
Alright, if this is broken, it won't be worse than it already is, so I'm merging. If it's still broken afterwards, please take another shot at fixing the links.

lrmoreno007:
" :ref: " is other way to make cross reference not discussed in #4409 . This need a label in the referenced file ("/doc/filesystem.rst") http://www.sphinx-doc.org/es/stable/markup/inline.html .

EXAMPLE:
In the file filesystem.rst:
```

Filesystem
==========

.. _flash-layout:

Flash layout
------------

Even though file system is stored on the same flash ......................
```

In the file /doc/ota_updates/readme.rst:

```

.......... EEPROM for example needs space too (one time) see :ref:`flash-layout`.

...........

```

But this work fine in Sphinx (ReadTheDocs) but not in Github.

lrmoreno007:
And with `flash layout <../filesystem.html#flash-layout>`_ , will work too in Sphinx but not in Github.

PR #4135, opened 2018-01-11
Rework arduino arg order compatibility in WiFiSTAClass::config() to fix #4114

### LINKED ISSUE #4114: PR #3860 broke DHCP client Wifi config
#### Hardware
Hardware:		ESP12
Core Version:      	master

### Description

As far as I can tell and experience, the change in #3860 broke DHCP. In order to get the dhcp client started one has to set the subnet to 0.0.0.0 in ESP8266WiFiSTAClass::config to pass this check: https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L242
and the new code (https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L221-L235) makes it impossible to get subnet==0.

devyte:


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -197,55 +197,57 @@ wl_status_t ESP8266WiFiSTAClass::begin() {
...
+        return true;
+    }
+
+
+    //To allow compatibility, check first octet of 3rd arg. If 255, interpret as ESP order, otherwise Arduino order.
+    IPAddress gateway = arg1;
+    IPAddress subnet = arg2;
+    IPAddress dns1 = arg3;
+
     if(subnet[0] != 255)
```
d-a-v: We could here check if ip and gateway are in the same network (ip & netmask) == (gateway & netmask) which gives :
- arduino if: (local_ip & arg3) == (arg2 & arg3)
- esp if: (local_ip & arg2) == (arg1 & arg2)

devyte: Agreed about the check, but I would add it later....

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -197,55 +197,57 @@ wl_status_t ESP8266WiFiSTAClass::begin() {
...
-      if(dns1[0] == 0)
-      {
-        //arg order is arduino and 4th arg not given => assign it arduino default
-        dns1 = IPAddress(255,255,255,0);
-      }
-
-      //current order is arduino:                   ip-dns-gway-subnet
-      swap(gateway, subnet); //after this, order is ip-gway-dns-subnet
-      swap(subnet, dns1);    //after this, order is ip-gway-subnet-dns (correct ESP order)
-    }
```
devyte: ... like here. At this point we already figured out the arduino compatibility, so it's a good place to check that.

devyte:
Something got messed up with git, and now I can't push further changes. Closing this, will redo.

PR #4138, opened 2018-01-11
Adding softAP SSID & PSK query API

IMAN4K:
Signed-off-by: Iman Ahmadvand iman72411@gmail.com (github: @IMAN4K)


libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -333,3 +333,23 @@ String ESP8266WiFiAPClass::softAPmacAddress(void) {
...
+}
+
+/**
+* Get the configurd(Not-In-Flash) softAP PSK or PASSWORD.
+* @return String psk.
+*/
+String ESP8266WiFiAPClass::softAPPSK() const {
+	struct softap_config config;
+	wifi_softap_get_config(&config);
+	return String(reinterpret_cast<char*>(config.password));
```
devyte: On second thought, this could cause an issue. When a 64-byte psk is used, it is not 0-terminated, and String assumes a null-terminator.
Instead, a char psk[65] should be declared, config.password should be memcpy()d into it, and then psk[64] = 0;. Then, return String(psk).

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -333,3 +333,23 @@ String ESP8266WiFiAPClass::softAPmacAddress(void) {
...
+* @return String SSID.
+*/
+String ESP8266WiFiAPClass::softAPSSID() const {
+	struct softap_config config;
+	wifi_softap_get_config(&config);
+	return String(reinterpret_cast<char*>(config.ssid));
+}
+
+/**
+* Get the configurd(Not-In-Flash) softAP PSK or PASSWORD.
```
devyte: Minor typo -> configured.

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
IMAN4K: @devyte 
Yes That's right.
This is the `String` constructor:
```
String::String(const char *cstr) {
    init();
    if(cstr)
        copy(cstr, strlen(cstr));
}
```

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -333,3 +333,33 @@ String ESP8266WiFiAPClass::softAPmacAddress(void) {
...
 }
+
+/**
+* Get the configured(Not-In-Flash) softAP SSID name.
+* @return String SSID.
+*/
+String ESP8266WiFiAPClass::softAPSSID() const {
+	struct softap_config config;
+	wifi_softap_get_config(&config);
+	char* name = reinterpret_cast<char*>(config.ssid);
```
d-a-v: why not just 
```
config.ssid[sizeof(config.ssid) - 1] = 0; // guard
return String((const char*)(void*)config.ssid) ?
```
String(char*) will anyway search for the last \0

IMAN4K: @d-a-v 
Based on @devyte  comment user must be able to configure `64` byte length `PSK` at max, not `63`

d-a-v: OK sorry I did not see the hidden-outdated comment.

libraries/ESP8266WiFi/src/ESP8266WiFiAP.h
```diff
@@ -47,6 +47,9 @@ class ESP8266WiFiAPClass {
         uint8_t* softAPmacAddress(uint8_t* mac);
         String softAPmacAddress(void);
 
+		String softAPSSID() const;
```
igrr: Indentation on this line is different from nearby lines

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -333,3 +333,33 @@ String ESP8266WiFiAPClass::softAPmacAddress(void) {
     sprintf(macStr, "%02X:%02X:%02X:%02X:%02X:%02X", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
     return String(macStr);
 }
+
+/**
+* Get the configured(Not-In-Flash) softAP SSID name.
+* @return String SSID.
+*/
+String ESP8266WiFiAPClass::softAPSSID() const {
+	struct softap_config config;
```
igrr: Indentation in this function is different from nearby functions

cores/esp8266/IPAddress.cpp
```diff
@@ -112,4 +112,51 @@ String IPAddress::toString() const
     return String(szRet);
 }
 
+/* ref: https://code.woboq.org/userspace/glibc/resolv/inet_pton.c.html#inet_pton4 */
+static bool inet_pton4(const char *start, const char *end) {
```
devyte: How is this function different from IPAddress::fromString(), found [here](https://github.com/esp8266/Arduino/blob/f4c391032aff382bcd243bb038d69feb136a96b5/cores/esp8266/IPAddress.cpp#L43)? If they serve equivalent purpose, I'd rather not duplicate.

devyte:
@IMAN4K will you be replying to this?

cores/esp8266/IPAddress.cpp
IMAN4K: @devyte 
So sorry for late reply.
So you say we should modify `bool IPAddress::fromString(const char *address)` function for adding validate support ? without duplicate

IMAN4K: @devyte 
My suggestion:
1.Adding a private validating method:
`static bool IPAddress::isValid(const char* arg, size_t len);`
2.Adding correspond public method:
`static bool IPAddress::isValid(const String& ip);`
3.Reuse method of 1 inside 2 and `bool IPAddress::fromString(const char *address)`

devyte: @IMAN4K said: 
> Because current IPAddress public API doesn't support validating IPV4 as string

Are you saying that the following does not validate a string?
`bool isvalid = IPAddress().fromString("17604123874106239");`

IMAN4K: @devyte 
Yes that's correct, my mistake!
But we need to make an object for every validation.
It's better with a public static member, then every where in code:
`if (IPAddress::isValid(...)) ...`

devyte: The object constructed is 4 bytes in size, essentially one int, which is a temporary instantiated on the stack and immediately destroyed, so that's not a problem (btw, destruction is trivial). 
What I suggest is a static convenience method like you propose, but something like this:
```
bool IPAddress::isValid(const char *str)
{
  return IPAddress().fromString(str);
}
bool IPAddress::isValid(String &str)
{
  return IPAddress().fromString(str);
}

```

IMAN4K: Ok.
I'll make changes very soon.

PR #4083, opened 2018-01-03
Add docs for Adafruit Feather HUZZAH ESP8266

nrobinson2000:
Adafruit Feather HUZZAH ESP8266 (ESP-12S)
-----------------------------------------

The Adafruit Feather HUZZAH ESP8266 is an Arduino-compatible Wi-Fi development board powered by Ai-Thinker's ESP-12S, clocked at 80 MHz at 3.3V logic. A SiLabs CP2104 USB-Serial chip is included so that you can upload code at 921600 baud for fast development time. It also has auto-reset so no noodling with pins and reset button pressings. A 3.7V Lithium polymer battery connector is included, making it ideal for portable projects. The Adafruit Feather HUZZAH ESP8266 will automatically recharge a connected battery when USB power is available.

Product page: https://www.adafruit.com/product/2821

PR #4143, opened 2018-01-12
Case change and and 'none' reset option for Digistump Oak

pfeerick:
The logo is 'digiStump', but is always written in text form as 'Digistump'. 

The generator fixed an issue with the release version of 2.4.0 not having a resetMethod for the Digistump Oak, and hence not being able to be uploaded to without modifying the boards.txt file. However, the Oak doesn't support auto-reset via serial - it was primarily designed to be a cloud-programmed board. Hence the addition of another resetMacro macro.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -3320,7 +3320,7 @@ gen4iod.menu.UploadSpeed.921600=921600
 gen4iod.menu.UploadSpeed.921600.upload.speed=921600
 
 ##############################################################
-oak.name=DigiStump Oak
+oak.name=Digistump Oak
 oak.build.board=ESP8266_OAK
 oak.build.variant=oak
 oak.upload.maximum_size=1040368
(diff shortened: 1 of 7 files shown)


pfeerick:
Damn github, should've done the changes to the local repo and pushed it from there, instead of editing directly on the github page in the first instance.

pfeerick:
I noticed some minor formatting corrections that I was going to do related to the Oak section of docs/boards.rst (title, fix link to serial adapter heading), but I'm not sure if the `boards.txt.py` generation is working properly or if boards.rst has been generated by it yet? As what should have been only a change to two lines has done a LOT of changes when running `--docgen` ?

d-a-v:
You are right, ```boards.rst``` is still the original one and has not yet been generated. The generator is quite fresh and the only generated data we use so far is ```boards.txt```.
This explains the lot of changes.
So while you are at it, then yes please do the commit.

pfeerick:
Just mentioning this PR here as the regen of boards.rst, if all is good, should close #4083 as well.

pfeerick:
Just a few final bits! :-O

PR #4145, opened 2018-01-12
Rework of arduino compatibility in WiFiSTAClass::config

devyte:


devyte:
No idea why the diff took the entire function.
@d-a-v I added your suggestion to check that gateway and ip are in the same subnet.

devyte:
Fixes #4114

d-a-v:
You are damn right. The test I proposed is not sufficient.
I don't feel right testing netmask[0][&255]==255 because it's a particular case (why not nm[0]&0b11111000==0b11111000)(248) or with 224, 240, 252...
So I should split hairs no more, testing the first netmask's byte would be sufficient and common enough.

PR #4148, opened 2018-01-12
add begin(port) to esp8266webserver, move some strings to flash, some refactoring

devyte:


devyte:
I'm including some string stuff here: moving strings to flash, remove duplicates, and so on.
WIP

devyte:
Alright, moving all the strings and so on saves 368b of heap on build with the HttpBasicAuth example, and increases the bin size by 496b. I actually expected the bin size to decrease as well, given that I removed several duplicate strings. No idea why it went the opposite direction.

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -127,74 +137,74 @@ bool ESP8266WebServer::authenticate(const char * username, const char * password
...
         authReq = String();
         return false;
       }
       // parameters for the RFC 2617 newer Digest
       String _nc,_cnonce;
-      if(authReq.indexOf("qop=auth") != -1){
-        _nc = _exractParam(authReq,"nc=",',');
-        _cnonce = _exractParam(authReq,"cnonce=\"");
+      if(authReq.indexOf(FPSTR(qop_auth)) != -1) {
+        _nc = _extractParam(authReq, F("nc="), ',');
```
earlephilhower: Thanks for fixing this typo in the method name!  _exractParam always gave me a headache!

devyte:
Any other comments? corrections? suggestions? observations? indications? tribulations?

earlephilhower:
The mimetable array and enum being in separate files will be a pain to keep in sync, if we ever add things (but that seems unlikely).  Only [gz] is used, so why not just hardcode ".gz".  It's not likely that the standard compression format identifier is going to change in the lifetime of this code (famous last words).  You could drop the enum then, no?

You were very thorough in moving strings to PMEM, but as .rodada does dedupe, I'm not sure you gained anything from changing "" to F("")  (and in some spots "" -> String()).

devyte:
About the mimetable in its own file, I thought it might be useful for some users. I even need it myself for something stupid. Maybe it shouldn't be under details directory, but if it's to be moved, I'm thinking do it later.
About the "", that was to soothe my OCD (it's done elsewhere). 
... and now that I just had another look, I seem to have missed one, and my right eye is twitching :P
Assuming no further comments, I'll do a couple of experiments to check size changes for heap and flash usage, maybe do an Arduino autoformat, and then merge.

devyte:
@earlephilhower looks like you were right: adding F() to short strings like "" or "/r/n" doesn't improve heap, and increases bin size. I'm gonna guess that the inflection point is strings of size 4, so 3 chars. 
In other words, strings of 4 chars or more should be moved to flash, while strings shorter than that should not.

More importantly: is the mimetable correctly declared to hold everything in flash?

earlephilhower:
@devyte Looks like your worries were well founded.  I thought your construction would push strings into PROGMEM, but you've only put the *pointers* to the strings, still in .RODATA, into PROGMEM. :(

````
$ objdump -s -j .rodata /tmp/arduino_build_25960/*.elf 
....
 3ffe8880 b49dd979 4378ea44 00000000 00002041  ...yCx.D...... A
 3ffe8890 74657874 2f706c61 696e000d 0a002e73  text/plain.....s
 3ffe88a0 72630074 6578742f 68746d6c 002e6373  rc.text/html..cs
 3ffe88b0 73007465 78742f63 7373002e 6a730061  s.text/css..js.a
 3ffe88c0 70706c69 63617469 6f6e2f6a 61766173  pplication/javas
 3ffe88d0 63726970 74002e70 6e670069 6d616765  cript..png.image
 3ffe88e0 2f706e67 002e6769 6600696d 6167652f  /png..gif.image/
 3ffe88f0 67696600 2e6a7067 00696d61 67652f6a  gif..jpg.image/j
 3ffe8900 70656700 2e69636f 00696d61 67652f78  peg..ico.image/x
 3ffe8910 2d69636f 6e002e78 6d6c0074 6578742f  -icon..xml.text/
 3ffe8920 786d6c00 2e706466 00617070 6c696361  xml..pdf.applica
 3ffe8930 74696f6e 2f706466 002e7a69 70006170  tion/pdf..zip.ap
 3ffe8940 706c6963 6174696f 6e2f7a69 70002f69  plication/zip./i
 3ffe8950 6e646578 2e68746d 00646f77 6e6c6f61  ndex.htm.downloa
 3ffe8960 64006170 706c6963 6174696f 6e2f6f63  d.application/oc
 3ffe8970 7465742d 73747265 616d0053 656e7420  tet-stream.Sent 
 3ffe8980 6c657373 20646174 61207468 616e2065  less data than e
 3ffe8990 78706563 74656421 00474554 00504f53  xpected!.GET.POS
 3ffe89a0 54005344 43415244 204e6f74 20446574  T.SDCARD Not Det
````

Your code looks good.  I do something similar to store about 500 different timezones in my psychoplug and it just worked there.  Let me investigate a bit...

devyte:
The mimetable is an array of structs with 2 string members, which is different than an array of pointer to strings, which is why that last look made my spidey sense tingle...
I'm not sure what the declaration is supposed to look like here.

earlephilhower:
You missed my updated comment.  You're all good, the strings are actually all repeated in the SDWebServer example that I compiled to test.  So, no problems w/the changes, but we may want to refactor the examples at some point...

devyte:
Oh, I did miss it. Great! Nothing further to do here then :)

PR #4150, opened 2018-01-12
lower lwip2 ram and flash footprint (by disabling old asserts)

d-a-v:
also restore zalloc() (lost with OOM debug commit, used by lwip/lwip2)


reaper7:
👍 
lost zalloc, used by https://github.com/i-n-g-o/esp-mqtt-arduino and https://github.com/martin-ger/uMQTTBroker 
that's why they did not compile :)

PR #4160, opened 2018-01-13
Move to PROGMEM aware libc, allow PSTR in printf()

earlephilhower:
A Newlib (libc) patch is in progress to move the _P functions from inside
Arduino into first-class citizens in libc.  This Arduino patch cleans up
code that's been migrated there.  Binaries for the new libs are included
because it seems they're part of the Arduino git tree, and should be
replaced with @igrr built ones when/if the Newlib changes are accepted.
https://github.com/igrr/newlib-xtensa/pull/5 

Notable changes/additions for Arduino:
Allow for use of PROGMEM based format and parameter strings in all
*printf functions.  No need for copying PSTR()s into RAM before printing
them out (transparently saves heap space when using _P functions) and
makes it easier to print out constant strings for applications.

Add "%S" (capital-S) format that I've been told, but cannot verify,
is used in Arduino to specify a PROGMEM string parameter in printfs,
as an alias for "%s" since plain "%s" can now handle PROGMEM.

Optimized the memcpy_P and strncpy_P functions to use 32-bit direct reads
whenver possible (source and dest alignment mediated), but there is
still room for improvement in others like *_P.

Finally, move several constant arrays from RODATA into PROGMEM and
update their accessors.  Among these are the ctype array, ~260 bytes,
mprec* arrays, ~300 bytes, and strings/daycounts in the time
formatting functions, ~200 bytes.  All told, sketches will see from
300 to 800 additional RAM heap free on startup (depending on their
use of these routines).


earlephilhower:
The builds succeeded, but the host_tests failed due to standard x86 libc not having the new sys/pgmspace.h include file that was added to newlib.  The build job would need to have the default, no-op newlib/libc/include/sys/pgmspace.h defined in the newlib patch submitted to newlib-xtensa.

devyte:
This sounds really cool, but moving things to libc feels weird to me, mostly because it'd mean we'd have a non-standard libc. I just did a quick search, and I realize that in the AVR world this is apparently nothing new, but it still feels weird.
Therefore, I can only give an opinion on style and coding, and not on purpose.
I did a review on the libc repo, linked above.

earlephilhower:
Latest commit includes some comments from newlib repo, as well as accelerated strnlen_P code/

tools/sdk/libc/xtensa-lx106-elf/include/locale.h
```diff
@@ -1,7 +1,7 @@
 /*
 	locale.h
 	Values appropriate for the formatting of monetary and other
-	numeric quantities.
+	numberic quantities.
```
d-a-v:
```suggestion
	numeric quantities.
```

earlephilhower: I can do this, but the problem stems from the newlib itself, from where this file is copied.  Doing the change in the newlib repo is easy, but it's unfortunately very hard to get igrr's time to look at PRs and merge them there.

For something like this, I'd rather not bother or bite the bullet and say we're using newlib from another repo (I can share commit permissions on mine, make a new one somewhere, etc.) and include it as an SDK directory module (like lwip and bssl).

liebman: Personally I like include as module - the less "binary" includes the better.

earlephilhower: It still wouldn't compile from the Arduino environment.  You **really** wouldn't want to build libc single-threaded every application, believe me!  But it would be kind of like BearSSL where there'd be a `make install` target that would automate updates to the core .a dirs.

d-a-v: About the initial trivial fix, I just experimented the inline PR beta github feature. Sorry for the noise.

tests/device/test_libc/memmove1.c
```diff
@@ -0,0 +1,194 @@
+/* A minor test-program for memmove.
```
devyte: The name of this file looks inconsisten vs. the others of its kind (memmove1.c vs. e.g.: memcpy-1.c).
Minor, but better addressed now than later :)

tests/host/Makefile
```diff
@@ -96,7 +96,7 @@ valgrind: $(OUTPUT_BINARY)
 	$(LCOV) --directory ../../cores/esp8266/ --zerocounters
 	$(VALGRIND) $(VALGRINDFLAGS) $(OUTPUT_BINARY)
 	$(LCOV) --directory $(CORE_PATH) --capture --output-file $(LCOV_DIRECTORY)/app.info
-	$(GENHTML) $(LCOV_DIRECTORY)/app.info -o $(LCOV_DIRECTORY)
+	-$(GENHTML) $(LCOV_DIRECTORY)/app.info -o $(LCOV_DIRECTORY)
```
devyte: Is the hyphen here on purpose?

devyte:
I didn't go in-depth into the code bits. I assume by the extensive device tests that functionality is sufficiently well covered.
Just a minor nitpick and a question, other than that, LGTM!

tests/device/test_libc/memmove1.c
earlephilhower: I'd rather leave the name as-is, because this is the actual name of a test in newlib self-test.

tests/host/Makefile
earlephilhower: Yes, it's on purpose because unfortunately LCOV throws an error due to the way we're using two different directories.  It runs and generates coverage properly, but complains that certain things can't be found (but it finds them elsewhere).  I wasn't able to figure out how to get rid of this w/o significant directory tree rework, which IMO's not worth the gried.

earlephilhower:
Closing this as it is now part of the toolchain update in PR#4160 .

earlephilhower:
Err, I mean it's in subsumed into PR #5376

PR #4171, opened 2018-01-15
Use fixed TCP port in ArduinoOTA to facilitate firewall configuration

kohlerjl:
Use a fixed network port for TCP firmware transfer in ArduinoOTA. This step of the transfer is not clearly documented, and presents an issue for systems behind a firewall. A fixed port allows configuring the developer computer's firewall or forwarding on a router to accept the connection from the esp device to download the new firmware, after authentication via UDP. Perhaps this could be made configurable.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -118,7 +118,7 @@ tools.esptool.upload.protocol=esp
 tools.esptool.upload.params.verbose=-vv
 tools.esptool.upload.params.quiet=
 tools.esptool.upload.pattern="{path}/{cmd}" {upload.verbose} -cd {upload.resetmethod} -cb {upload.speed} -cp "{serial.port}" -ca 0x00000 -cf "{build.path}/{build.project_name}.bin"
-tools.esptool.upload.network_pattern="{network_cmd}" "{runtime.platform.path}/tools/espota.py" -i "{serial.port}" -p "{network.port}" "--auth={network.password}" -f "{build.path}/{build.project_name}.bin"
+tools.esptool.upload.network_pattern="{network_cmd}" "{runtime.platform.path}/tools/espota.py" -i "{serial.port}" -p "{network.port}" -P "8266" "--auth={network.password}" -f "{build.path}/{build.project_name}.bin"
 
 tools.mkspiffs.cmd=mkspiffs
 tools.mkspiffs.cmd.windows=mkspiffs.exe


lrmoreno007:
It does not make sense to completely fix the port on platform.txt.
It may make sense to you, but not to other users. If you don't want to change the port don't change it, by default it will be 8266, but leave that option free for other users.

kohlerjl:
> by default it will be 8266

No, this is only true for the UDP port, which is used for the initial handshake in the ArduinoOTA protocol and is configured with the -p command line option to espota.py. During that handshake, the python script opens a random (by default) TCP port on the developer computer, and sends the port number to the device via UDP. The device then connects back to the developer computer via this port to download the firmware. This presents a challenge for a firewall, because the TCP port used is *not* fixed by default.

> it is clear for those who have a firewall that must open the ports they wish to use.

I don't think it is clear. I haven't seen any documentation about the TCP channel used in the transfer, and I had to read the source code and use a packet logger to figure out what was going wrong. The current defaults would require opening a large range of TCP ports for incoming traffic in the firewall.

I do think it would be better to allow configuration, as I originally mentioned, but as a simple first step, fixing the TCP port number (with -P) allows a fixed firewall configuration, and I don't see any issues for other users over a random port.

lrmoreno007:
Ok, now I understand you. This is the code in espota.py:

```
  group.add_option("-p", "--port",
    dest = "esp_port",
    type = "int",
    help = "ESP8266 ota Port. Default 8266",
    default = 8266
  )
  group.add_option("-P", "--host_port",
    dest = "host_port",
    type = "int",
    help = "Host server ota Port. Default random 10000-60000",
    default = random.randint(10000,60000)
  )
```

earlephilhower:
ArduinoOTA is not recommended for open internet usage, as I understand it, due to unencrypted transmission of username and password.  While an individual user can do as they wish, I'm against making it easier to do the wrong thing security-wise with this PR.  Closing.

bakadave:
I have the same issue: Windows Firewall blocks the reverse TCP connection from the device and I can't come up with a firewall rule that would work, other than opening every port which is a bad idea. I would like to point out that I'm on my own LAN and the firewall profile is set to "Private".

Firewall log:
```
#Version: 1.5
#Software: Microsoft Windows Firewall
#Time Format: Local
#Fields: date time action protocol src-ip dst-ip src-port dst-port size tcpflags tcpsyn tcpack tcpwin icmptype icmpcode info path

2021-07-31 12:02:16 DROP TCP 192.168.88.181 192.168.88.186 53146 40977 48 S 6541 0 2144 - - - RECEIVE
2021-07-31 12:02:19 DROP TCP 192.168.88.181 192.168.88.186 53146 40977 48 S 6541 0 2144 - - - RECEIVE
```

PR #4173, opened 2018-01-16
UPDATE_ERROR_ERASE wasn't used.

egemenertugrul:
Added two boolean results in _writeBuffer() function, thus utilized UPDATE_ERROR_ERASE. This helps to notice whether ESP.flashEraseSector or ESP.flashWrite was unsuccessful when an error occurs. Previously, it would only show UPDATE_ERROR_WRITE.

### DIFF
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -358,7 +358,7 @@ void UpdaterClass::_setError(int error){
 #endif
 }
 
-void UpdaterClass::printError(Stream &out){
+void UpdaterClass::printError(Print &out){
   out.printf_P(PSTR("ERROR[%u]: "), _error);
   if(_error == UPDATE_ERROR_OK){
     out.println(F("No Error"));
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -52,10 +52,7 @@ bool UpdaterClass::begin(size_t size, int command) {
   */
   int boot_mode = (GPI >> 16) & 0xf;
   if (boot_mode == 1) {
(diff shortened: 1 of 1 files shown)


cores/esp8266/Updater.cpp
```diff
@@ -223,18 +223,26 @@ bool UpdaterClass::end(bool evenIfRemaining){
...
+    eraseResult = ESP.flashEraseSector(_currentAddress/FLASH_SECTOR_SIZE);
   }
   
-  if (result) {
+  if (eraseResult) {
     if(!_async) yield();
-    result = ESP.flashWrite(_currentAddress, (uint32_t*) _buffer, _bufferLen);
+    writeResult = ESP.flashWrite(_currentAddress, (uint32_t*) _buffer, _bufferLen);
+  } else { // if erase was not successful
+    _error = UPDATE_ERROR_ERASE;
```
devyte: These three lines are repeated all over the place, and your change adds another repeat:
```
    _error = some_error;
#ifdef DEBUG_UPDATER
    printError(DEBUG_UPDATER);
#endif

```
Could you please create a new private method "setError(int err)" that wraps the three lines, and replace all cases in the file where the three lines appear by a call to the method? In this case, it would look like this:
```
_currentAddress = (_startAddress + _size);
setError(UPDATE_ERROR_ERASE);
return false;
```
There are a few cases where you'll need to swap lines (like in your change), and other cases where the #ifdef is even missing.

egemenertugrul: I believe I have fixed what you have mentioned, please check the latest commits. 

Thanks for the heads up.

PR #4196, opened 2018-01-19
Remove warnings when NDEBUG build option used

earlephilhower:
Sorry for the duplicate.  Something weird with git is making it impossible for me to clean up #4066, so I'm closing that one and opening this as a new PR.  It has *only* the WiFiClient changes to clear up the NDEBUG build warnings, and does not touch the core (that's moved into #4187  completely).

...snip...
When building using the new NDEBUG option recently added, the assert()
macro is defined to nothing. This leaves a few variables unused in the
WiFi stack causing compiler warnings. Add in empty casts to remove
these warnings. Does not affect actual assert use when NDEBUG is not
defined.


earlephilhower:
This minor tweak was approved in the initial #4066 PR, which was split into this and another one.  Going to commit this one as the warnings generated are a real PITA for cleaning up my own code with -NDEBUG.

PR #4200, opened 2018-01-20
Added network priority functionality.

voney:
Highest priority is 0, counts up from there.
APs with the same priority will function as per original.
Default priority is 255 (lowest).

Haven't yet made overload functions to allow priority with open networks.


libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
```diff
@@ -91,14 +91,16 @@ wl_status_t ESP8266WiFiMulti::run(void) {
                 for(auto entry : APlist) {
                     if(ssid_scan == entry.ssid) { // SSID match
                         known = true;
-                        if(rssi_scan > bestNetworkDb) { // best network
-                            if(sec_scan == ENC_TYPE_NONE || entry.passphrase) { // check for passphrase if not open wlan
-                                bestNetworkDb = rssi_scan;
-                                bestChannel = chan_scan;
-                                bestNetwork = entry;
-                                memcpy((void*) &bestBSSID, (void*) BSSID_scan, sizeof(bestBSSID));
+                        if (entry.priority <= bestNetwork.priority){
```
igrr: Shouldn't it be the other way around? It seems that lower values of 'priority' mean higher priority. If that's intended, maybe give this variable some other name (like 'order'?)

voney: Maybe this is just my background but i think of it like "number one priority" or "your first priority is x, your second is y" etc.

DNS MX records work in a similar way, lower number = higher priority

voney:
Hey @igrr I've changed it to order as per your suggestion. It's just worth noting that two networks can have the same order, hence the original name priority. Let me know if this works for you.

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
earlephilhower: UNIX nice's the same way (1=important, 20=only if nothing else to do).  How many Arduino users are aware of this, though, I can't say

voney: This was my first pull request to an open source project. Now that I've changed the variable to order instead of priority do I just wait? Or should I close it and re- request?

devyte:
@voney wait for now. The idea is a good one, so I'd like to review as well.

voney:
Any word on this?

voney:
Hi, I'm still unsure as to what I need to do to get this merged in... Can someone in the know please advise?

voney:
Still following this up. Is this request just not being seen due to it being flagged as changes requested? I've made the requested changes but we're still here months later.

devyte:
Serial/lwip stability fixes, heap/iram improvements, and other features, such as the monster bearssl implementation, timer use improvement, functional interrupt, mdns rewrite, take precedence. Please be patient, this is not forgotten, I'll get to it eventually.

earlephilhower:

voney:
> Thanks for your PR, but the core and libraries have changed enough that this PR now has a merge conflict.
> 
> Could you merge it manually with the latest core, so we can consider it for future releases?

No thanks.

Issue #3740, opened 2017-10-21
Improve available heap


devyte:
### Hardware
Hardware:		all
Core Version:      	git / 2.4-rc2

### Problem Description

Summary 

This issue is meant to track a RAM optimization effort. The goal is to increase available RAM as much as possible.

Details

The following are initial ideas to investigate.
- Move all const char strings to flash, and use the the access macros to access them. Implement alternate methods/functions with the correct argument signatures where applicable. This covers both the core as well as the Arduino lib code.
- Move const declarations to flash. Example: CHARS_PER_LINE in libb64/cencode.c.
- Revise all singleton object declarations. Make sure that all are wrapped by #defines that allow compile-time configuration to not use the global singletons. Example: Use of the Serial, but not Serial1, frees up 256+ bytes of heap.
- Revise all uses of constant size char [] arrays. Arrays like that should not be declared on the stack, but rather created dynamically, and only on demand. Example: 2K buffer in ESP8266Webserver.
-Revise all uses of String. When concatenating, the internal managed array will grow, sometimes far beyond what is needed. It is sometimes better to reserve, if there is an estimate of how much space will be needed.
- Revise object implementations. Make sure objects are passed by reference or pointer, and never by copy.

Pablo2048:
Actually I suggest to modify debug messages also to store text into PROGMEM. I've write macro for this reason:
`#ifdef DEBUG_ESP_PORT
#define DEBUG_MSG(_1, ...) DEBUG_ESP_PORT.printf_P(PSTR(_1), ##__VA_ARGS__)
#else
#define DEBUG_MSG(...)
#endif`

d-a-v:
There is this macro USE_OPTIMIZE_PRINTF  in osapi.h which, if defined, makes all os_printf's fmt stored in flash.

Pablo2048:
Yes, but IMHO this is not used in debug messages inside libraries - for example look into ESP8266WiFiGeneric.h macro DEBUG_WIFI_GENERIC. Every debug macro inside library uses DEBUG_ESP_PORT.printf() so it uses Serial, or Serial1 object, not os_printf...

earlephilhower:
You probably already know this, but the following command lines are very useful for finding stuff in RODATA that's eating up RAM:
For finding constant strings: objdump  -s -j .rodata /tmp/arduino_build_<code>/*.elf
For finding symbols(aka variables): readelf -a /tmp/arduino_build_<code>/*.elf| grep 3f

[...]

Issue #4134, opened 2018-01-11
Got Exception 28 when lost internet and reconnect internet using MQTT


tarzan115:
#### Hardware
Hardware:			`ESP-12E`
Core Version:      	`2.4.0`

### Description

I used library [async mqtt](https://github.com/marvinroger/async-mqtt-client) to connect mqtt, but when I already connected to the broker. I plug out the line of the internet, wifi still present. after 1 minute I  plug in to have internet. and my device auto reconnect to broker and crash with `Exception 28`

### Settings in IDE

Module:  			`Generic ESP8266 Module`
Flash Size: 		`4MB/3MB`
CPU Frequency:		`80Mhz`
Flash Mode:			`dio`
Flash Frequency:	`40Mhz`
Upload Using:		`SERIAL`
Reset Method:		`ck`
LwIP Variant:             both `v2 Prebuild (MSS=536)` and `v2 Prebuild (MSS=1460 , unstable)` got that error

### Sketch

```cpp

#include <Arduino.h>

void setup() {

}

void loop() {

}
```

### Debug Messages

```
Exception 28: LoadProhibited: A load referenced a page mapped with an attribute that does not permit loads
Decoding 12 results
0x40231e0a: pbuf_cat at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/lwip2-src/src/core/pbuf.c line 848 (discriminator 1)
0x4023189b: ip_chksum_pseudo at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/lwip2-src/src/core/inet_chksum.c line 542
0x4022d91c: udp_input at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/lwip2-src/src/core/udp.c line 393
0x40230f18: ip4_input at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/lwip2-src/src/core/ipv4/ip4.c line 679
0x4022a461: ethernet_input_LWIP2 at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/lwip2-src/src/netif/ethernet.c line 182
0x401045cc: esp2glue_ethernet_input at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/glue-lwip/lwip-git.c line 406
0x4025c492: ethernet_input at /home/david/dev/esp8266/origin/tools/sdk/lwip2/builder/glue-esp/lwip-esp.c line 349
[...]

PR #4204, opened 2018-01-20
Dynamically allocate the SD sector cache on use, save ~512 bytes when SD not used

earlephilhower:
Related to issue #3740 

The SD library includes a 512-byte cache array as a static class member
to hold a copy of a sector on the SD card. Because it is statically
allocated, the mere act of including <SD.h> in either your own application
or *any library* you use--even if you didn't ever use any of the SD
objects--would cause the linker to allocate a 512 byte array in BSS out
of heap.

Replace this static allocation with a static pointer, and only allocate
that pointer on the instantiation of a SdVolume object.  Free it when
the SdVolume is destroyed.

The simple test example below shows the issue:
cat > sdcardtest.ino <<EOF
#include <SD.h>

void setup() {
  Serial.begin(115200);
  Serial.println("Hello, world!");
}
void loop() {
}
EOF

Examining the memory map shows the large buffer that's allocated w/o anyone
ever actually allocating a SD object.
earle@/tmp$ readelf -a /tmp/arduino_build_705221/*.elf | grep cacheBuff
   614: 3ffee9b8   512 OBJECT  GLOBAL DEFAULT    3 _ZN8SdVolume12cacheBuffer


libraries/SD/src/utility/SdFat.h
```diff
@@ -431,14 +431,16 @@ union cache_t {
 class SdVolume {
  public:
   /** Create an instance of SdVolume */
-  SdVolume(void) :allocSearchStart_(2), fatType_(0) {}
+  SdVolume(void) :allocSearchStart_(2), fatType_(0) { if (!cacheBuffer_) cacheBuffer_ = (cache_t *)malloc(sizeof(cache_t)); }
```
d-a-v: null is not checked. I don't really know myself what to do in constructors when this happen.
Maybe you can also help with this in #4134 :)

platform.txt
```diff
@@ -86,7 +86,7 @@ recipe.S.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.fla
 recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{build.path}/arduino.ar" "{object_file}"
 
 ## Combine gc-sections, archives, and objects
-recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" -Wl,--start-group {object_files} "{build.path}/arduino.ar" {compiler.c.elf.libs} -Wl,--end-group  "-L{build.path}"
+recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" -Wl,-M -Wl,-Map "-Wl,{build.path}/{build.project_name}.map"  {compiler.c.elf.flags} {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" -Wl,--start-group {object_files} "{build.path}/arduino.ar" {compiler.c.elf.libs} -Wl,--end-group  "-L{build.path}"
```
d-a-v: #4186 :)

earlephilhower:
@d-a-v That's what I get for doing two things at once.  Just removed that change, sorry!

libraries/SD/src/utility/SdFat.h
earlephilhower: I believe errors in constructors can only throw an exception.  Unfortunately, that's not the Arduino way, so I'm as at a loss as yourself.

The alternative here, and elsewhere, is to use lazy allocation and/or defensive.  Before use of the cache it can be checked and the method can error out if it can't be allocated, assuming the API allows it.  Perusing the SD headers, it looks like SEGV'ing is really the only thing that's possible using the current API.

Something like link time optimization (-flto) might eliminate the need for this kind of change as it should be able to detect no call graph into this static class variable, but I've not been able to make it work (and I believe we're upgrading GCC soon, and at that point it should be "automatic" anyway).

libraries/SD/src/utility/SdFat.h
```diff
@@ -431,14 +431,16 @@ union cache_t {
 class SdVolume {
  public:
   /** Create an instance of SdVolume */
-  SdVolume(void) :allocSearchStart_(2), fatType_(0) {}
+  SdVolume(void) :allocSearchStart_(2), fatType_(0) { if (!cacheBuffer_) cacheBuffer_ = (cache_t *)malloc(sizeof(cache_t)); }
+  /** Delete an instance of SdVolume */
+  ~SdVolume() { free(cacheBuffer_); cacheBuffer_ = NULL; }
```
devyte: The original cache was a static array, accessible from any instance. This change makes it a static pointer, and the constructor checks if it has been allocated before, in which case nothing is done, or not, in which case it is allocated.
This destructor just blindly frees the mem. If there is more than one instance, the others will then try to access after free.
I think it's unlikely that there will be more than one instance of this class, but I think this is something that should be fixed to cover e.g.: temps (pass by value) and copy. 
I suggest use of a shared_ptr.

earlephilhower: I had only considered the single global SD object.  Since then, with the BSSL stuff I've been introduced to the wonders of the std::shared_ptr and this would be a good use of one like you suggest.  Will see if I can fix it up this weekend.

earlephilhower: Done, but using a separate refcnt instead of std::ptr as there needs to be a special-case for cnt==1 (like the BearSSL issue).

earlephilhower:
There's been some large changes to SD since this was submitted, so drop it for the near term.

PR #4205, opened 2018-01-21
Feature: I2S callback on DMA read

Crypter:
This will allow to fill the DMA buffer on demand using callback function (or at least set a flag in the callback and do stuff later outside of the interrupt routine). This will save a lot of overhead because the only way to do it now is to check in a loop if there's empty buffer waiting to be filled.


cores/esp8266/core_esp8266_i2s.c
```diff
@@ -92,10 +93,15 @@ void ICACHE_FLASH_ATTR i2s_slc_isr(void) {
       i2s_slc_queue_next_item(); //free space for finished_item
     }
     i2s_slc_queue[i2s_slc_queue_len++] = finished_item->buf_ptr;
+    if (i2s_callback) i2s_callback();
     ETS_SLC_INTR_ENABLE();
   }
 }
 
+void ICACHE_FLASH_ATTR i2s_set_callback(void (*callback) (void)){
```
earlephilhower: Now that 2.4.1 is out this can get into the merge queue.  Only minor change I would request is to remove the ICACHE_FLASH_ATTR from the set-callback function.  There's no need for this to be kept in precious IRAM.

cores/esp8266/i2s.h
```diff
@@ -51,6 +51,7 @@ bool i2s_write_lr(int16_t left, int16_t right);//combines both channels and call
 bool i2s_is_full();//returns true if DMA is full and can not take more bytes (overflow)
 bool i2s_is_empty();//returns true if DMA is empty (underflow)
 int16_t i2s_available();// returns the number of samples than can be written before blocking
+void i2s_set_callback(void (*callback) (void));
```
earlephilhower: (continuing prior thought)
The actual callback function passed in, though, needs to be in IRAM.  A comment in the code to that effect would be appreciated here, so it's obvious to people coming into it that it's an ISR and needs to be very special...

Crypter:
The requested changes are implemented, lets hope this gets merged soon :))

earlephilhower:
Thanks, squashed and committed.

PR #4221, opened 2018-01-23
fix the dead link docs

tarzan115:
fix the dead link for all docs


devyte:
I don't know of a way to really test this. I checked two of the addressed links: they are in fact broken, and the proposed fixes seem sane. The rest of the proposed fixes are in line with the two I checked.
I'm going to merge this as-is. If anything goes wrong, we can revert.

PR #4222, opened 2018-01-23
add missing constants to variants/oak

boneskull:
headers from the Oak didn't import `common.h`, among other problems.  for example, Firmata fails to build if `NUM_ANALOG_INPUTS` is not defined.

I also modified the SCL/SCA definitions to look like the other variants.

### DIFF
--- a/variants/oak/pins_arduino.h
+++ b/variants/oak/pins_arduino.h
@@ -26,6 +26,11 @@
 #ifndef Pins_Arduino_h
 #define Pins_Arduino_h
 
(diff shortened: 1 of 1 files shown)


boneskull:
For anyone using PlatformIO, you can workaround this until released by adding the following to your `platformio.ini` under the appropriate `[env:foo]`:

```ini
build_flags = -DNUM_ANALOG_INPUTS=1 -DNUM_DIGITAL_PINS=11
```

Issue #4089, opened 2018-01-04
Increase in binary size with 2.4.0


xoseperez:
### Description

Using a simple blink sketch shows an increment of ~25Kb in binary size. It is problematic when you are running out of space for OTA, for instance, or with 512Kb boards (I know there are not many of these now).

Sample sketch below is 226304 bytes with 2.3.0 and 251168 bytes with 2.4.0. That's 24864 (an 11%) bytes more to do the same.

Anyone else is having this problem? Is this a known "issue"? Is the code base that larger with 2.4.0? Is there a way to reduce binary size with 2.4.0?

#### Hardware
Hardware:		Any
Core Version:      	2.4.0 vs 2.3.0

### Settings in IDE
Module:  			Generic ESP8266 Module
Flash Size: 		1MB
CPU Frequency:	80Mhz/any
Flash Mode:		dout/any
Flash Frequency:	40Mhz/any
Upload Using:		N/A
Reset Method:		N/A

### Sketch
```
#define PIN 2

void setup() {
    pinMode(PIN, OUTPUT);
}

void loop() {
    digitalWrite(PIN, HIGH);
    delay(1000);
    digitalWrite(PIN, LOW);
    delay(1000);
}
```

xoseperez:
A more complex project shows the same problem. Memory usage also gets worst. This is the data for ESPurna framework:

```
                        |Cache IRAM |Init RAM   |R.O. RAM   |Uninit RAM |Flash ROM  | Binary size
--
2.3.0 - ESPurna Core   	|29789      |900        |2764       |30232      |238595     | 276192
2.4.0 - ESPurna Core 	|27519      |1292       |3536       |30712      |275891     | 312384
--
2.3.0 - ESPurna Default	|30289      |1032       |8856       |31648      |405443     | 449744
2.4.0 - ESPurna Default |27811      |1432       |9612       |32144      |443347     | 486352
```

arendst:
Look at the bright side: iram usage is a lot less allowing more use of interrupt driven code.

As for overall code size, mine grew by 22k too but I saw it coming using the rc versions and provide therefore a smaller binary to OTA load first before the full version can be OTA loaded. 

It always gets bigger....

devyte:
[...]

PR #4232, opened 2018-01-24
new board generator option: --nofloat

d-a-v:
~17K smaller sketches when floats are not needed
discussion in #4089


Raphyyy:
Hi,
Is there a way to implement this on PlatformIO ?

Issue #4235, opened 2018-01-24
Workaround for Alexa Bug


reloxx13:
Heya,

this library is used by the Tasmota Firmware to emulate HUE and WEMO Devices to control them with Amazon Alexa.

Currently their seems to be a bug in the Alexa Firmware where the header content is wrong resulting in a wrong parsing and missing args.

Another dev @OFreddy  made a workaround for it.

Please check this out and tell us, if their is any chance to include it in your library, since we cannot do anything in the Alexa Firmware.

https://github.com/arendst/Sonoff-Tasmota/issues/1639#issuecomment-360218536
https://github.com/arendst/Sonoff-Tasmota/issues/1434

devyte:
The referenced code rings a bell in my mind. There might already be a PR with the proposal, but I don't think Alexa was mentioned.
If there is no PR, please make one for review.

reloxx13:
seems u r right, but that pr is a lil bit diffrent.

https://github.com/esp8266/Arduino/pull/4151

i dont know which is better.

devyte:
@reloxx13 @Monarch73 please discuss the differences between this PR and #4151 . 
Are they both aimed at Alexa? 
Do they both work?
Are both code changes equivalent, or is one better than the other for some reason?

reloxx13:
Its independent from Alexa, but works for her, too.

As Monach said, it fixes the missing args if the server send the "application/x-www-form-urlencoded" header content-type instead of "application/json" and plain text (json) and not as suggestest for form-urlencoded a key/value pair.

i think monarches is better, cuz he made it in the _parseArguments method, while the one i posted is out of that method.

OFreddy:
I would prefer my solution as it saves code, because the code for adding the plain content is already existing.

reloxx13:
good point 👍

Monarch73:
I don't mind how it's done. Either way is fine with me.

OFreddy:
I am struggling with every byte to keep ota firmware update via web interface running ;-)
Not very satisfying

Monarch73:
I think if you do a direct size-comparison  of the binaries, you wouldn't get any differences. The optimizer in gcc picks up these kind of things pretty well...

how ever, like I said, I don't mind how it's done

OFreddy:
In that case I don't mind how it is done either.

themindfactory:
Can this work with Google home?

reloxx13:
any news about the PR ? :S

[...]

PR #4236, opened 2018-01-24
Workaround for Alexa Bug

reloxx13:
The PR for https://github.com/esp8266/Arduino/issues/4235

This https://github.com/esp8266/Arduino/pull/4151 is the same issue, but with a diffrent solution.

### DIFF
--- a/libraries/ESP8266WebServer/src/Parsing.cpp
+++ b/libraries/ESP8266WebServer/src/Parsing.cpp
@@ -190,7 +190,7 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
           searchStr += plainBuf;
         }
         _parseArguments(searchStr);
-        if(!isEncoded){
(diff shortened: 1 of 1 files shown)


ascillato:
Hi,

Any news on this PR ?

d-a-v:
Out of curiosity, is this an alexa bug or a core bug ?
if Alexa, isn't she supposed to receive firmware update ?

reloxx13:
hmm hard to say, i think its both :D

esp is throwing away data/arguments cuz of wrong or missing encoding in the request.

with an orginal hue device alexa does work.
the tasmota project emulates a hue device so i dont think amazon will fix it. its even hard to reach the devs to just report it. 
if it would get fixed, it will take a while till every echo owner will get the firmware. their is no manual method to look for an update and amazon is not rolling it out to everyone.
i got 3 echo dot 2g and 1 has a newer firmware since 3 month now. the other two are still on an old one.

PR #4241, opened 2018-01-25
Add a new resetmethod_menu_all macro

mribble:
Add a new resetmethod_menu_all macro to give the choice of all the reset options.  Then only offer all these options for the generic modules.

This is a continuation of pull request #4228


tools/boards.txt.py
```diff
@@ -761,6 +762,13 @@
     ####################### menu.resetmethod
 
     'resetmethod_menu': collections.OrderedDict([
+        ( '.menu.ResetMethod.ck', 'ck' ),
+        ( '.menu.ResetMethod.ck.upload.resetmethod', 'ck' ),
+        ( '.menu.ResetMethod.nodemcu', 'nodemcu' ),
+        ( '.menu.ResetMethod.nodemcu.upload.resetmethod', 'nodemcu' ),
+        ]),
+
+    'resetmethod_menu_all': collections.OrderedDict([
```
d-a-v: What do you think of calling this one ```_extra``` and leaving in it only ```none``` and ```dtrset```, then use it only in generic boards?

PR #4248, opened 2018-01-28
SPISlave.end() added

jandrassy:
resetting the registers is needed for the switch to master mode for example for AVRISP


libraries/SPISlave/src/SPISlave.cpp
```diff
@@ -72,6 +72,14 @@ void SPISlaveClass::begin()
     hspi_slave_onStatusSent(&_s_status_tx);
     hspi_slave_begin(4, this);
 }
+void SPISlaveClass::end()
+{
+    hspi_slave_onData(NULL);
```
devyte: Please use nullptr instead of NULL

libraries/SPISlave/src/hspi_slave.c
```diff
@@ -98,6 +98,23 @@ void hspi_slave_begin(uint8_t status_len, void * arg)
     ETS_SPI_INTR_ENABLE();
 }
 
+void hspi_slave_end()
+{
+  ETS_SPI_INTR_DISABLE();
+  ETS_SPI_INTR_ATTACH(_hspi_slave_isr_handler, NULL);
```
devyte: Shouldn't this be something like:
`ETS_SPI_INTR_ATTACH(nullptr, nullptr);`
?

jandrassy: sorry, the first argument should be NULL. I change it.

it is a C source code and nullptr is not declared

libraries/SPISlave/src/SPISlave.cpp
jandrassy: the hspi_slave_* functions are C functions called from C++. should nullptr  be used?

libraries/SPISlave/src/hspi_slave.c
devyte: Ah yes, C file, so NULL it is.

libraries/SPISlave/src/SPISlave.cpp
devyte: Yes. NULL is a #define 0 in C, so an int that gets cast into a 32bit ptr, and in C++ it is a #define 0x00 that is cast into a void*, while nullptr is a C++ builtin that is really a null ptr. It's a type and/or style thing. As a rule of thumb, use of nullptr is considered better in C++ code. You can google nullptr vs. NULL to read about the C++ decision.

jandrassy:
I made the requested changes as a new commit. How can we do it to look like one commit?

jandrassy:
I made the changes and changed the commit.

svofski:
I very much would love having this patch merged. Thanks!

jandrassy:
@svofski try to add your request as issue, with a reason why you need this

jandrassy:
@devyte, can you merge this?

svofski:
@jandrassy sure! https://github.com/esp8266/Arduino/issues/4511

PR #4265, opened 2018-01-31
ClientContext:_write_some: release buffer once data really accepted for sending

### LINKED ISSUE #1872: http.POST() limit of payload size?
Hi! 

I'm trying to send data as an HTTPClient to a server with http.POST() and it works great but the payload gets cut off around the 3000 character mark. Does this have something to do with "application/x-www-form-urlencoded"? It shouldn't, iirc, cause POST shouldn't have a size limit?

> HTTPClient http;
> http.begin("http://example.com/weather/espdata.php");
> http.addHeader("Content-Type", "application/x-www-form-urlencoded");
> http.POST(textToSend);
> http.end();

I just wanted to know if there is a limit on payload size so I can rule out some other issue with my server. The String I'm sending gets put out normally in the serial monitor, so that shouldn't be the problem.

Thanks!
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32680468-http-post-limit-of-payload-size?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

d-a-v:
may fix #1872

### DIFF
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -447,11 +447,13 @@ protected:
             }
             err_t err = tcp_write(_pcb, buf, next_chunk, TCP_WRITE_FLAG_COPY);
             DEBUGV(":wrc %d %d %d\r\n", next_chunk, will_send, (int) err);
-            _datasource->release_buffer(buf, next_chunk);
             if (err == ERR_OK) {
+                _datasource->release_buffer(buf, next_chunk);
                 _written += next_chunk;
                 need_output = true;
             } else {
+		// ERR_MEM(-1) is a valid error meaning
+		// "come back later". It leaves state() opened
                 break;
             }
             will_send -= next_chunk;


d-a-v:
This happens only with MSS!=1460 (with either lwip v1 or v2 - may be related with the 1460-sized buffers used in core, a PR for MSS instead will come later)

The ```tcp_write``` error is -1 (ERR_MEM) and is documented as follow:
> The tcp_write() function will fail and return ERR_MEM if the length of the data exceeds the current send buffer size or if the length of the queue of outgoing segment is larger than the upper limit defined in lwipopts.h (TCP_SND_QUEUELEN). If the function returns ERR_MEM, the application should wait until some of the currently enqueued data has been successfully received by the other host and try again.

PR #4275, opened 2018-02-02
Added possibility of sending POST with empty payload

aesapronov:
Specification does not prohibit the sending of an empty POST requests. In this case "Content-Length: 0" should be set. See thread here: http://lists.w3.org/Archives/Public/ietf-http-wg/2010JulSep/0276.html


aesapronov:
Yes, I'm also not an expert. But it not need to be an expert to agree with you. You suggestion is right. Do you make changes to the code or should I need to do this?

d-a-v:
If you commit the changes on your own repo, the PR will be updated accordingly.

PR #4276, opened 2018-02-02
LED_BUILTIN should be a #define

d-a-v:
Arduino only defines LED_BUILTIN, should BUILTIN_LED be removed ?


variants/ESPDuino/pins_arduino.h
```diff
@@ -33,7 +33,7 @@
 static const uint8_t SDA = PIN_WIRE_SDA;
 static const uint8_t SCL = PIN_WIRE_SCL;
 
-static const uint8_t LED_BUILTIN = 16;
-static const uint8_t BUILTIN_LED = 16;
+#define LED_BUILTIN 16
```
earlephilhower: Could a compiler #define such as "-DBUILTIN_LED=LED_BUILTIN" and remove the need for all these repeated 2nd #define BUILTIN_LED X?

d-a-v: BUILTIN_LED is not used in core, but seems to be historically used in esp8266 code outside.
What about moving this define into ```variants/generic/common.h``` (hidden, wrong name less exposed, less loaded gcc cmdline) ?

devyte: If it's not standard, how about putting it somewhere safe and deprecating its use?

d-a-v: Good idea

variants/espino/pins_arduino.h
```diff
@@ -34,15 +34,16 @@
...
 static const uint8_t SCL = PIN_WIRE_SCL;
 
-static const uint8_t LED_BUILTIN = 2;
+#define LED_BUILTIN 2
 static const uint8_t LED_BUILTIN_R = 2;
 static const uint8_t LED_BUILTIN_G = 4;
 static const uint8_t LED_BUILTIN_B = 5;
 
-static const uint8_t BUILTIN_LED = 2;
+/*
```
devyte: remove commented code

variants/espino/pins_arduino.h
```diff
@@ -34,15 +34,16 @@
...
 static const uint8_t BUILTIN_LEDR = 2;
 static const uint8_t BUILTIN_LEDG = 4;
 static const uint8_t BUILTIN_LEDB = 5;
 static const uint8_t BUILTIN_BUTTON = 0;
```
devyte: maybe there should be a BUTTON_BUILTIN?

d-a-v: Right. BUTTON_BUILTIN added, commented BUILTIN_* removed.

PR #4280, opened 2018-02-02
Minor cleanup of SSL server methods, missing macros

earlephilhower:
The first of these commits removes duplicated code for loading certs.  There's already a method to do this, so the add'l ones added for support of ESP8266WiFiServerSecure were not needed.

The second one fixes a compile error found in the WebServerSecure class when HTTP_SERVER debugging is enabled.  The output macro wasn't brought in from the original, non-secure file.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
@@ -329,14 +329,6 @@ public:
         return reinterpret_cast<SSLContext*>(fd)->io_ctx;
     }
 
-    int loadServerX509Cert(const uint8_t *cert, int len) {
-        return ssl_obj_memory_load(SSLContext::_ssl_ctx, SSL_OBJ_X509_CERT, cert, len, NULL);
-    }
-
-    int loadServerRSAKey(const uint8_t *rsakey, int len) {
-        return ssl_obj_memory_load(SSLContext::_ssl_ctx, SSL_OBJ_RSA_KEY, rsakey, len, NULL);
-    }
-
 protected:
     int _readAll()
     {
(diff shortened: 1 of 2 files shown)


d-a-v:
#4274 collision :) I'll update mine

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
```diff
@@ -471,24 +463,11 @@ WiFiClientSecure::WiFiClientSecure(ClientContext* client, bool usePMEM, const ui
...
-        uint8_t *stackData = (uint8_t*)alloca(max(certLen, rsakeyLen));
-        if (rsakey && rsakeyLen) {
-              memcpy_P(stackData, rsakey, rsakeyLen);
-              _ssl->loadServerRSAKey(stackData, rsakeyLen);
-        }
-        if (cert && certLen) {
-            memcpy_P(stackData, cert, certLen);
-            _ssl->loadServerX509Cert(stackData, certLen);
-        }
+        if (rsakey && rsakeyLen) _ssl->loadObject_P(SSL_OBJ_RSA_KEY, rsakey, rsakeyLen);
```
devyte: Please no 1-line if statements. Just put the _ssl line indented under the if.

devyte:
My one comment is just about readability

libraries/ESP8266WiFi/src/WiFiClientSecure.cpp
earlephilhower: My bad!  I try and cuddle all if{}s but sometimes one just slips by on me.  I've seen too many spots where folks have a 1-line, uncuddle'd if and decide "I'll put in a debug statement here" and end up changing the logic.

PR #4299, opened 2018-02-05
lwip2: use only wifi_station_get_hostname() to set netif hostname

### LINKED ISSUE #3970: dhcp hostname issue in 2.4.0
#### Hardware
Hardware:		ESP-12/NodeMCU v1.0
Core Version:      	2.4.0-github

### Description

When using github version of Arduino ESP8266, the dhcp client is not sending hostname in request, resulting in a blank name in the DHCP server, and no DNS entry created. If I recompile with 2.3.0, then the hostname is sent, and assigned as expected.

I noticed blank hostnames previously, but wasn't sure where they came from. It turns out that jumping between 2.3.0 and 2.4.0-github was masking the issue, as once the server has the name set, it maintains it thru renewals, even when the name is not specified subsequently. And something is causing my modules running 2.4.0-github to occasionally lose their lease (localIP() shows 0.0.0.0).

ALSO, the default generated hostname (ESP_xxxx) is invalid, as the underscore character is not allowed in a valid hostname. A valid hostname may only contain letters, numbers and the dash "-", though a valid hostname cannot begin or end with a dash. [https://en.wikipedia.org/wiki/Hostname](url)

### Settings in IDE

Arduino IDE:		1.8.5
Module:  			NodeMCU 1.0(ESP-12E)
Flash Size: 		4M(1M)
CPU Frequency:	80 Mhz
lwip:				v2
Upload Using:		Serial

### Sketch

```
#include <ESP8266WiFi.h>

#define HOST "test-hostname"
#define WIFI_SSID "SSID"
#define WIFI_PASS "PASS"

void setup() {
  Serial.begin(115200);
  while (!Serial);
  delay(100);
  Serial.println();
  Serial.print(F("setup: WiFi connecting to "));
  Serial.print(WIFI_SSID);
  Serial.print(F("..."));
  WiFi.hostname(HOST);
  WiFi.mode(WIFI_STA);
  WiFi.begin(WIFI_SSID, WIFI_PASS);
    Serial.print(F("."));
    delay(500);
  }
  Serial.println(F(" OK"));
  Serial.print(F("Last reset reason: "));
  Serial.println(ESP.getResetReason());
}

void loop() {
  String ip = WiFi.localIP().toString();
  int heap = ESP.getFreeHeap();
  Serial.printf("%s: %6d %6d \r\n", ip.c_str(), millis(), heap);
  delay(2000);
}
```

### Debug Messages

No related errors or other messages observed, even with full debug enabled.

d-a-v:
fix #3970


devyte:
Add a comment above esp2glue_netif_update(), above esp2glue_dhcp_start(), and above other functions calling those, that the netif hostname will be set from wifi_station_get_hostname(), with the implication that the hostname must be set externally via wifi_station_set_hostname() _before_ calling any of those functions.
Maybe put the same comment in the README, and other places it makes sense, i.e.: make it VERY visible, because I can feel issues in the future about the call order.

d-a-v:
Comments added inside lwip2, stating that there is no other way than setting hostname before dhcp client is started by WiFi.begin() (as it turns out, independantly from lwip2).

PR #4306, opened 2018-02-06
Refresh api.github.com x509 certificate

earlephilhower:
The certificate fingerprint included with the HTTPSRequest example seems
to be for an expired api.github.com certificate.  Replace with the current
one to avoid reporting "certificate mismatch" errors when running.

image


earlephilhower:
Platform.io builds are all now failing.  Will resubmit once the platform infrastructure is working again.

PR #4336, opened 2018-02-09
source level debugging via serial  port

freeck:
From Arduino-esp8266 version 2.0 on, gdbstub has been modified. It now supports catching exceptions from where you may use xtensa-x106-gdb.exe to inspect the state of the processor. Regular source level debugging(SL) is not possible anymore(not?). The Arduino-IDE does not support SL-debugging anyway, so perhaps a logical decision. 
SL-debugging should be a part of the tools for every embedded systems developer. Using this library enables command-line SL-debugging using the serial port. This library originated from VisualGDB, a Visual Studio-plug in. This library can also be used with Eclipse-Sloeber. A few modifications have been applied in order to minimize the IRAM-usage.
A tutorial has been added to explain the installation of gdbstub, and some aspects of gdb-command line-debugging. Be aware that SL-debugging with esp8266 is not always stable, some tips and tricks have been added to handle with that.


freeck:
Just read this: https://github.com/esp8266/Arduino/pull/4386 
I assume this pull request wont be relevant anymore? Is it?

devyte:
@freeck please discuss with @kylefleming . At a glance, I don't want to pick one over the other, and I think we should try to unify both approaches. Without looking too deep into the details, my instinct points me to adding uart hardware support to this PR, rather than go the other direction, but I rather you two hammer it out as the implementers.

earlephilhower:
I can see a ton of work was done in getting this to work.  But with GDB seeming to need both some toolchain fixes coupled with some libraries/GDBstub changes as well I think this needs to be looked at with the other PRs and potential new toolchain to be relevant and usable by most folks.  Punting to the next release, where we can hopefully square the circle and get it running.

earlephilhower:
This will block on PR #5559 with its new GDB support and documentation update for the plain CLI setup.  I know you're already looking at updating, @freeck, do you think it makes sense to start a new PR to track that work, or do you want to leave this older one?

freeck:
Leave the older one for time being....Most of that doc will be obsolete, I think one page of text and one or two screenshots can now cover the debug setup process :).

freeck:
Yep

earlephilhower:
@freeck, after going over your instructions.rst, I think it pretty much matches what's in the repo instructions already.  I'm closing this for now, but if you really want to update the existing gdb doc, then I think it would be easier for you to make a new PR (since all the code changes here are no longer needed).  Thanks!

PR #4363, opened 2018-02-13
Remove @PlatformIO library.json from GDBStub; fix pattern in LD

ivankravets:
Resolve #4355


kylefleming:
@ivankravets It seems like this is actually causing GDBStub to not link weak symbols properly anymore. I'm sure the exact cases where it does and doesn't link though.

ivankravets:
Do you use the latest PIO Core 3.5.2b4?

kylefleming:
Yep: `PlatformIO, version 3.5.2b4`

ivankravets:
Thanks! I reverted back `library.json` config in https://github.com/esp8266/Arduino/pull/4393

PR #4369, opened 2018-02-15
Add -Werror to acceptance builds for C and CPP

earlephilhower:
-edit- This is now completed and there are no errors or warnings reported when run with -Wall -Werror as part of the build process.  Ready for review 

Use platform.local.txt to add -Werror to GCC for the build of all
code.  Any warnings on a submitted patch will cause an error.

### DIFF
--- a/libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
+++ b/libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
@@ -87,10 +87,12 @@ bool LLMNRResponder::begin(const char* hostname) {
     _hostname.toLowerCase();
 
     _sta_got_ip_handler = WiFi.onStationModeGotIP([this](const WiFiEventStationModeGotIP& event){
+        (void) event;
         _restart();
     });
 
     _sta_disconnected_handler = WiFi.onStationModeDisconnected([this](const WiFiEventStationModeDisconnected& event) {
+        (void) event;
         _restart();
     });
 
@@ -122,6 +124,7 @@ bool LLMNRResponder::_restart() {
     _conn->setMulticastTTL(LLMNR_MULTICAST_TTL);
     _conn->onRx(std::bind(&LLMNRResponder::_process_packet, this));
     _conn->connect(multicast_addr, LLMNR_PORT);
+    return true;
 }
 
 void LLMNRResponder::_process_packet() {
@@ -242,8 +245,8 @@ void LLMNRResponder::_process_packet() {
 
     // Header
     uint8_t header[] = {
-        id >> 8, id & 0xff, // ID
-        FLAGS_QR >> 8, 0, // FLAGS
+        (uint8_t)(id >> 8), (uint8_t)(id & 0xff), // ID
+        (uint8_t)(FLAGS_QR >> 8), 0, // FLAGS
         0, 1, // QDCOUNT
         0, !!have_rr, // ANCOUNT
         0, 0, // NSCOUNT
@@ -269,7 +272,7 @@ void LLMNRResponder::_process_packet() {
             0, 1, // CLASS (IN)
             0, 0, 0, 30, // TTL (30 seconds)
             0, 4, // RDLENGTH
-            ip & 0xff, (ip >> 8) & 0xff, (ip >> 16) & 0xff, (ip >> 24) & 0xff, // RDATA
+            (uint8_t)(ip & 0xff), (uint8_t)((ip >> 8) & 0xff), (uint8_t)((ip >> 16) & 0xff), (uint8_t)((ip >> 24) & 0xff) // RDATA
         };
         _conn->append(reinterpret_cast<const char*>(rr), sizeof(rr));
     }
--- a/libraries/SD/src/utility/SdFile.cpp
+++ b/libraries/SD/src/utility/SdFile.cpp
@@ -259,11 +259,12 @@ uint8_t SdFile::make83Name(const char* str, uint8_t* name) {
       i = 8;   // place for extension
     } else {
       // illegal FAT characters
-      uint8_t b;
 #if defined(__AVR__)
+      uint8_t b;
       PGM_P p = PSTR("|<>^+=?/[];,*\"\\");
       while ((b = pgm_read_byte(p++))) if (b == c) return false;
 #elif defined(__arm__)
+      uint8_t b;
       const uint8_t valid[] = "|<>^+=?/[];,*\"\\";
       const uint8_t *p = valid;
       while ((b = *p++)) if (b == c) return false;
@@ -905,7 +906,7 @@ uint8_t SdFile::rmRfStar(void) {
       if (!f.remove()) return false;
     }
     // position to next entry if required
-    if (curPosition_ != (32*(index + 1))) {
+    if (curPosition_ != (uint32_t)(32*(index + 1))) {
       if (!seekSet(32*(index + 1))) return false;
     }
   }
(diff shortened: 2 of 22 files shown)


earlephilhower:
It caught the warning as an error.  Will remove the OTA test change and put in a final git commit:
````
...
/sketch/BasicOTA.ino.cpp.o"
/home/travis/build/esp8266/Arduino/libraries/ArduinoOTA/examples/BasicOTA/BasicOTA.ino: In function 'void loop()':
/home/travis/build/esp8266/Arduino/libraries/ArduinoOTA/examples/BasicOTA/BasicOTA.ino:64:5: error: unused variable 'x' [-Werror=unused-variable]
 int x = 0;
     ^
cc1plus: all warnings being treated as errors
Using library ESP8266WiFi at version 1.0 in folder: /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WiFi 
Using library ESP8266mDNS in folder: /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266mDNS (legacy)
Using library ArduinoOTA at version 1.0 in folder: /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/ArduinoOTA 
exit status 1
Sketch:  /home/travis/build/esp8266/Arduino/libraries/ArduinoOTA/examples/BasicOTA/BasicOTA.ino
Build dir:  /home/travis/build/esp8266/Arduino/build.tmp
Output:  /home/travis/build/esp8266/Arduino/build.tmp/BasicOTA.ino.bin
Building: /home/travis/arduino_ide/arduino-builder -compile -logger=human -build-path "/home/travis/build/esp8266/Arduino/build.tmp" -tools "/home/travis/arduino_ide/tools-builder" -libraries "/home/travis/Arduino/libraries" -hardware "/home/travis/arduino_ide/hardware" -hardware "/home/travis/build/esp8266/Arduino/tools/../cores" -fqbn=esp8266com:esp8266:generic:CpuFrequency=80,FlashFreq=40,FlashMode=qio,UploadSpeed=921600,FlashSize=512K64,ResetMethod=nodemcu -ide-version=10607 -warnings=all -verbose /home/travis/build/esp8266/Arduino/libraries/ArduinoOTA/examples/BasicOTA/BasicOTA.ino
The command "$TRAVIS_BUILD_DIR/tests/common.sh" exited with 1.
````

**Comment by earlephilhower** [comment]
Ugh, looks like many examples and even several libraries have problems when -Wall is enabled.  This is not as straightforward as I'd hoped.

**Comment by d-a-v** [comment]
I very much like the idea of -Wall for the tests. Should we be using -Wextra too ?

**Comment by earlephilhower** [comment]
At some point, maybe.  Right now, though, it's looking like 10s to 100s of changes will be required to get all the examples to build w/o warning-errors.  Some changes will be in the libraries, not just the sample code, too, so it's going to be somewhat intrusive.

I need to find a way to get the Travis tests to dump raw output from arduino-builder to get a good list, because individually building the dozens of examples is painful and slow...

File: libraries/TFT_Touch_Shield_V2/TFTv2.cpp [hunk 1]
```diff
@@ -204,10 +204,10 @@ void TFT::fillScreen(INT16U XL, INT16U XR, INT16U YU, INT16U YD, INT16U color)
         YD = YU^YD;
         YU = YU^YD;
     }
-    XL = constrain(XL, MIN_X,MAX_X);
-    XR = constrain(XR, MIN_X,MAX_X);
-    YU = constrain(YU, MIN_Y,MAX_Y);
-    YD = constrain(YD, MIN_Y,MAX_Y);
+    XL = (XL > MAX_X) ? MAX_X : XL;
```
devyte: how about using std::min here?

libraries/SD/src/utility/SdFile.cpp
```diff
@@ -905,7 +906,7 @@ uint8_t SdFile::rmRfStar(void) {
       if (!f.remove()) return false;
     }
     // position to next entry if required
-    if (curPosition_ != (32*(index + 1))) {
+    if (curPosition_ != (uint32_t)(32*(index + 1))) {
```
devyte: index is uint16_t, and the value is being expanded.
I _think_ this is ok, after reading up on usual arithmetic conversions, but I'm not positive. Please double check that the cast is ok on the final arithmetic result, and that it shouldn't be done directly on index _before_ the arithmetic operations.

tests/common.sh
```diff
@@ -40,7 +40,7 @@ function build_sketches()
     local build_arg=$3
     local build_dir=build.tmp
     mkdir -p $build_dir
-    local build_cmd="python tools/build.py -b generic -v -k -p $PWD/$build_dir $build_arg "
+    local build_cmd="python tools/build.py -b generic -v -w all -k -p $PWD/$build_dir $build_arg "
```
devyte: If all warnings are now eliminated, isn't it ok to build with -werror? I'd like to enforce no warnings, sooner rather than later.

devyte:
I'm approving despite my comments, which are minor. My one pet peeve is that I'd like to enforce no warning builds asap (i.e.: -werror), so as to not accumulate again in the future.
I think this usually falls under code quality metrics, so I suggest merging asap.

libraries/ESP8266LLMNR/ESP8266LLMNR.cpp
```diff
@@ -242,8 +245,8 @@ void LLMNRResponder::_process_packet() {
 
     // Header
     uint8_t header[] = {
-        id >> 8, id & 0xff, // ID
-        FLAGS_QR >> 8, 0, // FLAGS
+        (uint8_t)(id >> 8), (uint8_t)(id & 0xff), // ID
```
igrr: since there's a cast to uint8_t now, masking the low byte with `& 0xff` is not needed.

libraries/TFT_Touch_Shield_V2/TFTv2.cpp
igrr: By the way, why is constraint on MIN_Y being dropped?

tests/common.sh
igrr: It's already done below where platform.local.txt is modified.

tests/common.sh
```diff
@@ -97,6 +103,9 @@ function install_ide()
     mkdir esp8266com
     cd esp8266com
     ln -s $core_path esp8266
+    # Set custom warnings for all builds (i.e. could add -Wextra at some point)
```
igrr: The last part of the comment seems to be outdated as `-Werror` is already present.

libraries/TFT_Touch_Shield_V2/TFTv2.cpp
earlephilhower: GCC is pedantic here.  MIN_X/Y =0, and constrain is a macro that compares a to b  So you end up with "(uint16_t)XR < 0" as one of the conditions it checks, and that's by definition always false.  I will put std::min in on next push.

libraries/SD/src/utility/SdFile.cpp
earlephilhower: Didn't mean to expand it, thought curPosition_ was 32-bits.  Will correct to uint16_t.

tests/common.sh
earlephilhower: Also, that parameter is not passed to GCC but to build.yp,which passes it to arduino-builder, which then uses it to lookup into a GCC setting array.  Can't send in -werror (it's included now in platform.local.txt that's generated before running the builder.

tests/common.sh
earlephilhower: Actually, "-Werror" is included, but the comment says "-Wextra" :)  So this is really OK.

igrr: I'll stop reviewing PRs until I wake up. Sorry.

libraries/SD/src/utility/SdFile.cpp
earlephilhower: Misread your comment, will move the *cast* inside.  No 16 bits.

PR #4370, opened 2018-02-15
sdk: update to 2.2.0

igrr:
- update libraries and header files
- remove libmesh
- update PHY init data


igrr:
Looks like platformio build is failing, still trying to link libmesh.a. I suppose we need to convince it to pull linker flags from platform.txt somehow.

igrr:
@ivankravets could you point me to any way to override LD flags? Seems that platformio assumes certain list of flags, which include the list of libraries to link... We need to modify that.

d-a-v:
I needed to erase my flash (could not catch the AP), after that, first sketch with ssl is running well.

ivankravets:
@igrr  see https://github.com/esp8266/Arduino/blob/master/tools/platformio-build.py#L62

If you need to override `LINKFLAGS`, the better is to use `env.Replace`

```python
env.Append(
    LINKFLAGS=[
        ...
    ]
)
```

Currently, `LINKFLAGS` consist of 

* https://github.com/platformio/platform-espressif8266/blob/develop/builder/main.py#L97
* https://github.com/esp8266/Arduino/blob/master/tools/platformio-build.py#L66

This build script extends flags from `main.py` by default.

cores/esp8266/core_esp8266_phy.c
```diff
@@ -68,11 +68,11 @@ static const uint8_t ICACHE_FLASH_ATTR phy_init_data[128] =
     [32] = 0xf8, // Reserved, do not change
     [33] = 0xf8, // Reserved, do not change
 
-    [34] = 82, // target_power_qdb_0, 82 means target power is 82/4=20.5dbm
-    [35] = 78, // target_power_qdb_1, 78 means target power is 78/4=19.5dbm
-    [36] = 74, // target_power_qdb_2, 74 means target power is 74/4=18.5dbm
-    [37] = 68, // target_power_qdb_3, 68 means target power is 68/4=17dbm
-    [38] = 64, // target_power_qdb_4, 64 means target power is 64/4=16dbm
+    [34] = 78, // target_power_qdb_0, 82 means target power is 78/4=19.5dbm
```
devyte: Comments need updating (82 => 78)

cores/esp8266/core_esp8266_phy.c
```diff
@@ -31,7 +31,7 @@
 static const uint8_t ICACHE_FLASH_ATTR phy_init_data[128] =
 {
     [0] = 5,  // Reserved, do not change
-    [1] = 0,  // Reserved, do not change
+    [1] = 8,  // Reserved, do not change
```
devyte: do we know what this byte does?

igrr: According to https://github.com/espressif/ESP8266_NONOS_SDK/commit/d62e5a70d894028066bfe47dfd6e2a8a39dd5483, second byte is some version number.

PR #4371, opened 2018-02-16
additional mimetable fixes, additional string moves to progmem

devyte:
Some missing fixes

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
@@ -36,11 +36,9 @@
 #define DEBUG_OUTPUT Serial
 #endif
 
-//const char * AUTHORIZATION_HEADER = "Authorization";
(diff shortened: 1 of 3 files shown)


libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -36,11 +36,9 @@
 #define DEBUG_OUTPUT Serial
 #endif
 
-//const char * AUTHORIZATION_HEADER = "Authorization";
 static const char AUTHORIZATION_HEADER[] PROGMEM = "Authorization";
 static const char qop_auth[] PROGMEM = "qop=auth";
 static const char WWW_Authenticate[] PROGMEM = "WWW-Authenticate";
-static const char colon[] PROGMEM = ":";
```
earlephilhower: Thanks for containing your OCD here.

d-a-v: :) there's another one on L94 that could spare 2 more flash bytes

PR #4379, opened 2018-02-17
CI: Check if files generated by boards.txt.py are up to date

igrr:
Currently boards.txt, doc/boards.rst, pacakge JSON template, and LD scripts can be updated by boards.txt.py. This PR adds a mechanism to check that these files are committed when changes to boards.txt.py are made. If any difference is found between generated files and files in the repository, CI build will fail like this: https://travis-ci.org/esp8266/Arduino/jobs/342729114.

Additionally, the following tweaks  were made:

- Replace remaining tabs in boards.txt.py with spaces
- Make LD script output of the script look more like the existing LD scripts
- Some small formatting/comment tweaks to LD scripts to match generated output
- Fix issues with the generated JSON output
- Regenerate package JSON file


igrr:
> I can see a slight difference: './' is (re)added on 1st line comment in boards.txt when all files are regenerated.

This depends on how you run the tool: if you type `./tools/boards.txt.py <args>`, it will add the dot. If you run it as `tools/boards.txt.py <args>`, it will generate the output as in this PR.

I've pushed one more commit which strips that path from script name, so it should always print `...is likely to be overwritten by boards.txt.py` regardless of the way the script is invoked.

PR #4380, opened 2018-02-17
Add menu to choose which part of flash gets erased

igrr:
While troubleshooting some WiFi connection issues it is often required to erase flash contents, specifically SDK configuration (last 4 sectors of flash).

This PR adds a menu which allows selecting between:

- normal upload, when only the part of the flash required to upload the sketch is erased
- above plus SDK settings (last 4 sectors) are also erased
- the whole flash chip is erased

To support these changes, esptool-ck is updated to 0.4.13.


igrr:
Currently the new menu item is called "Erase Settings" and the options are "Only Sketch", "SDK Settings", and "All Flash Contents". Suggestions about better naming are very welcome.

Another option to give users a way to erase flash would be to create a plugin, similar to the SPIFFS upload plugin, but that needs some (even if small) effort to install.

One more option is to make the IDE perform flash erase when "Burn Bootloader" menu item is chosen. I thought this way was guaranteed to cause confusion, and went for the option implemented in this PR.

tools/boards.txt.py
```diff
@@ -1109,6 +1123,7 @@ def all_boards ():
     print 'menu.DebugLevel=Debug Level'
     print 'menu.LwIPVariant=lwIP Variant'
     print 'menu.led=Builtin Led'
+    print 'menu.FlashErase=Erase Settings'
```
devyte: I suggest "Erase Flash" here. "Erase Settings" is a bit confusing.

tools/boards.txt.py
```diff
@@ -889,6 +890,17 @@
         ( '.menu.UploadSpeed.921600.upload.speed', '921600' ),
         ]),
 
+    ####################### flash erase
+
+    'flash_erase_menu': collections.OrderedDict([
+        ( '.menu.FlashErase.none', 'Only Sketch' ),
+        ( '.menu.FlashErase.none.upload.erase_cmd', '' ),
+        ( '.menu.FlashErase.sdk', 'SDK Settings' ),
```
devyte: I suggest "Sketch + SDK Settings" here, to reflect reality.

d-a-v: +1 (or 'Sketch + WiFi settings' ?)

igrr:
CI check for boards.txt has successfully triggered here after I forgot to commit my boards.txt changes:
https://travis-ci.org/esp8266/Arduino/jobs/343289885

Also noticed that I was formatting `flash_size_bytes` as a hex number, but forgot to add leading `0x`. Fixed that as well.

d-a-v:
Here is a sketch to try with OTA.
The ```check_sdk22()``` needs to be in the first sketch flashed with sdk-2.2.
It uses EEPROM to check whether the erase operation is needed.

```
#include <EEPROM.h>
#include <ESP8266WiFi.h>

#define SSID "open"
#define PSK ""

#define EEPROM_SIZE 32          // EEPROM size
#define EEPROM_SDK_OFF 24       // EEPROM offset for sdk checker (size 4)

void check_sdk22 ()
{
  // all Serial.print can be commented out
  
  Serial.print("ESP reporting sdk version: ");
  Serial.println(ESP.getSdkVersion());

  // sketch is 128bytes bigger with PROGMEM, _P, pgm_read_byte
  static const char sdk22 [] = "2.2"; // 4 bytes in ram
  if (strncmp(ESP.getSdkVersion(), sdk22, sizeof(sdk22) - 1) == 0)
  {
    Serial.println("We are in sdk-2.2");

    EEPROM.begin(EEPROM_SIZE);
    char sdkee [sizeof(sdk22)];
    for (size_t i = 0; i < sizeof(sdk22) - 1; i++)
      sdkee[i] = EEPROM[EEPROM_SDK_OFF + i];
    sdkee[sizeof(sdk22) - 1] = 0;

    Serial.print("sdk in user-eeprom: ");
    Serial.println(sdkee);

    if (strncmp(sdkee, sdk22, sizeof(sdk22) - 1) == 0)
    {
      Serial.println("nothing to do");
    }
    else
    {
      for (size_t i = 0; i < sizeof(sdk22); i++)
        EEPROM[EEPROM_SDK_OFF + i] = sdk22[i];
      EEPROM.commit();
      bool done = ESP.eraseConfig();
      Serial.print("WiFi config erased: ");
      Serial.println(done);
      ESP.reset();
    }
  }
}

void setup() {
  Serial.begin(115200);
  Serial.setDebugOutput(true);
  check_sdk22();
  WiFi.mode(WIFI_STA);
  WiFi.begin(SSID, PSK);
}

bool connected = true;

void loop() {
  if (connected ^ (WiFi.status() == WL_CONNECTED))
  {
    connected = !connected;
    Serial.print("Connected: ");
    Serial.println(connected);
  }
}
```

PR #4384, opened 2018-02-19
Move cont_* funcs from ram to flash

kylefleming:
I think these can go in flash and not take up ram space since the functions that call them are all also located in flash.


cores/esp8266/cont_util.c
```diff
@@ -57,7 +57,7 @@ int ICACHE_RAM_ATTR cont_get_free_stack(cont_t* cont) {
     return freeWords * 4;
 }
 
-bool ICACHE_RAM_ATTR cont_can_yield(cont_t* cont) {
+bool cont_can_yield(cont_t* cont) {
```
igrr: This can be called from ISR, when cache is disabled, so it's best to keep it in IRAM.

cores/esp8266/cont_util.c
```diff
@@ -38,13 +38,13 @@ void ICACHE_RAM_ATTR cont_init(cont_t* cont) {
     }
 }
 
-int ICACHE_RAM_ATTR cont_check(cont_t* cont) {
+int cont_check(cont_t* cont) {
```
devyte: I'm calling this from an ISR, so it needs to be in RAM. 
My usage is as a kind of canary for stack usage. I suppose that would no longer be necessary when the breakpoint approach is implemented, but until then, I'd prefer this be kept in RAM.

cores/esp8266/cont_util.c
```diff
@@ -38,13 +38,13 @@ void ICACHE_RAM_ATTR cont_init(cont_t* cont) {
...
     if(cont->stack_guard1 != CONT_STACKGUARD || cont->stack_guard2 != CONT_STACKGUARD) return 1;
 
     return 0;
 }
 
-int ICACHE_RAM_ATTR cont_get_free_stack(cont_t* cont) {
+int cont_get_free_stack(cont_t* cont) {
```
devyte: I intend to call this from an ISR for statistical purposes (on my todo list for my application), so I'd prefer to keep it in RAM.

kylefleming:
Updated to integrate changes request by @igrr and @devyte.

devyte:
@igrr I think this is ok now.

PR #4389, opened 2018-02-19
WiFiClient: don’t destroy ClientContext on ::stop

igrr:
Reported in https://github.com/esp8266/Arduino/issues/4078.

WiFiClient::stopAll, called from a WiFi disconnected event handler,
could be called while WiFiClient::connect was in progress. This issue
was initially fixed in #4194, by testing `this` pointer for being
non-null in ClientContext::connect.

This change delegates deletion of ClientContext to WiFiClient
destructor. WiFiClient::stop only calls ClientContext::stop, which
closes/aborts the connection.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClient.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClient.cpp
@@ -272,8 +272,7 @@ void WiFiClient::stop()
     if (!_client)
         return;
 
-    _client->unref();
-    _client = 0;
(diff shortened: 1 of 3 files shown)


igrr:
@d-a-v This is the solution I mentioned in https://github.com/esp8266/Arduino/pull/4197, please have a look.

igrr:
Updated the PR to remove `if(this != 0)` check in `ClientContext::unref`.

PR #4391, opened 2018-02-19
Use Arduino code style for examples; check in CI

igrr:
Use code style defined in Arduino project to check code style of the examples:
https://github.com/arduino/Arduino/blob/master/build/shared/examples_formatter.conf

The check is done by formatting all example files using [astyle](http://astyle.sourceforge.net/astyle.html) and checking whether any changes have been introduced.

The change is split into two commits to make reviewing easier. First commit adds the code style rules and CI check, second commit re-formats all the examples.


tests/examples_style.conf
```diff
@@ -0,0 +1,49 @@
...
+indent-col1-comments
+
+# put a space around operators
+pad-oper
+
+# put a space after if/for/while
+pad-header
+
+# if you like one-liners, keep them
+keep-one-line-statements
```
earlephilhower: There was talk of replacing "keep-one-line-statements" w/"add-one-line-braces".  It's not 100% Arduino style, but it is IMO a better style as it avoids accidentally changing code flow when adding debug statements/etc.

tests/common.sh
```diff
@@ -213,8 +213,30 @@ function build_sketches_with_arduino()
...
+function check_examples_style()
+{
+    echo -e "travis_fold:start:check_examples_style"
+
+    find $TRAVIS_BUILD_DIR/libraries -name '*.ino' -exec \
+        astyle \
+            --suffix=none \
+            --options=$TRAVIS_BUILD_DIR/tests/examples_style.conf {} \;
+
+    git diff --exit-code -- $TRAVIS_BUILD_DIR/libraries
```
earlephilhower: I was wondering how you'd check this.  Neat!

devyte:
Per original comment, and per files changed, the styler seems to be applied only to .ino's.  Will it be applied to .h|.c|.cpp files as well?

igrr:
@earlephilhower regarding getting astyle from repos: I tried installing it using `apt`, but the version I got did not support some of the options used in Arduino astyle config file. I couldn't find a more recent astyle version in repos whitelisted by Travis, so went for building astyle from source.

With regards to changes going in before 2.4.1 — I've marked the PR with 2.5.0 milestone, I don't intend it to be merged now.

@devyte I think examples mostly consist of ino files. There could be .h and .cpp files here and there, and some better filter could be implemented to catch them, but this solution covers most cases and is simple enough. This can be revisited again when we start checking the code style for the core code and libraries.

Also I am intentionally not doing anything about the code style of core code and libraries in this PR. The style to be used for examples is fairly easy to decide on, since there is already a config file provided by Arduino project. As for the core code and libraries, it's a bit more tricky. We need to discuss this first.

tests/examples_style.conf
igrr: I agree with adding `add-one-line-braces`, but it is not the opposite of `keep-one-line-statements`. Without `keep-one-line-statements`, this:
```c++
do_foo(); do_bar();
```
will be changed to this:
```c++
do_foo();
do_bar();
```

Whereas `add-one-line-braces` will change
```c++
if (foo)
    bar();
```
into
```c++
if (foo)
    { bar(); }
```

If we *are* adding braces, i would probably go for `add-braces` instead, which would change the above into

```c++
if (foo) {
    bar();
}
```

devyte:
I agree about merging this post-2.4.1. 

As a side note, and just to vent it _somewhere_, I'm a big fan of Allman indentation style, and Arduino style makes me cringe. 
Having said that, if we're going to adopt a single style, of course Arduino style makes sense, and in any case, adopting Arduino style makes me cringe a lot less than looking at 20 different styles and weird inconsistent spacings and crazy indentations and other oddities I've come across. Whatever else, I consider this a _big_ improvement. 

> As for the core code and libraries, it's a bit more tricky. We need to discuss this first.

That answers my question. I'm eager to see the other files styled! Thank you for adopting the astyle idea!

tests/examples_style.conf
devyte: Personally, I prefer this:
if(condition)
  1-liner;
else
  other-1-liner;

over this:
if(condition) {
  1-liner;
} else {
  other-1-liner;
}

the reason is quite simple: it reduces complexity at a glance. But I don't consider it critical, and my glance can adapt.

Whatever else, this should be avoided:
if(condition) 1-liner;

igrr: Ah, i see now. Arduino config already has `add-brackets` option which was renamed to `add-braces` in latest astyle (but the old name is still supported).

So we are already converting 
```c++
if (foo)
    bar();
```
into
```c++
if (foo) {
    bar();
}
```

PR #4401, opened 2018-02-20
Reference to a arrow/_right that after doesn't exist

lrmoreno007:
Reference to a arrow/_right that after doesn't exist.

Please add label component:documentation if accepted.


earlephilhower:
The icon does exist and show fine for me.  Auto-pulls a github hosted image:
image

lrmoreno007:
Yes, the icon show fine! but it is not used later in any place. Everywhere "Read more" appears

Also that arrow does not exist in the Sphinx documentation

PR #4413, opened 2018-02-22
[WIP] ESP8266WiFi: initialize new fields, remove old workaround

igrr:
Initialize new .threshold field when setting up STA.


igrr:
Still WIP since there's something odd going on with WiFi.begin after WiFi.disconnect...

devyte:
This PR is needed anyways, so I'm merging now. The rest can be handled in a different PR.

Issue #1956, opened 2016-04-21
DHCP Options in AP mode


Manawyrm:
Hi,

with the ESP in AP mode, it acts as a DHCP server. 
when doing that, it hands out IP addresses to any client and also it tells them network information like the DNS servers and gateways. 

And thats exactly my problem. When a DHCP server hands out DNS server and default route-adresses, a mobile device like my android smartphone will try to use the ESP8266 as a router to get to the internet -- which will of course fail. 

This could be prevented by not sending the DHCP options 3 and 6. 
Screenshot: https://screenshot.tbspace.de/vetdnfuhizk.png

Is there anyway to disable or configure the DHCP payload which is sent by the ESP? 

Thanks,
Tobias
##

WereCatf:
In user_interface.h there is wifi_softap_set_dhcps_offer_option(), which, as far as I can tell, controls whether to send routing-information to clients or not. `wifi_softap_set_dhcps_offer_option(OFFER_ROUTER, 0)` should disable sending that.

I have not tried this, but after calling WiFi.begin() calling wifi_softap_set_dhcps_offer_option() might work.

igrr:
@Manawyrm did this solution work for you?

Manawyrm:
I unfortunatly can't test it at the moment, because I have lent out the ESP with the electronics to another team member who is developing on the hardware side of things.
I will come back to you as soon as possible. 

Sorry for the delay and keeping this open for so long!

datacute:
The code to include various options is here:
https://github.com/esp8266/Arduino/blob/18f66e99693020a758b8f5ac92e042cbdecb7760/tools/sdk/lwip/src/app/dhcpserver.c#L129

NarinLab:
Is there any clear method to customize the DHCP Configuration in ESP8266 Arduino?
I need to customize my DHCP range from 192.168.244.x to 192.168.43.x.
There is not enough documentation on how to setting NodeMCU/ESP8266 Arduino Language DHCP settings in AP Mode.

Thanks!

WereCatf:
@NarinLab Take a look at http://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/soft-access-point-class.html#softapconfig -- if you specify the ESP8266's IP-address as e.g. 192.168.43.1, the function softAPConfig() sets the DHCP-range as 192.168.43.100 - 192.168.43.200.

[...]

PR #4421, opened 2018-02-23
Allow to disable DHCP gateway offer when set up of AP.

tfry-git:
When running the ESP8266 in AP mode, it will advertise a default gateway via DHCP, even if it cannot route traffic anywhere. This patch simply disables the gateway offer, when the gateway is set to 0.0.0.0.

See also:
-  https://github.com/esp8266/Arduino/issues/1956
- http://www.esp8266.com/viewtopic.php?f=6&t=1798


tfry-git:
While this really does not seem smart on the of the DHCP client, mine (dhclient on Ubuntu 16.04) happily adds a default route via 0.0.0.0 when advertised. That's the reason for this patch.

tfry-git:
Ping!

Not to add noise, but I think this has simply been forgotten.

PR #4444, opened 2018-02-27
Fixes #641 "Change SD Card"

### LINKED ISSUE #641: Change SD Card
In some of my Projects I need to change the SD card. There is a simple allow this. I just require a single line in the SD.c library.

```
boolean SDClass::begin(uint8_t csPin) {
  /*
    Performs the initialisation required by the sdfatlib library.
    Return true if initialization succeeds, false otherwise.
   */
// ----> Line Added
  if (root.isOpen()) root.close();      // allows repeated calls
// <--- Line Added
  return card.init(SPI_HALF_SPEED, csPin) &&
         volume.init(card) &&
         root.openRoot(volume);
}
```

After changing the SD card all you need is to reinitialize the sd card.

Same Issue is posted in arduino as no 3607

sirzeta:
Added code from PR 38 from arduino-libraries/SD to fix SD card re-initialization.


devyte:
Are you aware of pr #3654 ? I've been wondering whether closing root was enough.

sirzeta:
Indeed, there must be a better way, at least with this fix I can call the begin function again if I change the SD memory, the PR 3654 touches the SPI connection, which I may not want if I have a configuration with multiple slaves. Anyway, we need to initialize a SD again, just a "end()" method would be nice or gain access to "card" and "root" somehow to know the SD card status and manually control it.

sirzeta:
I did other solution, I think it's a better approach withouth touch many files, now I can close the SD connection with some checks before close than just using root.close() (SdFile::close()) alone. SdFile::close() calls SdFile::sync() and makes some operations, not needed if not SD is present or Sd2Card has errors.

As stated in code: The sync() call causes all modified data and directory fields to be written to the storage device.

Calling constructors of Sd2Card, SdVolume, SdFile an set all related SD private variables, I think it's the safe way to initialize again by calling it's init() functions. 

I tried this code disconnecting/connecting SD Card, shut down and power up again the ESP8266, connecting SD, disconnecting again and modifying some file in PC and plug back again, and many more combinations, works OK so far.

There no way to check if the connected SD is the same, as author said in some post in arduino.cc, there is a way if the SD module has a additional pin that toggles it's state if SD is connected (by mechanical function), but this it's not related with SPI connection itself.

devyte:
Can you show a small example of the solution?

sirzeta:
Steps:

1. Press `m` to mount
2. Press `u` and `y` to unmount
3. Press `m` to mount again, if SD.end() is commented, you are in a endless "SD initialization failed!" loop.

```c++
File myFile;
bool mounted = false;

void setup() {
    Serial.begin(115200);

}

void loop() {
    if (mounted && Serial.peek() == 'u') {
        Serial.read();
        Serial.println("Unmount SD and press 'y' to continue");
        while (Serial.peek() != 'y') {
            delay(1000);
        }
        mounted = false;
        Serial.println("SD unmounted.");
        Serial.read();
    } else if (!mounted && Serial.peek() == 'm') {
        Serial.read();

        //Uncomment this to finish while (!SD.begin(D2)) endless loop
        //SD.end();

        while (!SD.begin(D2)) {
            Serial.println("SD initialization failed!");
            delay(1000);
        }
        mounted = true;
        Serial.println("Initialization succeded!");
    }

    if (mounted) {
        myFile = SD.open("test.txt");
        Serial.println("Printing test.txt contents:");
        while (myFile.available()) {
            Serial.write(myFile.read());
        }
        myFile.close();
        Serial.println();
        Serial.println("All contents of test.txt printed!");

        Serial.println("SD mounted, press 'u' to umount");
    } else {
        Serial.println("SD unmounted, press 'm' to mount");

    }

    Serial.println("Wait 3 seconds...");
    delay(3000);
}
```

devyte:
@sirzeta Thanks for this, and sorry I took so long to get to it.
You're right, in PR #3654 , the SPI is blindly shut down, which is bad if you have multiple slaves.
However, I don't like the fact that begin() inits the SPI, but end() doesn't shut it down, and this is the case that beginners will hit.
To cover both cases, how about a flag passed to end() that specifies whether SPI should be shut down or not? Set it to true as default to bring SPI down, which covers the use case for beginners. Advanced users with multiple slaves can call end(false) to keep the SPI up.

libraries/SD/src/SD.cpp
```diff
@@ -350,6 +350,17 @@ boolean SDClass::begin(uint8_t csPin, uint32_t speed) {
...
 }
 
+// Ends SD card connection
+void SDClass::end() {
+    if(card.errorCode() == 0 && root.isOpen()) {
+        root.close();
+    }
+
+    root = SdFile();
+    volume = SdVolume();
```
devyte: This may be incompatible with #4204 .

libraries/SD/src/SD.cpp
```diff
@@ -350,6 +350,17 @@ boolean SDClass::begin(uint8_t csPin, uint32_t speed) {
          root.openRoot(volume);
 }
 
+// Ends SD card connection
+void SDClass::end() {
+    if(card.errorCode() == 0 && root.isOpen()) {
```
devyte: Can root be open but card have an errorCode? What happens if root.close() is called regardless of the errorCode? i.e.:
```
if(root.isOpen())
  root.close();
```

devyte:
@sirzeta Are you available to continue work on this?

sirzeta:
> @sirzeta Are you available to continue work on this?

I'm little busy this days, next week I'll start working on it again.

devyte:
Ok. I've investigated a bit, and I think I understand what you were getting at about sync().
In the following abnormal scenario:
  //Insert SD Card A
  SD.begin()
  //do operations
  //remove card A
  //insert SD card B
  SD.end()

I think that it's possible that:
 - card A becomes corrupted due to removal before calling SD.end().
 - card B becomes corrupted if there were ops pending for card A at the time SD.end() is called with card B.

However, not calling sync on end() can cause pending ops to never be finished, which is bad for normal operation. I therefore think that calling sync() is necessary.

For now, I'm going to push a compromise implementation between this and #3654 , which also allows fixing #4204 without danger. It is not safe in the card swapping scenario described above, but it gets us one step forward. We can figure out the sync thing later.

devyte:
Closing in favor of #5402 .

Issue #3657, opened 2017-09-26
getCoreVersion reports 00000000 on Windows


tablatronix:
Is this normal ? Is there something wrong with my platform environment ?

davisonja:
Are you using the latest version from git, or the latest release package?

tablatronix:
ahhhhh, I am using platformio staging, which uses repo not release.

So this is normal I assume

davisonja:
I believe the coreversion gets computed and stored as part of the release
build process, so this would be expected for repo usage.
I haven't looked, so I'm not sure how simple it would be to change that,
but if people are using direct from repo it might be useful if it could be
something more meaningful (so you can actually tell the version from code,
which I presume is the aim :)

igrr:
It should work for git version, if you are using Arduino IDE. There is a
hook in platform.txt which launches git to figure out the version.
Might be something that hasn't been ported over to platformio build script.

davisonja:
Ah, okay. The version I currently have in operation has `## windows-compatible version may be added later
` which is why it's not done that for me using Windows :)

tablatronix:
I see the hook, let me see if i can figure something out with platformio

tablatronix:
Ahhh bash..

tablatronix:
Looks like I am having issues with this on osx platformio with release
`ARDUINO_ESP8266_GIT_VER`

d-a-v:
I ran into this too.
I just discovered that, from inside core/esp8266/*.cpp
```#include <core_version.h>``` is well behaving
```#include "core_version.h"``` is not
The default file with generic default value lives in the same directory, the other one is generated by platform.txt's recipes in arduino's os-dependant temporary directory. I don't know details about platformio.

d-a-v:
Can you try [my alpha release channel](https://d-a-v.github.io/) to check if this issue is fixed ?

d-a-v:
This issue has received multiple corrections over time, latest related ones are: #6735, #6745 already fixed in git master, verified on the alpha release channel.
Closing. Please wait for core-2.6.1.

PR #4467, opened 2018-03-06
provide full version descriptor, displayed in debug mode

d-a-v:
* (wip)
* This is intended to help users showing their running version when reporting.
* The version string is given by ```ESP.getFullVersion()```
* Version is displayed if debug on serial is enabled, or if ```Serial.setDebugOutput(true)``` is called.
* Uses ```git describe --tags``` on linux, or ```win-2.4.0post``` on windows (set inside platform.txt)

caveats:
* Travis may not like core_version.h disappearance
* some macro format are modified
* global variable ```core_version``` removed
* Strings may be too long, it currently shows:
  * ```Boot:31/SDK:2.2.1(cfd48f3)/Core:2.4.0-94-g7999f5c/lwIP:1.4.0rc2```
  * ```Boot:31/SDK:2.2.1(cfd48f3)/Core:2.4.0-95-gd1f9c14/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.0-5-gafb60c2)```


cores/esp8266/core_esp8266_main.cpp
```diff
@@ -41,6 +41,7 @@ extern "C" {
 
 struct rst_info resetInfo;
 
+#if 0
 extern "C" {
 extern const uint32_t __attribute__((section(".ver_number"))) core_version = ARDUINO_ESP8266_GIT_VER;
```
igrr: If you remove this, then eboot will no longer be able to print the core version:
https://github.com/esp8266/Arduino/blob/7999f5c29c653e6e0b8c77ea42334df8df256a23/bootloaders/eboot/eboot.c#L21-L33

igrr: (which i personally use to check that I'm running correct version; but if you think it should be removed i can find some other way)

cores/esp8266/HardwareSerial.cpp
```diff
@@ -94,6 +97,9 @@ void HardwareSerial::setDebugOutput(bool en)
     if(en) {
         if(uart_tx_enabled(_uart)) {
             uart_set_debug(_uart_nr);
+#if !defined(DEBUG_ESP_PORT) && !defined(NDEBUG)
+            os_printf("%s\r\n", ESP.getFullVersion().c_str());
+#endif
```
igrr: I'm fine with printing the version in Serial::begin if debug output is enabled... however if i'm calling setDebugOutput i probably know what i'm doing and can handle calling ESP.getFullVersion if needed.

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -150,7 +152,7 @@ void init_done() {
     system_set_os_print(1);
     gdb_init();
     do_global_ctors();
-    printf("\n%08x\n", core_version);
+    //printf("\n%08x\n", core_version);
```
igrr: This seems to be some debugging leftover — it is not used and can be removed.

platform.txt
```diff
@@ -6,7 +6,7 @@
 # https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5---3rd-party-Hardware-specification
 
 name=ESP8266 Modules
-version=2.5.0
+version=2.4.0-post
```
igrr: According to semver, 2.4.0-post is something *before* 2.4.0.

It might be 2.4.1-pre if we intend to merge this before 2.4.1?

platform.txt
```diff
@@ -70,9 +70,9 @@ compiler.elf2hex.extra_flags=
 
 ## generate file with git version number
 ## needs bash, git, and echo
-recipe.hooks.core.prebuild.1.pattern=bash -c "mkdir -p {build.path}/core && echo \#define ARDUINO_ESP8266_GIT_VER 0x`git --git-dir {runtime.platform.path}/.git rev-parse --short=8 HEAD 2>/dev/null || echo ffffffff` >{build.path}/core/core_version.h"
-## windows-compatible version may be added later
-recipe.hooks.core.prebuild.1.pattern.windows=
+recipe.hooks.core.prebuild.1.pattern=bash -c "mkdir -p {build.path}/core && echo \#define ARDUINO_ESP8266_GIT_VER `cd {runtime.platform.path}; git describe --tags 2>/dev/null || echo nix-{version}` >{build.path}/core/core_version.h"
```
igrr: Can we keep `ARDUINO_ESP8266_GIT_VER` to be a short hash (expressed as an integer), and add a new define for the `git describe` output instead? E.g. `ARDUINO_ESP8266_GIT_DESC`. I have some release scripts which rely on ARDUINO_ESP8266_GIT_VER being what it is.

igrr: Also, not necessarily `nix-` version. Also works on macOS, as far as i can tell.

igrr:
I think the idea with core_version.h was that the generated file would be placed into the build directory, so that the compiler picks it up before the one in the core. If this is broken for some reason, maybe it means that the IDE has started calling arduino-builder in some different way...

devyte:
@igrr maybe that's the reason for #3657 ? I had to delete core_version.h in my local repo to get the version to work correctly.

cores/esp8266/Esp.cpp
```diff
@@ -550,3 +556,41 @@ String EspClass::getSketchMD5()
...
+#else // !NDEBUG
+
+static const char arduino_esp8266_git_ver [] PROGMEM = STR(ARDUINO_ESP8266_GIT_VER);
+#if LWIP_VERSION_MAJOR != 1
+static const char lwip2_version [] PROGMEM = "/lwIP:" STR(LWIP_VERSION_MAJOR) "." STR(LWIP_VERSION_MINOR) "." STR(LWIP_VERSION_REVISION);
+#endif
+
+String EspClass::getFullVersion()
+{
+    return   String(F("Boot:")) + system_get_boot_version()
```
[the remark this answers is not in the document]
d-a-v: OK, will be removed

cores/esp8266/core_esp8266_main.cpp
d-a-v: I personally don't mind, I could not find where it was used in the core.

cores/esp8266/HardwareSerial.cpp
d-a-v: Removing this one will also allow the linker to get rid of the code if sources are moved to a separate file.

platform.txt
d-a-v: Will rename to GIT_DESC, and leave GIT_VER as before. About nix, it's a typo, unix was intended (for linux + macos).

d-a-v:
I hope we have a better view of what's happening in issues after next release.

devyte:
Some questions on this:
1. is a separate file for this really necessary? I'd prefer to stick to 1-cpp-per-object-definition.
2. This code introduces an entirely new way of retrieving versions, which is independent of the existing ones, and has duplication of function. Rather than implement a new way, does it make sense to enhance the current Esp::getBlahVersion() methods, add a new one for lwip, and implement this new Esp::getFullVersion() as a function of the others?
3. Why is this available only when building without NDEBUG? It seems inconsistent wrt the other Esp::getBlahVersion() methods, which are always available.

d-a-v:
> is a separate file for this really necessary? I'd prefer to stick to 1-cpp-per-object-definition.

If this function is not used, then it takes no flash (~400bytes just to print a ~40bytes string) even if another EspClass:: one is used. It's because we don't have ```-flto``` yet.

> This code introduces an entirely new way of retrieving versions, which is independent of the existing ones, and has duplication of function. Rather than implement a new way, does it make sense to enhance the current Esp::getBlahVersion() methods, add a new one for lwip, and implement this new Esp::getFullVersion() as a function of the others?

To be honest, I'm quite bothered when it comes to string manipulations and all those calls to malloc making tiny holes in ram tickle me. That's the reason why I enclosed them in a big function. It's rather useless, but it's all put in a single place. Still you're right, let's keep the modular way and let's expect gcc5 (or is it gcc49?) to come soon!
(: Or, we may use a big goodold C snprintf :)

> Why is this available only when building without NDEBUG? It seems inconsistent wrt the other Esp::getBlahVersion() methods, which are always available.

I put this NDEBUG conditional when the function was in Esp.cc because people using this option want to save more flash. It makes indeed no sense to keep it in the separate file.

cores/esp8266/Esp-version.cpp
```diff
@@ -0,0 +1,46 @@
+
```
igrr: License header missing here

cores/esp8266/HardwareSerial.cpp
```diff
@@ -39,6 +39,10 @@ void HardwareSerial::begin(unsigned long baud, SerialConfig config, SerialMode m
 {
     end();
     _uart = uart_init(_uart_nr, baud, (int) config, (int) mode, tx_pin, _rx_size);
+#if defined(DEBUG_ESP_PORT) && !defined(NDEBUG)
+    println();
+    println(ESP.getFullVersion());
```
igrr: Unrelated to printing the version, but since you're adding this code block here. Let's call `setDebugOutput(true);` or the equivalent `uart_` function. At the moment debug output only gets enabled before `loop` runs, so you don't get debug output while `setup` is running. Then these three lines can be removed: https://github.com/esp8266/Arduino/blob/7999f5c29c653e6e0b8c77ea42334df8df256a23/cores/esp8266/core_esp8266_main.cpp#L118-L120

d-a-v: To understand well, setDebugOutput(true) (= sdk debug messages) is then not automatically enabled anymore when Serial debug is enabled, we would not have espressif debug messages anymore, but only this "full version" displayed and selected arduino-core debug messages ?
For anywone willing to get sdk debug messages, then setDebugOutput(true) has to be called.

That's sound nice to me.

igrr: I'm suggesting that we move the call to `setDebugOutput(true);` a bit earlier, if DEBUG_ESP_PORT is set. 
Currently it happens only before `loop`, so some debug messages during setup may be lost.
If we do this in HardwareSerial::begin, then we will see all debug messages. So the feature merely becomes more consistent.

(We also need `if (&DEBUG_ESP_PORT == this) {` check before enabling debug output, in case DEBUG_ESP_PORT is the HardwareSerial which we are initializing).

d-a-v: Ok to summarize, we **should** automatically enable sdk messages along with core messages, but before setup not after.

OK about Serial/Serial1 distinction.

cores/esp8266/Esp-version.cpp
d-a-v: added

package/build_boards_manager_package.sh
```diff
@@ -88,6 +88,7 @@ $SED 's/recipe.hooks.core.prebuild.1.pattern.*//g' \
 ver_define=`echo $plain_ver | tr "[:lower:].\055" "[:upper:]_"`
 echo Ver define: $ver_define
 echo \#define ARDUINO_ESP8266_GIT_VER 0x`git rev-parse --short=8 HEAD 2>/dev/null` >$outdir/cores/esp8266/core_version.h
+echo \#define ARDUINO_ESP8266_GIT_DESC `git describe --tags 2>/dev/null` >>$outdir/cores/esp8266/core_version.h
```
igrr: A few lines above (can't comment on the line directly...) there is
```
$SED 's/recipe.hooks.core.prebuild.1.pattern.*//g' \
```
Since you have added 2.pattern, please add one more line, then run the script and check the the prebuild hooks are removed from platform.txt in the built archive.

d-a-v: Done. I had to rebase to push the fix.

PR #4471, opened 2018-03-06
lwip2: fix disconnection/reconnection issue

d-a-v:
also:
improve version string
remove useless message


igrr:
So if `ip == 0`, we clear `NETIF_FLAG_LINK_UP`, is that right? I'm wondering whether we should still indicate that the link is up even if IP is not yet obtained...

d-a-v:
The thing is upstream lwip-1 does not give much importance to netif flag UP regarding dhcp client, so espressif implementation does not care: On the esp side, an interface is down if it does not have the LINK_UP flag and I think it does not care at all about UP.

Upstream lwip-2's dhcp client needs netif UP and the glue (lwip2) deals with that.

You are right, I was too violent. So without too many changes into lwip-2 side, LINK_UP is always set, and we play with the UP flag to know whether it i enabled or not. Espressif blobs are not aware of lwip-2 anyway.

igrr:
Okay, the new diff makes sense to me. Is there something that needs to be tested here, or should i go ahead and merge once the CI passes?

PR #4474, opened 2018-03-07
Update ISSUE_TEMPLATE.md

devyte:
Proposal for issue template update


ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,60 @@
...
+- [ ] I have filled out all fields below.
 
-### Description
+#### Platform
 
-Problem description
+- Hardware:					[ESP-12]
+- Core Version:			[latest git hash or date]
+- Development Env:		[Arduino IDE|Platformio|Make|other]
+- Operating System:		[Windows|Ubuntu|MacOS]
```
earlephilhower: Can we please add LWIP version as well?

earlephilhower: Errr...maybe this belongs in the next section, not here, but we still need LWIP.

ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,62 @@
+----------------------------- Delete below -----------------------------
```
earlephilhower: I'm happy to approve as-is, as I think it presents the state of affairs for people who actually read it, but maybe right at the top you could add something like the below to direct people to the proper spot for 50% of the silly issues that are opened?  I should have thought of it on my first comment....

````
Have a general question?
Don't know how to do something?
Trouble with a third-party library?
Please go to https://www.esp8266.com for the support forum to get answers!

Only continue with this issue report if you have a failure or bug with the current core libraries (i.e. crashes, hangs, etc.)
````

File: ISSUE_TEMPLATE.md [hunk 3]
```diff
@@ -1,36 +1,62 @@
...
+----------------------------- Delete above -----------------------------
 
 ### Basic Infos
 
-#### Hardware
-Hardware:			?ESP-12?
-Core Version:      	?2.1.0-rc2?
+- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
+- [ ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
+- [ ] I have tested that the issue is present in current master branch (aka latest git).
```
igrr: That might be a bit harsh, no? I think it's fair that we also accept issue reports for release versions and ask to test with latest git if needed. How about "I have searched the issue tracker for a similar issue" instead?

ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,62 @@
...
 
-if you have a stack dump decode it:
 https://github.com/esp8266/Arduino/blob/master/doc/Troubleshooting/stack_dump.rst
 
-for better debug messages:
-https://github.com/esp8266/Arduino/blob/master/doc/Troubleshooting/debugging.rst
+6. Include a Minimal Complete Reproducible Example sketch that shows your issue. Do not include your entire project, or a huge piece of code.
+7. Include debug messages:
+
+https://github.com/esp8266/Arduino/blob/master/doc/Troubleshooting/debugging.rst
```
igrr: How about using a readthedocs link here?

ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,62 @@
...
+- Upload Speed:   [115200|other] (serial upload only)
 
 
-### Sketch
+### Problem Description
+
+Detailed problem description goes here.
+
+
+### MCVE Sketch
```
igrr: Let's make `MCVE` a link to https://stackoverflow.com/help/mcve

ISSUE_TEMPLATE.md
devyte: Normally I'd agree, but our release model is linear wrt to development, i.e.: our releases are a snapshot of master, and we don't use branches for released versions (the typical "stable release" model). As a result, It doesn't make much sense to track issues that apply to a particular release, except maybe the very last release, because there are no release branches to port a fix to.
Whatever else, the current reality is that we're getting a whole lot of duplicate issues. A release has an issue, we get a report, we fix it in master, then we keep getting multiple reports for the same issue, and we can't always tell that all those additional issues are the same one. Then, when an issue comes up that isn't obvious, the first thing we do is ask to test with latest git anyways. This wastes a whole lot of time, both on our end because of dealing with those issues, and on the user's end because of the turnaround and turnaround back again.

My thinking is that installing git is easy enough, and it's even documented. The user is supposed to do his homework, but many are a bit lazy, or just don't know, so don't do enough research prior to opening an issue. So anything we can do to "encourage" a user to do due diligence, that reduces our own overhead, is good from my pov.
So I think your idea to search the issue tracker is a good one, but as an additional item.

ISSUE_TEMPLATE.md
devyte: Agreed, same for the stack dump.

ISSUE_TEMPLATE.md
devyte: Oops, I thought I had put the link somewhere already.

PR #4493, opened 2018-03-10
Enhanced EEPROM library (ESP_EEPROM)

jwrw:
The current EEPROM library uses flash to emulate EEPROM storage.  However it needs to re-flash each time a change needs to be committed.  This is quite slow and interrupts need to be disabled whilst it happens (which breaks PWM etc.)
This revised library avoids re-flashing by writing committed data to a fresh area of flash each time.  The flash only needs to be re-erased when you run out of fresh space.  A bitmap is used to keep track of the space used.  All this is pretty much hidden from the user of the library (can be used as a drop-in replacement for EEPROM).

### DIFF
--- a/libraries/ESP_EEPROM/.gitattributes
+++ b/libraries/ESP_EEPROM/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
--- a/libraries/ESP_EEPROM/.gitignore
+++ b/libraries/ESP_EEPROM/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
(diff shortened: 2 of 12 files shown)


igrr:
Hi @jwrw, thanks for submitting this library, looks interesting! Have you considered distributing it through Arduino Library Manager? It might offer you an easier way of maintaining it (without having to submit PRs into this project each time) and more flexible release cycles, not tied to the rest of the core.

jwrw:
Yeah - thanks for looking. I wasn't sure myself of the best route as I was a little concerned it was ESP8266-specific.  But looking again it seems like this may not be a problem.  

I'll give that a go...

igrr:
Once you get this set up in library manager, feel free submit a PR with a link to your library into doc/libraries.rst file!

Issue #2421, opened 2016-08-22
clean up SPI_FLASH_SEC_SIZE and/or FLASH_SECTOR_SIZE


holgerlembke:
FLASH_SECTOR_SIZE is defined in cores/esp8266/flash_utils.h, while the SPI_FLASH_SEC_SIZE is defined in tools/sdk/include/spi_flash.h. They have both the same value and I assume they shall have the same meaning.

EEPROM uses SPI_FLASH_SEC_SIZE.

just to not confuse ppl...

devyte:
The referenced code is still present in latest git.
I think FLASH_SECTOR_SIZE  should be #defined in terms of SPI_FLASH_SEC_SIZE, which comes from the Espressif SDK.

PR #4509, opened 2018-03-13
Code cleanup

fabianoms:
FLASH_SECTOR_SIZE defined in terms of SPI_FLASH_SEC_SIZE (issue #2421).


devyte:
@fabianoms Thanks, but looks like something is still missing, because CI didn't build. Can you figure it out?

fabianoms:
@devyte just one check failing now. Looks like the host tests don't reference files from "tools/sdk/include/".  I'm not sure if I forgot to add something in flash_utils.h or I have to edit the "Arduino/tests/host" to add some reference to sdk files.

devyte:
@fabianoms I was just looking at the same. Honestly, I'm not sure what's the best way to go here.
@igrr would it be correct to copy c_types.h from the sdk include dir into the host/common dir, and edit as necessary, similar to how it's done with Arduino.h?

igrr:
TBH, i would much rather not pull the stuff defined in c_types.h into the global namespace. Easiest is probably replace all `uint32` with `uint32_t` in spi_flash.h. But then, spi_flash.h also declares a bunch of functions. Do we need them there, just for the flash sector size definition? I'm not sure.

devyte:
Closing in favor of #5327 .

PR #4510, opened 2018-03-13
add option to override fingerprint with *

r-downing:
For some projects which make requests from a remote server that requires https, putting a fingerprint in the code is not an option because it may change at any moment.

This gives the ability to make https requests using a "*" to intentionally override the fingerprint verification. Using the "*" rather than omitting it completely makes it apparent that it is intentionally being overridden. While this may be slightly less secure in some cases, depending on the user's application it may also add a lot of functionality.

For example, `http.begin("https://www.meethue.com/api/nupnp", "*");`


igrr:
Closing for the reasons described here: https://github.com/esp8266/Arduino/pull/3933#issuecomment-353241176

PR #4537, opened 2018-03-20
unefficient Print:write(data,len) shows message if used (only in debug mode)

d-a-v:
+fix HardwareSerial that used it


cores/esp8266/Print.cpp
```diff
@@ -33,10 +33,20 @@
...
+        once = true;
+        os_printf_plus(not_the_best_way);
+    }
+#endif
+
     size_t n = 0;
     while (size--) {
         size_t ret = write(*buffer++);
-        if (ret == 0) {
+        if (ret <= 0) {
```
devyte: ret is declared of type size_t, which should be unsigned, which means it can't be negative.

cores/esp8266/uart.c
```diff
@@ -196,23 +196,30 @@ void uart_stop_isr(uart_t* uart)
 }
 
 
-void uart_write_char(uart_t* uart, char c)
+void uart_do_write_char(uart_t* uart, char c)
```
devyte: Maybe declare this one as inline.

cores/esp8266/Print.cpp
d-a-v: Yes very true. Blinded with ::read, arduino's api should all get a major reshape for coherency.

cores/esp8266/uart.c
d-a-v: Thanks. I actually don't understand why inline does save flash space here.
This means one while loop and 2 calls with 2 parameters each take 16 more flash bytes than two while loops.
I was thinking gcc -Os would optimize this itself [...] Well, it would have if I had not forgotten ```static```. Then I'd prefer the static way.

d-a-v:
Last "remove duplicate code" commit spares ~100 bytes of flash.
I also gain ~1kbps @878kbps in a serial loop BW test (0.1% yeah :)

Issue #4496, opened 2018-03-10
Support I2S audio input


renne:
Please support I2S audio input on the ESP8266 for bidirectional audio streaming. This can be used e.g. for intercoms, VoIP, voice control with audio notifications, ...

devyte:
CC @earlephilhower

earlephilhower:
I've been looking at some existing code base that enables [I2S input](https://github.com/joextodd/listener).  Seems like it could be added to core_esp8266_i2s without too much grief as it's only setting up another set of dma buffers, starting the slc dma engine, and adding an api to get received data.

Setting this up for the n+2 release for now.

joextodd:
If I can help at all with what I have learned from my [listener](https://github.com/joextodd/listener) project, I would be happy to contribute.

earlephilhower:
If either of you already has a microphone hooked up, can you look at the new PR #4539 ?  I've taken the code in @joextodd's repo as a start, and combined it with the existing I2S code that support Xmit in Arduino core.  It's based on the same idea as the current xmit, where your app needs to i2s_read_sample() in its loop() to get the data and do work (i.e. requantize to 8-bits, packetize, and send as RTSP or whatever).

It's definitely not done, but it is returning samples that make sense to some degree in my limited testing (resistor from D6->Gnd/VCC on the D1 Mini I have running).  Overflow (i.e. not reading fast enough) will cause some interesting artifacts as-is, for example.

If the ESP8266 has a good HW reference that discusses its I2S block in more detail than the ESP8266-Tech-Ref, I sure as heck can't find it.

earlephilhower:
My test app:
````
void setup() {
  Serial.begin(115200):
  i2s_rxtx_begin(true, false);
  i2s_set_rate(10); // Your app should use 44100 or whatever, this lets me debug DMA descriptors
  while(1) {
    uint32_t l, r;
    i2s_read_sample(&l, &r, true);
    Serial.printf("%08x %08x\n", l, r);
  }
}
void loop() { }
````

pashinen:
Is there any other mic than SPH0645 as i need frequency range between 20HZ to 20 KHZ.

earlephilhower:
@pashinen Sorry, I've only used existing pcbs.  Dual Invensense mikes are in the Google AIY project so they definitely must be low cost.

Closing since the I2S input was added a while back and no feedback from the original requesters.

PR #4539, opened 2018-03-21
I2s input API and examples

earlephilhower:
-edit- Ready for testing/review.

Runs well enough to send I2S recorded data to a PC and play over its speakers.  Pretty cool!

Rewrite the I2S utils to support stereo I2S microphone inputs.

Add API calls:
* i2s_rxtx_begin(bool rx, bool tx);
* bool i2s_read_sample(int16_t *l, int16_t *r, bool blocking);

See issue #4496 for the trail.


earlephilhower:
Looks like the new code is working to some degree.  Graphing the i2s_read_sample(l,r,nb) gives:
image

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+
+typedef struct i2s_state {
+  uint32_t *       slc_queue[SLC_BUF_CNT];
+  volatile uint8_t slc_queue_len;
+  uint32_t *       slc_buf_pntr[SLC_BUF_CNT]; // Pointer to the I2S DMA buffer data
+  slc_queue_item_t slc_items[SLC_BUF_CNT]; // I2S DMA buffer descriptors
+  uint32_t *       curr_slc_buf; // Current buffer for writing
+  uint32_t         curr_slc_buf_pos; // Position in the current buffer
+  void             (*callback) (void);
+  // Callback function should be defined as 'void ICACHE_FLASH_ATTR function_name()',
```
devyte: Can the callback really be in Flash? Shouldn't this read "... defined as void ICACHE_RAM_ATTR function_name()"?

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+volatile int rx_irqs = 0;
+volatile int tx_irqs = 0;
+
+// Some constants that aren't defined in i2s_regs.h
+#define I2SI_DATA 12
+#define I2SI_BCK  13
+#define I2SI_WS   14
+
+
+static bool ICACHE_FLASH_ATTR _i2s_is_full(const i2s_state_t *ch) {
```
devyte: why do all of these have the ICACHE_FLASH_ATTR decorator, when that is our default?

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+  if (!ch) {
+    return false;
+  } else {
```
devyte: remove else, no need for it. Just let the flow get to the following return statement.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+}
+
+bool ICACHE_FLASH_ATTR i2s_rx_is_full() {
+  return _i2s_is_full( rx );
+}
+
+static bool ICACHE_FLASH_ATTR _i2s_is_empty(const i2s_state_t *ch) {
+  if (!ch) {
+    return false;
+  } else {
```
devyte: Remove else.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+bool ICACHE_FLASH_ATTR i2s_rx_is_empty() {
+  return _i2s_is_empty( rx );
 }
 
-bool ICACHE_FLASH_ATTR i2s_is_empty(){
-  return (i2s_slc_queue_len >= SLC_BUF_CNT-1);
+static int16_t ICACHE_FLASH_ATTR _i2s_available(const i2s_state_t *ch) {
+  if (!ch) {
+    return 0;
+  } else {
```
devyte: Remove else.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
     memset((void *)finished_item->buf_ptr, 0x00, SLC_BUF_LEN * 4);//zero the buffer so it is mute in case of underflow
-    if (i2s_slc_queue_len >= SLC_BUF_CNT-1) { //All buffers are empty. This means we have an underflow
-      i2s_slc_queue_next_item(); //free space for finished_item
+    if (tx->slc_queue_len >= SLC_BUF_CNT-1) { //All buffers are empty. This means we have an underflow
+      i2s_slc_queue_next_item(tx); //free space for finished_item
     }
-    i2s_slc_queue[i2s_slc_queue_len++] = finished_item->buf_ptr;
-    if (i2s_callback) i2s_callback();
+    tx->slc_queue[tx->slc_queue_len++] = finished_item->buf_ptr;
+    if (tx->callback) tx->callback();
```
devyte: No one-line ifs please.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+    ETS_SLC_INTR_ENABLE();
+  }
+  if (slc_intr_status & SLCITXEOF) {
+    ETS_SLC_INTR_DISABLE();
+    rx_irqs++;
+    slc_queue_item_t *finished_item = (slc_queue_item_t *)SLCTXEDA;
+    finished_item->owner = 1; // Or else RX just stops
+    finished_item->myid++;
+    i2s_slc_queue_append_item(rx, finished_item->buf_ptr);
+    if (rx->callback) rx->callback();
```
devyte: 1-line if

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
-    i2s_slc_items[x].unused = 0;
-    i2s_slc_items[x].owner = 1;
-    i2s_slc_items[x].eof = 1;
-    i2s_slc_items[x].sub_sof = 0;
-    i2s_slc_items[x].datalen = SLC_BUF_LEN*4;
-    i2s_slc_items[x].blocksize = SLC_BUF_LEN*4;
-    i2s_slc_items[x].buf_ptr = (uint32_t)&i2s_slc_buf_pntr[x][0];
-    i2s_slc_items[x].next_link_ptr = (int)((x<(SLC_BUF_CNT-1))?(&i2s_slc_items[x+1]):(&i2s_slc_items[0]));
+    ch->slc_buf_pntr[x] = malloc(SLC_BUF_LEN*4);
+    for (y=0; y<SLC_BUF_LEN; y++) ch->slc_buf_pntr[x][y] = 0;
```
devyte: 1-line for

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+    ch->slc_items[x].eof = 1;
+    ch->slc_items[x].sub_sof = 0;
+    ch->slc_items[x].datalen = SLC_BUF_LEN * 4;
+    ch->slc_items[x].blocksize = SLC_BUF_LEN * 4;
+    ch->slc_items[x].buf_ptr = (uint32_t*)&ch->slc_buf_pntr[x][0];
+    ch->slc_items[x].next_link_ptr = (x<(SLC_BUF_CNT-1))?(&ch->slc_items[x+1]):(&ch->slc_items[0]);
+    ch->slc_items[x].myid = 0;
+  }
+}
+#if 0
```
devyte: Does this #if 0 block serve any purpose? Can it be removed? If it's useful for debug, can it be surrounded by a debug conditional instead?

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
-  //Configure DMA
-  SLCC0 &= ~(SLCMM << SLCM); //clear DMA MODE
-  SLCC0 |= (1 << SLCM); //set DMA MODE to 1
-  SLCRXDC |= SLCBINR | SLCBTNR; //enable INFOR_NO_REPLACE and TOKEN_NO_REPLACE
-  SLCRXDC &= ~(SLCBRXFE | SLCBRXEM | SLCBRXFM); //disable RX_FILL, RX_EOF_MODE and RX_FILL_MODE
+  // Configure DMA
+  SLCC0 &= ~(SLCMM << SLCM); // Clear DMA MODE
+  SLCC0 |= (1 << SLCM); // Set DMA MODE to 1
+  SLCRXDC |= SLCBINR | SLCBTNR; // Enable INFOR_NO_REPLACE and TOKEN_NO_REPLACE
+  SLCRXDC &= ~(/*SLCBRXFE |*/ SLCBRXEM | SLCBRXFM); //disable RX_FILL, RX_EOF_MODE and RX_FILL_MODE
```
devyte: if RX_FILL is no longer disabled, then the comment should be updated

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
-  SLCTXL |= (uint32)&i2s_slc_items[1] << SLCTXLA; //set TX descriptor address. any random desc is OK, we don't use TX but it needs to be valid
   SLCRXL &= ~(SLCRXLAM << SLCRXLA); // clear RX descriptor address
-  SLCRXL |= (uint32)&i2s_slc_items[0] << SLCRXLA; //set RX descriptor address
+  if (!rx) {
+    SLCTXL |= (uint32)&tx->slc_items[1] << SLCTXLA; // Set fake (unused) RX descriptor address
+  } else {
+    SLCTXL |= (uint32)&rx->slc_items[0] << SLCTXLA; // Set real RX address
+  }
+  if (!tx) {
+//    SLCRXL |= (uint32)&rx->slc_items[1] << SLCRXLA; // Set fake (ununsed) TX descriptor address
```
devyte: Is this really not needed? If so, then a comment should be added to explain.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
+static void ICACHE_FLASH_ATTR i2s_slc_end(){
   ETS_SLC_INTR_DISABLE();
   SLCIC = 0xFFFFFFFF;
   SLCIE = 0;
   SLCTXL &= ~(SLCTXLAM << SLCTXLA); // clear TX descriptor address
   SLCRXL &= ~(SLCRXLAM << SLCRXLA); // clear RX descriptor address
 
   for (int x = 0; x<SLC_BUF_CNT; x++) {
-	  free(i2s_slc_buf_pntr[x]);
+    if (tx) {
```
devyte: Not really relevant in this particular case, but:
Rather than a for cycle with two if statements inside, it's better to have two if statements with each a for cycle inside.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -208,6 +333,33 @@ bool ICACHE_FLASH_ATTR i2s_write_lr(int16_t left, int16_t right){
...
+        }
+      }
+    }
+    ETS_SLC_INTR_DISABLE();
+    rx->curr_slc_buf = (uint32_t *)i2s_slc_queue_next_item(rx);
+    ETS_SLC_INTR_ENABLE();
+    rx->curr_slc_buf_pos=0;
+  }
+
+//  *left  = rx->slc_items[0].buf_ptr[rx->curr_slc_buf_pos++];
```
devyte: Can this commented code be removed, or is it some reference?

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -235,60 +387,105 @@ void ICACHE_FLASH_ATTR i2s_set_rate(uint32_t rate){ //Rate in HZ
     }
   }
 
-  //os_printf("Rate %u Div %u Bck %u Frq %u\n", _i2s_sample_rate, i2s_clock_div, i2s_bck_div, I2SBASEFREQ/(i2s_clock_div*i2s_bck_div*2));
-
   //!trans master, !bits mod, rece slave mod, rece msb shift, right first, msb right
-  I2SC &= ~(I2STSM | (I2SBMM << I2SBM) | (I2SBDM << I2SBD) | (I2SCDM << I2SCD));
-  I2SC |= I2SRF | I2SMR | I2SRSM | I2SRMS | ((sbd_div_best) << I2SBD) | ((scd_div_best) << I2SCD);
+  I2SC &= ~(I2STSM | I2SRSM | /*(I2SBMM << I2SBM) |*/ (I2SBDM << I2SBD) | (I2SCDM << I2SCD));
```
devyte: If bits mod is no longer being set, the above comment should be updated. Why is it not being set now?

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -235,60 +387,105 @@ void ICACHE_FLASH_ATTR i2s_set_rate(uint32_t rate){ //Rate in HZ
...
+  I2SC |= I2SRF | I2SMR | I2SRMS | ((sbd_div_best) << I2SBD) | ((scd_div_best) << I2SCD);
```
devyte: What does the change in this line mean? (add comment)

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -235,60 +387,105 @@ void ICACHE_FLASH_ATTR i2s_set_rate(uint32_t rate){ //Rate in HZ
...
 }
 
 void ICACHE_FLASH_ATTR i2s_set_dividers(uint8_t div1, uint8_t div2){
   div1 &= I2SBDM;
   div2 &= I2SCDM;
 
-  I2SC &= ~(I2STSM | (I2SBMM << I2SBM) | (I2SBDM << I2SBD) | (I2SCDM << I2SCD));
+  I2SC &= ~(I2STSM | I2SRSM | (I2SBMM << I2SBM) | (I2SBDM << I2SBD) | (I2SCDM << I2SCD));
```
devyte: what does the change in this line mean? (add comment)

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -30,147 +30,278 @@
...
-  SLCC0 |= (1 << SLCM); //set DMA MODE to 1
-  SLCRXDC |= SLCBINR | SLCBTNR; //enable INFOR_NO_REPLACE and TOKEN_NO_REPLACE
-  SLCRXDC &= ~(SLCBRXFE | SLCBRXEM | SLCBRXFM); //disable RX_FILL, RX_EOF_MODE and RX_FILL_MODE
+  // Configure DMA
+  SLCC0 &= ~(SLCMM << SLCM); // Clear DMA MODE
+  SLCC0 |= (1 << SLCM); // Set DMA MODE to 1
+  SLCRXDC |= SLCBINR | SLCBTNR; // Enable INFOR_NO_REPLACE and TOKEN_NO_REPLACE
+  SLCRXDC &= ~(/*SLCBRXFE |*/ SLCBRXEM | SLCBRXFM); //disable RX_FILL, RX_EOF_MODE and RX_FILL_MODE
 
   //Feed DMA the 1st buffer desc addr
```
devyte: Should this comment be updated to better reflect the new logic below?

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -235,60 +387,105 @@ void ICACHE_FLASH_ATTR i2s_set_rate(uint32_t rate){ //Rate in HZ
...
+  if (tx || rx) {
+    i2s_end(); // Stop and free any ongoing stuff
+  }
+
+  if (enableTx) {
+    tx = (i2s_state_t*)calloc(1, sizeof(*tx));
+    if (!tx) {
+      return; // OOM Error!
+    }
+    pinMode(2, FUNCTION_1);  // I2SO_WS (LRCK)
```
devyte: These pin numbers should probably be #defined like the I2SI_lol ones

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -235,60 +387,105 @@ void ICACHE_FLASH_ATTR i2s_set_rate(uint32_t rate){ //Rate in HZ
...
   I2SC &= ~(I2SRST);
 
-  pinMode(2, INPUT);
-  pinMode(3, INPUT);
-  pinMode(15, INPUT);
-
   i2s_slc_end();
+
+  if (tx) {
+    pinMode(2, INPUT);
```
devyte: #define pin numbers above

devyte:
Overall, looks pretty good, especially if it works!

cores/esp8266/core_esp8266_i2s.c
earlephilhower: You may have hit upon the cause of some random I2S crashes I've seen at idle.  The decorator comes from the original author (maybe cut-n-pasted from the mp3 example at some point?).  I just mirrored what was there in the new functions.  I'll go over and adjust and remove as needed before finishing.

earlephilhower: The actual I2S interrupt handler is defined "ICACHE_**FLASH**ATTR" in master!

cores/esp8266/core_esp8266_i2s.c
earlephilhower: Done.

cores/esp8266/core_esp8266_i2s.c
earlephilhower: done

cores/esp8266/core_esp8266_i2s.c
earlephilhower: done

cores/esp8266/core_esp8266_i2s.c
earlephilhower: done

cores/esp8266/core_esp8266_i2s.c
earlephilhower: done

cores/esp8266/core_esp8266_i2s.c
earlephilhower: Replaced w/memset

cores/esp8266/core_esp8266_i2s.c
earlephilhower: Still trying to figure out *why* it's working the way it is.  No complete register map AFAICT for this monster.  Actually two monsters.  The SLC is partially documented, the I2S is almost a black box...

cores/esp8266/core_esp8266_i2s.c
earlephilhower: Same as above.  Still trying to figure out *why* it's working the way it is.

cores/esp8266/core_esp8266_i2s.c
earlephilhower: My debugging, the DMA buffer handoff is still funky in the code and on the Rx side the receiver can be overwriting the block you're reading from.  It's to be fixed before real submission.

cores/esp8266/core_esp8266_i2s.c
earlephilhower: done

cores/esp8266/core_esp8266_i2s.c
earlephilhower: done

earlephilhower:
Sorry for the ton of updates.  There's no way to consolidate responses into a single one.  The ICACHE_FLASH_ATTR was a good catch, and may be a real bug in the existing ISR.  The rest, I'm still hacking away and will clean things up.  I've built DMA and serial units in Verilog and programmed them on many different systems, but always w/a good block diagram and register reference.  W/O those for the 8266, it's a lot more challenging.

earlephilhower:
Oh, and stereo input seems to be working too.  This is the demo app running on a hacked AIY voice hat:
image

devyte:
@earlephilhower does it make sense to add some unit tests for this?

earlephilhower:
I'll keep that in mind, but because this is running and capturing data via an external pin I'm not sure what can actually be tested.  If there was a loopback mode, that would be ideal.  Send data on one iface, read it back on the other, compare.

devyte:
@earlephilhower this has conflicts now

earlephilhower:
Yup, I figured it would get borked with the latest bugfix.  I'll see if I can untangle it.

earlephilhower:
I've tested and added an example I2S microphone send-over-UDP example.  Runs pretty well if I do say so!

At this stage review and testing by those in the original issue.

libraries/esp8266/examples/I2STransmit/I2STransmit.ino
```diff
@@ -0,0 +1,75 @@
...
+  udp.beginPacket(listener, port);
+  udp.write("I2S Receiver\r\n");
+  udp.endPacket();
+
+}
+
+void loop() {
+  static int cnt = 0;
+  // Each loop will send 100 raw samples (400 bytes)
+  // UDP needs to be < TCP_MSS which can be 500 bytes in LWIP2
```
d-a-v: I come after the battle with saying that I'm not sure MSS applies to UDP.
UDP is only limited by MTU which is always 1460 on ESP8266 (defined and given by PHY).

earlephilhower: Are you sure about that?  I started with ~1024 bytes per packet, but no UDP packets ever made it to the PC with the low-mem LWIP.  At high-mem it did work.  Dropping to 400 bytes or so made things work under both so I left it there.  It's also better for I2S latency, since you want to minimize the amount of buffering, anyway.

d-a-v: Yes, I'm sure that it should work, and I'm now sure I have to check why it does not :)

PR #4550, opened 2018-03-23
Resolve updater size check bug

Adam5Wu:
I accidentially tried to OTA a 2MB+ file, and found the Updater.begin() actually did not report any error.

Turned out there is an arithmetic underflow on an unsigned type.


devyte:
Will begin() report an error with this fix?

Adam5Wu:
Yes, the underflow fix will trigger the proper size error reporting logic in line 98-100
```
 if(updateStartAddress < currentSketchSize) {
...
}
```

devyte:
Thanks for finding this :)

PR #4562, opened 2018-03-25
defined SDA and SCL pins for Olimex MOD-WIFI-ESP8266(-DEV)

mabi:
Changed from generic I2C SDA and SCL pin mapping to specific mapping for board modwifi, Olimex MOD-WIFI-ESP8266(-DEV)

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1795,6 +1795,7 @@ nodemcuv2.menu.UploadSpeed.921600.upload.speed=921600
 ##############################################################
 modwifi.name=Olimex MOD-WIFI-ESP8266(-DEV)
 modwifi.build.board=MOD_WIFI_ESP8266
+modwifi.build.variant=modwifi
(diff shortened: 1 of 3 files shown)


devyte:
@mabi boards.txt is generated by a tool, you need to modify the tool instead.

mabi:
@devyte thank you for the explanation, I will fix shortly

mabi:
@devyte I will shortly squash the commits in one

d-a-v:
@mabi 
You need to regenerate boards.txt with the script and re-add/commit it.
CI also generates it and complains when result differs (yes this is minor diff :)

mabi:
@d-a-v thank you very much, I didn't expect that that line would change place re-running the tool :-)
@devyte is this commit OK for your change request?

devyte:
Yes, but I'll let @d-a-v merge this 😝

PR #4567, opened 2018-03-26
ldscripts: a Makefile to change c++ vtables location

d-a-v:
* using this makefile is sufficient
  no need to add more options to the boards generator as previously proposed
* this pr's ```eagle.app.v6.common.ld``` is regenerated with ```make iram``` 
  to be noted: cpp remove comments
* a warning is added in the generated file

```
$ make
available rules:
  iram: move c++ vtables to iram (= default in git master repository)
  heap: move c++ vtables to heap
$ make heap
c++ vtables moved to heap
$ make iram
c++ vtables moved to iram
```


tools/sdk/ld/Makefile
```diff
@@ -0,0 +1,17 @@
...
+COMMON = eagle.app.v6.common.ld
+WARN = "\/\* warning: AUTOGENERATED from $(COMMON)\*.h \*\/"
+
+all:
+	@echo "available rules:"
+	@echo "  iram: move c++ vtables to iram (= default in git master repository)"
+	@echo "  heap: move c++ vtables to heap"
+
+heap:
+	@(echo "$(WARN)"; $(CPP) -E -DVTABLES_IN_DRAM $(COMMON).h | grep -v '^#') > $(COMMON)
```
earlephilhower: Can you add in a -CC to preserve comments?

tools/sdk/ld/Makefile
```diff
@@ -0,0 +1,17 @@
...
+	@echo "c++ vtables moved to heap"
+
+iram:
+	@(echo "$(WARN)"; $(CPP) -E -DVTABLES_IN_IRAM $(COMMON).h | grep -v '^#') > $(COMMON)
+	@echo "c++ vtables moved to iram"
```
earlephilhower: -CC to preserve comments

Can you add a "flash:" branch, with the appropriate `#if defined(VTABLES_IN_FLASH)` to the makefile and app.ld file as well?

tools/sdk/ld/Makefile
d-a-v: done for both

Issue #2611, opened 2016-10-16
ESP8266WebServer parsing query part of URI  


holgerlembke:
If I send a 

   http://xxxx/index.html?foo=baa

I get an args() result. If I send a 

   http://xxxx/index.html?foo

the foo is lost, no args(), nothing.

I'm not suure how to solve that. Perhaps add a foo-argname with an empty arg.

holgerlembke:
`  int hasSearch = url.indexOf('?');
  if (hasSearch != -1){
    searchStr = url.substring(hasSearch + 1);
	if (searchStr.indexOf('=')==-1) {
		searchStr="query="+searchStr;
	}
    url = url.substring(0, hasSearch);
  }
  _currentUri = url;
`

devyte:
@holgerlembke You are right, but I know that 
http://xxxx/index.html?foo=
works.
A quick google shows that the relevant RFC is not specific on this matter, meaning that it is implementation defined. ASP seems to choke on what you propose, but other parsers accept it.
I think this should be fixed, makes for better compatibility.

holgerlembke:
For concept compatibility it might be an approach to make it like php.
```
$_SERVER["argv"] = Array
(
    [0] => foo
)
```

PR #4573, opened 2018-03-26
WebServer query arguments not containing an equals sign are now accepted…

jasonharper:
as a key with an empty value.  Previous behavior was to completely ignore the argument, yet still include it in the argument count.

(The use of "key=value" arguments is a convention, not an actual requirement in URLs.)

Fixes bug #2611.


reloxx13:
i think its the same as #4151 and  #4236...

PR #4574, opened 2018-03-27
I2S driver fixes for IRQs, protocol, factoring

### LINKED ISSUE #4571: I2S sending data 1 bit too early, results in clipping/garbage at high volumes
### Basic Infos

#### Platform

Platform == all [n/a]

### Settings in IDE

IDE == all [n/a]

### Problem Description

I've got a reasonably popular audio codec library for the ESP8266 that drives I2S DACs for decoded output.  Many people have found that at high levels their DACs generate noise/garbage, but when they drop the volume or add a delay (R/C on the I2SDOUT pin) to the data it works properly.  I thought it was wiring/cap issues, but it's a SW bug in core.

After looking through the ESP32 manual's I2S section the cause is the I2S on the ESP8266 is being configured improperly.

The I2S module, by default, is configured to send PCM format waveforms where the MSB and the Left/Right select are output on the same clock.  I2S, however, requires a 1-bit delay from LRCLK to MSB.

There's a config bit, undocumented on the ESP8266 but listed in ESP32, which should add this delay, I2STMS.  We're setting it on receive, for some reason, even though the code in core can't actually receive...

Will send in a PR.

N/A, but you can try https://github.com/earlephilhower/ESP8266Audio/blob/master/examples/PlayRTTTLToI2SDAC/PlayRTTTLToI2SDAC.ino if you're so inclined.

### Debug Messages

N/A

earlephilhower:
All redundant ICACHE_FLASH_ATTR decorators were removed, we already do
this by default for all routines, anyway,

The actual ISR and its called function moved to to IRAM.  Used to be in flash
due to the decorator, which could lead to crashes.

Fix the I2S on-the-wire protocol by enabling the transmit delay I2STMS because
I2S is supposed to send the MSB one clock after LRCLK toggles.  This was
causing I2S to be twice as loud as intended in the best of cases, and causing
garbage/noise output when the MSB was set since data was effectively shifted.

Refactor the clock divider setting to be done in one function only, as there
is no reason to do the same complicated bit setting in two spots.

Fixes #4571 and a few other things I've been noticing.


earlephilhower:
I've just got a USB 8-line logic analyzer and will try and capture the fixed I2S waveform tonight to verify the I2S protocol fix.  Just comparing the output using the simple playback example in the issue sounds much better (since it is sending out full-swing samples it was clipping something fierce before), but if at all possible I'd like to catch the entire waveform since there's not really documentation on these registers.

earlephilhower:
The change looks good.  Here is the I2S sending 32767 (0x7FFF).  The add'l bit shift is clearly visible and correct:
image

earlephilhower:
For comparison, the wrong, unshifted version that's in the repo now:

image

devyte:
My one comment is similar as I mentioned before: could you add comments for the register definitions/operations? At least for any changes/additions. If you're not sure, then suspicions are ok, as long as you explain that. Suspicions on how/why something works is better than no explanation at all!

earlephilhower:
I've added what I can glean for the I2S block.  Looks very close to the ESP32 so it's not so bad.  SLC (DMA) is not so simple, but this patch isn't touching that.

Issue #4551, opened 2018-03-23
section .text will not fit in region iram1_0_seg


dalbert2:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [ N/A ] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:			[ESP-12F]
- Core Version:			[2.4.1 Release]
- Development Env:		[Arduino IDE and VisualMicro (tried both)]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Generic ESP8266 Module|Nodemcu|WifInfo (tried all 3)]
- Flash Mode:			[qio]
- Flash Size: 		        [4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description
Section full when building with 2.4.1 (works fine with 2.4.0)

Detailed problem description goes here.
Codebase that builds perfectly under 2.4.0 will not build with 2.4.1

Building 1.0.22 release using 2.4.0 core for WifInfo (ESP12 4M/1M):

    Compiling 'Gateway2' for 'WifInfo'
    Program size: 442,404 bytes (used 42% of a 1,044,464 byte maximum) (164.09 secs)
    Minimum Memory Usage: 46580 bytes (57% of a 81920 byte maximum)

Building 1.0.22 release using 2.4.1 core for WifInfo (ESP12 4M/1M):

    Compiling 'Gateway2' for 'WifInfo'
    ld.exe: C:\Users\david\AppData\Local\Temp\VMBuilds\Gateway2\esp8266_wifinfo\Debug\Gateway2.ino.elf section .text will not fit in region iram1_0_seg
     

    Error linking for board WifInfo
    Build failed for project 'Gateway2'
    collect2.exe*: error: ld returned 1 exit status

Building 1.0.22 release using 2.4.1 core for Generic ESP8266 module using Arduino IDE
```
[...]

Issue #3369, opened 2017-06-23
Using PSTR and PROGMEM string with template class causes section type conflict


sticilface:
Using latest git and arduino IDE.  

I use PROGMEM strings and a debug define that places debug messages into progmem, however this does not seem to work when combined with templated classes. 

The error is the same as this issue: https://github.com/esp8266/Arduino/issues/2078

i define my debug as 
```cpp
#define DEBUG(_1, ...) { TEST.printf_P( PSTR(_1) ,##__VA_ARGS__); }
```

and a const char as
```cpp
const char testString[] PROGMEM = "Test String"; 
```

Individually they both compile.  i.e. if the debug define is enabled but i do not use testString. it works, and vice versa.  but if i use both 

```
In file included from /Users/amelvin/Documents/Arduino/templatesketch/templatesketch.ino:2:0:
test.h:7: error: testString causes a section type conflict with __c
 const char testString[] PROGMEM = "Test String"; 
            ^
In file included from /Applications/Arduino.app/Contents/Java/hardware/esp8266com/esp8266/cores/esp8266/Arduino.h:240:0,
                 from /var/folders/nv/8j3v5v0s18v7p4249kv8qbz80000gq/T/arduino_build_216642/sketch/templatesketch.ino.cpp:1:
/Applications/Arduino.app/Contents/Java/hardware/esp8266com/esp8266/cores/esp8266/pgmspace.h:21:51: note: '__c' was declared here
 #define PSTR(s) (__extension__({static const char __c[] PROGMEM = (s); &__c[0];}))
                                                   ^
/var/folders/nv/8j3v5v0s18v7p4249kv8qbz80000gq/T/arduino_build_216642/sketch/test.h:11:41: note: in expansion of macro 'PSTR'
 #define DEBUG(_1, ...) { TEST.printf_P( PSTR(_1) ,##__VA_ARGS__); }
                                         ^
/var/folders/nv/8j3v5v0s18v7p4249kv8qbz80000gq/T/arduino_build_216642/sketch/test.hpp:5:3: note: in expansion of macro 'DEBUG'
   DEBUG("abc");
   ^
exit status 1
testString causes a section type conflict with __c
```

A full working example can be found here
https://gist.github.com/sticilface/9a6410978d7235a469c1e154c1c4c396

igrr:
#2078
#3351

sticilface:
[...]

PR #4582, opened 2018-03-27
Add flash for vtable destination, make it default, and add build menu to control options

earlephilhower:
Add an option for placing vtables in flash to complement the existing iram and heap options, and allow selection using a build menu.  No makefile needed any more,.

Extension of #4551 issue and PR #4567


Adam5Wu:
@earlephilhower  I am not nitpicking on you or this PR. :)

In my recent exploration solving "__c causes a section type conflict with __c" (#3369) I found it is possible to define multiple sections for string literals, so that data with different tags can coexist.

Then I just wonders, whether it is possible to do similar things with vtable data? That is:
1. Put all vtable entries that are involved in ISR in one section, say `.vtable.isr`
2. Put all other vtable entries in another section, say `.vtable.regular`
3. Let the linker put `.vtable.isr` section in IRAM, and put `.vtable.regular` in flash

Just a crazy idea... not sure whether it is doable without hacking into the compiler.

earlephilhower:
Unfortunately the compiler doesn't know which are interrupt called functions. I'm not sure you could tell that with static graph analysis, in fact.  Turing strikes again!

Adam5Wu:
But could human annotation help?

Just like I know I am writing a template class, so instead of PSTR("Blah"), I use PSTR_T("Blah") to put my string literal to a different section to avoid conflict.

I know I am writing a class for ISR, then I add some annotation to the class so compiler can place the vtable of this class to the `.vtable.isr` section.

If that is possible I am happy enough...

earlephilhower:
I don't think there's a way to annotate other than by class names.  You can individually specify in the .ld file to include, say, `_ZL*__irq`, in `.RODATA` (and hence in RAM), but I don't think I've ever seen any linker annotations on anything other than *instances* which do not have their own vtables.

If you find a way, we can figure out how to mangle it in, for sure!  I think in practice, though, class functions in IRQs are pretty rare, and virtual tables in those are even less prevalent.

devyte:
Is there a way to be able to select this from the IDE menu? I'm thinking of the windows users.

earlephilhower:
I think this can be in the menus if we change platform.txt and boards.py to build the app.ld file before every compile, like got GIT version.

In boards.txt there'll be (yet another) popup menu, identical for all devices, that adds a -D switch.

In platform.txt, add a recipe.hooks.prebuild.x to run the command the makefile does now.

I can give it a whirl, but it'll take a few days.

@d-a-v Does this make sense?  You've been hacking the build.py and platform.txt much more than I have...

d-a-v:
Yes this could work this way and in fact would be much better because it'd work under windows without having to install make.

devyte:
@d-a-v that's exactly what I was getting at :)

earlephilhower:
@d-a-v , @devyte  I've just pushed a change which does what was discussed.  A new menu is used instead of a makefile, and no tools are required other than GCC so it should run fine under Windows, too.

earlephilhower:
@d-a-v I'm stuck on how to make build.py get the default vtable flag that was added.  Interactive builds seem to just work and take the first item by default, but the build.py script is barfing about it being undefined:
````
"/home/travis/arduino_ide/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/bin/xtensa-lx106-elf-gcc" -CC -E -P {build.vtable_flags} "/home/travis/arduino_ide/hardware/esp8266com/esp8266/tools/sdk/ld/eagle.app.v6.common.ld.h" -o "/home/travis/arduino_ide/hardware/esp8266com/esp8266/tools/sdk/ld/eagle.app.v6.common.ld"
xtensa-lx106-elf-gcc: error: {build.vtable_flags}: No such file or directory
````

**Comment by d-a-v** [comment]
Setting the default value in platform.txt seems to solve it:
```
build.vtable_flags=-DVTABLES_IN_FLASH
```
It is overridden by the IDE.
(just verified both)

**Comment by earlephilhower** [comment]
Thanks, @d-a-v .  Looks to be building now!

File: platform.txt [hunk 1]
```diff
@@ -76,6 +78,9 @@ recipe.hooks.core.prebuild.2.pattern=bash -c "mkdir -p {build.path}/core && echo
 recipe.hooks.core.prebuild.1.pattern.windows=cmd.exe /c mkdir {build.path}\core & (echo #define ARDUINO_ESP8266_GIT_VER 0x00000000 & echo #define ARDUINO_ESP8266_GIT_DESC win-{version} ) > {build.path}\core\core_version.h
 recipe.hooks.core.prebuild.2.pattern.windows=
 
+## Build the app.ld linker file
+recipe.hooks.linking.prelink.1.pattern="{compiler.path}{compiler.c.cmd}" -CC -E -P {build.vtable_flags} "{runtime.platform.path}/tools/sdk/ld/eagle.app.v6.common.ld.h" -o "{runtime.platform.path}/tools/sdk/ld/eagle.app.v6.common.ld"
```
igrr: `eagle.app.v6.common.ld` is now auto-generated, why is it still tracked in Git? I think it should be removed from the repo and added to .gitignore.

innodron:
I have  just pulled in the most recent commit ([e3c97021](https://github.com/esp8266/Arduino/commit/e3c970210fe74d35049afd0a7dc06759d4f5c437)) yesterday and now I am getting `xtensa-lx106-elf/bin/ld: cannot open linker script file ../ld/eagle.app.v6.common.ld: No such file or directory`error when compiling. Apparently this file is now removed from the repo and expected to be produced during a sketch build.

However, It seems `eagle.app.v6.common.ld` is not being auto-generated for some reason in my case. I  will appreciate to get guidance on how to approach to this issue. Apperently, no similar problems reported yet.

I am using Eclipse with Sloeber plugin. This is an ESP-12F module with the  [attached](https://user-images.githubusercontent.com/1423196/40218727-16350fd6-5a7b-11e8-8cdb-c93c5ccd4b90.png)  properties. Frankly I am not sure which value the new VTables option should take; I just assigned the default.

PR #4601, opened 2018-04-02
save 192 RAM bytes when attachInterrupt unused

phoddie:
If an application doesn't use GPIO interrupts, the `interrupt_handlers` table (192 bytes of RAM) is still allocated as a global and the `interrupt_handler` ISR is included in the binary. By deferring installation of GPIO interrupts from `initPins` to `attachInterrupt` both can be eliminated by the linker.


devyte:
Confirmed. Heap goes up when attachInterrupt is not used in the sketch, remains the same as before when used. 
Merging.

PR #4612, opened 2018-04-05
Update boards.txt

4dsystems:
Need to change flash mode from qio to dio, due to a change in supplier of Flash Memory. Old modules work just the same with no speed difference noticed, new modules will not work without the change to dio however.


earlephilhower:
The file you've modified is script generated and not to be edited.

You'll need to edit `tools/boards.txt.py` to effect your change.

4dsystems:
OK thank you

PR #4613, opened 2018-04-05
Update boards.txt.py

4dsystems:
Need to change flash mode from qio to dio, due to a change in supplier of Flash Memory. Old modules work just the same with no speed difference noticed, new modules will not work without the change to dio however.

### DIFF
--- a/tools/boards.txt.py
+++ b/tools/boards.txt.py
@@ -677,7 +677,7 @@ boards = collections.OrderedDict([
             },
         'macro': [
             'resetmethod_nodemcu',
-            'flashmode_qio',
+            'flashmode_dio',
             'flashfreq_80',
             '512K',
             ],


earlephilhower:
Almost there.  You also need to run the modified script once to generate the new boards.txt, since it's not normally executed as part of the setup (Windows users often don't have Python) and include it in the patch.  Just run it and `git add boards.txt; git commit -a --amend; git push` and you should be all set with this PR.

4dsystems:
Sorry I have no idea how to do that. I have just edited the file on github site, we dont have the tools installed on the machine. Are you able to assist please?

earlephilhower:
Unfortunately I can't edit your branch where this is coming from, but you should be able to do it from the web for this small of a patch.  For larger changes it's definitely not recommended!

Just go to your GitHub repo, select the `patch-2` branch,  edit the 1 spot this will change in boards from txt, save the edits to the existing branch, and they'll appear as a new commit in this PR automatically.

4dsystems:
I forked the clone into a repo on our account, and went to the patch-2 branch, and the change was already present there...

earlephilhower:
Yes, the file is there but it doesn't have the QIO->dio change in it.  You can run the `boards.txt.py` to generate it, `git commit -a --amend; git push` it, and you'll be set.  Right now your patch (see top link of the PR) only includes the .py change, not the output...

4dsystems:
how do you call 'git commit -a --amend; git push' etc from the website? Or do I have to have git installed on my computer or something? I am a hardware guy, and this I am really struggling with.

d-a-v:
@4dsystems if the change if only qio to dio, we can make it for you

PR #4616, opened 2018-04-07
board gen4iod: change flash mode from qio to dio

d-a-v:
on behalf of @4dsystems #4613:
Need to change flash mode from qio to dio, due to a change in supplier of
Flash Memory.  Old modules work just the same with no speed difference
noticed, new modules will not work without the change to dio however.


4dsystems:
Thank you kindly

PR #4619, opened 2018-04-07
pgm_read_*: cast address to const void*

### LINKED ISSUE #4572: Recent pgmspace.h broke backward compability for pgm_read_xxx() macros
There were some updates to pgmspace.h in mid 2017 that changed the way pgm_read_byte() and pgm_read_word() worked and were implemented.
When the dust settled, the final implementation breaks some use cases as it is no longer fully backward compatible with the AVR version.
The AVR version did not require the address argument to be any particular type and neither did the esp version until these recent updates.
This issue can be traced back to the commit on April 17, 2017
https://github.com/esp8266/Arduino/commit/0b672668bf636ef650b8b700a0af14677c3fafe5#diff-2b6ff2e1583a8a247dc1070d846b5160
That is the first version of code where the addr argument must be a specific type.
Then after that update, future updates built on that and ended up with the macros pgm_read_byte() and pgm_read_word() simply calling functions which expected an argument type.
This can be seen in the june 5 commit:
https://github.com/esp8266/Arduino/commit/c52b0da11cfd9666e45d576d3fdd2533415c4ab6#diff-2b6ff2e1583a8a247dc1070d846b5160
with these two AVR compatibility macros:
```
// Make sure, that libraries checking existence of this macro are not failing
#define pgm_read_byte(addr) pgm_read_byte_inlined(addr)
#define pgm_read_word(addr) pgm_read_word_inlined(addr)
```
While those macros solved a big concern in the discussion of having the  needed  macros to test for PROGMEM support at compile time, they break AVR compatibility by not casting the addr argument.
This means that any code that is not using the argument type expected by pgm_read_byte_inlined() or pgm_read_word_inlined() (which is a void *) will fail from a type mismatch.
There is quite a bit of code out there that was not using void * for this argument and some of that code was previously working on esp until these updates were done.
A simple solution is add a cast.
```
// Make sure, that libraries checking existence of this macro are not failing
// and support any type for addr argument for backward compatibility
#define pgm_read_byte(addr) pgm_read_byte_inlined((const void*)addr)
#define pgm_read_word(addr) pgm_read_word_inlined((const void*)addr)
```

d-a-v:
fix #4572

### DIFF
--- a/cores/esp8266/pgmspace.h
+++ b/cores/esp8266/pgmspace.h
@@ -112,8 +112,8 @@ static inline uint16_t pgm_read_word_inlined(const void* addr) {
 }
 
 // Make sure, that libraries checking existence of this macro are not failing
-#define pgm_read_byte(addr) pgm_read_byte_inlined(addr)
-#define pgm_read_word(addr) pgm_read_word_inlined(addr)
+#define pgm_read_byte(addr) pgm_read_byte_inlined((const void*)(addr))
+#define pgm_read_word(addr) pgm_read_word_inlined((const void*)(addr))
 
 #else //__ets__
 #define pgm_read_byte(addr)     (*reinterpret_cast<const uint8_t*>(addr))


cores/esp8266/pgmspace.h
```diff
@@ -112,8 +112,13 @@ static inline uint16_t pgm_read_word_inlined(const void* addr) {
 }
 
 // Make sure, that libraries checking existence of this macro are not failing
+#ifdef __PROG_TYPES_COMPAT__
```
igrr: Names starting with double underscore are usually reserved for the compiler and standard library, can we use something like ESP8266_PROGMEM_TYPES_COMPAT?

igrr: (deleted an earlier comment after seeing the reference to `__PROG_TYPES_COMPAT__` in the original issue).

Looking at https://www.nongnu.org/avr-libc/user-manual/group__avr__pgmspace.html, `__PROG_TYPES_COMPAT__` is used for different purpose (to enable some typedefs) than used here. Therefore suggest using a different name.

bperrybap: Since the esp8266 pgmspace.h always creates all the deprecated AVR pgmspace types that this define enables in the AVR version, this name seems ok for this usage to me. i.e. from a user perspective setting this define ensures you get the deprecated progmem types as well as the old behavior that was present when those types existed.
i.e. it provides full AVR compatibility including the old types.

I guess technically to be fully compatible with the AVR version the deprecated AVR pgmspace types should also be removed in the esp8266 version if the define is not set but currently the deprecated types always exist in the esp8266 version - which is a difference but is likely to be pretty harmless.

cores/esp8266/pgmspace.h
```diff
@@ -39,6 +41,8 @@ typedef uint16_t prog_uint16_t;
 typedef int32_t prog_int32_t;
 typedef uint32_t prog_uint32_t;
 
+#endif // defined(__PROG_TYPES_COMPAT__)
```
d-a-v: @igrr now that the should-have-been-deprecated typedefs are using this avr macro, is it OK to keep it for pgm_read_* type cast too ?

igrr: Yes, should be okay.

cores/esp8266/pgmspace.h
bperrybap: I've been following along. Everything seems reasonable for very close compatibility.
It may inadvertently trip up some existing esp8266 users that were depending on the deprecated progmem types, but overall it seems great.
The only last  potential remaining item for complete AVR pgmspace.h emulation is the location of the header file itself since with avr-gcc is pgmspace.h is in a subdirectory named "avr" since some Arduino code include 
```
#include <avr/pgmspace.h>
```
On my test/development system I have 80 Arduino libraries installed.
about 20 of those 80 include <avr/pgmspace.h>
Here is grep search for them on my system: 
(several may have other esp8266 issues, but still there is significant percentage of libraries using this location)
```
ffft/ffft.h:#include <avr/pgmspace.h>
glcd/glcd.h:#include <avr/pgmspace.h>
I2C_graphical_LCD_display/I2C_graphical_LCD_display.h:#include <avr/pgmspace.h>
LCD12864/LCD12864.h:#include <avr/pgmspace.h>
Newliquidcrystal_1.3.5/LCD.h:#include <avr/pgmspace.h>
PCF8574_I2C_LCD/PCF8574_I2C_LCD.h: #include <avr/pgmspace.h>
PCF8574_I2C_LCD/PCF8574_I2C_LCD.h: #include <pgmspace.h>
RGB-matrix-Panel/gamma.h:#include <avr/pgmspace.h>
SdFat/SdBaseFile.h:#include <avr/pgmspace.h>
TTS/english.h:#include <avr/pgmspace.h>
WaveHC/WaveUtil.h:#include <avr/pgmspace.h>
Adafruit_GFX_Library/Adafruit_GFX.cpp:  #include <avr/pgmspace.h>
Adafruit_GFX_Library/Adafruit_GFX.cpp:  #include <pgmspace.h>
Adafruit_GFX.sav/Adafruit_GFX.cpp: #include <avr/pgmspace.h>
Adafruit-MCP23017/Adafruit_MCP23017.cpp:#include <avr/pgmspace.h>
Adafruit_SSD1306/Adafruit_SSD1306.cpp:#include <avr/pgmspace.h>
DHT22/DHT22.cpp:#include <avr/pgmspace.h>
glcd/glcd.cpp:#include <avr/pgmspace.h>
glcd/gText.cpp:#include <avr/pgmspace.h>
LCD12864/LCD12864.cpp:#include <avr/pgmspace.h>
MD_MAX72xx/MD_MAX72xx_font.cpp:#include <avr/pgmspace.h>
RGB-matrix-Panel/RGBmatrixPanel.cpp:    "#include <avr/pgmspace.h>\n\n"
RTClib/RTClib.cpp: #include <avr/pgmspace.h>
Time/DateStrings.cpp:#include <avr/pgmspace.h>
```

This one may be a bit more tricky since I've seen that a few library authors have already gone in and modified their code with ifdefs for different location on esp8266.
With newer Arduino IDEs, pgmspace.h does not have to be included at all but, it still has to be included to be support older Arduino IDEs.
What esp8266 could do is add a new header file avr/pgmspace.h that simply includes the real <pgmspace.h> header file so it works either way.

d-a-v: well, you already have included in the merged PR :-)

bperrybap: Oops. I see that now. Sorry I missed that. I only saw the diffs in the header file itself.
Very complete - very nice.

PR #4622, opened 2018-04-08
Move continuation stack from .bss onto sys stack

igrr:
See https://github.com/esp8266/Arduino/pull/4553.

Allocate continuation stack on the system stack. This is done by overriding the application entry point. Also includes some cleanup in core_esp8266_main.cpp: group extern declarations and global/static variables, move `atexit` into libc_replacements.c.

### DIFF
--- a/cores/esp8266/cont_util.c
+++ b/cores/esp8266/cont_util.c
@@ -20,12 +20,15 @@
 
 #include "cont.h"
 #include <stddef.h>
+#include <string.h>
 #include "ets_sys.h"
 
 
 #define CONT_STACKGUARD 0xfeefeffe
 
 void cont_init(cont_t* cont) {
+    memset(cont, 0, sizeof(cont_t));
+    
     cont->stack_guard1 = CONT_STACKGUARD;
     cont->stack_guard2 = CONT_STACKGUARD;
     cont->stack_end = cont->stack + (sizeof(cont->stack) / 4);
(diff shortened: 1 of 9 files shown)


igrr:
Since the amount of available stack space on the system stack is reduced, it would be great if some heavy users of @me-no-dev's *Async libraries could test this.

devyte:
@igrr I don't know if I qualify as a "heavy user" of the Async libs, but I do use AsyncUDP, AsyncTCP, the AsyncWebserver, and an AsyncDNSServer that I wrote myself based on AsyncUDP. My app has been working fine with this for several days.
I suggest merging now to get better exposure. The next step is to implement the stack overflow canary with the hw breakpoint, plus the switching back and forth on context switch. If anything goes wrong with this, the canary will squawk.

Adam5Wu:
I think I may have found a negative impact:
The WPS function is completely broken now.
See: https://github.com/espressif/ESP8266_NONOS_SDK/issues/130

Took me 5 hours and 2 hotspots, before I realized this commit might be related.
And then I test verified: revert to the commit right before, WPS works; after this commit, WPS -> wdt reset.

It kind of make sense, according to the log, some pubkey crypto is involved during WPS, and it probably needed some big chunk of sys stack, which this commit took away...

Although I haven't tested, but it is possible other SDK functions are also broken by this commit, if they need heavy crypto: airkiss? smartconfig? mesh?

PR #4623, opened 2018-04-08
Update SoftwareSerial version to 3.4.1

plerup:


devyte:
@plerup thank you for this!
I looked at the changes in the remote repo between 3.3.1 and 3.4.1, and it looks good to me. I'm merging this now to get it exposed to the users.

PR #4626, opened 2018-04-09
fix connection reset by peer case

d-a-v:
... where pcb is set to null in ClientContext::_error but not reported to WiFiClient
also rename functions *_sent to *_acked (:sent to :ack in debug)


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -536,7 +536,7 @@ class ClientContext
         tcp_sent(_pcb, NULL);
         tcp_recv(_pcb, NULL);
         tcp_err(_pcb, NULL);
-        _pcb = NULL;
+        _pcb = 0;
```
devyte: _pcb is a pointer, so assignment was current with NULL.

d-a-v:
```
$ grep '_pcb =' ClientContext.h
            _pcb = 0;
            _pcb = 0;
        _pcb = NULL;
```
It is for coherency. 0 is c++ way for NULL (I prefer NULL myself). Should I change them all ?

devyte: I didn't realize, but yes please.

devyte: Oh, yes, for some reason I thought of ClientContext as C code. Nullptr please!

d-a-v:
Changes done but PR garbled, I will fix it clean.

PR #4627, opened 2018-04-10
Mention ESP.deepSleepMax() in the list of ESP-specific functions

marcelstoer:
Lack of documentation became obvious when I worked on https://thingpulse.com/2018/04/10/max-deep-sleep-for-esp8266/


devyte:
Hi @marcelstoer , thanks for this. I should have documented it when I put the method in.

PR #4628, opened 2018-04-10
device tests: add support for environment variables

igrr:
Previously device tests included information such as access point SSID/password at compile time. This made it difficult to compile test binaries once and then send them to multiple test runners for execution.

This change adds a command to the test library to set environment variable on the target device: “setenv key value”. C library setenv/getenv facility is used to store variables.

Test runner, tests, and makefile are updated to use this functionality.

Part of https://github.com/esp8266/Arduino/projects/8.


igrr:
I've added some facilities to pass environment variables back and forth between test runner and the device at run time. This should make some kinds of tests easier to write.

Issue #2819, opened 2016-12-31
printf_P doesn't expand varargs correctly


zzeekk:
#### Hardware
Hardware:		Sparkfun ESP8266 
Core Version:      	master

### Description

printf_P doesn't expand varargs correctly:

--- code ---
void setup() {
Serial.begin(115200);
Serial.setDebugOutput(true);
delay(10);
Serial.println("\nSandbox\n");
**printf("String %s\n", "Test");**
**printf_P(PSTR("String %s\n"), "Test");**
}

void loop() {
  delay(1000);
}

--- output ---
Sandbox
**String Test**
**String** 

Problem lies on line 237 of pgmspace.cpp (https://github.com/esp8266/Arduino/blob/master/cores/esp8266/pgmspace.cpp#L237)

**vprintf should be used instead of printf.**

Issue #2633, opened 2016-10-24
strncpy_P doesn't zero out memory


chschu:
Core Version: 2.3.0

According to the C standard, `strncpy(dst, src, count)` always writes exactly `count` bytes to `dst`, padding with zeroes after `src` has been copied.

The current implementation of `strncpy_P` doesn't do that, it just stops after copying `src` and the terminating zero. Code relying on the trailing zeroes works well with `strncpy`, but it doesn't work with `strncpy_P`.

PR #4631, opened 2018-04-11
tests: add a couple of pgmspace tests

igrr:
Regression tests for #2819 and #2633.


igrr:
I forgot to `git add` the file when fixing those two issues, and the file has been sitting in my local directory for a long time. Shouldn't hurt to check it in.

PR #4636, opened 2018-04-11
Added channel, ssid scan

dtworth:
Overloaded scanNetworks so scan can occur on a single channel and/or for a particular ssid.

I needed this so I could speed up the scan time in order to track the location of a mobile device using the measured signal strength.  A three-second scan was too long.  With this modification, the scan time is reduced to ~100 mS if the channel number is defined.

If folks find this useful, I might also fool around with the scan time parameter.

I hope that you will consider including this modification in the next release version.


libraries/ESP8266WiFi/src/ESP8266WiFiScan.cpp
```diff
@@ -102,6 +102,53 @@ int8_t ESP8266WiFiScanClass::scanNetworks(bool async, bool show_hidden) {
...
+    int status = wifi_station_get_connect_status();
+    if(status != STATION_GOT_IP && status != STATION_IDLE) {
+        wifi_station_disconnect();
+    }
+
+    scanDelete();
+
+    struct scan_config config;
+    memset(&config, 0, sizeof(config));
+    config.ssid = ssid;
```
devyte: The only difference between the original method and your overloaded one are these two lines:
 config.ssid = ssid;
 config.channel = channel;

where ssid and channel are new arguments.

Instead of duplicating the entire method, please just add the two arguments to the original method, and in the .h put set default values to keep the current behavior (i.e.: channel = 0 and ssid = nullptr, because of the memset).

happytm:
Is it now possible to scan only specific ssid. If so how to do it?

Thanks.

PR #4677, opened 2018-04-25
lwip2: fix static address management

d-a-v:


devyte:
@igrr that's because the code is really in the other repo.
I've been thinking that the glue code should be moved here, and the other repo should be "pure" lwip. I have the following reasons:
1. The glue code actually needs some maintenance, and it is easier to look at and review here
2. The glue code is really ESP-specific, sonI think it really belongs here
3. It would be more in line with e.g. spiffs
4. The other repo wouldd be just standalone lwip, configured for the ESP, but without ESP specifics

d-a-v:
@igrr there is no lwIP api to check whether dhcp is running (or I could not find it), so yes I currently don't know how I could make a device test with this. 
However I think it is worth adding the interactive sketch in libraries/esp8266/examples unless there would be a better place for it. I actually used it with some specific options to switch back and forth with static address. This test could become more consistent and evolve if it was in the core repo.
In the meantime I will update lwip2 to remove old comments and #if 0 dead code. There will also be the small packet capture changes which are harmless in normal conditions.

@devyte thanks to pinpoint. The "lwip2" repo is for esp8266-nonos-sdk only. lwIP -the real upstream library-  is a sub-module (hence a sub-sub-module of the arduino core) and is not held in the "lwip2" repo. "lwip2" is *only* the glue between esp8266-nonos-sdk and upstream lwIP hosted on savannah.
I currently host the glue, but it could be changed.

PR #4687, opened 2018-04-26
general core health status: KISS api

d-a-v:
This is a proposal, this is wip, a request for comment, the missing meditating guru.
We don't have exception but sometimes we would like to warn when the worst is known and about to happen.
This simple API can provide some valuable information to the user, so action can take place at his discretion.


devyte:
As discussed internally, let's pursue #4694 first. If we hit a wall there, we can revisit this.

devyte:
We now have C++ exception with #5376 . Closing.

PR #4691, opened 2018-04-27
Fixed SSID length limit to 32 bytes

jgesser:
Wifi was rejecting connections for SSID with 32 bytes.
It was also returning trash bytes at the end of SSIDs when scanning ...

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -104,7 +104,7 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
         return WL_CONNECT_FAILED;
     }
 
-    if(!ssid || *ssid == 0x00 || strlen(ssid) > 31) {
+    if(!ssid || *ssid == 0x00 || strlen(ssid) > 32) {
         // fail SSID too long or missing!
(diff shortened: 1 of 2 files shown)


devyte:
@jgesser Thanks! However, this PR is incomplete. There are other places where ssid is used, which need to be modified.
Examples off the top of my head:
- [ssid compare](https://github.com/esp8266/Arduino/blob/41a64707f149d01ace37c903f448d5e3f1cee5d8/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L62)
- [ssid copy](https://github.com/esp8266/Arduino/blob/41a64707f149d01ace37c903f448d5e3f1cee5d8/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L118)
- The [WiFiMulti ](https://github.com/esp8266/Arduino/blob/41a64707f149d01ace37c903f448d5e3f1cee5d8/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp#L186)implementation

There may be other places. A thorough search of code that sets/gets/uses the ssid must be done.

Also, there is the SoftAP ssid to consider in ESP8266WiFiAP.cpp. 

And having said that, I just realized that 64-char passphrase is not supported for SoftAP. Not sure if it makes sense, though.

devyte:
Closing in favor of #5411 .

Issue #4520, opened 2018-03-15
different sketch MD5 if built from different locations


marcobrianza:
I am using ESP.getSketchMD5() to manage software releases.

I have noticed that if I use the same Arduino IDE, but in a different place in the filesystem the sketch MD5 is different.

I have used a portable version of the IDE to be sure that the same files are used.
For example loading a sketch that prints MD5 shows this behaviour.

Is this a known thing?

```
void setup() {
  Serial.begin(115200);
  Serial.println();
  Serial.println( ESP.getSketchMD5());
}
void loop() {
}

```

marcobrianza:
using the strings command on a compiled binary shows some occurrence of the source path, it would be nice if this could be avoided.

d-a-v:
If it's about ```__FILE__``` gcc macro related, then I'd like to have to solution too.

igrr:
There's a GCC patch from Debian which can be used for this purpose:

https://lists.alioth.debian.org/pipermail/reproducible-builds/Week-of-Mon-20170814/009124.html

Eszartek:
The use of ESP.getSketchMD5() is interesting to me as well, so after running strings on my binaries I saw the long absolute paths being stored. Since I compile my projects on Linux using makeEspArduino.mk makefile instead of the Arduino IDE, it was trivial to edit my config.mk from:

`ESP_ROOT=/home/user/Arduino/hardware/esp8266com/esp8266`
`LIBS=~/Arduino/hardware/esp8266com/esp8266/libraries/ ~/Arduino/libraries/`
to be:

`ESP_ROOT=../hardware/esp8266com/esp8266`
`LIBS=../hardware/esp8266com/esp8266/libraries/ ../libraries/ `

Then, I compiled my current project on two different machines where the working directory is different, and the resulting md5sums matched. That's cool!

However, the more interesting result for me was that my binary size was now smaller:
from:
`Ram:        61220 bytes`
`Flash:  375172 bytes`
to:
`Ram:        61096 bytes`
`Flash:  374200 bytes`

I'm always working to save a byte here and there, so this was a nice gift for very little effort.

d-a-v:
@igrr thanks for the interesting reading !
While it should be feasible to backport [it](https://gcc.gnu.org/ml/gcc-patches/2016-11/msg00182.html) to (xtensa-)gcc-4.8.2 (at which cost?), @Eszartek's solution with makeEspArduino.mk seems very good at little cost.

[...]

PR #4694, opened 2018-04-27
exception playground

d-a-v:
Out of curiosity, I tried to enable exceptions.
With these changes and a random sketch, flash size is increased by 40KB, and a hundred bytes in ram.
It compiles and run until a `throw` test supposed to be `catch(...)`ed, which ends up in an `abort()` call (nothing is executed after the `throw`).
I'm not sure whether I'm playing with fire but I'd like to understand the (un)feasibility of the thing.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -43,7 +43,7 @@ compiler.c.elf.cmd=xtensa-lx106-elf-gcc
 compiler.c.elf.libs=-lhal -lphy -lpp -lnet80211 {build.lwip_lib} -lwpa -lcrypto -lmain -lwps -laxtls -lespnow -lsmartconfig -lairkiss -lwpa2 -lstdc++ -lm -lc -lgcc
 
 compiler.cpp.cmd=xtensa-lx106-elf-g++
-compiler.cpp.flags=-c {compiler.warning_flags} -Os -g -mlongcalls -mtext-section-literals -fno-exceptions -fno-rtti -falign-functions=4 -std=c++11 -MMD -ffunction-sections -fdata-sections
+compiler.cpp.flags=-c {compiler.warning_flags} -Os -g -mlongcalls -mtext-section-literals -fno-rtti -falign-functions=4 -std=c++11 -MMD -ffunction-sections -fdata-sections
 
 compiler.as.cmd=xtensa-lx106-elf-as
 
--- a/tools/sdk/ld/eagle.app.v6.common.ld.h
+++ b/tools/sdk/ld/eagle.app.v6.common.ld.h
@@ -102,6 +102,11 @@ SECTIONS
 #ifdef VTABLES_IN_FLASH
     *(.rodata._ZTV*) /* C++ vtables */
 #endif
+
+    *(.*except*)
+    *libgcc.a:unwind-dw2.o(.literal .text .rodata .literal.* .text.* .rodata.*)
+    *libgcc.a:unwind-dw2-fde.o(.literal .text .rodata .literal.* .text.* .rodata.*)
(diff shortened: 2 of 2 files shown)


devyte:
CI failed due to linker complaining

d-a-v:
This is needed somewhere (for Travis-CI too):
```
void __gnu_cxx::__verbose_terminate_handler ()
{
  //os_printf("exception\r\n");
  //i_was_here = 42;
  //panic();
}
```

igrr:
There are a few more things missing for exception support...

1) libstdc++ needs to be built without `-fno-exceptions` flag (this seems to be the case already)
2) need to [call `__register_frame_info`](https://github.com/espressif/esp-idf/blob/051d8d69000839985e9b6a281e89ccfe6c16080e/components/esp32/cpu_start.c#L421), before executing constructors for global objects
3) above call will result in libstdc++ allocating [16kB](https://github.com/jcmvbkbc/gcc-xtensa/blob/d230f4322ce0df8cdeb624d99310e4af9e33060c/libstdc%2B%2B-v3/libsupc%2B%2B/eh_alloc.cc#L63-L64) from heap for emergency exception pool. This probably makes sense on desktop but not much sense on an embedded platform. We probably need to reduce this size and/or make it adjustable, like this is done in [this patch](https://github.com/espressif/crosstool-NG/blob/xtensa-1.22.x/local-patches/gcc/5.2.0/0012-xtensa-Configurable-exception-emergency-pool-size.patch). Then a weak function [can be overridden](https://github.com/espressif/esp-idf/blob/051d8d69000839985e9b6a281e89ccfe6c16080e/components/esp32/cpu_start.c#L411-L414) to adjust emergency pool size from the application.
4) [a few more changes to the linker script](https://github.com/espressif/esp-idf/commit/d4e950d31187e83f96e7a19f7d0c54ee9a14ccb6#diff-d7a5610415927bcc3e2c66fe5f0680eb) to catch sections produced with -ffunction-sections -fdata-sections
5) either override `__verbose_terminate_handler` or, better, build gcc with `--disable-libstdcxx-verbose` to disable verbose terminate handler
6) make the exception support configurable (yet another menu option, boards.txt.local flag) or just bite the bullet and enable it for everyone.

devyte:
For me this is an area that has a sign that says "here be dragons", but it sounds like it would be more or less straightforward for someone with the right knowledge.
@earlephilhower any chance you could take this further?

d-a-v:
In addition to exceptions above, and if we come to try with gcc changes, let's not forget #4520:
- gcc patches for removing full path in `__FILE__`
- LTO support
- `#define JUMP_TABLES_IN_TEXT 1` to save more heap

@igrr one can always start from scratch (with jcmvbkbc/gcc-xtensa or? pfalcon/esp-open-sdk).
It is not clear to me if you have a [particular recipe](https://github.com/esp8266/Arduino/issues/4520#issuecomment-375849529) for esp8266/Arduino.
In the positive case, would it be wise to put this recipe somewhere in this (or another) repo, as a script, link, submodule, also with your [newlib-xtensa](https://github.com/igrr/newlib-xtensa), and with a set of patches to try / optionally use / include under a subdirectory ?

d-a-v:
@earlephilhower You also have the beginning of [an answer](https://github.com/esp8266/Arduino/issues/4520#issuecomment-375844928).

What I'm pushing for is a common place to start experiments with. That way interesting findings, even if not rightly used, can be stored and visible for a later review when someone have time to play with.

devyte:
Closing in favor of #5376 .

PR #4718, opened 2018-05-10
Update of mesh network library.

aerlon:
The mesh network library was not working well, so I took a look at it and managed to improve it to a relatively well working state.

As noted in the new library readme: The library has been tested and works with Arduino core for ESP8266 version 2.3.0 (with default lwIP) and 2.4.1 (with lwIP 1.4). Random disconnects and crashes seem to happen with the most recent master branch, so better stick to the release builds for now. Also, nodes seem to be unable to connect to more than one other node when using lwIP 2.0.


d-a-v:
About lwip2 not working properly, static IP bug was recently fixed in lwip2, after core-2.4.1 release.
Can you print out the full version string `Esp.getFullVersion()` (also visible in debug mode) ?

aerlon:
The output from `ESP.getFullVersion()` is: SDK:2.2.1(cfd48f3)/Core:win-2.5.0-dev/lwIP:1.4.0rc2

I noticed that a lwIP2 fix was released a few days ago (#4677) but I get the same result both with and without that fix.

In fact, I think the problem is unrelated to static IP. Even when I set the network to only use DHCP the problems persist. 

The main issue when using the current master branch seems to be that just after transmitting something to an AP with `client.print()`, the following condition sometimes turns true: `if (WiFi.status() == WL_DISCONNECTED || !curr_client.connected()) `. More precisely, `!curr_client.connected()` turns true even though `WiFi.status() == WL_DISCONNECTED` is false. The AP still receives the station's transmission.

Indeed, removing the `!curr_client.connected()` check from the condition above makes the mesh code work well even with the most recent master branch (and lwIP 1.4), but it seems to me that if a station can both send and receive transmissions from an AP then `!curr_client.connected()` should be false. Especially since this is the case in previous releases. Is this a topic for a bug report?

aerlon:
I also did a trial run with lwIP2 and the current master branch with the `!curr_client.connected()` check removed.

Though transmissions work well initially as with lwIP1.4, whenever a connection to a second AP is established (meaning static IP is disabled and DHCP is activated (via `WiFi.config(0u,0u,0u)`)) the node becomes unable to connect to any AP (connection attempts take more than 10 s).

Output from `ESP.getFullVersion()`: SDK:2.2.1(cfd48f3)/Core:win-2.5.0-dev/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.1-7-g2b827f8)

d-a-v:
Well, that's a lot of informations. Can we accurately summarize ?

* `(WiFiClient:) curr_client.connected()`  is not true but nonetheless connected and working when a new client has come (via `WiFiServer`?), in some mode (STA, AP, AP+STA?), with both lwip1.4 and 2.

* lwip2 does not behave correctly (at least not like lwip1.4) when connecting to a dhcp server (provided by another ESP in AP mode? locally in AP+STA mode? both ?) using `WiFi.config(0u,0u,0u)` (I'm not sure this is the recommended way to get to dhcp-client mode)

You'll understand that we need to reproduce bugs separately. As it is more easy to try and fix bugs with MCVEs, it'd be great if you could open issues with them.

aerlon:
Heh, yeah. And still I tried to be brief.

Your summary is correct. Answers: 
Via WiFiServer? Yes. 
Mode is STA for station, AP+STA for AP.

Server is provided by another ESP in AP+STA mode.
Checking [the source](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp) at line 223-230 `WiFi.config(0u,0u,0u)` seemed to be the most supported method. I'd be happy to hear of a more official one, though, because I found no documentation about it.

I'll see if I can reduce the code to simpler test cases for you and open issues with them. ETA a day or two.

d-a-v:
I tried calling `WiFi.config(0u,0u,0u)` when IP is static with lwIP-v2 and it gets back to dhcp-client mode.

aerlon:
@d-a-v Issues #4728 and #4727 have been filed regarding the problems discussed above.

While working on #4727 I noticed that with the latest master branch and lwIP2 it is possible for a station to switch between WiFi networks even while using static IP, provided `WiFi.mode(WIFI_OFF)` is called before the switch. This is not possible with any other lwIP or ESP8266 Arduino core. 

Previously I did not find a way to have a station switch networks while maintaining static IP (connections to all but the first network would always fail, thus the use of `WiFi.config(0u,0u,0u)` discussed above to activate DHCP), but it would seem possible after all. 

Is there an officially supported way to have a station switch WiFi network while maintaining static IP that I've missed?

d-a-v:
Thanks for the two mcve!
I will look into and try them as soon as I can make a 3-esp setup to play with.

> Is there an officially supported way to have a station switch WiFi network while maintaining static IP that I've missed?

I personally don't know. I think it is always safe and recommended to reset IP addresses (static or dhcp) after any network change (ethernet plugout-plugin, access-point change).

aerlon:
You're welcome!

OK.

devyte:
@aerlon is there anything pending for this?

aerlon:
@devyte Yes. I've been working on a large update during the last few weeks, trying to make the library more flexible. It is pretty much done now, so the plan is to push the update to this PR this week. The update will also modify the library documentation so that it is up to date with the recent bug fixes for lwIP2.

devyte:
Great! In that case I'll wait.

aerlon:
The PR has been updated. Next I’ve been planning on providing a better reference mesh implementation than in the current example. At the moment the library is more a framework to build your own mesh network, rather than a working mesh network. I thus intend to create an easy to use mesh network architecture based on the library and provide it as an example. I’m unsure when this will be completed, though. It is perhaps something for another pull request, so that this PR can be considered ready for release.

devyte:
@aerlon could you please look me up in gitter? there is something tangential to this PR that I'd like to discuss.

aerlon:
@devyte Certainly. I should be able to join later today or tomorrow.

aerlon:
@devyte As per our discussions I have added a compatibility layer to the new library code. The old mesh library API is now deprecated but valid until core version 2.5.0.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
 #include <ESP8266WiFi.h>
 #include <ESP8266WiFiMesh.h>
 
+String mesh_name = "Mesh_Node";
+
 unsigned int request_i = 0;
 unsigned int response_i = 0;
 
-String manageRequest(String request);
+String manageRequest(String request, ESP8266WiFiMesh *mesh_instance);
```
devyte: The request argument is declared for passing by value, which needlessly duplicated mem. It should probably be declared as reference, or better a const reference.
While strictly speaking it's ok to declare the mesh_instance as a pointer, it's better to declare as reference to make it explicit that the arg can never be a nullptr.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
 #include <ESP8266WiFi.h>
 #include <ESP8266WiFiMesh.h>
 
+String mesh_name = "Mesh_Node";
```
devyte: This should probably be declared as String meshName(F("meshNode"));
Also, please use camelCase for naming convention.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
+
 unsigned int request_i = 0;
 unsigned int response_i = 0;
 
-String manageRequest(String request);
+String manageRequest(String request, ESP8266WiFiMesh *mesh_instance);
+transmission_status_t manageResponse(String response, ESP8266WiFiMesh *mesh_instance);
+void networkFilter(int number_of_networks, ESP8266WiFiMesh *mesh_instance);
```
devyte: mesh_instanc declare as reference

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
-   Callback for when other nodes send you data
+   Callback for when other nodes send you a request
 
-   @request The string received from another node in the mesh
+   @param request The request string received from another node in the mesh
+   @param mesh_instance The "this" pointer of the ESP8266WiFiMesh instance that called the function.
    @returns The string to send back to the other node
 */
-String manageRequest(String request) {
+String manageRequest(String request, ESP8266WiFiMesh *mesh_instance) {
```
devyte: request => ref, mesh_instance => ref

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
   /* Print out received message */
-  Serial.print("received: ");
+  Serial.print("Request received: ");
   Serial.println(request);
 
   /* return a string to send back */
   char response[60];
-  sprintf(response, "Hello world response #%d from Mesh_Node%d.", response_i++, ESP.getChipId());
+  sprintf(response, "Hello world response #%d from %s%s.", response_i++, mesh_instance->getMeshName().c_str(), mesh_instance->getNodeID().c_str());
```
devyte: This string could easily be made larger than the size of response[60] by a user playing with the example. I suggest a more robust approach.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
   return response;
 }
 
+/**
+   Callback used to decide which networks to connect to once a WiFi scan has been completed.
+
+   @param number_of_networks The number of networks found in the WiFi scan.
+   @param mesh_instance The "this" pointer of the ESP8266WiFiMesh instance that called the function.
+*/
+void networkFilter(int number_of_networks, ESP8266WiFiMesh *mesh_instance) {
```
devyte: mesh_instance => ref

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
+}
+
+/**
+   Callback for when you get a response from other nodes
+
+   @param response The response string received from another node in the mesh
+   @param mesh_instance The "this" pointer of the ESP8266WiFiMesh instance that called the function.
+   @returns The status code resulting from the response, as an int
+*/
+transmission_status_t manageResponse(String response, ESP8266WiFiMesh *mesh_instance) {
```
devyte: response => ref, mesh_instance => ref

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
-
-  /* Scan for other nodes and send them a message */
-  char request[60];
-  sprintf(request, "Hello world request #%d from Mesh_Node%d.", request_i++, ESP.getChipId());
-  mesh_node.attemptScan(request);
-  delay(1000);
+  if (millis() - time_of_last_scan > 3000 // Give other nodes some time to connect between data transfers.
+      || (WiFi.status() != WL_CONNECTED && millis() - time_of_last_scan > 2000)) { // Scan for networks with two second intervals when not already connected.
+    char request[60];
+    sprintf(request, "Hello world request #%d from %s%s.", request_i, mesh_node.getMeshName().c_str(), mesh_node.getNodeID().c_str());
```
devyte: String could become larger than 60 chars => More robust print approach

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,129 @@
...
+
+    if (ESP8266WiFiMesh::latest_transmission_outcomes.empty()) {
+      Serial.println("No mesh AP found.");
+    } else {
+      for (TransmissionResult &transmission_result : ESP8266WiFiMesh::latest_transmission_outcomes) {
+        if (transmission_result.transmission_status == TS_TRANSMISSION_FAILED) {
+          Serial.println("Transmission failed to mesh AP " + transmission_result.ssid);
+        } else if (transmission_result.transmission_status == TS_CONNECTION_FAILED) {
+          Serial.println("Connection failed to mesh AP " + transmission_result.ssid);
+        }
```
devyte: else ? (Robustness)

libraries/ESP8266WiFiMesh/src/CompatibilityLayer.cpp
```diff
@@ -0,0 +1,162 @@
...
+ * Send the supplied message then read back the other node's response
+ * and pass that to the user-supplied handler.
+ *
+ * @target_ssid The name of the AP the other node has set up.
+ * @message The string to send to the node.
+ * @returns: True if the exchange was a succes, false otherwise.
+ * 
+ */
+// DEPRECATED!
+bool ESP8266WiFiMesh::exchangeInfo(String message, WiFiClient curr_client)
```
devyte: Can message be declared reference without losing backwards compat?
Same for curr_client.

libraries/ESP8266WiFiMesh/src/CompatibilityLayer.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Connect to the AP at ssid, send them a message then disconnect.
+ *
+ * @target_ssid The name of the AP the other node has set up.
+ * @message The string to send to the node.
+ * 
+ */
+// DEPRECATED!
+void ESP8266WiFiMesh::connectToNode(String target_ssid, String message)
```
devyte: Can these args be refs?

libraries/ESP8266WiFiMesh/src/CompatibilityLayer.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+	if (!exchangeInfo(message, curr_client))
+		return;
+
+	curr_client.stop();
+	WiFi.disconnect();
+}
+
+// DEPRECATED!
+void ESP8266WiFiMesh::attemptScan(String message)
```
devyte: Can message be a ref?

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.h
```diff
@@ -24,52 +24,193 @@
...
-	
-	WiFiServer  _server;
-	WiFiClient  _client;
+  static String last_ssid;
+  static bool static_IP_activated;
+  static IPAddress static_IP;
+  static IPAddress gateway;
+  static IPAddress subnet_mask;
 
-	void connectToNode(String target_ssid, String message);
```
devyte: I suggest typedefing (or using = ) these definitions, and doing so within the class:
...
using requestHandlerType = std::function<String(String, ESP8266WiFiMesh *)>;
...

The type can then be used and reused without duplication as:
ESP8266WiFiMesh::requestHandlerType someRequestHandler...

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.h
```diff
@@ -24,52 +24,193 @@
...
+  
+  /**
+   * WiFiMesh Constructor method. Creates a WiFi Mesh Node, ready to be initialised.
+   *
+   * @chip_id A unique identifier number for the node.
+   * @handler The callback handler for dealing with received messages. Takes a string as an argument which
+   *          is the string received from another node and returns the string to send back.
+   * 
+   */
+  ESP8266WiFiMesh(uint32_t chip_id, std::function<String(String)> handler);
```
devyte: separate handler arg type with typedef or using

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.h
```diff
@@ -24,52 +24,193 @@
...
+   *                    This can cause problems if several ESP8266WiFiMesh instances exist on the same ESP8266 and use different WiFi channels. 
+   *                    In such a case, whenever the station of one ESP8266WiFiMesh instance connects to an AP, it will silently force the 
+   *                    WiFi channel of any active AP on the ESP8266 to match that of the station. This will cause disconnects and possibly 
+   *                    make it impossible for other stations to detect the APs whose WiFi channels have changed.
+   * @param server_port The server port used by the AP of the ESP8266WiFiMesh instance. If multiple APs exist on a single ESP8266, each requires a separate server port. 
+   *              If two AP:s on the same ESP8266 are using the same server port, you must call deactivateAP on the active AP before calling activateAP on the inactive AP.                   
+   * 
+   */
+  ESP8266WiFiMesh(std::function<String(String, ESP8266WiFiMesh *)> requestHandler, std::function<transmission_status_t(String, ESP8266WiFiMesh *)> responseHandler,
```
devyte: This is precisely where separation of the arg types would be most useful. As it is, the declaration is very difficult to read.

devyte: Use references instead of value and pointer as needed

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.h
```diff
@@ -24,52 +24,193 @@
...
+
+  /**
+   * An empty IPAddress. Used as default when no IP is set.
+   */
+  static const IPAddress empty_IP;
+
+  static String Uint64ToString(uint64_t number, byte base = 16);
+  static uint64_t StringToUint64(String string, byte base = 16);
+
+  std::function<void(int, ESP8266WiFiMesh *)> getNetworkFilter();
```
devyte: Separate return type with typedef or using

libraries/ESP8266WiFiMesh/src/NetworkInfo.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <ESP8266WiFi.h>
+
+const int NETWORK_INFO_DEFAULT_INT = -1;
+
+class NetworkInfo {
+
+private:
+
+  uint8_t bssid_array[6] {0};
```
devyte: Current core convention is to use underscore _ as prefix for class/struct members

libraries/ESP8266WiFiMesh/src/TransmissionResult.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * THE SOFTWARE.
+ */
+
+#include "TransmissionResult.h"
+
+TransmissionResult::TransmissionResult(int new_network_index, transmission_status_t new_transmission_status, bool autofill) : 
+  NetworkInfo(new_network_index, autofill), transmission_status(new_transmission_status)
+{ }
+
+TransmissionResult::TransmissionResult(String new_ssid, int new_wifi_channel, uint8_t new_bssid[6], transmission_status_t new_transmission_status) :
```
devyte: can new_ssid be a ref?

libraries/ESP8266WiFiMesh/src/TransmissionResult.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+TransmissionResult::TransmissionResult(int new_network_index, transmission_status_t new_transmission_status, bool autofill) : 
+  NetworkInfo(new_network_index, autofill), transmission_status(new_transmission_status)
+{ }
+
+TransmissionResult::TransmissionResult(String new_ssid, int new_wifi_channel, uint8_t new_bssid[6], transmission_status_t new_transmission_status) : 
+  NetworkInfo(new_ssid, new_wifi_channel, new_bssid), transmission_status(new_transmission_status)
+{ }
+
+TransmissionResult::TransmissionResult(String new_ssid, int new_wifi_channel, uint8_t new_bssid[6], int new_network_index, transmission_status_t new_transmission_status) :
```
devyte: new_ssid ref?

libraries/ESP8266WiFiMesh/src/NetworkInfo.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+NetworkInfo::NetworkInfo(const NetworkInfo &other) : ssid(other.ssid), wifi_channel(other.wifi_channel), network_index(other.network_index)
+{
+  copyBSSID(other.bssid);
+}
+
+NetworkInfo & NetworkInfo::operator=(const NetworkInfo &other)
+{
+  ssid = other.ssid;
```
devyte: Should this be protected with if(this != &other) ?

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -18,151 +18,582 @@
...
+}
+
+/**
+ * Note that using a base higher than 16 increases likelihood of randomly generating ssid strings containing controversial words. 
+ * 
+ * @param string The string to convert to uint64_t. String must use radix "base".
+ * @param base The radix of "string". Must be between 2 and 36.
+ * @returns A uint64_t of the string, using radix "base" during decoding.
+ */
+uint64_t ESP8266WiFiMesh::StringToUint64(String string, byte base)
```
devyte: string => ref

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -18,151 +18,582 @@
...
+  }
+
+  return (result == "" ? "0" : result);
+}
+
+/**
+ * Note that using a base higher than 16 increases likelihood of randomly generating ssid strings containing controversial words. 
+ * 
+ * @param string The string to convert to uint64_t. String must use radix "base".
+ * @param base The radix of "string". Must be between 2 and 36.
```
devyte: Is this 2-36 range enforced somehow?

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -18,151 +18,582 @@
...
+  char current_character[1];
+  for(uint32_t i = 0; i < string.length(); i++)
+  {
+    result *= base;
+    current_character[0] = string.charAt(i);
+    result += strtoul(current_character, NULL, base);
+  }
+  
+  return result;
+}
```
devyte: Can this function fail? what happens if the string passed as arg is invalid or has a mistake?
Does it make sense to put these conversion files elsewhere? At least in standalone files.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -18,151 +18,582 @@
...
+  String result = "";
+
+  while(number > 0)
+  {
+    result = String((uint32_t)(number % base), base) + result;
+    number /= base;
+  }
+
+  return (result == "" ? "0" : result);
+}
```
devyte: Does it make sense to put this function in a standalone file?

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -18,151 +18,582 @@
...
+    else
+      Serial.print(string_to_print);
+  }
+}
+
+/**
+ * Note that using a base higher than 16 increases likelihood of randomly generating ssid strings containing controversial words. 
+ * 
+ * @param number The number to convert to a string with radix "base".
+ * @param base The radix to convert "number" into. Must be between 2 and 36.
```
devyte: is this 2-36 range enforced somehow?

devyte:
@aerlon I had originally thought to wait on making this review, because most of what I saw was guidelines and cosmetic. However, I just took a last look before merging and noticed the passing by value instead of reference, and the duplicate declarations of the std::function types, so I decided to make the full review now.
Do you think you can still make the changes for 2.4.2?

aerlon:
@devyte Today is a little busy, but I should have time to go through this tomorrow.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
aerlon: Makes sense.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
aerlon: Camel case or snake case work either way for me. I noticed snake case was used in the original mesh library, but I take it style conventions has changed since then?

Regarding the actual mesh name, though, I think it may be prudent to not write it in the same style as a programming object. It is just a network name, after all, and its lack of relationship to other program objects may be easier for the novice user to deduce when it does not follow the typical naming conventions for objects in the library. Thus `String meshName(F("Mesh_Node"));` or `String meshName(F("MeshNode_"));` or some such mesh name format may be useful.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
aerlon: In this particular case I’d argue that we only care about if there is an error occurring, and those are the cases checked for here. Adding an else would not really aid the user since it would only capture all the successful transmissions which we already get confirmations about elsewhere in the code (from the responseHandler).

libraries/ESP8266WiFiMesh/src/NetworkInfo.cpp
aerlon: I actually thought about that, but ultimately there seemed to be no real benefit from doing it (copying self would not cause any problems) so we could save some computing power by avoiding it. Would you prefer to have the check?

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
aerlon: No. The problem is that I’m not sure how I would do that in a nice way since I can’t throw exceptions in this environment (as far as I know). The core conversion functions used by my code (`String()` and `strtoul()`) also seem to accept arguments that are out of range, so I just propagated that behaviour through my functions. Of course I could clamp the values to the 2-36 range, but then I would silently convert invalid input arguments to other values which I think could confuse the user more than just getting wrong results. Ideas are welcome.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
aerlon: Sure, I could do that. Do you mean as a separate utility class or just definitions in a separate .cpp?

aerlon:
@devyte Thanks for the feedback! I left some comments for further discussion on a few of the points. I’ll have to do some testing before answering a few of the others.

Actually it is quite good that you postponed the merge because looking through the new API for this I got an idea for a change that would be nice to get in before the code goes public. I’ll see if I can find you on Gitter tomorrow to discuss how creative I should get while implementing that change.

Ready by 1st of August you say? No problem! Software is always ready on time, as I’m sure you know. (Seriously though, it should be OK.)

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
devyte: Style wasn't really pursued before, commenting on it is rather recent. In general, current naming convention is camelCase for C++ and snake_case for C.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
devyte: If there is ever another error status value added and returned, the if-else-if clause will silently not catch anything. For cases like this, the else should be akin to an assert. It's a similar case as having default: in a switch. It is rarely a good practice not to have one.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
devyte: Just movemto some separate file would be enough. The functions are generic and not specific to the contents of these files.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
devyte: WTB exceptions :) there have been several attempts, some I suspect are rather close.

I suggest a debug message and/or an assert for now.

libraries/ESP8266WiFiMesh/src/NetworkInfo.cpp
devyte: No, I just had to ask the question to make sure it wasn't an oversight.

devyte:
Code reviews always help in some way :)
Just look me up to discuss.

"We need it for yesterday, bug free, and performance tested"

libraries/ESP8266WiFiMesh/src/CompatibilityLayer.cpp
aerlon: No, that causes the example code to fail. However, we can change the private methods of CompatibilityLayer.cpp as long as we only care about keeping the public API backwards compatible. I’ve done this in the most recent commit.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
aerlon: As far as I can tell, any invalid input is treated as a 0.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
aerlon: I realized that a good way to get the robustness advantage you mentioned while avoiding too many notification events for the user was simply to add a condition for every current state, and then an else. This is now implemented in the most recent code update.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.h
```diff
@@ -24,52 +24,278 @@
...
+  
+  IPAddress getStaticIP();
+  void disableStaticIP();
+
+  /**
+   * An empty IPAddress. Used as default when no IP is set.
+   */
+  static const IPAddress emptyIP;
+
+  static String uint64ToString(uint64_t number, byte base = 16);
```
devyte: It looks like I wasn't clear about this. I meant move this completely out of the class. These two functions have nothing to do with the mesh nodes, or wifi, or even with the ESP really. They should be free standing functions. 
But let's leave that for a subsequent cleanup.

libraries/ESP8266WiFiMesh/src/UtilityMethods.cpp
```diff
@@ -0,0 +1,127 @@
...
+  }
+  
+  return result;
+}
+
+/**
+ * Calculate the current lwIP version number and store the numbers in the _lwipVersion array.
+ * lwIP version can be changed in the "Tools" menu of Arduino IDE.
+ */
+void ESP8266WiFiMesh::storeLwipVersion()
```
devyte: it should be possible to simplify this. As it is, it's rather fragile.
Note: investigate getting lwip version at compile time.

libraries/ESP8266WiFiMesh/examples/HelloMesh/HelloMesh.ino
```diff
@@ -1,50 +1,131 @@
...
+    } else {
+      for (TransmissionResult &transmissionResult : ESP8266WiFiMesh::latestTransmissionOutcomes) {
+        if (transmissionResult.transmissionStatus == TS_TRANSMISSION_FAILED) {
+          Serial.println("Transmission failed to mesh AP " + transmissionResult.SSID);
+        } else if (transmissionResult.transmissionStatus == TS_CONNECTION_FAILED) {
+          Serial.println("Connection failed to mesh AP " + transmissionResult.SSID);
+        } else if (transmissionResult.transmissionStatus == TS_TRANSMISSION_COMPLETE) {
+          // No need to do anything, transmission was successful.
+        } else {
+          Serial.println("Invalid transmission status for " + transmissionResult.SSID + "!");
```
devyte: this should be an assert, because if the condition ever happens it's due to a programming error or mem corruption.

devyte: The text of these prints should be moved to flash

devyte:
A few minor cleanups pending, but good enough to merge as-is.

libraries/ESP8266WiFiMesh/src/UtilityMethods.cpp
aerlon: Agreed. However there seemed to be no official way to get the lwIP version when I wrote this code, so I had to make my own. Of course, a more officially supported method would be better.

aerlon:
Nice. I will open a new PR when I’m done with the fixes for the remaining issues.

PR #4729, opened 2018-05-14
lwip2 fix and update

d-a-v:
* avoid crash when IP address is set to 0(any) by dhcp not getting its lease renewal in due time
* automatically remove oldest PCBs in time-wait state, limit their number (thanks to @me-no-dev)
  and update faq update about tcpCleanup()
* update an option in the interactive example to go back to DHCP client mode from static mode 
* add a macro HAS_PHY_CAPTURE=1 indicating capture facility is available


d-a-v:
I just manually rebased. Looks like #4737 is now required.
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

Issue #4726, opened 2018-05-14
Can't be installed successfully


lukeweston:
Arduino IDE Version: 1.8.5
Mac OSX 10.12.6
Python: 2.7.10
Which ESP8266 hardware I'm using is irrelevant, I'm not really up to that point.
I've shown the specific git clone path and board manager JSON file URL, below.

I have inserted the following entry into the additional board manager URLs:
`http://arduino.esp8266.com/stable/package_esp8266com_index.json
`
The ESP8266 entry is now visible for installation in the Board Manager GUI.
I've attempted to install, with v 2.4.1 selected. (Selecting 2.4.0 does not fix the issue.)

Installation fails with the following error message in red text:
`Error downloading https://github.com/esp8266/Arduino/releases/download/2.4.1/esp8266-2.4.1.zip
`

OK, that doesn't work, so let's try manually installing it.
I don't mind manual installation, as long as it gets the job done in a usable way.
(Remember this is OSX, so we're inside /Applications/Arduino.app/)
`cd /Applications/Arduino.app/Contents/Java
`
`cd hardware
`
`mkdir esp8266com
`
`cd esp8266com
`
`git clone https://github.com/esp8266/Arduino.git esp8266
`
`cd esp8266/tools
`
So far so good.

`get.py` fails. I'm not sure what Python packages etc, if any, are prerequisites for this script. No such requirements, beyond Python 2.7 itself, are documented on the documentation page.

Python is installed. (2.7.10)

I have followed the instructions as they are given on the documentation page, for both the manual install-from-git and the Board Manager cases, and in neither case can I get a successful installation.

```
$ python get.py
Traceback (most recent call last):
  File "get.py", line 15, in <module>
    import tarfile
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py", line 52, in <module>
    import copy
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 52, in <module>
    import weakref
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/weakref.py", line 14, in <module>
    from _weakref import (
ImportError: cannot import name _remove_dead_weakref
```

----------------------------- Delete above -----------------------------

### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
[...]

PR #4730, opened 2018-05-14
Updated README.md to clarify installation folder of git version within Mac OS X 

tzahari:
Clarify the root folder of the git installation method for Mac. Thanks to @lukeweston wich was noted this folder in #4726


README.md
```diff
@@ -33,8 +33,12 @@ Documentation: [https://arduino-esp8266.readthedocs.io/en/2.4.1/](https://arduin
 ### Using git version
 [![Linux build status](https://travis-ci.org/esp8266/Arduino.svg)](https://travis-ci.org/esp8266/Arduino)
 
-- Install Arduino 1.8.2 from the [Arduino website](http://www.arduino.cc/en/main/software).
-- Go to Arduino directory
+- Install the current upstream Arduino IDE at the 1.8 level or later. The current version is at the [Arduino website](http://www.arduino.cc/en/main/software).
+- Go to Arduino application directory
+  - For Mac OS X normally: 
+  ```bash
+  cd /Applications/Arduino.app/Contents/Java
```
igrr: Is it normal to install Arduino into global applications directory, instead of putting it somewhere in home directory?

Maybe we can relax this to `/path/to/Arduino.app`?

d-a-v: I used to use mac too some time ago, and that was the place where I installed Arduino, so the place seemed right to me.
But you are right, not everyone is root on his machine, or want to install a single version of arduino in the global app dir.
I will fix this.

tzahari: Yes, you are right. In Mac OS X you can put the app bundle everywhere.
It seems there is no easy way to get the folder of an app.

PR #4736, opened 2018-05-16
Add a build directory for libbearssl.a

earlephilhower:
Simple git submodule and makefile for building the bearssl library
from source in the Arduino tree.


earlephilhower:
@d-a-v , I tried to copy your setup in lwip2 for this.  Can you eyeball things and see if I've got it right?  Thanks!

d-a-v:
That works for me !
libssl.a is regenerated with `make install`
What do you think of adding a `make merge-upstream` or so to manually add the remote upstream and do the git pull (for test purpose only) ?

devyte:
LGTM!!

earlephilhower:
Several fixes/updates per the gitter discussion:
* Fixed the makefile `install` to also copy over the headers
* Add a git remote to upstream on submodule init, and a `merge-upstream` makefile target
* Generate a git hash ID on install, and include the hash in the ESP.getFullVersion() call

PR #4745, opened 2018-05-19
Add UART interrupt dispatcher

unaiur:
ESP8266 UART ports share a single interrupt service routine (ISR).
This code adds an interrupt dispatcher that allows to define a
different ISR for each UART port.

Why?

NeoPixelBus can use an ISR to send data though Serial1 asynchronously. But it only works when Serial0 is configured as TX_ONLY. Otherwise NeoPixelBus and Serial-RX ISRs conflicts and the ESP8266 reboots. This commit allows both ISRs to coexist.

### DIFF
--- a/cores/esp8266/HardwareSerial.h
+++ b/cores/esp8266/HardwareSerial.h
@@ -184,6 +184,16 @@ public:
         return uart_has_overrun(_uart);
     }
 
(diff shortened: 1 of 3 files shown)


earlephilhower:
Closing in favour of #5555 which is undergoing work presently.

PR #4754, opened 2018-05-22
Report connected when any bytes are available

### LINKED ISSUE #4738: BearSSL::WiFiClientSecure not working the same way axTLS did
#### Platform

- Hardware:				[ESP-12]
- Core Version:			[Mon May 14 20:46:47 2018 -0700]
- Operating System:		[MacOS]

### Settings in IDE

- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

BearSSL::WiFiClientSecure does not connect. No debug messages at all (debug tls+ssl+http client). My custom logs attached.

```cpp

#include <WiFiClientSecure.h>

void setup() {
 if (!client.connect(host, httpsPort)) {
    res.code = 0; res.content = F("connection failed");
    return (res); // 1 = connection failed
  }
}

void loop() {

}

```

### Debug Messages

```
00:01:374 Free HEAP: 36264
00:01:375 Client Secure started 1 ms
00:01:375 Free HEAP: 31704
00:01:376 connecting to api.monitec.systems
00:01:969 Connection failed 593 ms
00:01:973 Request failed. Reason: (0) connection failed
```

earlephilhower:
The SSL pipeline is multi-stage, and the TCP connection can go down
even though there is still data waiting to be decrypted or in the
decryption buffer.

Explicitly check that there if there can be any data made available
to the app, and if so report that we are still connected().  When
there is no data and there is no TCP connection, report disconnected.

Fixes #4738

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
@@ -228,10 +228,8 @@ bool WiFiClientSecure::_clientConnected() {
 }
 
 uint8_t WiFiClientSecure::connected() {
-  if (_recvapp_len) {
-    return true;
-  }
-  if (_client && _client->state() == ESTABLISHED && _handshake_done) {
+  if (WiFiClient::connected() || available() ||
+      (_clientConnected() && _handshake_done)) {
     return true;
   }
   return false;


earlephilhower:
I think the semantics are fine.  `connected()` means "should I bother trying to read or write?" to most people, not "a TCP pipe is in the established state."

For SSL it's just odd because the TCP can drop a long time before the last byte of data makes it to the app.   I took into account one way it could get into this state, but not the other (where the SSL engine needs to do one more decryption on the last TLS frame).

earlephilhower:
Removed the fixes comment because there's something else funky w/cloudflare hosts, but the code in the patch is correct WRT ::connected() and does fix a visible-by-inspection bug.

PR #4797, opened 2018-06-07
Windows 10 instructions

gojimmypi:
created separate section with explicit instructions for using git version of library with Windows 10 Arduino IDE (or Visual Micro Arduino IDE for Visual Studio)


Juppit:
By default Arduino installes under window in `C:\Program Files (x86)\Arduino` and so we find the harware folder as: 
`C:\Program Files (x86)\Arduino\hardware`

.

lrmoreno007:
```
and so we find the harware folder as: C:\Program Files (x86)\Arduino\hardware
```
If  you aren't admin, you can't install there. Hardware folder in Documents is the most correct place for everybody, except if you want install for all users.

Juppit:
If you aren't admin you are asked to install it as admin, and so you install it in
`C:\Program Files (x86)\Arduino\hardware`
I tried both.

gojimmypi:
My `C:\Program Files (x86)\Arduino\hardware` contains only items initially installed at application-install time. (I'm also local admin)  Other hardware installed by Arduino _Boards Manager_ end up in:

`C:\Users\{username}\AppData\Local\Arduino15\packages\esp8266\hardware` 

(then with version numbers for a directory name such as: 

`C:\Users\{username}\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\2.4.1`

There's apparently a download cache, as if I remove and re-install a hardware library with the Board Manager, the file dates are retained from when first downloaded. 

In Arduino IDE: File - Preferences, my `Sketchbook Location` is (I believe the default):

`C:\Users\{username}\Documents`

Which is, per my updated instructions in this PR:

`%USERPROFILE%\Documents\`

...and this is where I have my hardware directory. This is also where I installed this ESP8266/Ardunio repository per the current instructions. It seems to work.

See also:  https://github.com/arduino/arduino/wiki/arduino-ide-1.5-3rd-party-hardware-specification

> This specification allows a 3rd party vendor/maintainer to add support for new boards inside the Arduino IDE by providing a file to unzip into the hardware folder of Arduino's sketchbook folder.

Granted, this seems to be a "pre-Board Manager" spec, but I don't think the` C:\Users\{username}\AppData\Local` would be a good choice for manual installs due to the volatile nature when otherwise managed by the Board Manager.

It should be noted for completeness that I also have VisualGDB installed; it stores  files in `C:\SysGCC\esp8266` (but these files are also managed by sysprogs folks, so not mentioned in instructions)

In any case, my PR still includes the "_This is typically the sketchbook directory_" text.

Are there some suggestions for improved clarity?

Juppit:
After uninstall, cleaning the registry, deleting `C:\Users\{username}\AppData\Local\Arduino15`, deleting  `C:\Users\{username}\Documents\Arduino` ...
I install Arduino (`arduino-1.8.5-windows.exe` from `https://www.arduino.cc/`) and then Arduino sets the default sketchbook location to:
`C:\Users\{username}\Documents\Arduino` and there is only a folder `libraries` after installation.
This is the same behaver as the existing `installing.rst` describes.
And in my opinion this should be also the default location of the hardware folder:
`C:\Users\{username}\Documents\Arduino\hardware`

I personly use a script wich consists of the following steps:
1.) Figure out how will the package be called in build_boards_manager_package.sh
     `ver=...`
2.) Call `./build_boards_manager_package.sh`
3.) `cp versions/$ver/package_esp8266com_index.json versions/package_esp8266com_index.json`
4.) Start a small HTTPServer `python -m SimpleHTTPServer`

When you insert `http://localhost:8000/versions/package_esp8266com_index.json`
as `Additional Boards Manager URLs:` in the Arduino-IDE under `file->preferences`
then all the stuff will delivered to the right location `C:\Users\{username}\AppData\Local\Arduino15`.

lrmoreno007:
I use my own script for windows, valid for ESP8266 and ESP32:
https://github.com/lrmoreno007/ESP-Easy-Updater-for-Windows

I create shortcuts to the scripts in my desktop and update git version with one click.

When I change some code in the core and I want back to the official git version, only delete the folder of esp8266 or esp32 and click in the shortcut.

Regards.

gojimmypi:
@Juppit Indeed you are quite correct. Thank you. I did the same and confirmed the freshly installed Arduino app has a default sketchbook directory of `C:\Users\{username}\Documents\Arduino` I've never done a complete purge of all existing directories, nor do I have an explanation for the old value; perhaps it remembered it from my `C:\Users\{username}\AppData\Local\Arduino15\preferences.txt` (from an older version?) 

Any discussion regarding Arduino libraries should probably also reference:

https://www.arduino.cc/en/Guide/Libraries

I've made the changes:

https://github.com/gojimmypi/ESP8266-Arduino/blob/master/doc/installing.rst

PR #4804, opened 2018-06-08
Make BearSSL::write() blocking

earlephilhower:
When a message is sent by the app that is larger than the SSL buffer,
it will take multiple TLS fragments to transfer.  Writes will now
loop through and not complete until either all data is transferred
or there is an error.

The prior behavior caused an error in HTTPClient (and others) as it would only write the first chunk (i.e. 512b by default) and return "512" no matter how large the app write request was.  Unless the app was expecting this and trying it in a loop, this would either cause a protocol error if the app did not check write's return value, or an app error if it detected the short write.


earlephilhower:
Just tested this patch sending a 4096 byte char[] to an openssl s_server and it went through fine using the default 512b send fragment size.

PR #4805, opened 2018-06-08
Update boards.rst

harlock974:
Informations about Wemos D1 mini Lite

PR #4809, opened 2018-06-11
Wemos D1 Mini Lite informations

harlock974:


d-a-v:
Sorry if I was not clear enough in #4805.
You cannot edit this file, it is autogenerated.
You **must** edit `tools/boards.txt.py` line 552 and add your changes there.
You can find a multi-line description example in line 581.
Once this is done, you must call this script with option `--allgen` and this will regenerate/modify `boards.rst`.
```
./tools/boards.txt.py --allgen
```
You need to commit these two files in order for Travis-CI to accept your pull-request.

PR #4827, opened 2018-06-19
BearSSL: WiFiClientSecure insecure by default + setFingerprint(const char*) and example update

d-a-v:
Made to help transition from axTLS:

* set BearSSL client insecured by default
* overload `setFingerprint()` with axTLS's `char*` format (in a separate source file to save flash if unused)
* generate a linker error message when trying to use unimplementable `::verify()`

### DIFF
--- a/cores/esp8266/FunctionalInterrupt.cpp
+++ b/cores/esp8266/FunctionalInterrupt.cpp
@@ -1,24 +1,70 @@
 #include <FunctionalInterrupt.h>
-
+#include <Schedule.h>
+#include "Arduino.h"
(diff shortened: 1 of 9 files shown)


devyte:
@d-a-v @earlephilhower wasn't it discussed to have insecure disabled by default, to disallow insecure applications by default? I seem to remember somebody saying that, if insecure is needed, that it should be set by the user explicitly.

igrr:
Yeah, I mentioned a couple of times that making TLS insecure by default defeats the purpose.

PR #4832, opened 2018-06-20
Pass Client parameter to httpUpdate to use any BearSSL security method for a firmware update (with example)

Jeroen88:
Added two new update() and one updateSpiffs() function that uses the ESP8266HTTPClient adapted by me in PR #4825. With these adaptations, it is possible to do an encrypted and authenticated server firmware or SPIFFS update, using any of the supported security options of the WiFiClientSecureBearSSL library. httpUpdateSecure.ino is a complete example.

To try this, you need either a sketch with enough free memory to allocate the full TLS/SSL-buffer (16k), or a, currently not very widely supported Maximum Fragment Length Negotiation enabled, server (or do the same trick as I do, using a nodeJS server and set maxSendFragmentLength) that sends the correct response headers and serves the binary. You need a certificate store in SPIFFs that includes your root CA, or a self signed certificate, or a certificate fingerprint. In the latter two cases you need to edit the example to use allowSelfSignedCerts() or setFingerprint(). It is also possible to use encryption without server authentication. In this case use setInsecure(). Of course any of the other supported WiFiClientSecureBearSSL functions may be used as well.

I have a question: in ESP8266httpUpdate.cpp with these new functions, a Client& is passed into the update() functions. After the headers have been checked in handleUpdate(), a WiFiClient * is retrieved from the http client :

`
WiFiClient * tcp = (WiFiClient *) http.getStreamPtr();
`

This works fine, however: in PR #4825 I designed httpClient as a _generic_ client for _any_ Client. There is no fundamental reason not to use e.g. EthernetClient. But in handleUpdate() a WiFiClient * is necessary to do the actual update. Is it possible to use a base class Client instead?


Jeroen88:
Travis build failed because of macro redefinitions not within my scope, e.g.:
In file included from common/Arduino.h:241:0,
                 from ../../cores/esp8266/StreamString.cpp:23:
../../cores/esp8266/WCharacter.h:24:0: warning: "isascii" redefined [enabled by default]
 #define isascii(__c) ((unsigned)(__c)<=0177)
 ^

libraries/ESP8266HTTPClient/examples/BasicHttpClient/BasicHttpClient.ino
```diff
@@ -40,34 +42,39 @@ void loop() {
...
 
+    WiFiClient client;
+
     HTTPClient http;
 
     USE_SERIAL.print("[HTTP] begin...\n");
     // configure traged server and url
     //http.begin("https://192.168.1.12/test.html", "7a 9c f4 db 40 d3 62 5a 6e 21 bc 5c cc 66 c8 3e a1 45 59 38"); //HTTPS
-    http.begin("http://192.168.1.12/test.html"); //HTTP
+    if (http.begin((Client &)client, "http://tls.mbed.org/")) {  // HTTP
```
earlephilhower: Any reason this one picked?  Isn't tls.mbed.org https-only?

libraries/ESP8266HTTPClient/examples/Authorization/Authorization.ino
```diff
@@ -40,22 +42,24 @@ void loop() {
...
+    WiFiClient client;
+
     HTTPClient http;
 
     USE_SERIAL.print("[HTTP] begin...\n");
     // configure traged server and url
 
 
-    http.begin("http://user:password@192.168.1.12/test.html");
+    http.begin((Client&)client, "http://guest:guest@jigsaw.w3.org/HTTP/Basic/");
```
earlephilhower: Isn't there no need for the cast `(Client&)` here and in the other examples?  Your prototype will make the C++ compiler pass the pointer no matter what, no?

libraries/ESP8266HTTPClient/examples/BasicHttpClient/BasicHttpClient.ino
earlephilhower: Also, thanks for checking return codes!

libraries/ESP8266HTTPClient/examples/StreamHttpClient/StreamHttpClient.ino
```diff
@@ -65,7 +67,7 @@ void loop() {
         uint8_t buff[128] = { 0 };
 
         // get tcp stream
```
earlephilhower: Can we get rid of the `stream` temp variable completely since it's no longer needed with your fix?  i.e. there is no `getStreamPtr()` needed anymore?

libraries/ESP8266HTTPClient/examples/StreamHttpsClient/StreamHttpsClient.ino
```diff
@@ -0,0 +1,116 @@
...
+        if (httpCode == HTTP_CODE_OK) {
+
+          // get lenght of document (is -1 when Server sends no Content-Length header)
+          int len = http.getSize();
+
+          // create buffer for read
+          uint8_t buff[128] = { 0 };
+
+          // get tcp stream
+          WiFiClient * stream = &client;
```
earlephilhower: Same as above, no more "get tcp stream" with your change.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -129,9 +134,70 @@ void HTTPClient::clear()
...
+    }
+
+    String protocol = url.substring(0, index);
+    if(protocol != "http" && protocol != "https") {
+        DEBUG_HTTPCLIENT("[HTTP-Client][begin] unknown protocol '%s'\n", protocol.c_str());
+        return false;
+    }
+
+    _port = (protocol == "https" ? 443 : 80);
+    return beginInternal(url, protocol.c_str());
```
earlephilhower: There shouldn't be any need for the protocol anymore, no?  The `WiFiCilent*` passed in should already either be encrypted or not, and the code paths for both are the same, no?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -129,9 +134,70 @@ void HTTPClient::clear()
...
+    _tcpDeprecated.reset(nullptr);
+    _transportTraits.reset(nullptr);
+#endif
+    _tcp = &client;
+
+     clear();
+    _host = host;
+    _port = port;
+    _uri = uri;
+    _protocol = (https ? "https" : "http");
```
earlephilhower: Same question as above about the protocol, is it needed anymore?

libraries/ESP8266HTTPClient/examples/StreamHttpClient/StreamHttpClient.ino
earlephilhower: Actually, I see later on that you still have a "getStreamPtr()" method with the new one, too.  Any reason not to just keep using it?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -912,15 +1016,22 @@ bool HTTPClient::connect(void)
...
         return false;
     }
+#endif
 
 
 #ifdef ESP8266
-    _tcp->setNoDelay(true);
+#ifdef KEEP_PRESENT_API
+    if(_tcpDeprecated)
+        _tcpDeprecated->setNoDelay(true);
```
earlephilhower: Wouldn't you want to set it on `_tcp` as well (outside this `#define`, of course)?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
```diff
@@ -20,14 +20,22 @@
...
 
+#define KEEP_PRESENT_API
+
 #include <memory>
 #include <Arduino.h>
+
+#ifdef KEEP_PRESENT_API
 #include <WiFiClient.h>
+#else
+#include <Client.h>
```
earlephilhower: Why move to abstract `Client` from `WiFiClient`?  I don't remember ever seeing anything else take this root class.

libraries/ESP8266httpUpdate/examples/httpUpdateSecure/httpUpdateSecure.ino
```diff
@@ -0,0 +1,138 @@
+/**
```
earlephilhower: While I'm happy to see someone using CertStore, for ESP8266 updates that's probably not what we want people to do since it's saying you'd accept an update from anyone on the internet.

What most people would want to do is have a single, private certificate that they generate and keep secure on their own company's system.  It would make this new example shorter and simpler, too, and focus it on the new way of doing the update vs. the overhead for making a CertStore...

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp
```diff
@@ -276,7 +304,7 @@ HTTPUpdateResult ESP8266HTTPUpdate::handleUpdate(HTTPClient& http, const String&
                 ret = HTTP_UPDATE_FAILED;
             } else {
 
-                WiFiClient * tcp = http.getStreamPtr();
+                WiFiClient * tcp = (WiFiClient *) http.getStreamPtr();
```
earlephilhower: I'm fine with casts like this, but I think @devyte will probably ask for static_cast<>.  No change from me, just a warning...

Jeroen88:
@earlephilhower thnx for the review! I will make updates and give answers to your questions / remarks soon!

libraries/ESP8266HTTPClient/examples/BasicHttpClient/BasicHttpClient.ino
Jeroen88: I will change this in http://jigsaw.w3.org/HTTP/connection.html

libraries/ESP8266HTTPClient/examples/Authorization/Authorization.ino
Jeroen88: You will be right. I'll check it and then change it

libraries/ESP8266HTTPClient/examples/StreamHttpClient/StreamHttpClient.ino
Jeroen88: I do not quite understand you here. 
The getStreamPtr() is embedded in the #ifdef KEEP_PRESENT_API, so it's just there for backward compatibility. 
I added the streaming example on purpose, because a server response may be too large to fit into a String, making it impossible to use getString(), but still possible to stream it. Moreover, since it is String getString() instead of String &getString() the response might even be copied (allthough the compiler may optimize this, I don't know)
Could you explain what you are suggesting?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
Jeroen88: Yes it is still needed. The ESP8266HTTPClient is setting up (and maintaining with Keep-Alive) the connection, and it should know what port to use. The idea (also in the original version) of this begin() method , is to pass all the needed information to set up the connection through a fully qualified URL.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
Jeroen88: The _port is needed, see in connect(): if(!_tcp->connect(_host.c_str(), _port)) { ...
If you mean that that _protocol can be dropped as class variable, I think you are right. I'll check that and wrap it into a #ifdef KEEP_PRESENT_API if possible

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
Jeroen88: Client::setNoDelay() does not exist. If I change Client in WiFiClient its is possible, see my remark later

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
Jeroen88: The idea was that this class could be used for other Clients as well, eg an EthernetClient as defined here (https://github.com/esp8266/Arduino/blob/master/libraries/Ethernet/src/EthernetClient.h). But because most people will use the WiFiClient, I will change Client to WiFiClient

libraries/ESP8266httpUpdate/examples/httpUpdateSecure/httpUpdateSecure.ino
Jeroen88: I do not understand:
You will not accept an update from anyone on the internet because the URL determines where you fetch the update, and the Client identifies the host. Also, there is no need to populate the cert store with all root certificates, just your company's one is possible too.
Do you want an extra, simpler example just using a cert? Could you provide it?

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp
Jeroen88: I will try the static_cast

libraries/ESP8266HTTPClient/examples/Authorization/Authorization.ino
Jeroen88: The cast has been removed

libraries/ESP8266HTTPClient/examples/StreamHttpClient/StreamHttpClient.ino
Jeroen88: I understand you now. The getStreamPtr() is kept as part of the existing API, however in the examples the unneeded call to getStreamPtr() has been removed.

libraries/ESP8266HTTPClient/examples/StreamHttpsClient/StreamHttpsClient.ino
Jeroen88: getStreamPtr() has been removed from the example

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
Jeroen88: The _protocol may be dropped in the new class, but beginInternal() is still used, and this function uses _protocol as well. In this function _protocol could be changed into a variable local to that function. I did not do this to stay as close as possible to the original code.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
Jeroen88: Now setNoDelay() is alway called

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp
Jeroen88: Because http.getStreamPtr() returns a WiFiClient now, casting is no longer necessary

Jeroen88:
Requested review changes are made and posted as a new PR #4980.

PR #4841, opened 2018-06-22
Add a callback function for connection loop

videosoftarg:


earlephilhower:
All builds fail as this looks like an incomplete set of changed files pushed.  Looks abandoned, too.  Closing.

PR #4864, opened 2018-06-29
Support for concatenation of headers

mickwheelz:
If the server returns several headers of the same key (e.g Set-Cookie) only the last one is returned, causing issues in communicating with some servers where cookies are required.

This change concatenates the headers of the same key separated by ";" to alleviate this issue


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1026,7 +1026,7 @@ int HTTPClient::handleHeaderResponse()
 
                 for(size_t i = 0; i < _headerKeysCount; i++) {
                     if(_currentHeaders[i].key.equalsIgnoreCase(headerName)) {
-                        _currentHeaders[i].value = headerValue;
+                        _currentHeaders[i].value += headerValue + ";";
```
earlephilhower: Can we follow the HTTP spec and use a `,` instead for concatenation?

[RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

earlephilhower:
Verified multiple `Set-Cookie` lines concatenate properly and single-line headers don't get any add'l unneeded commas.

earlephilhower:
I'll let this go as-is, but it doesn't actually help for cookies proper.  `Set-Cookie` headers seem to include both `,` and `;` and so you can't use a concat'ed string to set individual cookies because splitting it is non-deterministic.

To do it right and support multiple cookies, we will need to have this call return an iterator and potentially a whole series of individual strings for parsing.  It's doable, but not something I can fix in the online editor.

For example, see the following header set from GOOG:

````
[HTTP-Client][handleHeaderResponse] RX: 'Date: Thu, 04 Oct 2018 03:35:13 GMT'
[HTTP-Client][handleHeaderResponse] RX: 'Expires: -1'
[HTTP-Client][handleHeaderResponse] RX: 'Cache-Control: private, max-age=0'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Type: text/html; charset=ISO-8859-1'
[HTTP-Client][handleHeaderResponse] RX: 'P3P: CP="This is not a P3P policy! See g.co/p3phelp for more info."'
[HTTP-Client][handleHeaderResponse] RX: 'Server: gws'
[HTTP-Client][handleHeaderResponse] RX: 'X-XSS-Protection: 1; mode=block'
[HTTP-Client][handleHeaderResponse] RX: 'X-Frame-Options: SAMEORIGIN'
[HTTP-Client][handleHeaderResponse] RX: 'Set-Cookie: 1P_JAR=2018-10-04-03; expires=Sat, 03-Nov-2018 03:35:13 GMT; path=/; domain=.google.com'
:c 1, 536, 1072
[HTTP-Client][handleHeaderResponse] RX: 'Set-Cookie: NID=140=g70gxMT35ZROzCbLx8AO7KTjmIbw4GoOCddaseHafS7QceQFbO7covx5i9xydvSun0nKEJP4b7vKqvD0On8X6H3kZ1JCOWXMQnTAE_Kcw6oiqXiDOYOiE8E_HsMCFVZ8PsiK; expires=Fri, 05-Apr-2019 03:35:13 GMT; path=/; domain=.google.com; HttpOnly'
[HTTP-Client][handleHeaderResponse] RX: 'Accept-Ranges: none'
[HTTP-Client][handleHeaderResponse] RX: 'Vary: Accept-Encoding'
[HTTP-Client][handleHeaderResponse] RX: 'Connection: close'
````

PR #4872, opened 2018-07-01
Compatibility and IRQ fixed for waveform/tone/pwm

### LINKED ISSUE #4830: cli() in tone/waveform, needed?
Hi!

I found a cli() (disable interrupts right?) call here:
https://github.com/esp8266/Arduino/blob/91bb97ddb30082f629b3ff2c8fc992c20fbad822/cores/esp8266/core_esp8266_waveform.c#L214

Is that needed at all? Why is it there? :)
It seems the only use in that source file, nowhere to re-enable interrupts, isn't that a problem?

Mostly trying to understand the code here, hope it's ok to have an issue for that :)

Thanks!

earlephilhower:
Fix a compiler ambiguity introduced with a floating point frequency option
for tone().  Thanks to @Rob58329 for discovering this and proposing the
fix.

Match original analogWrite behavior by going from 0...1023 (PWMRANGE) and
not 0...1024, and also explicitly set the analogWrite pin to an OUTPUT.
Thanks to @jandrassy for finding this.

Fixes #4830 0 discovered by @cranphin where interrupts were disabled on a
stopWaveform().  Remove that completely and bracket the update of non-atomic
fields in the structure with disable/enable IRQs for safety.


earlephilhower:
@Rob58329, @jandrassy , @cranphin if you could give this PR a go and report if it corrects the problems you found, I'd be much obliged.  Thx!

cores/esp8266/Tone.cpp
```diff
@@ -70,6 +70,12 @@ void tone(uint8_t _pin, double frequency, unsigned long duration) {
 }
 
 
+// Fix ambiguous tone() binding when adding in a duration
+void tone(uint8_t _pin, int frequency, unsigned long duration) {
+  tone(_pin, frequency, duration);
```
d-a-v: Correct me, I think this one is calling itself.

earlephilhower: D'oh!  I've been away from the keyboard too long.  Will fix that.

cranphin:
Eep, looks good to me, I only read the code, so nothing I can test :)

jandrassy:
my test run OK

earlephilhower:
Thanks for the update, all.

uzi18:
@earlephilhower what is PWMRANGE value?

uzi18:
But where is it definied?

jandrassy:
in Arduino.h
but use analogWriteResolution() if you want to change the range

Issue #4876, opened 2018-07-02
OTA succeeds with a different flash mode, then kills the module


teejaydub:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [n/a] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12F]
- Core Version:			[1eb0645dcb82e03e239e7932ad20802c769c89e0]
- Development Env:		[Arduino IDE]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[custom]
- Flash Mode:			[qio vs. dio]
- Flash Size: 		[4MB]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA vs. SERIAL]

### Problem Description

The symptom is that OTA update works normally, using `ESPhttpUpdate`:

```
      [begin] roundedSize:       0x00061000 (397312)
      [begin] updateEndAddress:  0x00300000 (3145728)
      [begin] currentSketchSize: 0x00062000 (401408)
      [begin] _startAddress:     0x0029F000 (2748416)
      [begin] _currentAddress:   0x0029F000 (2748416)
      [begin] _size:             0x00060780 (395136)
      MD5 Success: f3a150dffd2ddd1136011efe3c483f4e
      Staged: address:0x0029F000, size:0x00060780
      Firmware update: OK.
      state: 5 -> 0 (0)
      rm 0
      pm close 7
      del if0
      uslc_ÇrSöf
```
and then after reboot I see:

```
       ets Jan  8 2013,rst cause:2, boot mode:(3,6)

      load 0x4010f000, len 1384, room 16 
      tail 8
      chksum 0xef
      csum 0xef
      csum err
      ets_main.c 
```
And the module stops.

The conditions under which this occur are:

* Recently switched to a new batch of modules for a new manufacturing run
* Had discovered that the new modules only support flash mode DIO, while the old ones supported QIO (the default).
* Was now using `--flash-mode dio` on the `esptool.py` command line to program modules initially, with binaries built as before.
[...]

PR #4877, opened 2018-07-02
Copy the existing flash mode over the one set in an OTA update image.

teejaydub:
Suggest fix for #4876: tolerate different flash modes in OTA update images and convert them to match what the module is currently using.

Allows the flash mode to be ignored during OTA, so the same binary image can be used for both modules that support QIO and those that require DIO.


devyte:
I'm ok with this.
@igrr any thoughts on this one?

teejaydub:
Thanks, guys, I appreciate it!

robertpoll:
Hi. This change has caused some issues for us so just wanted to discuss. We buy modules from AI-Thinker with a simple loader firmware that we've written loaded by them when they manufacture the modules. Once they are assembled into our product the loader firmware gets a certificate, stores in SPIFFS and then does an OTA update to load the real firmware. The problem is that AI-Thinker is insisting that the firmware they load is DOUT, but our final firmware is QIO. Obviously this change means our final firmware runs DOUT. 

I think the change is a good one but was wondering whether an override option would be a sensible addition for people in our position (which I would assume would affect quite a few people)?

teejaydub:
Good to know - two clearly valuable use cases that need opposite behavior. So, it could be on a compile-time switch.

Although it seems like the underlying bug might be that AI-Thinker is incorrectly labeling your firmware as DOUT when it's QIO, right? But you still need a stopgap solution for any existing modules in inventory.

PR #4888, opened 2018-07-04
Eliminate gigantic blocking time inside ESP8266WiFiMulti::run

braddr:
I'm using an esp8266 + Arduino in a project and need to keep the loop latencies low (low 10's of ms).  Using ESP8266WiFiMulti connections result in multi-second pauses due largely to the connection loop inside run().

This change (developed and tested against 2.4.1 but looks to apply to head of master cleanly):

1) pulls a couple functions out of the long run(), and
2) introduces a new state variable to track an in-progress connection attemp

This takes the latency down to only once hitting a wifimulti.run() latency > 10ms, and that at only 14ms.


braddr:
For what it's worth, I'm still seeing a single 880+/- 5 ms latency that's occurring between loop() calls that I haven't identified the source of yet.  It happens between the wifi.begin() and the completion of the connection.  Between the 0->2 and 2->3 state transitions:

state: 0 -> 2 (b0)
892 - loop
state: 2 -> 3 (0)

braddr:
The failing sub-build:

core/test_md5builder.cpp:65: FAILED:
  REQUIRE( builder.toString() == "bc4a2006e9d7787ee15fe3d4ef9cdb46" )
with expansion:
  1 == "bc4a2006e9d7787ee15fe3d4ef9cdb46"

Am I going blind or aren't those two strings the same, just like the require wants them to be?

d-a-v:
This CI bug is yet unsolved and happens from time to time. I retriggered.

beegee-tokyo:
Like this! 
I had problems with WiFiMulti blocking my system when the connection got lost. Your solution is much better.

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.h
```diff
@@ -47,6 +47,8 @@ struct WifiAPEntry {
 
 typedef std::vector<WifiAPEntry> WifiAPlist;
 
+struct BestNetwork;
```
devyte: If selectBest is removed from the class and made a static free standing function, there is no need to declare this here, because the definition in the .cpp is enough. Please remove.

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.h
```diff
@@ -58,6 +60,12 @@ class ESP8266WiFiMulti {
 
     private:
         WifiAPlist APlist;
+        unsigned long connect_timeout_ms;
+
+        void selectBest(BestNetwork& best, int8_t scanResult);
+        void handleConnectComplete(wl_status_t status);
```
devyte: This is a debug method, and there is also no need for it to be part of the class, or exported in the .h. Please remove and make the function a static free standing function in the .cpp.

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
```diff
@@ -38,8 +38,104 @@ bool ESP8266WiFiMulti::addAP(const char* ssid, const char *passphrase) {
     return APlistAdd(ssid, passphrase);
 }
 
+void ESP8266WiFiMulti::handleConnectComplete(wl_status_t status) {
```
devyte: This function has no dependencies on other class members or methods, and is not public. Please make it a static free standing function.

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
```diff
@@ -38,8 +38,104 @@ bool ESP8266WiFiMulti::addAP(const char* ssid, const char *passphrase) {
...
+wl_status_t ESP8266WiFiMulti::connectLoop() {
+    wl_status_t status = WiFi.status();
+    // wait for connection, fail, or timeout
+    if (status == WL_CONNECTED      ||
+        status == WL_NO_SSID_AVAIL  ||
+        status == WL_CONNECT_FAILED ||
+        (millis() > connect_timeout_ms))
+    {
+        handleConnectComplete(status);
+        connect_timeout_ms = 0;
```
devyte: This is a constructor-inited member, yet it is cleared here. Will it never be used again? Is there a need to init it again?

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
```diff
@@ -38,8 +38,104 @@ bool ESP8266WiFiMulti::addAP(const char* ssid, const char *passphrase) {
...
+    }
+
+    return status;
+}
+
 wl_status_t ESP8266WiFiMulti::run(void) {
 
+    if (connect_timeout_ms != 0) {
```
devyte: The timeout value is cleared above and never set again. Is that correct behavior?

earlephilhower:

devyte:
Closing in favor of #7619 , which should be equivalent.

Issue #4779, opened 2018-05-31
WPS causes wdt reset


d-a-v:
Tracking the WPS issue:
ref: https://github.com/espressif/ESP8266_NONOS_SDK/issues/130
cause: PR[#4622](https://github.com/esp8266/Arduino/pull/4622#issuecomment-392303906)
discussion: [bearssl#2](https://github.com/earlephilhower/bearssl-esp8266/issues/2)
fix: #4889

PR #4889, opened 2018-07-04
optionally allow WPS

d-a-v:
* Disable WPS at compile time
* At the price of 4K in heap, allow WPS by using a `boards.txt.py` option
* documentation in FAQ

ref: #4779

### DIFF
--- a/cores/esp8266/core_esp8266_main.cpp
+++ b/cores/esp8266/core_esp8266_main.cpp
@@ -156,10 +156,40 @@ void init_done() {
  * Peripherals (except for SPI0 and UART0) are not initialized.
  * This function does not return.
  */
+/*
+   A bit of explanation for this entry point:
 
-/* provided or not by boards.txt */
-/* #define NO_EXTRA_4K_HEAP */
+   SYS is the RTOS task/context used by the upperlying system to run its
+   administrative tasks (at least WLAN and lwip's receive callbacks and
+   Ticker).  NONOS-SDK is designed to run users's non-threaded code in
+   another specific task/context with its own stack in BSS.
 
+   Some clever fellows found that the SYS stack was a large and quite unused
+   piece of ram that we could use for the user stack, and proposed to use it
+   to store the user stack instead of using the users' main memory, thus
+   saving around 4KB or ram/heap.
+
+   A problem arose later, which is that this stack is heavily used by the
+   system for some features.  One of these features is WPS.  We still don't
+   know if other features are using this, or if this memory is going to be
+   used in future releases.
+
+   WPS beeing flawed by its poor security, or not beeing used by lots of
+   users, it has been decided that we are still going to use that memory for
+   the users's stack and disable the use of WPS, with an option to revert
+   that back at the user's discretion.  This selection can be done by using
+   the board generator script.  This could be also done by setting up a new
+   option in the IDE's Tools menu.
+
+   The behavior is controlled by the globally NO_EXTRA_4K_HEAP define below.
+
+   References:
+   https://github.com/esp8266/Arduino/pull/4553
+   https://github.com/esp8266/Arduino/pull/4622
+   https://github.com/esp8266/Arduino/issues/4779
+   https://github.com/esp8266/Arduino/pull/4889
+
+*/
 
 #ifdef NO_EXTRA_4K_HEAP
 /* this is the default NONOS-SDK user's heap location */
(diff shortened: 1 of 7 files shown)


tools/sdk/include/user_interface.h
```diff
@@ -575,9 +575,16 @@ enum wps_cb_status {
...
 bool wifi_wps_disable(void);
 bool wifi_wps_start(void);
+#else
+bool WPS_is_unavailable_in_this_configuration__Please_check_FAQ_or_board_generator_tool ();
+#define wifi_wps_enable(...) WPS_is_unavailable_in_this_configuration__Please_check_FAQ_or_board_generator_tool()
+#define wifi_wps_disable() WPS_is_unavailable_in_this_configuration__Please_check_FAQ_or_board_generator_tool()
+#define wifi_wps_start() WPS_is_unavailable_in_this_configuration__Please_check_FAQ_or_board_generator_tool()
+#endif
 
 typedef void (*wps_st_cb_t)(int status);
```
devyte: Should these 2 lines be moved into the #ifdef ?

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -156,11 +156,25 @@ void init_done() {
  * Peripherals (except for SPI0 and UART0) are not initialized.
  * This function does not return.
  */
+
```
devyte: Please add a descriptive comment about what the #define means (explain it's either WPS or stack optimization, but not both), or at least reference the issue.

doc/faq/readme.rst
```diff
@@ -40,6 +40,22 @@ entering an issue report, please perform initial troubleshooting.
...
+Using ``*printf()`` with floats is enabled by default.  Some KBs of flash can
+be saved by using the option ``--nofloat`` with the boards generator:
+
+``./tools/boards.txt.py --nofloat --allgen``
+
+Why can't I use WPS ?
+~~~~~~~~~~~~~~~~~~~~~
+
+WPS is disabled by default, this offers an extra 4KB in ram/heap.  To enable
+WPS, use this boards generator option:
```
devyte: Please state explicitly that enabling WPS removes the 4KB heap optimization.

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -571,6 +571,8 @@ int32_t ESP8266WiFiSTAClass::RSSI(void) {
 // -------------------------------------------------- STA remote configure -----------------------------------------------
 // -----------------------------------------------------------------------------------------------------------------------
 
+#ifdef NO_EXTRA_4K_HEAP
```
devyte: Please add a descriptive comment about what the #define means, or at least reference the issue.

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -92,7 +93,13 @@ class ESP8266WiFiSTAClass {
 
     public:
 
+#ifdef NO_EXTRA_4K_HEAP
```
devyte: Please add a descriptive comment about what the #define means, or at least reference the issue.

tools/sdk/include/user_interface.h
```diff
@@ -575,9 +575,16 @@ enum wps_cb_status {
     WPS_CB_ST_UNK,
 };
 
+#ifdef NO_EXTRA_4K_HEAP
```
devyte: Please add a descriptive comment about what the #define means, or at least reference the issue.

@igrr In this case, should the change be moved into a patch that gets applied over the original user_interface.h file? I'm thinking of migration to future sdk versions, and what will happen with our own changes to this file.

devyte:
@d-a-v this is pretty good. My sole requests is to add comments to explain what it means. I'm thinking of the future when we come back and look at this, to avoid head scratching/WTFs and so on.

cores/esp8266/core_esp8266_main.cpp
d-a-v: Would this (above, below) be a clear enough explanation with its references ?

doc/faq/readme.rst
```diff
@@ -40,6 +40,22 @@ entering an issue report, please perform initial troubleshooting.
...
+Using ``*printf()`` with floats is enabled by default.  Some KBs of flash can
+be saved by using the option ``--nofloat`` with the boards generator:
+
+``./tools/boards.txt.py --nofloat --allgen``
+
+Why can't I use WPS ?
+~~~~~~~~~~~~~~~~~~~~~
+
+WPS is disabled by default, this offers an extra 4KB in ram/heap.  To enable
+WPS (and loose 4KB of useable ram), use this boards generator option:
```
devyte: loose -> lose

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -156,11 +156,62 @@ void init_done() {
  * Peripherals (except for SPI0 and UART0) are not initialized.
  * This function does not return.
  */
+/*
+   A bit of explanation for this entry point:
+
+   SYS is the RTOS task/context used by the upperlying system to run its
```
devyte: RTOS task/context -> SDK task context

There is confusion about RTOS and NONOS, I'd prefer to not add to that.

PR #4890, opened 2018-07-04
ScheduledInterrupts should still use current schedule_function(..)

hreintke:
@devyte 
The ScheduledInterrupt PR uses the new ScheduledFunctions to do the actual scheduling.
But.. that is not yet integrated into the Schedule.h -> The attached function is never called.

This PR fixes this by using the current schedule_function(..).


devyte:
This looks good to me.

PR #4894, opened 2018-07-05
clear mdns.queryservice()'s previous result when called

a-c-sreedhar-reddy:


libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -289,7 +289,15 @@ int MDNSResponder::queryService(char *service, char *proto) {
 #ifdef DEBUG_ESP_MDNS_TX
   DEBUG_ESP_PORT.printf("queryService %s %s\n", service, proto);
 #endif  
-  
+  MDNSAnswer *answer;
+  int numAnswers = _getNumAnswers();
+  for (int n = numAnswers - 1; n >= 0; n--) {
```
devyte: Answers is implemented as a linked list. This implementation will search the linked list from front to back for each iteration in order to find the last one for removal, which is an exponential algorithm.
Could you please reimplement the clearing of the answers by just iterating through the list and freeing each?

PR #4640, opened 2018-04-13
Support multiple tone(), analogWrite(), and Servo

earlephilhower:
-edit 5/14 to remove WIP -

Remove and rewrite all the parts of the core/libraries using TIMER1
and consolidate into a single, shared waveform generation interrupt
structure.  Tone, analogWrite(), Servo all now just call into this
shared resource to perform their tasks.

This setup enables multiple tones, analogWrites, servos, and
to be concurrent with reasonable accuracy.  It uses both TIMER1
and the internal ESP cycle counter to handle timing of waveform edges.
TIMER1 is used in non-reload mode and only edges cause interrupts.  The
interrupt is started and stopped as required, minimizing overhead when
these features are not being used.

A generic "startWaveform(pin, high-US, low-US, runtime-US)" and
"stopWaveform(pin)" allow for further types of interfaces.

Using a simple USB DSO shows that with only a single channel running,
a "tone(1000)" generates a waveform of 999.8Hz (500.1us per high/low).
Running additional generators or steppers will potentially increase the
jitter when multiple edges need to be generated in close proximity (.5-1us).

PR #4895, opened 2018-07-05
allow analogWrite to pin 16

liebman:
Was broken here  #4640


earlephilhower:
They say there's only two really difficult problems in computer science:  cache coherency, naming, and off-by-one errors.

Thanks for the fix to the silly bug!

PR #4901, opened 2018-07-08
Fix minor BearSSL API issues

### LINKED ISSUE #4882: BearSSL bugs found by @jeroen88
@jeroen88 has been using BearSSL and adding some PRs and discovered a few issues that he's sent in email that I want to make sure get tracked here as I'm still kind of swamped.  @jeroen88, please add to this if I missed something:
1. The set{insecure,fingerprint,etc.} methods should clear any other authentication methods.  Presently, if you setFingerprint then setInsecure, for example, it will have both flags set and actually check the FP and not be insecure as the last call requested.
2. Shouldn't _sc be tested for nullptr before &_sc->eng in  WiFiClientSecure::_connectSSL(const char* hostName)
3. _waitForHandshake() can cause a hang if the server times out, and needs a WDT feed call in the while() loop somewhere

earlephilhower:
Fixes #4882 and updates GitHub certificate fingerprint to the current one
in BearSSL_Validation example.

When setting a authentication mode or stopping, clear all others out in case
the object is being re-used.

Add in a yield during the SSL handshake to allow a graceful timeout and not
a WDT error when the remote server hiccups.


Jeroen88:
@earlephilhower , the clearance of the parameters for reuse of the client looks good! Maybe you should add a warning that the fingerprint in the example might be outdated and how to find the present one? I will try asap to force a server timeout to check the optimistic_yield.

Jeroen88:
@earlephilhower I just checked and forced a timeout on my server while doing a secure POST with ESP8266HTTPClient. The client did not crash and returned a HTTPC_ERROR_READ_TIMEOUT as it should!

earlephilhower:
Thanks for the update, @Jeroen88!

earlephilhower:
Oops, hit wrong button there.

PR #4913, opened 2018-07-10
Avoid unnecessary waveform de-initialization 

shimarin:
which harms frequent digitalRead/Write like SoftwareSerial.

### DIFF
--- a/cores/esp8266/core_esp8266_waveform.c
+++ b/cores/esp8266/core_esp8266_waveform.c
@@ -205,7 +205,7 @@ int startWaveform(uint8_t pin, uint32_t timeHighUS, uint32_t timeLowUS, uint32_t
 // Stops a waveform on a pin
 int stopWaveform(uint8_t pin) {
   for (size_t i = 0; i < countof(waveform); i++) {
-    if (((pin == 16) && waveform[i].gpio16Mask) || ((pin != 16) && (waveform[i].gpioMask == 1<<pin))) {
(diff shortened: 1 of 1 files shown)


earlephilhower:
Looks good to me, thanks!  Like I was mentioning in the original discussion, we may want to remove the `stopWaveform` call completely from `digitalRead` as it feels like anyone doing a read on a pin playing an output should expect to get garbage, anyway.

Also, we could move the `.enabled` check to the beginning of the condition since it evaluates LTR and it's not often that pins are actually playing a tone.  Would save some cycles in the digitalWrite case as well.

We could also set up an inverse map and make this O(1) for pin->waveform entry.  No for needed then.

If the SWSerial still has issues, let's try some of these.

shimarin:
After posting PR,  I noticed that by adding 

```
if (!timerRunning) return false;
```

on the very start of ```stopWaveform``` helps to avoid searching the table everytime ```digitalRead``` is called entirely. (I assume the timer always is running during any of pins use waveform)

Just removing ```stopWaveform``` from ```digitalRead``` still is a reasonable option, though.

earlephilhower:
@shimarin Can you please give PR #4920 a look?  I included your very good suggestions and also an optimization for when there actually is a PWM running so it may help you even more with SoftwareSerial and others.  Thanks!

shimarin:
@earlephilhower looks good!

Issue #2677, opened 2016-11-09
SPI methods are inconsistent with Arduino AVR SPI


Makuna:
### Basic Infos
APIs across platforms should try to be consistent so that library writers don't have to pepper their code with ifdefs and custom calls for standard uses.
The Esp8266 SPI has method names for functions that are inconsistent with AVR SPI but are functionally the same.

on AVR 
`void transfer(void *buf, size_t count) ` 
is the same as on Esp8266
`void writeBytes(uint8_t * data, uint32_t size);`

mkeyno:
sorry to be picky but what's big problem with that ?  writeBytes much faster than normal transfer function 

https://github.com/esp8266/Arduino/issues/2624

Makuna:
The standard Arduino SPI library does not have a `writeBytes`, it does have a `transfer` that takes a byte pointer and count of bytes.  The Esp8266 SPI library does not expose this transfer and further adds its own `writeBytes`.  This is inconsistent and makes writing Arduino code across platforms harder.

I am not saying there is a need to remove writeBytes; but at least expose the standard of a transfer that takes a buffer and count of bytes.

selste:
According to the writeBytes method comment the data has to be aligned to 32 bits - so it differs significantly from the transfer method that ships with the Arduino.
Would be great if the implementations would provide the same functionality.

PaulStoffregen:
My improved performance Ethernet library doesn't compile on ESP8266, only because SPI.transfer(data, size) is missing.

Here's a quick implementation I added in SPI.h for testing.  Maybe something like this could be merged, so Arduino sketches and libraries using this well established Arduino API can compile and run on ESP8266?

  void transfer(uint8_t * data, uint32_t size) {
        uint8_t buf[40];
        while (size > sizeof(buf)) {
                memcpy(buf, data, sizeof(buf));
                transferBytes(buf, data, sizeof(buf));
                data += sizeof(buf);
                size -= sizeof(buf);
        }
        if (size) {
                memcpy(buf, data, size);
                transferBytes(buf, data, size);
        }
  }

d-a-v:
The above comments are right, `writeBytes()` does the job without copy, using 32 bits registers, but require 32bits aligned buffers.
We can arrange something for the non aligned bytes and use this function.
Could `SPI.transfer()` take a `const uint8_t*` ?

off-topic:
[...]

PR #4925, opened 2018-07-13
arduino API's SPI::transfer(void*, size) (optimized)

d-a-v:
ref: #2677 
`SPI::transferBytes_()` needs 32bits optimization in its input part, like the one in its output part (done)


libraries/SPI/SPI.cpp
```diff
@@ -511,6 +530,14 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
...
+/**
+ * Note:
+ *  in and out need to be aligned to 32Bit
+ *  or you get an Fatal exception (9)
+ * @param out uint8_t *
+ * @param in  uint8_t *
+ * @param size uint8_t (max 64)
+ */
 void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
     while(SPI1CMD & SPIBUSY) {}
```
earlephilhower: I suggest an ASSERT() here checking the alignment of the inputs would be handy here and print a useful error when debugging was enabled...

devyte:
Assert and optimization will be tracked with an issue. Merging this.

PR #4931, opened 2018-07-16
sd: implement readBytes

icosaeder:
There is a speed issue when sending the files from SD card over HTTP.
The problem is, that the SD-specific class `File` inherited from `Stream` does not override the 
`virtual size_t readBytes(char *buffer, size_t length)`
method, while this method is used under the hood of `ESP8266WebServer::streamFile`:

```
Stream::readBytes
BufferedStreamDataSource::get_buffer
ClientContext::_write_some
ClientContext::_write_from_source
ClientContext::write
WiFiClient::write
ESP8266WebServer::streamFile
```

The default implementation of `readBytes` in the `Stream` class does not actually use buffered reading, just a loop reading bytes one by one (through `Stream::timedRead()` method), which is obviously slow.

The proposed fix redirects `readBytes` to buffered read of SD-specific File class. 

This speeds up the `ESP8266WebServer::streamFile` more than 3 times. Tested on streaming the 800+ Kb file from SD (FAT32), average time without a fix was 9000 ms, with the fix is 2600 ms (maximal possible SPI speed used), which is as fast as streaming the same file from internal SPIFFS. Hardware: WeMos D1 mini.


devyte:
I have some concerns about the casts, but that is a matter for discussion, and likely eventual cleanup elsewhere.
=> approving.

icosaeder:
Thank you for approving :)

PR #4935, opened 2018-07-18
Expose post args during upload

luc-github:
Currently post args are only available at the end of upload but they are already listed - this PR just expose them with minimal changes
It also set a define for post args array size originaly set to 32
This is useful to add extra info for upload like file size or other info like file time etc...

FYR Same PR has been done to ESP32 https://github.com/espressif/arduino-esp32/pull/1650


luc-github:
I think `-Werror=reorder` should be part of platform.txt

PR #4936, opened 2018-07-18
Return error in class deepSleep() when max sleeptime is exceeded

marcvtew:
Proposed codechange returns 0 when max sleeptime  is exceeded.

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -107,11 +107,22 @@ void EspClass::wdtFeed(void)
 
 extern "C" void esp_yield();
 
-void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
+bool EspClass::deepSleep(uint64_t time_us, WakeMode mode)
 {
-    system_deep_sleep_set_option(static_cast<int>(mode));
-    system_deep_sleep(time_us);
-    esp_yield();
+    if (time_us > deepSleepMax())
+    {
+       #ifdef DEBUG_SERIAL
+          DEBUG_SERIAL.println("Error: max sleeptime exceeded");
+       #endif
+       return 0; // error: max sleeptime exceeded
+    }
+   else
+   {
+      system_deep_sleep_set_option(static_cast<int>(mode));
+      system_deep_sleep(time_us);
+      esp_yield();
+      return 1; // never gets called
+   }
 }
 
 //this calculation was taken verbatim from the SDK api reference for SDK 2.1.0.


cores/esp8266/Esp.cpp
```diff
@@ -107,11 +107,22 @@ void EspClass::wdtFeed(void)
...
-    system_deep_sleep(time_us);
-    esp_yield();
+    if (time_us > deepSleepMax())
+    {
+       #ifdef DEBUG_SERIAL
+          DEBUG_SERIAL.println("Error: max sleeptime exceeded");
+       #endif
+       return 0; // error: max sleeptime exceeded
+    }
+   else
```
devyte: Please remove else and braces.

marcvtew: You're putting the esp8266 in deepsleep **_before_** testing wether time_us > deepSleepMax(). Testing never gets done because deepsleep resets the esp8266 and code starts from skratch.

cores/esp8266/Esp.cpp
```diff
@@ -107,11 +107,19 @@ void EspClass::wdtFeed(void)
...
 
-void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
+bool EspClass::deepSleep(uint64_t time_us, WakeMode mode)
 {
+    if (time_us > deepSleepMax())
+    {
+       #ifdef DEBUG_SERIAL
+          DEBUG_SERIAL.println("Error: max sleeptime exceeded");
+       #endif
+       return 0; // error: max sleeptime exceeded
```
devyte: return type is bool, but returned value is int => return false here

cores/esp8266/Esp.cpp
```diff
@@ -107,11 +107,19 @@ void EspClass::wdtFeed(void)
...
+    }
     system_deep_sleep_set_option(static_cast<int>(mode));
     system_deep_sleep(time_us);
     esp_yield();
+    return 1; // never gets called
```
devyte: return type is bool, but returned value is int => return true here

cores/esp8266/Esp.cpp
marcvtew: You're right.
Strange though that this is not detected by git code check.

liebman:
I have a question.  I've noticed that the deepSleepMax() can vary.  For instance deep sleep an hour, take an action and then go back to sleep.  I print the deepSleepMax() value and it changes.  Does this value only change during a sleep or could it very between calling deepSleepMax() and then calling deepSleepMax()?  Asking because that could create a race condition even with the check being added here.

devyte:
I think it can only vary within the sys context, i.e.: between loop()s, or between a yield() and return. That means that it shouldn't vary between the call to deepSleepMax() and the actual sleep.
The above is an educated guess though. It's based on my experience with how the SDK works in general, and not on any deep knowledge of what goes on under the hood wrt sleep, so I could be wrong.

cores/esp8266/Esp.cpp
```diff
@@ -107,11 +107,19 @@ void EspClass::wdtFeed(void)
 
 extern "C" void esp_yield();
 
-void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
+bool EspClass::deepSleep(uint64_t time_us, WakeMode mode)
 {
+    if (time_us > deepSleepMax())
```
earlephilhower: Would it not be more useful (and help fix peoples code which is broken already) if, instead of possibly printing an error and `return false;` , you just printed a *warning* and adjusted the time down to `deepSleepMax()`? 

Existing code will not expect to ever get a return value and potentially go off into lala-land, since this was a `void`.

"Be conservative in what you do, be liberal in what you accept from others."

marcvtew: @earlephilhower Point taken. I'll make some changes to the code.

cores/esp8266/Esp.cpp
```diff
@@ -109,6 +109,13 @@ extern "C" void esp_yield();
 
 void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
 {
+    if (time_us > deepSleepMax()) // we need to prevent the esp8266 from not waking up from deepsleep
+    {
+       time_us = (deepSleepMax() - (round(deepSleepMax() * 5 / 100))); // 5% correction because of inaccurate timekeeping by the esp8266
```
earlephilhower: No need for round() here.  You're doing integer division already, so this takes the integer, promotes it to a double, then promotes deepSleepMax() to a double, subtracts them as doubles, then truncates it all back to an integer.  A shorter way which won't cause a program to include the whole floating point libs (if not being included already):
`time_us = (deepSleepMax() * 95L) / 100L;`
or if time_us can be > 2^24 (so the *95 but could overflow)
`time_us = deepSleepMax() - (deepSleepMax() / 20);`

I bring this up because there was much howling of pain when we included floating point support in printf with the latest release, as some folks binaries ended up being just a hair too big for OTA...

devyte:
@marcvtew did you actually encounter a case where the deepsleep would get stuck with a value close to the max?
I'm not an expert in this, but my understanding is that the max value is related to the max number that can be put in some timer without overflowing it. If it does overflow, then behavior is undefined. I say related because there is another internal factor that gets updated every so often.
What I mean is that, in theory, sleeping for max time should always be possible. If it's not, then that should be documented and reported to Espressif. 
Or is this 5% meant to be more accurate wrt how long the ESP actually sleeps? If so, then most likely it should be shaved off in all cases, not just when the specified value is > max.

marcvtew:
@devyte Made some code to read and display maxSleeptime(). The esp8266 on my Lolin board has a max sleeptime of 13840678905 microseconds. I'll set my sleeptime to 13840678904 microseconds and see if it wakes up.

marcvtew:
@devyte I made some code to measure the value of maxDeepsleep() every two seconds, remembering the absolute max value and the absolute min value measured. I let the code run for about 15 minutes to get a stable result:
max deep sleep MAX: 14386462713 microseconds
max deep sleep MIN: 13704364025 microseconds
max deep sleep DIF: 682098688 microseconds

So, in my case the shortest calculated maxsleeptime differs more than 11 minutes (within max around 3,8 to around 4 hours deepsleep) from the longest calculated maxsleeptime. It looks to me that it's not sufficient to only test (and set) deepsleeptime once (i.e. before going into sleep) because during deepsleep the calculated maxsleeptime may become shorter (worst case: more than 11 minutes) resulting in a timer overflow (esp8266 never restarts allthough we passed if (time_us > (deepSleepMax())).

Safe deepsleep time for me seems to be13704364025 microseconds. With this sleeptime the esp8266 should always wake up. I'll test this tomorrow.
Esp.cpp needs an adjustment like this i think:
`if (time_us > (deepSleepMax() - 682098688 or % of deepSleepMax()) // choose one of the two methods
{
       time_us = deepSleepMax() - 682098688 or % of deepSleepMax()); // choose one of the two methods
}`

artua:
thanks for reference. waiting for your test results. I'm always using 4294967295 before. But it would be better to make it sleep even more.

cores/esp8266/Esp.cpp
```diff
@@ -109,6 +109,13 @@ extern "C" void esp_yield();
 
 void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
 {
+    if (time_us > deepSleepMax()) // we need to prevent the esp8266 from not waking up from deepsleep
```
earlephilhower: Hate to dogpile here, but if you do find out that `deepSleepMax()` doesn't really work and need to derate by 5% or whatever, please use that # to compare and not `deepSleepMax()` in this `if` condition.    Otherwise someone will pass in, say, `deepSleepMax() - 1` which won't work but which won't be caught by this check.

devyte:
@marcvtew after re-reading this, it occurs to me that I'm not sure I understand your intent. Do you want to assure a "safe" deepSleep() call to make sure that the ESP always wakes up? Do you want to find a safe lower bound to the max deepSleep() time? or something else?

marcvtew:
@devyte Testing wether a warning/error should be given only makes sense when we can be absolutely sure that our test produces the same result under all conditions. Last week i realized that deepSleepMax() is not a constant but a variable. So for my specific esp8266 i tested wether this variable stayed between an upper and a lower constant value. This seemed to be the case:

max deep sleep MAX: 14386462713 microseconds
max deep sleep MIN: 13704364025 microseconds

Based on this result i thought it might be possible to come up with a test based on these more or less constant upper and lower boundaries. 
However restarting the esp8266 and performing the same test again gave totally different values:

max deep sleep MAX: 13830193145 microseconds
max deep sleep MIN: 12875988986 microseconds

Based on these results i conclude that testing if time_us > deepSleepMax() only makes sense when the value of deepSleepMax (calculated before going into deepsleep) stays valid (i.e. constant) during deepsleep. To test this i adjusted my code for deepsleep to `ESP.deepSleep(ESP.deepSleepMax(), WAKE_RF_DEFAULT)`. If the value of deepSleepMax() never changes during deepsleep than with this code the esp8266 will allways wake up from deepsleep. On the other hand, if the value of deepSleepMax() does change during deepsleep there's the possibility that the value of deepSleepMax() needed to wake up the esp8266 is smaller than the calculated (and used) value of deepSleepMax() resulting in the esp8266 not waking up. The latter seems to be the case. This morning, after two succesful wakeups from deepsleep my esp8266 never woke up for the third time. For now my conclusion is that using a value calculated with deepSleepMax() does NOT mean the esp8266 will allways wake up from deepsleep.

cores/esp8266/Esp.cpp
d-a-v: Would it also be wise to increasingly loop the whole function until users's `time_us` is reached ?

devyte:
@marcvtew that's precisely why I asked about your trying to find boundaries. I suspect that the rtc clock changes based on several variables including temperature.

The deepSleepMax shouldn't change while the ESP is sleeping.
Per Espressif, ESP.deepSleepMax() is the max theoretical sleep time, and should always work, at least in their context of usage (i.e.: their example apps, where there is only a sys context and no cont context).
Ref: [NONOS SDL api reference doc](https://www.espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf) section 3.3.9.

Which makes me think of a race condition before actually going to sleep. I see 2 possibilties:
1. Section 3.3.9, in Note, 3rd bullet says:
> After configuration, the chip will not enter Deep-sleep mode immediately, but will wait for a while till the Wi-Fi core is closed safely.
Is it possible that during this wait time the max value changes? If so, would shutting down wifi before deepsleeping help? If that does not help, then I think the formula to calculate deepSleepMax is useless and there isn't much we can do but use a very low "safe" value.
2. I see [this](https://github.com/esp8266/Arduino/blob/75a6a3f0730a2c20af7b444f2e0044e39e77607c/cores/esp8266/Esp.cpp#L110):
```
void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
{
    system_deep_sleep_set_option(static_cast<int>(mode));
    system_deep_sleep(time_us);
    esp_yield();
}
```

That yield forces a context switch back to the sys context. Could that somehow allow the deepSleepMax to change before sleeping?

devyte:
Also, it's always possible that the [deepSleepMax ](https://github.com/esp8266/Arduino/blob/75a6a3f0730a2c20af7b444f2e0044e39e77607c/cores/esp8266/Esp.cpp#L119) formula is wrong. That was implemented from the [NONOS SDK api reference](https://www.espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf) section 3.3.9 explanation.

5chufti:
some observations:
cali (a.k.a. system_rtc_clock_cali_proc(), base for calculating deepSleepMax()) varies not only during loops but also on consecutive reads in same loop. 
It also varies between wake_ups after deepsleep. 
after new flash it is rather high (0x6079), dropping and stabilising with time (0x5c8*, looks like chip gets hot during flash)
it also drops/rises with module temperature.
same behaviour is true even with wake_rf_disabled
which makes me wonder: what is the timer calibrated against???
how do they determine the timer periode?

for this specific module, with no ambient temperature change, the safe max deep_sleep time in seconds can be approximated as tmax < cali (crediting d-a-v https://github.com/esp8266/Arduino/issues/4969#issuecomment-409490760)

marcvtew:
@5chufti In my opinion `max_us_sleeptime < system_rtc_clock_cali_proc() * 1000000` does not neccesarily mean that the esp8266 will always wake up from deepsleep. Using this calculation means that we are correcting with `((2^32-1)/2^12)/1000000)` = less than 5%. In my tests so far i've seen deviations of more than 5% measuring system_rtc_clock_cali_proc().
Besides, as @devyte mentioned, the esp8266 does not immediately go into deepsleep. Some time after execution of the deepsleep code is needed by the esp8266 to actually go into deepsleep. During this time the value of `system_rtc_clock_cali_proc()` may change. This might make our test for safe max deepsleep time unreliable.

5chufti:
IMHO the sleeptime will be transformed by the api call to a simple timer value loaded into the wake-up timer. So as long as the value recalculates to a valid 32bit value the sleep and wake-up will work as intended. As I mentioned, after an initial phase of steady decrease of cali, it settles within a range of correctjy recalculated some ten seconds. If one uses the tmax(s)<cali one is allways on the safe side given the cali reading and deepsleep is subsequent so as not to allow for major temp drift.
I had no problems entering deep-sleep with delay(10) or less, so that is "marginal". If you want to avoid this, use system_deep_sleep_instant() directly. 
If your esp didn't wake up from deep_sleep I doubt it has been because of sleeptime overflow.

marcvtew:
@5chufti Looking closer to the code of deepSleepMax() i noticed that the calculation is being/must be done with `(2^31-1)/2^12)` and not with `(2^32-1)/2^12)`. For the safe max deepsleep time in seconds `(2^31-1)/2^12)` indicates an approximation of tmax < 0.5 * cali. `(2^32-1)/2^12)` indicates an approximation of tmax < 1.0 * cali as suggested at #4969.

d-a-v:
True. datasheet tells 2^31-1.
So we would have these equiv:
* `(time_in_us / cali) << 12	= 2^31 - 1` (datasheet)
* `time_in_us < (cali << (31-12))`
* `time_in_us < (cali<<19)`
* `time_in_s * 1000000 < cali * 524288`
* `time_in_s < cali * 0.524288`

implies
* `time_in_s < cali/2`

But (datasheet):

> the cali is the RTC clock period (in us); bit11 ~ bit0 are decimal.

So we need either interpret this decimal part (datasheet gives the way to interpret it, it gives a maximum of 1000s / 16mn40s - not 100% sure about this)), or remove it:

* `time_in_s < ((cali & ~((1<<12)-1)) / 2)`, same as
* `time_in_s < ((cali >> 12) << 11)`

My board gives 10240s with this formula (2h50mn). I remember having seen it fluctuating a bit.

5chufti:
for me this is purely academic, I cannot see any use in ESP.deepSleep(ESP.deepSleepMax());
For a start I would have expected the API call to return false if the max_deep_sleep_time is exceeded, but I forgot we are dealing with espressif ...

So the decent thing to do would be to either 
a) replace the passed value with tmax if exceeded 
    (I think this is what API does instead of returning false)
b) return with "false" if tmax exceeded

just for the unlikely case that someone really exeeds tmax for ESP.deepSleep();
They should beforhand check on the use and probable range in documentation and/or div forum posts. I don't even see need of deepSleepMax().

regarding 2^31 vs. 2^32 see here https://github.com/esp8266/Arduino/issues/4969#issuecomment-410188697 (or 3.3.9 vs 3.3.50)

artua:
@5chufti for me, knowing the deepSleepMax() is about to save battery on
some applications. For now, I wake ESP once a hour with modem sleeping,
check the sleep cycles left and put it to deepSleep next cycle. But anyway,
it’s ~30mA consumption while I’m checking these cycles.
So it’s not academic, it’s purely practical question.

d-a-v:
> check the sleep cycles left

How are you doing that ?

5chufti:
@artua 
still I wouldn't use deepSleepMax() but a value that I can deal with, like 9000s (2.5)h.
without any true reference like ntp or rtc you'll never get repeatable cycles anyhow.

marcvtew:
I'm using my esp8266's for IoT purposes, mostly battery powered. For me not knowing the safe max sleeptime makes the esp8266 totally useless.

5chufti:
thats why I argue for only one function, returning a safe value.
Who needs a function returning a value "without any value" (not being reliable)?

marcvtew:
@5chufti I agree that it's best to have only one meaningful function. However ths might mean that we will be deviating from the api specification of deepSleepMax().
I've tested ESP.deepSleep(ESP.deepSleepMax(), WAKE_RF_DEFAULT) twice. Both times the esp8266 eventually did NOT wake up from deepsleep anymore. Is it possible for you to check my results?

d-a-v:
Since we are still speaking in uS, and bits 0..11 are BCD and(?) we don't wish to deal with them, we could use:
`time_in_us < ((((uint64_t)system_rtc_clock_cali_proc()) >> 12) << (12+19))`

artua:
@5chufti I'm ok with that. But how can I know the max safe value? If the
deepSleepMax() will return any safe constant - it would be great.
Let it be deepSleepSafe() or anything like that.

5chufti:
@d-a-v 
BCD? any confidential info? not even the rtc "example" in A.2 does suggest that?

@marcvtew 
```
SDK:2.2.1(cfd48f3)/Core:win-2.5.0/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.1-13-g163bb82)/BearSSL:6d1cefc

# Connecting to WiFi bcn 0
del if1
usl
mode : sta(5c:cf:7f:80:d7:0c)
add if0
...............scandone
state: 0 -> 2 (b0)
.state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 3
cnt 

connected with WiFi, channel 1
dhcp client start
ip:192.168.22.136,mask:255.255.255.0,gw:192.168.22.123
# Setup done
# sntp ok after 19417 ms
# Fri Aug  3 15:04:27 2018
########################################
deep sleep too long
pm open,type:2 0
bcn_timout,ap_probe_send_start
ap_probe_send over, rest wifi status to disassoc
state: 5 -> 0 (1)
rm 0
pm close 7
scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 3
cnt 

connected with WiFi, channel 1
dhcp client start...
ip:192.168.22.136,mask:255.255.255.0,gw:192.168.22.123
pm open,type:2 0
bcn_timout,ap_probe_send_start
ap_probe_send over, rest wifi status to disassoc
state: 5 -> 0 (1)
rm 0
pm close 7
scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 3
cnt 

connected with WiFi, channel 1
dhcp client start...
ip:192.168.22.136,mask:255.255.255.0,gw:192.168.22.123
```

this is not looking good: 
with deepsleep > tmax the chip wakes up ~immediately but does not start up sketch
(output from sketch has preceding #).

5chufti:
@marcvtew 
unfortunately your right, no wake-up when tmax is exceeded.
I call this a severe API flaw, system_deep_sleep() is of type bool and should return false - when, if not at least when called with too high deepsleep time? 
but I really wonder: how did they do it, so it never wakes up again? 
If you load a timer register with a value exeeding the register width, one would usually assume the tineout would occur after the "modulo" ...

devyte:
@5chufti some timers don't continue to count past overflow, or have undefined behavior if they overflow. Another explanation is that some interrupt is triggered on overflow, but no ISR is assigned, or a garbage ISR is assigned.
Whatever else, I see two distinct usage cases here:
1. sleep for a specific time, which should always be "safe"
2. sleep for the max time possible, whatever that time is, even if that max time changes over time

The key point here is that neither of the above cases is really addressed in the current Espressif SDK API or docs. 
- The formula for the max theoretical sleep time is useless, because cali could change between the calculation and sleep start. Per @5chufti it could change even between calls within the same loop(). I didn't observe this in my own tests, for me it was always constant within a loop, but could change between loops, or during a yield(), but then again I didn't exactly spend a lot of time on that testing.
Given that race condition, there is no way that 1. or 2. above can be done "safely".
- If the max time is exceeded, in some cases you get an error message on Serial, in other cases the ESP doesn't wake up. I suspect that the code that checks the sleep time on the SDK side suffers from the same race condition due to cali changing.

5chufti:
breaking news:
>system_deep_sleep_instant(time_us);

behaves much better as it at least restarts if tmax is exceeded.

@devyte 
we're talking about hw timer / compare registers, so no isr etc.

5chufti:
I now personally have settled for this approach
```
void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
{
    system_deep_sleep_set_option(static_cast<int>(mode));
    if (time_us & 0x0800000000ull)
	time_us = (uint64_t)system_rtc_clock_cali_proc() * 0x7A120ull;
    if (time_us & 0x1000000000ull)
    	system_deep_sleep(time_us);
    else
    	system_deep_sleep_instant(time_us);
    esp_yield();
}
```

now I can easily select the functionality I want with 
```
ESP.deepSleep(sleepTime)                         // backward compatible   
ESP.deepSleep(sleepTime | 0x0800000000ull);       // do max safe deepsleep
ESP.deepSleep(sleepTime | 0x1000000000ull);      // do instant deepsleep
ESP.deepSleep(sleepTime | 0x1800000000ull);      // do max safe instant deepsleep

```

devyte:
The current implementation uses system_deep_sleep(), which internally shuts down the wifi core "safely" (whatever that means), and _then_ goes to sleep.
I think that to get this to work as expeted, @5chufti is correct in that we should use system_deep_sleep_instant() to go to sleep immediately, which should remove the race condition.
My only worry is the meaning of shutting down the wifi core "safely", but there are several possibilities here:
1. going to sleep immediately has no impact on sleep power consumption, so we don't care about safe or non-sage
2. the user is responsable for shutting down wifi before going to sleep
3. we shut down wifi before calling these api calls, and hide that from the user

marcvtew:
The past 30 hours i've tested wether the esp8266 wakes up with deepSleepMax() = ((uint64_t)system_rtc_clock_cali_proc()/2*(0xF4240ULL)-1). My esp8266 restarted 11 times so using deepSleepMax() = ((uint64_t)system_rtc_clock_cali_proc()/2*(0xF4240ULL)-1) seems to always wake up the esp8266.

In one of the api docs i remember to have read that system_deep_sleep_instant() function is void. Is this function still being supported by Espressif?

I did some testing shutting down wifi before going into system_deep_sleep(). Made no difference: esp8266 did not wake up with time_us = deepSleepMax(). Maybe someone else is willing to test this as well?

PR #4939, opened 2018-07-19
Add valgrind and string tests to host_tests

earlephilhower:
Valgrind identified an error that was causing intermittent failures in
the host tests, and can check for subtle memory access and allocation bugs.
Add it to the standard host_test suite, and any errors will cause test
failure.

Also start adding string tests, since two undefined behaviors have been
found so far just by inspection.


earlephilhower:
The lcov stuff won't help the CI, but it does produce some very nice color-coded source HTML files that makes it way easier to see what's covered and what isn't.

earlephilhower:
screenshot from 2018-07-25 20-03-24

PR #4945, opened 2018-07-21
Fix waveform missing edges on cycle rollover

### LINKED ISSUE #4944: analogWrite is reset to the minimum or maximum value
#### Platform

- Hardware:				ESP-12E
- Core Version:			win-2.5.0-dev
- Operating System:		Windows

### Settings in IDE

- Module:  			Nodemcu 1.0
- Flash Mode:			?
- Flash Size: 		        4MB
- lwip Variant:                  v2 Lower Memory
- Reset Method:		nodemcu?
- Flash Frequency:	        40Mhz?
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:              256000

### Problem Description

AnalogWrite is reset to zero or maximum value. The plugged led gets stuck on the maximum PWM level or turns off. At the same time the buzzer continues to play sounds. Thus, it becomes clear that the module does not hang. To replicate the problem run the attached code and wait for a while. It may take a few hours. The problem is mentioned in #4640

### Sketch

```cpp

#define LED_PIN D1 // GPIO5
#define BUZ_PIN D2 // GPIO4
int bright;
bool increase;
int note;
unsigned long cur_millis = millis();
unsigned long prev_millis = cur_millis;

void setup() {
  Serial.begin(115200);
  Serial.print("\n\nSDK Version=");
  Serial.println(ESP.getCoreVersion());
  pinMode(LED_PIN, OUTPUT);
  pinMode(BUZ_PIN, OUTPUT);
  bright = 0;
  increase = true;
}

void loop() {
  cur_millis = millis();
  if (abs(cur_millis - prev_millis) > 100) {
    analogWrite(LED_PIN, bright);
    if (increase) {
      bright++;
    } else {
      bright--;
    }
    if (bright >= 105) {
      increase = false;
    } else if (bright <= 10) {
      increase = true;
    }
    prev_millis = cur_millis;
    tone(BUZ_PIN, (double)bright+200, 85);
  } 
}
```

### Debug Messages

```
SDK:2.2.1(cfd48f3)/Core:win-2.5.0-dev/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.1-10-g0c0d8c2)/BearSSL:94e9704


SDK Version=00000000
scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 4
cnt 

connected with HUT, channel 1
dhcp client start...
ip:192.168.1.7,mask:255.255.255.0,gw:192.168.1.1
pm open,type:2 0

```

earlephilhower:
When the ESP cycle counter rolls over, the "now" can be smaller than the
next-edge time of a waveform generator.  This would cause the edge to be
missed on that specific pin, and make it look like PWM was hung.

Use proper comparison between current time and edge time.

Fixes #4944

Also remove the "sigma-delta.c.unused" file which was replaced by a
working one some time ago.


earlephilhower:
Got a little over 8hrs recorded before the logic analyzer aborted.  Checked w/o any pwm freezes (on the same run I started about 24hrs ago) so I'm pretty sure we're clean.

PR #4961, opened 2018-07-26
change "WeMos" to new brand name "LOLIN"

wemos:


devyte:
@wemos can you please explain this? Did wemos and lolin merge? Did one buy the other? Were they always the same and nobody knew?

devyte:
Also, you have to edit boards.txt.py, regenerate boards.txt with it, and then submit both files.

PR #4964, opened 2018-07-27
fix digitalPinToBitMask(), portOutputRegister(), portInputRegister() …

wemos:
fix digitalPinToBitMask(), portOutputRegister(), portInputRegister() and portModeRegister() error when the pin is GPIO16.

### DIFF
--- a/cores/esp8266/Arduino.h
+++ b/cores/esp8266/Arduino.h
@@ -224,12 +224,13 @@ void loop(void);
 void yield(void);
 void optimistic_yield(uint32_t interval_us);
 
-#define digitalPinToPort(pin)       (0)
-#define digitalPinToBitMask(pin)    (1UL << (pin))
+#define _PORT_GPIO16    1
+#define digitalPinToPort(pin)       (pin==16)?(_PORT_GPIO16):(0)
+#define digitalPinToBitMask(pin)    (pin==16)?(1):(1UL << (pin))
 #define digitalPinToTimer(pin)      (0)
-#define portOutputRegister(port)    ((volatile uint32_t*) &GPO)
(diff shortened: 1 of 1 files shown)


cores/esp8266/Arduino.h
```diff
@@ -224,12 +224,13 @@ void loop(void);
 void yield(void);
 void optimistic_yield(uint32_t interval_us);
 
-#define digitalPinToPort(pin)       (0)
-#define digitalPinToBitMask(pin)    (1UL << (pin))
+#define _PORT_GPIO16    1
+#define digitalPinToPort(pin)       (pin==16)?(_PORT_GPIO16):(0)
```
earlephilhower: Sorry, I was just about to merge this but realized that all these macros have now lost their `(`..`)` parenthesis.  In the rest of this header all macros are completely enclosed in parens.  ex `#define bit` and `#define _min`.  The port*Register(port) macros all had them as well.

Can we bracket these ternary operators with parens, just to avoid any weirdness later on if they're used in operations and simply assigned to a value?

cores/esp8266/Arduino.h
```diff
@@ -224,12 +224,13 @@ void loop(void);
 void yield(void);
 void optimistic_yield(uint32_t interval_us);
 
-#define digitalPinToPort(pin)       (0)
-#define digitalPinToBitMask(pin)    (1UL << (pin))
+#define _PORT_GPIO16    1
+#define digitalPinToPort(pin)       ((pin)==16)?(_PORT_GPIO16):(0)
```
devyte: @earlephilhower request to surround the entire ternary is not done. This should be as follows:
` (((pin)==16)?(_PORT_GPIO16):(0))`

Same is valid for all new defines below.

PR #4965, opened 2018-07-27
Issues with LWIP

taubel:
I have experienced issues while developing a program with LWIP and wakaamaNode libraries. At some point the program tries to dereference and undefined element "eb" of a "pbuf" structure, which usually resolves to a very low address value, causing the program to crash. The following resolution seems to have fixed the issue, although I do experience problems with the device failing to respond to wifi frames after a random time period, while also printing an error message "no buf for action frame".

The issues are reproduced using the stock "liblwip_gcc.a" library. WakaamaNode can be found here: https://github.com/Openhab-Nodes/wakaamaNode


taubel:
I have noticed a similar fix already included in the lwip2 version, but the platformio script for Arduino framework has "-llwip_gcc" hard-coded, which makes portability an issue

devyte:
CC @d-a-v

d-a-v:
I can't really comment on this because since the beginning I had been forbidding myself to use the esp8266 because of this patched-lwip1.4's unstabilities. `pbuf->eb` is part of the espressif's patches to lwip about the wifi/radio packet header which encapsulate the ethernet frame. This lower layers encapsulation management has been improved so no more radio patches are needed within lwip-v2 (a macro defines the radio header length, `system_pp_recycle()` is called on received packets and that's pretty much it).


We may not be currently up-to-date with lwip-v1.4 sources regarding nonos-sdk, and we sometimes need to update them because lwip2 uses lwip-v1.4 structures - specifically the v1.4 side of the lwip2 glue -, and espressif sometimes change some critical definitions (last one was `lwip_err_t` some months ago).
So I don't see anything wrong with this PR but if we start to patch we have to keep trace of them for re-applying on upstream updates.

Although not beeing a platformio user, to my knowledge you can use lwip2 with it ([ref](https://github.com/platformio/platform-espressif8266/issues/65#issuecomment-383179529)).

This patch would better be proposed to nonos-sdk repository. See first if you are able to give lwip2 a try with platformio.

taubel:
I have switched to LWIP2 and this version seems to have already fixed the issue that I have proposed, thank you for your support.

PR #4975, opened 2018-07-29
Update I2SInput.ino

SConaway:


devyte:
In this case, I2SI  refers to the I2SI_ prefix for the pins, e.g.: I2SI_DATA, so the current text is correct.

SConaway:
Ok. Got it.

Issue #4176, opened 2018-01-16
ESP8266HttpClient never seems to finish large files when using lwip2 when also using websocket


schlaegerz:
#### Hardware
Hardware:		NodeMCU 1.0 (ESP-12E)
Core Version:      	2.4

### Description

I am trying to download a file that is around 10MB, and on lwip2 the file download seems to get around 1MB done and then just stops. If i go back to 1.4 then it finishes the download, but takes about 4min.

### Settings in IDE

Module:  			NodeMCU 1.0 (ESP-12E)
Flash Size: 		4M
CPU Frequency:		160Mhz
Flash Mode:			
Upload Using:		Serial

Code:

while (success && 
           http.connected() && 
            (downloadRemaining > 0 || downloadRemaining == -1) 
         ) {
        // get available data size
        
        auto size = stream->available();
        if (size > 0) {
            auto c = stream->readBytes(cur_buffer, ((size > CHUNK_SIZE) ? CHUNK_SIZE : size));
            cur_buffer = cur_buffer + c;
            auto totalSize = cur_buffer - buff;
            if (totalSize + 2* CHUNK_SIZE > buffSize)
            {
                cur_buffer = buff;
            }
             
            hasDownload = true;
            success &= c > 0;
            if (downloadRemaining > 0) {
                downloadRemaining -= c;
            }
            lastDownloadTime = millis();
        }
        yield();
        //delay(1);
        static int i = 0;
        i++;
        if (i % 2000 == 0)
        {
            Serial.println(downloadRemaining);
        }
    }

d-a-v:
Could you please post a full sketch so the issue can be reproduced ?

schlaegerz:
This sketch downloaded 2Mb for me and then stalled.
```

#include <Arduino.h>

#include <ESP8266WiFi.h>
#include <ESP8266WiFiMulti.h>

#include <ESP8266HTTPClient.h>

#define USE_SERIAL Serial

ESP8266WiFiMulti WiFiMulti;


#define CHUNK_SIZE HTTP_TCP_BUFFER_SIZE/2
uint8_t *buff;
const int buffSize = 4096;
void setup() {

    USE_SERIAL.begin(115200);
    // USE_SERIAL.setDebugOutput(true);

    USE_SERIAL.println();
    USE_SERIAL.println();
    USE_SERIAL.println();

    for (uint8_t t = 4; t > 0; t--) {
[...]

Issue #2771, opened 2016-12-17
SSL error 40, SSL handshake failure


ghost:
#### Hardware
Hardware:		Wemos D1 mini
Core Version:      	2.3.0-master

### Description

Simply unable to initiate the SSL handshake...

### Settings in IDE

Module:  			Generic ESP8266 Module
Flash Size: 		4MB
CPU Frequency:	80Mhz
Flash Mode:		qio
Flash Frequency:	80Mhz
Upload Using:		SERIAL
Reset Method:		nodemcu

### Sketch

```cpp

#include <ESP8266WiFi.h>
#include <WiFiClientSecure.h>

#define WLAN_SSID    "H********8"
#define WLAN_PASS    "T***********2"

#define SERVER_HOST  "apiv2.boltbutton.com"
#define SERVER_PORT  443

// WiFiFlientSecure for SSL/TLS support
WiFiClientSecure client;

//HTTPS Port fingerprint
const char* fingerprint = "‎07 59 1b 25 a7 36 7a 95 b0 ef 2d ae b2 58 d4 50 a0 89 d4 bb";

void setup() {
  Serial.begin(115200);
  delay(10);
  Serial.setDebugOutput(true);
 
  // Connect to WiFi access point.
  Serial.print("Connecting to ");
  Serial.println(WLAN_SSID);
  
  delay(1000);
  
  WiFi.begin(WLAN_SSID, WLAN_PASS);
  delay(2000);
  
  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
  }
  Serial.println("WiFi connected");
  Serial.print("IP address: "); Serial.println(WiFi.localIP());  
}

void loop() {
  
  const char* host = SERVER_HOST;
  const int httpPort = SERVER_PORT;
  
  // create a secure connection using WiFiClientSecure
  WiFiClientSecure client;
  
  if (client.connect(host, httpPort)) {
    
    // verify the signature of the ssl certificate
    if (client.verify(fingerprint, host)) {
      Serial.println("ssl cert matches");
    } else {
      Serial.println("ssl cert mismatch");
    }
  }
  else {
    Serial.println("Connect fail before verification");
  }
  
  delay(60000);
}

```

### Debug Messages

```
WiFi connected
IP address: X.X.X.X
:ref 1
please start sntp first !
State:	sending Client Hello (1)
:sent 72
:rn 7
:rcl
:abort
:rd 5, 7, 0
:rdi 7, 5
[...]

PR #4980, opened 2018-07-29
Feature/http update wi fi client parameter

Jeroen88:
Pass Client parameter to httpUpdate to use any BearSSL security method for a firmware update (with example). This PR replaces #4832.

This is a combination with PR #4979, because The ESP8266HTTPUpdate class uses the ESP8266HTTPClient class from this PR to pass the BearSLL WiFiClient.

Added two new update() and one updateSpiffs() function that uses the ESP8266HTTPClient adapted by me in PR #4825. With these adaptations, it is possible to do an encrypted and authenticated server firmware or SPIFFS update, using any of the supported security options of the WiFiClientSecureBearSSL library. httpUpdateSecure.ino is a complete example.

To try this, you need either a sketch with enough free memory to allocate the full TLS/SSL-buffer (16k), or a, currently not very widely supported Maximum Fragment Length Negotiation enabled, server (or do the same trick as I do, using a nodeJS server and set maxSendFragmentLength) that sends the correct response headers and serves the binary. You need a certificate store in SPIFFs that includes your root CA, or a self signed certificate, or a certificate fingerprint. In the latter two cases you need to edit the example to use allowSelfSignedCerts() or setFingerprint(). It is also possible to use encryption without server authentication. In this case use setInsecure(). Of course any of the other supported WiFiClientSecureBearSSL functions may be used as well.


arnolde:
@d-a-v Thank you so much for helping me try this PR in advance! I succeeded in switching the ESP library from 2.4.2 to the git version (after wasting an hour trying to copy it to the ~/Library/Arduino15 directory structure, where the 2.4.2 library was, before realising the docs say to put it in the sketch structure) and checkout out your branch "ssl".

```
#include <ESP8266WiFi.h>
#include "credentials.h"
using namespace BearSSL;
#define DEBUG true
[...]
[ensure wifi connected]

    HTTPClient http;  //Declare an object of class HTTPClient

    Serial.println("Sending HTTP Request...");
    http.begin("http://heise.de");

    const char* headerNames[] = { "Location", "Last-Modified" };
    http.collectHeaders(headerNames, sizeof(headerNames)/sizeof(headerNames[0]));

    int httpCode = http.GET();   
    String loc;
    
    if (httpCode > 0) {
      if (http.hasHeader("Location")) {
        loc = http.header("Location");
       }
      String payload = http.getString();   //Get the request response payload
      Serial.println(payload);                     //Print the response payload
    }

    http.end(); // not sure if this is necessary

    if (httpCode=302) {
      Serial.println("Redirecting to "+loc+"...");

      http.begin(loc,TELEKOM_FP);
      int httpCode = http.GET();                       // this fails with SSL error 40 (cipher mismatch?)
      Serial.println((String)"Resultcode: "+httpCode);
      Serial.println(payload);
    }
```

How must I now change this code so the new changes become effective?
This is the current debug output:

```
16:41:17.563 -> SDK:2.2.1(cfd48f3)/Core:2.4.2-73-gfd795c1c/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.1-13-g163bb82)/BearSSL:6d1cefc
16:41:17.563 -> scandone
scandone
16:41:17.709 -> state: 0 -> 2 (b0)
16:41:17.709 -> state: 2 -> 3 (0)
16:41:17.709 -> state: 3 -> 5 (10)
16:41:17.709 -> add 0
16:41:17.709 -> aid 101
16:41:17.709 -> 
16:41:17.709 -> connected with Telekom, channel 1
16:41:17.709 -> dhcp client start...
16:41:17.747 -> cnt 
16:41:17.747 -> ip:10.227.184.255,mask:255.255.252.0,gw:10.227.187.254
3[HTTP-Client][end] tcp is closed
16:41:17.887 -> [HTTP-Client] failed connect to https://www.heise.de:443
16:41:17.925 -> [HTTP-Client][returnError] error(-1): connection refused
16:41:17.925 -> [HTTP-Client][returnError] error(-4): not connected
16:41:17.925 -> Result Code: -1
16:41:17.925 -> Payload: 
16:41:17.925 -> Sending HTTP Request...
16:41:17.925 -> [HTTP-Client][end] tcp is closed
16:41:17.925 -> [HTTP-Client][begin] url: http://heise.de
16:41:17.925 -> [HTTP-Client][begin] host: heise.de port: 80 url: 
[HTTP-Client] connected to heise.de:80
16:41:17.996 -> [HTTP-Client] sending request header
16:41:17.996 -> -----
16:41:17.996 -> GET / HTTP/1.1
16:41:17.996 -> Host: heise.de
16:41:17.996 -> User-Agent: ESP8266HTTPClient
16:41:17.996 -> Connection: close
16:41:17.996 -> Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0
16:41:18.030 -> 
16:41:18.030 -> -----
[HTTP-Client][handleHeaderResponse] RX: 'HTTP/1.1 302'
16:41:18.248 -> [HTTP-Client][handleHeaderResponse] RX: 'Date: Wed, 19 Sep 2018 14:41:17 GMT'
16:41:18.248 -> [HTTP-Client][handleHeaderResponse] RX: 'Location: https://hotspot.t-mobile.net/wlan/redirect.do?origurl=http%3A%2F%2Fheise.de%2F&ts=1537368077987'
16:41:18.284 -> [HTTP-Client][handleHeaderResponse] RX: 'Content-Type: text/html;charset=UTF-8'
16:41:18.284 -> [HTTP-Client][handleHeaderResponse] RX: 'Content-Length: 0'
16:41:18.284 -> [HTTP-Client][handleHeaderResponse] RX: 'Connection: close'
16:41:18.284 -> [HTTP-Client][handleHeaderResponse] RX: 'Server: Apache/2.0.143 (Win2010 SP2)'
16:41:18.284 -> [HTTP-Client][handleHeaderResponse] RX: ''
16:41:18.284 -> [HTTP-Client][handleHeaderResponse] code: 302
16:41:18.316 -> [HTTP-Client][writeToStreamDataBlock] connection closed or file end (written: 0).
16:41:18.316 -> [HTTP-Client][end] tcp stop
16:41:18.316 -> 
16:41:18.316 -> [HTTP-Client][end] tcp is closed
16:41:18.316 -> Redirecting to https://hotspot.t-mobile.net/wlan/redirect.do?origurl=http%3A%2F%2Fheise.de%2F&ts=1537368077987...
16:41:18.316 -> [HTTP-Client][end] tcp is closed
16:41:18.316 -> [HTTP-Client][begin] url: https://hotspot.t-mobile.net/wlan/redirect.do?origurl=http%3A%2F%2Fheise.de%2F&ts=1537368077987
16:41:18.354 -> [HTTP-Client][begin] host: hotspot.t-mobile.net port: 443 url: /wlan/redirect.do?origurl=http%3A%2F%2Fheise.de%2F&ts=1537368077987
16:41:18.354 -> [HTTP-Client][begin] httpsFingerprint: FF F4 F8 43 10 C5 A1 44 3B F4 78 2A 83 DE A0 0E 97 7D ED AC
16:41:18.392 -> Starting GET...
State:	sending Client Hello (1)
Alert: handshake failure
16:41:18.612 -> Error: SSL error 40
16:41:18.612 -> Alert: unexpected message
16:41:18.612 -> Error: SSL error 40
16:41:18.612 -> Alert: close notify
16:41:18.612 -> [HTTP-Client] failed connect to hotspot.t-mobile.net:443
16:41:18.612 -> [HTTP-Client][returnError] error(-1): connection refused
16:41:18.612 -> Resultcode: -1
16:41:18.612 -> [HTTP-Client][returnError] error(-4): not connected
16:41:18.612 -> 
16:41:18.612 -> [HTTP-Client][end] tcp is closed
pm open,type:2 0
```

d-a-v:
This is @Jeroen88's branch :)
You need to look at @Jeroen88's example I showed you in #2771: [link](https://github.com/esp8266/Arduino/pull/4980/files#diff-46402b56934d1fb76007c6229f811545) and study this code.
Basically you need to initialize a `BearSSL::WiFiClientSecure` client and give it to your `http.begin()`.

arnolde:
Got it!
```
// from example file
    client.setFingerprint(fingerprint);
    HTTPClient https;
    https.begin(client, loc);
    int httpCode = https.GET();
    String payload = https.getString();
    Serial.println(payload);
```
Now I finally get a response from https://hotspot.t-mobile.net.. and... another 302 redirect... now I have to implement that... but now I'm finally rolling again! Thanks :-)

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -478,7 +613,7 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
 
     // send Payload if needed
     if(payload && size > 0) {
-        if(_tcp->write(&payload[0], size) != size) {
+        if(_client->write(&payload[0], size) != size) {
```
earlephilhower: I'm definitely not a fan of the `#ifdef` bits, but in prior instances in this file I can understand your logic.  You had to have both `_tcpDeprecated` and `_client` operations in the code, with appropriate `if ()`s to ensure the right one was used.  You were kind of stuck with doing both cases due to the way it was being implemented having 2 potential `WiFiClient`s to use.

But don't you need the `if (tcp) tcp->write(); else _client->write();` pattern here and elsewhere below in the code where you `s/_tcp/_client/g`'d?  The way I see it now, old code will simply never write or read anymore, since `_tcpDeprecated` is not being read/written.

d-a-v:
@arnolde 
> I would like to comment though that I would prefer having fingerprint as a string (which can be copy+pasted from browser certificate details) instead of a byte array. That would also be compatible with "the old way".

on its way #5204

Jeroen88:
Hi  Earle,
Good to hear from you and for the delay: no problem at all.
In connect() the _tcpDeprecated is copied into _client. Next, all tcp calls use _client.
I have no computer right now, but reviewing my own PR I think it can be simplified. I hope I can spend some time on it next weekend. 
Did you see my email about your suggestion to allocate the object on the heap? I got an error because your BearSSL library referenced a null pointer during destruction of the object. I think the cause is in your library.
Further, I use this PR myself to do a secure firmware upgrade. Quite a big download! I noticed that once the (WiFi?) communication is disturbed the download is not recovering. The problem is that reproducing the error is difficult, but I tried to find the cause. It seems that going back to lwIP 1.4 solves the issue, but because reproducing the error is difficult I am not sure. I see various issues in the github that might be similar. I am sure it has nothing to do with my PR, the issues also occur without my changes. I reproduce the error quite clumsy: I deliberately disturb the radio signal during a download of a firmware by holding my cell phone very close to the esp. Once the connection is lost the download dies not recover using lwIP 2.0. lwIP 1.4 does. I am going to try the open source version, but: could you give it a try? If I found an issue it is a serious one. One of the github issues reports problems on a noisy connection (somewhere in india). You can check it also without my PR.
Best, Jeroen
Verzonden vanaf mijn Samsung Galaxy-smartphone.
-------- Oorspronkelijk bericht --------Van: "Earle F. Philhower, III" <notifications@github.com> Datum: 03-10-18  18:17  (GMT+01:00) Aan: esp8266/Arduino <Arduino@noreply.github.com> Cc: Jeroen88 <info@el-mago.nl>, Mention <mention@noreply.github.com> Onderwerp: Re: [esp8266/Arduino] Feature/http update wi fi client parameter
  (#4980) 
@earlephilhower requested changes on this pull request.

Sorry for the delay.  Things got busy and I ended up missing this, but @devyte is kicking our butts to get things merged for the next rev so I finally came around to checking this one out too.
If you've run this library with the old example code (not fixed to use the new method), then my review comment is moot so let me know.  But visually, it looks like right now this breaks old code.  The two different WiFiClients are a pain, I understand!

In libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp:
> @@ -478,7 +613,7 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
 
     // send Payload if needed
     if(payload && size > 0) {
-        if(_tcp->write(&payload[0], size) != size) {
+        if(_client->write(&payload[0], size) != size) {

I'm definitely not a fan of the #ifdef bits, but in prior instances in this file I can understand your logic.  You had to have both _tcpDeprecated and _client operations in the code, with appropriate if ()s to ensure the right one was used.  You were kind of stuck with doing both cases due to the way it was being implemented having 2 potential WiFiClients to use.
But don't you need the if (tcp) tcp->write(); else _client->write(); pattern here and elsewhere below in the code where you s/_tcp/_client/g'd?  The way I see it now, old code will simply never write or read anymore, since _tcpDeprecated is not being read/written.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/esp8266/Arduino","title":"esp8266/Arduino","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/email/message_cards/header.png","avatar_image_url":"https://assets-cdn.github.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/esp8266/Arduino"}},"updates":{"snippets":[{"icon":"PERSON","message":"@earlephilhower requested changes on #4980"}],"action":{"name":"View Pull Request","url":"https://github.com/esp8266/Arduino/pull/4980#pullrequestreview-161244467"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/esp8266/Arduino/pull/4980#pullrequestreview-161244467",
"url": "https://github.com/esp8266/Arduino/pull/4980#pullrequestreview-161244467",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
},
{
"@type": "MessageCard",
"@context": "http://schema.org/extensions",
"hideOriginalBody": "false",
"originator": "AF6C5A86-E920-430C-9C59-A73278B5EFEB",
"title": "@earlephilhower requested changes on 4980",
"sections": [
{
"text": "Sorry for the delay.  Things got busy and I ended up missing this, but @devyte is kicking our butts to get things merged for the next rev so I finally came around to checking this one out too.\r\n\r\nIf you've run this library with the old example code (not fixed to use the new method), then my review comment is moot so let me know.  But visually, it looks like right now this breaks old code.  The two different `WiFiClient`s are a pain, I understand!",
"activityTitle": "**Earle F. Philhower, III**",
"activityImage": "https://assets-cdn.github.com/images/email/message_cards/avatar.png",
"activitySubtitle": "@earlephilhower",
"facts": [

]
}
],
"potentialAction": [
{
"targets": [
{
"os": "default",
"uri": "https://github.com/esp8266/Arduino/pull/4980#pullrequestreview-161244467"
}
],
"@type": "OpenUri",
"name": "View on GitHub"
},
{
"name": "Unsubscribe",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"MuteNotification\",\n\"threadId\": 362131149\n}"
}
],
"themeColor": "26292E"
}
]

d-a-v:
@Jeroen88 This is quite interesting:
>  I noticed that once the (WiFi?) communication is disturbed the download is not recovering.

Could you by any chance try to reproduce that bug using lwip2 in #5126 ?

> It seems that going back to lwIP 1.4 solves the issue

This is not the first time I read that, and I have no strong explanation. I will revisit #4176 and try to understand what happens, maybe with more benchmarks, because that issue does not happen when TCP is slow

earlephilhower:
Howdy @Jeroen88.  I see the assignment now, but don't understand why there are several spots where you check if it is != null and then do something w/it, when _client could be used everywhere except at deletion.

Thanks for the heads-up on your mail.  Yahoo seems to think you're spam and keeps moving your messages into the junk folder.  
> I nailed the problem down to the destructor ~HTTPClient() in ESP8266HTTPClient. This calls _client->stop().
In WiFiClientSecure::stop() the function _clearAuthenticationSettings() is called. This function clears the pointers _sk and _ta. Next the destructor ~WiFiClientSecure() is called. This calls delete _sk and delete _ta (if _deleteChainKeyTA is true, which I asume is) but these pointers were cleared already.
If I skip _client->stop() in ~HTTPClient() in ESP8266HTTPClient everything works fine.

It's perfectly safe to `delete nullptr;` same as `free(NULL);`.  It should be a no-op (and in fact, the way we do it `delete` just calls `free` anyway.  I see some example code in the mail, let me try it and see if I can hit it and go through some stack traces.  What I think **is** there, though, is a possible memory leak when you use the axTLS shim (should only assign _sk/etc=nullptr after freeing it in _clearAuthSet() if the flag is set).

If the app was mixing some axTLS compatibility and some BearSSL type calls, though, then it may also go boom since I didn't set an individual flag for each possible compatibly-set variable.  Probably should, but the axTLS compatibility was definitely done as an afterthought and in about 10 minutes...

earlephilhower:
@Jeroen88 I just pulled this PR and ran the sample you sent (checking to make sure the client->stop() was present in ~HTTPClient).  Runs fine, dumps the website every loop without a problem.  Can you check it's still happening to you when you apply GIT head to this PR?

Jeroen88:
@earlephilhower I still have a crash running the example streamHttpsClient.ino of this PR. I did not apply GIT head because I do not know how to do that (I am still a git newbie). 
You are right that the problem is not the delete's. I am still not sure of the root cause, but: when the streamHttpsClient.ino-sketch calls delete client at the end of the loop ~WiFiClientSecure() is called before ~HTTPClient(). The ~HTTPClient() destructor calls _client->stop() with _client being a WiFiClientSecure and stop() is defined in WiFiClient.cpp and stop() calls flush() and _client->close(). However: the client context is already gone, because the ~WiFiClientSecure() destructor was already called. I think this is the problem. I do not know how it should be solved, in WiFiClientSecureBearSSL of in ESP8266HTTPClient. In case of the latter I also do not know how...

Jeroen88:
@d-a-v 

> Could you by any chance try to reproduce that bug using lwip2 in #5126 ?

I would like to try it, but I do'nt know how. The clue 'git fetch origin pull/5126/head:lwip210' in the PR gives an error: 'fatal: Couldn't find remote ref pull/5126/head'

d-a-v:
> I would like to try it, but I don't know how. The clue 'git fetch origin pull/5126/head:lwip210' in the PR gives an error: 'fatal: Couldn't find remote ref pull/5126/head'

Can you try with this ?
```
git checkout master
git pull
git fetch origin pull/5126/head:lwip210
git checkout -b testonlybranch
git merge feature/HTTPUpdate_WiFiClient_parameter
git merge lwip210
```

Jeroen88:
@d-a-v I had to use git fetch _upstream_ pull/5126/head:lwip210 to find the PR. After _restarting_ the Arduino IDE I have 6 lwIP options now: lwIP v2 Lower Memory, v2 Higher Bandwidth, v2 Lower Memory (no features), v2 higher Bandwidth (no features), v1.4 Higher Bandwidth, v1.4 Compile from source. If I use v2 High Bandwidth, so **without** (no features), it seams "impossible" to fully disturb the signal. I am able to slow it down, but it finishes and reboots with the new firmware.
As you already guessed, I am using the LED blink feature of the Updater to assess the speed of the download. Without disturbance it blinks slightly longer on than off, with disturbance clearly longer on than off

Jeroen88:
@earlephilhower 

> LGTM, but maybe you could update the example code to use this new feature? W/o an example, it's not exceptionally clear from the header what's going on.

Sure! All 3 examples updated in this PR with the usage of setLedPin() and a comment on how to use it

Jeroen88:
@d-a-v You're welcome, happy to help! I am glad that the lwIP devs SACKed us ;)
Maybe origin is the default, but I am a rather newbie with github and I do not comprehend it fully yet. My 'own' fork is origin and upstream is the original repo I cloned from. Before doing any changes I pull from upstream and push to origin to sync them

earlephilhower:
You had a trailing space in a comment, causing the CI style_check to fail.  Fixed it for you.

earlephilhower:
I just tried this code with the original example code (simulating someone running their existing app) and it looks like the `_tcpDeprecated` is not quite there yet:

````
[HTTP] begin...
[HTTP-Client][end] tcp is closed
[HTTP-Client][begin] url: http://192.168.1.12/test.html
[HTTP-Client][begin] host: 192.168.1.12 port: 80 url: /test.html
[HTTP] GET...
[HTTP-Client] connect: HTTPClient::begin was not called or returned error
[HTTP-Client][returnError] error(-1): connection refused
[HTTP] GET... failed, error: connection refused
[HTTP-Client][end] tcp is closed
````

earlephilhower:
Actually, I see the old sample code was silly and required a local 192.168.x.x server.

Changing the server to one that exists doesn't change the results, unfortunately:
````
[HTTP] begin...
[HTTP-Client][end] tcp is closed
[HTTP-Client][begin] url: http://jigsaw.w3.org/HTTP/connection.html
[HTTP-Client][begin] host: jigsaw.w3.org port: 80 url: /HTTP/connection.html
[HTTP] GET...
[HTTP-Client] connect: HTTPClient::begin was not called or returned error
[HTTP-Client][returnError] error(-1): connection refused
[HTTP] GET... failed, error: connection refused
[HTTP-Client][end] tcp is closed
````

earlephilhower:
Original code output (StreamHTTPClient with the URL moved to the jigsaw.w3.org server)
````

SDK:2.2.1(cfd48f3)/Core:2.4.2-70-g9bc8ea1/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.1-13-g163bb82)/BearSSL:f55a6ad



[SETUP] WAIT 4...
wifi evt: 2
scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 3
cnt 

connected with NOBABIES, channel 10
dhcp client start...
wifi evt: 0
[SETUP] WAIT 3...
ip:192.168.1.138,mask:255.255.255.0,gw:192.168.1.1
wifi evt: 3
[SETUP] WAIT 2...
[SETUP] WAIT 1...
[WIFI][APlistAdd] add SSID: NOBABIES
[HTTP] begin...
[HTTP-Client][begin] url: http://jigsaw.w3.org/HTTP/connection.html
[HTTP-Client][begin] host: jigsaw.w3.org port: 80 url: /HTTP/connection.html
[HTTP] GET...
[hostByName] request IP for: jigsaw.w3.org
[hostByName] Host: jigsaw.w3.org IP: 128.30.52.21
:ref 1
[HTTP-Client] connected to jigsaw.w3.org:80
[HTTP-Client] sending request header
-----
GET /HTTP/connection.html HTTP/1.1
Host: jigsaw.w3.org
User-Agent: ESP8266HTTPClient
Connection: close
Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0

-----
:wr 160 0
:wrc 160 160 0
:ack 160
:rn 536
:rch 536, 284
:rcl
:abort
[HTTP-Client][handleHeaderResponse] RX: 'Connection: extensionheader,close'
[HTTP-Client][handleHeaderResponse] RX: 'Date: Sat, 06 Oct 2018 00:06:57 GMT'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Length: 550'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Type: text/html'
[HTTP-Client][handleHeaderResponse] RX: 'Etag: "1giilod:q0efehi8"'
[HTTP-Client][handleHeaderResponse] RX: 'Last-Modified: Tue, 20 Jun 2000 13:33:22 GMT'
[HTTP-Client][handleHeaderResponse] RX: 'Server: Jigsaw/2.3.0-beta3'
[HTTP-Client][handleHeaderResponse] RX: 'ExtensionHeader: ExtensionValue'
[HTTP-Client][handleHeaderResponse] size: 550
[HTTP] GET... code: 200
<HTML>
<HEAD>
  <!-- Created with AOLpress/2.0 -->
  <TITLE>Connection Header</TITLE>
</HEAD>
<BODY>
<P>
<IMG ALT="Jigsaw" BORDER="0" WIDTH="212" HEIGHT="49" SRC="/icons/jigsaw">
<H1>
  The <I>Connection</I> header
</H1>
<P>
This page will be served to you:c 1, 536, 820
 with the following headers:
<P>
<CODE>ExtensionHeader: ExtensionValue<BR>
Connection: ExtensionHeader</CODE>
<P>
If you're getting this page through a proxy, you should <I>not</I> see the
<I>ExtensionHeader</I> !
<P>
  <HR>
<BR>
<A HREF="mailto:jigsaw@w3.:c0 1, 284
org">jigsaw@w3.org</A>
</BODY></HTML>

[HTTP] connection closed or file end.
[HTTP-Client][end] tcp is closed
:ur 1
:del
pm open,type:2 0
````

earlephilhower:
@Jeroen88, hope you don't mind but the fix was simple so I just moved the problem NULL check to after the _client is actually set and committed.  Now old samples seem to work fine with the new class.

Jeroen88:
@earlephilhower did you see my comment:

> I still have a crash running the example streamHttpsClient.ino of this PR. I did not apply GIT head because I do not know how to do that (I am still a git newbie).
> You are right that the problem is not the delete's. I am still not sure of the root cause, but: when the streamHttpsClient.ino-sketch calls delete client at the end of the loop ~WiFiClientSecure() is called before ~HTTPClient(). The ~HTTPClient() destructor calls _client->stop() with _client being a WiFiClientSecure and stop() is defined in WiFiClient.cpp and stop() calls flush() and _client->close(). However: the client context is already gone, because the ~WiFiClientSecure() destructor was already called. I think this is the problem. I do not know how it should be solved, in WiFiClientSecureBearSSL of in ESP8266HTTPClient. In case of the latter I also do not know how...

This is with streamHttp**s**Client.ino. Any suggestions?

earlephilhower:
@Jeroen88, I'm going to commit this now, but if after this is committed you pull git head and still see the crash open an issue and we can look at it.

I'm running the StreamHTTPSClient (i.e. BSSL) example now and it's running just fine, so I have no idea what you're seeing, unfortunately.

scottchiefbaker:
Now that the ESP8266 call requires a `WiFiClient` object for the `begin()` method it's no longer the same as the ESP32 class. If I'm trying to make code that compiles on both ESP8266 and ESP32 do I have to wrap **every** `begin()` call in:

	#ifdef ESP32
	http.begin(url);
	#else
	http.begin(client, url);
	#endif

Or is there another easier way to make cross-platform code?

Jeroen88:
@scottchiefbaker ESP32 HTTPClient supports the same ::begin() function as ESP8266HTTPClient as you can see [here](https://github.com/espressif/arduino-esp32/blob/9b75c65fc76ea67b7b8420dbc2c8ccc917182765/libraries/HTTPClient/src/HTTPClient.h#L137)

scottchiefbaker:
@Jeroen88 oh I didn't know the ESP32 libraries we're already update. I fixed [my code](https://github.com/scottchiefbaker/ESP32-Light-Monitor/blob/master/LightLevel.ino#L91) to reflect the new calling parameters and now it compiles on both platforms correctly. Thank you.

Jeroen88:
@scottchiefbaker you're welcome!

PR #4984, opened 2018-07-30
change "WeMos" to new brand name "LOLIN"

wemos:
The D1 mini from version 3.1.0 and D1 mini Pro from version 2.0.0 using new brand name "LOLIN".

https://wiki.wemos.cc/products:d1:d1_mini
https://wiki.wemos.cc/products:d1:d1_mini_pro


devyte:
@wemos is there any difference between older hardware versions and these new ones? Are there any additional pins? or different pins? flash mode (DIO/QIO) or speed? Puya-like flash chips? Something else?
Or are these new versions functionally identical and just optimized for production?
If these new ones are functionally identical as the previous, then this PR is ok. But if there are any differences in functionality, including changes in flash mode, speed, use of the infamous Puya flash chips, etc, then additional things need to be changed.

devyte:
For example, I don't see the Dx pin numbering on the board. Does that mean that the D0-D8 pin macros should no longer be #defined?

wemos:
@devyte The flash mode, speed and pins are as same as previous versions, only optimization circuit and using ESP8266EX IC instead of the original ESP-12S module.

devyte:
Great! Thank you for the answer.

Misiu:
@wemos @d-a-v does [D1 Mini](https://wiki.wemos.cc/products:d1:d1_mini#pin) and [D1 Mini Pro](https://wiki.wemos.cc/products:d1:d1_mini_pro#pin) expose GPIO9 and GPIO10? I don't see them listed.

d-a-v:
They will never. Only d1 mini lite could,
but the board itself would have to be updated by @wemos.

PR #4985, opened 2018-07-30
Update ESP8266WiFiSTAClass::waitForConnectResult()  for the same behavior with ESP32.

ShinDaiIchi:
Add  counter i for trying WIFi-connection 100 times

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -374,7 +374,8 @@ uint8_t ESP8266WiFiSTAClass::waitForConnectResult() {
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    while(status() == WL_DISCONNECTED) {
+    int i = 0;
+    while((!status() || status() >= WL_DISCONNECTED) && i++ < 100) {
         delay(100);
     }
     return status();


d-a-v:
Ref: #4983 

You seem to have committed something else along with this proposal. Please do another PR for this.

To not break very precious backward compatibility,
what do you think of adding an optional parameter that would not change behavior for existing sketches ?
```
uint8_t ESP8266WiFiSTAClass::waitForConnectResult(uint8_t maxSeconds = 0) { // 0 is infinity
```
You'd propose the same fix to esp-idf with 10 instead of 0.
@me-no-dev what do you think ?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -374,7 +374,8 @@ uint8_t ESP8266WiFiSTAClass::waitForConnectResult() {
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    while(status() == WL_DISCONNECTED) {
+    int i = 0;
+    while((!status() || status() >= WL_DISCONNECTED) && i++ < 100) {
```
devyte: This is a max of 10s time. Some connections could take longer, especially under poor signal conditions. Some users could not want to spend so much time trying to connect and want to give up sooner. I suggest taking an argument that defaults to this when not set.

cores/esp8266/WString.h
```diff
@@ -237,7 +237,8 @@ class String {
         void toLowerCase(void);
         void toUpperCase(void);
         void trim(void);
-
+        
+        int td_split(String delimiter, String** str_array); // v2.3   split String to String Array by multi-delimiters
```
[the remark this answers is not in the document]
ShinDaiIchi: Already removed  this String addition change.

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
d-a-v: What would happen with a default non inifinite value when user's network connectivity / the router is for any reason powered down or disconnected for a while (a longer while than this default value), or takes long to power on, or whatever ?

My point is that usually sketches patiently wait for connectivity. With this change, all sketches would need to be updated to loop on the connectivity loop.

I don't question this PR, only the historical default behaviour / backward compatibility.

devyte: Oops @d-a-v is correct: either we maintain backwards compatibility and merge now, or we keep this breaking change and wait until release 3.0.0 to merge. I prefer the former.
So: 
1. the function should take a timeout argument, in ms, which defaults to some value that disables the timeout (std::numeric_limits?)
2. If the user wants a timeout, he can specify a value for the argument.
3. In a future release we'll change the default to some value like 10 or 15s.

@ShinDaiIchi Could you please implement the changes and update the PR?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -369,12 +369,13 @@ bool ESP8266WiFiSTAClass::getAutoReconnect() {
...
  */
-uint8_t ESP8266WiFiSTAClass::waitForConnectResult() {
+uint8_t ESP8266WiFiSTAClass::waitForConnectResult(uint32_t wait_secs) {
     //1 and 3 have STA enabled
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    while(status() == WL_DISCONNECTED) {
+    int i = 0; int try_times = wait_secs*10;
+    while((!status() || status() >= WL_DISCONNECTED) && ((wait_secs!=0)? (i++ < try_times): true )) {
```
devyte: don't use inverted logic please:
... && ((wait_secs==0) ? true : (i++ < try_times) )

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -369,12 +369,13 @@ bool ESP8266WiFiSTAClass::getAutoReconnect() {
...
  * @return wl_status_t
  */
-uint8_t ESP8266WiFiSTAClass::waitForConnectResult() {
+uint8_t ESP8266WiFiSTAClass::waitForConnectResult(uint32_t wait_secs) {
     //1 and 3 have STA enabled
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    while(status() == WL_DISCONNECTED) {
+    int i = 0; int try_times = wait_secs*10;
```
devyte: PLease declare/init try_times on its own line

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -369,12 +369,14 @@ bool ESP8266WiFiSTAClass::getAutoReconnect() {
...
-uint8_t ESP8266WiFiSTAClass::waitForConnectResult() {
+uint8_t ESP8266WiFiSTAClass::waitForConnectResult(uint32_t wait_secs) {
     //1 and 3 have STA enabled
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    while(status() == WL_DISCONNECTED) {
+    int i = 0; 
+    int try_times = wait_secs*10;
+    while((!status() || status() >= WL_DISCONNECTED) && ((wait_secs==0)? true : (i++ < try_times))) {
```
yoursunny: This line is difficult to read.
* `!status()` tests whether the status is non-zero, but status is an enum and it’s unobvious which value is zero.
* The part after && is really complicated.

How about:
```
for (int timeout = wait_secs * 10; wait_secs == 0 || timeout > 0; --timeout) {
  auto st = status();
  switch (st) {
    case WL_CONNECTED:
    case WL_NO_SSID_AVAIL:
    case WL_CONNECT_FAILED:
      return st;
    default:
      delay(100);
      break;
  }
}

devyte: Agree with @yoursunny . However:

I have a PR with a class that encapsulates the polled timeout. I suggest waiting a bit until it's merged, then using it, instead of reinventing the wheel for the nth time.
the switch should define cases for all enum values, and leave the default case for values not in the enum (error).

devyte:
@ShinDaiIchi the polledTimeout class has been available for a while now. Could you please resolve conflicts and update with that?

PR #4992, opened 2018-07-31
(WIP) optional use of esptool.py for flashing 

d-a-v:
**Untested in all situations**

This is an option.

* tested with linux and wemos d1 mini
* original esptool is still used to generate the image (it sets at least flash freq and flash mode)
* erase command not treated (yet)
* reset method not treated (yet) (working with d1 which uses nodemcu method, esptool.py by default? triggers DTR)

How to try:

* run `get.py` to download `esptool.py`.
* run `tools/board.txt.py --esptoolpy --allgen`
* restart IDE

How to get back to default:
* run `tools/board.txt.py --allgen`
* restart IDE

(the big changes in `boards.txt.py` is only the `macros` variable moved from global to `all_boards()` function because it uses now the new `esptool` variable)

### DIFF
--- a/package/package_esp8266com_index.template.json
+++ b/package/package_esp8266com_index.template.json
@@ -268,6 +268,20 @@
               "size": "43609"
             }
           ]
+        },
+        {
+          "name": "esptoolpy",
+          "version": "2.5.0",
+          "systems": [
+            {
+              "host": "any",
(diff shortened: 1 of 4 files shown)


d-a-v:
closed in favor of #5635 which is running very well

PR #5009, opened 2018-08-05
optionally allow redirects on HTTPClient & OTA updates

liebman:
Support is disabled by default.  Call `ESPhttpUpdate.followRedirects(true);` to enable.


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -369,6 +369,29 @@ void HTTPClient::setTimeout(uint16_t timeout)
...
+    String protocol = url.substring(0, index);
+    if(protocol != "http" && protocol != "https") {
+        DEBUG_HTTPCLIENT("[HTTP-Client][begin] unknown protocol '%s'\n", protocol.c_str());
+        return false;
+    }
+
+    _port = (protocol == "https" ? 443 : 80);
+    return beginInternal(url, protocol.c_str());
+}
+
```
d-a-v: Is there any way to factorize parsing code used in `begin(url...)` ?

liebman: I'll take a look this weekend.

liebman: So i've refactored a bit. It does bring up the question of when the redirect is from 'http' to 'https', or vise versa, as traits are incorrect in the current version and when #4980 is merged the current client would not support the new protocol.  So, i'm wondering if #4980 should support a ClientFactory parameter so that the client could be created/recreated when the protocol changes.  Or maybe its not worth the work and we just have to make a new HTTPClient for redirects that change protocol. 
Currently I have setURL() returning false if the protocol would change.

d-a-v:
Redirection code is currently handled in `ESP8266httpUpdate.cpp` in your PR.
I was wondering why not moving it straight into `ESP8266HTTPClient.cpp` ?
After a discussion with @earlephilhower, this is doable if the protocol does not change after redirection (http vs https).
If it changes (security involved), then you can return the same code as before (301 or 302).
What do you think ?

liebman:
@d-a-v Its still a work in progress but its functional.  Before I clean up a few optional TBDs I want to see what others think.  Also I'm thinking of allowing redirects for POST requests by turning them into GET requests at the new location if redirects are enabled. (https://en.wikipedia.org/wiki/Post/Redirect/Get)

liebman:
@d-a-v Updated.  Added support for POST 303 redirect.  Also added device tests for various redirect features.

liebman:
@d-a-v any more changes needed?

PR #5020, opened 2018-08-09
Analog input voltage can be 3.3V

SarenT:
According to this tutorial and my personal test with my Amica, 1.0V delivers 320 and 3.3V delivers 1024 at 10-bit resolution. I guess 3.3V is possible.

### DIFF
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -42,7 +42,7 @@ either to read voltage at ADC pin, or to read module supply voltage
 (VCC).
 
 To read external voltage applied to ADC pin, use ``analogRead(A0)``.
-Input voltage range is 0 — 1.0V.
+Input voltage range is 0 — 1.0V or 0 — 3.3V.
 
 To read VCC voltage, use ``ESP.getVcc()`` and ADC pin must be kept
 unconnected. Additionally, the following line has to be added to the
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -42,7 +42,12 @@ either to read voltage at ADC pin, or to read module supply voltage
 (VCC).
 
 To read external voltage applied to ADC pin, use ``analogRead(A0)``.
-Input voltage range is 0 — 1.0V, however some breakout boards may implement voltage dividers.
+Input voltage range of bare ESP8266 is 0 — 1.0V, however some many 
+boards may implement voltage dividers. To be on the safe side, <1.0V 
+can be tested. If e.g. 0.5V delivers values around ~512, then maximum 
+voltage is very likely to be 1.0V and 3.3V may harm the ESP8266. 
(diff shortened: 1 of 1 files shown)


devyte:
The voltage range that can be applied to the ESP8266 ADC is 0-1.0V. If you go above 1V you will fry the electronics.
Some boards incorporate a voltage divider* so that a range of 0-3.3V applied to the board ADC pin scales down to 0-1V on the chip ADC. Some boards don't have this divider and break out the chip ADC directly to the board pin.
So the voltage range allowed on the board ADC pin depends on the board.

*many users don't know about, or at least don't know the implications of, this internal voltage divider, and apply a voltage to the board pin from another external voltage divider. This interacts with the board divider and can produce apparently wrong readings from the ADC.

SarenT:
Can we write it in a way that people need to be careful when trying to read ADC? I was looking around but this issue isn't addressed at all and I was confused that my readings at 1.0V are 320. As a safe method, one can start reading 1.0V to make sure that they don't fry with 3.3V. If the readings are 1/3 of 1024 -> then they can be sure about 3.3V.

duncan-a:
Or people could have the sense to look at the schematic of the board
they're using...

SarenT:
In my case, I didn't find the schematics and many new electronic hobbyists might need further explanation/warnings to avoid fatal mistakes.

duncan-a:
Yes, I agree - RoboIndia don't seem to provide much information about
anything.

You could trace the track from the A0 pin but that really shouldn't be
necessary.

The problem with your suggested method is 'how many people have a 1V
source?'

I think that something about "1V on the actual ESP8266 - breakout boards
MAY implement voltage dividers" should be sufficient.

On 9 Aug 2018 16:06, "SarenT" <notifications@github.com> wrote:

In my case, I didn't find the schematics and many new electronic hobbyists
might need further explanation/warnings to avoid fatal mistakes.

—

<https://github.com/esp8266/Arduino/pull/5020#issuecomment-411769705>, or mute
the thread

doc/reference.rst
```diff
@@ -42,7 +42,7 @@ either to read voltage at ADC pin, or to read module supply voltage
 (VCC).
 
 To read external voltage applied to ADC pin, use ``analogRead(A0)``.
-Input voltage range is 0 — 1.0V.
+Input voltage range is 0 — 1.0V or 0 — 3.3V.
 
 To read VCC voltage, use ``ESP.getVcc()`` and ADC pin must be kept
```
jindrichsirucek: There also should be mentioned, that many boards (it means basicaly all boards which are not bare esp) has voltage divedr in theirs design, so ESP.getVcc() will produce totaly wrong results - possible resistor connected to A0 pin has to be removed before using this function

PR #5024, opened 2018-08-10
FunctionalScheduledTicker

hreintke:
Implementation of functional callbacks to Ticker library
Adding option to schedule Ticker callback

PR #5030, opened 2018-08-12
Implementation of Functional and Scheduled option in Ticker lib

hreintke:


hreintke:
@devyte : Next try for Ticker update.
Replaces #5024. 

Based on master -> No issue with active() PR.

I do not understand why the travis check fails on the example.
It compiles in my environment , in the travis details I see a lot of duplicate code.
If you give a direction to search I'll update

earlephilhower:
https://travis-ci.org/esp8266/Arduino/jobs/415077264

CI failure is related to style on the example code.  We check for Arduino formatting to try and get a consistent look to the code.  Intent of 2, braces on the function definition line, etc.  The link above shows the problem bits.

hreintke:
@earlephilhower 
Thx, I did see the travis log but did not understand the red/geen code.

Will update the example.

hreintke:
Updated the example.
The failed travis check is due to connection issue, not PR related.

fabiuz7:
Hi,
probably it is my fault, but I don't get the differences between the "scheduled" and "not-scheduled" methods. Looking at your example (Ticker Functional.ino), it seems they are the same:

From your example, you declare the 2 functions:
```
void staticBlink() {
  digitalWrite(LED2, !digitalRead(LED2));
}

void scheduledBlink() {
  digitalWrite(LED3, !digitalRead(LED2));
}
```
And in the setup:
 ```
pinMode(LED2, OUTPUT);
  staticTicker.attach_ms(100, staticBlink);

  pinMode(LED3, OUTPUT);
  scheduledTicker.attach_ms_scheduled(100, scheduledBlink);
```

Links or comments in code could be useful to understand.

However thanks for your work, it seems a _needed and important_ improvement.

hreintke:
Indeed, the only difference in the code is the 
`.attach_ms(100, `  vs the `.attach_ms_scheduled(100,`

Although the behavior of the both options is different of course.

I made the code this identical in order to keep the focus on the "schedule functionality"

fabiuz7:
That's perfectly fine, but I would understand what are this "schedule functionality" :)

I mean, I've just learnt that in esp8266 exists _schedule.h_. Now I'm wondering what are the differences between "scheduling" a function instead the simple ticker.attach(..)

Any links or suggestion is appreciated

devyte:
The non scheduled form executes asynchronously in the system context. The code that can execute there has certain restrictions, e.g. you can't yield().
The scheduled form executes as thoughnit were in the loop. The time latency until the code starts executing is a bit higher, but the code has the same freedoms as loop() code.

fabiuz7:
Clear explanation, thank you very much!

I'm searching on readthedocs... Is already there a page documenting this?

PR #5038, opened 2018-08-13
Infinite loop while passing File(FS.h) resolved

mayankgour13:
Stream.available() never reaches to -1 which makes it an infinite loop. When File (FS.h) goes empty it reaches to 0 not -1.
Please do check with other Stream implementation and if all is fine, Change requested to be applied.


d-a-v:
Someone enlighten me:

`WiFiClient::available()` returns 0 when data is unavailable *now*, that does not mean that all data is yet received. @earlephilhower Is it the same with `WiFiClientSecure::available()` ?

Indeed `->available() > -1` is always true and useless here
```
while(connected() && (stream->available() > -1) && (len > 0 || len == -1)) {
```
This loop is "busy waiting / uselessly active" when data is not available (at the time of testing). When the stream is closed, `->connected()` will return false, making that loop *not* infinite. This is the arduino way, no RTOS env.

This code indeed need revisiting (I just discovered this arduino `get/setWriteError()` API which is also useless here, WiFi library don't use them).

We need to implement a generic `Print::write(Stream&)` *with timeout* (as previously stated in gitter with @igrr some month ago) that would find its use in many places (where virtuals are still in use :-) ) in the core and particularly here. 

To my mind, changing -1 to 0 will prematurely end the connection even when all data is not yet received, it has been seen in many tcp related cases in other issues that sometimes TCP can be slow.

PR #5043, opened 2018-08-14
build: place generated ld script into build directory

### LINKED ISSUE #5008: IDE 1.8.5 recompiles every time after update to core 2.4.2
### Basic Infos

- [x ] If there is a stack dump, I have decoded it.

#### Platform

- Hardware:				All
- Core Version:			[2.4.2]
- Development Env:		[Arduino IDE 1.8.5]
- Operating System:		[Windows10]

### Settings in IDE

N/A, apllies to all

### Problem Description

I'm sure I have done nothing to any settings in the IDE (1.8.5) but since updating to esp8266 core 2.4.2 everything is recompiled on a build, every time even if it has not been touched.

Not a "showstopper" but it sure slows down my dev cycle. Please tell me it's nothing something daft I have done...

On the plus side, I am now able to compile a huge program that just grew to busting memory limits, so well done for clawing back lots of heap/ram etc, it saved my project, thanks!

igrr:
arduino-builder 1.3.25 (shipped with Arduino 1.8.5) forces full
recompilation when any file in the core directory is modified. Avoid
full recompilation by placing generated ld script into build
directory, not source directory.

Also fix an issue where git version description would not be generated
if there were spaces in build path.

Closes #5008.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -74,13 +74,13 @@ compiler.elf2hex.extra_flags=
 ## generate file with git version number
 ## needs bash, git, and echo
 recipe.hooks.core.prebuild.1.pattern=bash -c "mkdir -p {build.path}/core && echo \#define ARDUINO_ESP8266_GIT_VER 0x`git --git-dir {runtime.platform.path}/.git rev-parse --short=8 HEAD 2>/dev/null || echo ffffffff` >{build.path}/core/core_version.h"
-recipe.hooks.core.prebuild.2.pattern=bash -c "mkdir -p {build.path}/core && echo \#define ARDUINO_ESP8266_GIT_DESC `cd {runtime.platform.path}; git describe --tags 2>/dev/null || echo unix-{version}` >>{build.path}/core/core_version.h"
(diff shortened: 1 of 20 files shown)


igrr:
With these changes the following happens on each rebuild:

* core-version.h is generated
* Esp-version.cpp is compiled
* core_esp8266_main.cpp is compiled
* core object files are archived
* LD script is generated
* elf file is linked
* output file is generated

Which still leaves some room for improvement, but is much faster than "recompiling all".

d-a-v:
> To fix recompilation due to LD script generation, we can probably ship a small tool which will "copy file if contents have changed", i.e. compare src and dst, and copy src to dst if they have different content.

Without asking user to install something more, we have gnu(linux) `cp -u`, windows `xcopy /u` and your portable linux/macos `diff || cp`.

.. or ask users to install the arduino.cc's "hourly build" 1.8.6 ?

igrr:
diff || cp is a good idea, thanks.

1.8.6 does not fix recopilation on every build due to updated core_version.h. It's only an issue for git version though.

Issue #4630, opened 2018-04-10
multiple definition of `_putc_r'


trlafleur:
See issue defined below....
tested with 2.4.0 and 2.4.1
~~~
https://github.com/grzesl/ESPMail/issues/4
~~~

~~~
C:\Users\anonymous\AppData\Local\Temp\arduino_build_962164/arduino.ar(libc_replacements.c.o): In function `_putc_r:

C:\Users\anonymous\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\2.4.0\cores\esp8266/libc_replacements.c:96: multiple definition of `_putc_r'

C:\Users\anonymous\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\2.4.0/tools/sdk/libc/xtensa-lx106-elf/lib\libc.a(lib_a-putc.o):/Users/igrokhotkov/e/newlib-xtensa/xtensa-lx106-elf/newlib/libc/stdio/../../../.././newlib/libc/stdio/putc.c:97: first defined here

collect2.exe: error: ld returned 1 exit status

~~~

igrr:
Thanks for reporting this! Started investigating this and found another bug: `printf` returns `-1` instead of the number of characters written, when writing to the console.

Komse:
Hi, I had the same issue. I needed to remove the "#include <iostream>" to compile.

cristianCarrerasCastillo:
remover of what file? @Komse

Komse:
Oh yes sorry, to remove the "#include <iostream>" @cristianCarrerasCastillo

shineangelic:
+1 on this. I didn't get what "#include" I shall remove, thanks

Komse:
For me it was the iostream include @shineangelic

shineangelic:
thank you @Komse but that did not help. I have other weird linking errors as reported in #4663

shineangelic:
I was able to solve this commenting out _put_c method already defined in "tools\xtensa-lx106-elf-gcc\1.20.0-26-gb404fb9-2\"

Issue #3358, opened 2017-06-16
Missing __throw_bad_cast stub


TimWoosh:
### Basic Infos
I hope I'm providing sufficient information. I tried my best to collect as much as possible

#### Hardware
Hardware:			ESP-8266MOD
Core Version:      	3.4.0b12

### Description

Problem description

Compiler error when  std::ostringstream is used.
When commenting out all depenend lines and replacing ```+ convert.str() + ``` with a 1 (inside the quotes) in the line below, all compiles fine.:
```  
std::string msgStr = "{\"message\" : \"hello\", \"msgNum\": \"" + convert.str() + "\"}"; 
```
### Settings in IDE

Module:  			d1_mini
Flash Size: 		?4MB/1MB?
CPU Frequency:		?80Mhz?
Flash Mode:			?qio?
Flash Frequency:	?40Mhz?
Upload Using:		n/a
Reset Method:		n/a

### Sketch

```cpp

#include <Arduino.h>
#include <PubSubClient.h>
#include <ESP8266WiFi.h>
#include <string>
#include <sstream>

const char* ssid = "ssid";
const char* password = "blabla";
const char* mqtt_server = "192.168.0.100:1883";

WiFiClient espClient;
PubSubClient client(espClient);

const byte ledPin = 15;
int counter = 0;
const char *msg;

void callback(char* topic, byte* payload, unsigned int length) {
 Serial.print("Message arrived [");
 Serial.print(topic);
 Serial.print("] ");
 for (int i=0;i<length;i++) {
  char receivedChar = (char)payload[i];
  Serial.print(receivedChar);
  if (receivedChar == '0')
  // ESP8266 Huzzah outputs are "reversed"
  digitalWrite(ledPin, HIGH);
  if (receivedChar == '1')
   digitalWrite(ledPin, LOW);
  }
  Serial.println();
}

void reconnect() {
 // Loop until we're reconnected
 while (!client.connected()) {
 Serial.print("Attempting MQTT connection...");
 // Attempt to connect
 if (client.connect("ESP8266 Client")) {
  Serial.println("connected");
  // ... and subscribe to topic
  client.subscribe("ledStatus");
 } else {
  Serial.print("failed, rc=");
  Serial.print(client.state());
  Serial.println(" try again in 5 seconds");
  // Wait 5 seconds before retrying
  delay(5000);
  }
 }
}

void setup() {
  Serial.begin(9600);
  client.setServer(mqtt_server, 1883);
[...]

PR #5047, opened 2018-08-14
Fix iostream related issues

igrr:
* add stubs for more exception throw calls (https://github.com/esp8266/Arduino/issues/3358)
* make putc_r implementation weak (https://github.com/esp8266/Arduino/issues/4630)
* fix incorrect return value of _write_r call
* add device test for output to std::cout


tests/device/test_iostream/test_iostream.ino
```diff
@@ -0,0 +1,33 @@
...
+    Serial.begin(115200);
+    BS_RUN(Serial);
+}
+
+TEST_CASE("can print to std::cout", "[iostream]")
+{
+    std::stringstream test_stream("");
+    test_stream << "hello stream";
+
+	Serial.readString();
```
d-a-v: Why is this needed ?

devyte: Same comment as @d-a-v

cores/esp8266/libc_replacements.c
```diff
@@ -84,15 +84,18 @@ int ICACHE_RAM_ATTR _read_r(struct _reent* unused, int file, char *ptr, int len)
 
 int ICACHE_RAM_ATTR _write_r(struct _reent* r, int file, char *ptr, int len) {
     (void) r;
+    int pos = len;
     if (file == STDOUT_FILENO) {
-        while(len--) {
+        while(pos--) {
```
devyte: Does this address a current issue or ongoing discussion? Maybe a link could be added in the PR description if that's the case.

tests/device/test_iostream/test_iostream.ino
igrr: Was meant to clean up UART RX buffer just in case there was something in there. Chances are, the test will also work without this.

And I have messed up the indentation in this line...

cores/esp8266/libc_replacements.c
igrr: Haven't seen this reported, probably because including iostream header caused compilation error, so no one got as far as using iostreams. Found when writing the test case.

PR #5049, opened 2018-08-15
Enclose PROGMEM segment names in quotes

earlephilhower:
__FILE__ is used to name the segments used for each PROGMEM constant,
but __FILE__ may have a space in it.  This would cause compilation
errors.

Add quotes around the entire segment name to work around this.


jjsuwa:
LGTM. Thanks.

Issue #3408, opened 2017-07-09
ESP.deepSleep(xxxx, WAKE_NO_RFCAL) not working since 2.0.0


wb8wka:
### Summary

RF Calibration cannot be turned off after 2.0.0 in wake from deepSleep  causing current spike and longer boot

#### Hardware
Hardware:			Wemos D1 mini and ESP-12S, 470uf cap between Vcc and Ground
Core Version:      	2.0.0  (works -  No RF Calibration current spike), 2.3.0  (Not working, current spike)
Test equipment: Rigol Scope, 1ohm current shunt, "fuzz" is because I didn't do calibration on scope but traces are accurate. 

### Description

Using identical code function ESP.deepSleep(6000000, WAKE_NO_RFCAL);  in 2.0.0 ESP can wakeup without large current spike from RF_CAL, however after 2.0.0 RF_CAL executes causing large current spike and longer run time.   Please see attached current traces. Correct data is sent in both instances, only different is current spike, which appears to be RF_CAL executing. 

Sketch sends simple beacon using wifi_send_pkt_freedom. GPIO16 connected to reset.  

WAKE_RF_DISABLED does behave as expected in both 2.0.0 and 2.3.0

Also 

    system_deep_sleep_set_option(2);  // Option 2 is WAKE_NO_RFCAL
    //system_deep_sleep(6000000);
    system_deep_sleep_instant(6000000);  //  Trims 98ms off go to sleep

has same characteristics in that it cannot turn off RF_CAL 

### Settings in IDE

Module:  			WeMos D1 Mini & ESP-12S
Flash Size: 		4MB
CPU Frequency:		80Mhz
Flash Mode:			qio
Upload Using:		SERIAL

### Images

![061517_1ohm_100mvdiv](https://user-images.githubusercontent.com/6334083/27991244-71872e98-643e-11e7-8919-b6abe641115b.png)
### Version 2.0.0 without RF_CAL current spike at head

![060917_1ohm_100mvdiv](https://user-images.githubusercontent.com/6334083/27991231-05b1adb0-643e-11e7-910d-a4e612e977bd.png)
### Sketch

```

#include <ESP8266WiFi.h>
#include <WiFiClient.h> 
extern "C" {
  #include "user_interface.h"
// (Commented out code below for 2.0.0 to expose send_pkt_freedom, not needed in 2,3.0
//  typedef void (*freedom_outside_cb_t)(uint8 status);
[...]

Issue #4969, opened 2018-07-27
ESP.deepSleepMax() - "deep sleep too long"


artua:
- Hardware:					[ESP-12]
- Core Version:			[ff74813d54220adac13f788dae2022aeca25ad2e]
- Development Env:		[Arduino IDE]
- Operating System:		[MacOS]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[qio]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

ESP.deepSleep(ESP.deepSleepMax()) not working, returns "deep sleep too long".

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

void setup() {
    Serial.begin(74880);
    uint64_t deepSleepMax = ESP.deepSleepMax();
    ESP.deepSleep(deepSleepMax, WAKE_RF_DISABLED);
}
```

### Debug Messages

```
SDK:2.2.1(cfd48f3)/Core:2.4.1-130-gff74813d/lwIP:2.0.3(STABLE-2_0_3_RELEASE/glue:arduino-2.4.1-13-g163bb82)/BearSSL:6d1cefc
deep sleep too long
scandone
error: pll_cal exceeds 2ms!!!
no esp_update found, reconnect after 1s
reconnect
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!

[...]

PR #5052, opened 2018-08-16
Add instantly option to deepSleep

torntrousers:
As mentioned in https://github.com/esp8266/Arduino/issues/3408#issuecomment-413279368 here's an update to add an 'instantly' option to the deepSleep function.

### DIFF
--- a/cores/esp8266/Esp.h
+++ b/cores/esp8266/Esp.h
@@ -92,7 +92,7 @@ class EspClass {
         void wdtDisable();
         void wdtFeed();
 
-        void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT);
+        void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT, bool instantly = false);
         uint64_t deepSleepMax();
 
         bool rtcUserMemoryRead(uint32_t offset, uint32_t *data, size_t size);
--- a/cores/esp8266/Esp.h
+++ b/cores/esp8266/Esp.h
@@ -92,7 +92,8 @@ class EspClass {
         void wdtDisable();
         void wdtFeed();
 
-        void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT, bool instantly = false);
+        void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT);
+        void deepSleepInstant(uint64_t time_us, RFMode mode = RF_DEFAULT);
         uint64_t deepSleepMax();
 
         bool rtcUserMemoryRead(uint32_t offset, uint32_t *data, size_t size);
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -107,10 +107,14 @@ void EspClass::wdtFeed(void)
 
 extern "C" void esp_yield();
 
-void EspClass::deepSleep(uint64_t time_us, WakeMode mode)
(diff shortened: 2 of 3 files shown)


cores/esp8266/Esp.h
```diff
@@ -92,7 +92,7 @@ class EspClass {
         void wdtDisable();
         void wdtFeed();
 
-        void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT);
+        void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT, bool instantly = false);
```
igrr: Suggest adding new function, `deepSleepInstant`. This way, only one of `system_deep_sleep_instant`/`system_deep_sleep` will be linked into the final application, depending on what the user needs.

torntrousers: Ok, done.

5chufti:
together with the request for a safe max option like requested in #4969 I opt for something like proposed here https://github.com/esp8266/Arduino/pull/4936#issuecomment-410435875

torntrousers:
Doesn't that still have the issue that @igrr just mentioned of linking in a bit more than necessary. Also ```0x1800000000ull``` is a little obscure compared to explicitly mentioning `instant`` somewhere.

5chufti:
yes but then why not also seperate ESP.deepSleep() and ESP.setWakeUp()
to me this is academic hair-splitting. To get least pulled in from SDK, one should not combine any SDK functions within single ESP.class method at all, ending up in ESP.class just passing through simple SDK api. With the suggested workaround for #3408 it's even one more ...

I don't mind having 2 additional bools, 
> void deepSleep(uint64_t time_us, RFMode mode = RF_DEFAULT, bool max_safe = FALSE, bool instant = FALSE);

no hard feelings for obscure flags in arguments, just wanted to keep it max compat.

igrr:
Dear @5chufti, if a user encounters the following piece of code

```c++
ESP.deepSleep(duration, RF_DEFAULT, true);
```

they will need to go find the function declaration to figure out what "true" means.

Compare this to

```c++
ESP.deepSleepInstant(duration, RF_DEFAULT);
```

where the method name conveys this information.

Admittedly, the core has accumulated numerous APIs with optional boolean arguments, most via simple backwards-compatible enhancements like this one (in the original form of this PR). WiFi class is a good example. It doesn't mean though that the problem should be ignored.

With regards to pulling in SDK functions — this is not academic hair splitting. Let me give you an extreme example. In the first version of HTTP Client library, there was code like this:
```c++
// parse URL string, get host, port, protocol, URI
if (protocol == "https") {
    client = new WiFiClientSecure(host, port);
} else {
    client = new WiFiClient(host, port);
}
```

This is an example of bad design. User usually only needs A or B, but the code includes functionality for both A and B at link time. In this specific case, it pulled in *lots* of SSL-related code, greatly increasing the binary size.

It's going to be less extreme with system_deep_sleep, but still, this is a thing to consider.

5chufti:
@igrr 
> ESP.deepSleepInstant(duration, RF_DEFAULT);
sorry, but here the user has to find the useable definitions that are not even the same as in SDK !

torntrousers:
Cheers @igrr . Don't suppose you could also pop over [here](https://github.com/esp8266/Arduino/issues/3408#issuecomment-355725812) and comment again on that issue?

igrr:
WiFi is turned off as well, just not gracefully.

d-a-v:
OK, I was mistaken by "false friend" `instant` word between french and english.

Issue #619, opened 2015-07-27
system_rtc_mem_write not working in latest code


torntrousers:
Has something change in how to use system_rtc_mem_write?

The following sketch works with the old 1.6.1 release and the serial prints show the rtc byte incrementing, but with the latest code it doesn't work and always outputs "first time".

```
extern "C" {
  #include "user_interface.h"
}

byte rtcStore[2];

void setup() {
  Serial.begin(115200);
  Serial.println();

  system_rtc_mem_read(64, rtcStore, 2);

  if (rtcStore[0] != 123) {
     Serial.println("first time");
     rtcStore[0] = 123;
     rtcStore[1] = 0;
  } else {
     rtcStore[1] += 1;
     Serial.print("rtc = ");
     Serial.println(rtcStore[1]);
  }
    
  system_rtc_mem_write(64, rtcStore, 2);

  Serial.println("deep sleep 3 seconds...");

  //ESP.deepSleep(5000000, WAKE_RF_DISABLED);
  system_deep_sleep_set_option(4); // 4 = low power, radio off mode
  system_deep_sleep(2000000); 
  delay(1000);
}

void loop() {
   // shouldn't get here
}
```

igrr:
Yes, i think while adding a workaround for "RTC MEM CHECK FAIL", i made it wipe user data as well:
https://github.com/esp8266/Arduino/blob/0cc192525fe4d9f784361408dbb82fff2fb96b39/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_phy.c#L247

torntrousers:
Just as an FYI - this is still not working in the latest trunk code

torntrousers:
I guess for a work around i can change to use EEPROM. What are the pros and cons of EEPROM compared to the RTC memory? EEPROM seems quite a lot slower for one thing.

antelder:
Playing around with this some more just now and i find that the example code at the top of this issue works ok if the RTC memory address is changed from 64 to be 65. Thats odd isn't it? The ESP8266 SDK Guide is pretty clear that 64 is the start of the user area.

torntrousers:
Playing around with this some more just now and i find that the example code at the top of this issue works ok if the RTC memory address is changed from 64 to be 65. Thats odd isn't it? The ESP8266 SDK Guide is pretty clear that 64 is the start of the user area.

aperepel:
Still an issue in a current release. But... using to the next segment (not 64) works...

[...]

PR #5056, opened 2018-08-16
Fix for #3408 

torntrousers:
As per https://github.com/esp8266/Arduino/issues/3408#issuecomment-413661491 undo the change in https://github.com/esp8266/Arduino/commit/e02932fcddc24740852ec0e39e494e57a8906df2. I've tried the sketch in #619 that the commit was original done to fix and it does work now without the commit.

### DIFF
--- a/cores/esp8266/core_esp8266_phy.c
+++ b/cores/esp8266/core_esp8266_phy.c
@@ -301,10 +301,7 @@ void user_rf_pre_init()
     // *((volatile uint32_t*) 0x60000710) = 0;
     spoof_init_data = false;
     volatile uint32_t* rtc_reg = (volatile uint32_t*) 0x60001000;
-    if((rtc_reg[24] >> 16) > 4) {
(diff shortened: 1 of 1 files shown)


devyte:
Merging this after confirmation from several users.

torntrousers:
Not that this change appears to have broken RF_CAL, see the on going discussion in https://github.com/esp8266/Arduino/issues/3408#issuecomment-414159385

devyte:
@torntrousers should this be rolled back, new changes added on top, or is further investigation needed to figure out the problem/solution?

torntrousers:
Its really unclear what RF calibration does, for me it appears to work ok without calibration running, and there's the work around of using an RF_PRE_INIT function to force calibration to run - https://github.com/esp8266/Arduino/issues/3408#issuecomment-414477566. And without this change here its impossible to get the fast startup times which make running off small batteries practical. SO IMHO I'd prefer this fix to stay on, at least for now.

PR #5064, opened 2018-08-20
fix min and max for Servo library

tagno25:
When I went to change the pulse width for my servos, the pulse width did not change with the pre-existing code.


earlephilhower:
LGTM, thanks!

PR #5081, opened 2018-08-27
doc: improve WiFi.persistent documentation

### LINKED ISSUE #3641: Documentation for WiFi.persistent() is misleading
Documentation at http://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/generic-class.html says
```Setting persistent to false will get SSID / password written to flash only if currently used values do not match what is already stored in flash.```

But the code does something different:
https://github.com/esp8266/Arduino/blob/35d5fabe33c24931bc629704b645119c0218a0ed/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L136-L151

It writes settings to flash if they don't match the values in flash if persistent is set to **true**. If it is false it will not write to flash independent of the current value in flash.

### PR DESCRIPTION
**Pull request by yoursunny**
fixes #3641

### DIFF
--- a/doc/esp8266wifi/generic-class.rst
+++ b/doc/esp8266wifi/generic-class.rst
@@ -10,7 +10,7 @@ onEvent
 
 .. code:: cpp
 
-    void  onEvent (WiFiEventCb cb, WiFiEvent_t event=WIFI_EVENT_ANY) __attribute__((deprecated)) 
+    void  onEvent (WiFiEventCb cb, WiFiEvent_t event=WIFI_EVENT_ANY) __attribute__((deprecated))
 
 To see how to use ``onEvent`` please check example sketch `WiFiClientEvents.ino <https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiClientEvents/WiFiClientEvents.ino>`__ available inside examples folder of the ESP8266WiFi library.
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: doc/esp8266wifi/generic-class.rst [hunk 1]
```diff
@@ -34,25 +34,22 @@ persistent
...
 
-This might result in some wear of flash memory depending on how often such functions are called.
-
-Setting ``persistent`` to ``false`` will get SSID / password written to flash only if currently used values do not match what is already stored in flash.
-
-Please note that functions ``WiFi.disconnect`` or ``WiFi.softAPdisconnect`` reset currently used SSID / password. If ``persistent`` is set to ``false``, then using these functions will not affect SSID / password stored in flash.
-
-To learn more about this functionality, and why it has been introduced, check issue report `#1054 <https://github.com/esp8266/Arduino/issues/1054>`__.
+Setting ``WiFi.persistent(false)`` disables writing Wi-Fi settings into flash memory.
+Calling ``WiFi.begin``, ``WiFi.disconnect``, ``WiFi.softAP``, or ``WiFi.softAPdisconnect`` only changes the current in-memory Wi-Fi settings, but does not affect SSID / passphrase stored in flash memory.
```
igrr: Suggest "Once WiFi.persistent(false) is called, calling .... only changes the current in-memory Wi-Fi settings ..."

Issue #4460, opened 2018-03-04
Number of Max Concurrrent Connection in Station mode?


dannybres:
Hello,

I am using ESP8266WiFi in station mode to connect to my AP at home. It is controlling 3 RGB leds sets  via the GPIO (9 channels for 3 sets). There is an situation where the server that controls them all polls the ESP8266WiFi server for the status of the lights with many concurrent http requests, e.g.

 - /rgb1/status
 - /rgb1/colour
 - /rgb1/brightness
 - /rgb2/status
 - /rgb2/colour
 - /rgb2/brightness
 - /rgb3/status
 - /rgb3/colour
 - /rgb3/brightness

My code is below and as new connections are created on the server, I add them to the array `clients`, once I have received a whole request and attempt to write back to that client, the line `clients[i]->println("HTTP/1.1 200 OK");` takes a long time to execute and return 0, as I do not believe it still has that connection open and therefore cannot respond to the client? Is this because I am hitting a concurrent connection limit?

**Is there a maximum number of connections in Station mode? 
Can I increase this?**

Thanks,

Daniel

```
#include <ESP8266WiFi.h>

const char* location = "tv room LED controller";
const int locationID = 2;

int whiteLedPin[3] = {5,14,3}; // Pin 8
int redLedPin[3] = {4,12,1}; // Pin 6
int greenLedPin[3] = {0,13,9}; // Pin 7
int blueLedPin[3] = {2,15,10}; // Pin 8

int rgbLedStatus[3] = {0,0,0};
String rgbColour[3] = {"FFFFFF","FFFFFF","FFFFFF"};

int WWLedStatus[3] = {0,0,0};
int ledWWBrightness[3] = {100,100,100};
#define MAX_CLIENTS 10
#define MAX_LINE_LENGTH 50
// Create an instance of the server
// specify the port to listen on as an argument
WiFiServer server(80);
WiFiClient *clients[MAX_CLIENTS] = { NULL };
String st[MAX_CLIENTS];

void printJsonForTemperatureAndHumidityToClient(WiFiClient client,int locationID) {
  client.print("{\"room\": \"");
  client.print(location);
  client.print("\",\"id\": ");
  client.print(locationID);
  client.print(", \"led status\": ");
  client.print(rgbLedStatus[0]);
  client.print(", \"rgb colour\": \"");
  client.print(rgbColour[0]);
  client.print("\"}");
}

unsigned int hexToDec(String hexString) {
  unsigned int decValue = 0;
[...]

Issue #570, opened 2015-07-19
How many clients can connect in Access point mode


STASIUK:
How many clients can connect to the module in Access point mode?
##

acoulson2000:
anyone?

Links2004:
4 clients are possible right now.
https://github.com/esp8266/Arduino/blob/esp8266/hardware/esp8266com/esp8266/libraries/ESP8266WiFi/src/ESP8266WiFi.cpp#L267

acoulson2000:
It seems like it is possible to call wifi_softap_set_config() with a config.max_connections value greater than 4 to override the default? Do you know if this is, in fact the case, and what the ramifications might be? I understand that more clients require more RAM, but from googling around, it's not clear how much. I have a relatively simple host app which I would like to use to send UPD broadcasts to many clients.

I appreciate any sight,
Andy

Links2004:
you can try it, if its working we can add an API to change the value.
keep in mind the TCP sockets are limited to 5 and the UDP sockets to 4 form the SDK. 
but multicast / broadcasts shut reaches all with one socket.

acoulson2000:
Markus,

Thanks for you're feedback. I'm not very C-savy and certainly the whole ESP8266 toolchain is a bit intimidating. To try it, is it necessary to modify the source code and build the firmware? Or can I simply set up my own config structure and pass it in from my own user code using something like below?  I ask because your response regarding the possibility of providing an API seems to imply this will not work.

struct softap_config cfgESP;
//.... set ssid, auth, etc
cfgESP.max_connection=32;
// ....
wifi_softap_set_config(&cfgESP);

Thanks again,
Andy

Links2004:
you can use the SDK directly like you shown. 

to use the SDK you need to do the include like this:
```cpp
extern "C" {
#include "user_interface.h"
}
```

if its working i will add it to the begin function of the AP, so it will be easy to use with out to deal with the SDK.

brianensor:
Any luck @acoulson2000? I'd like to increase the number of connections for my project if possible.

acoulson2000:
So far, I have only connected 3 simultaneously, but so far, so good. It will probably be a couple of weeks at least before I'm ready to try more, but I'll post back with my results.

bertamf:
Hi @acoulson2000! I've also been trying to connect more than 4 devices to the AP, but I haven't succeed so far. What about you?
Thanks!

acoulson2000:
[...]

PR #5088, opened 2018-08-28
Soft-AP documentation update

aerlon:
In several places of the documentation one can read "the maximum number of stations that may be connected to ESP8266 soft-AP is five." or something along those lines. However, the [softAP method](https://github.com/esp8266/Arduino/blob/9634f79f45eead2e053e30460b973ea000a0a8f6/libraries/ESP8266WiFi/src/ESP8266WiFiAP.h#L39) has a `int max_connection = 4` argument which represents ["Max simultaneous connected clients"](https://github.com/esp8266/Arduino/blob/9634f79f45eead2e053e30460b973ea000a0a8f6/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp#L86). It would thus seem that the maximum number of stations that may be connected to ESP8266 soft-AP is four by default. This PR brings the documentation in line with the source code.

While looking into this issue I noticed that the topic may not be quite as simple as it first seems. Although the softAP method claims the max number of clients is 4, I've read other [claims](https://github.com/esp8266/Arduino/issues/570#issuecomment-338116826) that the [max is actually 8](https://bbs.espressif.com/viewtopic.php?f=46&t=481&p=1832&hilit=max_connection#p1832). @D-A-V even seems to have been able to get [12 simultaneous connections working](https://github.com/esp8266/Arduino/issues/4460#issuecomment-370382586) (though I'm not sure if that test is applicable here), and there are [some claims](https://bbs.espressif.com/viewtopic.php?t=4472 ) that the main limiting factor is actually only available memory.

Sadly I do not have enough ESPs to try where the max limit is for myself, so it would be good if @D-A-V or someone else with knowledge on this topic could confirm whether the max actually is 4, 8, 12 or something else.

### DIFF
--- a/doc/esp8266wifi/readme.rst
+++ b/doc/esp8266wifi/readme.rst
@@ -125,7 +125,7 @@ Soft Access Point
 ~~~~~~~~~~~~~~~~~
 
 An `access point (AP) <https://en.wikipedia.org/wiki/Wireless_access_point>`__ is a device that provides access to Wi-Fi network to other devices (stations)
-and connects them further to a wired network. ESP8266 can provide similar functionality except it does not have interface to a wired network. Such mode of operation is called soft access point (soft-AP). The maximum number of stations connected to the soft-AP is five.
+and connects them further to a wired network. ESP8266 can provide similar functionality except it does not have interface to a wired network. Such mode of operation is called soft access point (soft-AP). The maximum number of stations that can simultaneously be connected to the soft-AP is 4 by default.
 
 .. figure:: pictures/esp8266-soft-access-point.png
    :alt: ESP8266 operating in the Soft Access Point mode
--- a/doc/esp8266wifi/readme.rst
+++ b/doc/esp8266wifi/readme.rst
@@ -125,7 +125,7 @@ Soft Access Point
 ~~~~~~~~~~~~~~~~~
 
 An `access point (AP) <https://en.wikipedia.org/wiki/Wireless_access_point>`__ is a device that provides access to Wi-Fi network to other devices (stations)
-and connects them further to a wired network. ESP8266 can provide similar functionality except it does not have interface to a wired network. Such mode of operation is called soft access point (soft-AP). The maximum number of stations that can simultaneously be connected to the soft-AP is 4 by default.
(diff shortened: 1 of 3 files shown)


d-a-v:
I was telling that I could run 12 tcp connections simultaneously (layer3, in STA mode, with a PC), not having 12 wlan clients simultaneously connected to the Soft-AP (layer2).
I currently have not that much esp to test with. If there is a limitation, it should come from the internal firmware.
If there are no other mention in the nonos-sdk documentation, then you at least found an official statement from espressif.
IMHO, that information should also be stated in the sources (with reference).

aerlon:
OK. Thanks for the clarification. 

I agree, providing a reference is a good practice. The question is which reference… 

Looking at [the latest NonOS API reference](https://www.espressif.com/sites/default/files/documentation/2C-ESP8266_Non_OS_SDK_API_Reference__EN.pdf) one can find the following at page 143/166: `uint8 max_connection; // default 4, max 4`. Which would seem to go against a lot of the other sources I've found indicating a max of 8, including [the previous one from Espressif](https://bbs.espressif.com/viewtopic.php?f=46&t=481&p=1832&hilit=max_connection#p1832) . It is all very confusing.

devyte:
I think max wifi clients is 8, limited by the sdk. Espressif document 2C api reference should explain it. I can't look up the specific section right now, though, still limited to my broken phone.

devyte:
To clarify, the sdk max used to be 4, and it was increased to 8, but the max default of 4 was kept. Now, if you want more than 4, you have to set a higher max explicitly. At least that's how I remember it.

aerlon:
OK. If Espressif document 2C api reference is the following: https://www.espressif.com/sites/default/files/documentation/2C-ESP8266_Non_OS_SDK_API_Reference__EN.pdf
I was unable to find any other information than the one at page 143 I mentioned above.

However, I've since discovered [an Espressif forum thread](https://bbs.espressif.com/viewtopic.php?t=1894) where an employee states "ESP8266 soft-AP can connected to 8 stations at most." . I'm guessing page 143 in the NonOS API reference has not been updated since they increased the max.

The PR has been updated to reflect that valid values for `max_connection` is 1 to 8, and a reference has been added.

d-a-v:
Thanks !

While you are at it, please check this recent #5082 PR
(not really wanting to be picky :)

[before](https://github.com/esp8266/Arduino/blob/master/doc/esp8266wifi/soft-access-point-class.rst#softap)
[after](https://github.com/aerlon/Arduino/blob/ap_documentation_update/doc/esp8266wifi/soft-access-point-class.rst#softap)
[rendered(before)](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/soft-access-point-class.html#softap)

devyte:
Ok so espressif's doc needs to be updated :P 
There is mention in the sdk releases changelog that the max allowed is 8 stations.

PR #5089, opened 2018-08-28
ClientContext (tcp) updates

d-a-v:
* make Nagle work properly (default is unchanged: disabled = NoDelay enabled)
* add `client.setSync(bool)` (= "always flush" mode)
  slower when enabled but allocates less memory and data are always flushed
* `client.stop()` implicitely calls `.flush()` (flush has a small timeout anyway)
* add static `WiFiClient::setDefaultSync(bool)`/`::setDefaultNoDelay(bool)`
* internal: remove `PuSH` flag from data packet


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -427,40 +449,47 @@ class ClientContext
             return false;
         }
 
-        size_t left = _datasource->available();
-        size_t can_send = tcp_sndbuf(_pcb);
-        if (_pcb->snd_queuelen >= TCP_SND_QUEUELEN) {
-            can_send = 0;
-        }
-        size_t will_send = (can_send < left) ? can_send : left;
         DEBUGV(":wr %d %d %d\r\n", will_send, left, _written);
```
earlephilhower: The debug lines fail to compile due to them trying to dump variables that were just removed (`left`)

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -427,40 +449,47 @@ class ClientContext
...
+            // Notes:
+            //   PUSH is for peer, telling to give data to user app as soon as received
+            //   PUSH "may be set" when sender has finished sending a meaningful data block
+            //   Nagle is for delaying local data, to send less/bigger packets
+            uint8_t flags = TCP_WRITE_FLAG_MORE; // do not tcp-PuSH
+            if (!_sync)
+                // user data will not stay in place when data are sent but not acknowledged
+                flags |= TCP_WRITE_FLAG_COPY;
+            err_t err = tcp_write(_pcb, buf, next_chunk_size, flags);
+            DEBUGV(":wrc %d %d %d\r\n", next_chunk_size, will_send, (int)err);
```
earlephilhower: `will_send` undefined, too

earlephilhower:
Functionally, when compiled w/o debugging, this patch doesn't seem to cause any problems with BearSSL or axTLS in simple testing (bssl_validatiaon, https_request).

libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -42,6 +42,8 @@ extern "C"
 #include "c_types.h"
 
 uint16_t WiFiClient::_localPort = 0;
+bool WiFiClient::_defaultNoDelay = true;
```
devyte: I suggest a big fat comment here about what the four combinations of nodelay and sync do, because I can't for the life of me keep them straight in my head.

earlephilhower:
@d-a-v, there's something weird in the tree and it's failing with the EthernetClient example builds. Very odd!  Maybe it was using the ClientContext and something got broken?

libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -264,19 +311,20 @@ size_t WiFiClient::peekBytes(uint8_t *buffer, size_t length) {
     return _client->peekBytes((char *)buffer, count);
 }
 
-void WiFiClient::flush()
+bool WiFiClient::flush(unsigned int maxWaitMs)
 {
-    if (_client)
-        _client->wait_until_sent();
+    return !_client || _client->wait_until_sent(maxWaitMs?: WIFICLIENT_MAX_FLUSH_WAIT_MS);
```
earlephilhower: I get the shorthand, but using the early-termination feature to avoid an if-then could be confusing

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -175,23 +175,19 @@ void WiFiClientSecure::setBufferSizes(int recv, int xmit) {
...
 
-void WiFiClientSecure::stop() {
-  flush();
-  if (_client) {
-    _client->wait_until_sent();
-    _client->abort();
-  }
-  WiFiClient::stop();
+bool WiFiClientSecure::stop(unsigned int maxWaitMs) {
+  bool ret = WiFiClient::stop(maxWaitMs); // calls our virtual flush()
```
earlephilhower: I see `::stop` calls `::flush` so we're OK there.  Why I did `_client->abort()` as well, I couldn't tell you.  LGTM.

earlephilhower:
The build problem is the changes to the `flush` and `stop` signature.  Need to update the EthernetClient sources to have the new timeout parameter w/defaults...

d-a-v:
doc is visible there: [client](https://github.com/d-a-v/Arduino/blob/ClientContext/doc/esp8266wifi/client-class.rst) and [server](https://github.com/d-a-v/Arduino/blob/ClientContext/doc/esp8266wifi/server-class.rst).

PR #5091, opened 2018-08-29
ESP8266HTTPClient: allow getString() more than once

### LINKED ISSUE #4951: HTTPClient.getString() returns payload on first call only and returns empty after
#### Platform

- Hardware:				Wemos  mini
- Core Version:			latest git
- Development Env:		Sloeber eclipse
- Operating System:		Windows

### Settings in IDE

- Module:  			Wemos D1 mini 
- Flash Size: 		4MB
- lwip Variant:    v2 Lower Memory
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   921600

### Problem Description

Might be working as intended but took me some time to find out when I was creating a small test program.

When the HTTPClient.GET() has finished, the payload can be retrieved with HTTPClient.getString() returns the payload.
However, when calling multiple times, only the first will return the payload, each subsequent call returns empty string.

```cpp
#include "LocalDefines.h"
#include <ESP8266WiFi.h>
#include "ESP8266HTTPClient.h"

HTTPClient htpc;

void setup()
{
	Serial.begin(115200);

	Serial.printf("Initiating WiFi\r\n");
	WiFi.mode(WIFI_STA);
	Serial.printf("Wifi mode set\r\n");
    WiFi.begin(ssid,password);
	    delay(1000);
	    Serial.println("Connecting to WiFi..");
	 }
	Serial.printf("IP = %s\r\n",WiFi.localIP().toString().c_str());


	htpc.begin("http:/www.google.nl");

	if (htpc.GET() > 0)
	{
		String first = htpc.getString();
		String second = htpc.getString();

		Serial.printf("------------------------------\r\n");
		Serial.printf("First len = %d, txt = %s\r\n",first.length(),first.c_str());
		Serial.printf("------------------------------\r\n");
		Serial.printf("Second len = %d, txt = %s\r\n",second.length(),second.c_str());
	}
}

void loop()
{
}

```

Results in Serial output.

```
Initiating WiFi
Wifi mode set
Connecting to WiFi..
Connecting to WiFi..
IP = 10.0.0.211
------------------------------
First len = 219, txt = <HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
 <TITLE>301 Moved</TITLE></HEAD><BODY>
                                      <H1>301 Moved</H1>
                                                        The document has moved
                                                                              <A HREF="https://www.google.nl/">here</A>.
</BODY></HTML>

------------------------------
Second len = 0, txt =

```

yoursunny:
fixes #4951

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -33,7 +33,7 @@
 class TransportTraits
 {
 public:
-    virtual ~TransportTraits() 
+    virtual ~TransportTraits()
     {
     }
 
@@ -126,6 +126,7 @@ void HTTPClient::clear()
     _returnCode = 0;
     _size = -1;
     _headers = "";
+    _payload.reset();
 }
 
 
@@ -281,6 +282,16 @@ bool HTTPClient::begin(String host, uint16_t port, String uri, const uint8_t htt
  * called after the payload is handled
  */
 void HTTPClient::end(void)
+{
+    disconnect();
+    clear();
+}
+
+/**
+ * disconnect
+ * close the TCP socket
+ */
+void HTTPClient::disconnect()
 {
     if(connected()) {
         if(_tcp->available() > 0) {
@@ -734,7 +745,7 @@ int HTTPClient::writeToStream(Stream * stream)
         return returnError(HTTPC_ERROR_ENCODING);
     }
 
-    end();
+    disconnect();
     return ret;
 }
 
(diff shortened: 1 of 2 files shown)


yoursunny:
I choose to store `unique_ptr<StreamString>` rather than `String` on the `HTTPClient` class, so that `getString` does not need to copy `StreamString` into `String`.
Program size of https://github.com/esp8266/Arduino/issues/4951#issue-343413924 sketch increases from 268400 bytes to 268496 bytes due to this change.

devyte:
Agreed on making clear public, sounds like a good enough compromise.

devyte:
Refactoring basics :) please go ahead!

yoursunny:
"end = disconnect + clear" design has been implemented.

PR #5116, opened 2018-09-10
Make unique sections for ICACHE_* variables

earlephilhower:
Similar to PROGMEM changes, make the name of each ICACHE_* variable/fuction
unique to avoid issues with section conflicts.

Fixes #5115

PR #4618, opened 2018-04-07
remove autogenerated eagle.app.v6.common.ld and git-ignore it

d-a-v:

PR #5117, opened 2018-09-10
Make unique sections for ICACHE_* variables

### LINKED ISSUE #5115: Code does not compile when compiling without optimisations -Og instead of optimize small -Os
*edit: quoting updated*
Hi
When trying to make a hardware debug platform.txt I bumped into the compilation problem below.
```
'Building file: C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\cores\esp8266\core_esp8266_postmortem.c'
'Starting C compile'
"C:\eclipse\arduinoPlugin\packages\esp8266\tools\xtensa-lx106-elf-gcc\1.20.0-26-gb404fb9-2/bin/xtensa-lx106-elf-gcc" -D__ets__ -DICACHE_FLASH -U__STRICT_ANSI__ "-IC:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2/tools/sdk/include" "-IC:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2/tools/sdk/lwip2/include" "-IC:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2/tools/sdk/libc/xtensa-lx106-elf/include" "-IL:/test/runtime-EclipseApplication_eng/esp/ff/core" -c -Wall -Wextra   -DDEBUG -Og -ggdb -g -Wpointer-arith -Wno-implicit-function-declaration -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals -falign-functions=4 -MMD -std=gnu99 -ffunction-sections -fdata-sections -DF_CPU=160000000L -DLWIP_OPEN_SRC -DTCP_MSS=536 -DDEBUG_ESP_PORT=Serial  -DARDUINO=10802 -DARDUINO_ESP8266_GENERIC -DARDUINO_ARCH_ESP8266 -DARDUINO_BOARD="\"ESP8266_GENERIC\"" -DLED_BUILTIN=9  -DESP8266   -I"C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\cores\esp8266" -I"C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\variants\generic" -I"C:\Users\jan\git\hardware\jantje\ESP8266\libraries\gdb" -MMD -MP -MF"core\core\core_esp8266_postmortem.c.d" -MT"core\core\core_esp8266_postmortem.c.o" -D__IN_ECLIPSE__=1 "C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\cores\esp8266\core_esp8266_postmortem.c"  -o  "core\core\core_esp8266_postmortem.c.o"
C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\cores\esp8266\core_esp8266_postmortem.c:83:23: error: istr causes a section type conflict with iram_read_byte
     static const char istr[] ICACHE_RAM_ATTR = (str); \
                       ^
C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\cores\esp8266\core_esp8266_postmortem.c:190:5: note: in expansion of macro 'ets_printf_P'
     ets_printf_P("<<<stack<<<\n");
     ^
C:\eclipse\arduinoPlugin\packages\esp8266\hardware\esp8266\2.4.2\cores\esp8266\core_esp8266_postmortem.c:66:29: note: 'iram_read_byte' was declared here
 static char ICACHE_RAM_ATTR iram_read_byte (const char *addr) {
                             ^
core\core\subdir.mk:356: recipe for target 'core\core\core_esp8266_postmortem.c.o' failed
make: *** [core\core\core_esp8266_postmortem.c.o] Error 1
```

The problem lies into the fact that the define ICACHE_RAM_ATTR  is used 2 times for a different variable.
Apparently optimized for small optimizes this to 1 variable which makes this problem is not visible when using -Os. Obviously optimizing for debugging does not allow this.

This can be fixed by changing the line
```
    static const char istr[] ICACHE_RAM_ATTR = (str);
```
to
```
    static const char istr[] __attribute__((section(".iram.text.2"))) = (str);
```

I didn't want to make this a push request because I am not sure "the change is an acceptable solution" as yesterday I had not heard about something called
` __attribute__((section(XXX)))`

Best regards
Jantje

jjsuwa:
Similar to PROGMEM changes, make the name of each ICACHE_* variable/fuction
unique to avoid issues with section conflicts.

Fixes #5115 

Added `.irom0.text.*` and `.iram.text.*` to ldscript.

### DIFF
--- a/tools/sdk/include/c_types.h
+++ b/tools/sdk/include/c_types.h
@@ -84,9 +84,11 @@ typedef enum {
 #define SHMEM_ATTR
 
 #ifdef ICACHE_FLASH
-#define ICACHE_FLASH_ATTR   __attribute__((section(".irom0.text")))
-#define ICACHE_RAM_ATTR     __attribute__((section(".iram.text")))
-#define ICACHE_RODATA_ATTR  __attribute__((section(".irom.text")))
+#define __ICACHE_STRINGIZE_NX(A) #A
+#define __ICACHE_STRINGIZE(A) __ICACHE_STRINGIZE_NX(A)
+#define ICACHE_FLASH_ATTR   __attribute__((section("\".irom0.text." __FILE__ "." __ICACHE_STRINGIZE(__LINE__) "." __ICACHE_STRINGIZE(__COUNTER__) "\"")))
+#define ICACHE_RAM_ATTR     __attribute__((section("\".iram.text." __FILE__ "." __ICACHE_STRINGIZE(__LINE__) "." __ICACHE_STRINGIZE(__COUNTER__) "\"")))
+#define ICACHE_RODATA_ATTR  __attribute__((section("\".irom.text." __FILE__ "." __ICACHE_STRINGIZE(__LINE__) "." __ICACHE_STRINGIZE(__COUNTER__) "\"")))
 #else
 #define ICACHE_FLASH_ATTR
 #define ICACHE_RAM_ATTR
--- a/tools/sdk/ld/eagle.app.v6.common.ld.h
+++ b/tools/sdk/ld/eagle.app.v6.common.ld.h
@@ -127,7 +127,7 @@ SECTIONS
     *libwpa.a:(.literal.* .text.*)
     *libwpa2.a:(.literal.* .text.*)
     *libwps.a:(.literal.* .text.*)
-    *(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom.text .irom.text.*)
+    *(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom0.text.* .irom.text .irom.text.*)
     _irom0_text_end = ABSOLUTE(.);
     _flash_code_end = ABSOLUTE(.);
   } >irom0_0_seg :irom0_0_phdr
@@ -164,8 +164,8 @@ SECTIONS
     *(.init.literal)
     *(.init)
     *(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.*)
-    *.cpp.o(.iram.text)
-    *.c.o(.iram.text)
+    *.cpp.o(.iram.text .iram.text.*)
+    *.c.o(.iram.text .iram.text.*)
 #ifdef VTABLES_IN_IRAM
     *(.rodata._ZTV*) /* C++ vtables */
 #endif


devyte:
@jjsuwa how is this different from #5116?

earlephilhower:
I think it's the same, now.  It seems he committed the fix he suggested and re submitted the patch as a new one.  He did find it, so I don't mind if you merge this and drop mine. It would have been a bear to find if he didn't!  I'm on my phone so can't double check the PR, so please use due diligence...

jjsuwa:
> @jjsuwa how is this different from #5116?

ldscript.

wemos:
@jjsuwa Could you test it and upload compiled code to board?
I test on simple "Blink", the code compiled successfully, but upload to esp8266 board,  it can't work.

It work well on previous commit https://github.com/esp8266/Arduino/commit/91519309d000c65c283c6721ae58d4bf708bf719

I think it may be some wrong with this commit.

jjsuwa:
> @wemos Could you test it and upload compiled code to board?

- board: ESP-WROOM-02 breakout PCB (Akizuki AE-ESP-WROOM-02)
- commit: 622569cde2b5af2d1b50a86130346751a53cf5c8
```
#include <Ticker.h>

Ticker ticker;  // uses ICACHE_RAM_ATTR internally

// ICACHE_FLASH_ATTR... used by the framework everywhere

#define TICK_FREQENCY 10

void setup() {
  pinMode(LED_BUILTIN, OUTPUT);
  digitalWrite(LED_BUILTIN, LOW);
  Serial.begin(115200);
  delay(250);
  Serial.print(F("\n\ntest sketch...\n"));  // uses ICACHE_RODATA_ATTR
  ticker.attach_ms((1000 / 2) / TICK_FREQENCY, []() {
    static int state = 0;
    digitalWrite(LED_BUILTIN, state ^= 1);
  });
}

void loop() {
}
```

The above seems to work properly (IO2 pin generates 10Hz).

earlephilhower:
@jjsuwa , @wemos , @devyte - I can reproduce the failure @wemos shows.  With this change the blink.ino example fails to work, as well as some BearSSL tests I've tried, too.

There is some kind of interaction with the linker change and the way the final BIN gets generated and you end up in an infinite reset loop.

Until that's figured out, I say we remove these changes since it's breaking things.  I've put in a PR #5130 that reverses this.

When we figure it out, we can re-apply with whatever add'l fixes needed.  My guess is the image layout got moved and the init code is not ending up at the right spot (since link is clean it's not like we're missing the function...it's just not where it should be).

jjsuwa:
> @earlephilhower - I can reproduce the failure @wemos shows. With this change the blink.ino example fails to work,

But i can't reprod on `libraries\esp8266\examples\Blink\Blink.ino`... mysterious.

- commit: 622569cde2b5af2d1b50a86130346751a53cf5c8
- IDE settings: Generic ESP8266 Module, 160 MHz, Flash, nodemcu, 26 MHz, 80 MHz, QIO, 4M(1M SPIFFS), 2, v2 Lower Memory, Disabled, NoAssert-NDEBUG, Only Sketch, 921600

earlephilhower:
I believe you, @jjsuwa , but it seems like for other configurations of debug options/modules/etc. it fails.  Need to get to the bottom of it and add whatever add'l tweaks are required to make it stable on all settings.

I'm running D1Mini, 4M no spiffs, Debug port=Serial, Debug level=all, lwip2lomem, vtables=flash, 80mhz.

It may be related to debugging, then.  Can you try your config w/debug port=serial, debug=all?

jjsuwa:
> @earlephilhower - Can you try your config w/debug port=serial, debug=all?

- 4M(no SPIFFS)
- SSL+TLS_MEM+HTTP_CLIENT+HTTP_SERVER+CORE+WIFI+HTTP_UPDATE+UPDATER+OTA+OOM
- Serial

but no difference (`Blink.ino` runs properly), hmm...

jjsuwa:
fact:

- Both `.irom0.text*` and `.iram.text*` affect code arrangement but `.irom.text*` not (const data).

hypothesis:

- Distance from `.irom0.text.*` to other `.irom*`s (or `text*`s) may be changed worse (possibly too far)

proposal experiment:

- apply the below to 622569cde2b5af2d1b50a86130346751a53cf5c8
```
-     *(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom0.text.* .irom.text .irom.text.*)
+     *(.irom0.literal .irom.literal .irom.text.literal .irom0.text.* .irom0.text .irom.text .irom.text.*)
```
```
-    *.cpp.o(.iram.text .iram.text.*)
+    *.cpp.o(.iram.text.* .iram.text)
```
```
-    *.c.o(.iram.text .iram.text.*)
+    *.c.o(.iram.text.* .iram.text)
```

earlephilhower:
Prior WiFi state may also be part of the issue.  My prior testing tests executed `WiFi.begin(bad ssid, bad pass);` and I can see in debug logs it's trying to reconnect (and failing), even in the silly `blink.ino` sketch.  That's one thing that's not reset by upload or code changes..."hidden state."

earlephilhower:
Looks like you posted while I was typing, didn't see it.  Let me try your changes and see.  The linker should fix up any short branches that break.  Or at least give an error.  But maybe not...

jjsuwa:
> @earlephilhower - That's one thing that's not reset by upload or code changes..."hidden state."

OK, `Erase Flash: "All Flash Contents"`... but works fine. ;)

earlephilhower:
No joy yet.

Tools->Full flash erase:  Same failure
Change in order of linker patterns:  Same failure

The only difference I think is left is the generic vs. WemosD1 variant.  I'll need to dig into the variant code a bit later.

jjsuwa:
Oh, Windows Update bashes me... time to shut my PC down :)

conclusion:

- These results suggest there is a hidden/potential problem with ICACHE_* link mechanism, I guess.

earlephilhower:
I rebuilt and tried blink by swapping out the ICACHE_* macros one by one.  ICACHE_RODATA_ATTR can be uniqified, but uniquifying either the _RAM_ or _FLASH_ cause the crash errors.

jjsuwa:
Interesting...

In my envs, uniqified three `ICACHE_`s w/o `.irom0.text.*` and `.iram.text.*` (such as #5116) cause infinite reset loop. That is why I committed #5117.

earlephilhower:
I actually just used the `#defines` in the PR to test (ie I just replaced the original section attribute with these new unique names one by one).

earlephilhower:
Just found the problem here, @jjsuwa.  There was a generated `eagle.app.v6.common.ld` file present in my `tools/sdk/ld` directory because my Arduino core was installed prior to #4618 .

@d-a-v, you git-removed the "eagle.app.v6.common.ld" but it seems like @wemos and I and others still have it around, even after git updates.  It's now generated in the `/tmp/arduino_build_*` tree, so the original patch works fine *unless* you have the old obsolete one in the `ld` dir, which people with long-running dev systems will.  Do you have a good idea how to get around this?  Should we delete any old .ld file there unconditionally?  Need to merge whatever way we use to fix this with any update so the problem @wemos and I had won't spread to others.

d-a-v:
@earlephilhower it must have been tough to find this one!
I don't see any other way than removing it from platform.txt scripts.
The generated one is in build dir (/tmp on linux).

jjsuwa:
> @earlephilhower - Just found the problem here, @jjsuwa.

Good job :)

PR #5119, opened 2018-09-10
Add files via upload

sadykowich:
This file need to using this library in Eclipes CDT Arduino.
Without this file the library manager doesn't listing this library.


sadykowich:
Ok, i updated.

d-a-v:
will merge after CI is re-triggered (reason: savannah was unreachable)

PR #5120, opened 2018-09-11
Fix connection options and update github pubkey

### LINKED ISSUE #5086: bearssl_validation example invalid public key
This public key seems to be invalid:
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino#L134-L140

This is valid public key:
```
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAy+3Up8qBkIn/7S9AfWlH
Od8SdXmnWx+JCIHvnWzjFcLeLvQb2rMqqCDL5XDlvkyC5SZ8ZyLITemej5aJYuBv
zcKPzyZ0QfYZiskU9nzL2qBQj8alzJJ/Cc32AWuuWrPrzVxBmOEW9gRCGFCD3m0z
53y6GjcmBS2wcX7RagqbD7g2frEGko4G7kmW96H6dyh2j9Rou8TwAK6CnbiXPAM/
5Q6dyfdYlHOCgP75F7hhdKB5gpprm9A/OnQsmZjUPzy4u0EKCxE8MfhBerZrZdod
88ZdDG3CvTgm050bc+lGlbsT+s09lp0dgxSZIeI8+syV2Owt4YF/PdjeeymtzQdI
wQIDAQAB
```

earlephilhower:
As part of the "clear connection configuration for reused objects"
patch, a ::stop would reset the self-signed, trust anchors, etc.
WiFiClient, unfortunately, calls ::stop as part of the connection
process, so all of these settings were lost.

Now only clear the connection settings on ::stop if we've already
been connected.

Also update the github public key which changed yet again.

Fixes #5086

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
@@ -179,7 +179,10 @@ void WiFiClientSecure::stop() {
     _client->abort();
   }
   WiFiClient::stop();
-  _clearAuthenticationSettings();
+  // Only if we've already connected, clear the connection options
+  if (_handshake_done) {
+    _clearAuthenticationSettings();
+  }
   _freeSSL();
 }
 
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -179,7 +179,10 @@ void WiFiClientSecure::stop() {
     _client->abort();
   }
   WiFiClient::stop();
-  _clearAuthenticationSettings();
+  // Only if we've already connected, clear the connection options
+  if (_handshake_done) {
```
devyte: What happens if I set up a connection, attempting to connect fails (_handshake_done never becomes true), and then decide to abort and call stop()? Won't the flags remain?
What if the certs are then changed and a different connection is attempted? won't the previous flags be still there?

PR #5121, opened 2018-09-11
Move SSLContext to its own header

### LINKED ISSUE #3648: SSLContext builtin to the WifiClientSecure.cpp file makes it harder to extend WifiClientSecure class
I wanted to make some very specific changes to the WifiClientSecure class, but since SSLContext is defined completely inside the WifiClientSecure.cpp file it makes it significantly harder to deal with it. 

It would be a great help to define the SSLContext class in a header so I don't have to reimplement everything if I want to extend the class.

earlephilhower:
Simple refactor to make WiFiClientSecureAxTLS use an external header to
define its SSLContext, just as it does for several other classes.

Fixes #3648


devyte:
I would argue that it's a bad idea to extend the axtls-based WiFiClientSecure class, and that @schlaegerz should first migrate to bearssl, and then make any specific changes he needs on that.
However, since the effort is already spent in the PR, I'll approve.

schlaegerz:
Yep I plan to move to Bearssl, this was before that even existed. I end up needing to make a few changes to Bearssl to get what I want, but I haven't thought through exactly what the right things are

PR #5129, opened 2018-09-12
Create library.properties

sadykowich:
This file need to using this library in Eclipes CDT Arduino.
Without this file the library manager doesn't listing this library.


d-a-v:
Thanks!
One PR for all three files would be much better with CI

PR #5133, opened 2018-09-13
implement EspClass::getFreeContStack method

### LINKED ISSUE #5118: Function cont_get_free_stack is not available anymore
The new release 2.4.2 includes this change:

 . Move continuation stack from .bss onto sys stack, saving 4kB of RAM, if WPS is not needed (#4622)

But there is a side effect: the function **cont_get_free_stack(&g_cont)** [cont.h] does not work because the variable **g_cont** is not defined anymore (file core_esp8266_main.cpp).

My programs include stack statistics but I am unable to maintain this feature if I use the new release 2.4.2.

devyte:
Fixes #5118

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -25,6 +25,7 @@
 #include "interrupts.h"
 #include "MD5Builder.h"
 #include "umm_malloc/umm_malloc.h"
+#include "cont.h"
 
 extern "C" {
 #include "user_interface.h"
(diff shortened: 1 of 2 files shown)


cores/esp8266/Esp.cpp
```diff
@@ -177,6 +178,12 @@ uint16_t EspClass::getMaxFreeBlockSize(void)
     return umm_max_block_size();
 }
 
+uint32_t EspClass::getFreeContStack()
+{
+    extern cont_t* g_pcont;
```
d-a-v: This declaration is already done in `cont.h`

As a general question, why don't we write these trivial functions directly inside the class ?
This would have the effect of reducing the final image size if they are not used.

d-a-v: Damn right !
I was totally unaware of that and I thought that was the LTO feature that we don't have.
So I uselessly separated some code from Esp.cpp namely Esp-frag.cpp and Esp-version.cpp.
Thanks for opening my eyes, I will seek details on the differences between these two gcc/ld features.

PR #5142, opened 2018-09-16
Added naive content length to upload struct

RickeyWard:
Content length provides context into the size of the upload, it's not the exact form body size because it's off by the request header and other request info, but it was already parsed by the library and provides little to no additional overhead to pass on for use in request handlers.
This allows  you the opportunity to abort if the file will be too large, or approximate the upload percentage. The only impact of this change is extending the HTTPUpload struct by a size_t, and the one-time cost of assigning it.


RickeyWard:
What would it take to get this merged?

PR #5146, opened 2018-09-18
Add full debug builds for all tests

### LINKED ISSUE #5143: Add a full debug build of some number of tests to CI
The debug macros sometimes cause trouble when the main code is changed, since they are NOPs until debugging is enabled.

To ensure changes don't break debug builds, we should add a build of some representative tests with full debugging enabled to the CI.

Probably don't need to build all with full debug, but one or two examples from each major subsystem.

earlephilhower:
In order to ensure the -DEBUG builds don't fail with checked-in code, add
two jobs which build even/odd with full debug flags to Travis.

Fixes #5143

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -118,6 +118,9 @@ function install_ide()
     # Set custom warnings for all builds (i.e. could add -Wextra at some point)
     echo "compiler.c.extra_flags=-Wall -Werror $debug_flags" > esp8266/platform.local.txt
     echo "compiler.cpp.extra_flags=-Wall -Werror $debug_flags" >> esp8266/platform.local.txt
+    echo -e "\n----platform.local.txt----"
+    cat esp8266/platform.local.txt
+    echo -e "\n----\n"
     cd esp8266/tools
     python get.py
     export PATH="$ide_path:$core_path/tools/xtensa-lx106-elf/bin:$PATH"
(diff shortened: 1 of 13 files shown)


earlephilhower:
As CI is "free" for now, no reason not to build everything with debugging enabled.  If it becomes an issue, it is not a big change to pick specific tests to compile (but unless needed, I'd rather not try and guess what's needed for good coverage).

tests/common.sh
```diff
@@ -103,16 +103,24 @@ function install_ide()
...
     ln -s $core_path esp8266
+    local debug_flags=""
+    if [ "$debug" = "debug" ]; then
+        debug_flags="-DDEBUG_ESP_PORT=Serial -DDEBUG_ESP_SSL -DDEBUG_ESP_TLS_MEM -DDEBUG_ESP_HTTP_CLIENT -DDEBUG_ESP_HTTP_SERVER -DDEBUG_ESP_CORE -DDEBUG_ESP_WIFI -DDEBUG_ESP_HTTP_UPDATE -DDEBUG_ESP_UPDATER -DDEBUG_ESP_OTA -DDEBUG_ESP_OOM"
+    fi
     # Set custom warnings for all builds (i.e. could add -Wextra at some point)
-    echo "compiler.c.extra_flags=-Wall -Werror" > esp8266/platform.local.txt
-    echo "compiler.cpp.extra_flags=-Wall -Werror" >> esp8266/platform.local.txt
+    echo "compiler.c.extra_flags=-Wall -Werror $debug_flags" > esp8266/platform.local.txt
+    echo "compiler.cpp.extra_flags=-Wall -Werror $debug_flags" >> esp8266/platform.local.txt
```
d-a-v: should `-Wextra`have its place here too ?

earlephilhower: Sure.  It wasn't in the original "Add all warnings" patch but we can pull it in here...

PR #5151, opened 2018-09-19
Allow cipher specification for BearSSL

### LINKED ISSUE #5110: BearSSL takes more than twice as long for handshake as axTLS
This is a follow-up to https://github.com/esp8266/Arduino/issues/4738#issuecomment-418855207

### Claim
I observed that BearSSL in standard configuration takes more than twice as long for a handshake as axTLS.

### Basic Infos

#### Platform

- Hardware:				WeMos D1 Mini Pro
- Core Version:			2.4.2
- Operating System:		macOS

### Sample sketch
The below sketch is inspired by what @artua presented at #4738 but it's much simpler. I feel that when just starting out measuring / comparing performance you shouldn't collect too much data at once as you might not see the wood for the trees. Once you identify a suspect you then adjust the rig accordingly and collect more data around the suspected culprit.

The program establishes a connection to a SSL/TLS protected host 50 times with a delay of 1.5s. For every attempt it reports total failure/success count, accumulated handshake time, average handshake time and free heap _at the end_. No server certificates or fingerprints are verified!

Around line 40 you define whether to use BearSSL (with `setInsecure()`) or axTLS.

```c++
#include <ESP8266WiFi.h>

#define HOST "thingpulse.com"
#define PORT 443
#define WIFI_SSID "ssid"
#define WIFI_PWD "password"

const char* ssid = WIFI_SSID;
const char* password = WIFI_PWD;

int numberOfSamples = 50;
int totalCounter = 0;
int failCounter = 0;
uint64_t totalDuration = 0;
void setup () {
  Serial.begin(115200);
  Serial.println("");
  WiFi.begin(ssid, password);

  Serial.print("Connecting to Wifi");
    delay(300);
    Serial.print(".");
  }
  Serial.println("done.");
}


void loop() {
  if (totalCounter < numberOfSamples) {
    bool success = connectToHost();
    delay(1500);
  }
}

bool connectToHost() {
  uint64_t startTime = millis();
//  BearSSL::WiFiClientSecure client;
//  client.setInsecure();
  WiFiClientSecure client;

  boolean success = false;
  totalCounter++;
  if (client.connect(HOST, PORT)) {
    // Serial.printf("Connection to %s success.\n", HOST);
    success = true;
  } else {
    failCounter++;
    // Serial.printf("Connection to %s failed.\n", HOST);
  }
  client.flush(); // not actually needed
  client.stop();
  totalDuration += millis() - startTime;
  Serial.printf("Failed: %d, Total: %d (%f %%), Total Dur: %d, Avg. Dur: %f, Heap: %d\n", 
    failCounter, 
    totalCounter, 
    (100.0 * failCounter / totalCounter), 
    totalDuration,
    1.0 * totalDuration / totalCounter, 
    ESP.getFreeHeap());
  return success;
}
```

### Test results

I ran the above sketch with all combinations of BearSSL/axTLS, lwIP v2 lower memory/higher bandwidth and 80/160MHz on core 2.4.2. So, this makes 2^3 * 50 samples.

The summary is as follows:
0 connection failures. That's remarkable as this whole endeavour started out with this issue https://stackoverflow.com/q/52143894.

**BearSSL**
- lwIP v2 lower memory, 80 MHz: ~860ms
- lwIP v2 higher bandwidth, 80 MHz: ~845ms
- lwIP v2 lower memory, 160 MHz: ~510ms
- lwIP v2 higher bandwidth, 160 MHz: ~500ms

**axTLS**
- lwIP v2 lower memory, 80 MHz: ~360ms
- lwIP v2 higher bandwidth, 80 MHz: ~350ms
- lwIP v2 lower memory, 160 MHz: ~250ms
- lwIP v2 higher bandwidth, 160 MHz: ~240ms

-> BearSSL takes more than twice as long
-> As expected higher bandwidth has close to zero impact as only a socket connection with handshake was established. No resources were loaded from the host.

The detailed result is in the below PDF.
[BearSSL-vs-axTLS.pdf](https://github.com/esp8266/Arduino/files/2354505/BearSSL-vs-axTLS.pdf)

Over at #4738 @earlephilhower stated that

> BearSSL often negotiates a more secure but slower running cypher... limiting the receiving end's connection capabilities in your SSL configuration

I would first have to look into that. Not familiar with SSL config yet.

earlephilhower:
BearSSL has many more ciphers than axTLS, but they are more compute intensive
and slower.  Add an option to use only the same, limited security, axTLS ciphers
as well as allow users to specify any suite of ciphers they want using standard
BearSSL formats.

Fixes #5110

### DIFF
--- a/libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino
+++ b/libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino
@@ -193,11 +193,12 @@ BearSSL does verify the notValidBefore/After fields.
   fetchURL(&client, host, port, path);
 }
 
-void fetchAxTLS() {
+void fetchFaster() {
   Serial.printf(R"EOF(
 The ciphers used to set up the SSL connection can be configured to
-be the same as axTLS. They are faster, but less secure, so if you care
-about security you won't want to do this.
+only support faster but less secure ciphers.  If you care about security
+you won't want to do this.  If you need to maximize battery life, these
+may make sense
 )EOF");
   BearSSL::WiFiClientSecure client;
   client.setInsecure();
@@ -206,10 +207,10 @@ about security you won't want to do this.
   uint32_t delta = millis() - now;
   now = millis();
   client.setInsecure();
-  client.setAxTLSCiphers();
+  client.setCiphersLessSecure();
   fetchURL(&client, host, port, path);
   uint32_t delta2 = millis() - now;
-  Serial.printf("Using more secure: %dms\nUsiing axTLS ciphers: %dms\n", delta, delta2);
+  Serial.printf("Using more secure: %dms\nUsing less secure ciphers: %dms\n", delta, delta2);
 }
 
 void setup() {
@@ -239,7 +240,7 @@ void setup() {
   fetchSelfSigned();
   fetchKnownKey();
   fetchCertAuthority();
-  fetchAxTLS();
+  fetchFaster();
 }
 
 
--- a/libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino
+++ b/libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino
@@ -193,6 +193,25 @@ BearSSL does verify the notValidBefore/After fields.
   fetchURL(&client, host, port, path);
 }
 
+void fetchAxTLS() {
+  Serial.printf(R"EOF(
+The ciphers used to set up the SSL connection can be configured to
+be the same as axTLS. They are faster, but less secure, so if you care
+about security you won't want to do this.
+)EOF");
+  BearSSL::WiFiClientSecure client;
+  client.setInsecure();
+  uint32_t now = millis();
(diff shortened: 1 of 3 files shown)


artua:
@earlephilhower Can't use setCipher:
```
  client.setCipher();
```

```
_'class BearSSL::WiFiClientSecure' has no member named 'setCipher'_
```

earlephilhower:
client.setCipher**s**.  Plural.  Easy mistake to make.

artua:
@earlephilhower Yeah. 
But not my mistake. :)

https://github.com/esp8266/Arduino/pull/5151/commits/af699f458099b0612bdaa55381805cb99d3f91ca
``` 
For C++ afficionados, allow std::vectors to be passed in to the setCipher()
routine.
```

marcelstoer:
I'm a bit lost figuring out what your release planning is (apart from the major milestones you track on GitHub). Is there any chance fixes like this one could go into a 2.4.3?

devyte:
Hi @marcelstoer 
our current [release model](https://github.com/esp8266/Arduino/tree/master/package#release-model tree) is linear, not branched. That means there won't be a 2.4.3, and there are no backports of fixes.

Issue #5148, opened 2018-09-18
Stack usage in WiFi.onStationModeGotIP too high, probably overflow


toomasz:
### Basic Infos

- [ X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ X] I have tested that the issue is present in current master branch (aka latest git).
- [ X] I have searched the issue tracker for a similar issue.
- [ X] If there is a stack dump, I have decoded it.
- [ X] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[1a44f79]
- Development Env:		[VS Micro]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[NodeMCU 1.0 (ESP-12E Module]
- Flash Mode:			[qio|dio|other]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

I was experiencing random crashes in my project which is using ESPAsyncTCP and async-mqtt-library so i decided to check stack usage in event handlers on mqtt library, especially in publish received. Found out that free stack size is below zero. I was getting results like current free stack = -4000

I also checked stack usage in onStationModeGotIP event handler and saw free stack is below zero.

Im not sure if this method of measuring free stack is good enough. In setup method is shows areound 4Kb of free stack which seems to be correct.

Is it a bug? Does 'negative' free stack means that stack overlaps with heap region?

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <ESP8266WiFi.h>

#define WIFI_SSID "toya88806789"
#define WIFI_PASSWORD "56744461"


WiFiEventHandler wifiConnectHandler;

extern "C"
{
#include <cont.h>
	extern cont_t* g_pcont;
	void DebugFreeStack()
	{
		register uint32_t *sp asm("a1");
		int freestack = 4 * (sp - g_pcont->stack);
		Serial.printf("current free stack = %d\n", freestack);
	}
}

void connectToWifi() 
{
	Serial.println("Connecting to Wi-Fi...");
[...]

PR #5153, opened 2018-09-19
implement EspClass::getFreeSysStack method

toomasz:
Not sure if implemetation is correct.

Did simple test:

```

void StackTest()
{
	char cc[100] = "sdfsdfsfsf";
	auto freeSysStack1 = ESP.getFreeSysStack();
	Serial.printf("free sys stack in submethod: %d\n", freeSysStack1);
	Serial.println(cc);
}

void onWifiConnect(const WiFiEventStationModeGotIP& event) 
{
	auto freeSysStack = ESP.getFreeSysStack();
	Serial.printf("free sys stack: %d\n", freeSysStack);
	StackTest();

	Serial.println("Connected to Wi-Fi.");
	DebugFreeStack();
}

```

And output was:

```
Connecting to Wi-Fi...
free sys stack: 12000
free sys stack in submethod: 11872
```
See comments in #5148

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -183,6 +183,11 @@ uint32_t EspClass::getFreeContStack()
     return cont_get_free_stack(g_pcont);
 }
 
+uint32_t EspClass::getFreeSysStack()
+{
+	register volatile uint32_t stackAddress asm("a1");
+	return stackAddress-SYSTEM_STACK_END_ADDRESS;
+}
 uint32_t EspClass::getChipId(void)
 {
     return system_get_chip_id();
--- a/cores/esp8266/Esp.h
+++ b/cores/esp8266/Esp.h
@@ -65,6 +65,9 @@ enum RFMode {
 #define WAKE_NO_RFCAL    RF_NO_CAL
 #define WAKE_RF_DISABLED RF_DISABLED
 
+// for getFreeSysStack
+#define SYSTEM_STACK_END_ADDRESS 0x3FFFC000
+
 enum ADCMode {
     ADC_TOUT = 33,
     ADC_TOUT_3V3 = 33,
@@ -111,6 +114,7 @@ class EspClass {
         void getHeapStats(uint32_t* free = nullptr, uint16_t* max = nullptr, uint8_t* frag = nullptr);
 
         uint32_t getFreeContStack();
+		uint32_t getFreeSysStack();
 
         const char * getSdkVersion();
         String getCoreVersion();
--- a/cores/esp8266/Esp.h
+++ b/cores/esp8266/Esp.h
@@ -114,7 +114,7 @@ class EspClass {
         void getHeapStats(uint32_t* free = nullptr, uint16_t* max = nullptr, uint8_t* frag = nullptr);
 
         uint32_t getFreeContStack();
-		uint32_t getFreeSysStack();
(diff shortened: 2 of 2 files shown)


toomasz:
I didn't realise getFreeContStack works that way. For now have two problems with implementing same approach for sys stack. 

- Don't know region of sys stack. Need this to fill it with magic pattern. @earlephilhower said is starts 0x3FFFC000 but what would be the stack size?
- No idea where to put code that fills SYS stack with watermark pattern. Is that even possible?

earlephilhower:
Actually, there's a new API that makes more sense than trying to find the high water mark.   Look at #5133 where the "get free stack" is added to the ESP class.  You can't really fill the OS stack unless you patch the startup code, which is probably not worth the effort.

devyte:
@earlephilhower #5133 is a thin wrapper around cont_get_free_stack(), which looks at the cont stack watermark. The cont stack is filled with a guard value on init, and usage overwrites those values. The watermark is checked by starting at the end and counting the guard values while moving inwards. My previous effort is aimed at doing something similar for the sys stack.
About where to init, we can do it in a naive place as a first approach, given that what we really care about is the stack usage during the application callbacks. We can figure out an earlier place later.
About how to init, I'd calculate the current stack usage with the approach given here, then I'd allocate an array of size with that value, then I'd fill that array with the guard value. That's just an abstract, you could do the same with just a pointer startingto at the end and moving inwards while assigning to it.
About how to calculatenthe free stack, I'd do the same thing as with the cont stack: start at the end and count guard values moving inwards.

earlephilhower:
Did not see that, @devyte .  While I can see its use, from my own experience the answer to the question "How much stack do I have now?" is much more interesting and given the name of the ESP function that's actually what I would expect...

For sanity this should match the existing behavior, though.  There is no task stack structure on startup, so @toomasz would need to find the current stack and then somehow safely fill from end-of-stack to `a1` with the guard bytes...

devyte:
Thinking back, I've actually found the need for both. Maybe the current method in #5133 should be renamed to getMaxContStackUse(), then a new method should be implemented for the current use, and then do the same for the sys stack. What do you think?

d-a-v:
Cont stack is initialized [here](https://github.com/esp8266/Arduino/blob/1a44f79a9e26ebb72900caef8e03a2b7df78724d/cores/esp8266/core_esp8266_main.cpp#L197) (within sys stack) or [here](https://github.com/esp8266/Arduino/blob/1a44f79a9e26ebb72900caef8e03a2b7df78724d/cores/esp8266/core_esp8266_app_entry_noextra4k.cpp#L30) (within user ram).
This might also help understanding addresses ([link](https://github.com/esp8266/Arduino/blob/1a44f79a9e26ebb72900caef8e03a2b7df78724d/cores/esp8266/core_esp8266_postmortem.c#L150))

d-a-v:
We may have both views:
how much do we have left right now,
and how far/deep did my app go inside stack until now.

devyte:
Reference: [sys stack area and painting](https://www.esp8266.com/viewtopic.php?f=13&t=16734) by @eriksl

toomasz:
I think we should have 4 methods then:
`ESP.getMaxContStackUse()` - rename method from #5133
`ESP.getMaxSysStackUse()` - to implement, tricky

`ESP.getFreeContStack()` - to implement, easy
`ESP.getFreeSysStack()` - method from this PR

PR #5156, opened 2018-09-21
cont.h: cont_yield() should be __attribute__d as ((returns_twice))

jjsuwa:
GCC manual says, "The returns_twice attribute tells the compiler that a function may return more than one time."


jjsuwa:
Sorry, my mistake. Cancelled.

PR #5159, opened 2018-09-22
Revert "Remove @PlatformIO library.json from GDBStub; fix pattern in LD"

diblumberg:


devyte:
@diblumberg what is this? What does it address?

tools/sdk/ld/eagle.app.v6.common.ld
```diff
@@ -96,7 +96,7 @@ SECTIONS
     *.pioenvs\\*\\lib*.a:(EXCLUDE_FILE (umm_malloc.o) .literal*, EXCLUDE_FILE (umm_malloc.o) .text*)
     *.pioenvs/*/lib*.a:(EXCLUDE_FILE (umm_malloc.o) .literal*, EXCLUDE_FILE (umm_malloc.o) .text*)
     /* project dependencies */
-    *.pioenvs\\*\\lib*\\lib*.a:(.literal*, .text*)
+    *.pioenvs\\*\\lib*\lib*.a:(.literal*, .text*)
```
earlephilhower: Besides the fact that this file is no longer part of the core, the conversion of `\\` to `\` cannot possibly be right, it makes the `l` in lib into an escape code.

PR #5163, opened 2018-09-24
Wrong directory for installation??

teo1978:
PLEASE REVIEW, I'm not completely  sure, but on Linux there's no `hardware` directory inside `Java`, and it doesn't seem to make sense that there should be on MacOS. There is, instead, a `hardware` directory in the parent.


devyte:
Thanks @teo1978

igrr:
The old version (one with Java) looks correct for MacOS version of Arduino, IMO. I think this should be reverted.

devyte:
It's not correct for Linux, and I checked on Linux. In that case, they should be specified individually. Also, I think it shouldn't even be inside the Arduino application directory, but rather in the arduino dir. I'll revert and edit.

PR #5164, opened 2018-09-25
Update to BearSSL 0.6+ release, add AES_CCM modes

earlephilhower:
Pull in latest BearSSL head (0.6 + minor additions) release and add AES_CCM
modes to the encryption options.


devyte:
Are there flashsize or heap use implications here?

earlephilhower:
There should be no real heap difference.  There is going to be a slight increase as these new codes will be included in the library.

If flash size becomes an issue I can move from `-O2` to `-Os`.  In fact, in another bug we may want to track the performance difference between the two builds.  In any sane processor, `-Os` would run slower.  In this CPU, though, running large straight-line code from ROM can be slower due to the poor SPI BW.

d-a-v:
I observed that -Os can run faster than -O2 with other CPUs too (the gameboy advance's one  arm7tdmi I believe it was)

earlephilhower:
Tests with the latest update show <2KB flash growth on the BearSSL_Validation example.

PR #5167, opened 2018-09-25
Fix TCP race condition, remove fixed delay in CC

earlephilhower:
ClientContext::_wait_for_sent() could dereference a TCP's _pcb after
the connection was dropped by the OS, resulting in a crash.

Move the connection dropped check to catch this case, and replace
a fixed millisecond delay() with a yield and timeout value to minimize
wasted time when transmission completes.

### DIFF
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -309,7 +309,6 @@ public:
             return true;
 
         int prevsndbuf = -1;
-        max_wait_ms++;
 
         // wait for peer's acks to flush lwIP's output buffer
         uint32_t last_sent = millis();
@@ -319,7 +318,8 @@ public:
                 // wait until sent: timeout
                 DEBUGV(":wustmo\n");
 #endif
-                break;
+                // All data was not flushed, timeout hit
+                return false;
             }
 
             // force lwIP to send what can be sent
@@ -328,9 +328,9 @@ public:
             int sndbuf = tcp_sndbuf(_pcb);
             if (sndbuf != prevsndbuf) {
                 // send buffer has changed (or first iteration)
-                // we received an ack: restart the timeout
                 prevsndbuf = sndbuf;
-                last_sent = millis(); // We just sent a bit, move timeout forward
+                // We just sent a bit, move timeout forward
+                last_sent = millis();
             }
 
             yield();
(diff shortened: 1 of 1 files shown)


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -308,37 +308,37 @@ class ClientContext
         if (!_pcb)
             return true;
 
-        int loop = -1;
         int prevsndbuf = -1;
         max_wait_ms++;
```
d-a-v: `max_wait_ms++;` no more necessary

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -308,37 +308,37 @@ class ClientContext
...
+            }
 
             // force lwIP to send what can be sent
             tcp_output(_pcb);
 
             int sndbuf = tcp_sndbuf(_pcb);
             if (sndbuf != prevsndbuf) {
                 // send buffer has changed (or first iteration)
-                // we received an ack: restart the loop counter
+                // we received an ack: restart the timeout
```
d-a-v: just remove this comment, your new one is better

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -308,37 +308,37 @@ class ClientContext
...
         // wait for peer's acks to flush lwIP's output buffer
-
+        uint32_t last_sent = millis();
         while (1) {
+            if (millis() - last_sent > (uint32_t) max_wait_ms) {
+#ifdef DEBUGV
+                // wait until sent: timeout
+                DEBUGV(":wustmo\n");
+#endif
+                break;
```
d-a-v: this break would be `return false;`

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -308,37 +308,37 @@ class ClientContext
...
-        if (loop <= 0) {
-            // wait until sent: timeout
-            DEBUGV(":wustmo\n");
+            if ((state() != ESTABLISHED) || (sndbuf == TCP_SND_BUF)) {
+                break;
+            }
         }
-        #endif
 
         return max_wait_ms > 0;
```
d-a-v: `return true;` here

Issue #5175, opened 2018-09-27
Downloading large files to spiffs via https 


MarcFinns:
I am observing at behaviour that I would like to confirm being as expected.
I need to download a file from an HTTPS service into SPIFFS. This file is fairly large and  when downloading using HTTP it is not a problem. This is what I have done in other projects for example with the json streaming parser. It looks like HTTPS is instead loading the whole thing in RAM, and then it crashes.
Is this what others are experiencing/ the expected behaviour? Is there a way to do it without lording it in RAM? ?
Thank you.

MarcFinns:
Apologies. Here it is.

### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[2.4.2]
- Development Env:		[Arduino IDE]
- Operating System:		[MacOS]

### Settings in IDE

- Module:  			[Nodemcu 1.0]
- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[20Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [other] (serial upload only)


### Problem Description

I need to download a file from an HTTPS service directly into SPIFFS, as I used to do successfully with plain HTTP with files larger than the available RAM (e.g. json files).
The sketch shows about 30K free heap before the download starts. The download fails if the file is above 8k or so. The sketch gets stuck, no stack dump. Enabling the serial debug shows a ram issue (ssl->need_bytes=10464 > 6859).
It seems the file is loaded first into RAM and does not go directly to SPIFFS. Is this the expected behaviour? Is there a way to download it without copying it into RAM?


### [MCVE](https://stackoverflow.com/help/mcve) Sketch

#define FS_NO_GLOBALS
#include <FS.h>

#include <ESP8266WiFi.h>
#include <ESP8266HTTPClient.h>
#include <Arduino.h>

[...]

PR #5185, opened 2018-09-29
Really free stack after last BearSSL obj destroyed

earlephilhower:
The BearSSL second stack, once allocated, was never deallocated.  The
reference count of the stack pointer never hit 0 due to the initial
creation counting as one.  Now, check to see if there is only one use_count
and if so then delete the stack.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
@@ -91,12 +91,16 @@ WiFiClientSecure::WiFiClientSecure() : WiFiClient() {
   _clear();
   _clearAuthenticationSettings();
   _certStore = nullptr; // Don't want to remove cert store on a clear, should be long lived
+  _ensureStackAvailable();
+  _local_bearssl_stack = _bearssl_stack;
+}
+
+void WiFiClientSecure::_ensureStackAvailable() {
   if (!_bearssl_stack) {
     const int stacksize = 4500; // Empirically determined stack for EC and RSA connections
     _bearssl_stack = std::shared_ptr<uint8_t>(new uint8_t[stacksize], std::default_delete<uint8_t[]>());
     br_esp8266_stack_proxy_init(_bearssl_stack.get(), stacksize);
   }
-  _local_bearssl_stack = _bearssl_stack;
 }
 
 WiFiClientSecure::~WiFiClientSecure() {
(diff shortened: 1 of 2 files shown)


MarcFinns:
> This fixes the secondary issue found in #5175 by @MarcFinns.

That was quick... Thanks!
DO you plan to release the fix in the main distribution anytime soon or shall I just use the github version?
thanks again!

earlephilhower:
December is the next "scheduled" release, so git head is going to be the best bet for some time.

MarcFinns:
Hi Earle, I erased the installation and followed the instructions to install from git (on macOS).
I still see the issue in the test sketch. what did i miss?
thanks again.

earlephilhower:
You need to apply the PR patch until this is merged.  See https://help.github.com/articles/checking-out-pull-requests-locally/ for more info.

MarcFinns:
Thanks. I noticed that the version from GitHub (without this patch) uses more ram than the 2.4.2. Is this expected?
As part from these teething issues, BearSSL seems a great addition.

MarcFinns:
I confirm the specific issue is fixed.
However, the GIT version uses about 4k RAM more than the 2.4.2 release. I can see that right after compilation and at runtime.
Is this another issue?

earlephilhower:
Looks basically the same to me.  You're going to need a new bug and MCVE.

Using your test sample and git head there's some difference, but it's <1KB:
````
Setup - before instantiation - Free Heap: 49688
Setup - After instantiation - Free Heap: 44920
Setup - After delete - Free Heap: 49688
Loop - Free residual Heap: 49704
````
tags/2.4.2:
````
Setup - before instantiation - Free Heap: 49192
Setup - After instantiation - Free Heap: 44480
Setup - After delete - Free Heap: 44664
Loop - Free residual Heap: 44680
````

MarcFinns:
Hi Devyte, thanks for the hint, i think this might indeed be related.
I use the WifiManager library. In the latest version there is a patch to disable WPS to comply with ESP 2.4.2. 
I see from the source that it kicks in if NO_EXTRA_4K_HEAP is defined.
It appears this is defined in 2.4.2 but not in the git version. can this be?
thanks

devyte:
In current git, if you use wps, a heap optimization is disabled, which allocates the cont stack on the heap instead of on top of the sys stack. This is because it was found that wps apparently has large stack use, so requuires the original huge size.
Latest git detects the use of wps and disables automagically. In 2.4.2 you had to select the option from the IDE menu (I think).
Bottom line: if you're using wps, the optimization isn't in effect, and your heap is 4KB smaller.

PR #5193, opened 2018-09-30
Dns server cleanup

Makuna:


Makuna:
warnings needed to be cleaned up

PR #5194, opened 2018-09-30
Dns server cleanup

Makuna:


earlephilhower:
Just a note, @Makuna, you can keep a PR open and just `git push` onto your private branch and it will update the PR with your changes.  Not a big deal here, but when it's a long set of changes with comments using the same PR makes it way easier to follow any comments.

Makuna:
Also, Since I am the author of this change, I feel other collaborators should review and merge.  Is that the normal policy?

libraries/DNSServer/src/DNSServer.cpp
```diff
@@ -48,82 +46,106 @@ void DNSServer::downcaseAndRemoveWwwPrefix(String &domainName)
...
+  
+  const uint8_t* pos = buffer + sizeof(DNSHeader);
+  const uint8_t* end = buffer + packetSize;
+
+  // to minimize reallocations due to concats below
+  // we reserve enough space that a median or average domain
+  // name size cold be easily contained without a reallocation
+  // - max size would be 253, in 2013, average is 11 and max was 42
+  //
+  parsedDomainName.reserve(32);
```
earlephilhower: Yay!  String::reserve() examples are few and far between.

libraries/DNSServer/src/DNSServer.h
```diff
@@ -6,6 +6,8 @@
 #define DNS_QR_RESPONSE 1
 #define DNS_OPCODE_QUERY 0
 
+#define MAX_DNSNAME_LENGTH 253
+
```
earlephilhower: Not used anywhere as far as I can see, but shouldn't do any harm.

PR #5195, opened 2018-09-30
restore spiffs data in boards.txt

d-a-v:
(removed by mistake)


earlephilhower:
This fixes my issue uploading SPIFFS.

Trying an upload with a `4M (no SPIFFS)` setting, which should error, gives:
````
[SPIFFS] data   : /home/earle/Arduino/BearSSL_CertStore1/data
[SPIFFS] size   : -20
[SPIFFS] page   : 256
[SPIFFS] block  : 8192
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc

SPIFFS Create Failed!
````

With the removed `if spiffs_kb>0`, though, it would give a slightly more informative error
`SPIFFS Not Defined for LOLIN(WEMOS) D1 R2 & mini, 80 MHz, Flash, 4M (no SPIFFS), v2 Lower Memory, Disabled, None, Only Sketch, 921600`

Really a matter of preference, since it's an error and both are pretty obvious when you look at them.

File: tools/boards.txt.py [hunk 1]
```diff
@@ -1119,12 +1119,11 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
         ( menu + '.upload.maximum_size', "%i" % max_upload_size ),
         ( menub + 'rfcal_addr', "0x%X" % rfcal_addr)
         ])
-    #if spiffs_kb > 0:
```
devyte: Per @earlephilhower : putting this if back improves the error message for 4M0

d-a-v: fixed

PR #5198, opened 2018-10-02
PolledTimeout Class for wrapping millis() loops (WIP)

devyte:
Class to wrap typical polled timing constructs:
1. wait n ms before continuing (oneshot)
2. retry some test until n ms timeout has expired {oneshot)
3. execute some code body once every n ms (periodic)

Classes are declared inside namespace esp8266. Use model is meant for stack declaration (declaration initializes, i.e.: RAII) with one exception: do something in the loop every n ms. In that last case, declaration must be global and the global must be reset() at the end of setup().

I expect to add a device test as soon as I figure out htf they work.

Usage is something like this:
```
void test1()
{
  using namespace esp8266;
  polledTimeoutOneShot timeout(3000);

  Serial.println("OneShot Timeout 3000ms");

  Serial.print("before 1\n");
  while(!timeout.expired())
    yield();
  Serial.print("after 1\n");

  Serial.print("reset\n");
  timeout.reset();

  Serial.print("before 2\n");  
  while(!timeout)
    yield();
  Serial.print("after 2\n");
}

void test2()
{
  using namespace esp8266;
  polledTimeoutPeriodic timeout(3000);

  Serial.println("Periodic Timeout 1T 3000ms");

  Serial.print("before 1\n");
  while(!timeout)
    yield();

  Serial.print("no reset needed\n");

  Serial.print("before 2\n");  
  while(!timeout)
    yield();
  Serial.print("after 2\n");
}

void test3()
{
  using namespace esp8266;
  polledTimeoutPeriodic timeout(1000); 

  Serial.println("Periodic 10T Timeout 1000ms");

  int counter = 10;
  while(1)
  {
    if(timeout)
    {
      Serial.print("*");
      if(!--counter)
        break;
      yield();
    }
  }
}
```


d-a-v:
Does it make sense to add such extenders for ease of use,
or should this be left to the user as an exercise in personal sketches ?

```
class EspTimeout: public esp8266::polledTimeoutOneShot
{
public:

    EspTimeout(timeType timeout = 0): polledTimeoutOneShot(timeout)
    {
    }

    EspTimeout& operator= (timeType timeout)
    {
        _timeout = timeout;
        return *this;
    }

    bool expired()
    {
        yield();
        return expiredOneShot();
    }
    
    operator bool()
    {
        return expired(); 
    }
};

class EspEvery: public esp8266::polledTimeoutPeriodic
{
public:

    EspEvery(timeType timeout = 0): polledTimeoutEvery(timeout)
    {
    }

    EspEvery& operator= (timeType timeout)
    {
        _timeout = timeout;
        return *this;
    }

    operator bool()
    {
        bool ret = expired(); 
        if (ret)
            reset();
        return ret;
    }
};
```

```
// yield not needed, semantically it's in the name ("Esp"Timeout)
void test()
{
  for (EspTimeout timeout = 3000; !timeout; )
    doStuff();
}
```


```
EspEvery ledChange = 1000;

void test()
{
  if (ledChange)
    digitalWrite(LED_PIN, 1 - digitalRead(LED_PIN)
}
```

I would call this the glue between perfect c++ and the arduino world which is primarily aimed at beginners.

yoursunny:
The API looks good.
I wonder why type names are starting with a lower case letter? It’s different from most other code that uses an upper case letter.

PR #5208, opened 2018-10-05
Updater: fix small bug and add visual feedback

Jeroen88:
The function writeStream() in Updater.cpp tries to read a full buffer every time, even at the end of the stream. Because the last buffer is almost never a full buffer, writeStream() exits on a timeout in stead of on end of data. This timeout can be potentially long, eg BearSSL uses a timeout of 10,000 ms. With this PR this is fixed: only the remaining bytes are read for the last buffer.
A visual feedback is added to the update process: the blue LED on pin 2 is turned on just before starting to read the next buffer and turned of just before writing this buffer to flash. In this way the blue LED starts to blink during an update, 'on' means receiving from the network, 'off' means writing to flash. On a normal, stable connection the LED should blink in a regular way. On a bad connection the LED will show irregular 'on' times, that may be potentially long (on for seconds)

### DIFF
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -370,10 +370,14 @@ size_t UpdaterClass::writeStream(Stream &data) {
         return 0;
     }
 
-    pinMode(LED_BUILTIN, OUTPUT);
+    if(_ledPin != -1) {
+        pinMode(_ledPin, OUTPUT);
+    }
 
     while(remaining()) {
-        digitalWrite(LED_BUILTIN, _ledOn); // Switch LED on
+        if(_ledPin != -1) {
+            digitalWrite(LED_BUILTIN, _ledOn); // Switch LED on
+        }
         size_t bytesToRead = _bufferSize - _bufferLen;
         if(bytesToRead > remaining()) {
             bytesToRead = remaining();
@@ -389,7 +393,9 @@ size_t UpdaterClass::writeStream(Stream &data) {
                 return written;
             }
         }
-        digitalWrite(LED_BUILTIN, _ledOn == HIGH ? LOW : HIGH); // Switch LED off
+        if(_ledPin != -1) {
+            digitalWrite(LED_BUILTIN, _ledOn == HIGH ? LOW : HIGH); // Switch LED off
+        }
         _bufferLen += toRead;
         if((_bufferLen == remaining() || _bufferLen == _bufferSize) && !_writeBuffer())
             return written;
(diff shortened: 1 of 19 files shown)


cores/esp8266/Updater.cpp
```diff
@@ -360,18 +360,27 @@ size_t UpdaterClass::writeStream(Stream &data) {
         return 0;
     }
 
+    pinMode(2, OUTPUT);
```
d-a-v: LED_BUILTIN is preferred here.

Also pin state should be backuped in case OTA fails and sketch should go on,
or,
The led pin (and also it's lit state which can be 0 for ON) could be passed through arguments for users who are using the led pin for something else like a relay.

Jeroen88:
@d-a-v made LED feedback optional, LED pin & LED on status selectable using an adapted function bool begin(size_t size, int command = U_FLASH, int ledPin = -1, uint8_t ledOn = LOW). If the update fails the pin status is reset to the status prior to the call. The default value for ledPin is -1 (no pin), ensuring current behavior of the class)
I commit #4980 as well to make use of this feature by the added function void setLedPin(int ledPin = -1, uint8_t ledOn = HIGH) of the ESP8266HTTPUpdate class in that PR

d-a-v:
Thanks for this bugfix and the the optional led !

Regarding saving the pin digital state, I'm not sure we should keep this as is. Pin can be of different types (not only digital - also analog or i2s at least).
Semantically, because setting the led is now an option given to the user, then saving and restoring pin function should be left at the user responsibility and you should not care about saving its digital state anymore.

> I commit #4980 as well to make use of this feature by the added function void setLedPin(int ledPin = -1, uint8_t ledOn = HIGH) of the ESP8266HTTPUpdate class in that PR

@earlephilhower this one needs merge for #4980 to pass CI

Jeroen88:
@d-a-v you're welcome :)

earlephilhower:
@Jeroen88 LGTM, but maybe you could update the example code to use this new feature?  W/o an example, it's not exceptionally clear from the header what's going on.

Jeroen88:
@earlephilhower Sure! All 3 examples updated with the usage of setLedPin() and a comment on how to use it

earlephilhower:
@Jeroen88 since you've got this change and example already in your PR #4980, maybe this one should be closed?  There's no reason to keep two separate branches up to data with your changes, just extra work for you.

Jeroen88:
@earlephilhower this PR is independent of #4980 and could be merged separately. This PR does not change the API, it just solves a small bug and adds an optional extra feature. Shouldn't it be better merged?

earlephilhower:
@Jeroen88 you included these changes in the other PR, so when I did the "Update Branch" you now have "0 files changed" :)  Nothing to merge here!

Issue #4485, opened 2018-03-09
LIGHT_SLEEP_T causes endless delay


MikhailNatalenko:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:				[ESP-12]
- Core Version:			[latest git hash or date]
- Development Env:		[Arduino IDE]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Nodemcu]

### Problem Description

LIGHT_SLEEP_T wifi mode brakes sketch. After 10-15 seconds from connection it just stop's without any output (like infinite loop with yield()). No WDT reset or exception. 

This bug appers in last commit (170911a6) and 2.4.1.  2.4.0 works fine (SDK: 2.1.0 deb1901)

Delay before `wifi_set_sleep_type(LIGHT_SLEEP_T);` doesn't help (5-10 seconds)

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <ESP8266WiFi.h>

extern "C" {
#include "user_interface.h"
}

void printStatus();

void setup() {
  wifi_set_sleep_type(LIGHT_SLEEP_T);
  WiFi.begin("my-ssid", "password");
  delay(500);
  
  WiFi.printDiag(Serial);

  Serial.begin(115200);
  Serial.println(F("\xAA\xAA\xAA\xAA\xAA\xAA\xAA\xAA\r\n"));
  delay(100);
  Serial.print(F("Core: "));
  Serial.println(ESP.getCoreVersion());

  Serial.print(F("SDK: "));
  Serial.println(ESP.getSdkVersion());

  WiFi.printDiag(Serial);
  Serial.println(F("Starting Firmware"));

}

void loop(void) { 
  delay(250);
  printStatus();
}

void printStatus() {
  static auto i = 0;
  Serial.print(WiFi.isConnected());
 
  if (i > 20) {
    Serial.println();
    i = 0;
  }
  i++;
}

```
```
Core: 00000000
SDK: 2.2.1(cfd48f3)
Mode: STA
[...]

PR #5210, opened 2018-10-06
Update to the last version of non-sdk V2, WiFi addons

d-a-v:
New feature: "listen interval" setting in
`WiFi.setSleepMode(WIFI_LIGHT/MODEM_SLEEP, listen-interval)`
Default value in FW is 3 so I tend to think this is a DTIM-related value like it is explained in https://routerguide.net/dtim-interval-period-best-setting/.

note: I'm not a radio expert

With new api: `WiFi.getListenInterval()` / `WiFi.isSleepLevelMax()`
(the setters are here but should not be used, prefer WiFi.setSleepMode())

This is interesting for at least #2330 #1381 #4485 

This WiFi issue with light or modem sleep modes does not happen for everybody,
so testing testers with current-meter or not are needed.


doc/esp8266wifi/generic-class.rst
```diff
@@ -61,7 +61,7 @@ Other Function Calls
 .. code:: cpp
 
     int32_t  channel (void)
-    bool  setSleepMode (WiFiSleepType_t type)
+    bool  setSleepMode (WiFiSleepType_t type, int listenInterval=-1) (DTIM setting for LIGHT or MODEM sleep mode)
```
devyte: Rather than add a short explanation here, how about adding a section above with the explanation, and adding a short comment for the args?

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -426,6 +432,65 @@ bool ESP8266WiFiGenericClass::forceSleepWake() {
...
+ * datasheet:
+   It only works when sleep level is set as MAX_SLEEP_T
+   It should be called following the order:
+     wifi_set_sleep_level(MAX_SLEEP_T)
+     wifi_set_listen_interval
+     wifi_set_sleep_type
+   forum: https://github.com/espressif/ESP8266_NONOS_SDK/issues/165#issuecomment-416121920
+   default value seems to be 3 (as recommended by https://routerguide.net/dtim-interval-period-best-setting/)
+ */
+void ESP8266WiFiGenericClass::setListenInterval (uint8_t listenInterval) {
```
devyte: Does it make sense to have a setter just for this when the call must be part of a specific call sequence? It would be trivialfor a user to call it in wrong order.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -426,6 +432,65 @@ bool ESP8266WiFiGenericClass::forceSleepWake() {
...
+     receive beacon.  Broadcast data may be lost because station may be in sleep
+     state at DTIM time.  If listen interval is longer, more power will be saved, but
+     it’s very likely to lose more broadcast data.
+
+   - Default setting is minimum sleep level.
+
+   Further reading: https://routerguide.net/dtim-interval-period-best-setting/
+
+ */
+void ESP8266WiFiGenericClass::setSleepLevelMax (bool setToMax) {
```
devyte: Same here, sdk docs say this is part of a call sequence, so exposing it stand alone allows the user wrong usage.

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -86,6 +86,7 @@ static bool sta_config_equal(const station_config& lhs, const station_config& rh
 // -----------------------------------------------------------------------------------------------------------------------
 
 bool ESP8266WiFiSTAClass::_useStaticIp = false;
+bool ESP8266WiFiSTAClass::_useInsecureWEP = false;
```
devyte: Is this here on purpose, or did it slip in? I'm ok with what I see of the code, but given the nature of this functionality I suggest splitting it into its own PR.

ascillato:
Hi,

I have tested the actual Stage Core with this PR in order to check the Modem Sleep Feature and works very good. No crashes if using lwIP 1.4.

ascillato:
Hi,

After performing several tests of the actual stage core with @d-a-v's PR, it results that it is working really good for [Sonoff-Tasmota](https://github.com/arendst/Sonoff-Tasmota).

All functions are working very good and also _**the sleep feature finally works**_ in a core different than 2.3.0 !

I have tested with lwIP 1.4, lwIP 2 (low memory) and also lwIP (high bandwidth) and works very stable in all those, with and without sleep enabled.

So, I hope this Pull Request got merged ASAP into the core, and also that a new core got released ASAP.

Thanks a lot for all the hard work you all have done here, and thanks @d-a-v for this solution!

Jason2866:
Hello @d-a-v 
would be perfect is your PR is merged very very soon and a new core version is released!
I use Tasmota (https://github.com/arendst/Sonoff-Tasmota) which uses Modem Sleep. **With your modification Sleep works again without issues.** Sleep was unusable since the last working Version 2.3.0.
Thanks for your work and all the others here!

andrethomas:
This is really good news - will be very good if these fixes can be included to finally solve the sleep problem on 2.4.2!

arendst:
It would be great if this PR could be implemented really soon as currently most Tasmota users are forced to use core 2.3.0 for light sleep functionality while so many new features have been released in later core versions.

mike2nl:
I would like to have it too, because by 193 sonoff devices you have a lot of power use and that is pricy.
When i could cut it by a half that would be a great step in the right direction.

So please don't waste time and money. Add it wher we need it.

d-a-v:
documentation [overview](https://github.com/d-a-v/Arduino/blob/fw2/doc/esp8266wifi/generic-class.rst#wifi-power-management-dtim)

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
d-a-v: All points above are addressed your way.
About WEP, it is a requirement to set this value in sdk structures for initialization.

clumsy-stefan:
Just a question: this actually updated to Nonos SDK V3.0.0 not V2 ? At least the libraries in the tools/sdk/lib directory show V3...

d-a-v:
It is called "V3-dev" by espressif.
It is not V3 but their latest commit that can be currently used with the 
arduino core.
It brings some useful updates and fixes regarding WiFi sleep modes which 
were broken around arduino core v2.4.0.
Official V3 will come later when the arduino core will be updated to 
reliably boot with it.

AugustoCiuffoletti:
It seems that the problem has been solved, so it's probably my mistake, but I still have problems with LIGHT_SLEEP, and I can't find any document around. I use the Arduino IDE 1.8.7, and installed the ESP8266 package v2.5.0 (stable) using the File->Preferences way. Doing of my best with many suggestions that I found googling around I produced the following testing sketch:
```c++
#include <ESP8266WiFi.h>
#include "secret.h"  //defines MYSSID and MYPASSWD

int joinAP(){
  Serial.print("Connecting");
  WiFi.begin(MYSSID, MYPASSWD);
  {
    delay(500);
    Serial.print(".");
  }
  Serial.println();
  Serial.print("Connected as ");
}

void setup()
{
  Serial.begin(115200);
  WiFi.mode(WIFI_STA);
  pinMode(LED_BUILTIN, OUTPUT); 
  digitalWrite(LED_BUILTIN,LOW);
  WiFi.setSleepMode(WIFI_LIGHT_SLEEP,10);
}

void loop()
{
  WiFiClient client;
  HTTPClient http;

  joinAP();
  digitalWrite(LED_BUILTIN,!digitalRead(LED_BUILTIN));
  if (http.begin(client, "http://www.example.com")) {
    Serial.print(" - return code: "); Serial.println(http.GET());
  }

unsigned long t0=millis();
WiFi.forceSleepBegin(5000000UL); 
Serial.println(millis()-t0);
}
```
However there is no 5 seconds delay: the output is as follows:
```
Connecting.............................
Connected as 192.168.43.113
 - return code: 200
91
.....
```
The 91 is the difference in ms. In fact, there is no visible delay.

Thank you.

AugustoCiuffoletti:
I did an experiment today replacing the WiFi.forceSleepBegin(5000000) with a plain
```
delay(5000);
```
The result is that the delay time is measured correctly, and power consumption drops around 15mA, with short (20ms) spikes around 60mA every one second. Outside the delay time power consumption is around 60mA with spikes around 200mA.
So power consumption is reduced to 25%, but I'd expect better than that. My guess is that the result is consistent with a MODEM_SLEEP, and not with LIGHT_SLEEP as I asked with the setSleepMode.
I used an oscilloscope with a 1Ohm shunt to measure currents.

netprince17:
I haven't had any luck making lightsleep work yet either.  

We need a good example that works with a specific version of the esp8266 code.

PR #5213, opened 2018-10-06
Add cryptographically signed update support

earlephilhower:
--edit 11/05-- Remove WIP

Using a pluggable architecture, allow updates delivered via the Update
class to be verified as signed by a certificate.  By using plugins, avoid
pulling either axTLS or BearSSL into normal builds.

A signature is appended to a binary image, followed by the size of the
signature as a 32-bit int.  The updater takes a verification function
and checks this signature using whatever method it chooses, and if it
fails the update is not applied.

A SHA256 hash class is presently implemented for the signing hash (since
MD5 is a busted algorithm).

A BearSSLPublicKey based verifier is implemented for RSA keys.  The
application only needs the Public Key, while to sign you can use
OpenSSL and your private key (which should never leave your control
or be deployed on any endpoints).

This code has been tested by manually signing a Blink.ino.bin file using the BearSSL server example pub/private keypair, but it needs to have a better detailed example and be generalized for different RSA and EC keys.

Inspired by @madpilot's PR #3105 which, unfortunately seems to be abandoned and for some reason had to pull in the entire axTLS source tree and new binary.  Maybe the functions weren't exported in axTLS.  They are exported in BearSSL, so the changes are minimal here.


earlephilhower:
Signing example using /tmp/sk as the signing private key.  Mostly for my benefit
````
sha256sum /tmp/file.bin  | cut -f1  -d" " | xxd -r -p > /tmp/sha256.bin
openssl rsautl  -sign -inkey /tmp/sk -in /tmp/sha256.bin  > /tmp/sha256.sign
echo 00010000 | xxd -r -p > /tmp/256.bin
cat //tmp/file.bin /tmp/sha256.sign /tmp/256.bin  > /tmp/esp8266.bin
````

The corresponding public key to the private one used should also, of course, be in the source code.

**Comment by devyte** [comment]
Please don't merge this yet, I have several comments.

**Comment by earlephilhower** [comment]
Sure, @devyte .  Apologies in advance for the virtual functions and base classes.  Given we've got stuck with a single, globally allocated `Updater` variable templates aren't an option.

Potential things I think still need checking for are 4-byte length alignment, but I think that's actually guaranteed by the binary format.  OTW it's only a readme away from general usability, methinks.  Tested three different signed uploads, all worked (and one unsigned and one signed but doctored binary failed with SIGN_ERROR as expected).

**Comment by madpilot** [comment]
So glad you have picked this up! You are correct about importing all of AxTLS, but sounds like BearSSL could be a winner. I'll be following along with interest (I wish I had more bandwidth to help more)

**Comment by d-a-v** [comment]
IMHO, KISSest would be 3/ and python.
un-knowledgeable users don't have python.
Keys defined/located with a name/directory, generated files into arduino {build} (/tmp) directory.

**Comment by devyte** [comment]
I'm also inclined towards 3. In order for it work, the user still has to have some knowledge.
If found necessary, 2 can come later to add flexibility and mitigate the dangers. In the meantime, we can put together a guide describing what to do and what not to do.
I don't like 1, mostly because I'm pretty sure I'd have trouble using it :p

**Comment by liebman** [comment]
Mac's have python by default. 10.13 has Python 2.7.

**Comment by earlephilhower** [comment]
Great, that makes it a little easier.  I also forgot the post-bingen step, actually signing the executable.  Doing that in Python would be easier and saner than the bash I'm using now.

Unfortunately I'm still stuck on how to conditionally change the compiled code.  Any ideas that don't include overwriting include files?  If I can run bash scripts in the build, I suppose it would be possilbe (but ugly) to have 2 copies of the compile command line prefixed with  something like `if -e ./signed.private.key gcc ... -D SIGNED` and one that is `if -not -e ./signed.private.key ...` (or whatever bash ! is)

**Comment by d-a-v** [comment]
Does `test -r somefile && INCLUDEIT='-include somefile'; gcc $INCLUDEIT ...` would help ?
For windows10/linux/mac we "officially" have `bash` (`sh` in fact). That would live windows-pre10 (7/8) aside though...
About linux, I think python2.7 is preinstalled on most major distributions by default.

**Comment by earlephilhower** [comment]
(3) is implemented.

The latest push can now sign the executable automatically in the build process assuming that openssl is installed and there is a "public.key" and "private.key" file in the sketch directory.  The manual method of installing a key and signing from the command line is also available.

I'm not too happy about having to include BearSSL calls inside `cores/esp8266`, but that's the only way I can see of making it work.  The example is also wonky:  because the IDE does NOT copy any files other than the `.ino` when you save the example, it won't duplicate the key files.  But, if you build directly w/o saving to a custom directory, it does include the key.

**Comment by earlephilhower** [comment]
Looks like Platform.IO doesn't use platform.txt, so have to hack that as well to build the signing header.  Using the existing Python script instead of bash looks like it'll turn out useful as I won't have to duplicate any weird bash in the pio build setup.

**Comment by earlephilhower** [comment]
Updated the documentation describing the signing process, how to make a key pair, etc.  Ready for serious review I think.

**Comment by earlephilhower** [comment]
Add 2.5.0 since it's related to a 2.5.0 bug, #2103

**Comment by corbolais** [comment]
@earlephilhower, thank you for you efforts. Greatly appreciated.

**Comment by Misiu** [comment]
@earlephilhower I'd like to ask for more details about PlatformIO and Windows 10.
PlatformIO has [advanced scripting](http://docs.platformio.org/en/latest/projectconf/advanced_scripting.html) maybe this feature could be used on Windows to allow automatic signing?
What tools are needed on Windows (https://github.com/esp8266/Arduino/pull/5213/commits/15ca564f74ed3ef3d13c793ba506e24e663956ef#diff-9f3d735bebc6c9e39a84eecbbb897a6dR81).
If this isn't supported by default maybe an easy walkthrough could be added (I was thinking about PlatformIO)?
Some of use use WIndows 🙂

**Comment by earlephilhower** [comment]
@Misiu , your best bet to get this followed up on is to open a new issue or else it's going to be lost on a closed PR.

We need python and native openssl executables.  #5635 adds a native, self-contained Python tool which is one part.  Win32 OpenSSL executables are still needed and very rare among Arduino users (OpenSSL exes are part of most Linux (and I think Mac) installations by default).

**Comment by Misiu** [comment]
@earlephilhower OpenSSL isn't a problem, it is easily installable on Windows.
I don't mind manual installation on OpenSSL, especially is it done only once.
If installing OpenSSL manually and adding it to the path (so that OpenSSL executable is available from the command line) is the only step then this should be definitively added to docs.
I'll open a new issue to track this.

**Comment by earlephilhower** [comment]
It's more than just getting the OpenSSL binaries.  Python needs to be there and the platform.txt needs to be updated to run under Windows.  But please do open up an issue so it can be tracked after the other changes land.

**Comment by Misiu** [comment]
@earlephilhower I've already did - https://github.com/esp8266/Arduino/issues/5694.
If I'm not mistaking Python is installed with PlatformIO, but as with OpenSSL installation of Python is done only once, so no problem with that 🙂

**Comment by Najsr** [comment]
If anyone is looking for PHP script for automated signing, here you go. It can be added to wiki if anyone cares (will work with version newer than 2.5.2:
```php
<?php
$key_path = __DIR__ . '/key.pem';
$compiled_bin_path = 'esp8266.bin';
$bin_data = base64_decode('AAEAAA==');

$key = openssl_pkey_get_private('file://' . $key_path);
$file = fopen($compiled_bin_path, 'r');
$fw_data = fread($file, filesize($compiled_bin_path));
fclose($file);
$data;

$hash = hash_file('sha256', $compiled_bin_path, true);
openssl_private_encrypt($hash, $data, $key);

$result = $fw_data . $data . $bin_data;
$myfile = fopen($compiled_bin_path . '.signed', "w");
fwrite($myfile, $result);
fclose($myfile);
```

Issue #2019, opened 2016-05-11
Add an on() function to ESP8266 Webserver that supports startsWith


bmuessig:
Hello!

Having a way to pass paramters via URL seems like a very important thing to have and I think a on() handler that uses startsWith should therefore be added.
As this was essential in one of my projects, I did the changes to my local copy of the esp8266webserver library but I really strongly suggest adding this feature to allow for more interactive websites and easier programming of things like: http://esp8266.local/configure/mdns_name/this_is_my_new_name and similar use cases like passing strings via the url.
##

Duality4Y:
link does not work.

Muessigb:
Well it isn't supposed to be a real link but an example of what I am using
the new functions for.

2016-05-11 21:26 GMT+02:00 Robert <notifications@github.com>:

Duality4Y:
yes i know now :) didn't read it right. pretty nice function to have.

randuhmm:
I have some experimental code that makes it possible to create a `RequestHandler` object that can handle URI path parameters. The basic usage is to create a the handler object like this:

    void myHandler(LinkedList<String>& segments) {
        // Handle the request and path parameters here
    }
    
    PathSegmentRequestHandler* handler = new PathSegmentRequestHandler(
        myHandler,
        "/my/path/widgets/{}/things/{}",
        HTTP_ALL);
    http.addHandler(handler);

The `PathSegmentRequestHandler` is a custom class that I created that is able to parse individual segments of the URI and extract any parameters. The `{}` acts as a placeholder for the parameter.

While this is working for me, I think there is some room for optimization so that the path segments do not need to be parsed individually by each handler. I think that the server should first check all of the static paths for matches and then, if no match is found, the URI can be separated into segments and the dynamic paths can be checked.

nardev:
Would be nice and cleaner to be able to aloate function only for serving files per path preferences...

like

[ip]/scripts/{something_goes_here}

to pass full uri to a handler for anything under "scripts" .. and than determine if it fulfills our criteria and  what to do with it.. fetch from eprom or not... etc...

bytemage:
Hope this will make it in the next milestone. It's sorely missing.

devyte:
There are some VERY hairy issues being investigated that are much higher priority than this. To help, I encourage anyone who needs this to make a PR.
[...]

PR #5214, opened 2018-10-06
Feature/path arguments

Bmooij:
Adding path arguments to WebServer.  (In respond to https://github.com/esp8266/Arduino/issues/2019)
Ex.
```
server.on("/led/2/actions/{}", []() {
    String action = server.pathArg(0);
    ...
 });
```

I don't know if this is the correct approach, feel free to give comments.

### DIFF
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -23,6 +23,7 @@ private:
 	
 protected:
     std::vector<String> pathArgs;
+    bool _isRegex = false;
 
 public:
     const String& pathArg(unsigned int i) { 
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -18,6 +18,16 @@ public:
 
 private:
     RequestHandler<ServerType>* _next = nullptr;
+	
+protected:
+    std::vector<String> pathArgs;
+
+public:
+    String pathArg(int i) { 
+        if (i < pathArgs.size())
+            return pathArgs[i];
+        return "";
+    }
 };
 
 #endif //REQUESTHANDLER_H
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -24,7 +24,7 @@ protected:
     std::vector<String> pathArgs;
 
 public:
-    String pathArg(int i) { 
+    String pathArg(unsigned int i) { 
         if (i < pathArgs.size())
             return pathArgs[i];
         return "";
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -3,6 +3,7 @@
 
 #include <ESP8266WebServer.h>
 #include <vector>
+#include <assert.h>
 
 template<typename ServerType>
 class RequestHandler {
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -25,7 +25,7 @@ protected:
     std::vector<String> pathArgs;
 
 public:
-    String pathArg(unsigned int i) { 
+    const String& pathArg(unsigned int i) { 
         assert(i < pathArgs.size());
         return pathArgs[i];
     }
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -23,7 +23,6 @@ private:
 	
 protected:
     std::vector<String> pathArgs;
-    bool _isRegex = false;
 
 public:
     const String& pathArg(unsigned int i) { 
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -25,9 +25,8 @@ protected:
 
 public:
     String pathArg(unsigned int i) { 
-        if (i < pathArgs.size())
-            return pathArgs[i];
-        return "";
+        assert(i < pathArgs.size());
+        return pathArgs[i];
     }
 };
 
--- a/libraries/ESP8266WebServer/src/detail/RequestHandler.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandler.h
@@ -2,6 +2,7 @@
 #define REQUESTHANDLER_H
 
 #include <ESP8266WebServer.h>
+#include <vector>
 
 template<typename ServerType>
 class RequestHandler {
--- a/libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
+++ b/libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
@@ -2,6 +2,8 @@
 #define REQUESTHANDLERSIMPL_H
 
 #include <ESP8266WebServer.h>
+#include <string>
+#include <regex>
 #include "RequestHandler.h"
 #include "mimetable.h"
 #include "WString.h"
@@ -18,15 +20,16 @@ public:
     , _uri(uri)
     , _method(method)
     {
-        int numParams = 0, start = 0;
-        do {
-            start = _uri.indexOf("{}", start);
-            if (start > 0) {
-                numParams++;
-                start += 2;
-            }
-        } while (start > 0);
-        pathArgs.resize(numParams);
+        _isRegex = uri.startsWith("^") && uri.endsWith("$");
+        if (_isRegex) {
+            std::regex rgx((_uri + "|").c_str());
+            std::smatch matches;
+            std::string s{""};
+            std::regex_search(s, matches, rgx);
+            pathArgs.resize(matches.size() - 1);
+        } else {
+            pathArgs.resize(0);
+        }
     }
 
     bool canHandle(HTTPMethod requestMethod, String requestUri) override  {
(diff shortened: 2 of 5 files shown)


libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -15,16 +15,55 @@ class FunctionRequestHandler : public RequestHandler {
     , _uri(uri)
     , _method(method)
     {
+		int numParams = 0, start = 0;
```
d-a-v: indentation (or do not use tabs)

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -15,16 +15,55 @@ class FunctionRequestHandler : public RequestHandler {
...
+        do {
+            start = _uri.indexOf("{}", start);
+            if (start > 0) {
+                numParams++;
+                start += 2;
+            }
+        } while (start > 0);
+        pathArgs = std::vector<String>(numParams);
```
d-a-v: Is it better than `pathArgs.resize(numParams)` ?

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -15,16 +15,55 @@ class FunctionRequestHandler : public RequestHandler {
...
             return false;
 
-        if (requestUri != _uri)
-            return false;
+        if (_uri == requestUri)
+            return true;
+
+        size_t uriLength = _uri.length();
+        unsigned int pathArgIndex = 0;
+        unsigned int j = 0;
```
d-a-v: could `j` be renamed to something more explicit ?
some more comments are welcome

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
Bmooij: Removed tabs

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
Bmooij: Changed to ```pathArgs.resize(numParams)```

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
Bmooij: renamed j to requestUriIndex

Bmooij:
One note: **the order of the paths are key!**

Given
```
ESP8266WebServer server(80);
server.on("/users/{}", handleUsersCall);
server.on("/users/{}/devices/{}", handleDevicesCall);
server.begin();
```

When calling ```http://YOUR_IP/users/3/devices/1``` only the function ```handleUsersCall``` will be called.
With ```server.pathArg(0) == "3/devices/1"```

In order to make this work:
```
ESP8266WebServer server(80);
server.on("/users/{}/devices/{}", handleDevicesCall);
server.on("/users/{}", handleUsersCall);
server.begin();
```

d-a-v:
> In order to make this work:

Could you add this explanation in your example ?

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -490,6 +490,11 @@ void ESP8266WebServer::_streamFileCore(const size_t fileSize, const String & fil
   send(200, contentType, "");
 }
 
+String ESP8266WebServer::pathArg(unsigned int i) {
```
d-a-v: Since you are returning an already allocated string, then return-type can be `const String&`.
That would save time and heap.
About `return ""` below, maybe we could define a global `extern String nullString;` in `WString.h`
@devyte what do you think ?

devyte: No, that would uselessly allocate an empty String object from the heap.
I think the correct thing to do here is to return String().

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -15,16 +15,55 @@ class FunctionRequestHandler : public RequestHandler {
     , _uri(uri)
     , _method(method)
     {
+        int numParams = 0, start = 0;
+        do {
+            start = _uri.indexOf("{}", start);
```
devyte: Why {}? Why not * for glob style notation, or .* for regex notation?

libraries/ESP8266WebServer/src/detail/RequestHandler.h
```diff
@@ -14,6 +16,16 @@ class RequestHandler {
 
 private:
     RequestHandler* _next = nullptr;
+
+protected:
+    std::vector<String> pathArgs;
+
+public:
+    String pathArg(unsigned int i) { 
+        if (i < pathArgs.size())
```
devyte: If the index is out of bounds, do we really want to silently fail? Or is it a programming error? If the latter, there should be an assert checking the index. Then, due to the assurance, the return type can be a const reference instead of by value.

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
devyte: See my other comment about returning a const ref.

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
Bmooij: The reason that I didn't have chosen for the regex **.*** notation is that it would imply that regex is used. You would expect that ```/users/[0-9]+``` will work too.

The reason that I did use the {} was because I originally wanted a named path arguments:
```
server.on("/users/{userId}/devices/{deviceId}/sensors/{sensorId}", []() {
    String userId = server.pathArg("userId");
    String deviceId = server.pathArg("deviceId");
    String sensorId = server.pathArg("sensorId");
    ...
 });
```
I removed the naming because, you know the index location of the path args.

libraries/ESP8266WebServer/src/detail/RequestHandler.h
```diff
@@ -22,9 +22,8 @@ class RequestHandler {
 
 public:
     String pathArg(unsigned int i) {
```
d-a-v: Then you can return `const String&` in all `pathArg()`

Bmooij:
> One note: **the order of the paths are key!**
> 
> Given
> 
> ```
> ESP8266WebServer server(80);
> server.on("/users/{}", handleUsersCall);
> server.on("/users/{}/devices/{}", handleDevicesCall);
> server.begin();
> ```
> When calling `http://YOUR_IP/users/3/devices/1` only the function `handleUsersCall` will be called.
> With `server.pathArg(0) == "3/devices/1"`
> 
> In order to make this work:
> 
> ```
> ESP8266WebServer server(80);
> server.on("/users/{}/devices/{}", handleDevicesCall);
> server.on("/users/{}", handleUsersCall);
> server.begin();
> ```

The order no longer matters.
Path arguments may not contain '/'.
If a '/' needs to be a path argument, a URL encoded value can be used (%2F)

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
d-a-v: I manually patched constref and used `emptyString`.

devyte:
@Bmooij I've been looking at this from different angles and... the {} syntax bothers me, it's just not standard. 
I'd personally prefer regex, but that can get complicated to use, and the patterns tend to be a bit error prone.
How about using glob (glob.h is available)? It's simple and standard, and makes use of a standard lib for the algorithm, so no custom parsing/string manipulation.

Bmooij:
@devyte I'm open for every thing. I'm not familiair with glob. Can you show an example how the path would look like?

devyte:
/users/*/devices
If you've ever done something like this in a Linux console, you've used glob patterns:
ls -l myfile*
or:
ls -l *cpp

devyte:
This requires further discussion, so pushing back.

Bmooij:
After using glob patterns, my personal preference goes to regex.
I added the regex support. Please let me know your thoughts on this.

usage:
```
server.on("^\\/users\\/([0-9]+)\\/devices\\/([0-9]+)$", []() {
    String user = server.pathArg(0);
    String device = server.pathArg(1);
    server.send(200, "text/plain", "User: '" + user + "' and Device: '" + device + "'");
  });
```

_**note**_:
The ```^``` upfront and ```$``` at the end are required to detect it is a regex pattern.

earlephilhower:

devyte:
This has a merge conflict, and needs further consideration to evaluate against the alternatives. 
As a side note, I'm evaluating coexistence of all three.

Pushing back.

Bmooij:
I forgot this pull request. Tomorrow I will fix the merge conflicts and implement the `glob patterns`, like @denyte proposed. 

A problem that I encountered from a similar pull request, was that regex is a large include (100k).

devyte:
@Bmooij #5467 I already implemented glob. All 3 alternatives are in PRs.
The question is: which do we use. 
I was thinking of all three, with a different method nameing scheme, e. g. :
```cpp
onRegEx() 
onGlob() 
onDontKnowNameOfThisOne()
```

With this, the regex bin increase should happen only if it is actually used, and yet all three alternatives are available.

Bmooij:
@devyte What do you think of the following solution:

```
server.on("a/static/uri", callbackfn); // this will automatically transform to Uri("a/static/uri")
server.on(Uri("an/other/static/uri"), callbackfn);
server.on(UriRegex("^\\/users\\/([0-9]+)\\/devices\\/([0-9]+)$"), callbackfn);
server.on(UriGlob("/users/*/devices/*"), callbackfn);
server.on(UriBracket("/users/{}/devices/{}"), callbackfn);
```

devyte:
That is a very good idea! 
Notes for the uri objects:
- the constructors need to be explicit
- overloads in each for const char * and String args
- probably const String member
- UriBrace or UriBraces instead of brackets
- mind move semantics
- getter methods

devyte:
@Bmooij could you please pick up the three implementations from the PRs and make a new one with your idea? That new one would then supersede the three.

Bmooij:
There is an issue when I want to include `<regex>`

lib\ESP8266WebServer\src/detail/RequestHandlersImpl.h:25:13: error: 'regex' is not a member of 'std'

devyte:
Closing in favor of #6696.

Issue #5216, opened 2018-10-07
StreamHttpsClient.ino crashes


Jeroen88:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					NodeMCU
- Core Version:			git pull upstream HEAD
- Development Env:		Arduino IDE
- Operating System:		Ubuntu

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			qio
- Flash Size: 		4MB/1MB
- lwip Variant:   v2 Lower Memory
- Reset Method:		ck
- Flash Frequency:	80Mhz
- CPU Frequency:		160MHz
- Upload Using:		SERIAL
- Upload Speed:   230400 (serial upload only)

### Problem Description

Running StreamHttpsClient.ino crashes. Remarkable is that when I enable Debug Level: SSL+HTTP_CLIENT it runs fine. @earlephilhower as requested!

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```/**
   StreamHTTPClient.ino

    Created on: 24.05.2015

*/

#include <Arduino.h>

#include <ESP8266WiFi.h>
#include <ESP8266WiFiMulti.h>

#include <ESP8266HTTPClient.h>

ESP8266WiFiMulti WiFiMulti;

void setup() {

  Serial.begin(115200);
  // Serial.setDebugOutput(true);

  Serial.println();
  Serial.println();
  Serial.println();

  for (uint8_t t = 4; t > 0; t--) {
    Serial.printf("[SETUP] WAIT %d...\n", t);
    Serial.flush();
    delay(1000);
  }

  WiFi.mode(WIFI_STA);
  WiFiMulti.addAP("SSID", "PASSWORD");

}

void loop() {
  // wait for WiFi connection
  if ((WiFiMulti.run() == WL_CONNECTED)) {

    HTTPClient http;

    BearSSL::WiFiClientSecure *client = new BearSSL::WiFiClientSecure ;

[...]

PR #5220, opened 2018-10-10
Bugfix/esp8266 http client

Jeroen88:
Fix for issue #5216 + minor changes: removing unnecessary tests and make sure that end() also closes the connection if it was a Keep-Alive connection.
Changed BasicHttpsClient.ino to also allocate the BearSSL::WiFiClientSecure on the heap in stead of on the stack.

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -123,7 +123,7 @@ HTTPClient::HTTPClient()
 HTTPClient::~HTTPClient()
 {
     if(_client) {
-        _client->stop();
+        DEBUG_HTTPCLIENT("[HTTP-Client][~HTTPClient] end() not called before destruction of HTTPClient\n");
     }
     if(_currentHeaders) {
         delete[] _currentHeaders;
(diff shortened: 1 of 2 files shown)


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -367,6 +367,7 @@ bool HTTPClient::begin(String host, uint16_t port, String uri, const uint8_t htt
  */
 void HTTPClient::end(void)
 {
+    _canReuse = false;
```
earlephilhower: Even if the user calls `::end`, _client is never cleared in this code so the object will always print the debug message nastygram when you destroy the object.  Suggest the simplest thing here is to set _client to NULL at this point in the code.

I would also suggest setting _tcpDeprecated to NULL here as well, since that will take care of deletion when the emulated object is no longer needed.

Jeroen88: ::end() calls ::disconnect(). If not connected both pointers are already NULL, otherwise _client will be set to nullptr and _tcpDeprecated will be reset to nullptr. Agree?

earlephilhower: Ah, got it.  Didn't catch that since disconnect() isn't part of the diff so it's not shown, but checking the full code it looks OK.

Jeroen88:
@earlephilhower I looked at the diff and see that everywhere _canReuse is set to false just before the call to ::end(). But ::end() also sets _canReuse to false. It doesn't harm, but it isn't necessary. I create another PR.

PR #5221, opened 2018-10-10
Add stack repainting call to cont_utils

earlephilhower:
Allow the unused stack to be reset to the check value at any time in
the application, allowing for delta-stack calculations to be done.

Add in some dumping in the BearSSL_Validation example to show the
usage for those that care.


earlephilhower:
Added an ESP class helper, and am now using it in the example.

PR #5226, opened 2018-10-11
I2C slave support (originally by bjoham)

suculent:
Resolved conflicts, fixed style, cleaned up in-progress comments and useless lines of code.


devyte:
Ok, given @suculent 's feedback (in prior incarnations of this PR, not in this one), I'm merging this as-is. If there are issues, they can be addressed with new PRs.

freeck:
From version 2.5.0 on the I2C-library core_esp8266_si2c.c has been extended with slave-functionality. For a performance reason there was decided to put a lot of functions in IRAM!
I don't use these functions but they are linked into the image nevertheless....but this costs 2596 bytes of our scarce IRAM!

r3na:
@suculent Did you try this code using different I2C drivers as master? I will like another comment I wrote:
https://github.com/esp8266/Arduino/issues/5762#issuecomment-476330325

rustyx:
ESP8266 can never be a compliant I2C slave; there is no hardware support, and a bit-by-bit software solution isn't going to be fast enough for the standard I2C speed of 100kHz because 1) the interrupt won't arrive fast thought and 2) the GPIO bus clock is slower than 100kHz.

That's why this solution can handle 14kHz max, which is useless since most other devices run at 100kHz.

Try ESP32, it has I2C hardware.

PR #5232, opened 2018-10-12
Bugfix/esp8266 http client

Jeroen88:
Removed unnecessary code


Jeroen88:
@earlephilhower The current version of the library does not handle Connections that may be reused correctly. If the programmer calls end() directly, the connection may be _reused_, so it should **NOT** be _client->stop() nor should _client = nullptr and also not _tcpDprecated.reset(nullptr)!!!

If end() is called from begin() the connection should also not be closed, for the same reason.

However, just before destruction ~ESP8266HTTPClient() the connection should be _client->stop(), because now the connection will be definitely closed. In the previous versions the destructor took care of this, but not anymore because of the bugfix.
I think it would be the best solution if it would still be possible to have the destructor stop the connection, but I do not know how. The only option I can think of is a new function that should be called before the ESP8266HTTPClient is destructed, but this changes the API and is error prone. Any ideas?

PR #5234, opened 2018-10-12
Update debugging.rst

teo1978:
PLEASE REVIEW
The current link definitely points to the wrong line, where there's nothing remotely resembling a list of DEBUG_ESP_* definitions. There's not an actual "full list" in the whole boards.txt file really, but this is the closest thing that can be found in that file.
Some additional explanation is needed though, unless there is an actual list and it's somewhere else.

### DIFF
--- a/doc/Troubleshooting/debugging.rst
+++ b/doc/Troubleshooting/debugging.rst
@@ -74,7 +74,7 @@ Debug Level
 All defines for the different levels starts with ``DEBUG_ESP_``
 
 a full list can be found here in the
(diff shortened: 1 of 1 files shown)


dragondaud:
This would be the link in boards.txt.py script:
https://github.com/esp8266/Arduino/blob/master/tools/boards.txt.py#L1045-L1047

teo1978:
I think I have applied the requested change. Otherwise feel free to change it yourself and reject this PR.
I'm not familiar with git/github.

devyte:
Thanks @teo1978 !

Issue #2581, opened 2016-10-06
WEMOS D1 mini Pro - 16M supporting


4refr0nt:
#### Hardware

Hardware:           WEMOS D1 mini Pro - 16M
#### What's new with D1 mini Pro
- 16M bytes (128M bit) Flash
- External antenna connector
- Built-in ceramic antenna
- New CP2104 USB-TO-UART IC
- Same size as D1 mini, but more light

[View D1 mini Pro on WeMos Aliexpress store](http://ali.ski/gsnRGA)

WereCatf:
This is not an issue and this is not the place to advertise your affiliate-links.

TotallyInformation:
Hi, I've got 2 of these now and would like to understand how to update the boards.txt to make use of the 16M memory?

Not as simple as I'd hoped. I'm guessing that these are a copy of the D1 Mini but with the larger FLASH chips attached.

WereCatf:
@TotallyInformation It'd be best to make a proper variant out of it. I can take a stab at it in a few hours.

TotallyInformation:
Fab, thanks WereCatf.

WereCatf:
@TotallyInformation Okay, copy https://github.com/WereCatf/Arduino/blob/WemosD1MiniPro/boards.txt over your existing one and that should be it. I do not have one of these boards, so I can't actually verify that it works, so it would be nice if you'd chime back here if it works as it should.

pieman64:
@WereCatf I have some of the Pro's and did some manual adjustments to boards.txt. Am I right in thinking we still can't access the memory from 4M to 16M?

WereCatf:
@pieman64 I don't know, I ain't got any 16MB-boards. If you've got the pros then you could just try.

wfraga18:
why WEMOS never answer the forum questions ?? launch a product, but don't have support !

HOW TO USE 16MB ???

WereCatf:
@wfraga18 Ask them, not us. We don't work for Wemos. As for using it: how about you actually read the replies here, instead of raging? I provided a new boards.txt above you can try.

TotallyInformation:
Thanks @WereCatf, I'm at a conference for a few days, I'll try after that. Good work.

4refr0nt:
[...]

PR #5249, opened 2018-10-14
fix SPIFFS when not enabled in build options

### LINKED ISSUE #5246: SPIFFS crashes Fatal exception 0(IllegalInstructionCause) if SPIFFS not assigned
### Basic Infos

- [x ] If there is a stack dump, I have decoded it.

#### Platform

- Hardware:				WemosD1 and NodeMCU V1.0
- Core Version:			12/10/2018 00:34
- Development Env:		Visual Micro
- Operating System:		Windows

### Problem Description

More of a 'gotcha' than a problem so it turns out.
On previous builds if a SPIFFS region wasn't assigned calling SPIFFS.begin() worked silently (ok, I guess there would have been problems later on).
On the current git calling begin() causes:
Fatal exception 0(IllegalInstructionCause):
epc1=0x4000e25d, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000000, depc=0x00000000

As I said, more of a gotcha than a bug but it would be kinder if there was some sort of compiler warning.

d-a-v:
fix #5246 
~also  may help debugging #2581 
by adding optional `SPIFFS.begin(false)`: will not try to format (default is true)~


devyte:
@d-a-v is anything pending here?

Issue #5237, opened 2018-10-13
WifiClient crashes - _yield and millis()


mateusz-szafraniec:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[2eb5b5600a3631cdfb9ddd1150505d82e38f3406]
- Development Env:		[Arduino IDE]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Wemos D1 mini r2]
- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description
My program randomly crashes while trying to post data via WifiClient.
Sometimes at client.connect(), sometimes at client.println(), but always latest stack trace points to millis and _yield
Sometimes program can run hours without crashes, sometimes few minutes.

### Sketch

```cpp

void wyslijPomiar()
{
  Serial.println(F("wyslijPomiar"));
  WiFiClient client;
  unsigned long connectStart = millis();
  if (!client.connect(adresSerwera, portSerwera)) { //sometime crashes here
      Serial.println(F("Brak połączenia z serwerem"));
      return;
  }
  client.flush();
  client.setNoDelay(true);
  bool aktualizuj = false;
  String PostData ="pm1_0_std=" + String(PM01Avg);
  PostData += "&pm2_5_std=" + String(PM25Avg);
  PostData += "&pm10_std=" + String(PM10Avg);
  PostData += "&pm1_0=" + String(PMa01Avg);
  client.println(F("POST /esp/upload.php HTTP/1.1"));
  client.println("Host: "+ String(adresSerwera));
  client.println(F("Cache-Control: no-cache")); //sometime crashes here
  client.print(F("Authorization: Basic "));
  client.println(base64::encode(AUTH));
[...]

PR #5254, opened 2018-10-17
fix: WiFiClient::flush() yields but can be called from events

d-a-v:
for #5237 

Bug introduced by #5167 which replaced `delay()` by `yield()`,
and that should have been `esp_yield()` which is the one `delay()` calls.
To be noted for current maintainers included myself:
`delay()` and `esp_yield()` can be called from sys context

### DIFF
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -331,11 +331,9 @@ public:
                 prevsndbuf = sndbuf;
                 // We just sent a bit, move timeout forward
                 last_sent = millis();
-
-                // yield because we are staying longer
-                // (only if possible because we can be called from events)
-                esp_yield();
             }
+
+            esp_yield(); // from sys or os context
 
             if ((state() != ESTABLISHED) || (sndbuf == TCP_SND_BUF)) {
                 break;
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -331,9 +331,11 @@ public:
                 prevsndbuf = sndbuf;
                 // We just sent a bit, move timeout forward
                 last_sent = millis();
-            }
 
-            yield();
+                // yield because we are staying longer
+                // (only if possible because we can be called from events)
+                esp_yield();
+            }
 
             if ((state() != ESTABLISHED) || (sndbuf == TCP_SND_BUF)) {
                 break;


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -331,9 +331,11 @@ class ClientContext
                 prevsndbuf = sndbuf;
                 // We just sent a bit, move timeout forward
                 last_sent = millis();
-            }
 
-            yield();
+                // yield because we are staying longer
+                // (only if possible because we can be called from events)
+                esp_yield();
```
earlephilhower: Did you mean to move the esp_yield inside the if-block?  Before it would yield each loop, here it will only yield if it actually sends something so it seems like it'll hard-lock, no?

devyte: The esp_yield() should not be conditional.
Also, while in the sys context, i.e.: in an async callback: if we're relying on receiving something and detecting that reception during the yield, that won't happen => infinite loop. The async way of doing this is to break the loop body into its own function, and schedule the function to get called by the sys context. Then, in the function body, if nothing has happened and we need to continue waiting, the function is re-scheduled, and return.
Personally, I don't like the idea of trying to allow sync calls during async callbacks. In fact, I would disallow such usage, period. The call sequence and use models are very different, and there's bound to be something that will bite hard. Async callbacks in the sys context have stricter timing requirements (e.g.: sending several packets and waiting for the acks or whatever could go boom), you can't yield and then continue execution, if you're waiting on something you have to return and come back later, etc.
What I would _love_ to have is an async socket that can work from async callbacks, and a sync socket that is built on top of it by serializing (i.e.: waiting for) the async calls. One possibility would be to implement a new WiFiClient/Server built on top of @me-no-dev 's async libs. I'm not sure how that serialization would be implemented, I'd have to think about it, but if we did manage it, we could drop a lot of code, unify with the async libs, get an async framework into the core, and overall get lots of goodies.

ascillato:
Hi

@d-a-v, @earlephilhower, @devyte 

We have done several tests and we found that this last commit make the device unresponsive to webserver.

Reverting to https://github.com/esp8266/Arduino/commit/ad7cb635eda1212eb82584adf28e965bad674634 all works again.

Please, could you revert this last commit?

ascillato:
Hi,

As discussed in https://github.com/esp8266/Arduino/commit/e5493552215f279b0ca3c069d5ebcdfe06773d62 changing `esp_yield()` to `delay(0)` make it the webserver to work again.

Thanks. :+1:

PR #5255, opened 2018-10-17
spiffs auto format selector (true by default)

d-a-v:
To help debugging #2581 (wemos/lolin d1 mini pro and/or 16MB flash):
`ESP.setAutoFormatSPIFFS()` will not format SPIFFS when mount fails.

Helpers set in ESP:: because this is the most simple way.
SPIFFS real interface is hard to reach from userland.

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -574,3 +574,13 @@ String EspClass::getSketchMD5()
     result = md5.toString();
     return result;
 }
+
+bool _spiffsAutoFormat = true;
+
+void EspClass::setAutoFormatSPIFFS (bool autoFormat) {
+    _spiffsAutoFormat = autoFormat;
+}
+
+bool EspClass::getAutoFormatSPIFFS () {
+    return _spiffsAutoFormat;
+}
--- a/cores/esp8266/Esp.h
+++ b/cores/esp8266/Esp.h
@@ -153,6 +153,9 @@ class EspClass {
         bool eraseConfig();
 
         inline uint32_t getCycleCount();
+
+        bool getAutoFormatSPIFFS ();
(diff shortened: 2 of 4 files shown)


d-a-v:
To the question: why not `SPIFFS.begin((autoFormat=)false)`, the answer is:
SPIFFS's type is not `SPIFFSImpl::` but `FS::` and this is a bag of bones.

This PR is mainly here to help debugging.
We can remove that once we have reliable 16MB SPIFFS.
Or rewrite it when FS (sd, spiffs) is refactored.

earlephilhower:
@d-a-v, I've got the same issue.  Maybe if we did something like make a per-FS configuration object?
````
class FSConfig {
public:
  FSConfig();
}; // empty base

class SPIFFSConfig : public {
public:
 SPIFFSConfig(bool autoFormat = false) {_autoFormat = autoFormat; }
 void disableAutoFormat() { _autoformat = false; }
 bool _autoFormat;
};

class SDFSConfig : public FSConfig {
public:
  SDFSConfig(int cs = 4, SPISettings spi =SPI_MHZ(10)) {
    _cs = cs; _spi = spi;
   }
  int _cs;
  SPISettings spi;
};
````

Then the FSImpl object begin signature becomes:
`begin( FSConfig &cfg = nullptr );`

`SPIFFS::begin` can then `static_cast` its pointer to SPIFFSConfig, etc.

User code then becomes
`SPIFFS.begin()` or `SPIFFS.begin( SPIFFSConfig(false) );` or `SPIFFSConfig cfg; cfg.disableAutoFormat(); SPIFFS.begin(cfg);`

earlephilhower:
@d-a-v, take a look at #5525 which is my proposal for this (SPIFFS and SDFS now, LittleFS once one or the other merges).

PR #5260, opened 2018-10-18
Actually let OS advance in ::flush

earlephilhower:
Replace esp_yield with delay(0) which calls both the yield and
the scheduler.

Supersedes PR #5254


earlephilhower:
@ascillato I think we may have a second spot where we were hoping to delay

PR #5268, opened 2018-10-22
Allow override of macros in core_version.h

SenH:
So we can override the build description when compiling dev branches with a command flag like `-DARDUINO_ESP8266_GIT_DESC=...`


d-a-v:
This file is overridden during the build process in the build path by `platform.txt`L75-79
(`/tmp/arduino-build.../core/core_version.h` for linux, elsewhere for windows and macOS).
So if you need this feature, I think you need to update this file.
Unless I miss something, I don't think your changes have any effect.
Did you actually noticed a change with them ?

If you need to be able to check what core you are using at compile time, you may find some interest in #5269 (coincidence)

SenH:
This doesn't change anything for building via Arduino.app or platform.txt but I'm using a [custom makefile](https://github.com/plerup/makeEspArduino).

 I can add `BUILD_EXTRA_FLAGS=-DARDUINO_ESP8266_GIT_DESC=$(ESP_ARDUINO_VERSION)
`in my makefile and then `ESP.getFullVersion()`will output the current Git version used for the build. Without the PR it's impossible to redefine these macros.

PR #5292, opened 2018-10-29
Add missing decrement operator to I2C clockCount

### LINKED ISSUE #4063: twi_status clockCount  wrong usage ??

devyte:
Fixes #4063 .
I am not changing the clockCount init value of 20 to 10 because I don't know the impact. Given that it's essentially a timeout, it shouldn't be relevant.


miky2k:
It's not timeout , it's minimum number of clock pulse to put a I2C bus in a consistent state.

Issue #5094, opened 2018-08-30
Serial.flush should change to Serial.flush(uint32_t data_width, uint32_t parity, uint32_t stop)


yldzmuhammed:
#### Platform

- Hardware:				[LoLin new NodeMCU V3 board]
- Core Version:			[Updated today.So should be 2.4.2?]
- Development Env:		[Arduino IDE|Platformio|Make|other]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Nodemcu 0.9]
- Flash Size: 		[4MBB]
- lwip Variant:   [v2 Lower Memory|Higher Bandwidth]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600]

### Problem Description

I am building a Rs485 project. So esp has to control RE and DE pins. I have done some test and finaly find a easy way (for me at least). I added second flush function like:
```

void HardwareSerial::flush(uint32_t data_width, uint32_t parity, uint32_t stop)
{
    if(!_uart || !uart_tx_enabled(_uart)) {
        return;
    }

    uart_wait_tx_empty(_uart);
    //Workaround for a bug in serial not actually being finished yet
    //Wait for 8 data bits, 1 parity and 2 stop bits, just in case
    delayMicroseconds(((data_width + parity + stop) * 1000000) / uart_get_baudrate(_uart) + 1);
}

```

I belive in the HardwareSerial.h file the prototype should change:

`void flush(uint32_t data = 8_width, uint32_t parity = 0, uint32_t stop = 1);`

I user using serial port with default settings (8N1) then user dont have to anything. Otherwise, function must call with serial config parameters.

With this function, esp can handle tx enable, rx enable nicely.

What do you think? Is there any better way to handle RS485 comm direction problem?

NOTE: I am not a cpp expert.

d-a-v:
Those parameters should not come to this function that way.
They are already internally stored and their values should be extracted from there.

What would be nice is to move this workaround into `uart.c:uart_wait_tx_empty()`, in there you can read `USC0(uart->uart_nr)` which is the serial mode (8N1 or whatever).
With a couple of defines, you can extract from it the width, the parity and the stop bit.
```
#define MODE2WIDTH(mode) (((mode) >> 2) + 5) // see UART_NB_BIT_* 
```
then `MODE2WIDTH(USC0(uart->uart_nr))` would give you the needed `data_width`.

We would ~love~be thrilled if you could make a pull request !
[...]

PR #5293, opened 2018-10-29
Serial.flush modification

dsv19:
Added a few defines to extract width, parity and stop from uart mode in uart.c as suggested in the comments in #5094 and modified flush in HardwareSerial.cpp file.


cores/esp8266/uart.c
```diff
@@ -354,15 +358,16 @@ uart_tx_free(uart_t* uart)
     return UART_TX_FIFO_SIZE - uart_tx_fifo_available(uart->uart_nr);
 }
 
-void 
+uint8_t 
 uart_wait_tx_empty(uart_t* uart)
 {
+    uint8_t tmp = MODE2WIDTH(USC0(uart->uart_nr)) + MODE2PARITY(USC0(uart->uart_nr)) + MODE2STOP(USC0(uart->uart_nr));
```
d-a-v: You can't do that here, `uart` can be NULL

d-a-v:
CI fails: uart.h needs to be committed too.
Should we count the implicit start bit too (+1 in the result, 12 instead of 11) ?
Is it necessary to wait (to change to 11/12) when the returned value is 0 ?

dsv19:
I'm not sure about the implicit start bit but it is not necessary to wait when the returned value is 0 since the HardwareSerial::flush function already handles a situation when uart is null.

d-a-v:
Start bit is always there, so IMHO 1 should be added to the `tmp` calculus.
Btw could you rename it for the readers, like `bit_length` or anything more explicit.
After a second read, you don't even need this variable, the value can be directly returned,
then there would be a comment to describe what's returned.

dsv19:
Ignore last commit

cores/esp8266/uart.c
```diff
@@ -422,15 +426,15 @@ uart_tx_free(uart_t* uart)
...
 
-void 
+uint8_t 
 uart_wait_tx_empty(uart_t* uart)
 {
     if(uart == NULL || !uart->tx_enabled)
-        return;
-
+        return 0;
+    uint8_t bit_length = MODE2WIDTH(USC0(uart->uart_nr)) + MODE2PARITY(USC0(uart->uart_nr)) + MODE2STOP(USC0(uart->uart_nr)) + 1;
```
devyte: This calculation has nothing to do with the function in which it is being executed. I suggest reverting uart_wait_tx_empty(), and putting this code in its own function, e.g.:
```
inline uint8_t uart_get_bit_length(const int uart_nr)
{
  return MODE2WIDTH(USC0(uart_nr)) + MODE2PARITY(USC0(uart_nr)) + MODE2STOP(USC0(uart_nr)) + 1;  //+1 for the start bit, which is always there
}
```

Then, use it in HardwareSerial like so.
```
     uart_wait_tx_empty(_uart);	
     //Workaround for a bug in serial not actually being finished yet
     //Wait for 8 data bits, 1 parity and 2 stop bits, just in case	  
     delayMicroseconds(1000000 * uart_get_bit_length(_uart->uart_nr) / uart_get_baudrate(_uart) + 1);
```

Issue #3213, opened 2017-05-08
ESP8266WiFi: add support for String arguments to begin and softAP methods


treii28:
This is for the Arduino platform. The Arduino standard library includes a type String. Why do I still need to add a .c_str() to the end of all my parameters???

e.g.: WiFi.softAP(ssid.c_str(). passwd.c_str());

Humancell:
You don't have to add it to all of your parameters, but in this case the API that you referenced is expecting two pointers to char buffers.

The API is designed to be similar to the Arduino WiFi APIs ... which also take char buffers.

https://www.arduino.cc/en/Reference/WiFiBegin

igrr:
The latest official WiFi101 library does support String arguments:
https://github.com/arduino-libraries/WiFi101/blob/master/src/WiFi101.h#L116
so i suppose this is a valid enhancement request.

treii28:
but aren't most of the Arduino WiFi APIs built to talk to a connected ESP8266 module?

Humancell:
No.  The Arduino WiFi APIs were designed to support the Arduino WiFi shield.  You know that you can easily Google this type of information:  https://www.arduino.cc/en/Reference/WiFi

This project is very different:

_> This project brings support for ESP8266 chip to the Arduino environment. It lets you write sketches using familiar Arduino functions and libraries, and run them directly on ESP8266, no external microcontroller required.
> 
> ESP8266 Arduino core comes with libraries to communicate over WiFi using TCP and UDP, set up HTTP, mDNS, SSDP, and DNS servers, do OTA updates, use a file system in flash memory, work with SD cards, servos, SPI and I2C peripherals._

devyte:
Closing via #5295

PR #5295, opened 2018-10-30
ESP8266WiFi softAP and begin

dav1901:
Added support for String arguments to begin and softAP methods in ESP8266WiFi library as suggested in #3213 issue.


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -38,6 +38,7 @@ class ESP8266WiFiSTAClass {
 
         wl_status_t begin(const char* ssid, const char *passphrase = NULL, int32_t channel = 0, const uint8_t* bssid = NULL, bool connect = true);
         wl_status_t begin(char* ssid, char *passphrase = NULL, int32_t channel = 0, const uint8_t* bssid = NULL, bool connect = true);
+        wl_status_t begin(const String& ssid, const String& passphrase = String('\0'), int32_t channel = 0, const uint8_t* bssid = NULL, bool connect = true);
```
d-a-v: `emptyString` can be used instead of `String('\0')` or `String()` (recently added in WString.h for that purpose)

Issue #1902, opened 2016-04-12
Cannot call more than once ESP8266HttpClient POST()


far5893:
If i define ESP8266HttpClient websebd;
websend.begin("http://192.168.1.1/htpxml.php");
websend.addHeader("Content-Type", "text/xml");
websend.addHeader("Cookie:","XDEBUG_SESSION_START=session0"); 

then i call it two times:

websend.POST("some xml text");
websend.end();

websend.POST("another  xml text");
websend.end();

POST /htpxml.php HTTP/1.1
Host: 192.168.1.1
User-Agent: ESP8266HTTPClient
Connection: close
Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0
Content-Type: text/xml
Cookie:: XDEBUG_SESSION_START=session0
Content-Length: 744
Content-Length: 744

(Wireshark log)

if(payload && size > 0) {
        addHeader("Content-Length", String(size));
    }

don't know if it's the right mode to use object, but i would like to avoid 
recreate the same object every time i have to send same file to same 
url with same header!!

 

The code in sendrequest that add "definitely" content size to header. 
 if(payload && size > 0) {
        addHeader("Content-Length", String(size));
 }

TheAustrian:
I don't know if it's cause you call the .end() but regarding setting up stuff only once I made myself a function that I then use in the loop().

int SendSensorData (String Sensordata) {
  // returns negative value if unsuccessful
  // returns 0 if successful
  String returnvalue = "-1";
  if (Sensordata != "" ) {
    HTTPClient http;
    http.begin(phppage);
    http.addHeader("Content-Type", contenttype);
    int httpCode = http.POST(Sensordata);
    if (httpCode != 200) {
      Serial.println("not successful");
      } else {
      returnvalue = http.getString();
      Serial.println("successful:" + returnvalue);   
      }
    http.end();
  } 
  int returnval = returnvalue.toInt();
  return returnval;
}

far5893:
Apache does not agree with more than one Content-Length  line in request, 
i'dont know why your reuse work!!

PR #5309, opened 2018-11-05
Fix device/test_http_client tests

liebman:
update HTTPClient API usage
skip the second POST as end() has different semantics and nulls the client pointer
use bearssl in ssl tests
add delay in python side when shutting down http web server so MacOS does not complain about address already in use

### DIFF
--- a/tests/device/test_http_client/test_http_client.ino
+++ b/tests/device/test_http_client/test_http_client.ino
@@ -60,7 +60,7 @@ TEST_CASE("HTTP GET & POST requests", "[HTTPClient]")
         http.end();
 
         httpCode = http.POST("bar");
-        // its not expected to work but should not ccrash
+        // its not expected to work but should not crash
         REQUIRE(httpCode == HTTPC_ERROR_CONNECTION_REFUSED);
         http.end();
     }
--- a/tests/device/test_http_client/test_http_client.ino
+++ b/tests/device/test_http_client/test_http_client.ino
@@ -24,8 +24,9 @@ TEST_CASE("HTTP GET & POST requests", "[HTTPClient]")
 {
     {
         // small request
+        WiFiClient client;
         HTTPClient http;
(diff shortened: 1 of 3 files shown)


tests/device/test_http_client/test_http_client.ino
```diff
@@ -48,17 +50,20 @@ TEST_CASE("HTTP GET & POST requests", "[HTTPClient]")
...
         // can do two POST requests with one HTTPClient object (#1902)
+        WiFiClient client;
         HTTPClient http;
-        http.begin(getenv("SERVER_IP"), 8088, "/");
+        http.begin(client, getenv("SERVER_IP"), 8088, "/");
         http.addHeader("Content-Type", "text/plain");
         auto httpCode = http.POST("foo");
         Serial.println(httpCode);
         REQUIRE(httpCode == HTTP_CODE_OK);
+/* TBD: This test is broken because end() nulls the client*
```
earlephilhower: Thanks for doing the changes, you beat me to it.  I noticed this failing while making my own device tests for two PRs.

Any reason you don't want to reset the client after `.end()` and run the POST?  I'm not sure the test is caring about the `.end()` behavior here so much as a "Can I do a successful POST"?

liebman: The comment in the check says two back to back POSTs and points to this ticket: https://github.com/esp8266/Arduino/issues/1902 which does POST();end();POST(); and that's broken.  Should we test both old and new APIs?  And both BearSSL and AXTLS? (or will they go away soon enough?)  In any case the specific usage listed in the above ticket is broken.

liebman: I'm about to commit an update that will check that the second POST() after end() without another begin() does not crash. This includes a fix in HTTPClient::connect() where the crash was ocuring.

PR #5327, opened 2018-11-09
cleanup/unify flash sector size define value

### LINKED ISSUE #2421: clean up SPI_FLASH_SEC_SIZE and/or FLASH_SECTOR_SIZE

devyte:
Fixes #2421
Replaces #4509


devyte:
Further rework is possible to unify eboot defines as well, but that should be done separately.

devyte:
@d-a-v you've dabbled with sdk updates. Does this factoring out make sense? I'm worried about migrating to a newer sdk. Maybe this should be a patch or something?

devyte:
There are three code pieces involved: eboot, the sdk, and our core, and all three had duplicate definitions in one way or the other. The sdk comes from Espressif and could be updated, eboot is hosted by us and is not included in the sdk, and our core is our own.
My thinking was: 
- eboot was dependent only on itself only through code duplication. It's ok to have it depend on the sdk, but must reduce external dependencies when at all possible.
- core depends on sdk and eboot
- sdk depends only on itself

From the above, the flash geometry must be in the sdk. If I were to put them in eboot, the sdk would depend on eboot, which we don't want.

devyte:
Any further comments? Feel free to destroy my thinking above.
I'm still worried about maintaining this change across sdk updates.

devyte:
Given that there are no further comments, and in the interest of moving forward, I'm merging this as-is. Any issues that arise from updating the sdk in the future will be handled at that time.

PR #5330, opened 2018-11-09
Fix check for rtc mem sdk api in ESPClass

### LINKED ISSUE #2875: Inconsistent use of offset in ESP.rtcUserMemoryRead and ESP.rtcUserMemoryWrite in Esp.cpp
The variable `offset` in ESP.rtcUserMemoryRead and ESP.rtcUserMemoryWrite is used inconsistently. Either it should be in _bytes_ or _block of 4 bytes_. (Probably the latter)

The corresponding functions in (system_rtc_mem_read and system_rtc_mem_write) "user_interface.h" uses a multiple of 4 bytes for `src_addr` and `dst_addr`. The total rtc memory amount is 768 bytes, the first 256 bytes reserved for system use and 512 bytes available to the user!

In the "Esp.cpp" `offset` is used as a multiple of 4 bytes in these lines:
`    return system_rtc_mem_read(64 + offset, data, size);`
`    return system_rtc_mem_write(64 + offset, data, size);`
(64 blocks being the reserved memory, 256 bytes / 4) 

A few lines above these, a test uses `offset` as a single byte:
`    if (size + offset > 512)`
`        return false;`

If `offset`  represents _blocks of 4 bytes_, it should read
`if (size + offset*4 > 512)`
or else
`    return system_rtc_mem_read(64 + offset/4, data, size);`
`    return system_rtc_mem_write(64 + offset/4, data, size);`
(The latter being rather stupid)

By the way, does `offset` need to be `uint32_t` when it is converted to `uint8` in the system calls?

devyte:
Also:
-added extra check to disallow zero-size read/writes
-added some comments to explain the rtc mem layout, api, interface
Fixes #2875

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -155,7 +155,15 @@ offset is the user block number (block size is 4 bytes) must be >= 0 and <128
 data is a pointer to data, 4-byte aligned
 size is number of bytes in the block pointed to by data
 
-Same for write
+Same for write.
+
+Note: If the Updater class is in play, e.g.: the application uses OTA, the eboot
+command will be stored into the first 128 bytes of user data, then it will be
+retrieved by eboot on boot. That means that user data present there will be lost.
+Ref: 
+- discussion in PR #5330.
+- https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map#memmory-mapped-io-registers
+- Arduino/bootloaders/eboot/eboot_command.h RTC_MEM definition
 */
 
 bool EspClass::rtcUserMemoryRead(uint32_t offset, uint32_t *data, size_t size)
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -132,9 +132,35 @@ uint64_t EspClass::deepSleepMax()
 
 }
 
+/* 
+Layout of RTC Memory is as follows:
+Ref: Espressif doc 2C-ESP8266_Non_OS_SDK_API_Reference, section 3.3.23 (system_rtc_mem_write)
+
+|<------system data (256 bytes)------->|<-----------------user data (512 bytes)--------------->|
+
+SDK function signature:   
+bool	system_rtc_mem_read	(
+				uint32	des_addr,	
+				void	*	src_addr,	
+				uint32	save_size
+)   
+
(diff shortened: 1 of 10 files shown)


devyte:
Looks like I messed something up with git, but I think it should be resolved once #5327 is merged.

mcspr:
I think it should also mention Updater/eboot interaction, see https://github.com/esp8266/Arduino/issues/619#issuecomment-222657888
https://github.com/esp8266/Arduino/blob/74819a763bfb6e9890a57411dcea4aba221a778d/bootloaders/eboot/eboot_command.c#L48
https://github.com/esp8266/Arduino/blob/055748ff6fab345511985d684c9e67fb84cddaa5/cores/esp8266/Updater.cpp#L199
OTA update will overwrite first 128 bytes of user section

mcspr:
Per https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map#memmory-mapped-io-registers
`0x60001100` is system section, for 0x100 / 256 bytes
`0x60001200` is user section

https://github.com/esp8266/Arduino/blob/055748ff6fab345511985d684c9e67fb84cddaa5/bootloaders/eboot/eboot_command.h#L15

eboot writes to the user portion here, am i wrong?
e.g. writing using 64 offset and reading directly from that address would return the same `data` value:
```cpp
uint32_t data = 1751720192;
system_rtc_mem_write(64, &data, 4);
data = ((uint32_t*)RTCMEM_ADDR)[0];
Serial.println(data);
```
([nodemcu source used as reference](https://github.com/nodemcu/nodemcu-firmware/blob/master/app/include/rtc/rtcaccess.h), checking how rtcmem used there)

devyte:
@mcspr I misread the RTC_MEM define at a glance. 
So you're right: that means that the first 128 bytes of the RTC user mem can't be used if Updater is in play. Please open an issue with request for documentation, and reference this discussion. In the meantime, I'll update the comments in this PR.

Note: comments should be added to the [Updater](https://arduino-esp8266.readthedocs.io/en/latest/ota_updates/readme.html#updater-class) section and [ESP-specific APIs](https://arduino-esp8266.readthedocs.io/en/latest/libraries.html?highlight=RTC#esp-specific-apis) section.

devyte:
Closing in favor of #5372 .

Issue #1715, opened 2016-03-03
Host side tests


igrr:
This issue is tracking implementation of a set of mocks which enable testing parts of the core on the PC, and writing relevant tests.

Easy testing targets include core classes (String, Stream, Print), heap management functions, and filesystem APIs. 

Testing WiFiClient, WiFiClientSecure, WiFiServer, ESP8266WebServer, ESP8266httpClient is also possible, but requires a bit more work to build Linux port of LwIP and link against it.

To do list:

- [x] initial batch of mocks
- [x] (some) filesystem tests
- [ ] add makefile, run tests on Travis
- [ ] correct license headers in all mock files
- [ ] mock for HardwareSerial
- [ ] more filesystem tests
- [ ] optional replacement of malloc/free/new/delete in test environment — to test against simulated 40k heap. Need to figure out how to support different behaviour of `operator new` and `operator delete` in parts of code under test and in test runner.
- [ ] tests for number formatting functions in core_esp8266_noniso.c
- [ ] tests for String class
- [ ] LwIP port for Linux and infrastructure to link tests against it
- [ ] tests for WiFiClient/WiFiServer
- [ ] axTLS Linux build
- [ ] tests for WiFiClientSecure

devyte:
Progress has been done, but more work remains. Pushing back for next steps.

devyte:
@earlephilhower @d-a-v is there anything further to add here before moving on? If not, I'll push back.

d-a-v:
There'll always be something pending in this feature request.
Let's just not close it and keep regularly pushing it back so we can discuss on its state.

Currenly as @earlephilhower stated a short while ago, we have
- host tests (history)
- emulation on host (quite recent)
- device tests (recent)

Device tests use emulation on host, but host tests uses a different API for a subset of function.
This needs to be unified.

PR #5342, opened 2018-11-15
Sketch emulation on host

d-a-v:
Sketch emulation on host
--------------------

This environment let compile esp8266/Arduino sketches into native
environment.  Network (tcp, udp, including ssl and multicast) is linked to
local host interfaces.  WiFi is trivialy emulated and reported as "just"
already connected and usable.

Currently network emulation is a complete rewrite of
WiFiClient+WiFiServer/ClientContext and WifiUdp/UdpContext using socket
posix API.  Further work will optionally propose native lwIP library
instead.

([readme](https://github.com/d-a-v/Arduino/blob/host/tests/host/README.txt))


d-a-v:
#1715

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -440,7 +440,7 @@ void ESP8266WebServer::sendContent(const String& content) {
   if(_chunked) {
     char * chunkSize = (char *)malloc(11);
```
earlephilhower: Would this be an appropriate spot to reduce some technical debt?  This is a silly `malloc` whose pointer takes an appreciable amount of memory as the actual data itself?

Changing to `char chunkSize[11];` and dropping the check for `chunkSize!=NULL`and the `free(chunkSize);` would seem to me to be trivial enough.

libraries/ESP8266mDNS/ESP8266mDNS.cpp
```diff
@@ -431,11 +431,11 @@ MDNSTxt * MDNSResponder::_getServiceTxt(char *name, char *proto){
   for (servicePtr = _services; servicePtr; servicePtr = servicePtr->_next) {
     if(servicePtr->_port > 0 && strcmp(servicePtr->_name, name) == 0 && strcmp(servicePtr->_proto, proto) == 0){
       if (servicePtr->_txts == 0) 
-        return false;
+        return nullptr;
```
earlephilhower: Good catch, that's a silly return false.

tests/common.sh
```diff
@@ -134,7 +134,7 @@ function build_docs()
 function run_host_tests()
 {
     pushd host
-    make
+    make ARCH32=0 CI
```
earlephilhower: Is there any reason not to only do testing in 32-bit mode?  Seems like it's as close as we're likely to get, and should remove the need for all of the casting dances you had to do with `(uint32_t)(unsigned long)&abcd` throughout the core.

earlephilhower: We can make sure the host CI tests have 32-bit dev libs if needed.  We're already installing valgrind and lcov, if that was your concern.

tests/host/Makefile
```diff
@@ -31,34 +50,64 @@ CORE_C_FILES := $(addprefix $(CORE_PATH)/,\
...
+PREINCLUDES := \
+	-include common/mock.h \
+	-include common/c_types.h \
+
+ifneq ($(D),)
+DEBUG += -DDEBUG_ESP_PORT=Serial
+DEBUG += -DDEBUG_ESP_SSL -DDEBUG_ESP_TLS_MEM -DDEBUG_ESP_HTTP_CLIENT -DDEBUG_ESP_HTTP_SERVER -DDEBUG_ESP_CORE -DDEBUG_ESP_WIFI -DDEBUG_ESP_HTTP_UPDATE -DDEBUG_ESP_UPDATER -DDEBUG_ESP_OTA -DDEBUG_ESP_MDNS
+endif
+
+CXXFLAGS += $(DEBUG) -std=c++11 -Wall -Werror -coverage -O0 -fno-common -g $(M32)
```
earlephilhower: I know the original makefile has `-O0` (oh-zero), but is there some reason for this?  Wouldn't `-Os` be closer to the real system?  Same w/the LDFLAGS.

tests/host/common/ArduinoMain.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "lwip/mem.h"
+#include <include/UdpContext.h>
+#include <poll.h>
+
+#include <unistd.h> // usleep
+#include <getopt.h>
+
+#include <map>
+
+#if 0
```
earlephilhower: Looks like leftover code block

tests/host/common/MockEEPROM.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+void EEPROMClass::write (int x, uint8_t c)
+{
+	if (x > (int)_size)
+		fprintf(stderr, MOCK "### eeprom beyond\r\n");
+	else if (pwrite(_fd, &c, 1, x) != 1)
+		fprintf(stderr, MOCK "eeprom: %s\n\r", strerror(errno));
+}
+
+#if 0
```
earlephilhower: Leftover?

tests/host/common/user_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+}
+
+uint8 wifi_get_listen_interval (void)
+{
+	return 1;
+}
+
+bool wifi_get_macaddr(uint8 if_index, uint8 *macaddr)
+{
+	macaddr[0] = 0xde;
```
earlephilhower: Is this your MAC?  Did you want to put in an obviously fake one?

earlephilhower: `memcpy(macaddr, "ESPMOK");` ?

tools/sdk/ssl/Makefile
```diff
@@ -11,7 +11,7 @@ install: all version-header
 
 bearssl/README.txt:
 	git submodule update --init --recursive bearssl
-	cd bearssl && git remote add bearssl https://www.bearssl.org/git/BearSSL
+	cd bearssl && (git remote add bearssl https://www.bearssl.org/git/BearSSL || true)
```
earlephilhower: Thanks, I never did get around to cleaning that up!

tests/common.sh
d-a-v: This is the reason. CI was using native compiler, it still does. It can be later sorted out.
```
valgrind:  Fatal error at startup: a function redirection
valgrind:  which is mandatory for this platform-tool combination
valgrind:  cannot be set up.  Details of the redirection are:
valgrind:  
valgrind:  A must-be-redirected function
valgrind:  whose name matches the pattern:      strlen
valgrind:  in an object with soname matching:   ld-linux.so.2
valgrind:  was not found whilst processing
valgrind:  symbols from the object with soname: ld-linux.so.2
valgrind:  
valgrind:  Possible fixes: (1, short term): install glibc's debuginfo
valgrind:  package on this machine.  (2, longer term): ask the packagers
valgrind:  for your Linux distribution to please in future ship a non-
valgrind:  stripped ld.so (or whatever the dynamic linker .so is called)
valgrind:  that exports the above-named function using the standard
valgrind:  calling conventions for this platform.  The package you need
valgrind:  to install for fix (1) is called
```

tests/host/Makefile
d-a-v: I'll add options to change that.
> run more of core through host-based testing and code analysis

I guess `-O0` was here for debugging with gdb on host and valgrind detailed report.

tests/host/common/ArduinoMain.cpp
d-a-v: It's WIP code for the ability to store SPIFFS content in a file and retrieve it at next run.
It's not working yet but it's close.

tests/host/common/user_interface.cpp
d-a-v: It's deba7ab1e (at most 0x42 times) ;)

tests/host/common/MockEEPROM.cpp
d-a-v: Yes `#if 0` is a leftover. `.put` is arduino [api](https://www.arduino.cc/en/Reference/EEPROMPut).
There are indeed many things to improve in this PR.

penfold42:
This is cool - top work.

Running on a pi you could even support real gpio, spi, i2c, i2s for much quicker dev cycles (with much less flash wear !)

cores/esp8266/Updater.cpp
```diff
@@ -84,20 +84,20 @@ bool UpdaterClass::begin(size_t size, int command, int ledPin, uint8_t ledOn) {
...
+    uintptr_t updateEndAddress = (uintptr_t)&_SPIFFS_start - 0x40200000;
     //size of the update rounded to a sector
     uint32_t roundedSize = (size + FLASH_SECTOR_SIZE - 1) & (~(FLASH_SECTOR_SIZE - 1));
     //address where we will start writing the update
     updateStartAddress = (updateEndAddress > roundedSize)? (updateEndAddress - roundedSize) : 0;
 
 #ifdef DEBUG_UPDATER
         DEBUG_UPDATER.printf("[begin] roundedSize:       0x%08X (%d)\n", roundedSize, roundedSize);
-        DEBUG_UPDATER.printf("[begin] updateEndAddress:  0x%08X (%d)\n", updateEndAddress, updateEndAddress);
+        DEBUG_UPDATER.printf("[begin] updateEndAddress:  0x%08X (%d)\n", (int)updateEndAddress, (int)updateEndAddress);
```
devyte: 0x%08X => %p?

cores/esp8266/Updater.cpp
```diff
@@ -133,7 +133,7 @@ bool UpdaterClass::begin(size_t size, int command, int ledPin, uint8_t ledOn) {
 #ifdef DEBUG_UPDATER
   DEBUG_UPDATER.printf("[begin] _startAddress:     0x%08X (%d)\n", _startAddress, _startAddress);
   DEBUG_UPDATER.printf("[begin] _currentAddress:   0x%08X (%d)\n", _currentAddress, _currentAddress);
-  DEBUG_UPDATER.printf("[begin] _size:             0x%08X (%d)\n", _size, _size);
+  DEBUG_UPDATER.printf("[begin] _size:             0x%08X (%d)\n", (int)_size, (int)_size);
```
devyte: %d=>%u? 
and is the first cast needed?

cores/esp8266/Updater.cpp
devyte: %d => %u or %ul?

cores/esp8266/Updater.cpp
```diff
@@ -159,7 +159,7 @@ bool UpdaterClass::end(bool evenIfRemaining){
 
   if(hasError() || (!isFinished() && !evenIfRemaining)){
 #ifdef DEBUG_UPDATER
-    DEBUG_UPDATER.printf("premature end: res:%u, pos:%u/%u\n", getError(), progress(), _size);
+    DEBUG_UPDATER.printf("premature end: res:%u, pos:%u/%u\n", getError(), (unsigned)progress(), (unsigned)_size);
```
devyte: the args are all unsigned (size_t). Are these casts necessary? Maybe %u=>%ul ?

cores/esp8266/Updater.cpp
```diff
@@ -199,10 +199,10 @@ bool UpdaterClass::end(bool evenIfRemaining){
     eboot_command_write(&ebcmd);
 
 #ifdef DEBUG_UPDATER
-    DEBUG_UPDATER.printf("Staged: address:0x%08X, size:0x%08X\n", _startAddress, _size);
+    DEBUG_UPDATER.printf("Staged: address:0x%08X, size:0x%08X\n", _startAddress, (unsigned)_size);
```
devyte: instead of casting, add "l" (lower case L) for size_t, which is long, i.e.: 
0x%08X => 0x%08Xl

cores/esp8266/Updater.cpp
```diff
@@ -199,10 +199,10 @@ bool UpdaterClass::end(bool evenIfRemaining){
...
   }
   else if (_command == U_SPIFFS) {
-    DEBUG_UPDATER.printf("SPIFFS: address:0x%08X, size:0x%08X\n", _startAddress, _size);
+    DEBUG_UPDATER.printf("SPIFFS: address:0x%08X, size:0x%08X\n", _startAddress, (unsigned)_size);
```
devyte: add "l" (lower case L) for size_t, which is long, i.e.: 
0x%08X => 0x%08Xl

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -1386,7 +1405,7 @@ extern "C" {
     int freeheap = ESP.getFreeHeap();
     static int laststack, lastheap, laststack2;
     if ((laststack != freestack) || (lastheap != freeheap) || (laststack2 != (int)br_esp8266_stack_proxy_usage())) {
-      Serial.printf("%s:%s(%d): FREESTACK=%d, STACK2USAGE=%d, FREEHEAP=%d\n", file, fcn, line, freestack, br_esp8266_stack_proxy_usage(), freeheap);
+      Serial.printf("%s:%s(%d): FREESTACK=%d, STACK2USAGE=%d, FREEHEAP=%d\n", file, fcn, line, freestack, (int)br_esp8266_stack_proxy_usage(), freeheap);
```
devyte: instead of casting to int, add "l" (lower case L) for size_t, which is long, i.e.: 
0x%08X => 0x%08Xl

libraries/ESP8266WebServer/src/ESP8266WebServer.cpp
```diff
@@ -438,12 +438,9 @@ void ESP8266WebServer::sendContent(const String& content) {
   const char * footer = "\r\n";
   size_t len = content.length();
   if(_chunked) {
-    char * chunkSize = (char *)malloc(11);
-    if(chunkSize){
-      sprintf(chunkSize, "%x%s", len, footer);
-      _currentClientWrite(chunkSize, strlen(chunkSize));
-      free(chunkSize);
-    }
+    char chunkSize[11];
```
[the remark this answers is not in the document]
d-a-v: That was an @earlephilhower 's notice above.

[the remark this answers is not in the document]
earlephilhower: Yup.  4 bytes vs 12 on the stack, I'm not seeing that change things.  If you're that close, one interrupt and you're toast anyway.

libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
```diff
@@ -826,6 +826,8 @@ bool X509List::append(const uint8_t *derCert, size_t derLen) {
   return true;
 }
 
+#if !CORE_MOCK
```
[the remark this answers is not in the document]
d-a-v: I tried. The assembly `make_stack_thunk()` helper macro is not simply and cleanly mockable.

tests/host/common/ArduinoWire.cpp
```diff
@@ -0,0 +1,71 @@
+/*
```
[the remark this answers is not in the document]
d-a-v: For file name consistency, I renamed it to MockWire.cpp

libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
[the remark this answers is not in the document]
devyte: Alright, let's keep it as-is.

tests/host/common/MockUDPSocket.cpp
```diff
@@ -0,0 +1,182 @@
+/*
```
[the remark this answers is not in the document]
devyte: I didn't mean the filename, although that's ok. I was referring to the function names. Instead of having mockSomething() mockOtherthing(), you could do 
```
namespace mock
{
void something();
void otherthing();
}
```
and eventually, you could do the following for equivalent functionality with different implementation on the host:
```
namespace host
{
class ClientContext
{...};
}
```
which could be the counterpart to `namespace esp8266 {class ClientContext {...}; }`.
Then, you can map into the global namespace as needed.

tests/host/common/ArduinoWire.cpp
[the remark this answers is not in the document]
devyte: That's not a good idea. MockWire hints at Wire, which is the I2C implementation. The pinMode() function is in core_esp8266_wiring_digital.c, and analogRead() is in core_esp8266_wiring_analog.c, which is why I suggested hostWiring.cpp. Alternatives: core_host_wiring.cpp, or coreHostWiring.cpp. Or better yet, put pinMode in coreHostWiringDigital.cpp and analogRead() in coreHostWiringAnalog.cpp (or their lower case names) to keep the filenames consistent with the core.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -1377,6 +1381,21 @@ bool WiFiClientSecure::loadPrivateKey(Stream& stream, size_t size) {
...
+  }
+  void _BearSSLCheckStack(const char *fcn, const char *file, int line) {
+    (void)fcn;
+    (void)file;
+    (void)line;
+  }
+
+#else // !CORE_MOCK
+
   extern size_t br_esp8266_stack_proxy_usage();
```
earlephilhower: Ah, poo.  These 2 (BSSLChkStack, BSSLPrint) routines are no longer used with BearSSL as a thunk.  Would you like to save the effort and remove them here, or do you want me to do a separate PR?

earlephilhower:
Per: https://en.wikipedia.org/wiki/Printf_format_string , "%zd"/"%zu" === size_t format specifier, if there are still issues w/printfs.

tests/host/common/MockUDPSocket.cpp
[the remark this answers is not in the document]
earlephilhower: I think @d-a-v is going to run into many *.C files, where `namespace` isn't really going to be usable.  So maybe mockXXX is good enough, and will be consistent?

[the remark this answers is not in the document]
devyte: Ok, agreed.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
d-a-v: Because you are offering, I'd be glad to leave it to you

earlephilhower: Will do after this merges.  You've got enough commits already!

Issue #5206, opened 2018-10-04
Enhancement - Add support for writing/reading multiple I2S samples


earlephilhower:
PR #2501 had some good ideas which were integrated via later reworks of the I2S subsystem.  The one thing that's still missing is a call to the I2S subsystem which can read or write multiple samples at a time.

This is important because most audio codecs can produce many samples at a time, and we can reduce CPU usage significantly by simply writing as many of those samples at a time as we have I2S buffer space.  Instead of calling an add-sample() call 44,100 times per second, this would allow us only to have <1/100th those number of calls.

PR #5349, opened 2018-11-19
i2s can send now buffers 

aguaviva:
This is more efficient than sending samples individually. Actually my https://github.com/aguaviva/ESP8266-phone wont work fast enough without this call I added.


cores/esp8266/core_esp8266_i2s.c
```diff
@@ -331,6 +331,45 @@ bool i2s_write_lr(int16_t left, int16_t right){
   return i2s_write_sample(sample);
 }
 
+#define MIN(a,b) (a<b?a:b)
```
earlephilhower: This is only used the one time later on, can we dispense with the macro?  MIN() as a macro is always dangerous, but obviously std::min() isn't an option for .C code.

cores/esp8266/core_esp8266_i2s.c
```diff
@@ -331,6 +331,45 @@ bool i2s_write_lr(int16_t left, int16_t right){
...
+
+// writes a buffer of samples into the DMA memory, returns the amount of samples written
+int16_t i2s_write_buffer_mono(int16_t *sample, int16_t len) {
```
earlephilhower: The existing code doesn't do any mono, and has blocking and nonblocking modes.  For a mono array I see why you need a special function.  Can we abstract this, though, to be more in-line with the current options?

i.e.:
````
int16_t _i2s_write_buffer(int16_t *sample, int16_t len, bool mono, bool nb)
{
...existing code, changed to handle mono or stereo in the storing...handle blocking to write all if nb==true...
}

int16_t i2s_write_buffer_mono(int16_t *sample, int16_t len) { return _i2s_write_buffer(sample, len, true, true); }

int16_t i2s_write_buffer_mono_nb(int16_t *sample, int16_t len) { return _i2s_write_buffer(sample, len, true, false); }

int16_t i2s_write_buffer(int16_t *sample, int16_t len) { return _i2s_write_buffer(sample, len, false, true); }

int16_t i2s_write_buffer_nb(int16_t *sample, int16_t len) { return _i2s_write_buffer(sample, len, false false); }
````

cores/esp8266/core_esp8266_i2s.c
aguaviva: I can inline that macro in the code, is that fine?

cores/esp8266/core_esp8266_i2s.c
aguaviva: sounds good, just bear with me because I am still learning how resolving code reviews work on Github :)

earlephilhower: No problem.  I opened a bug for 2.6.0 to add just this kind of option, assigned to myself since I redid the I2S and maintain a bunch of audio decoders for the 8266.  So you're fixing a bug I assigned to myself, thanks!  Issue #5206 .

aguaviva:
I made a small optimization. As I said the code is correct, the sniffing of the I2C bus looks good and it looks like the distortion is coming from my I2C HW (pcm5100) who is causing it, I can't even get a proper sine wave (Checked with my oscilloscope). Please lets move this to the next  step :)

earlephilhower:
Looks good, should be very useful.  Thanks!

aguaviva:
BTW this was my first pull request ever so thanks for your patience :) Silly question, who what am I supposed to do now? Close the pull request?

devyte:
Congrats on your first PR! You're done now. This is now available in master branch (aka latest git) and will be released in the next core version.

aguaviva:
wohoo! :D

PR #5355, opened 2018-11-20
document wificlient loop

### LINKED ISSUE #5257: Did WiFiClient.connected recently change for small payloads?
### Basic Infos

#### Platform

- Hardware:				ESP-WROOM-2
- Core Version:			2..4.2
- Operating System:		macOS

### Settings in IDE

- Module:  			ESP-WROOM-2
- Flash Mode:			qio
- Flash Size: 		        4MB
- lwip Variant:                  v2 Lower Memory
- Reset Method:		ck
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:             other

### Problem Description
This is a follow-up to https://github.com/ThingPulse/esp8266-weather-station/issues/140 which in turns follows up on as long thread in our support forum.

It appears that a (nested) HTTP/TCP parse loop that first checks for `client.connected()` and then for `client.available() > 0` **no longer** works reliably for small payloads.

What _seems_ to be happening on the networking stack is that "sometimes" the `WiFiClient` reports `connected() == false` immediately when the payload is sufficiently small. However, even though the client is no longer connected its internal buffer contains all the data and, hence, `client.available()` returns more than 0.

Can you confirm that such a double nested loop, which is a very common sketch template, should these days be changed to only check for `client.available() > 0`?

### Sketch fragment

```cpp
    // WiFiClient * client = http.getStreamPtr();
    // or something like
    // BearSSL::WiFiClientSecure client;
    // ...
    // if (client.connect(host, port))
    while(client.connected()) {
      while((size = client.available()) > 0) {
        // client->read() and do stuff
        // ...
        // give WiFi and TCP/IP libraries a chance to handle pending events
        yield();
      }
    }
  }
```

d-a-v:
fix #5257, rendered
- [there](https://github.com/d-a-v/Arduino/blob/loopdoc/doc/esp8266wifi/client-examples.rst#read-reply-from-the-server)
- and [there](https://github.com/d-a-v/Arduino/blob/loopdoc/doc/esp8266wifi/client-examples.rst#general-client-loop)

### DIFF
--- a/doc/esp8266wifi/client-examples.rst
+++ b/doc/esp8266wifi/client-examples.rst
@@ -87,11 +87,11 @@ If connection is successful, we should send request the host to provide specific
 Read Reply from the Server
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Then, while connection by our client is still alive (``while (client.connected())``, see below) we can read line by line and print out server's response:
+Then, while connection by our client is still alive or while data are available to read (``while (client.connected() || client.available())``, see below) we can read line by line and print out server's response:
 
 .. code:: cpp
 
-    while (client.connected())
+    while (client.connected() || client.available())
     {
       if (client.available())
       {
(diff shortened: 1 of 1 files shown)


marcelstoer:
Very nice, thanks a lot!

marcelstoer:
One useful follow-up would be to code the two examples linked on that page according to that blueprint: https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiClient/WiFiClient.ino, https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiClientBasic/WiFiClientBasic.ino

d-a-v:
@marcelstoer True. Examples need review with this and also ssl.

marcelstoer:
Do you keep this pending item somewhere on your (personal) task list or is there an issue to follow?

Issue #5340, opened 2018-11-14
I2C clock stretching not implemented at all occurrences


maarten-pennings:
When using as I2C slave the [CCS811](http://ams.com/ccs811) gas sensor from ams, e.g. using [this board](https://www.aliexpress.com/item/CJMCU-811-CCS811-Carbon-Monoxide-CO-VOCs-Air-Quality-Numerical-Gas-Sensors/32820864503.html), there are communication errors. These are due to the CCS811 doing _clock stretching_.

The driver file [core_esp8266_si2c.c](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_si2c.c) "Software I2C library for esp8266" implements clock stretching. However, at two places clock stretching code is missing. Functions `twi_writeTo()` and `twi_readFrom()` set the clock line high using `SCL_HIGH()`, without checking whether SCL really goes high. All other places in the driver that use `SCL_HIGH()` do have the clock stretch check.

In both functions, one line should be added ([here](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_si2c.c#L269) and [here](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_si2c.c#L289)):
```
  while(SDA_READ() == 0 && (i++) < 10){
    SCL_LOW();
    twi_delay(twi_dcount);
    SCL_HIGH();

    unsigned int t=0; while(SCL_READ()==0 && (t++)<twi_clockStretchLimit); // Clock stretching

    twi_delay(twi_dcount);
  }
```

See this [screenshot](https://github.com/maarten-pennings/CCS811/blob/master/ESP8266-AddClockStretch.png) for a diff.

devyte:
@maarten-pennings I don't have a way to test this. Of note, there are currently no I2C examples or tests. Is there any chance you can implement something that doesn't depend on specialized hardware? 
Example: two ESPs where one is master and the other is slave, and one sends a hello world to the other. Something like that could then be used as a starting point to test functionality like the issue you describe here.

devyte:
@suculent @mrwgx3 @bjoham You've contributed to our I2C. Could one of you please help out with assessing the changes proposed here, and with my request for a Wire master/slave example with 2 ESPs?

suculent:
I have working implementation of master-slave communication, it’s just a bit complex as it involves simple protocol with CRC check and request-retransfer logic. 

It would all need to be stripped down first. 

[...]

PR #5362, opened 2018-11-21
i2c_clock_stretching_fix

suculent:
Adds time-braces in places where aces caused stasis.

No, really. There are 10 points in the code where stretching should be supported and this adds the twi more missing. Correctly ignores i value to stretch behind usual clock bounds (that's what clock-streching does, effectively stopping I2C communication until slave device does its job).


suculent:
Ugh :o) Nightly failures.

devyte:
Fixes #5340

PR #5366, opened 2018-11-22
deprecate axTLS, update examples

d-a-v:
~None of these changes have been tested yet,
I give them a visibility here,~ opened to discussion.


d-a-v:
This PR also makes BearSSL the default.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -27,8 +27,9 @@
 
 #ifdef HTTPCLIENT_1_1_COMPATIBLE
 #include <ESP8266WiFi.h>
-#include <WiFiClientSecure.h>
+#include <WiFiClientSecureAxTLS.h>
```
devyte: Something doesn't look right in this file with the HTTPCLIENT_1_1_COMPATIBLE define. Did you test both defined and undefined cases?

libraries/ESP8266WiFi/examples/NTPClient/NTPClient.ino
```diff
@@ -21,8 +21,13 @@
...
 #include <WiFiUdp.h>
 
-char ssid[] = "*************";  //  your network SSID (name)
-char pass[] = "********";       // your network password
+#ifndef STASSID
+#define STASSID "your-ssid"
+#define PSK  "your-password"
+#endif
+
+char ssid[] = STASSID; // your network SSID (name)
```
devyte: prior examples were ssid*, but here it0s ssid[]. Same for pass right below.

libraries/ESP8266WiFi/examples/WiFiTelnetToSerial/WiFiTelnetToSerial.ino
```diff
@@ -20,83 +20,190 @@
...
-  //check clients for data
-  for (i = 0; i < MAX_SRV_CLIENTS; i++) {
-    if (serverClients[i] && serverClients[i].connected()) {
-      if (serverClients[i].available()) {
-        //get data from the telnet client and push it to the UART
-        while (serverClients[i].available()) {
-          Serial.write(serverClients[i].read());
+
+  //check TCP clients for data
+#if 1
```
devyte: what's this?

d-a-v: it's temporary, see the comment below
// Incredibly, this code is faster than the bufferred one below - #4620 is needed

d-a-v:
This PR
- deprecates axTLS
- moves former axTLS examples to BearSSL
- adds an axTLS only example
- replaces all SSID/PSK entries with defines STASSID/STAPSK
  those defines can be overriden in your `platform.local.txt` that way:
```
compiler.c.extra_flags=-DSTASSID="mylocalssid" -DSTAPSK="mylocalssidpassword"
compiler.cpp.extra_flags={compiler.c.extra_flags}
```
- changes example `astyle` config: `#preprocessordirectives` are no more indented
- adds some scripts in `tests`/
  - `examples_restyle.sh`
  - `run_CI_locally.sh`
- updates WiFiTelnelToSerial example

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
d-a-v: This patch is now reverted, axTLS is however explicitely mentioned in this file.
I just fixed the deprecation, nothing more.

libraries/ESP8266WiFi/examples/NTPClient/NTPClient.ino
d-a-v: Except from WiFiTelnetToSerial and axTLS examples which have bigger changes,
I only added `APSSID`/`APPSK` defines to all the others without much further updates.
There is no general template on how they use ssid and password.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -27,7 +27,7 @@
...
```
earlephilhower: I'd have done these changes differently, but I guess this is OK too and much less work.  Since axTLS is going away anyway, fine by me.

The "proper" way to do this, I believe, would be to change `WiFiClientSecure.h` to `using BearSSL`, then no changes to headers in any other files.

There would be changes to all object instantiations from `WiFiClientSecure` to `axTLS::WiFiClientSecure` in the examples, and then removal of the `BearSSL::` prefix everywhere.

libraries/ESP8266WiFi/examples/NTPClient/NTPClient.ino
earlephilhower: Yes, we're bad there since dozens of people have submitted these things.  `const` is appropriate here, too.

Don't get me started on `DEBUG`-type macros, they're a complete mess that makes this look like nothing!

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -152,6 +148,21 @@ class WiFiClientSecure : public WiFiClient {
       return loadCACert(file, file.size());
     }
 
+    // AxTLS API deprecated warnings to help upgrading
+
+    bool verify(const char* fingerprint, const char* domain_name)
```
earlephilhower: Thanks

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -121,10 +121,6 @@ class WiFiClientSecure : public WiFiClient {
     static bool probeMaxFragmentLength(const char *hostname, uint16_t port, uint16_t len);
     static bool probeMaxFragmentLength(const String& host, uint16_t port, uint16_t len);
 
-    // AXTLS compatible wrappers
-    // Cannot implement this mode, we need FP before we can connect: bool verify(const char* fingerprint, const char* domain_name)
-    bool verifyCertChain(const char* domain_name) { (void)domain_name; return connected(); } // If we're connected, the cert passed validation during handshake
-
     bool setCACert(const uint8_t* pk, size_t size);
```
earlephilhower: Can you please deprecate the `setCACert`, `setCertificate`, and `setPrivateKey` methods too?  They're axTLS compatibility wrappers and hackish.

tests/common.sh
```diff
@@ -104,7 +104,7 @@ function install_ide()
     local ide_path=$1
     local core_path=$2
     local debug=$3
-    wget -O arduino.tar.xz https://www.arduino.cc/download.php?f=/arduino-nightly-linux64.tar.xz
+    test -r arduino.tar.xz || wget -O arduino.tar.xz https://www.arduino.cc/download.php?f=/arduino-nightly-linux64.tar.xz
```
earlephilhower: Good catch and speedup.

tests/examples_restyle.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
earlephilhower: THANK YOU!

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -121,48 +121,69 @@ class WiFiClientSecure : public WiFiClient {
...
     static bool probeMaxFragmentLength(const String& host, uint16_t port, uint16_t len);
 
-    bool setCACert(const uint8_t* pk, size_t size);
-    bool setCertificate(const uint8_t* pk, size_t size);
-    bool setPrivateKey(const uint8_t* pk, size_t size);
+    ////////////////////////////////////////////////////
+    // AxTLS API deprecated warnings to help upgrading
+
+#define AXTLS_DEPRECATED
+    #define xAXTLS_DEPRECATED \
```
earlephilhower: Looks like something's mixed up here.  Shouldn't everything be in the AXTLS_DEPRECATED macro and not xAXTLS...?

d-a-v: fixed :)

libraries/ESP8266WiFi/examples/NTPClient/NTPClient.ino
```diff
@@ -21,8 +21,13 @@
...
 #include <WiFiUdp.h>
 
-char ssid[] = "*************";  //  your network SSID (name)
-char pass[] = "********";       // your network password
+#ifndef STASSID
+#define STASSID "your-ssid"
+#define STAPSK  "your-password"
+#endif
+
+const char ssid[] = STASSID; // your network SSID (name)
```
devyte: const char[] => const char * ?

libraries/esp8266/examples/interactive/interactive.ino
```diff
@@ -10,8 +10,13 @@
...
 #include "user_interface.h"
 
-const char SSID[] = "open";
-const char PSK[] = "";
+#ifndef STASSID
+#define STASSID "your-ssid"
+#define STAPSK  "your-password"
+#endif
+
+const char SSID[] = STASSID;
```
devyte: const char[] => const char * ?

PR #5371, opened 2018-11-24
Add timeout to waitForConnectResult

### LINKED ISSUE #4309: WiFi.waitForConnectResult() takes up to 1 hour to return
Hardware: ESP12E
Core Version: Latest master

Hi,

I noticed since the 2.4.0 (even the latest) I get random network disconnection. Not that hard to handle. But in few case, I get WiFi.waitForConnectResult() taking up to 1 hour to return the result : 

Code : 
```
Serial.printf(F("Connecting to %s with key %s...\n"), conf.SSID, conf.WKEY);
WiFi.begin(conf.SSID, conf.WKEY);
delay(50);

if (WiFi.waitForConnectResult() == WL_CONNECTED)
{
	Serial.println(F("WiFi connected"));
	return true;
}
```

Serial monitor : 
```
[06/02/18 - 01:34:15:776] Connecting to *** with key ***...

[06/02/18 - 02:20:21:509] WiFi connected
```
```
[06/02/18 - 06:26:26:208] Connecting to *** with key ***...

[06/02/18 - 06:45:57:847] WiFi connected
```
It could takes 1 minute or 5 minutes or in that case 45 minutes to return status code.
I have others ESP8266 with same code that are able to connect and send data during those disconnection.
This problem doesn't appear at the beggining. In that case, the problem appears after 10 hours of running with no problems.

Thank you

dav1901:
Aims to fix #4309


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -376,14 +376,19 @@ bool ESP8266WiFiSTAClass::getAutoReconnect() {
...
- * @return wl_status_t
+ * @return wl_status_t or -1 on timeout
  */
-uint8_t ESP8266WiFiSTAClass::waitForConnectResult() {
+int8_t ESP8266WiFiSTAClass::waitForConnectResult() {
     //1 and 3 have STA enabled
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
+    unsigned long timeout = 60*1000; // number of milliseconds to wait before returning timeout error
```
devyte: Please reimplement this in terms of polledTimeout, e.g.:
```
using esp8266::polledTimeout::oneShot;
oneShot timeout(60*1000);
while(!timeout)
{
  if(status() != WL_DISCONNECTED)
    return status();
}
return -1;
```

dav1901: done

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -381,17 +382,21 @@ bool ESP8266WiFiSTAClass::getAutoReconnect() {
...
     //1 and 3 have STA enabled
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    while(status() == WL_DISCONNECTED) {
-        delay(100);
+    using esp8266::polledTimeout::oneShot;
+    oneShot timeout(60*1000); // number of milliseconds to wait before returning timeout error
+    while(!timeout) {
+        if(status() != WL_DISCONNECTED) {
```
devyte: I forgot a yield() in my example :laughing: I suggest as first line inside the while.

dav1901:
Actually, I think it might be better to allow passing timeout length as an optional parameter, with default being 60 seconds. Any thoughts @devyte?

dav1901:
default is 60000ms.

Issue #4253, opened 2018-01-29
WifiMulti addAP (check if exists)


Hiddenvision:
Not my finest example but I found constantly adding Ap's could crash the party.

I added in a dirty check for existing matching entries, 
there should be another test for password matching but I am super lazy.

This works in both 2.3 and 2.4

I see that the CleanApList is also being made public.
Just seems wrong to clean a list, simply to create it again.

```
// ESP8266WiFiMulti.h
public:
bool existsAP(const char* ssid, const char *passphrase = NULL);

private:
bool APlistExists(const char* ssid, const char *passphrase = NULL);

// ESP8266WiFiMulti.cpp
// added a call to new function APlistExists ()
// just above above the line that follows in function ESP8266WiFiMulti::APlistAdd
if (APlistExists(ssid,passphrase)){return true;}    
newAP.ssid = strdup(ssid);


// This was the public function.
// I normally put this just above the public addAP function near the top.
bool ESP8266WiFiMulti::existsAP(const char* ssid, const char *passphrase) {    return APlistExists(ssid, passphrase);}

// This was the private function.
// I normally put this at the bottom
bool ESP8266WiFiMulti::APlistExists(const char* ssid, const char *passphrase) {
    for(uint32_t i = 0; i < APlist.size(); i++) {
        WifiAPlist_t entry = APlist[i];
        if(entry.ssid && entry.ssid == ssid){return true;}
    }
 return false;
}
```

Hope it comes out code friendly.

devyte:
Closing via #5373 .

PR #5373, opened 2018-11-24
Check if AP exists before adding it

dav1901:
Check if AP exists before adding it with WiFiMulti.addAP as suggested in #4253.


libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
```diff
@@ -38,6 +38,10 @@ bool ESP8266WiFiMulti::addAP(const char* ssid, const char *passphrase) {
     return APlistAdd(ssid, passphrase);
 }
 
+bool ESP8266WiFiMulti::existsAP(const char* ssid, const char *passphrase) {
+    return APlistExists(ssid, passphrase);
+}
+
```
d-a-v: What about only one `public: bool existsAP()` ?
(without the other private one)

dav1901: I tried to make it similar to adding an AP, `public: bool addAP()` , `private: bool APlistAdd()`
Do you want me to change it or leave it as is?

d-a-v: I think @devyte will enlighten us on that very subject

devyte: @d-a-v is right: this is used to map different overloads to a single implementation in a private method, but in this case there is only one public method without overloads...
Sigh, IMHO leave it as-is, other public methods could be implemented with e.g.: String args.

PR #5389, opened 2018-11-29
update HTTPCLIENT_1_1_COMPATIBLE

d-a-v:
use #if instead of #ifdef, set to 1 by default
(set to 0 in host emulation where axTLS is unavailable)


tests/host/README.txt
```diff
@@ -41,19 +41,19 @@ Optional 'FORCE32=0' will use native/default gcc (default is FORCE32=1 unless gc
 
 
 Non exhaustive list of working examples:
-	make D=1 ../../libraries/ESP8266WiFi/examples/udp/udp
-	make D=1 ../../libraries/ESP8266WiFi/examples/WiFiClient/WiFiClient
-	make D=1 ../../libraries/ESP8266WebServer/examples/HelloServer/HelloServer
-	make D=1 ../../libraries/ESP8266WebServer/examples/AdvancedWebServer/AdvancedWebServer
-	make D=1 ../../libraries/ESP8266mDNS/examples/mDNS_Web_Server/mDNS_Web_Server
-	make D=1 ../../libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation
+	make -j D=1 ../../libraries/ESP8266WiFi/examples/udp/udp
```
earlephilhower: I remember saying `-j` would auto-scale the jobs, but it turns out I was wrong.  `-j` will run as many jobs in parallel as `make` can figure out.  It's not really an issue for a single compile like this as most folks will have at least 2-4 cores if they're doing this.
````
-j [jobs], --jobs[=jobs]
     Specifies the number of jobs (commands) to run simultaneously.  If there is more
     than one -j option, the last one is effective.  If the -j option is given  with‐
     out  an argument, make will not limit the number of jobs that can run simultane‐
     ously. When make invokes a sub-make, all instances of make  will  coordinate  to
     run  the  specified  number of jobs at a time; see the section PARALLEL MAKE AND
     THE JOBSERVER for details.
````

d-a-v: > I remember saying -j

I took your wise word :)

PR #5393, opened 2018-11-29
Try caching core.a for CI builds

earlephilhower:


earlephilhower:
Didn't seem to affect runtime, it may already be doing this in arduino-builder.

PR #5395, opened 2018-11-30
weak hook preinit() #2111 #2133 #2136

d-a-v:
Called in `init_done()` callback after global constructors initialized.
Or should it be called in `user_init()` (before global constructors) ?
#2111 #2133 #2136


cores/esp8266/core_esp8266_main.cpp
```diff
@@ -145,6 +145,7 @@ void init_done() {
     system_set_os_print(1);
     gdb_init();
     do_global_ctors();
+    early_setup(); // <-- here after global constructors, or in user_init() ?
```
earlephilhower: Bottom of user_init(), not here.  This is called after WiFi power probing/etc. is set up.

d-a-v: fixed

cores/esp8266/AddrList.h
```diff
@@ -24,7 +24,7 @@
   for (auto a: ifList)
     out.printf("IF='%s' index=%d legacy=%d IPv4=%d local=%d hostname='%s' addr= %s\n",
                a->iface().c_str(),
-               a->number(),
+               a->ifnumber(),
```
earlephilhower: Were these changes intended in the PR?

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -243,18 +243,26 @@ extern "C" void ICACHE_RAM_ATTR app_entry (void)
...
     uart_div_modify(0, UART_CLK_FREQ / (115200));
 
-    init();
+    init(); // in core_esp8266_wiring.c
 
     initVariant();
 
     cont_init(g_pcont);
 
+    preinit(); // user redefinable
```
devyte: To reduce confusion, please add a comment that this function refers to Before C++ Dynamic Init phase, and is not related to the init() above, which is about initializing hw registers and sdk timers.
Refs: 
https://en.cppreference.com/w/cpp/language/initialization
https://stackoverflow.com/questions/5945897/what-is-dynamic-initialization-of-object-in-c

PR #5400, opened 2018-11-30
Add doc for ESP8266WebServer

dav1901:
Basic documentation about ESP8266WebServer library.


libraries/ESP8266WebServer/keywords.txt
```diff
@@ -15,18 +15,27 @@ HTTPMethod	KEYWORD1
 #######################################
 
 begin	KEYWORD2
+close KEYWORD2
```
earlephilhower: All of these keywords.txt need top be done using TAB chars, not spaces.  The editors are kind of pedantic about the formatting, so I don't believe these will work.  See the esp8266/keywords.txt/etc. for more examples.

PR #5404, opened 2018-11-30
added 2M generic boards

apicquot:
2M (SPIFFS 128K)
2M (SPIFFS 256K)

used python board.txt.py --allgen as requested and added the proper flash mapping in board.txt.py


apicquot:
many files generated because names for ld files have changed 1M, 2M becoming 1.0M 2.0M...

devyte:
I see .orig files here.

apicquot:
sorry for the mistakes. should be good now

PR #5410, opened 2018-12-02
Split AddrList into object-iterator-container components

devyte:
Cleanup of the AddrList implementation


cores/esp8266/AddrList.h
```diff
@@ -85,84 +85,220 @@
...
 
-            protected:
 
-                bool equal (const const_iterator& o) {
-                    return    _netif == o._netif
-                           && (!_netif || _num == o._num);
-                }
 
-                const ip_addr_t* _ip_from_netif_num () const {
+#if 0
```
d-a-v: Maybe this should be removed

PR #5411, opened 2018-12-02
Implement for ssid a similar approach as for passphrase

### LINKED ISSUE #2200: ESP8266WiFiScan does not terminate a 32 character long SSID
ESP8266WiFiScan's `ssid(uint8_t i)` function assumes `bss_info`'s ssid member is null terminated with the following statement: 

` return String(reinterpret_cast<const char*>(it->ssid));`

Now, this works fine for any SSID less than 32 characters, as bss_info will have at least one null byte at the end. In the event that an SSID is exactly 32 characters, `bss_info->ssid` will not be null terminated, and the above line will include garbage in the String as it tries to find a null byte beyond the member. 

A memcpy of 32 bytes like below guarantees at most 32 characters are read and correctly terminated, and seems to work fine during my testing.

```
char ssidBuffer[33] = {0};
memcpy(&ssidBufer[0], it->ssid, 32); 
return String(ssidBuffer);
```

devyte:
Fixes #2200 (in theory).
This is not tested. The PR doesn't affect the pre-PR case (ssid up to 31chars). If the implementation doesn't work, and assuming no bug in the approach (same code as for passphrase with 64 bytes), then there is something broken in the SDK, and likely nothing can be done here.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -108,7 +108,7 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
         return WL_CONNECT_FAILED;
     }
 
-    if(!ssid || *ssid == 0x00 || strlen(ssid) > 31) {
+    if(!ssid || *ssid == 0x00 || strlen(ssid) > 32) {
         // fail SSID too long or missing!
         return WL_CONNECT_FAILED;
     }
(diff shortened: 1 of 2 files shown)


devyte:
CC @mictlan666 @C0rn3j @liquidfalcon

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -119,10 +119,12 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
     }
 
     struct station_config conf;
-    strcpy(reinterpret_cast<char*>(conf.ssid), ssid);
+    if(strlen(ssid) == 32)
```
earlephilhower: `strncpy(reinterpret_cast<char*>(conf.ssid), ssid, 32);` will do the same w/o any if required.

From the man page, this is one of the intended uses:
>        One  valid  (and  intended) use of strncpy() is to copy a C string to a fixed-length buffer while ensuring both that the buffer is not overflowed and that unused bytes in the target buffer are zeroed out (perhaps to prevent information leaks if the buffer is to be written to media or transmitted to another process via an interprocess communication technique).

PR #5412, opened 2018-12-02
inconsistent block size for spiffs in board.txt. and ld files

apicquot:
boardtxt.py is not consistent for generating spiffs block size in boards.txt and ld files:

it takes the flash size for boards.txt but the spiffs size for the lf files: there is currently a problem for boards with  flash size >= 1M and spiffs size < 1M

impacted boards 2M (SPIFFS 128K), 2M (SPIFFS 256K) 2M (SPIFFS 512K)

Suggested change is to make block size dependable on spiffs size for board.txt:
if spiffs size < 1M  block size=4096  else block size=8192


tools/boards.txt.py
```diff
@@ -1129,10 +1129,13 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
     if flashsize_kb <= 1024:
         max_upload_size = (flashsize_kb - (spiffs_kb + eeprom_size_kb + rfcal_size_kb + sdkwifi_size_kb)) * 1024 - reserved
         spiffs_start = spiffs_end - spiffs_kb * 1024
-        spiffs_blocksize = 4096
     else:
         max_upload_size = 1024 * 1024 - reserved
         spiffs_start = (flashsize_kb - spiffs_kb) * 1024
+
+    if spiffs_kb < 1024:
```
devyte: Should this be < or <=?

apicquot:
it should be < 1024: for example 2M (SPIFFS 1M) should sill have a block size of 8192.

devyte:
@apicquot what about creating a SPIFFS image from the IDE for flashing? what decision is made for the block size there?

apicquot:
@devyte the block size when flashing the spiffs from the ide  (parameters passed to mkspiffs) is coming from board.txt. That is why it has to be the same value than in the ld files

d-a-v:
While understandable, this is to me dangerous as-is, because users already using existing from some time 2M512 will have their spiffs reformated on boot after this patch.

You are basically stating that spiffs block size **should** be 4096 when spiffs size < 1MB.
Why not choosing `if spiffs_kb < 512:` instead of `< 1024` ?

We can see below that it is 4096/0x1000 when size is `< 512` and 0x2000 otherwise.
I reckon you patch *is very useful and needed* to make spiffs-blocksize calculated from a real understandable rule, but 1024 scares me.

```
tools/sdk/ld$ grep SPIFFS_block *.ld
eagle.flash.16m14m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.16m15m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.1m.ld:PROVIDE ( _SPIFFS_block = 0x0 );
eagle.flash.1m128.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.1m144.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.1m160.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.1m192.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.1m256.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.1m512.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.1m64.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.2m.ld:PROVIDE ( _SPIFFS_block = 0x0 );
eagle.flash.2m128.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.2m1m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.2m256.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.2m512.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.4m.ld:PROVIDE ( _SPIFFS_block = 0x0 );
eagle.flash.4m1m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.4m2m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.4m3m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.512k.ld:PROVIDE ( _SPIFFS_block = 0x0 );
eagle.flash.512k128.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.512k32.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.512k64.ld:PROVIDE ( _SPIFFS_block = 0x1000 );
eagle.flash.8m6m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
eagle.flash.8m7m.ld:PROVIDE ( _SPIFFS_block = 0x2000 );
```

apicquot:
good point @devyte 

changing to  if spiffs_kb < 512 is a good idea

earlephilhower:
#5429 will take care of the indent.  `platform....8266.json` is being edited using string replacement not a JSON parser presently so there will be odd indent differences.

The SPIFFS blocksize change hasn't been tested by me, can't comment on it.

tools/boards.txt.py
```diff
@@ -1153,7 +1153,10 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
     else:
         max_upload_size = 1024 * 1024 - reserved
         spiffs_start = (flashsize_kb - spiffs_kb) * 1024
-        spiffs_blocksize = 8192
+        if spiffs_kb < 512:
```
earlephilhower: The problem introduced here is you have made existing small SPIFFS filesystems unreadable (and probably will cause hangs/crashes when they're attempted to be mounted).

earlephilhower: I went thru the logic and on review it seems OK.

This only applies to existing @M or greater SPIFFS chips whose FS was <512KB

4M and above do not support 512K SPIFFS.

You only just added <512KB to these 2-3 commits ago, so it's never been released.

Therefore, this only changes 2M(<512K) spiffs, and it's your addition less than a week ago.  I say it's OK by me!

TD-er:
> While understandable, this is to me dangerous as-is, because users already using existing from some time 2M512 will have their spiffs reformated on boot after this patch.
> 
> You are basically stating that spiffs block size **should** be 4096 when spiffs size < 1MB.
> Why not choosing `if spiffs_kb < 512:` instead of `< 1024` ?
> 
> We can see below that it is 4096/0x1000 when size is `< 512` and 0x2000 otherwise.
> I reckon your patch _is very useful and needed_ to make spiffs-blocksize calculated from a real understandable rule, but 1024 scares me.

The tests I did with the 2M/512k layout simply did not work.
So I think it is very unlikely there are working setups running this layout with working SPIFFS.

Also a practical question about the block size.
Does any file take at least 1 block? (like the name suggests)
If so, then I would even opt for smaller blocks if possible.
On 128k SPIFFS there is only room for max. 32 (very small) files (more likely less due to filesystem overhead) with minimum filesize of 4k.

apicquot:
2M 512K works great for me and so do the 2 new board configs 2M 128K and 2M 256K. I use the IDE for launching mkspiffs and setting up properly the parameters. As of the maximum number of files, my understanding is that they can more files than number of blocks, the minimum file size is the page size.  I have 47 small files that works fine with the 128K so going below 4096 block size won't do any good. Hope it helps

devyte:
@apicquot thanks for the info!

PR #5414, opened 2018-12-02
fix device test test_http_server

liebman:
use OrderedDict to force order of get params, they were hash order + random so test would fail a lot.
The body of the request handler for each test was almost identical, refactored to share


tests/device/test_http_server/test_http_server.ino
```diff
@@ -26,21 +27,27 @@ void setup()
...
-            siteData = "";
-            for (uint8_t i=0; i<server.args(); i++){
-                if(i > 0)
-                    siteData += "\n";
-                siteData += server.argName(i) + " = " + server.arg(i);
-            }
-            siteHits++;
-            server.send(200, "text/plain", siteData);
-        });
+        siteData = "";
```
devyte: I suspect this code will work anyways, but to be equivalent to the original code, this line should go in the handle_request() function.

liebman: That would work for all but the last test, "HTTP Upload", where the upload code, which sets ```siteData``` is called before ```handle_request```.

liebman: @devyte To clarify - setting ```siteData``` in ```handle_request``` would break the "HTTP Upload" test.  I found this empirically as my initial implementation was as you had suggested and that's where I found that the upload callback was called before the request callback.

devyte: @liebman thought I had answered this sorry.
You're right, I didn't realize the last test was different.

PR #5418, opened 2018-12-03
Add ESPectro Core board support.

alwint3r:
We noticed that the old PR for this ESPectro Core is closed. So we re-create the board definition in the boards.txt.py like everybody does.


variants/espectro/pins_arduino.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef ESPECTRO_CORE_VERSION
+#define ESPECTRO_CORE_VERSION   3
+#endif
+
+static const uint8_t SDA = 4;
+static const uint8_t SCL = 5;
+
+static const uint8_t LED_BUILTIN = 15;
+static const uint8_t BUILTIN_LED = 15;
```
d-a-v: LED_BUILTIN is preferably a `#define`
BUILTIN_LED is automatically defined in `"common.h"`

alwint3r:
I know how to resolve a commit but I think I'm gonna play it safe by taking out the boards.txt.py and variants and apply it to a new core.

alwint3r:
Will open a new PR. Thanks

PR #5434, opened 2018-12-04
Make exceptions a configurable menu

earlephilhower:
Add a menu, Exceptions, which allows exceptions to be disabled for ROM
sensitive scripts.  Default is enabled.


andrethomas:
Excuse my ignorance, but why is this enabled by default? :)

devyte:
@andrethomas because Exceptions are a standard C++ language feature, because they're needed to gracefully handle error conditions in class constructors, because we want to replace some custom code with STL standard calls, some of which rely on Exceptions, and because we want to bring in a subset of boost, some of which relies on Exceptions.
Also, migration to a newer C++ standard is on the horizon, and Exceptions are a part of that as well.

andrethomas:
@devyte Thanks for the clarity, makes sense now! The reason I was asking is that the current codebase fails for me which I guess is because libstdc++-nox.a does not exist yet. I'm weighing up the ram utilization against 2.4.2.

andrethomas:
Even if I manually edit the boards.txt to disable it irrespective of the selection it still seems to result in an exception enabled binary

```
generic.menu.exception.enabled=Enabled
generic.menu.exception.enabled.build.exception_flags=-fno-exceptions
generic.menu.exception.enabled.build.stdcpp_lib=-lstdc++-nox
generic.menu.exception.disabled=Disabled
generic.menu.exception.disabled.build.exception_flags=-fno-exceptions
generic.menu.exception.disabled.build.stdcpp_lib=-lstdc++-nox
```

andrethomas:
Hmm, sorry false alarm! My json package build script did not update platform.txt :|

earlephilhower:
Thanks for the update.  On closed bugs/merges sometimes comments get overlooked, so sorry for not replying before you figured it out yourself!

andrethomas:
@earlephilhower No sweat - I still think I'm missing something there but I'll figure it out. Great work... can only dream to have as much time and dedication to a single project as you guys have!

Here's hoping for a bug-free release!

PR #5445, opened 2018-12-07
Improved readString() for File

apicquot:
readString() method for the class File is inherited from the class Stream that requires a timeout of 1 second at the end of the file, hence a very poor performance. Stream implementation also uses multiple memory resizing that could be anticipated based on the size of the file.

The proposed implementation solves these two issues.


apicquot:
may want to use a define rather than sizeof for the size of the block to read ?

cores/esp8266/FS.cpp
```diff
@@ -121,6 +121,22 @@ const char* File::name() const {
     return _p->name();
 }
 
+String File::readString()
+{
+    String ret;
+    ret.reserve(size() - position());
+    char temp[256];
+    int countRead = readBytes(temp, sizeof(temp));
+    while (countRead > 0)
```
devyte: You're forgetting to read the last chunk when it has less than sizeof(temp). Eg.: consider a file with size 257bytes, this code will only read the first 256, and the last one will be dropped.
You have to calculate the delta remaining.

devyte:
The function sizeof is compile-time  code, so it gets optimized to a single value. Your use is correct. The number 256 shoukd be in a macro, though.

devyte:
ping!

apicquot:
I don't understand, the code works for me for files<256 or files>= 256: I read and push blocks while countRead>0.

The test countRead<sizeof(temp) is just to place a string terminator for the last block.

devyte:
Apologies, my previous review comment was messed up. It's corrected now.

apicquot:
You are absolutely right @devyte : String::concat(char*) needs a string terminator to work. I fixed it.

string::concat(char*, int length) may have been more elegant but it is a protected method

cores/esp8266/FS.cpp
apicquot: I fixed it, thank you @devyte

Issue #4283, opened 2018-02-03
[2.4.0] Communication issue between ArduinoOTA and espota.py


eXtatic-GitHub:
### Basic Infos
I'm not a python developer and also new to all this here so please have mercy.
While migrating my project from 2.3.0 to 2.4.0 I experienced the OTA to behave different.
espota.py is not closed like it was in 2.3.0

#### Hardware
Hardware:			ESP-07
Core Version:      	2.4.0

### Description

**ArduinoOTA**
*espota.py* is waiting for an "OK" from *ArduinoOTA library* after upload was successful but only receives "K", thus not closing connection.
The "O" seems to get lost somewhere.

With 2.3.0 version of *espota.py* this was not an issue since *espota.py* closed anyway.
#4053 introduced this new handling

**Hack**
By trial and error I just added a leading character to the char array ArduinoOTA returns and it was working:
![image](https://user-images.githubusercontent.com/5675570/35767289-e1674f40-08e9-11e8-8153-deb2db0f3638.png)

**Fix?**
Next try was then to provide a String instead of char*
![image](https://user-images.githubusercontent.com/5675570/35767354-4282f364-08eb-11e8-9917-3a6994c57330.png)

### Settings in IDE

Module:  			Generic ESP8266 Module
Flash Size: 		1 MB
CPU Frequency:		80Mhz
Flash Mode:			qio
Flash Frequency:	40Mhz
Upload Using:		OTA
Reset Method:		ck

### Sketch

basicOTA.ino

### Debug
*espota.py 2.3.0* - prints error "K" but flashed successful
![image](https://user-images.githubusercontent.com/5675570/35767298-162e1768-08ea-11e8-96c6-b0d393f46127.png)

*espota 2.4.0* - hangs there forever but sketch is updated and ESP restarts fine (I canclled execution some minutes after)
![image](https://user-images.githubusercontent.com/5675570/35767341-f9e907ba-08ea-11e8-8aef-9360b91d4648.png)

*espota 2.4.0 + ArduioOTA hack* - nice and smooth
![image](https://user-images.githubusercontent.com/5675570/35767361-531ff352-08eb-11e8-86bc-67097f09abe7.png)

*espota 2.4.0 + ArduioOTA fix?* - nice and smooth
![image](https://user-images.githubusercontent.com/5675570/35767367-6a95a45a-08eb-11e8-8d43-f870234c4a05.png)

devyte:
[...]

PR #5454, opened 2018-12-08
fix OTA

d-a-v:
https://github.com/esp8266/Arduino/issues/4283#issuecomment-445447030

@washcroft
- OTA protocol is old, it may not be changed or things would break on the user side
- OTA received multiple fixes
- You are right about the latest fix, it was too simple and didn't take into account this `res=connection.recv(4)` which is, apart from swallowing data without using them, useless
- OTA code is quite hairy to read, you are not a python dev but I surely could have proposed a fix

> [...] all the recent issues with OTA seems to be caused by the changes @d-a-v made to the espota.py

Well... Yes. I didn't change it for pleasure :) I was fixing some other issue.

> [...] but I can understand the logic and I am merely pointing out why so many people are having issues with OTA uploads. Take it or leave it.

I can't imagine finding this bug was easy, so a way to fix it might also have popped out from your research, that was @devyte's question.

### DIFF
--- a/tools/espota.py
+++ b/tools/espota.py
@@ -144,6 +144,8 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
     sock.close()
     return 1
 
+  received_ok = False
+
   try:
     f = open(filename, "rb")
     if (PROGRESS):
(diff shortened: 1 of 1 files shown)


devyte:
@d-a-v <--- what he said :smile:

PR #5457, opened 2018-12-08
Add fnmatch, regexp to newlib library

earlephilhower:


devyte:
@earlephilhower tested ok both in build and functionally.

PR #5458, opened 2018-12-09
Convert to standard PSTR/PROGMEM macros

earlephilhower:
The bug referenced in the code that required a custom macro to enable
PSTR/PROGMEM compilation ( https://github.com/esp8266/Arduino/issues/3369 )
was fixed a while back and all PROGMEM references now live in their own
save .sections.

This patch simply is the output of `sed -i s/PSTR_LEA/PSTR/g *` and
`sed -i s/PROGMEM_LEA/PROGMEM/g *` on the codebase and the removal
of those defines.


earlephilhower:
As discussed in gitter, the custom PSTR/PROGMEM macros aren't required anymore.

devyte:
Thanks @earlephilhower !

PR #5482, opened 2018-12-12
Update release instructions README.md

devyte:
Fix json url typo
Clarify some points
Add readthedocs instructions

### DIFF
--- a/package/README.md
+++ b/package/README.md
@@ -57,35 +57,42 @@ Here is an overview of the release process. See the section below for detailed i
 
    * When done, put release notes into a private Gist and send the link to other maintainers for review.
 
-   * Update `version` to the release in platform.txt and commit. E.g. `2.5.0`.
+The following points assume work in a direct clone of the repository, and not in a personal fork.
 
-2. Tag the latest commit on the master branch. In this project, tags have form `X.Y.Z`, e.g. `2.4.0`, or `X.Y.Z-rcN` for release versions. Notice that there's no `v`at the beginning of the tag. Tags must be annotated, not lightweight tags. To create a tag, use git command (assuming that the master branch is checked out):
+2. Update `version` to the release in platform.txt and commit. E.g. `2.5.0`. Then git push origin to get the commit merged to master.
 
-   ```
-   git tag -a -m "Release 2.4.0" 2.4.0
-   ```
-
-3. Push the tag created in step 2 to esp8266/Arduino Github repository:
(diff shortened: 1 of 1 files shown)


package/README.md
```diff
@@ -57,35 +57,42 @@ Here is an overview of the release process. See the section below for detailed i
 
    * When done, put release notes into a private Gist and send the link to other maintainers for review.
 
-   * Update `version` to the release in platform.txt and commit. E.g. `2.5.0`.
+The following points assume work in a direct clone of the repository, and not in a personal fork.
 
-2. Tag the latest commit on the master branch. In this project, tags have form `X.Y.Z`, e.g. `2.4.0`, or `X.Y.Z-rcN` for release versions. Notice that there's no `v`at the beginning of the tag. Tags must be annotated, not lightweight tags. To create a tag, use git command (assuming that the master branch is checked out):
+2. Update `version` to the release in platform.txt and commit. E.g. `2.5.0`. Then git push origin to get the commit merged to master.
+
+3. Tag the latest commit on the master branch. In this project, tags have form `X.Y.Z`, e.g. `2.4.0`, or `X.Y.Z-rcN` for release versions. Notice that there's no `v`at the beginning of the tag. Tags must be annotated, not lightweight tags. To create a tag, use git command (assuming that the master branch is checked out):
```
igrr: @devyte If you want to keep using `-beta` instead of `rc`, you can update this paragraph. If you, can you please fix "for release versions" to "for release candidate versions"?

package/README.md
```diff
@@ -57,35 +57,42 @@ Here is an overview of the release process. See the section below for detailed i
...
 
    * In main README.md:
 
-     - in "Contributing" section, update the "easy issues" link to point to the next milestone.
-
      - in "Latest release" section, change version number in the readthedocs link to the version which was just released.
 
+     - Replace any other instances of the version number in the doc.
+     
+     - Check that links to the latest readthedocs build are correct.
```
igrr: Can be combined with the point above, e.g.

- in "Latest release" section, change version number in the readthedocs link to the version which was just released, and verify that the link works

package/README.md
devyte: @igrr we decided to go with -betaX because -rcX was being confused as post-release instead of pre release (e.g. we would receive comments like "I tried 2.4.0 but it crashed, so I updated to 2.4.0-rc2, but then it didn't compile").

What do you think of not doing any more -rcX or -betaX, and just doing normal simple releases? 
- Breaking changes in majors (usually)
- Other changes in minors (avoid breaking changes, except when feature is too broken)
- Majors and minors could be a bit unstable
- Sub releases with stabilization, and critical changes/bug fixes
- Document the linear release model as well as the above guidelines
?

PR #5484, opened 2018-12-12
close files

a-c-sreedhar-reddy:


a-c-sreedhar-reddy:
what does no-ops mean

igrr:
Agree with the comment from @earlephilhower, are files closing automatically, this is done precisely for the reason to avoid having explicit close() calls in the code.

devyte:
The calls aren't needed because the File object destructor chain eventually closes the underlying file.
As a side note, I think that is currently valid for SPIFFS File objects, but not SD File objects.
Closing this as not needed.

PR #5485, opened 2018-12-12
Add RTC time read API

IMAN4K:
Signed-off-by: Iman iman72411@gmail.com (Github: @IMAN4K)


d-a-v:
I was wondering where this weird printf() was coming as I remember having seeing it before.
It's in the SDK documentation (A.2. RTC APIs Example) !

```

uint32 rtc_t1,rtc_t2;
uint32 st1,st2;
uint32 cal1, cal2;
uint32 rtc_time_base = system_get_rtc_time();
uint64 rtc_time_acc = 0;
rtc_t1 = system_get_rtc_time();
st1 = system_get_time();
cal1 = system_rtc_clock_cali_proc();

os_delay_us(300);

st2 = system_get_time();
rtc_t2 = system_get_rtc_time();
cal2 = system_rtc_clock_cali_proc();
os_printf(" rtc_t2-t1 : %d \r\n",rtc_t2-rtc_t1);
os_printf(" st2-t2 : %d \r\n",st2-st1);
os_printf("cal 1 : %d.%d \r\n", ((cal1*1000)>>12)/1000, ((cal1*1000)>>12)%1000);
os_printf("cal 2 : %d.%d \r\n",((cal2*1000)>>12)/1000,((cal2*1000)>>12)%1000);
os_printf("==================\r\n\r\n");

rtc_time_acc += ( ((uint64)(rtc_t2 - rtc_time.time_base)) * ((uint64)((cal2*1000)>>12)));
os_printf("rtc time acc : %lld \r\n",rtc_time_acc);
os_printf("power on time : %lld us\r\n", rtc_time_acc/1000);
os_printf("power on time : %lld.%02lld S\r\n", (rtc_time_acc/10000000)/100, (rtc_time_acc/10000000)%100);
```

This code, as is, won't be included in the arduino core.

Please rewrite it without sprintf nor floats.

I reckon the documentation is quite obscure to me.
Quoting it:
```
system_rtc_clock_cali_proc: Function Get RTC clock period.

Prototype: uint32 system_get_rtc_time(void)    (<--- typo)

Parameter: none

Return: RTC clock period (in us); bit11 ~ bit0 are decimal.

Example: os_printf("clk cal: %d\r\n", system_rtc_clock_cali_proc()>>12);
- RTC clock period has decimal part.
- RTC clock period tends to drift with changes in temperature, so RTC timer is

Note:
- not very precise.
- See RTC demo in Appendix.A.
```

IMAN4K:
Did you notice about this example part of `Docs` for `system_get_rtc_time()`?!
> If system_get_rtc_time returns 10 (which means 10 RTC cycles), and system_rtc_clock_cali_proc returns 5.75 (which means 5.75μs per RTC cycle),the real time is 10 x 5.75 = 57.5μs

So we extract one period time as float (for example 5.75us) and how we suppose to rewrite this without float?
If we don't want to use float and round the result, we get 50us instead of 57.5us!

d-a-v:
if `cali_proc` returns 5.75us, would we have ?:
```
      20bits            3* BCD
<------------------> <--> <--> <-->
xxxxxxxxxxxxxxxxxxxx bbbb cccc dddd
00000000000000000101 0111 0101 0000
          5            7    5    0
```

Maximum value is `2^20 * 1000` which is ~`2^30` and < `2^32`, that fits in 32bits, unit would be not `us` but `ns`.
So we want `system_get_rtc_time`*`system_rtc_clock_cali_proc`, we can multiply them into a 64 bits result and obtain nanosecs (divisible by 1000 for usecs).

That would work without pulling floats and atof (which can be disabled, so a core function would not work in some case).

IMAN4K:
I just don't agree with the first part, if we run a test and call `system_rtc_clock_cali_proc()` it'll return:
```
0x5E29:
00000000000000000101 1110 0010 1001
```
which is not 5.885:
```
00000000000000000101 1000 1000 0101
          5            8    8    5
```
So the correct time would 
**BE**: `5.885 * 10 = 58.85us`
**NOT**: `(24105(0x5e29) * 10) / 1000 = 241.05us`

So, am i wrong?

devyte:
There are several things at play in this discussion:
1. The code proposed here seems seems to be based on code in the Espressif SDK document 2s nonos api reference section A.2.

2. The explanation in the doc for the cali return value is severely lacking, and there are several different ways to interpret it. The following are some possibilities, and it's unclear which is correct:
  a. bits 31-12 are the integer part, bits 11-0 are the fractional part
  b. bits 31-12 are the integer part, bits 11-0 are 3 BCD digits of 4 bits each for the fractional part
  c. bits 31-12 are a scaled up value (i.e.: value in ns), bits 11-0 are the same value in some other format (BCD?)
  d. something else entirely

3. The proposal here does conversion of the high bits into a float going through print to text and scan to float. This also seems to come from the Espressif example. This code is the cause of the objection, and won't be accepted into our core, because it is very inefficient.
Off the top of my head, I think the following would do the same thing without print-scan:
uint32_t high20 = (cal1*1000)>>12)
return (float)high20 / 1000.0;

4. Regardless of efficiency, it is unknown if the example code is correct. Notice that in the example the cali value is first multiplied by 1000 _before_ being right shifted by 12.

IMAN4K:
@d-a-v 
There's one more thing if we read `0x603b` as `cali` the `ns` conversion return `6014ns` while us conversion, as in examples, will return `6.14us` which there is a `126ns` different(and in 1000 RTC cycles it'll be `126,000ns` in total!).
Don't know if the fractional-part is `0.140us` OR `0.014us`?!

d-a-v:
To be honest, I don't know.
You are right with "6.14" vs "6.014". I tend to think the API example is wrong and they should have used `"%d.%03d"` (to get 6.014).
I made some tests and I have no such values (staying between 5.7xx and 5.8xx). Maybe I should cook of freeze the ESP and see what happens :) 

Interestingly, nodemcu went [the simple way](https://github.com/nodemcu/nodemcu-firmware/blob/11592951b90707cdcb6d751876170bf4da82850d/app/modules/tmr.c#L315)

Maybe that would be a question to ask espressif: `"%d.%d"` is not logic, would it be `"%d.%03d"` instead ?

IMAN4K:
So what do you suggest @d-a-v ?!
just close this for lack of documentation or do the `nodemcu` way?

d-a-v:
We could ask espressif about this sprintf on their repository, what do you think ?

PR #5489, opened 2018-12-12
Update hspi_slave.c

GuaAck:
I found that SPI communication did not work due te a wrong timing of the MISO signal with respect to SCK. With slower processores like Arduino UNO (and a small delay of sck due to th level converter) it works, but with a faster Arduino DUE (without converter) it did not. Together with Jiri Bilek I found that MISODM_S has to be set. In the documentation for the ESP32 (should have a very similar SPI interface) I found that also MOSIDN_S should be set in SPI1C2.  The MISO timing is correct, I tested it for a clock of 4 MHZ and 320 kHz. With experiments I found that also the timing of MOSI is robust. Neither a delay of 70 ns in SCK nor in MOSI did disturb the communication. Some oscillograms are here: http://homepage.o2mail.de/g.ackermann/ESP8266.htm

Jiri Bilek added the ICACHE_RAM_ATTR to the function hspi_slave_setStatus() because this function may be indirect called from an interrupt handler.


JiriBilek:
Adding a picture from a logic analyzer. In the first picture you can see the MISO signal changes on rising edges. This is bad because sampling is provided in the very moment.

image

libraries/SPISlave/src/hspi_slave.c
```diff
@@ -92,7 +92,7 @@ void hspi_slave_begin(uint8_t status_len, void * arg)
     SPI1S1 = (((status_len * 8) - 1) << SPIS1LSTA) | (0xff << SPIS1LBUF) | (7 << SPIS1LWBA) | (7 << SPIS1LRBA) | SPIS1RSTA;
     SPI1P = (1 << 19);
     SPI1CMD = SPIBUSY;
-
+    SPI1C2=(0x2<<SPIC2MOSIDN_S) | (0x1<<SPIC2MISODM_S); // GuaAck. 30.10.2018; timing of MISO
```
devyte: The PR already has the info of the code change author as well as the date and specific change diff, so putting that info in the comment isn't very useful. I suggest replacing the comment with more helpful info.
Example:
What do these registers mean? (Set Register blah, which controls/sets/manipulates bleh)
What do the operations do? (Set bit blah to do bleh)

libraries/SPISlave/src/hspi_slave.c
```diff
@@ -114,7 +114,7 @@ void hspi_slave_end()
   SPI1P = B110;
 }
 
-void hspi_slave_setStatus(uint32_t status)
+void ICACHE_RAM_ATTR hspi_slave_setStatus(uint32_t status) // GuaAck. 30.10.2018
```
devyte: You provided an explanation for why this function should have the decorator. I suggest replacing the comment with that.
Also, I suggest inline.

libraries/SPISlave/src/hspi_slave.c
JiriBilek: Explanation of the registers:
We couldn't find any documentation about SPI interface in 8266 but @GuaAck found out it is similar or same as in ESP32. I am particularly referring to "ESP32 Technical Reference Manual Version 3.1", chapter 7.8 (SPI) Registers, register SPI_CTRL2_REG. 
So we set the SPI_MISO_DELAY_MODE to 1 and on logic analyzer I saw the desired change.

devyte: @JiriBilek thank you for the answer. Could you please update the PR and add it as a comment? I am aware that there is no documentation, so I am looking to explain these things directly in the code. The esp32 reference is also helpful.

libraries/SPISlave/src/hspi_slave.c
devyte: Same here, olease change the comment to add the explanation.

libraries/SPISlave/src/hspi_slave.c
JiriBilek: Can I update the PR, not being the OP?

libraries/SPISlave/src/hspi_slave.c
```diff
@@ -85,36 +90,25 @@ void hspi_slave_begin(uint8_t status_len, void * arg)
...
+//  es it should for SPI-Mode = 1
+//	SPIC2MOSIDN_S (SPI_MOSI_DELAY_NUM) = 2: Probably not required, but all tests are done with this setting
     ETS_SPI_INTR_ATTACH(_hspi_slave_isr_handler,arg);
     ETS_SPI_INTR_ENABLE();
 }
+
+void ICACHE_RAM_ATTR hspi_slave_setStatus(uint32_t status) 
+// put in RAM as setStatus is often called from an interrupt routine
 
-void hspi_slave_end()
```
JiriBilek: Why did you delete hspi_slave_end function?

libraries/hspi_slave.c
```diff
@@ -0,0 +1 @@
+Dummy, should be deleted.
```
JiriBilek: I hope you can delete the file in your repository and the deletion is then propagated to master.

devyte:
@JiriBilek @GuaAck what is the status here? I see a deleted function and an unneeded dummy file.

Makuna:
This timing relationship of the data pulse to clock pulse is defined by the SPI mode.  See [WikiPedia SPI](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface).
I could not find what the mode was supoosed to be for SPI slave or even how its exposed.

GuaAck:
> 
> 
> @JiriBilek @GuaAck what is the status here? I see a deleted function and an unneeded dummy file.

I have nothing to do with deleted functions and dummy files, Perhaps I did siomething by accident, Ia am very beginner in github.

With my changes, the SPI communication in my project works reliable for different clock rates. For me the matter is closed. Me and Jiri provided in github a I lot of notes and diagrams with measured wave forms .  The owner of hspi_slave has to decide what to do. 

I am open for any technical questions,
GuaAck

Makuna:
@GuaAck @JiriBilek What SPI mode are your changes setting the SPI Slave to?  When you tested with a SPI Master, did you test all the modes?

GuaAck:
Yes, I tested the original hspi_slave with all modes (selected in the master), all failed.

The changed hspi_slave  is only valid for mode 0. I found no switch to select the mode and mode 0 seems to be most common.

To remember: The problem came up, when I changed over from my UNO to the very much faster DUE. Then I found the incorrect timing.

Best regards
GuaAck

devyte:
@GuaAck with commit 01cf0e6e062ab867ebfacad43abfb8f3b15ceca7 you deleted the end function, and the commit after that deleted the content of hspi_slave.c, leaving the empty file.

JiriBilek:
@devyte, sorry haven't seen the PR for a while. I am afraid, @GuaAck made a mistake and moved some files to a bad directory. I'm gonna fix it in another PR soon as I have no more info from the OP.

devyte:
Closing in favor of #6022 .

PR #5490, opened 2018-12-12
CI: reduce git load

d-a-v:
- git: clone with depth=1
- git: no recursive submodule
  no more dependency on savannah.nongnu.org nor bearssl.org

### DIFF
--- a/package/build_boards_manager_package.sh
+++ b/package/build_boards_manager_package.sh
@@ -48,15 +48,6 @@ srcdir=$PWD
 rm -rf package/versions/$ver
 mkdir -p $outdir
 
-# Get submodules
-modules=libraries/SoftwareSerial
-for mod in $modules; do
-    echo "refreshing submodule: $mod"
-    git submodule update --init -- $mod
-    (cd $mod && git reset --hard)
-done
-echo "done with submodules"
-
 # Some files should be excluded from the package
 cat << EOF > exclude.txt
 .git
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,13 @@ language: bash
 os: linux
 dist: trusty
 
+git:
+  depth: 1
+  submodules: false
+
+before_install:
+  - git submodule update --init  # no recursive update
+
 cache:
   directories:
     - $HOME/astyle


package/build_boards_manager_package.sh
```diff
@@ -135,11 +126,15 @@ cat $srcdir/package/package_esp8266com_index.template.json | \
 
 # Use Github API token, if available
 curl_gh_token_arg=()
-if [ ! -z "$CI_GITHUB_API_KEY" ]; then
+if [ -z "$CI_GITHUB_API_KEY" ]; then
+    echo "curl: API key not present, exit with no error"
+    exit 0
```
igrr: I think that not preparing the package and exiting with 0 exit code is a bad behavior for the script. If I run it locally, it should either generate a package or exit with an error.

Skipping this script if the API key is not set should be handled by `if: env(CI_GITHUB_API_KEY) IS present` in travis yml. If you have observed that that does not work, it should be fixed there, I think.

igrr:
Also need to push the same branch to esp8266/Arduino and check that CI passes, before merging.

package/build_boards_manager_package.sh
igrr: Nevermind, I see now the link to the failed job in gitter. It seems that the condition in the yml does not work correctly, as variables get removed after the job has started. I think we need some other condition there, like `fork = false`...

d-a-v: No more condition in yml.
* In a PR, package script is verbose and does not fail whether or not curl is failing.
* In a branch, curl failing makes the job failed.

d-a-v:
d-a-v approved this pull request.

> Also need to push the same branch to esp8266/Arduino and check that CI passes, before merging.

Countless tests above were always made simultaneously in a PR and in a branch.
Sorry for the noise in your mailboxes, thanks for your patience, CI cannot be tested locally.

No more *Travis failed with "The job exceeded the maximum time limit for jobs, and has been terminated."*

package/build_boards_manager_package.sh
```diff
@@ -132,13 +123,32 @@ fi
...
+curl --silent -D curl-headers.txt "${curl_gh_token_arg[@]}" https://api.github.com/repos/esp8266/Arduino/releases > releases.json
+
+echo "======== >>> curl dumped header >>>"
+cat curl-headers.txt
+echo "======== <<< curl dumped header <<<"
+if ! grep "HTTP.*200 OK" curl-headers.txt > /dev/null; then
+    if [ -z "$CI_GITHUB_API_KEY" ]; then
+        # github api is rate-limiting
+        echo "---- Bad moon phase, in a PR, exit successfully"
+        exit 0
```
igrr: My comment above about failing with non-zero exit code is still valid, I think. If this is started from command line, it should not pretend to succeed without doing anything useful.

d-a-v:
requested changes addressed in #5496

Issue #5476, opened 2018-12-11
Build with 2.5.0-dev is about 200k larger compared to 2.4.2


TD-er:
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [ ] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12|ESP8285 device|other]
- Core Version:			[2.5.0-beta1]
- Development Env:		[Platformio] (test with [own PIO platform](https://github.com/TD-er/ESPEasy/tree/feature/testbuild_core2_5_0_beta1) )
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Generic ESP8266 Module|other]
- Flash Mode:			[dio]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v2 Lower Memory]

### Problem Description

I was testing to see if I could build ESPeasy with core 2.5.0-beta1 using PlatformIO.

- [Test branch ESPeasy](https://github.com/TD-er/ESPEasy/tree/feature/testbuild_core2_5_0_beta1)
- [Test branch platform-expressif8266 with core 2.5.0-beta1](https://github.com/TD-er/platform-espressif8266/tree/feature/2.5.0-beta1)
- packageRepositories [manifest.json](https://td-er.nl/ESPeasy/manifest.json)

Almost all versions I tested of our project were either unable to build due to exceeding max_size limit, or have become so large they will be useless for their intended purpose.

As an example the `minimal_ESP8266_1024_OTA` env in our project:
```
ATA:    [=====     ]  54.8% (used 44852 bytes from 81920 bytes)
PROGRAM: [========= ]  92.8% (used 828304 bytes from 892912 bytes)
```

The last Travis build of the same environment using core 2.4.2 was:
```
DATA:    [=====     ]  54.7% (used 44836 bytes from 81920 bytes)
PROGRAM: [=======   ]  69.0% (used 616192 bytes from 892912 bytes)
```

Am I missing something very obvious?
Like including some library twice (maybe moved into the core) ?

[...]

PR #5492, opened 2018-12-13
Move libstdc++-nox out of IRAM

earlephilhower:
The new non-exception libstdc++ was not referenced in the linker script,
allowing it to end up in IRAM when not needed.  Add the line to match
and move it into IROM where it belongs.


earlephilhower:
Fixes #5476 by getting the std::map classes out of IRAM.

Issue #5487, opened 2018-12-12
[OTA] Device no longer found by Arduino IDE 2.5.0 (Beta 1)


Swiftnesses:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:				[NodeMCU V1.0]
- Core Version:			[Latest Arduino IDE build]
- Development Env:		[Arduino IDE]
- Operating System:		[MacOS]

### Settings in IDE

- Module:  			[NodeMCU V1.0]
- Flash Mode:			[qio|dio|other]
- Flash Size: 		        [4MB (No SPIFFS)]
- lwip Variant:                  [v2 Higher Bandwidth]
- CPU Frequency:		[80Mhz|160MHz]
- Upload Using:		[OTA]

### Problem Description

Following the update to 2.5.0 (Beta 1), devices are no longer found by Arduino IDE (1.8.9).

Restarted IDE, no change. Reverting back to 2.4.2 (full flash erase via OTA), devices show within 10 seconds. Note: the valve is remote to the house, so I have not tried a serial flash (I will, once the weather improves!).

Sketch contains the following ESP parameters:

```
  // Prevent wifi sleeping
  WiFi.setSleepMode(WIFI_NONE_SLEEP);
  // 802.11n
  WiFi.setPhyMode(WIFI_PHY_MODE_11N);
  // Station mode
  WiFi.mode(WIFI_STA);
```

pfeerick:
I can reproduce this with a Wemos D1 mini on Windows using Arduino IDE 1.8.7 and the latest 2.5.0 Beta1 ESP8266 core. OTA on 2.4.2 was working perfectly, but the device is no longer visible with 2.5.0 Beta1. This is with lwIP 1.4 Higher Bandwidth. Re-installing 2.4.2 and re-compiling/re-uploading the code (erasing only the sketch) resulted in the device being visible for OTA within a few seconds.

devyte:
[...]

PR #5494, opened 2018-12-14
Update ArduinoOTA and examples with MDNS.update() calls

devyte:
Allow use of ArduinoOTA without MDNS (meant for controlling MDNS separately)
Add MDNS.update() call in ArduinoOTA.handle()
Add MDNS.update() call in loop in all examples


devyte:
#Fixes #5487 (examples)

Swiftnesses:
Hi @devyte,

Does this mean if I include ArduinoOTA, I don't need to add the mDNS, as a rookie I find this a little confusing (sorry).

devyte:
It does now, but only in latest git. The fix will be released with beta2.

PR #5501, opened 2018-12-15
use pip3 for doc generation

d-a-v:


d-a-v:
That was a try, manually applied in #5496

PR #5502, opened 2018-12-15
Update to 2.5.0-3 toolchain build

### LINKED ISSUE #5497: Binary files of win32/win64 toolchain is unstripped
For example, the size of cc1.exe (packages\esp8266\tools\xtensa-lx106-elf-gcc\1.20.0-26-gb404fb9-2\libexec\gcc\xtensa-lx106-elf\4.8.2) is 72,783,631 bytes. After stripping with the "-d" option, the size of this file is 9,383,845 bytes.

### LINKED ISSUE #5500: GDB needs to be built fully static
GDB it has a shared library link to libmpfr.so on Linux.  Need to make sure it's built fully static as this shared lib is not included nor generally available.

earlephilhower:
Converted to a Makefile-based cross-compile toolchain setup.

GDB now build statically to avoid issues w/shared libs.

Stripped Windows executables.

Fixes #5497
Fixes #5500


earlephilhower:
Windows ZIPs are now down to about the same size as Linux ones (20+MB shaved off the gcc ones).

igrr:
Question: do we need to keep including C library archives into this repo? If the archives are produced by the toolchain build process, we can use copies included into the toolchain. This will reduce repo size increase rate.

earlephilhower:
They're rebuilt from the same sources and configs.  It's cleaner to upload everything, but for space sake I'll drop them unless there's a functional change in this and the future ones.

earlephilhower:
Darn, it's been too long a day.  One more try, sorry for the noise!

Issue #5493, opened 2018-12-14
Rewrite PUYA patch to be more universal and mem friendly.


TD-er:
The (current) patch:
1. uses a static 4KB buffer, which means everyone using the patch has 4KB less heap, including those users who do _not_ have a puya chip. Instead, a dynamically allocated buffer should be used.
2. evaluates the first if condition every single call of the function, although only the first one after boot actually does anything. Instead of a private static, a global could be used that is set elsewhere only once at bootup, and then if condition is no longer needed. Alternatively, The ESPClass could get a new data member (i.e.: instead of the global flag) which is set on construction, and a method to return its value.
3. calculates the flash_chip_id condition in the second if every single call of the function, which makes no sense, because it can't ever change at runtime. Instead, the condition should be fully calculated once and the result stored in a flag, which should then be checked.
4. detects the condition based on flash_chip_id, which isn't very robust, as has been shown with the flash chip size. Instead, a general purpose detection approach should be implemented that transparently covers other potential brands and manufacturers that have the same quirk, should they appear.
Example: on first bootup after flashing, choose a sector, e.g.: within the update area of the flash layout., preferably one that is unlikely to be used. Do a write test and read back sequence known to behave differently with puya chips. If puya is detected, set a quirk flag somewhere. On all successive bootups, detect that the quirk flag is set, so don't do the test again. Use the quirk flag in the write function to decide whether the additional logic is done or not.

Given the above, the patch won't be accepted as-is into our core. There have been several internal discussions on exactly what to do, but given that the patch is available, there have been, and still are, higher priorities, e.g.: figuring out how to ease IRAM usage.

There have been a lot of comments from many users in this discussion, and I see the patch is even being applied elsewhere, but I still don't see a PR here. Instead of wondering what will happen, I invite those involved to make a PR with the proposed patch, and implement at least points 1-3 above on top of it. If you come up with a good solution for point 4, even better.

_Originally posted by @devyte in https://github.com/esp8266/Arduino/issues/4061#issuecomment-447076931_

Will have a look at it tomorrow (well, after sleep)

devyte:
Note: From #4061 :
Patch v3 in use by ESPEasy, which needs rework per points quoted above:
https://github.com/letscontrolit/ESPEasy/blob/mega/patches/puya_v3.patch#L8

mrWheel:
@devyte 
[...]

PR #5504, opened 2018-12-15
 Rewrite PUYA patch to be more universal and mem friendly.

TD-er:
See #5493

Implemented all suggestions, except the extended detection of PUYA like flash operations.
Current detection is only to match the vendor ID from the flash chip ID.


igrr:
> enable/disable that supporrt via the define

That would make sense, in this case the more common pattern would be:

```
#ifndef FOO_SUPPORT
#define FOO_SUPPORT 1
#endif

// Later in the code
#if FOO_SUPPORT
```

TD-er:
We used the define to see if the patch was applied.
And the `flashIsPuya` function was used to display info about the used core to see if the patch was applied and if it has a PUYA chip. (to have the detection of that chip in one place)
So I would like to have that function public accessible to have some indicator on future reported issues.

> missing ets_isr_unmask in the failed path of malloc-ing the temporary buffer

That's a serious one, I will have a look at the suggested changes right now.

TD-er:
@igrr 
> please move puya-related write logic into a new function, e.g. spi_flash_write_puya, and call either that or spi_flash_write from flashWrite.

Should it be in the same Esp.cpp file?
The `spi_flash_write` function is taken from the sdk, but we cannot put it there for obvious reasons.
I assume the puya version of this function should not be public?

igrr:
It's okay to put spi_flash_write_puya into Esp.cpp as a static function.

Regarding detection of whether PUYA is supported or not, okay, understood. Can you modify the define to be conditional as shown in my previous comment? This was it will be possible to override the define and disable PUYA support via e.g. platform.txt.local file.
Also, not a fan of having flashIsPuya function in the public interface. Something like getFlashVendor() with an enum or list of constants for known vendors would be preferable.

TD-er:
All suggestions should now be implemented.

Did it all on a Celeron Chromebook, so now I can appreciate a proper i7 or Xeon ;)

cores/esp8266/Esp.cpp
```diff
@@ -569,9 +578,58 @@ bool EspClass::flashEraseSector(uint32_t sector) {
...
+    return rc;
+}
+#endif
+
 bool EspClass::flashWrite(uint32_t offset, uint32_t *data, size_t size) {
     ets_isr_mask(FLASH_INT_MASK);
-    int rc = spi_flash_write(offset, (uint32_t*) data, size);
+    int rc = 0;
+#if PUYA_SUPPORT
+    if (getFlashChipVendorId() == 0x85) { // 0x146085 PUYA
```
igrr: as a courtesy to user who might be checking the same, please create a definition in Esp.h (as a macro `#define SPI_FLASH_VENDOR_PUYA 0x85` or a class member `static const uint8_t SPI_FLASH_VENDOR_PUYA = 0x85;`)

TD-er:
Is this also merged back to core 2.4.2, since 2.5.0 is beta?
Or should I create a new PR after this gets merged, to a core 2.4.2 branch?

igrr:
We aren't maintaining bugfix releases for older versions, so this should only apply to 2.5.0.

cores/esp8266/Esp.cpp
TD-er: I added all vendor IDs I could find in the Flashrom project.
I keep searching for these IDs myself, so I thought it would be best to included them all.

cores/esp8266/Esp.h
```diff
@@ -23,6 +23,48 @@
 
 #include <Arduino.h>
 
+#ifndef PUYA_SUPPORT
+  #define PUYA_SUPPORT 1
```
devyte: This means that the default is enabled, unless explicitly disabled. I suggest other way around: disabled, unless explicitly enabled. I prefer to avoid having this enabled by default by mistake for every board out there, so please change to 0.
I could be wrong, but the only boards today that seem to have the puya chip are ESP-01s and some ESP-7s, which look to be mislabeled ESP-01s or ESP-01 variants. With the above change, users who need the support can provide the define externally to the build, i.e.: -DPUYA_SUPPORT=1. A control could be added to the IDE menu, or users can simply add the flag locally in boards.txt.

cores/esp8266/Esp.h
```diff
@@ -23,6 +23,48 @@
...
+#endif
+#ifndef PUYA_BUFFER_SIZE
+  // Good alternative for buffer size is: SPI_FLASH_SEC_SIZE (= 4k)
+  // Always use a multiple of flash page size (256 bytes)
+  #define PUYA_BUFFER_SIZE 256
+#endif
+
+// Vendor IDs taken from Flashrom project
+// https://review.coreboot.org/cgit/flashrom.git/tree/flashchips.h?h=1.0.x
+#define SPI_FLASH_VENDOR_ALLIANCE    0x52    /* Alliance Semiconductor */
```
devyte: This is a great help, thank you! there are some known issues with the flash size detection for some vendors, and there was talk of building such a table.

devyte: As an after-thought: maybe it would be better to have this as an enum with assigned values.

cores/esp8266/Esp.cpp
```diff
@@ -569,9 +578,58 @@ bool EspClass::flashEraseSector(uint32_t sector) {
...
+        }
+        rc = spi_flash_read(pos, flash_write_puya_buf, bytesNow);
+        if (rc != 0) {
+            return rc;
+        }
+        for (size_t i = 0; i < bytesNow / 4; ++i) {
+            flash_write_puya_buf[i] &= *ptr;
+            ++ptr;
+        }
+        rc = spi_flash_write(pos, flash_write_puya_buf, bytesNow);
```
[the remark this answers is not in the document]
devyte: ah yes, you're right.

cores/esp8266/Esp.h
devyte: Adding Unknown is a good idea. Please add it as the last/biggest value, e.g.: 0xFF.
Don't touch getVendorId() function, it's ok as it is.
I suggest
- calling the enum flashVendor or similar
- use enum class instead of classic enum (new-ish C++ feature), which forces the enum values to be contained in the class name in a similar way as a namespace (reduces chance of name clash). This also simplifies the enum values, i.e.: class enum flashVendor {AMD = 0x01, AMIC = 0x37,...};
- adding a `flashVendor getFlashVendorFromId(uint8_t id)` function which implements a switch, where all known types fall through to single case that does `return (flashVendor)id;` and a default case that does `return Unknown`

TD-er: OK, will do... tomorrow.
I was just about to test what I had, but this will take a bit more and it is already close to 1am.
The rest is already present, but not yet tested.

TD-er: By the way, should we also have a toString function for it? (will take quite some flash space)

And also I wondered, is there some consensus about how to handle internal strings (used for internal purposes, not to be translated ever) and display strings?
Tasmota uses defines for the display strings, so you can simply use another .h file for translated versions.

devyte: No to toString(). These aren't meant for display, or even user info really, but rather to check conditions and make certain decisions in the internal core behavior. If a core user needs the string table, it can be implemented on his side.
About internal strings in general, strings smaller than... I think it was 4 bytes, are ok to be const char *, while strings longer than that should go to flash with PROGMEM or F() or whatever.
About translations, this has been brought up before. Impementation of complete and generic support for translation requires huge refactoring of current code. It also requires design of the specific approach, there are several possibilities. It can be done in time, but there are higher priorities at the moment, so for now the language for internal strings is English, and it can't be changed.

mrWheel:
[PUYA] Support for PUYA now in core 2.5.0 (?)
Can anybody please explain how to activate this? I have several reports from people trying to get this working with a project of mine but it looks like it is still not.
It would really help if only some documentation would explain how to activate this.

Simply `#define PUYA_SUPPORT 1` in the ino sketch does not show any changing in the building logs. No `-DPUYA_SUPPORT=1` flags.   
Changing line 27 in Esp.h to `  #define PUYA_SUPPORT 1` does not show any change in build/compile flags.

So, for the simple people among us (including myself): How to activate the PUYA chip functionality?

d-a-v:
In `platform.local.txt`:
```
compiler.c.extra_flags=-DSTASSID="yourssid" -DSTAPSK="yourpassword" -DPUYA_SUPPORT=1
compiler.cpp.extra_flags={compiler.c.extra_flags}
```

mrWheel:
@d-a-v 
Thanks for that quick reply!

If there is no `platform.local.txt` should I just make one (in the same map?)?

d-a-v:
`platform.local.txt` needs to be in the same directory where `platform.txt` is.
STASSID and STAPSK to allow examples running on your network without modifying them.

> Is that correct behaviour?

Yes, no need to add this in your source code. It will be useless.

mrWheel:
@d-a-v 
Ok, thanks again.
Is there a drawback to adding this compiler option if you're compiling for/uploading to a ESP8266 that does **not** have a PUYA chip? Probably some performance degradation but no real problems?

TD-er:
Nope, not a real problem.
In ESPeasy we build all versions including Puya support and it works just fine.

Issue #4748, opened 2018-05-20
Files created without leading / in filename are nowhere


borisneubert:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware: ESP8266 (NodeMCU 4M)
- Core Version: latest, 2.4.1
- Development Env: Platformio
- Operating System: Ubuntu

### Settings in IDE

- Module:  			Wemos D1 mini
- Flash Mode:			dio
- Flash Size: 		        4MB
- lwip Variant:  
- Reset Method:	
- Flash Frequency:	       40Mhz
- CPU Frequency:	       80Mhz
- Upload Using:	      SERIAL
- Upload Speed:   921600

### Problem Description

The documentation states that a filename needs to start with a slash but this is not quite obvious. Trying to open a file for writing whose filename has no leading slash is successful. I can even write to the file, close it and determine its size afterwards. It only does not show up in a directory listing. 

Either opening a file whose filename does not start with a leading slash should fail in the first place or the leading slash should be implicitely added.

### Code Snippet

```String filename= "bar.txt";
File file= SPIFFS.open(filename, "w");
file.write((uint8_t*)buffer, 8);
file.close();
```

PR #5511, opened 2018-12-17
Add LittleFS as an optional filesystem, API compatible w/SPIFFS (but not on-flash-format compatible)

earlephilhower:
Adds a `LittleFS` object which uses the [ARMmbed littlefs](https://github.com/ARMmbed/littlefs) to enable a new filesystem for onboard flash, utilizing the exact same API as the existing SPIFFS filesystem.  LittleFS is built for low memory systems that are subject to random power losses, is actively supported by the ARMmbed community, supports directories (not implemented in this PR to minimize the SPIFFS underlying FS class changes), and seems to be much faster in the large-ish read-mostly applications I use.

Simply replace `SPIFFS.begin()` with `LittleFS.begin()` in your sketch, use `LittleFS.open` in place of `SPIFFS.open` to open files, and everything else just works thanks to the magic of @igrr's `File` base class.

LITTLEFS FLASH LAYOUT IS INCOMPATIBLE WITH SPIFFS.  Since it is a completely different filesystem, you will need to reformat your flash (and lose any data therein) to use it.  Tools to build the flash filesystem and upload are at https://github.com/earlephilhower/arduino-esp8266littlefs-plugin and https://github.com/earlephilhower/mklittlefs/

While SPIFFS seems to work well for small files accessed infrequently, its performance can be very slow on large files when doing random seeking and updates.  The included example shows a contrived read-mostly example and demonstrates how the same calls work on either SPIFFS.* or LittleFS.*

Toolchain integration is provided by [mklittlefs](https://github.com/earlephilhower/mklittlefs/) and [arduino-plugin-littlefs](https://github.com/earlephilhower/arduino-esp8266littlefs-plugin/).

May be of interest in #4291, #4748, and others.


libraries/LittleFS/src/LittleFS.cpp
```diff
@@ -0,0 +1,133 @@
...
+    /* NOOP */
+    (void) c;
+    return 0;
+}
+
+
+// these symbols should be defined in the linker script for each flash layout
+#ifndef CORE_MOCK
+#ifdef ARDUINO
+#ifndef SPIFFS_MAX_OPEN_FILES
```
devyte: Rather than use all of these SPIFFS symbols directly, I suggest implementing LITTLEFS versions of them, and have those implementations in terms of the SPIFFS ones. Example:
```
#define LITTLEFS_MAX_OPEN_FILES SPIFFS_MAX_OPEN_FILES

uint32_t _LITTLEFS_start = _SPIFFS_start;

inline int littlefs_hal_read(blah)
{
  return spiffs_hal_read(blah);
}

int LittleFSImpl::lfs_flash_read(const struct lfs_config *c,
    lfs_block_t block, lfs_off_t off, void *dst, lfs_size_t size) {
    LittleFSImpl *me = reinterpret_cast<LittleFSImpl*>(c->context);
    uint32_t addr = me->_start + (block * me->_blockSize) + off;
    return littlefs_hal_read(addr, size, static_cast<uint8_t*>(dst));
}

```

Put all of the definitions elsewhere in an isolated file, so that spiffs symbols are ever only referenced in that one place.
Why do this? The implementation is then better self-contained, which would help if we decide to drop spiffs in favor of littlefs.

libraries/LittleFS/src/LittleFS.h
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        if (wasMounted) {
+            return _tryMount();
+        }
+
+        return true;
+    }
+
+    // Because "/" is an actual full-fledged dir separator in LFS, we need to
```
devyte: I understand the idea, but not the specifics. What is the purpose of the reserved char? Does littlefs interpret it in some special way?

libraries/LittleFS/src/LittleFS.h
```diff
@@ -0,0 +1,446 @@
...
+#include <limits>
+#include <FS.h>
+#include <FSImpl.h>
+#include <debug.h>
+#include <flash_utils.h>
+
+#define LFS_NAME_MAX 32
+#include "../lib/littlefs/lfs.h"
+
+using namespace fs;
```
devyte: Please put everything from this point on, including this "using fs" line, until just before the GLOBAL_INSTANCE line at the end inside a namespace. I suggest LittleFSImplementation, but it could be anything, really.
See comment at the end as well.

libraries/LittleFS/src/LittleFS.h
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    String                      _pattern;
+    LittleFSImpl               *_fs;
+    std::shared_ptr<lfs_dir_t>  _dir;
+    lfs_info                    _dirent;
+    bool                        _valid;
+    bool                        _opened;
+};
+
```
devyte: The end of the implementation namespace should be here.

libraries/LittleFS/src/LittleFS.cpp
```diff
@@ -0,0 +1,133 @@
...
+extern "C" uint32_t _SPIFFS_block;
+
+#define SPIFFS_PHYS_ADDR ((uint32_t) (&_SPIFFS_start) - 0x40200000)
+#define SPIFFS_PHYS_SIZE ((uint32_t) (&_SPIFFS_end) - (uint32_t) (&_SPIFFS_start))
+#define SPIFFS_PHYS_PAGE ((uint32_t) &_SPIFFS_page)
+#define SPIFFS_PHYS_BLOCK ((uint32_t) &_SPIFFS_block)
+#endif
+
+#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_LITTLEFS)
+FS LittleFS = FS(FSImplPtr(new LittleFSImpl( SPIFFS_PHYS_ADDR, SPIFFS_PHYS_SIZE, SPIFFS_PHYS_PAGE, SPIFFS_PHYS_BLOCK, SPIFFS_MAX_OPEN_FILES)));
```
devyte: With the namespace comment for the .h file, this would become:
```
FS LittleFS = FS(FSImplPtr(new LittleFSImplementation::LittleFSImpl( SPIFFS_PHYS_ADDR,....)
```

tests/host/common/littlefs_mock.h
```diff
@@ -0,0 +1,56 @@
...
+    //    by 0x43E9FF: SpiffsMock::save() (littlefs_mock.cpp:116)
+    //    = if (::write(fs, &m_fs[0], m_fs_size) != (ssize_t)m_fs_size)
+    // so switched to a regular array
+    // and that bug is still here
+    // XXXWIPTODO
+
+    uint8_t* m_fs;
+    size_t m_fs_size;
+    bool m_storage;
+    const char *type;
```
devyte: What's this for?

tests/host/common/spiffs_mock.h
```diff
@@ -44,10 +44,10 @@ class SpiffsMock {
     uint8_t* m_fs;
     size_t m_fs_size;
     bool m_storage;
+    const char *type;
```
devyte: What's this for?

tests/host/common/littlefs_mock.h
earlephilhower: Leftover, removed

tests/host/common/spiffs_mock.h
earlephilhower: Leftover, removed

libraries/LittleFS/src/LittleFS.h
earlephilhower: Removed, you shamed me into using real directories and not faking SPIFFS flat filesystem.  There'll be new code replacing this in the next push.

libraries/LittleFS/src/LittleFS.cpp
earlephilhower: Refactored to this kind of setup in the latest push.

libraries/LittleFS/src/LittleFS.cpp
earlephilhower: Done latest commit

libraries/LittleFS/src/LittleFS.h
earlephilhower: Done latest commit

libraries/LittleFS/src/LittleFS.h
earlephilhower: Done latest commit

earlephilhower:
Example output from the sample showing from ~3x to ~24x(!!) faster LittleFS operation (YMMV, depends on flash chip and access pattern):
````
Beginning LittleFS test
Creating 512KB file, may take a while...
==> Time to write 512KB in 256b chunks = 5613 milliseconds
==> Created file size = 524288
Reading 512KB file sequentially in 256b chunks
==> Time to read 512KB sequentially in 256b chunks = 190 milliseconds = 2759000 bytes/s
Reading 512KB file MISALIGNED in flash and RAM sequentially in 256b chunks
==> Time to read 512KB sequentially MISALIGNED in flash and RAM in 256b chunks = 191 milliseconds = 2744000 bytes/s
Reading 512KB file in reverse by 256b chunks
==> Time to read 512KB in reverse in 256b chunks = 693 milliseconds = 756000 bytes/s
Writing 64K file in 1-byte chunks
==> Time to write 64KB in 1b chunks = 1111 milliseconds = 58000 bytes/s
Reading 64K file in 1-byte chunks
==> Time to read 64KB in 1b chunks = 236 milliseconds = 277000 bytes/s
Beginning SPIFFS test
Creating 512KB file, may take a while...
==> Time to write 512KB in 256b chunks = 17552 milliseconds
==> Created file size = 524288
Reading 512KB file sequentially in 256b chunks
==> Time to read 512KB sequentially in 256b chunks = 564 milliseconds = 929000 bytes/s
Reading 512KB file MISALIGNED in flash and RAM sequentially in 256b chunks
==> Time to read 512KB sequentially MISALIGNED in flash and RAM in 256b chunks = 563 milliseconds = 931000 bytes/s
Reading 512KB file in reverse by 256b chunks
==> Time to read 512KB in reverse in 256b chunks = 28469 milliseconds = 18000 bytes/s
Writing 64K file in 1-byte chunks
==> Time to write 64KB in 1b chunks = 7496 milliseconds = 8000 bytes/s
Reading 64K file in 1-byte chunks
==> Time to read 64KB in 1b chunks = 5759 milliseconds = 11000 bytes/s
````

**Comment by earlephilhower** [comment]
Looks like V2 alpha is one step closer to committal.  https://github.com/ARMmbed/littlefs/pull/85#issuecomment-457411638

**Comment by earlephilhower** [comment]
Updated to v2-alpha branch of LittleFS, but I think I'm running into a bug with small files.  We'll see what happens...https://github.com/ARMmbed/littlefs/pull/85#issuecomment-469040744

**Comment by earlephilhower** [comment]
V2 now works and passes host tests.  Now to fix the 10 file merge conflicts....

**Comment by earlephilhower** [comment]
LittleFS V2 was merged upstream, and with this latest commit it's fully integrated and working w/the core.  Added a bunch of SDFS tests as a byproduct (all the SPIFFS/LittleFS tests) for extra goodness.

**Comment by earlephilhower** [comment]
The latest push seems to use even less heap than SPIFFS while providing almost flawless playback of MIDI tunes using the ESP8266Audio library, vs. completely unusable and choppy SPIFFS version.

**Comment by earlephilhower** [comment]
Argh, looks like SWSerial library has gone waky or been `git push --force`d and deleted the tad we were using.  
````
error: Server does not allow request for unadvertised object b178c0ff023ff87a4754636b624a6d18e6036a07
Fetched in submodule path 'libraries/SoftwareSerial', but it did not contain b178c0ff023ff87a4754636b624a6d18e6036a07. Direct fetching of that commit failed.
The command "git submodule update --init" failed and exited with 1 during .
````

File: libraries/LittleFS/src/LittleFS.h [hunk 8]
```diff
@@ -0,0 +1,557 @@
...
+        if (!_mounted) {
+            return false;
+        }
+        info.maxOpenFiles = _maxOpenFds;
+        info.blockSize = _blockSize;
+        info.pageSize = _pageSize;
+        info.maxOpenFiles = _maxOpenFds;
+        info.maxPathLength = LFS_NAME_MAX;
+        info.totalBytes = _size;
+        info.usedBytes = _size - _getUsedBlocks() * _blockSize;
```
TD-er:
```c++
info.usedBytes = _getUsedBlocks() * _blockSize;
```

earlephilhower: Thanks, just fixed in latest push.

connornishijima:
Really looking forward to built-in LittleFS support, any May updates?

earlephilhower:
We've got a bugfix only rev due any time now, but thien this is scheduled for merge.  Give it a couple weeks.  You can always pull the PR and use it yourself to test things.

PR #5516, opened 2018-12-17
fix bug. in sta mode, empty passphrase should not use secure auth mode

acevest:
if the passphrase is empty, use secure auth mode could not connect to wifi.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -113,14 +113,14 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
         return WL_CONNECT_FAILED;
     }
 
-    int passphrase_len = passphrase == NULL ? 0 : strlen(passphrase);
-    if(passphrase_len > 64) {
+    int passphraseLen = passphrase == nullptr ? 0 : strlen(passphrase);
+    if(passphraseLen > 64) {
         // fail passphrase too long!
         return WL_CONNECT_FAILED;
     }
 
     struct station_config conf;
-    conf.threshold.authmode = passphrase_len == 0 ? AUTH_OPEN : _useInsecureWEP ? AUTH_WEP : AUTH_WPA_PSK;
+    conf.threshold.authmode = (passphraseLen == 0) ? AUTH_OPEN : (_useInsecureWEP ? AUTH_WEP : AUTH_WPA_PSK);
 
     if(strlen(ssid) == 32)
         memcpy(reinterpret_cast<char*>(conf.ssid), ssid, 32); //copied in without null term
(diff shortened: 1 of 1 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -113,21 +113,22 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
         return WL_CONNECT_FAILED;
     }
 
-    if(passphrase && strlen(passphrase) > 64) {
+    int passphrase_len = passphrase == NULL ? 0 : strlen(passphrase);
```
devyte: Nitpick: Please use nullptr instead of NULL in C++ code

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -113,21 +113,22 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
...
 
-    if(passphrase && strlen(passphrase) > 64) {
+    int passphrase_len = passphrase == NULL ? 0 : strlen(passphrase);
+    if(passphrase_len > 64) {
         // fail passphrase too long!
         return WL_CONNECT_FAILED;
     }
 
     struct station_config conf;
+    conf.threshold.authmode = passphrase_len == 0 ? AUTH_OPEN : _useInsecureWEP ? AUTH_WEP : AUTH_WPA_PSK;
```
devyte: This line is a bit difficult to understand. Please add parenthesis around the last part: `(_useInsecureWEP ? AUTH_WEP : AUTH_WPA_PSK)`

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
devyte: Also, please prefer camelCase: passphrase_len => passphraseLen

acevest: i've fixed.

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
acevest: ok

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -113,21 +113,22 @@ wl_status_t ESP8266WiFiSTAClass::begin(const char* ssid, const char *passphrase,
...
     struct station_config conf;
+    conf.threshold.authmode = (passphraseLen == 0) ? AUTH_OPEN : (_useInsecureWEP ? AUTH_WEP : AUTH_WPA_PSK);
+
     if(strlen(ssid) == 32)
         memcpy(reinterpret_cast<char*>(conf.ssid), ssid, 32); //copied in without null term
     else
         strcpy(reinterpret_cast<char*>(conf.ssid), ssid);
 
-    conf.threshold.authmode = AUTH_OPEN;
     if(passphrase) {
```
dav1901: Isn't it much simpler just to add `&& strlen(passphrase) > 0` to the if statement ?

acevest: three places need passphraseLen in this function. just add strlen is simple but slow.

acevest: three places need passphraseLen in this function. just add strlen is simple but slow.

PR #5520, opened 2018-12-18
Include <time.h> for missing type time_t

jej:


earlephilhower:
@jej, did you have a compile error with this missing that caused you to add `time.h`?  Since this file is the one exporting the `Serial` object I'm pretty sure it's been building OK in the IDE for quite a while as-is.

jej:
Sorry for the lack of information. Since 2.5.0-beta2 I get
```
In file included from /tmp/arduino_build_644510/sketch/xxx.cpp:2:0:
.../esp8266/2.5.0-beta2/cores/esp8266/HardwareSerial.h:209:34: error: 'time_t' has not been declared
     unsigned long detectBaudrate(time_t timeoutMillis);
```
If I include <time.h> in xxx.cpp it solves the problem. But the time_t is used in HardwareSerial.h not in my code...

jej:
Sorry false report... I cleaned my includes and it's ok now... sorry for the noise...

PR #5522, opened 2018-12-19
Fix negative SPIFFS size for 2M/4M (no SPIFFS) boards

mcspr:
Noticed by @arihantdaga in https://github.com/xoseperez/espurna/issues/1436

Same as <=1024 KiB boards, just make start=end

### DIFF
--- a/tools/boards.txt.py
+++ b/tools/boards.txt.py
@@ -1206,6 +1206,7 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
             sys.stdout = open(lddir + ld, 'w')
 
         if spiffs_kb == 0:
+            spiffs_start = spiffs_end
             page = 0
             block = 0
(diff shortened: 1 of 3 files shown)


arihantdaga:
Can this affect the working of OTA or Usage of EEPROM. Because i noticed that few of my devices stopped working after giving an OTA and on looking into the flash's last 5 sectors(1020 -> 1024). I found there was junk and some parts from the firmware code which should have been limited well before the last 5 sectors. 
I was using semi latest esp core(Just few commits behind the current master). I am not yet able to reproduce this issue, but it happened in few devices before. I am trying to figure out the reason yet.

domingosl:
Will this affect every board size? 1Mb, 2, 3 and 4? I know it breaks 4Mb (No SPIFFS) boards after an OTA update. While waiting for beta-3 is there any combination of flash size + SPIFFS size I can use that work?

PR #5525, opened 2018-12-19
SD Filesystem compatible with 8266 File, using latest SdFat

earlephilhower:
--edit-- Ready for trial, memory error cleaned up, SDWebServer demo and host tests running locally

Arduino forked a copy of SD lib several years ago, put their own wrapper
around it, and it's been languishing in our ESP8266 libraries ever since
as SD. It doesn't support long file names, has class names which
conflict with the ESP8266 internal names, and hasn't been updated in
ages.

The original author of the SD library has continued work in the
meantime, and produced a very feature rich implementation of SdFat. It
unfortunately also conflicts with the class names we use in ESP8266
Arduino and has a different API than the internal SPIFFS or proposed
LittleFS filesystem objects.

This PR puts a wrapper around the latest and greatest SdFat library,
by forking it and wrapping its classes in a private namespace "sdfat,"
and making as thin a wrapper as possible around it to conform to
the ESP8266 FS, File, and Dir classes.

This PR also removes the Arduino SD.h class library and rewrites it
using the new SDFS filesystem to make everything in the ESP8266
Arduino core compatible with each other.

By doing so it lets us use a single interface for anything needing a
file instead of multiple ones (see SDWebServer and how a different
object is needed vs. one serving from SPIFFS even though the logic is
all the same). Same for BearSSL's CertStores and probably a few others
I've missed, cleaning up our code base significantly.

Like LittleFS, silently create directories when a file is created with
a subdirectory specifier ("/path/to/file.txt") if they do not yet exist.

Adds a blacklist of sketches to skip in the CI process (because SdFat
has many examples which do not build properly on the ESP8266).

Now that LittleFS and SDFS have directory support, the FS needs to be
able to communicate whether a name is one or the other.  Add a simple
bool FS::isDirectory() and bool FS::isFile() method.  SPIFFS doesn't
have directories, so if it's valid it's a file and reported as such.

Add ::mkdir/::rmdir to the FS class to allow users to make and destroy
subdirectories.  SPIFFS directory operations will, of course, fail
and return false.

Emulate a 16MB SD card and allow test runner to exercise it by using
a custom SdFat HOST_MOCK-enabled object.

Throw out the original Arduino SD.h class and rewrite from scratch using
only the ESP8266 native SDFS calls.  This makes "SD" based applications
compatible with normal ESP8266 "File" and "FS" and "SPIFFS" operations.

The only major visible change for users is that long filenames now are
fully supported and work without any code changes.  If there are static
arrays of 11 bytes for old 8.3 names in code, they will need to be
adjusted.

While it is recommended to use the more powerful SDFS class to access SD
cards, this SD.h wrapper allows for use of existing Arduino libraries
which are built to only with with that SD class.

Additional helper functions added to ESP8266 native Filesystem:: classes
to help support this portability.

The rewrite is good enough to run the original SDWebServer and SD
example code without any changes.


earlephilhower:
It's not for internal flash, only external SD cards.  You would not want to run FAT on flash w/o wear levelling, it would destroy your chip in no time!

The idea is this lets you use internal flash w/SPIFFS or LittlleFS or external SD card with no code changes by unifying the underlying objects.

liebman:
I wonder if its possible to submit a PR to the SdFat to make the namespace a compile time option.  That would eliminate the need for a fork.

earlephilhower:
@liebman, good point.  I can wrap my namespace changes (3 lines in every lib source and 1 in each example) in a `#ifdef`.  There are still issues with the base SdFat examples, though, where some don't compile for non-AVR(or other specific HW) and emit `#error`s.  That's what's making the CI fail now.

@devyte , @d-a-v, any issues with me adding a blacklist to the CI test runner to stop trying to build *carefully selected* examples (AVR-only, etc.)?  Then I can see about getting a PR into @greiman with the minor namespace patch and we can use the original lib sources (yay, I do not want to maintain other people's libs!).

cores/esp8266/FS.cpp
```diff
@@ -121,6 +121,20 @@ const char* File::name() const {
     return _p->name();
 }
 
+bool File::isFile() const {
+    if (!_p)
+        return nullptr;
```
d-a-v: return false

earlephilhower: Thanks, fixed in latest push.

earlephilhower:
@liebman, per https://github.com/greiman/SdFat/pull/121, it looks like SdFat is hitting maintenance mode in favor of a new exFAT compatible version.  For now I'll just keep my own private fork (since it's going maintenance, don't expect to have to pull many changes from upstream).  When SdExFat or whatever it is called appears, it should be a simple process to migrate to that new lib (either again with a fork or a PR to the main repo).

earlephilhower:
Wow, looks like this PR deletes a net of over 4,000 lines of code from the repo.  Yay!

For brave enough folks, the code now runs pretty stably and well.  I'm able to just build and run the existing SDWebServer demo without any changes and it runs even better than before (it now has long file name support, with no user code changes).

cores/esp8266/FS.h
```diff
@@ -99,12 +119,22 @@ struct FSInfo {
     size_t maxPathLength;
 };
 
+class FSConfig
+{
+public:
+    FSConfig(bool autoFormat = true) {
+        _autoFormat = autoFormat;
+    }
+
```
d-a-v: Thanks !
Did you consider using an enum for the constructor parameter, to make it more explicit/visual for the end user (like `FS_AUTOFORMAT` / `FS_NOFORMAT` instead of a single `true` or `false`) ?

earlephilhower: How about an set/get function instead of an enum?  `setAutoFormat(true);`?

Also, I need to add a poor man's RTTI by putting a type in the shared header so you can't accidentally pass in the wrong config type (well, you could, but the library would catch it).

And after looking at what I had to do w/the testing I think I need a separate `setConfig()` method in FS to make things like "format()" sensible.  Begin will go back to a no-parameter method using a default config if nobody calls setConfig().

libraries/SDFS/src/SDFS.h
```diff
@@ -0,0 +1,404 @@
...
+    {
+        _dir->close();
+    }
+
+    FileImplPtr openFile(OpenMode openMode, AccessMode accessMode) override
+    {
+        if (!_valid) {
+            return FileImplPtr();
+        }
+	char tmpName[128];
```
devyte: Is 128 really ok? This looks like a limitation of the implementation. If so, it should be documented.

libraries/SDFS/src/SDFSFormatter.h
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    bool clearFatDir(uint32_t bgn, uint32_t count) {
+        clearCache(false);
+        if (!card->writeStart(bgn, count)) {
+            DEBUGV("SDFS: Clear FAT/DIR writeStart failed");
+            return false;
+        }
+        for (uint32_t i = 0; i < count; i++) {
+            if ((i & 0XFF) == 0) {
```
devyte: We really need a better mechanism than this. I still think optimistic_yield should get an overhaul to yield every x ms.

cores/esp8266/FS.cpp
```diff
@@ -175,6 +233,22 @@ bool Dir::next() {
     return _impl->next();
 }
 
+bool Dir::rewind() {
```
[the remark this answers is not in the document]
earlephilhower: It may be *technically* correct, but *logically* I'd have to disagree.  The externally visible state is changed by this (i.e. a pointer inside an ope directory), even if this class's own pointer to that encapsulated class doesn't change.

The pointer doesn't change in all other methods as well, and they're none of them `const`.

cores/esp8266/FS.cpp
```diff
@@ -114,13 +114,57 @@ File::operator bool() const {
     return !!_p;
 }
 
+bool File::truncate(uint32_t size) {
```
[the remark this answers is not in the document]
earlephilhower: Same as above.  The total state is different here, even if the indirect pointer to that state we keep doesn't change...

libraries/SD/src/SD.h
```diff
@@ -1,154 +1,134 @@
...
-  SdVolume volume;
-  SdFile root;
-  
-  // my quick&dirty iterator, should be replaced
-  SdFile getParentDir(const char *filepath, int *indx);
-public:
-  // This needs to be called to set up the connection to the SD card
-  // before other methods are used.
-  boolean begin(uint8_t csPin = SD_CHIP_SELECT_PIN, uint32_t speed = SPI_HALF_SPEED);
+    boolean exists(const char *filepath) {
```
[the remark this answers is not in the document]
earlephilhower: For backward compatibility these should be `boolean` and not `bool` (see original SD.h).

I think now w/the boolean->bool (from boolean->uint8_t)  they're identical, but I have a worry that two typedefs, even if identically implemented, might not match and you'd get warnings...

libraries/SDFS/src/SDFSFormatter.h
earlephilhower: Need a new API defined or you'll break existing code.  Could you think of a way of using your templated-timeout for something here?  I think it would fit, but I can't quite get it...

libraries/SDFS/src/SDFS.h
```diff
@@ -0,0 +1,404 @@
...
+            _fd->sync();
+        }
+    }
+
+    bool seek(uint32_t pos, SeekMode mode) override
+    {
+        if (!_opened) {
+            return false;
+        }
+        switch (mode) {
```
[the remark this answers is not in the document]
earlephilhower: It's handled now with the `return false` catch-all at the end of the method.  I think it's more explicit there than moved into a case (since the fcn has more than just the case in it).  Thoughts?

libraries/SDFS/src/SDFS.h
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    const char* name() const override
+    {
+        if (!_opened) {
+            return nullptr;
+        } else {
+            const char *p = _name.get();
+            const char *slash = strrchr(p, '/');
+            return (slash && slash[1]) ? slash + 1 : p;
```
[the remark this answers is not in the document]
earlephilhower: Will update.  Sometimes it's more obvious when writing than when reading.

libraries/SDFS/src/SDFS.h
earlephilhower: Bumped to 260 bytes which is what Wikipedia has to say on the issue, with comment.  The SDFat lib doesn't really care, since the app (us) has to pass it in and keep track of it, not the lib.

libraries/SDFS/src/SDFS.h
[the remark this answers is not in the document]
devyte: Ok, I should explain better. The default case in a switch should be implemented in all cases. For those cases where it doesn't make sense, I.e. should never happen, it usually contains an assert. It's a code robustness thing, it guards against a case that, well, should never happen. If it does happen, it means something is very wrong code wise, e.g. corruption, in which case a silent return false is not going to give any clues.

libraries/SDFS/src/SDFSFormatter.h
devyte: Instantiate a periodic and call yield() on timeout. It's kind of the inverse of the canonical usage.

libraries/SDFS/src/SDFS.h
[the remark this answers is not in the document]
earlephilhower: Added in some debug messages on bad calls in general and a default handler in this specific case.

libraries/SDFS/src/SDFS.h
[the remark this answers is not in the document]
earlephilhower: Don't know why this isn't showing updated, comments were added to this section of code...

libraries/SDFS/src/SDFSFormatter.h
earlephilhower: I've put in what I think will give a 5ms delay between yields.  Can you check the new code in this area?  Seems too simple to work, but I didn't have any problem with running it.

libraries/SDFS/src/SDFS.h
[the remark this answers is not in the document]
devyte: You don't think an assert is valid here?

libraries/SDFS/src/SDFSFormatter.h
devyte: Looks correct :smiley:

libraries/SDFS/src/SDFS.h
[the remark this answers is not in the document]
earlephilhower: OK, you win.  Just added with a condition that's guaranteed to fail should it be hit.  Still returning false afterwards for release builds so the file open will fail instead of be in an indeterminate state.

cores/esp8266/spiffs_api.h
```diff
@@ -124,6 +127,27 @@ class SPIFFSImpl : public FSImpl
...
+        return false;
+    }
+
+    bool rmdir(const char* path) override
+    {
+        (void)path;
+        return false;
+    }
+
+    bool setConfig(const FSConfig *cfg) override
```
devyte: How about passing by const reference instead of by const pointer? No nullptr allowed.

libraries/SD/src/SD.h
```diff
@@ -1,154 +1,134 @@
...
+#undef FILE_WRITE
+#define FILE_WRITE (sdfat::O_READ | sdfat::O_WRITE | sdfat::O_CREAT)
 
-#define FILE_READ O_READ
-#define FILE_WRITE (O_READ | O_WRITE | O_CREAT)
+class SDClass {
+public:
+    boolean begin(uint8_t csPin, SPISettings cfg = SPI_HALF_SPEED) {
+        auto fsCfg = SDFSConfig(csPin, cfg);
+	SDFS.setConfig(&fsCfg);
```
devyte: If setConfig() were to receive by const ref, this wouldn't need to pass the address of fsCfg, which looks weird.

libraries/SDFS/src/SDFS.cpp
```diff
@@ -0,0 +1,155 @@
...
+ */
+#include "SDFS.h"
+#include "SDFSFormatter.h"
+#include <FS.h>
+
+using namespace fs;
+
+
+#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SDFS)
+static sdfs::SDFSImpl *_SDFSImpl = new sdfs::SDFSImpl();
```
devyte: Why save this pointer? It doesn't seem to serve any purpose.
How about something like this:
```
FS SDFS = FS(FSImplPtr(new sdfs::SDFSImpl()));
```
?

cores/esp8266/spiffs_api.h
earlephilhower: Good idea, makes nicer code.  Updated in latest push.

libraries/SD/src/SD.h
earlephilhower: Ditto, and now we can make a SDFSConfig object directly in-line with the setConfig method.

libraries/SDFS/src/SDFS.cpp
earlephilhower: No idea, but it's cleaned up in the latest push.

PR #5529, opened 2018-12-19
Add support for newer mobile OS changes.

nouser2013:
Wanted to update the examples with the ones actually working on modern mobile OS. Tested on iOS 12.1, Android 5 and Android 7.0.


libraries/DNSServer/examples/CaptivePortal/CaptivePortal.ino
```diff
@@ -3,7 +3,7 @@
 #include <ESP8266WebServer.h>
 
 const byte DNS_PORT = 53;
-IPAddress apIP(192, 168, 1, 1);
+IPAddress apIP(172, 217, 28, 1);
```
earlephilhower: This PR doesn't make a lit of sense.

192.168.x.x is a private class B subnet, 172.x.x.x is a private class C subnet.  Both are equally valid AP addresses, and should have no effect upon DHCP clients who obey the relevant standards, which I certainly hope all OSes, mobile or otherwise, do.

d-a-v:
172.217.x.x is not part of a private range (172.16.0.0 .. 172.31.255.255 is, [ref](https://en.wikipedia.org/wiki/Private_network)).
@nouser2013 can you elaborate on your proposed changes ?

nouser2013:
Hi, gladly. Resolving the captive checks to an IP in a public range apparently triggers the notification to the user correctly, as stated in the referenced comment [here](https://github.com/espressif/arduino-esp32/issues/1037).

Therefore, I'd propose to adapt the changes also to these examples in order to save future developers the headache of debugging the captive portal examples not functioning on Android.

d-a-v:
I tried with android 7.1.2(LineageOS), and these sketches are working with 192.168.4.1 *and* data plan disabled on mobile. Notification to log in appears. (later:  this test was working maybe by chance, see below).

We are opened for improvements, but setting a random public IP address instead of an official private one (192.168.x.x, 10.x.x.x, 172.16.x.x..172.16.31.x.x) is not a good idea.

At first I was thinking that it might be related with complexity of URL handling in modern browsers because of unification into a single user input field of 1) URL, 2) searchbar 3) cache, and 4) general simplification of the so-called "user experience" leading to hidden-maybe-unwise assumptions.

Then I tried with a laptop (doing lots and lots of dns requests):
- captiveportal example sadly dies with WDT from within a too long loop
- with WiFiManager - which using DNSServer from this core - I observed that sometimes captive-dns works, sometimes not.
- more importantly, and sadly, the `dig` tool reports errors: `Warning: Message parser reports malformed message packet`

Aside from this, there are some [weirdnesses](https://github.com/esp8266/Arduino/blob/20b7e480b5a3e9aeffc056020d816a2fad4cbe12/libraries/DNSServer/src/DNSServer.cpp#L122) in our DNSServer.

We must solve this DNSServer issue first, and check afterwards whether the issue from which this PR originates is still valid.

@devyte @LaborEtArs I read somewhere that mDNS and DNS have the same packet format. Our DNSServer is not meant to be a DNS forwarder so both are similar regarding pure-local-only-DNS functionality. Is there any chance that we could include local-only legacy DNS server from a subset of mDNS code ?

@tablatronix what are your thinkings ?

nouser2013:
> [...] but using a random non class C ip is not a good idea officially, considered a kludge.
Out of curiosity, could you elaborate? E.g., is there a standardized captive-portal behaviour I've overlooked?

As to why it seems to work on (my) 7.0, 5.0 and 4.3 devices with the public IP only, I cannot even speculate. It _shouldn't_ make a difference.

tablatronix:
I think this is still being discussed in that other post, it seems to be some kind of security restriction in some versions of android, or maybe a setting, I don't think anyone has figured it out yet, as we could not find specific definitive docs on it. It should work.

LaborEtArs:
@d-a-v Your right, mDNS uses the DNS protocol. However the implemented logic is quite different... so, to get a DNS server would be a lot of additional work...

Misiu:
any updates on this? I've seen multiple places mentioning that Captive Portal doesn't work on some devices.
https://github.com/tripflex/wifi-captive-portal/issues/7
https://github.com/Hieromon/AutoConnect/issues/85
https://github.com/tripflex/captive-portal/blob/dev/src/mgos_captive_portal.c#L368-L380

DNSServer got a rework so maybe this can be finally fixed?

d-a-v:
Given the links above, and also [this one](https://github.com/esp8266/Arduino/pull/6173#issuecomment-506307686), and if the examples work better, I tend to think we can merge it if we can add documentation (at least a ref to this issue) in these changes.

@nouser2013, or @MisterD66 from #6173, can you do that ?

d-a-v:
@nouser2013 and @MisterD66 thank you for bringing these fixes.

That helped understanding and solving other issues (I finally experimented it too).
I'm going to merge this one (two fixes) and I'll add comments in a further one.

d-a-v:
closed by #5529

PR #5539, opened 2018-12-21
Speed up CI builds with caching hacks

earlephilhower:
Hopefull give a significant speedup to the Arduino CI process.


earlephilhower:
There's something funky in the way we build in CI so arduino-builder thinks the cache is always invalid.  Making the GIT version header on each INO, probably.  Closing.

earlephilhower:
It now preserves all intermediate .o files in the libraries and core, speeding up the Arduino builds by >50%!

devyte:
This is aramaic to me, so @d-a-v : tag, you're it.

earlephilhower:
Amharic?  I was going for Coptic..  I've updated the comments to better explain the logic behind the steps for @d-a-v's sanity.

tests/common.sh
```diff
@@ -53,7 +53,30 @@ function build_sketches()
         if [ $testcnt -ne $build_rem ]; then
             continue  # Not ours to do
         fi
-        rm -rf $build_dir/*
+
+        if [ -e $cache_dir/core/*.a ]; then
+            # We need to preserve the build.options.json file and replace the last .ino
+            # with this sketch's ino file, or builder will throw everything away.  This
+            # head/tail easier than a complicated SED script to replace the 1 line we need.
```
d-a-v: `sed -i "s,^.*sketchLocation.*$,  \"sketchLocation\": \"$sketch\"\,,g" build.options.json`
`touch -r $build_dir build.options.json`

I did not test this `touch` (I expect it will work beyond 2038)

earlephilhower: Thanks for the SED, I wasn't able to find a good way of escaping the /s in the path OTW!

The touch won't work because the builder regens the gitver.h file after this bit of the script runs.  I've updated it to be 'now + 1 day' which will cover us until the end of time...

PR #5542, opened 2018-12-23
mDNS: Add support for String arg to begin()

jjsuwa:
#5295 should be applied to mDNS::begin() I guess :)


devyte:
@jjsuwa thanks!

PR #5546, opened 2018-12-23
[Proposal] Clarify (the existence of) the constant empty String

jjsuwa:
instead of use of naked global symbol.


cores/esp8266/WString.h
```diff
@@ -70,6 +72,9 @@ class String {
         explicit String(double, unsigned char decimalPlaces = 2);
         ~String(void);
 
+        // empty string singleton
+        static constexpr const String& empty = ::__emptyString;
```
devyte: Please don't call the static member "empty". It should be ok to also call it "emptyString", because it is scoped, or call it something else (emptyStr, nullString, defString, etc)

devyte: Given the deprecation, can this be implemented the other way around? make this the member the "real" object, and make the global a reference to this one.

d-a-v:
@jjsuwa Are you going to make a new PR for this ?

PR #5557, opened 2018-12-26
Fixes for IPv6, added in CI

d-a-v:


davefiddes:
I'd just been trying IPv6 support and run into build problems. I had a hack fix for it but this is neater and looks OK to me.

PR #5558, opened 2018-12-27
(re)introduce timeout in HardwareSerial::readBytes(buffer, size)

d-a-v:
and add 
- `HardwareSerial::read(buffer, size)`
- visual test

per [comment](https://github.com/esp8266/Arduino/commit/4c8d8f1e8aadf5236d5b1b9406197885385dd45c#commitcomment-31767817)


cores/esp8266/HardwareSerial.cpp
```diff
@@ -132,6 +133,22 @@ unsigned long HardwareSerial::detectBaudrate(time_t timeoutMillis)
...
+size_t HardwareSerial::readBytes(char* buffer, size_t size)
+{
+    size_t got = 0;
+
+    while (got < size)
+    {
+        esp8266::polledTimeout::oneShot timeOut(_timeout);
+        size_t avail;
+        while ((avail = available()) == 0 && !timeOut);
+        if (avail == 0)
```
devyte: This is the correct condition. It's common to check for the counterpart, i.e.: if(timeOut) in this case, but this is better.

Issue #5513, opened 2018-12-17
[core 2.5.0] Loop Count much lower than with 2.4.2


TD-er:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12|ESP-01|ESP-07|ESP8285 device|other]
- Core Version:			[0fd86a07f0b22aa43f24dc6158b6d8093b607765]
- Development Env:		[Platformio]
- Operating System:		[Windows|Ubuntu]

### Settings in IDE

- Module:  			[Generic ESP8266 Module|other]
- Flash Mode:			[dio]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY_LOW_FLASH]
- Reset Method:		[-]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA|SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

Loop Count in loops per sec. is much lower than with core 2.4.2. It was also mentioned here: https://github.com/esp8266/Arduino/issues/5476#issuecomment-447115849

I have several test setups running core 2.4.2 and core 2.5.0 (see latest nightly builds of ESPeasy)
For example 2 Sonoff POW r2 nodes running the exact same configuration, the loopcount (per sec) of both differs quite a lot:

Core 2.4.2:
![image](https://user-images.githubusercontent.com/3751318/50083889-7a2e9800-01f5-11e9-89b0-54a97aaa4850.png)

Core 2.5.0 beta1:
![image](https://user-images.githubusercontent.com/3751318/50083935-9af6ed80-01f5-11e9-9ef8-5bf31dfd8ddf.png)

N.B. Not sure yet why in the node running 2.4.2 less things are being shown.

In general (among my test nodes) the loop-count is:
- 4000 - 11000 when using core 2.4.2 
- 1100 - 2300 when using core 2.5.0 (4200 when doing nothing)

[...]

PR #5559, opened 2018-12-27
GDB support w/new toolchain and UART driver

earlephilhower:
This extends #4386 by manually fixing the merge issues in @kylefleming 's original PR, fixing the linker literal assignment errors, and updating the PR to use the open source GDB protocol for Xtensa (as is included in the new toolchain).

It's still WIP because I've not tried everything out nor written up how to use it, but it's able to stop a program in mid-run, set a breakpoint, continue, show values, and change them.

````
#include <GDBStub.h>
extern "C" void gdbstub_init();
void setup() {
gdbstub_init();
Serial.begin(115200);
Serial.printf("Starting...\n");
}

void loop() {
static  uint32_t cnt = 0;
Serial.printf("%d\n", cnt++);
//*(int*)0=1;
delay(100);
}
````

````
earle@server:~/src/esp-quick-toolchain/repo/binutils-gdb/gdb$ ~/Arduino/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/bin/xtensa-lx106-elf-gdb
GNU gdb (GDB) 8.2.50.20180723-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "--host=x86_64-linux-gnu --target=xtensa-lx106-elf".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
(gdb) set remote hardware-breakpoint-limit 1
(gdb) set remote hardware-watchpoint-limit 1
(gdb) set remote interrupt-on-connect on
(gdb) set remote kill-packet off
(gdb) set remote symbol-lookup-packet off
(gdb) set remote verbose-resume-packet off
(gdb) file /tmp/arduino_build_257110/sketch_dec26a.ino.elf
Reading symbols from /tmp/arduino_build_257110/sketch_dec26a.ino.elf...done.
(gdb) mem 0x20000000 0x3fefffff ro cache
(gdb) mem 0x3ff00000 0x3fffffff rw
(gdb) mem 0x40000000 0x400fffff ro cache
(gdb) mem 0x40100000 0x4013ffff rw cache
(gdb) mem 0x40140000 0x5fffffff ro cache
(gdb) mem 0x60000000 0x60001fff rw
(gdb) set serial baud 115200
(gdb) target remote /dev/ttyUSB0
Remote debugging using /dev/ttyUSB0
0x40000f68 in ?? ()
(gdb) where
#0  0x40000f68 in ?? ()
#1  0x40000f58 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) cont
Continuing.
290
291
292
293
294
295
296
297
298
299
300
301
302
303
304
305
306
307
308
309
310
311
^C
Program received signal SIGINT, Interrupt.
0x40000f68 in ?? ()
(gdb) where
#0  0x40000f68 in ?? ()
#1  0x40000f58 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) lsit
Undefined command: "lsit".  Try "help".
(gdb) list
1	#include <Arduino.h>
2	#line 1 "/home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino"
3	#line 1 "/home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino"
4	#include <GDBStub.h>
5	extern "C" 
6	void gdbstub_init();
7	#line 4 "/home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino"
8	void setup();
9	#line 11 "/home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino"
10	void loop();
(gdb) cont
Continuing.
bcn_timout,ap_probe_send_start
312
313
314
315
^C
Program received signal SIGINT, Interrupt.
0x40000f68 in ?? ()
(gdb) break loop
Breakpoint 1 at 0x402024d7: file /home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino, line 11.
(gdb) cont
Continuing.
Note: automatically using hardware breakpoints for read-only addresses.

Breakpoint 1, loop () at /home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino:11
11	void loop() {
(gdb) where
#0  loop () at /home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino:11
#1  0x40202a40 in loop_wrapper () at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_main.cpp:124
#2  0x401013f9 in cont_wrapper () at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/cont.S:80
(gdb) info registers
pc             0x402024d7          0x402024d7 <loop()+3>
lbeg           <unavailable>
lend           <unavailable>
lcount         <unavailable>
sar            0x1e                30
litbase        0x0                 0
ps             0x20                32
threadptr      <unavailable>
scompare1      <unavailable>
a0             0x40202a40          1075849792
a1             0x3fffffb0          1073741744
a2             0x3ffee3b8          1073669048
a3             0x40202a20          1075849760
a4             0x0                 0
a5             0x0                 0
a6             0x0                 0
a7             0x0                 0
a8             0xfffffffe          -2
a9             0xffffffff          -1
a10            0x3fffc6fc          1073727228
a11            0x1                 1
a12            0x3ffee558          1073669464
a13            0x0                 0
a14            0x3fffdad0          1073732304
a15            0x3ffee58c          1073669516
(gdb) print cnt
No symbol "cnt" in current context.
(gdb) next
loop () at /home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino:14
14	Serial.printf("%d\n", cnt++);
(gdb) print cnt
$1 = 316
(gdb) set cnt=2000
(gdb) print cnt
$2 = 2000
(gdb) cont
Continuing.
2000

Breakpoint 1, loop () at /home/earle/Arduino/sketch_dec26a/sketch_dec26a.ino:11
11	void loop() {
(gdb) cont
Continuing.
2001
````

### CODE REVIEW

File: cores/esp8266/uart.c [hunk 1]
```diff
@@ -571,36 +645,17 @@ uart_uninit(uart_t* uart)
...
-        break;
+    if(uart->tx_enabled && (!gdbstub_has_uart_isr_control() || uart->uart_nr != UART0)) {
+        uart_disable_pin(uart->tx_pin);
     }
 
-    if(uart->rx_enabled)
-    {
-        uart_stop_isr(uart);
+    if(uart->rx_enabled) {
         free(uart->rx_buffer->buffer);
```
devyte: Why are these two frees not conditional together with the uart_disable_pin and uart_stop_isr calls? Is the buffer not needed when gdb is active? Or does the isr keep firing after this uninit call for gbd?

cores/esp8266/uart.c
```diff
@@ -614,47 +669,39 @@ uart_swap(uart_t* uart, int tx_pin)
...
-                pinMode(uart->tx_pin, (tx_pin == 2)?FUNCTION_4:SPECIAL);    //TX
-
-            if(uart->rx_enabled) 
-                pinMode(3, SPECIAL);    //RX
-
             IOSWAP &= ~(1 << IOSWAPU0);
         }
 
+        if(uart->tx_enabled) { //TX
+            uart0_enable_tx_pin(uart->tx_pin);
```
devyte: In the previous code, if pin 2, then FUNCTION_4, else SPECIAL.
In the current code, if pin 2 ||15, then FUNCTION_4, else SPECIAL.
The mode for pin 15 is different. This looks fishy to me.

cores/esp8266/uart.c
```diff
@@ -798,22 +841,26 @@ void
...
         break;
     case UART_NO:
     default:
-        system_set_os_print(0);
-        ets_install_putc1((void *) &uart_ignore_char);
+        func = &uart_ignore_char;
         break;
     }
+    if(!gdbstub_has_putc1_control()) {
+        system_set_os_print((uint8)((uart_nr == UART0 || uart_nr == UART1)?1:0));
```
devyte: This logic for the argument is hard to understand. Please pull it out and assign it to a local uint8 with a meaningful name, then use that in the call.

doc/gdb.rst
```diff
@@ -0,0 +1,365 @@
...
+.. code:: cpp
+
+    (gdb)     target remote /dev/ttyUSB0
+    Remote debugging using /dev/ttyUSB0
+    0x40000f68 in ?? ()
+    (gdb)
+
+Don't worry that GDB doesn't know what is at our present address, we broke
+into the code at a random spot and we could be in an interrupt, in the
+ROM, or elsewhere.  The important bit is that we're not connected and
```
devyte: not => now?

libraries/GDBStub/src/internal/gdbstub.c
```diff
@@ -111,118 +86,85 @@ struct XTensa_exception_frame_s gdbstub_savedRegs;
...
 #define ST_OK -3
 #define ST_CONT -4
+#define ST_DETACH -5
 
 //Grab a hex value from the gdb packet. Ptr will get positioned on the end
 //of the hex string, as far as the routine has read into it. Bits/4 indicates
 //the max amount of hex chars it gobbles up. Bits can be -1 to eat up as much
 //hex chars as possible.
-static long ATTR_GDBFN gdbGetHexVal(unsigned char **ptr, int bits) {
+static long /*ATTR_GDBFN*/ gdbGetHexVal(unsigned char **ptr, int bits) {
```
devyte: Commented ATTR_GDBFN? should it be removed?

devyte: Same for other functions below.

libraries/GDBStub/src/internal/gdbstub.c
```diff
@@ -233,73 +175,157 @@ static long ATTR_GDBFN gdbGetHexVal(unsigned char **ptr, int bits) {
...
-	uint32_t a[16];
-	uint32_t pc;
-	uint32_t sar;
-	uint32_t litbase;
-	uint32_t sr176;
-	uint32_t sr208;
-	uint32_t ps;
-};
+
+static inline bool ATTR_GDBFN gdbRxFifoIsEmpty() {
```
devyte: This whole FIFO code looks very similar to what's in the uart.* files. Is it worth investigating unification?

libraries/GDBStub/src/GDBStub.h
```diff
@@ -1,6 +1,6 @@
 #ifndef GDBSTUB_H
```
devyte: This #define is the same as the one in libraries/GDBStub/src/internal/gdbstub.h (file is near bottom of this diff). Is that on purpose? This can cause shadowing in the include chain.

devyte:
@earlephilhower In #4386 there are some observations, e.g.: 
- by @igrr to have a switch to enable the smaller panic-only handler
- about 3.7KB RAM usage

Does it make sense to pursue them in view of the current status of this implementation?

earlephilhower:
@devyte wrote:
> by @igrr to have a switch to enable the smaller panic-only handler
He was talking about the fatal exception handler (postmortem).  It's already moved out in 2.5.0-b2

> about 3.7KB RAM usage
That's *IRAM* usage AFAICT from the thread.  The movement of things to flash with stubs that call os-enable-flash was already done.  We're at ~700b of IRAM and ~1KB of heap (and ~3KB of flash).  So it's not something we would want to enable by default, but it's not unreasonable.  Might be heap/flash savings possible, but I've not yet looked into any optimizations other than minimizing the IRAM footprint..

Since the other review comments look like work, I'll leave them for tomorrow.

libraries/GDBStub/src/GDBStub.h
earlephilhower: Removed this copy and moved the internal one into an easily accessed spot.  Only one [Gg][Dd][Bb]stub.h file now exists and it has all of @kylefleming's functions defined in it now.

doc/gdb.rst
earlephilhower: Yes, updated.  Thx.

cores/esp8266/uart.c
```diff
@@ -452,8 +481,10 @@ uart_flush(uart_t* uart)
     if(uart->tx_enabled)
         tmp |= (1 << UCTXRST);
 
-    USC0(uart->uart_nr) |= (tmp);
-    USC0(uart->uart_nr) &= ~(tmp);
+    if(!gdbstub_has_uart_isr_control() || uart->uart_nr != UART0) {
```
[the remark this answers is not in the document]
d-a-v: it can be read as `if (!( gdbstub_set_uart_isr_callback() && uart->uart_nr == UART0 )) do-this"
= "do this if gdb hasn't control over uart0".
It seems OK.

cores/esp8266/uart.c
```diff
@@ -513,20 +583,16 @@ uart_init(int uart_nr, int baudrate, int config, int mode, int tx_pin, size_t rx
...
-            {
-                uart->tx_pin = 1;
-                pinMode(uart->tx_pin, FUNCTION_0);
-            }
+            if (tx_pin == 2) {
+                 uart->tx_pin = 2;
+             } else {
+                 uart->tx_pin = 1;
+             }
+             uart0_enable_tx_pin(uart->tx_pin);
```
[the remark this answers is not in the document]
d-a-v: You are right. This is tricky.
The only change I see is uart0/tx/gpio1 was set to F0, now SPECIAL.
@earlephilhower can check if it works from SPECIAL to FUNCTION_0 in `uart0_enable_tx_pin()` when `tx`=1?
That's the most common case (uart0: tx=1, rx=3)

cores/esp8266/uart.c
```diff
@@ -555,12 +620,21 @@ uart_init(int uart_nr, int baudrate, int config, int mode, int tx_pin, size_t rx
...
     USC0(uart->uart_nr) = config;
-    uart_flush(uart);
-    USC1(uart->uart_nr) = 0;
-    USIC(uart->uart_nr) = 0xffff;
-    USIE(uart->uart_nr) = 0;
-    if(uart->uart_nr == UART0 && uart->rx_enabled)
-        uart_start_isr(uart);
+
+    if(!gdbstub_has_uart_isr_control() || uart->uart_nr != UART0) {
+        uart_flush(uart);
```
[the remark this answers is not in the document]
d-a-v: `if (!( gdbstub_has_uart_isr_control() && uart->uart_nr == UART0 ))` ==
if gdb not enabled on uart0, do as before

cores/esp8266/uart.c
d-a-v: It indeed seems those free should go into is the gdb-has-uart test

cores/esp8266/uart.c
```diff
@@ -555,12 +620,21 @@ uart_init(int uart_nr, int baudrate, int config, int mode, int tx_pin, size_t rx
...
+        USC1(uart->uart_nr) = 0;
+        USIC(uart->uart_nr) = 0xffff;
+        USIE(uart->uart_nr) = 0;
+    }
+    if(uart->uart_nr == UART0) {
+        if(uart->rx_enabled && !gdbstub_has_uart_isr_control()) {
```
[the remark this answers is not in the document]
d-a-v: They indeed look mutually exclusive.
@earlephilhower could you add some prints to check what happens around here on init ?
What's suspicious is that `ETS_UART_INTR_ENABLE()` is called below, but it seems that the handler cannot have been set.

cores/esp8266/uart.c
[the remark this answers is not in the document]
earlephilhower: Undid this and other refactoring so the code here is identical and setting the same values as before.

cores/esp8266/uart.c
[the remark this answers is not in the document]
earlephilhower: Same as prior comment, it's intended.

cores/esp8266/uart.c
earlephilhower: I've reworked the `uart_uninit` function to always stop the ISR (which doesn't really stop the ISR when GDB is active, it just tells GDB there is no IRQ callback to stuff received bytes into).  Will be in a push in a few mins.

The `free`s are correct as-is, GDB does NOT use any of the uart.c `struct uart_` state.

cores/esp8266/uart.c
earlephilhower: This was fixed in the last commit where the uart refactoring was removed.  All register settings are now the same as in master branch.

cores/esp8266/uart.c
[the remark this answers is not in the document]
earlephilhower: Agreed.  Changed in an upload to come in a few mins to always call the start isr.  Tested by doing a `Serial.readString` while GDB was enabled and it worked fine w/the change.

cores/esp8266/uart.c
earlephilhower: Reordered to positive condition and broke out the trinay in the next push to come.

libraries/GDBStub/src/internal/gdbstub.c
earlephilhower: Not worth it in my opinion after going through it a few times.

The GDB infrastructure is completely independent from the `struct uart_` stuff in `uart.c` in order to simplify the design (i.e. user can free a `uart_` structure by doing a Serial.end(). but GDB still needs to continue to run).

This also minimizes the logic changes in `uart.c` (an already hairy piece of code!) so it's also got that as a positive.

libraries/GDBStub/src/internal/gdbstub.c
earlephilhower: Yes, this is a leftover from me trying things out as I moved code to flash from IRAM.  Removing the commented out parts in my next push.

freeck:
Hi there,
Great work!! Got it working using Eclipse-Sloeber-IDE, so no complicated userinterventions using gdb...
Also the exceptionhandler seems to  work ok. IRAM-usage much better than previous versions.
Need more testing with a serious application.

earlephilhower:
@freeck, thanks for the report!  Is there anything needed to make it work with your IDE?  I can update the docs easily enough (but run CLI for most things so haven't used Eclipse or an IDE in a long time).

freeck:
I will adapt my document (the document should reside on this site somewhere) based on the previous version of gdbstub.  It is quit easy to configure Eclipse. 
A question: are the mem settings as defined in the cmds-file fixed for all the esp8266 chips and all boards?

earlephilhower:
Yes, the mem settings are applicable to all boards.  While the 512M flash boards don't have the full 1MB program flash, the address space that doesn't exist should never be accessed by GDB (since your program can never reference it).

freeck:
Question perhaps a bit off topic: Is there any documentation on how to handle exceptions on application level (try catch) , with and without using this debugger? With the example provided I observe that the debugger halts at line 4....(-fexceptions). Adding a try catch statement around line 4 show the same result.
1:void loop() {
2:static  uint32_t cnt = 0;
3:Serial.printf("%d\n", cnt++);
4:*(int*)0=1;
5: delay(100);
6:}

earlephilhower:
They're 2 different things.  Hardware exceptions vs C++ exceptions.
1. `*(int*)0=1;` == HW exception (think segfault, bluescreen, etc.)
2. `char *toobigary = new char[9999999];` == C++ exception (it throws a std::exception out-of-mem exception).

You can't catch() the 1st type, they're below the language and hardware actually does a jump to a HW exception vector.

freeck:
Unfortunately I observe quit a larger IRAM-usage of 2.5.0.beta2 compared to 2.4.2. , almost 2k extra IRAM with my application. Can you confirm this? (So no space left for GDBstub....).

freeck:
I observe that when including the twowire-library the IRAM usage exceeds by almost 0x2800 bytes....
Suppose for some (timing?) reason a lot of twi-functions have been put in IRAM. Can this be switched off? Otherwise the debugger won't fit even when using a very basic application...

earlephilhower:
Sorry, I don't use anything with TWI.  If they're included in IRAM it's because they need to be called during IRQs, I imagine.  If not, then they may be able to be moved out  by removing the IRAM_ATTR decorator on the function definition.

Seems like GDB wouldn't affect it either way, though, as GDB now uses WAY less than 0x2800 bytes, only around 0x600 bytes IIRC.

freeck:
I should should address my question in another thread...still very happy with GDB!

freeck:
I would like to include the debugger in the release version. But as in case of a exception the processor halts, it would be nice to have an option to disable the exception handler in the release version....
Could that be done on application level? I.c. Attach the debugger and then manually set a exception-enable-flag?

earlephilhower:
I'm not sure I understand what you're trying to do, but if you don't include gdbtub.h and do the call then the GDB stub will never be installed and the default exception handler will be called.

The simplest thing you could do is use `#ifdef RELEASE` around the `gdbstub_init()` call if that's what you want.  Or you can call gdbstub_init() somewhere else in your code, say if you get a special char sequence on the serial port.  There may be some hiccup if you're already using the serial port when this happens, but it's programmatic at that point...

freeck:
> The simplest thing you could do is use #ifdef RELEASE around the gdbstub_init()

That implies I have to build an new image and download to the target...

> Or you can call gdbstub_init() somewhere else in your code, say if you get a special char sequence on the serial port. There may be some hiccup if you're already using the serial port when this happens, but it's programmatic at that point...

OK, that's an option! But now:  when I disconnect the serial communication line, the gdbserver remains active, not? Ergo the GDB-server will be activated on an exception and the application halts. So how to detach programmatically in such cases?

earlephilhower:
Actually, scratch that.  By including a call to gdbstub_init() anywhere in the code, the linker will replace dummy no-op GDB hooks with real GDB hooks in the code, even if it doesn't get called.  So gdbstub_init() later on won't affect things, that only hooks the serial port and (I think) the breakpoint exception.

With the way it's architected now, I don't believe there's a clean, sane way to do what you're trying.

freeck:
Ok no problem, I am happy as it is.

earlephilhower:
@devyte, what do you think of getting this in 2.5.0?  It's self contained and introduces no changes in default operation (you need to mod your code to enable GDB).  It's been tested by me and @freeck with different environments (Eclipse and CLI).  Even in the worst case GDB would just not work (same as before), so it seems like only upside.

earlephilhower:
Haven't seen any problem when running from CLI (printf's seem to come out at normal speed, but I haven't run it with a stopwatch or anything).  Hitting a breakpoint and getting app as well as debug-command output seems instantaneous.  Could be some weird JAVA thing or the way Eclipse is running things (maybe single-insn stepping through all code?).

earlephilhower:
https://github.com/esp8266/Arduino/blob/a3ed4b46c73013c6f2c2f8bffb1b8fc031ff0b01/libraries/GDBStub/src/internal/gdbstub.c#L362

So only if the GDB debugger is connected will it encapsulate chars with the GDB packet.  That's just the way it has to be when GDB is connected since it's parsing everything coming over the line and would interpret bytes as garbage and barf.

At 115200 and the CLI, it's as close as I can tell to normal speed.  Maybe you're running at a low baud?

devyte:
My one comment is almost the same as @freeck 's: it would be nice to be able to de-init the gdbstub to allow normal operation. I know that calling gdbstub_init() replaces the hooks with real ones, but that shouldn't mean that the gdbstub functionality needs to be enabled all the time. In other words, it should be possible to have the gdbstub compiled into the app, then have something like a master switch that enables/disables the capability to attach gdb.
But that's a nice-to-have additional enhancement that needs to be investigated and should be handled in a different PR.

freeck:
@devyte It is indeed a nice to have. So far I am very happy with the debugger as it is.

earlephilhower:
Please, feel free to open an issue for the next release.  It's not something that's impossible, only tedious, but it will need new code and testing.  You won't get back any IRAM, of course, which was the big reason for not including GDBStub by default.

freeck:
Observation: my application relies on the softwareserial library (ver 18-12-2018). I receive 1000 bytes messages, baudrate is 115200,  length message 1000 bytes, each message ends with a crc.
Using v2.4.2 the crc-error rate was aprox. 1%
Using v2.5.0 the crc-error rate was aprox. 50%
Using this tree with/without Gdbstub the crc error rate is >90%. As you know softwareserial relies havily on handling interrupts each edge of a signal.
Any idea what might be the correlation between the upgrades of the core, and softwareserial?

freeck:
Don't you think there is a strong relation with this issue #5513 …..knowing that SW-serial does not use interrupts....

earlephilhower:
> Don't you think there is a strong relation with this issue #5513 …..knowing that SW-serial does not use interrupts....

I don't think so.  I was mistaken about SWserial.  It transmits in a busy-loop, but Rx is handled right inside a single ISR handler.

https://github.com/plerup/espsoftwareserial/blob/23ae000cb2cf4d5823a2744f6b8ae831575ff135/SoftwareSerial.cpp#L224

It doesn't return from the ISR until a full byte is shifted in....not seeing how anything at all could be affected there no matter what's done in the core.  

> I don't use these functions but they are linked into the image nevertheless....but this costs 2596 bytes of our scarce IRAM.

Open up a new issue on IRAM w/the i2c blowing up, that way it can be followed.  2.5K is a big amount...

freeck:
SWserial issue: After some experimenting I found out that I defined the size of the receive buffer too small....

> Open up a new issue on IRAM w/the i2c blowing up, that way it can be followed. 2.5K is a big amount...

I will..

freeck:
> I don't think so. I was mistaken about SWserial. It transmits in a busy-loop, but Rx is handled right inside a single ISR handler.

Ouch....This means that SWserial is busy in its ISR  during 1/115200*10 bits = 9 uSeconds! In my case this means 90% CPU is wasted! Is there a better alternative to SWserial?

Interesting observation: I found out that the interrupt handling of GDBstub influences the interrupt handling of SWserial.  After including  (and not even started) GDBstub,  SWserial doesn't work 100% reliable anymore i.e. a lot of characters are lost or corrupted....

freeck:
> I don't think so. I was mistaken about SWserial. It transmits in a busy-loop, but Rx is handled right inside a single ISR handler.

SWserial introduces a 10% CPU load in the ISR. In my case I sent 1000 chars/second, baud rate is 115200.   
I observe three things:
1: Without GDBstub installed: the error rate of the messages is aprox 25% 
2: With GDBstub installed (not started/attached): the error rate is aprox 95%. So it seems the interrupt-handlers are influencing each other...
3: Comparing v2.4.2 with v2.5.0: Using an idle-loopcounter in function loop(), v2.5.0  shows a decrease of performance of 60% (from 100000 to 40000 each second) compared to v2.4.2. (see also #5513)

PR #5563, opened 2018-12-28
Fix LeamDNS examples

hreintke:
Fix error in esp8266 hostname update
I also added the hostname to the http response so the user sees the actual one.


hreintke:
Is there a way that I can run the style checks locally before sending PR and committing to my repo ?
Then I can prevent the multiple commits.

d-a-v:
Yes, check in test/

LaborEtArs:
Thanks a lot 🙂

PR #5569, opened 2019-01-01
Split common.sh into separate scripts for each job

Androbin:
Fixes the TODO created by @igrr in #5474.
By the way, the `BUILD_TYPE` values `package`, `host` and `style` seem to be unhandled?


Androbin:
@earlephilhower Done. BTW no need to pass `cache_dir` since it's in scope anyway.

Issue #5568, opened 2018-12-31
PWM is blocked during SPIFFS write operations


Gei0r:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.  (no stack dump)
- [X] I have filled out all fields below.

#### Platform

- Hardware:					ESP-12
- Core Version:			b26c19e82e1b6c94f86a66dc2652336e8ca5bba3
- Development Env:		Platformio / Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			Wemos D1 mini
(removed rest because it's not a flashing problem)

### Problem Description

The ESP8266 does not have hardware support for PWM, so the Arduino Core implements PWM in software with timer interrupts.

However, these PWM interrupts seem to be blocked by *some* SPIFFS write operations. In the picture below, every yellow spike shows the start of a small write operation. The green channel shows a PWM at 500 Hz.

As you can see, every 16th or so write operation takes a lot longer than the other ones (about 110 ms), and disturbs the PWM output.

[![oscilloscope image][1]][1]

I understand that SPIFFS is not realtime capable and it is unavoidable that some write operations take longer than others. But maybe the SPIFFS code can be changed so that the PWM is not affected by that?

  [1]: https://i.stack.imgur.com/MIlET.png

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
#include <Arduino.h>
#include <FS.h>

void setup() {
    pinMode(14, OUTPUT);

    analogWriteRange(255);
    analogWriteFreq(500);
    analogWrite(5, 128);   // GPIO5 (pin "D1" on D1 mini)

    SPIFFS.begin();
}

String toWrite = "Hello world!";

void loop() {
    // Trigger for oscilloscope (yellow channel):
    digitalWrite(14, 1);
    delayMicroseconds(100);
    digitalWrite(14, 0);

    File f = SPIFFS.open("/test.txt", "w");
    f.write((uint8_t*)toWrite.c_str(), toWrite.length());
    f.close();
}

```

### Debug Messages

(see image)

devyte:
[...]

PR #5577, opened 2019-01-03
Remove interrupt disable around flash operations

earlephilhower:
All interrupt service routines are supposed to be in IRAM now, so there
is no need to keep interrupts disabled while doing flash operations.
Remove the IRQ disable/enable from the ESP.flash* methods.

Related to #5568


earlephilhower:
A test w/WIFI being pinged, SPIFFS reading/writing like mad, and the older, default-IRQ (non-NMI) waveform generator with this PR passed without incident.

PR #5578, opened 2019-01-03
Make waveform generator a NMI to run always, increase accuracy

### LINKED ISSUE #5568: PWM is blocked during SPIFFS write operations

earlephilhower:
Make the waveform generator an NMI using the same code as in 2.4.0.
Making it NMI will ensure it runs even when interrupts are disabled.

Fixes #5568


earlephilhower:
Ran the SpeedTest.ino on SPIFFS in PR #5511 with a 1KHz PWM output on a single pin and it's looking good.

Attempted the WiFi Server test at https://gist.github.com/arunoda/3f17eba4f3f1fd5e1a7adc86e7b62ca7 with a 1KHz output on a single pin as well, TCP v2 hi BW, 160MHz, and got about 670KByte/sec (5.3Mbit/sec) downloads from the 8266.

At 10KHz I can get occasional SPIFFS WDT timeouts when writing 512KByte at a time in 256byte chunks (but adding in yields() every 32KB fixes this while reducing write BW by under 10%, so it's not really a showstopper), but WiFi works fine and delivers ~620KBytes/sec (4.96Kbit/sec) doe to increased CPU usage by the PWM generator.

earlephilhower:
Argh, looking over it one last time I realize it's now going to be a problem as-adjusted.

Before we could twiddle with IRQ data structures as long as we bracketed it with disable-irq/.../enable-irq.

For NMI, disable/enable irq doesn't do anything.  So we could catch an interrupt while in the middle of changing values and bad stuff will happen.

Need to make it lockless somehow, may be a significant effort.

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -38,84 +38,41 @@
...
-// Set/clear *any* GPIO
-#define SetGPIOPin(a) do { if (a < 16) { GPOS |= (1<<a); } else { GP16O |= 1; } } while (0)
-#define ClearGPIOPin(a) do { if (a < 16) { GPOC |= (1<<a); } else { GP16O &= ~1; } } while (0)
-// Set/clear GPIO 0-15
+// Set/clear GPIO 0-15 by bitmask
 #define SetGPIO(a) do { GPOS = a; } while (0)
 #define ClearGPIO(a) do { GPOC = a; } while (0)
 
+// Convert from us to ESP8266 clocks
+#define MicrosecondsToCycles(m) (clockCyclesPerMicrosecond() * (m))
```
devyte: This is already defined in Arduino.h as  `clockCyclesPerMicrosecond`

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -38,84 +38,41 @@
...
-  {0, 0, 1<<10, 0, 0, 0, 0, 0},
-#endif
-  // GPIO 11 not allowed, used for flash
-  {0, 0, 1<<12, 0, 0, 0, 0, 0},
-  {0, 0, 1<<13, 0, 0, 0, 0, 0},
-  {0, 0, 1<<14, 0, 0, 0, 0, 0},
-  {0, 0, 1<<15, 0, 0, 0, 0, 0},
-  {0, 0, 0, 1, 0, 0, 0, 0}  // GPIO16
-};
+static Waveform waveform[17];        // State of all possible pins
```
devyte: The isFlashInterfacePin(9) macro no longer applies here?

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -149,122 +104,153 @@ void setTimer1Callback(uint32_t (*fn)()) {
...
+  uint32_t timeoutCycle = GetCycleCountIRQ() + MicrosecondsToCycles(14);
+
+  if (waveformToEnable || waveformToDisable) {
+    // Handle enable/disable requests from main app.
+    waveformEnabled = (waveformEnabled & ~waveformToDisable) | waveformToEnable; // Set the requested waveforms on/off
+    waveformState &= ~waveformToEnable;  // And clear the state of any just started
+    waveformToEnable = 0;
+    waveformToDisable = 0;
+    startPin = __builtin_ffs(waveformEnabled) - 1;
+    endPin = 32 - __builtin_clz(waveformEnabled);
```
devyte: I don't understand these two lines. Maybe add comments?

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -149,122 +104,153 @@ void setTimer1Callback(uint32_t (*fn)()) {
...
+      Waveform *wave = &waveform[i];
+      uint32_t now = GetCycleCountIRQ();
+
+      // Disable any waveforms that are done
+      if (wave->expiryCycle) {
+        int32_t expiryToGo = wave->expiryCycle - now;
+        if (expiryToGo < 0) {
+            // Done, remove!
+            waveformEnabled &= ~mask;
+            if (i==16) GP16O &= ~1;
```
devyte: if-else astyle
```
if(cond)
  blah
else
  bleh
```

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -149,122 +104,153 @@ void setTimer1Callback(uint32_t (*fn)()) {
...
+          else SetGPIO(mask);
+
           wave->nextServiceCycle = now + wave->nextTimeHighCycles;
           nextEventCycles = min_u32(nextEventCycles, wave->nextTimeHighCycles);
         } else {
-          ClearGPIO(wave->gpioMask);
-          if (wave->gpio16Mask) {
-            GP16O &= ~wave->gpio16Mask;
-          }
+          if (i==16) GP16O &= ~1; // GPIO16 write slow as it's RMW
```
devyte: if-else astyle

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -273,47 +259,27 @@ static ICACHE_RAM_ATTR void timer1Interrupt() {
...
-      } else {
-        uint32_t newTimeLeftCycles = wave->timeLeftCycles - deltaCycles;
-        wave->timeLeftCycles = newTimeLeftCycles;
-      }
-    }
-  }
+    uint32_t now = GetCycleCountIRQ();
+    done = false;
+    int32_t cycleDeltaNextEvent = timeoutCycle - (now + nextEventCycles);
+    if (cycleDeltaNextEvent < 0) done = true;
```
devyte: No single line ifs please

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -273,47 +259,27 @@ static ICACHE_RAM_ATTR void timer1Interrupt() {
...
+    int32_t cyclesLeftTimeout = timeoutCycle - now;
+    if (cyclesLeftTimeout < 0) done = true;
```
devyte: No single line ifs pls

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -149,122 +104,153 @@ void setTimer1Callback(uint32_t (*fn)()) {
...
       if (cyclesToGo < 0) {
-        wave->state = !wave->state;
-        if (wave->state) {
-          SetGPIO(wave->gpioMask);
-          if (wave->gpio16Mask) {
-            GP16O |= wave->gpio16Mask; // GPIO16 write slow as it's RMW
-          }
+        waveformState ^= mask;
+        if (waveformState & mask) {
+          if (i==16) GP16O |= 1; // GPIO16 write slow as it's RMW
```
devyte: if-else astyle

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -149,122 +104,153 @@ void setTimer1Callback(uint32_t (*fn)()) {
...
+  if (!(waveformEnabled & mask)) {
+    return false; //It's not running, nothing to do here
   }
-  return false;
+  waveformToDisable |= mask;
+  // Ensure tiomely service....
+  if (T1L > MicrosecondsToCycles(10)) {
+    timer1_write(MicrosecondsToCycles(10));
+  }
+  while (waveformToDisable) {
```
devyte: Does it make sense to wait with timeout?

earlephilhower: Good eyes, but at this point we know the timer is running and it has, at most, 10us of runtime before the NMI will be called.  So no need to have a timeout here.

cores/esp8266/core_esp8266_waveform.c
earlephilhower: I'm happily surprised to stump you. :) They're GCC's Find-First-Set (plus one) and Count-Leading-Zeros.  I'll add a comment.

cores/esp8266/core_esp8266_waveform.c
earlephilhower: No, that's taken care of in `startWaveform` where we return `false` if it's an invalid pin or `isFlashInterfacePin()`.  The simpler array makes for simpler (faster) code in the IRQ.

cores/esp8266/core_esp8266_waveform.c
earlephilhower: Yes, ` clockCyclesPerMicrosecond` is defined, but I need it for a fixed # of milliseconds (m) bit.

devyte: I don't follow, the Arduino.h macro looks the same as this one:
```
#define microsecondsToClockCycles(a) ( (a) * clockCyclesPerMicrosecond() )
```

Also: milliseconds? I'm confused now.

earlephilhower: Ah, darn it.  I did mean *micro*seconds and I didn't see that macro for some reason.  I'll replace the lot out...

cores/esp8266/core_esp8266_waveform.c
```diff
@@ -53,70 +50,35 @@
...
 
-static inline ICACHE_RAM_ATTR uint32_t min_u32(uint32_t a, uint32_t b) {
-  if (a < b) {
-    return a;
-  }
-  return b;
-}
+// Enable lock-free by only allowing updates to waveformState and waveformEnabled from IRQ service routine
+static volatile uint32_t waveformToEnable = 0; // Message from startWaveform to IRQ to actuall begin a wvf
+static volatile uint32_t waveformToDisable = 0; // Message from startWaveform to IRQ to actuall begin a wvf
```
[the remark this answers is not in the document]
earlephilhower: This and other commenting/if{}else{} bracketing taken care of in latest push

cores/esp8266/core_esp8266_waveform.c
earlephilhower: Done in latest push.

Jason2866:
With this change Tasmota is nearly unresponsible. Before it worked very well.

earlephilhower:
@Jason2866, please open an issue and be much more specific or we're not going to be able to help you out.  Also, are you sure it's related to this specific PR and not #5513 which was noticed at the 1st 2.5.0-dev release.   Thx, -EFP3

Jason2866:
@earlephilhower Thx for your comment. Did more testing. At the end the behaviour was a part defective Wemos Mini!! Everything is fine now :-)

davidmpye:
This is causing me a problem :-(

If you try to use something timing critical like FastLED (or Adafruit_Neopixel) *and* PWM at the same time, you run into issues.

When I write the LED data out, I need to have interrupts disabled, otherwise, the NMI firing causes the FastLED data sequence to glitch, making the LEDs flash randomly.    The fastled library suggests disabling interrupts during the write sequence, but obviously this fails if using NMI.

Is it possible to create a #define or similar to set whether the PWM code uses NMI or not?

David

earlephilhower:
@davidmpye the most compatible is going to be stopWaveform/etc. right before the code in question followed by startWaveform/etc. right after.  There may be other code in the processor (WiFi, etc.) which also comes in the binary SDK.  This isn't as bad as it sounds since the tone/analogWrite would be garbage while writing the LEDs anyway since w/IRQs disabled they won't do anything at all and will be stuck at full high or low.

Alternately, look in the core_esp8266_waveform.cpp file where it's pretty straightforward to hack code in to make it a normal IRQ (but will fail during SPIFFS writes or other periods when IRQs are disabled for any period).

PR #5590, opened 2019-01-05
Add RTC time read API

IMAN4K:
Signed-off-by: Iman iman72411@gmail.com (Github: @IMAN4K)


cores/esp8266/Esp.cpp
```diff
@@ -184,6 +184,14 @@ bool EspClass::rtcUserMemoryWrite(uint32_t offset, uint32_t *data, size_t size)
     }
 }
 
+uint64_t EspClass::getRTCTime() {
+	uint64_t cali = system_rtc_clock_cali_proc();
+	uint64_t cycles = system_get_rtc_time();
+	cali = (cali * 1000) >> 12;
+
+	return (cali * cycles) / 1000;
+}
```
d-a-v: You calculates `((((cali * 1000)/4096)*cycles)/1000)` which is `(cali * cycle) >> 12`.
Using `uint32_t` instead, I don't know whether it would be less expensive than `(((uint64_t)cali) * cycle) >> 12`.
Maybe @earlephilhower has better knowledge on this.

devyte:
Ref: 
https://github.com/espressif/ESP8266_NONOS_SDK/issues/204
#5485

cores/esp8266/Esp.cpp
devyte: @d-a-v is correct, 64bit ops are rather expensive on the ESP. That is the primary reason why [millis()](https://github.com/esp8266/Arduino/blob/8a64a1236f447009bd53fec4863aff9bf5428b59/cores/esp8266/core_esp8266_wiring.c#L65) was fixed the way it was for the drift correction some time ago.
Also, I haven't fully digested the answer given in the [nonos sdk repo](https://github.com/espressif/ESP8266_NONOS_SDK/issues/204), but is this equivalent to the info given there? (that means: this code needs a big piece of comment explaining the official answer given in the nonos repo, and how this calculation implement it)

IMAN4K: Just asking if the above multiplication will overflow over time OR not(for example if esp works for days)?
because of that, I used `uint64_t`  for the result

IMAN4K:
@devyte @d-a-v 
Good to go?

cores/esp8266/Esp.cpp
earlephilhower: This is a simple case of multiply fixed point 16b * 32b numbers and then returning the integer part.  There is no way around doing a 64b multiplication.

The math is 4p12 (calibration) * 32p0 (clock) which gives a result of 36p12 = 48 bits (drop the 12 bits after the binary point).

However, the code here does not use a 32x32->64 bit multiply (just checked the disassembly).

On the 8266 there is no 32x32->64b intrinsic multiply (only 32b*32b->32b (lowest bits)) so it is implemented by GCC as a set of 4 32x32->32 mults and shifts/adds.

Knowing this is a 16x32b multiply and needs 48b of temporary result you **might** be able to reduce it to 3 mults with shifts and adds, but I doubt it is worth it here, and you'd be doing it all on your own and probably would end up with serious bugs.

So the original code he had with 64b temp variables is the only way to get the correct answer as far as I can see.

devyte:
@IMAN4K could you please resolve conflicts so that we can take another look at this?

earlephilhower:
@IMAN4K , @devyte , my review is still relevant.  You're multiplying a fixed point 4p12 and 32p0 number, so need 48b of precision and hence `uint64_t` for the multiplicands.  W/`uint32_t`s you will silently overflow on the multiply and when you shift(divide) right, you're gonna shift in 0s and not the whole part of the fixed-point result.

PR #5595, opened 2019-01-07
Create an async version of WiFi::connect(...) method

ruggi99:
HostByName is a blocking-call function and I have rewritten the code in order to make it non-blocking. Valid for NodeMCU 1.0 as I tested it and it works.


ruggi99:
> HostByName is a blocking-call function and I have rewritten the code in order to make it non-blocking.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -569,12 +580,9 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
...
 void wifi_dns_found_callback(const char *name, CONST ip_addr_t *ipaddr, void *callback_arg)
 {
-    (void) name;
-    if (!_dns_lookup_pending) {
-        return;
-    }
+    (reinterpret_cast<struct arg_callback*>(callback_arg))->called = true;
     if(ipaddr) {
-        (*reinterpret_cast<IPAddress*>(callback_arg)) = IPAddress(ipaddr);
+        (reinterpret_cast<struct arg_callback*>(callback_arg))->addr = ipaddr->addr;
```
d-a-v: `(reinterpret_cast<struct arg_callback*>(callback_arg))->addr = addr;`
should work. `IPAddress` can take a `ip_addr_t*` and it is also IPv6 compatible.

ruggi99: In my opinion there's something wrong with your reply. In this function "callback_arg" is a pointer to a struct and you're trying to save a IpAddress into a struct pointer, instead of saving it into the "addr" member of the struct.

d-a-v: see https://github.com/esp8266/Arduino/blob/7c2e071351fa5db136b685dcc57fd30e12c451c6/cores/esp8266/IPAddress.h#L153 and https://github.com/esp8266/Arduino/blob/7c2e071351fa5db136b685dcc57fd30e12c451c6/cores/esp8266/IPAddress.h#L167 when IPv6 is enabled.

If it does not work, then we can also overload the `operator =`.

ruggi99: I modified the line. Is it OK? Or have I misunderstood your message?

d-a-v: It is OK.
More generally, you must update all of your `.addr` to use `IPAddress` instead otherwise It will not compile with IPv6.
IPAddress has lots of constructors, but maybe lacks of some overloaded `operator=`.
I'll eventually add them if needed.
Or you can, that would be adding in `IPAddress.h` L152:
```
IPAddress& operator= (const ipv4_addr& fw_addr) { setV4(); v4() = fw_addr.addr; return *this; }
IPAddress& operator= (const ipv4_addr* fw_addr) { setV4(); v4() = fw_addr->addr; return *this; }
...
#if LWIP_IPV6
...
IPAddress& operator= (const ip_addr_t& lwip_addr) { ip_addr_copy(_ip, lwip_addr); return *this; }
IPAddress& operator= (const ip_addr_t* lwip_addr) { ip_addr_copy(_ip, *lwip_addr); return *this; }
```

With them you can 
```
 if(ipaddr) {
        //(*reinterpret_cast<IPAddress*>(callback_arg)) = IPAddress(ipaddr);
        *(*reinterpret_cast<IPAddress*>(callback_arg)) = ipaddr;
  }
```

ruggi99: Ok, thank you so much

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -526,39 +529,47 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
...
-        aResult = IPAddress(&addr);
-    } else if(err == ERR_INPROGRESS) {
-        _dns_lookup_pending = true;
-        delay(timeout_ms);
-        _dns_lookup_pending = false;
-        // will return here when dns_found_callback fires
-        if(aResult.isSet()) {
-            err = ERR_OK;
-        }
+        aResult = addr.addr;
```
d-a-v: `aResult = addr` would be enough and IPv6 compatible

ruggi99:
Hi,
I have added a new function hostByNameAsync like you suggested.
Can be improved making the code more flexible, like supporting, for example, more than one DNS request at the same time. Can someone check my code? Thank you

d-a-v:
I fixed the conflicts. Style change makes it hard to read and for some reason github doesn't give for this PR the option for diffing without taking spaces into account.

ruggi99:
In my project I have some buttons attached to my nodemcu and a library (Blynk) that makes a connection automatically with blynk cloud for IoT purposes. I use polling technique for the buttons, but nodemcu blocks when HostByName is called (timeout is 10 s) and so buttons are unavailable during this time. But if I had a non blocking function like HostByNameAsync I would be able to do background stuff (polling buttons) and at the same time try to connect to the server. Blynk automatically reconnects if the connection fails, so I do not have to do cycles.

d-a-v:
Suggestions
- (you must use `strcmp()` when comparing `char*`)
Use a `String` in your structure, you never know what can happen to the pointed data. You may use `const String&` (or `String&&`) for the function parameter.
- instead of numeric constants, use enums or already existing defines like lwIP's you are already using. `err_t` provides `ERR_INPROGRESS`, `ERR_OK`, `ERR_TIMEOUT`. You could just store `gethostbyname()`'s return value in your structure and play with it ?

ruggi99:
I think I've finished.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -600,6 +600,69 @@ void wifi_dns_found_callback(const char *name, CONST ip_addr_t *ipaddr, void *ca
...
+
+int ESP8266WiFiGenericClass::hostByNameAsync(const char* aHostname, IPAddress& aResult) {
+    if (aHostname == NULL) {
+        DEBUG_WIFI_GENERIC("[hostByNameAsync] hostname is NULL!");
+        return 0;
+    }
+    static ip_addr_t addr;
+    if (!host_var.hostname.length()) {
+        host_var.hostname = aHostname;
+        err_t err = dns_gethostbyname(host_var.hostname.c_str(), &addr, &wifi_dns_found_callback_async, &host_var);
```
d-a-v: Why not `host_var.status = dns_gethostbyname...`
Because `status` is never set to `ERR_INPROGRESS` but is is needed for a test below.

(btw thanks for cleaning style and space changes)

libraries/ESP8266WiFi/examples/WiFiClientAsync/WiFiClientAsync.ino
```diff
@@ -0,0 +1,91 @@
...
+
+void loop() {
+  Serial.print("connecting to ");
+  Serial.print(host);
+  Serial.print(':');
+  Serial.println(port);
+
+  // Use WiFiClient class to create TCP connections
+  WiFiClient client;
+  if (!client.connectAsync(host, port)) { // 3rd argument to specify async mode
```
d-a-v: This call will not be blocking, but
- it does not wait,
- it does not retry,
- it doesn't show you can still do something else while waiting for the answer.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
ruggi99: Done

ruggi99:
I'm trying a new approach to create the async method. Also modified the blocking one without making use of the callback function. Could you have a look at it? Suggestions?

Issue #5555, opened 2018-12-26
Hardware Uart ISR not Shareable


Makuna:
### Problem Description

The Hardware Uart support shares a single interrupt between both Uart0 and Uart1.  Current implemementation does not allow for it to be shared so that Serial is usable at the same time that features of Uart1 can be used for other purposes, like updating NeoPixels (Ws2812 and the like).

@unaiur has created [this pull](https://github.com/esp8266/Arduino/pull/4745) in an attempt to solve the issue.  Unfortunately the Uart code has changed significiantly and this pull needs work before it can be merged.

There doesn't seem to be an original issue that pull was being applied.

Makuna:
NeoPixelBus has a branch created that uses these changes
https://github.com/Makuna/NeoPixelBus/tree/SharedUartIsr

PR #5600, opened 2019-01-07
Shared uart isr

Makuna:
Issues:
https://github.com/esp8266/Arduino/issues/5555
https://github.com/Makuna/NeoPixelBus/issues/119

Notes:
Sorry about the small format changes, but my editor when swapping tabs to spaces also does standard formating, like putting spaces after if and the like.

### DIFF
--- a/cores/esp8266/uart.c
+++ b/cores/esp8266/uart.c
@@ -72,13 +72,15 @@ struct uart_
     struct uart_rx_buffer_ * rx_buffer;
 };
 
-struct uartIntContext_t
+struct uartIsrContext_t
 {
     uartInterruptHandler callback;
     void* arg;
 };
 
-static volatile struct uartIntContext_t s_uartInterruptContext[2];
+// NOTE: GCC will generated an invalid warning for the following line
+// see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119
+static volatile struct uartIsrContext_t s_uartInterruptContext[2] = { 0 };
 
 /*
    In the context of the naming conventions in this file, "_unsafe" means two things:
@@ -91,15 +93,12 @@ static volatile struct uartIntContext_t s_uartInterruptContext[2];
    wrap the unsafe ones with disabling/enabling of the uart interrupt for safe public use.
 */
 
-
-
 // called by ISR
 inline size_t ICACHE_RAM_ATTR
 uart_rx_fifo_available(const int uart_nr)
 {
     return (USS(uart_nr) >> USRXC) & 0xFF;
 }
-
 
 /**********************************************************/
 /************ UNSAFE FUNCTIONS ****************************/
@@ -312,7 +311,8 @@ uart_isr(void* arg)
                 }
                 else
                 {
-                    // Clear all interrupts flags (just in case)
+                    // No registered handler, so
+                    // clear all interrupts flags (just in case)
                     USIE(uartNum) = 0;
                     USIC(uartNum) = 0xffff;
                 }
@@ -355,11 +355,9 @@ uart_subscribeInterrupt(int uart_nr, uartInterruptHandler callback, void* param)
     }
 }
 
-void
+bool
 uart_unsubscribeInterrupt(int uart_nr, uartInterruptHandler callback)
 {
-    ETS_UART_INTR_DISABLE();
-
     if (s_uartInterruptContext[uart_nr].callback == callback)
     {
         // turn off uart
@@ -376,12 +374,13 @@ uart_unsubscribeInterrupt(int uart_nr, uartInterruptHandler callback)
             // detach our ISR
             ETS_UART_INTR_ATTACH(NULL, NULL);
 
-            // return so we don't enable interrupts since there is no ISR anymore
-            return;
+            // return false so we don't enable interrupts 
+            // since there is no need for the ISR anymore
+            return false;
         }
     }
 
-    ETS_UART_INTR_ENABLE();
+    return true;
 }
 
 
@@ -485,7 +484,6 @@ uart_wait_tx_empty(uart_t* uart)
 
     while (uart_tx_fifo_available(uart->uart_nr) > 0)
         delay(0);
-
 }
 
 void
@@ -652,7 +650,11 @@ uart_uninit(uart_t* uart)
 
     if (uart->rx_enabled)
     {
-        uart_unsubscribeInterrupt(uart->uart_nr, uart_isrDefault);
+        ETS_UART_INTR_DISABLE();
+        if (uart_unsubscribeInterrupt(uart->uart_nr, uart_isrDefault))
+        {
+            ETS_UART_INTR_ENABLE();
+        }
         free(uart->rx_buffer->buffer);
         free(uart->rx_buffer);
     }
--- a/cores/esp8266/uart.c
+++ b/cores/esp8266/uart.c
@@ -21,25 +21,25 @@
  */
 
 
-/**
- *  UART GPIOs
- *
- * UART0 TX: 1 or 2
- * UART0 RX: 3
- *
- * UART0 SWAP TX: 15
- * UART0 SWAP RX: 13
- *
- *
- * UART1 TX: 7 (NC) or 2
- * UART1 RX: 8 (NC)
- *
- * UART1 SWAP TX: 11 (NC)
- * UART1 SWAP RX: 6 (NC)
- *
- * NC = Not Connected to Module Pads --> No Access
- *
- */
+ /**
+  *  UART GPIOs
+  *
+  * UART0 TX: 1 or 2
+  * UART0 RX: 3
+  *
+  * UART0 SWAP TX: 15
+  * UART0 SWAP RX: 13
+  *
+  *
+  * UART1 TX: 7 (NC) or 2
+  * UART1 RX: 8 (NC)
+  *
+  * UART1 SWAP TX: 11 (NC)
+  * UART1 SWAP RX: 6 (NC)
+  *
+  * NC = Not Connected to Module Pads --> No Access
+  *
+  */
 #include "Arduino.h"
 #include <pgmspace.h>
 #include "uart.h"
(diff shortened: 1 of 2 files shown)


cores/esp8266/uart.h
```diff
@@ -110,43 +110,52 @@ extern "C" {
...
-int uart_detect_baudrate(int uart_nr);
+    void uart_set_debug(int uart_nr);
+    int uart_get_debug();
 
+    void uart_start_detect_baudrate(int uart_nr);
+    int uart_detect_baudrate(int uart_nr);
+
+    // uart_subscribeInterrupt & uart_unsubscribeInterrupt are not safe and must
+    // be called within ETS_UART_INTR_DISABLE()/ETS_UART_INTR_ENABLE() protection
+    //
```
d-a-v: Other unsafe functions have `_unsafe_` part of their name.
I think that is a good reminder for their users.
Or, why not calling INTR_EN/DISABLE() from inside these functions and remove that warning ?

cores/esp8266/uart.c
```diff
@@ -283,42 +289,105 @@ uart_resize_rx_buffer(uart_t* uart, size_t new_size)
...
+void
+uart_subscribeInterrupt(int uart_nr, uartInterruptHandler callback, void* param)
+{
+    s_uartInterruptContext[uart_nr].callback = callback;
+    s_uartInterruptContext[uart_nr].arg = param;
+
+    int other_nr = (uart_nr + 1) % 2;
+    if (s_uartInterruptContext[other_nr].callback == NULL)
+    {
+        ETS_UART_INTR_ATTACH(uart_isr, s_uartInterruptContext);
```
d-a-v: Maybe some comments like
`// attach interrupt if not already attached by the other port (1-uart_nr)`

Makuna: will do

cores/esp8266/uart.c
```diff
@@ -72,6 +72,15 @@ struct uart_
     struct uart_rx_buffer_ * rx_buffer;
 };
 
+struct uartIsrContext_t
+{
+    uartInterruptHandler callback;
+    void* arg;
+};
+
+// NOTE: GCC will generated an invalid warning for the following line
```
Makuna: This is causing the build failures. I will be wrapping it with a pragma ignore like...
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmissing-braces"
    static volatile struct uartIsrContext_t s_uartInterruptContext[2] = { 0 };
#pragma GCC diagnostic pop
```

cores/esp8266/uart.h
d-a-v: - Rename them _unsafe, add new ones calling the unsafe version in a INTR_EN/DISABLE sandwhich

- too bad it's C, otherwise a parameter `bool protect` with a default value would do the job

- keep as it is, and let's rely on users' intelligence for reading doc and comments

devyte: Please implement like bullet 1 above. The unsafe word in the function name is meant to indicate it's not interrupt-safe and it shouldn't be exported, if at all possible.
The unsafe functions are meant for resuse from other unsafe functions and/or for sandwich wrapping into safe versions.

cores/esp8266/uart.c
```diff
@@ -72,6 +72,18 @@ struct uart_
...
+// see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-braces"
+static volatile struct uartIsrContext_t s_uartInterruptContext[2] = { 0 };
```
earlephilhower: You can remove all these pragmas by simply dropping `{0}`.  It's a global static, so is 0-initted before the app begins.

Makuna: They are not guaranteed to be initialized, it based on the flags used to compile.  So it wouldn't be safe to do that.
Consensus from my searches is to use the pragma until/if it gets fixed.

earlephilhower: It's guaranteed on any working C compiler (this is a *.c file).  There might be something weird in C++, but that's not material in this file.  The only concern would be if your code comes before the main C startup routines, (i.e. in early_preinit() callbacks/etc.).  You can verify by looking at the .MAP file and see that it's in the bss segment.

igrr: @earlephilhower is right here — you can drop the initializer.

Makuna: fixed

cores/esp8266/uart.c
```diff
@@ -283,42 +287,109 @@ uart_resize_rx_buffer(uart_t* uart, size_t new_size)
...
+{
+    s_uartInterruptContext[uart_nr].callback = callback;
+    s_uartInterruptContext[uart_nr].arg = param;
+
+    // check if we are already attached to the interrupt for
+    // the other uart and only attach if not
+    int other_nr = (uart_nr + 1) % 2;
+    if (s_uartInterruptContext[other_nr].callback == NULL)
+    {
+        ETS_UART_INTR_ATTACH(uart_isr, s_uartInterruptContext);
```
devyte: In uart_start_isr(), the current code calls ETS_UART_INTR_ATTACH() and then ETS_UART_INTR_ENABLE() right after. This new code does not.

cores/esp8266/uart.h
devyte: This is not quite resolved. I'll try to explain again.
I said:
> The unsafe word in the function name is meant to indicate it's not interrupt-safe and it shouldn't be exported

In this case, there could be undefined behavior if there is an IRQ in the middle of calling these functions.
Instead, mark the current implementations of these functions as static, and implement new functions without the _unsafe() word. Then export those in the .h. These exported versions call the _unsafe() functions inside a sandwich.

devyte: Also, it is ok to skip argument sanity checks in the _unsafe functions, but the safe versions should implement checks on the received args.

Makuna: @devyte Maybe I was not clear enough.  You can't wrap the subscribe as the site of where you call it has to be already wrapped with it due to several processes of setting up the uart hardware are required to be called within the same DISABLE/ENABLE.  This is what I was pointing out by referencing the call spots withing UART.C.  This same setup; but different based on needs of the caller is done externally by anyone needing to use uart ISR.
IF DISABLE/ENABLE were actually ref-count like entities, then it would be a good practice, as the ENABLE inside would only ref-count and not truly enable; but this is not the case.

cores/esp8266/uart.c
```diff
@@ -283,42 +287,109 @@ uart_resize_rx_buffer(uart_t* uart, size_t new_size)
...
+}
 
-    if(usis & (1 << UIFF))
+void ICACHE_RAM_ATTR
+uart_isrDefault(void * arg)
+{
+    uart_t* uart = (uart_t*)arg;
+    uint32_t usis = USIS(uart->uart_nr);
+
+    if (usis & (1 << UIFF))
```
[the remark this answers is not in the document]
Makuna: We have ASSERTs that turn off in a normal build?
I would agree with an ASSERT, but I disagree otherwise.   This is an internal handler and not exposed, and all locations include an object.  Param checks should be reserved for the outermost APIs.   If hit such a check, then there is corruption.

cores/esp8266/uart.c
Makuna: See line 619, this is where it is enabled.

cores/esp8266/uart.c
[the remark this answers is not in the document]
devyte: As I said, _some_ sanity check. An assert is ok by me, as long as it's clear what it checks (i.e.: include comments).

cores/esp8266/uart.c
devyte: Ah, I had missed the significance of that, the disable-enable sandwich is moved there.
In that case, it seems to me that the semantics of uart_start_isr() changes, given that the isr isn't really started. I suggest a a change in the function name, and a comment for the function explaining the new behavior (i.e.: inits this, prepares that, blah, does not enable the interrupt yet, etc)

cores/esp8266/uart.h
Makuna: YES, these two unsafe functions are called by the user.  I pointed to my project that consumes these in the pull comments at the top.

Can I just point them to my code?  Ok, its not simple ;-)

The real use of this is where you aren't doing simple serial, otherwise you would just use Serial.  In general, using the hardware to talk to devices that take a serial stream; often where translation is done from data on the MCU as you send.  Which means configuring the UART to your custom needs (using all those esoteric features to invert, turn off stop bits, etc) that are not exposed by Serial or even the current uart.h.

cores/esp8266/uart.c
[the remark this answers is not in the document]
Makuna: BTW, the original check here was invalid.  It would have dereferenced the null on lines above this check, this is why it was removed.

cores/esp8266/uart.h
Makuna: Here is a link to my code that does this specifically
https://github.com/Makuna/NeoPixelBus/blob/959cfac07b97e95e10fc71f311f81fa3e2757895/src/internal/NeoEsp8266UartMethod.cpp#L75

devyte:
@Makuna can you please take another look at this now? There are conflicts that need to be resolved, and some of your code could need adapting.
BTW, the conflicts are due to #5559 , which adds support for debugging with gdb while sharing the hw uart with the app. I'm not sure how/if that impacts the idea here.

Makuna:
I will need to rewrite this some to address all the changes made on the other pull.  Ignore this for now and once I create the new one I will close this.

earlephilhower:

Makuna:
Last time I looked, the changes that were taken made my changes/solution IMPOSSIBLE.  I really felt the debug stuff should not have gone in, it had inverted API use at the minimum and lots of poor assumptions.  I felt rather put down since we had two pushes BEFORE it and we were the ones who had to react; TWO TIMES already.

Was that fixed?

devyte:
@Makuna I don't remember the status anymore, I got sidetracked too many times with stability pursuits. Can you please handle this so that we can look at it? We're currently trying to recover proposals that got put aside for various reasons.
If this is dependent on something else, please explain.

PR #5601, opened 2019-01-08
generic esp8266 board: Use flash mode DOUT by default, in place of QIO

d-a-v:


igrr:
@d-a-v Which issue does this fix?

d-a-v:
https://gitter.im/esp8266/Arduino?at=5c33150782a6c30b9099abb3
@makuna

also https://gitter.im/esp8266/arduino-maintainers?at=5c34059f357dd7655d334bca

igrr:
Okay, I think this is fine, if prominently mentioned in the release notes.

igrr:
Side note, does the new menu item order DOUT - QIO - QOUT - DIO look logical? Maybe DOUT - DIO - QOUT - QIO? Or DOUT - QOUT - DIO - QIO?

d-a-v:
> Side note, does the new menu item order DOUT - QIO - QOUT - DIO look logical? Maybe DOUT - DIO - QOUT - QIO? Or DOUT - QOUT - DIO - QIO?

TBH I don't know.
The most compatible is DOUT, position 1, purpose of this PR.
Previous first, the most simultaneously common and fastest is QIO, position 2 ?

We can also add comments in menus with those options, could be
```
1- DOUT: most compatible
2- QIO: fastest
3- QIO: ????
4- DIO: ????
```

This PR is not mandatory and does not fix anything apart from common user error when using a generic board. It arose from the above link when it comes to sonoff devices, and they seem to not be alone, which seem to have only two wires between esp8266 and flash chip (or using a specific flash chip, I don't know).

We can also add defines that can be checked in sketches and #warn if configuration does not match something (when sketch can be matched to specific hardware).
```
#ifndef FLASH_MODE_DOUT
#error Please use DOUT flash mode for this board
#endif 
```

I'm open for updates, or closing.

earlephilhower:
@d-a-v, sorry but I don't have measured numbers, and it depends on the cache hit ratio.  Assuming the cache line is 32bytes, though, the time to do a cache fill is going to be very roughly (I'm not looking at a datasheet now or factoring protocol overhead):

DOUT: (write addr) 24clks (read 32*8bits @ 2-bits/clk) 128clks = 152 Flash Clocks @ 40MHz = 3800 ns
DIO: (write addr) 12 clks (read 32*8bits @ 2-bits/clk) 128 clks = 140 Flash Clocks @ 40MHz = 3500 ns
QOUT: write(addr) 24clks (read 32*8bits @ 4bits/clk) 64 clks = 88 Flash Clocks @ 40MHz = 2200 ns
QIO: (write addr) 6 clks (read 32*8 @ 4clks/bit) 64 clks = 70 Flash Clocks @ 40MHz = 1750 ns

Straight-line code in this case (not in cache yet) would end up almost 50% slower on DOUT vs QIO.  

That said, though, if you're using "generic" as the board then I'm all for a safe (but slow) default.  You can change it to QIO/etc. manually if you know enough about the board (or add it to boards.py).

tools/boards.txt.py
```diff
@@ -947,14 +947,14 @@
     ####################### menu.FlashMode
 
     'flashmode_menu': collections.OrderedDict([
+        ( '.menu.FlashMode.dout', 'DOUT' ),
+        ( '.menu.FlashMode.dout.build.flash_mode', 'dout' ),
         ( '.menu.FlashMode.qio', 'QIO' ),
```
earlephilhower: Would suggest @igrr's ordering here as it is intuitive:  DOUT/DIO/QOUT/QIO.  Most compatible, slowest to least compatible, fastest...

devyte: Given numbers mentioned somewhere, I agree with that order 😄

d-a-v: Order fixed
performance indication in menus
and defines/macro available if needed

Issue #5583, opened 2019-01-04
Folder of SoftwareSerial are empty in Windows 10


lrmoreno007:
### Basic Infos

- [X ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ X] I have tested that the issue is present in current master branch (aka latest git).
- [ X] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ X] I have filled out all fields below.

#### Platform

- Hardware:					[NodeMCU v.1]
- Core Version:			[latest git]
- Development Env:		[Arduino IDE]
- Operating System:		[Windows 10 1809]

### Settings in IDE

- Module:  			[Nodemcu]
- lwip Variant:   [v2 Higher Bandwidth]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA]

### Problem Description

Folder esp8266com\esp8266\libraries\SoftwareSerial are empty in Windows 10.

I receive the following message in Arduino IDE:
Librería inválida encontrada en C:\Users\XXXXXX\Documents\Arduino\hardware\esp8266com\esp8266\libraries\SoftwareSerial: no headers files (.h) found in C:\Users\XXXXXX\Documents\Arduino\hardware\esp8266com\esp8266\libraries\SoftwareSerial

I inspect the github repository and found that the folder is called: "SoftwareSerial @ 23ae000" and isn't empty.

Pablo2048:
It's because the SoftwareSerial is git submodule see https://github.com/esp8266/Arduino/issues/3880 and https://github.com/esp8266/Arduino/issues/3596

devyte:
Note: need to update git install doc

lrmoreno007:
> 
> 
@Pablo2048  Thanks!! it's solved with: `git submodule update --init --recursive`

I update:
https://github.com/lrmoreno007/ESP-Easy-Updater-for-Windows

@devyte yes, update doc is needed.

PR #5602, opened 2019-01-08
Add fix for SoftwareSerial submodule errors to documentation

phrxmd:
Add a notice on how to initialize git submodules to the installation instructions.
This fixes the messages that people have been getting about missing files for the ``SoftwareSerial`` submodule, e.g. #5583.


doc/installing.rst
```diff
@@ -112,6 +112,16 @@ Instructions - Windows 10
 
 - If using the Arduino IDE for Visual Studio (https://www.visualmicro.com/), be sure to click Tools - Visual Micro - Rescan Toolchains and Libraries 
 
+-  You may see some error messages about missing files related to the
```
devyte: I think the instructions should go further up, right after Download binary tools. Also, I suggest rephrasing in a more straightforward manner, something along the lines of:

- Initialize the submodules:
cd blah
git submodule update --init

If error messages about missing files related to the SoftwareSerial lib are encountered, it's because .... blah ... To get rid of the error messages, this step is mandatory.

doc/installing.rst
```diff
@@ -174,3 +184,13 @@ Instructions - Other OS
        python get.py
 
 -  Restart Arduino
+
+-  You may see some error messages about missing files related to the
```
devyte: Same as previous case, these should be right before Restart Arduino.

KristiyanFxy:
@devyte, thank you for updating the doc as I got bitten by "no header files found in..." error. That said, the git command has an extra character

git submodule**s** update --init

PR #5607, opened 2019-01-11
Release referenced resources in the destructor for ESP8266SSDP

mcbp223:
I need this to resolve a crash when I dynamically turn WiFi off, with the SSDP destructor called before WiFi.disconnect().


libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -150,7 +150,30 @@ SSDPClass::SSDPClass() :
 }
 
 SSDPClass::~SSDPClass() {
-  delete _timer;
+#ifdef DEBUG_SSDP
```
devyte: 1. Instead of putting this directly in the destructor, I suggest implementing the cleanup in a standalone method, e.g.: close(), then calling that method from the destructor. To be specific, the cleanup done in the destructor should reflect the resource acquisition in the constructors + the call to close(), while the cleanup in close() should reflect the acquisition done in begin().
2. close() should be called at the start of begin() to allow calling begin() repeatedly.
3. There is some duplication between the implemented cleanup and the begin() method body. That should be cleaned up.
4. strings should be moved to flash, even (especially?) DEBUG ones.

d-a-v:
> Regarding those, I am using path shortening in one of my projects
> let me know if you want me to make a PR with this change in debug.h.

Yes !

mcbp223:
@d-a-v Now I like more the [solution](https://github.com/esp8266/Arduino/issues/4520#issuecomment-373388954) with relative paths, or anything with gcc support as commented later in the thread, since what I've been using is C++ only, and parameter pack metaprogramming hurts your eyes. But for tests, with the following code I get no more _/home_ occurrences inside the _.ino.bin_ file.
```cpp
#include "Indexes.h"

#ifdef __cplusplus

template <size_t N>
inline constexpr int lastIndexOfSep(const char (&str)[N], int crt) {
	return crt < 0 ? -1
	               : str[crt] == '/' || str[crt] == '\\'
	                   ? crt
	                   : lastIndexOfSep(str, crt - 1);
}

template <size_t N>
inline constexpr int lastIndexOfSep(const char (&str)[N]) {
	return lastIndexOfSep(str, N - 2); // allow for one character filenames
}

template <size_t N_LONG, int LAST_SEP
        , typename Indexes = typename andrivet::ADVobfuscator::Make_Indexes<N_LONG - (LAST_SEP + 1)>::type>
struct FileNameShortener;

  // Partial specialization, used for indices argument deduction
template<size_t N_LONG, int LAST_SEP, int... I>
struct FileNameShortener<N_LONG, LAST_SEP, andrivet::ADVobfuscator::Indexes<I...>> {
	using storage_type = char[sizeof...(I)];
	storage_type buffer_;

	constexpr FileNameShortener(const char (&str)[N_LONG])
		: buffer_ { str[I + (LAST_SEP + 1)]... } {}
};

#ifndef PROGMEM_G
// per solution in https://github.com/esp8266/Arduino/issues/3369
#define PROGMEM_G __attribute__((section(".irom.text.group")))
#endif

#define __SHORT_FILE__ (__extension__({static constexpr FileNameShortener<sizeof(__FILE__), lastIndexOfSep(__FILE__)> PROGMEM_G sf__(__FILE__); FPSTR(sf__.buffer_);}))
#else // not C++
#define __SHORT_FILE__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
//#define __SHORT_FILE__ __PRETTY_FUNCTION__
//#define __SHORT_FILE__ __FUNCTION__
#endif
```

d-a-v:
> and parameter pack metaprogramming hurts your eyes

Show me where I wrote that

I like your snippet. While c++ constexpr functions are elegant, are they really needed regarding the simplicity of the c-only define which is also c++ compatible ?

> Now I like more the solution with relative paths

It seems easier to replace `__FILE__` by `__SHORT_FILE__` than to patch gcc no ?
(and according to @earlephilhower these patches are rather hairy and now unmaintained)

mcbp223:
> While c++ constexpr functions are elegant, are they really needed regarding the simplicity of the c-only define which is also c++ compatible 

As far as I understand, the name of the game here is to convince the compiler to put in the binary the short file name _only_, for several reasons - to reduce memory consumption, privacy purposes, be able to produce the same exact binary on different machines with different path arrangements. To cut the filename short you'd have to execute a compile-time processing, and `constexpr` is designed to require the compiler to do a computation at compile-time, instead of at run-time. Also, without the `int... I` parameter pack gymnastics I wasn't able to completely remove the full filename from the output, no matter what else I tried. Too bad I can't use this in the C parts of the program though.

In the solution I've linked to, @Eszartek didn't modify gcc, but only a config file in his development chain. You seemed to like his method too, in a comment you've made in the thread.

d-a-v:
Ah! Of course you are right.
I don't know why I was thinking `__builtin_strrchr` was called at compile time...

earlephilhower:
@d-a-v, It's a 1-line change to convert GCC to only use the filename everywhere `__FILE__` is used while leaving debug info untouched as far as I can see.
````
diff --git a/gcc/gimplify.c b/gcc/gimplify.c
index e7119283ff3..ce9678dec13 100644
--- a/gcc/gimplify.c
+++ b/gcc/gimplify.c
@@ -2562,7 +2562,14 @@ gimplify_call_expr (tree *expr_p, gimple_seq *pre_p, bool want_value)
       case BUILT_IN_FILE:
        {
          expanded_location loc = expand_location (EXPR_LOCATION (*expr_p));
-         *expr_p = build_string_literal (strlen (loc.file) + 1, loc.file);
+         char *fname;
+#if defined(_WIN32)
+         fname = strrchr(loc.file, '\\');
+#else
+         fname = strrchr(log.fine, '/');
+#endif
+         if (!fname) fname = loc.file;
+         *expr_p = build_string_literal (strlen (fname) + 1, fname);
          return GS_OK;
        }
       case BUILT_IN_FUNCTION:
````

The uncommitted GCC patches allow much more granular transformations, not just eliminations, of multiple paths.

This is a neat discussion but not really related to SSDP.  Maybe jump to #4520 if you want to get into it, for everyone's sanity?

earlephilhower:
> @d-a-v, It's a 1-line change to convert GCC to only use the filename everywhere `__FILE__` is used while leaving debug info untouched as far as I can see.
....

After actually trying this patch I can see the lines changed don't actually get called on `__FILE__` expansion somehow.  Too bad, it looked so simple and clean (famous last words). 😞

PR #5611, opened 2019-01-13
LEAmDNS fixes 1

LaborEtArs:
- Better separation of ESP wifi thread code from user thread code
- Added a flag for 'update()'-less use (disabled by default)
- The too fast updates for service queries are fixed
- Switched fully to PolledTimeout; LEATimeFlag not needed anymore (BTW: a const 'expired()' method would be helpful)
- The device should stay visible now even after the first TTL timeout
- Improved service querying (queries five times now)

### DIFF
--- a/LEAmDNS.cpp
+++ b/LEAmDNS.cpp
@@ -1,0 +1,1130 @@
+/*
+ * LEAmDNS.cpp
+ *
+ *  License (MIT license):
+ *    Permission is hereby granted, free of charge, to any person obtaining a copy
+ *    of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


devyte:
Something is strange in this PR. I see brand new files, and no changes at all to existing files.
Maybe the files are in the wrong dir?

d-a-v:
Indeed. @LaborEtArs Can you "git move" your files into `src/` and repush ?

LaborEtArs:
Can do and would love to do, but haven’t got a clue, what your talking about (at least don’t know how to move files via git) ☹️

d-a-v:
closed in favor of #5619

PR #5631, opened 2019-01-18
Add fix to ESP8266HTTPClient sendHeader()

destotelhorus:
Fixed a part of code that lead - for me - to successive HTTP-Requests failing.
Logfile for first request read in apache like:
"GET /smtdata.php HTTP/1.1"
but after the first request turned deterministicly into:
"GET/smtdata.php HTTP/1.1"

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -1085,7 +1085,7 @@ bool HTTPClient::sendHeader(const char * type)
         return false;
     }
 
-    String header = String(type) + " " + (_uri.length() ? _uri : F("/")) + F(" HTTP/1.");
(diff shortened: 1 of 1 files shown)


destotelhorus:
ci checks failed due to http error 500 which is unrelated to this pull request.

devyte:
@destotelhorus the code change doesn't make sense to me. If String isn't concatenating that " " then there is something else wrong. Whatever else, your code change shouldn't make a difference in the resulting concatenated string.
Can you please construct a MCVE sketch that shows the issue you see? Emphasis on the "M", as in minimal.

destotelhorus:
@devyte I did some further tests and it seems to me that the situation I am facing is due to some memory issue of another kind.

I do have a char-array (that has nothing to do with the httpclient) in my application, and once that goes over a certain amount of bytes, the http GET requests fail again due to strings not containing the data they should.
I guess my checkin only solved it briefly and for my situation.

PR #5641, opened 2019-01-19
LEAmDNS Fixes 1.2

LaborEtArs:
- The removal of the cache flush bit from the whole query response (Fixes 1.1) was a bit too much... Removed the cache flush now from PTR TYPE and PTR NAME answers only.

- Reenabled the 'Unicast Response' flag for probes


LaborEtArs:
Added a flag to avoid multiple initialization

LaborEtArs:
Flag free 'multi init stopper' now :-)
@devyte: Thanks for making me think ;-)

PR #5651, opened 2019-01-22
Release buffer/cache on SPIFFS.end()

### LINKED ISSUE #4585: SPIFFS.end() does not free memory
When do SPIFFS.begin() it will consume some like 2kB of memory, but SPIFFS.end() will not release it. That is kinda problem for low-memory environment as for my application these 2kB is difference between SSL handshake can be done later or not.

Tested SDK: 2.3.0, 2.4.1 from Arduino IDE
Tested ESP_12E

Test code:
```
// Filesystem
#include "FS.h"

void printHeap() {
  Serial.print(F("Heap size: "));
  Serial.println(ESP.getFreeHeap());
}

void setup() {
  Serial.begin(115200);
  delay(1000);
  Serial.println();
  Serial.println("Setup begin");
  Serial.print(F("SDK: ")); Serial.println(ESP.getSdkVersion());
  Serial.print(F("Core: ")); Serial.println(ESP.getCoreVersion());
  
  printHeap();
  
  Serial.println("SPIFF BEGIN");
  SPIFFS.begin();
  printHeap();

  Serial.println("SPIFF END");
  SPIFFS.end();
  printHeap();

  Serial.println("Setup end");
}

void loop() {
  Serial.println("Loooop");
  Serial.println("SPIFF BEGIN");
  SPIFFS.begin();
  printHeap();
  
  delay(1000);
  
  Serial.println("SPIFF END");
  SPIFFS.end();
  printHeap();

}
```

Sample output:
```
Setup begin
SDK: 1.5.3(aec24ac9)
Core: 2_3_0
Heap size: 47072
SPIFF BEGIN
Heap size: 44960
SPIFF END
Heap size: 44960
Setup end
Loooop
SPIFF BEGIN
Heap size: 44960

Setup begin
SDK: 2.2.1(cfd48f3)
Core: 2_4_1
Heap size: 47160
SPIFF BEGIN
Heap size: 44984
SPIFF END
Heap size: 44984
Setup end
Loooop
SPIFF BEGIN
Heap size: 44984
SPIFF END
Heap size: 44296
Loooop
SPIFF BEGIN
Heap size: 44296
SPIFF END
Heap size: 44224
```

earlephilhower:
Releases the buffers allocated by the SPIFFS object when/if SPIFFS::end()
is called.

Fixes #4585


cores/esp8266/spiffs_api.h
```diff
@@ -154,6 +154,9 @@ class SPIFFSImpl : public FSImpl
             return;
         }
         SPIFFS_unmount(&_fs);
+        delete[] _workBuf.release();
```
devyte: Why not do _workBuf.reset(nullptr), which is preferred?

earlephilhower: My understanding is that reset(x) does not implicitly call delete[], only delete.  https://en.cppreference.com/w/cpp/memory/unique_ptr/reset .

devyte: The type of _workBuf is the specialization for dynamic arrays `std::unique_ptr<uint8_t[]>`, and so the default deleter object is `delete[]`. This is actually why, if you're going to delete manually, it is preferred to use `_workBuf.get_deleter()(_workBuf.release())`, at least for normal usage of unique_ptr.

The case that is dangerous is when you misuse unique_ptr, like when you don't use the specialization, e.g.:
```
uint8_t *raw = new uint8_t[256];
std::unique_ptr<uint8_t> buf(&raw[0]); //gain ownership
buf.reset(); //calls delete instead of delete[] => mem leak
```
in which case you *do* want to do:
`delete[] buf.release();`

About reset():
> 4) In the specialization for dynamic arrays, std::unique_ptr<T[]>, this overload is necessary to allow reset to nullptr (which would otherwise be prohibited by the template overload). Equivalent to reset(pointer())

The specialization is provided for dynamic arrays for the case when you pass in nullptr as well. As of C++17, you can even omit nullptr, as it's defaulted if you call without args. 
Per my understanding, the reason that this is needed is that the dynamic array type is T[], but you're passing in nullptr, which is not arrayed, so a naive implementation of unique_ptr would not allow nullptr as arg (different type, notice the deleted reset(U) which is disallowed because the arg type is U and not U[]).

If you want to put the doubt gremlins to rest, try a small test with the above and run it through valgrind. There's always the chance that I'm wrong 😄

PR #5652, opened 2019-01-23
dynamic WiFi.hostname("newname")

d-a-v:
Effect of `WiFi.hostname("new name")` on the network:

hostname

My AP runs `openwrt` with `dnsmasq`, and I can ping the new hostname as soon as `Wifi.hostname("name")` is called on the esp.

Sketch:
```

#include <ESP8266WiFi.h>
#include <PolledTimeout.h>

#ifndef STASSID
#define STASSID "your-ssid"
#define STAPSK  "your-password"
#endif

const char* ssid     = STASSID;

void connect ()
{
  WiFi.mode(WIFI_STA);
  WiFi.begin(ssid, password);

    delay(500);
    Serial.print(".");
  }

  Serial.println("");
  Serial.println("WiFi connected");
  Serial.println("IP address: ");
}

void setup() {
  Serial.begin(115200);
  Serial.println();
  Serial.println();
  Serial.print("Connecting to ");
  Serial.println(ssid);

  WiFi.persistent(false);
  connect();
  WiFi.hostname("chameleon");
  Serial.printf("host name is %s, will change every 20 seconds, ping me !\n", WiFi.hostname().c_str());
}

esp8266::polledTimeout::periodic renew(20000);
int count = 0;

void loop()
{
  if (renew)
  {
    String name = String(++count) + "chameleonchameleonchameleon";
    Serial.printf("setting hostname to: %s\n", name.c_str());
    Serial.printf("ret=%d\n", WiFi.hostname(name));
    Serial.printf("hostname=%s\n", WiFi.hostname().c_str());
  }
}
```

Logs:
```
...
setting hostname to: 17chameleonchameleonchameleon
hostname '17chameleonchameleonchameleon' is not compliant with RFC952
ret=0
...
```
(but esp name stays "nicely" responding to name `17chameleonchameleonchameleon`)

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -467,7 +467,7 @@ IPAddress ESP8266WiFiSTAClass::dnsIP(uint8_t dns_no) {
  * Get ESP8266 station DHCP hostname
  * @return hostname
  */
-const char* ESP8266WiFiSTAClass::hostname(void) {
+String ESP8266WiFiSTAClass::hostname(void) {
     return wifi_station_get_hostname();
 }
 
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -37,6 +37,7 @@
 extern "C" {
 #include "lwip/err.h"
 #include "lwip/dns.h"
+#include "lwip/dhcp.h"
 #include "lwip/init.h" // LWIP_VERSION_
 #if LWIP_IPV6
 #include "lwip/netif.h" // struct netif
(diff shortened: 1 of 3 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -467,38 +468,91 @@ IPAddress ESP8266WiFiSTAClass::dnsIP(uint8_t dns_no) {
...
-bool ESP8266WiFiSTAClass::hostname(const String& aHostname) {
-    return hostname((char*) aHostname.c_str());
+    if (aHostname.length() > 24) {
+        // nonos-sdk allows 32, but
+        // dhcp server may require RFC compliance
+        aHostname.remove(24);
+    }
+    for (size_t i = 0; i < aHostname.length(); i++)
+        // replace unallowed chars by dashes
+        if (!isalnum(aHostname[i]) && aHostname[i] != '-')
```
devyte: What is the reasoning behind silently changing the given hostname string?
How about pulling this first part of the code out and splitting it into two functions:
1. validateHostname() which takes a hostname string and returns true if the string complies with the RFC limitations, or false if not. Call this from WiFi.hostname(), and return false if validation fails.
2. fixHostname() which fixes a hostname string by replacing invalid chars, truncating to 24chars, etc. Make this function public. If the user wants to have a "soft" behavior, where the argument gets auto-fixed, then he can do something like WiFi.hostname(fixHostname(myhostname)).

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -467,38 +468,91 @@ IPAddress ESP8266WiFiSTAClass::dnsIP(uint8_t dns_no) {
...
+            // renew already started DHCP leases
+            err_t lwipret = dhcp_renew(intf);
+            if (lwipret != ERR_OK) {
+#ifdef DEBUG_ESP_CORE
+                static const char fmt[] PROGMEM = "WiFi.hostname(%s): lwIP error %d on interface %c%c (index %d)\n";
+                DEBUGV(fmt, intf->hostname, (int)lwipret, intf->name[0], intf->name[1], intf->num);
+#endif
+                ret = false;
+            }
+        }
```
devyte: How about an else here to cover the case of static IP setup, where dhcp_inform() is called?
Does dhcp_inform() work instead of dhcp_renew() in the dhcp case?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -467,38 +468,91 @@ IPAddress ESP8266WiFiSTAClass::dnsIP(uint8_t dns_no) {
...
  * Set ESP8266 station DHCP hostname
- * @param aHostname max length:32
+ * @param aHostname max length:24
  * @return ok
  */
-bool ESP8266WiFiSTAClass::hostname(char* aHostname) {
-    if(strlen(aHostname) > 32) {
+bool ESP8266WiFiSTAClass::hostname(String aHostname) {
+  /*
+  vvvv RFC952 vvvv
```
devyte: Really nice comment!

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
d-a-v: I too have thought to split this checking/updating out.

Users usually do not check `WiFi.hostname()` return value, on purpose, because:
- sketches would be unnecessarily complex just to take actions in case this sole function would be failing,
- there's not much to do from the sketch point of view when the API refuses this request,
- there's no easy way to verify network servers did acknowledge this request (on some dhcp servers, this hostname request is simply ignored, that's one of the reasons why they made mDNS)

In the end, I took the efficient way:
- arduino simplicity
- if a user does not know this RFC, we just stuff it to them

I'm open to any counter-argument :)

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
d-a-v: We can't, dhcp servers have absolutely no knowledge on static IPs originating from devices.
On static leases however, this PRs still applies.
This dhcp option "hostname" is not supposed to modify DNSs. It works when your DNS and DHCP servers talk to each other (like dnsmasq offering both services). Otherwise, DNS servers cannot be aware of these dhcp names.

devyte: > dhcp servers have absolutely no knowledge on static IPs originating from devices 
My understanding, both from a quick overview of the dhcp inform message, and of the dhcp_inform() lwip function comment, is that this is precisely the reason for sending the message: it lets local dhcp servers know that there is a static IP in use.
A search online didn't give me much details about how dhcp_inform() is supposed to work, but I did find one hint about this use case, and it's as you said: the default behavior for dhcp servers is to ignore all inform messages. Therefore, this isn't useful.

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
devyte: My thinking is precisely what you said: simplicity, and the RFC, i.e.: if the hostname doesn't meet the RFC requirements, then do nothing and just fail with return false. If anyone doesn't check the return value, and/or doesn't understand why it doesn't work, we direct the person to the RFC. 
It's a bit like trying to set up a wifi connection with wpa2, and a password that is too short. 

What I'm suggesting is to just remove the code that silently replaces illegal characters with '-', and do a simple validation instead.

d-a-v: > It's a bit like trying to set up a wifi connection with wpa2, and a password that is too short.

That's different, in that case, the whole sketch connectivity won't work. An `hostname(name)` failing does not prevent the sketch working and being accessible through its IP.

> if the hostname doesn't meet the RFC requirements, then do nothing and just fail with return false

I agree on that we can have silently an unexpected hostname result with this PR.

I would propose something in the middle, which is returning false when the name is not compliant (without anymore name fixing, with debug message when they are enabled), but still using this name as we were before, because dhcp servers are flexible with this RFC restriction.

devyte: Agreed!

PR #5689, opened 2019-01-28
LEAmDNS_Fixes_1_3

LaborEtArs:
Re-Fix 'avoid multiple initialization' code in LEAmDNSResponder::begin()

### DIFF
--- a/libraries/ESP8266mDNS/src/LEAmDNS.cpp
+++ b/libraries/ESP8266mDNS/src/LEAmDNS.cpp
@@ -89,7 +89,7 @@ bool MDNSResponder::begin(const char* p_pcHostname) {
     
     bool    bResult = (0 != m_pcHostname);
     
-    if (0 == m_pcHostname) {
+    if (0 == m_pUDPContext) {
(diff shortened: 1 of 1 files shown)


LaborEtArs:
@devyte : You're right, the code was TOO 'quick and dirty', but I don't like your proposed code too, When 'begin()' was called once, another call to begin() should fail. At least it shouldn't change something silently (as it would, if OTA will be initialized after mDNS by the user). Unfortunately I can't change the code here (webeditor?, maybe only for branch owners? Maybe not with IE...?)...
The initial 'bool bResult = (0 != m_pcHostname);' should be 'bool bResult = (0 != m_pUDPContext);' if multiple calls should lead to 'false' or simple 'bool bResult = false;' (as I would prefer).
In addition the secend log should to moved to the end of the 'if-clause' to avoid misleading logs.

LaborEtArs:
Updated (and included the 'schedule_function(std::bind(&MDNSResponder::_restart, this));' to avoid calling _restart() in SYS context.

PR #5696, opened 2019-01-30
Move .C to .CPP in the core

earlephilhower:
Use g++ to compile core files to get additional C++ checks on the code.


cores/esp8266/StackThunk.cpp
```diff
@@ -115,8 +118,10 @@ void stack_thunk_dump_stack()
   }
   ets_printf(">>>stack>>>\n");
   while (pos < stack_thunk_ptr) {
-    ets_printf("%08x:  %08x %08x %08x %08x\n", pos, pos[0], pos[1], pos[2], pos[3]);
+    ets_printf("%08x:  %08x %08x %08x %08x\n", (int32_t)pos, pos[0], pos[1], pos[2], pos[3]);
```
devyte: Is it possible to use %p for pos here?

mhightower83:
With the renaming of `cores/esp8266/umm_malloc/umm_malloc.c` to `cores/esp8266/umm_malloc/umm_malloc.cpp` I think you missed updating its entry in the linker file:
https://github.com/esp8266/Arduino/blob/f37b4921e300ff5dfce38e8a0280f438f4bd4da1/tools/sdk/ld/eagle.app.v6.common.ld.h#L138-L139

earlephilhower:
Good eyes, @mhightower83 !  #6161 should fix.

PR #5709, opened 2019-02-01
Allow unaligned input/output to SPI::transferBytes

### LINKED ISSUE #4967: SPI::transferBytes pending changes
### Basic Infos

#### Platform

- Hardware:			        all
- Core Version:			latest git
- Development Env:		all
- Operating System:	        all

### Settings in IDE

- Module:  			all
- Flash Mode:			all
- Flash Size: 		all
- lwip Variant:   all
- Reset Method:		all
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz|160MHz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

This issue is meant to track pending changes to SPI::transferBytes, specifically a missing assert and some api optimizations.
Ref: #4925 #2677 .

*edit 19/01/22 d-a-v*: The API has to be clarified. `transfer(buffer)` and `transferBytes(buffer)` have the same signature, the first accepts unaligned buffer, not the second.

earlephilhower:
Fixes #4967

Support any alignment of input and output pointers and transfer lengths
in SPI::transferBytes.  Use 32-bit transfers and FIFO as much as
possible.

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -528,93 +528,95 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
 }
 
 /**
+ * Note:
+ *  in and out need to be aligned to 32Bit
+ *  or you get an Fatal exception (9)
  * @param out uint8_t *
  * @param in  uint8_t *
  * @param size uint8_t (max 64)
  */
-void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
-    while(SPI1CMD & SPIBUSY) {} // Make sure we're IDLE on the SPI
 
-    int bytesToTransfer = size; // How much to read/write from the FIFO in bytes
-
-    if (out && !in) {
-        int bytesLeft = size;
-
-        // Only transmitting, get out 32b aligned
-        while (bytesLeft && ((uint32_t)out & 3)) {
-            SPI.transfer(*(out++));
-            bytesLeft--;
-        }
-
-        if (!bytesLeft) return;
-        bytesToTransfer = bytesLeft;
-
-        // Now do 32b writes until we have 0 or fewer bytes left
-        // Note we can read and store past the end of string because it will not be sent (setDataBits)
-        uint32_t *dataPtr = (uint32_t*)out;
-        volatile uint32_t *fifoPtr = &SPI1W0;
-        while (bytesLeft > 0) {
-            *(fifoPtr++) = *(dataPtr++);
-            bytesLeft -= 4;
-        }
-        // Remainder will be sent out of loaded FIFO below
-    } else if (in && !out) {
-        int bytesLeft = size;
-
-        // Only receiving, get in 32b aligned
-        while (bytesLeft && ((uint32_t)in & 3)) {
-            *(in++) = SPI.transfer(0xff);
-            bytesLeft--;
-        }
-
-        if (!bytesLeft) return;
-        bytesToTransfer = bytesLeft;
-
-        // Now send 32b writes of all 0xff, actual read will be done below
-        // Note we can write add'l 0xffs and they won't be sent thanks to setDataBits below
-        volatile uint32_t *fifoPtr = &SPI1W0;
-        while (bytesLeft > 0) {
-            *(fifoPtr++) = 0xffffffff;
-            bytesLeft -= 4;
-        }
-    } else if (in && out && (((uint32_t)in & 3) || ((uint32_t)out & 3))) {
-        // Bidirectional and we have one or both pointers misaligned
-        while (size) {
-            *(in++) = SPI.transfer(*(out++));
-            size--;
-        }
-        // All the data was transferred byte-wise, we're done
+void SPIClass::transferBytesAligned_(const uint8_t * out, uint8_t * in, uint8_t size) {
+    if (!size)
         return;
+
+    while(SPI1CMD & SPIBUSY) {}
+    // Set in/out Bits to transfer
+
+    setDataBits(size * 8);
+
+    volatile uint32_t *fifoPtr = &SPI1W0;
+    uint8_t dataSize = ((size + 3) / 4);
+
+    if (out) {
+        uint32_t *dataPtr = (uint32_t*) out;
+        while (dataSize--) {
+            *fifoPtr = *dataPtr;
+            dataPtr++;
+            fifoPtr++;
+        }
+    } else {
+        // no out data only read fill with dummy data!
+        while (dataSize--) {
+            *fifoPtr = 0xFFFFFFFF;
+            fifoPtr++;
+        }
     }
 
-    // At this point we've got aligned (or null) in and out and the FIFO is filled
-    // with whatever data we're going to transmit as aligned
-
-    // Set in/out bits to transfer
-    while (SPI1CMD & SPIBUSY) {} // Paranoia, make sure SPI is idle
-    setDataBits(bytesToTransfer * 8);
-
-    // Start the transfer and wait for done
     SPI1CMD |= SPIBUSY;
-    while (SPI1CMD & SPIBUSY) { /* noop, waiting */ }
+    while(SPI1CMD & SPIBUSY) {}
 
     if (in) {
-        // Bulk read out by 4 until we have 0..3 left
-        uint32_t *dataPtr = (uint32_t*)in;
-        volatile uint32_t *fifoPtr = &SPI1W0;
-        while (bytesToTransfer >= 4) {
+        uint32_t *dataPtr = (uint32_t*) in;
+        fifoPtr = &SPI1W0;
+        dataSize = size;
+        while (dataSize >= 4) {
             *(dataPtr++) = *(fifoPtr++);
-            bytesToTransfer -= 4;
-            in += 4; // Keep track of the in ptr for any stragglers
+	    dataSize -= 4;
+            in += 4;
         }
-
-        // Bytewise read out the remainder
-        volatile uint8_t *fifoPtrB = (uint8_t*)fifoPtr;
-        while (bytesToTransfer--) {
+        volatile uint8_t *fifoPtrB = (volatile uint8_t *)fifoPtr;
+        while (dataSize--) {
             *(in++) = *(fifoPtrB++);
         }
     }
 }
+
+
+void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
+    if (!((uint32_t)out & 3) && !((uint32_t)in & 3)) {
+        // Input and output are both 32b aligned or NULL
+        transferBytesAligned_(out, in, size);
+    } else if (!out && ((uint32_t)in & 3)) {
+        // Input only and misaligned, do bytewise until in aligned
+        while (size && ((uint32_t)in & 3)) {
+            *(in++) = transfer(0xff);
+	    size--;
+        }
+        transferBytesAligned_(out, in, size);
+    } else if (!in && ((uint32_t)out & 3)) {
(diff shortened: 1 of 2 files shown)


libraries/SPI/SPI.cpp
```diff
@@ -531,48 +528,90 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
...
 
-    if(out) {
-        uint32_t * dataPtr = (uint32_t*) out;
-        while(dataSize--) {
-            *fifoPtr = *dataPtr;
-            dataPtr++;
-            fifoPtr++;
+        // Only transmitting, get out 32b aligned
+        while (bytesLeft && ((uint32_t)out & 3)) {
+            SPI.transfer(*(out++));
```
devyte: What? Did you mean just:
`transfer(*(out++));`
?

libraries/SPI/SPI.cpp
```diff
@@ -531,48 +528,90 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
...
+            bytesLeft--;
         }
-    } else {
-        // no out data only read fill with dummy data!
-        while(dataSize--) {
-            *fifoPtr = 0xFFFFFFFF;
-            fifoPtr++;
+
+        if (!bytesLeft) return;
```
devyte: return on its own line please

libraries/SPI/SPI.cpp
```diff
@@ -531,48 +528,90 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
...
-    while(SPI1CMD & SPIBUSY) {}
-    // Set in/out Bits to transfer
+    while(SPI1CMD & SPIBUSY) {} // Make sure we're IDLE on the SPI
 
-    setDataBits(size * 8);
+    int bytesToTransfer = size; // How much to read/write from the FIFO in bytes
 
-    volatile uint32_t * fifoPtr = &SPI1W0;
-    uint8_t dataSize = ((size + 3) / 4);
+    if (out && !in) {
```
devyte: This if-else is kind of big. I suggest breaking it up into smaller functions, one for each of the cases, and give the functions meaningful names. That would leave the top logic in here, and the specific lower level steps in each of the functions. E.g.: 
```
if(out && !in)
    trasmitOnly(blah);
else if(!out && in)
    receiveOnly(blah)
else if(in && out && (isMisaligned(in) || isMisaligned(out)))
    transmitAndReceiveMisaligned(blah);
...
```

libraries/SPI/SPI.cpp
```diff
@@ -531,48 +528,90 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
...
+            *(fifoPtr++) = *(dataPtr++);
+            bytesLeft -= 4;
         }
+        // Remainder will be sent out of loaded FIFO below
+    } else if (in && !out) {
+        int bytesLeft = size;
+
+        // Only receiving, get in 32b aligned
+        while (bytesLeft && ((uint32_t)in & 3)) {
+            *(in++) = SPI.transfer(0xff);
```
devyte: Shouldn't this be just transfer(0xff)?

libraries/SPI/SPI.cpp
```diff
@@ -531,48 +528,90 @@ void SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size) {
...
+        // Note we can write add'l 0xffs and they won't be sent thanks to setDataBits below
+        volatile uint32_t *fifoPtr = &SPI1W0;
+        while (bytesLeft > 0) {
+            *(fifoPtr++) = 0xffffffff;
+            bytesLeft -= 4;
+        }
+    } else if (in && out && (((uint32_t)in & 3) || ((uint32_t)out & 3))) {
+        // Bidirectional and we have one or both pointers misaligned
+        while (size) {
+            *(in++) = SPI.transfer(*(out++));
```
devyte: remove SPI

libraries/SPI/SPI.cpp
earlephilhower: Refactored to remove this bit completely.

libraries/SPI/SPI.cpp
earlephilhower: Refactored to remove this bit completely.  There now is a much smaller function which just handles alignment with a call to the main aligned method.

libraries/SPI/SPI.cpp
earlephilhower: Refactored to remove this bit completely.

libraries/SPI/SPI.cpp
earlephilhower: Refactored to remove this bit completely.

earlephilhower:
I realized after testing that you can't do bytewise updates to the FIFO space because it's RmW.  When you read, you're not reading what was writen by the core but the SPI HW read FIFO.  So the original refactoring could not work.

Rewrote it with a small shim with handles alignment and then calls the existing transfer operation with aligned input and output buffers, and then handles unaligning output as needed.

earlephilhower:
Testing with SDFS and forcing both misaligned read and write buffers as well as no-by-4 lengths, looks good in all testing I can throw at it.

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
         }
     }
 }
 
+
+void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
+    if (!((uint32_t)out & 3) && !((uint32_t)in & 3)) {
+        // Input and output are both 32b aligned or NULL
+        transferBytesAligned_(out, in, size);
+    } else if (!out && ((uint32_t)in & 3)) {
```
devyte: Why do you care about misaligned condition here? I think the rhs of the && should be removed. The while loop will handle both misaligned and aligned (skip) cases.

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
+        // Input only and misaligned, do bytewise until in aligned
+        while (size && ((uint32_t)in & 3)) {
+            *(in++) = transfer(0xff);
+	    size--;
+        }
+        transferBytesAligned_(out, in, size);
+    } else if (!in && ((uint32_t)out & 3)) {
```
devyte: Same as above: I think the rhs of the && should be removed.

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
+        // Output only and misaligned, bytewise xmit until aligned
+        while (size && ((uint32_t)out & 3)) {
+            transfer(*(out++));
+	    size--;
+        }
+        transferBytesAligned_(out, in, size);
+    } else {
```
devyte: If in both cases above the rhs of the && is removed, then this case means both in and out are non-nullptr, and one or both are misaligned.
Because both in and out are known to be non-ullptr, no need to to check them for that.

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
+        // HW FIFO has 64b limit, so just align in RAM and then send to FIFO aligned
+        uint8_t outAligned[64]; // Stack vars will be 32b aligned
+        uint8_t inAligned[64]; // Stack vars will be 32b aligned
+        if (out) {
+            memcpy(outAligned, out, size);
```
devyte: This looks wrong to me. If size > 64, you'll have stack corruption.
Shouldn't there be a loop here or something, where you take 64bytes at a time from out and memcpy them to outAligned, then push out outAligned?

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
+        } else {
+            memset(outAligned, 0xff, size); // 0xff = no xmit data
+        }
+        transferBytesAligned_(outAligned, inAligned, size);
+        if (in) {
+            memcpy(in, inAligned, size);
```
devyte: what if size > 64b?

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
+    } else {
+        // HW FIFO has 64b limit, so just align in RAM and then send to FIFO aligned
+        uint8_t outAligned[64]; // Stack vars will be 32b aligned
+        uint8_t inAligned[64]; // Stack vars will be 32b aligned
+        if (out) {
+            memcpy(outAligned, out, size);
+        } else {
+            memset(outAligned, 0xff, size); // 0xff = no xmit data
+        }
+        transferBytesAligned_(outAligned, inAligned, size);
```
devyte: what if size > size of the aligned stack buffers, which is 64b?

libraries/SPI/SPI.cpp
```diff
@@ -565,18 +566,58 @@ void SPIClass::transferBytes_(const uint8_t * out, uint8_t * in, uint8_t size) {
...
-        uint32_t * dataPtr = (uint32_t*) in;
+    if (in) {
+        uint32_t *dataPtr = (uint32_t*) in;
         fifoPtr = &SPI1W0;
-        dataSize = ((size + 3) / 4);
-        while(dataSize--) {
-            *dataPtr = *fifoPtr;
-            dataPtr++;
-            fifoPtr++;
+        dataSize = size;
```
devyte: dataSize = size here, but it was (size+3)/4 before. From the below code, I suspect it's to handle the last 1-3 bytes at the end when size is not a multiple of 4.
If that is correct, why is the above code for out the same as before? Shouldn't that be changed in the same way?

libraries/SPI/SPI.cpp
earlephilhower: This and below are actually by the API and HW limitations.  The HW FIFO buffer is 64bytes, and calls of >64b are split into 64b by the actual `transferBytes()` call.  (This is the private `transferBytes_` call that does the real work in 64b or less chunks).  You call "transferBytes(128)", this specific function invoked 2x.

libraries/SPI/SPI.cpp
earlephilhower: See above, size>64 is not possible in this bit of code.

libraries/SPI/SPI.cpp
earlephilhower: See above, size>64 is not possible in this bit of code.

libraries/SPI/SPI.cpp
earlephilhower: D'oh, that is correct and we do know that at this point in is misaligned.  If we had (out=null, in=align), the 1st case would grab this.  It's a no-op, but doesn't affect functionality.

libraries/SPI/SPI.cpp
earlephilhower: Yup, save a few insns.

libraries/SPI/SPI.cpp
earlephilhower: Ah, yes.  fair enough.  These checks below aren't needed, but don't affect functionality.

libraries/SPI/SPI.cpp
earlephilhower: Renamed and commented for clarity.

PR #5728, opened 2019-02-06
fix DEBUG macros

### LINKED ISSUE #5658: Reminder: rework all debug directives
The good way to display a debug message is:
```
#define DEBUG_HTTPCLIENT(fmt, ...) \
    DEBUG_ESP_PORT.printf_P((PGM_P)PSTR(fmt), ## __VA_ARGS__)
```

Every `DEBUG_` directives should follow this model.

### LINKED ISSUE #5855: OTA crash with UDP server and CORE debug
### Basic Infos

- [ x] I have searched the issue tracker for a similar issue.
-  [   ] If there is a stack dump, I have decoded it.

#### Platform

- Hardware:				ESP-12
- Core Version:			git | last commit I have  `2528013c5dc1ce7703bf211b41a3aefcd6afe0e5`
- Operating System:		MacOS

### Settings in IDE

- Flash Mode:			qio
- Flash Size: 		4MB
- lwip Variant:   v2 Higher Bandwidth
- Reset Method:	ck
- Flash Frequency:	40Mhz
- CPU Frequency:   80Mhz
- Upload Using:       OTA
- Debug Port:          Serial
- Debug Level:        Core
- Espressif FW:         2.2.1

### Problem Description

OTA updates crash with `Fatal exception 3(LoadStoreErrorCause)` when there is a UDP server running **AND** CORE debug is enabled.
If CORE debug is disabled, it works just fine.

Haven't been able to decode the stack, because with the the git version the ESPExceptionDecoder is not able to locate the file `xtensa-lx106-elf-addr2line`:`ERROR: xtensa-lx106-elf-addr2line not found!`

```cpp

#include <ESP8266WiFi.h>
#include <WiFiClient.h>
#include <ESP8266WebServer.h>
#include <ESP8266HTTPUpdateServer.h>
#include <WiFiUDP.h>

unsigned int localPort = 8888;

#ifndef STASSID
#define STASSID "ssid"
#define STAPSK  "password"
#endif

const char* ssid = STASSID;

ESP8266WebServer httpServer(80);
ESP8266HTTPUpdateServer httpUpdater;
WiFiUDP Udp;
char packetBuffer[UDP_TX_PACKET_MAX_SIZE];

void setup(void) {
  Serial.begin(9600);
  Serial.println();
  Serial.println("Booting Sketch...");
  WiFi.mode(WIFI_STA);
  WiFi.begin(ssid, password);

  while (WiFi.waitForConnectResult() != WL_CONNECTED) {
    WiFi.begin(ssid, password);
    Serial.println("WiFi failed, retrying.");
  }
  Serial.print("IP: ");
  httpUpdater.setup(&httpServer);
  httpServer.begin();
  Udp.begin(localPort);

}
void loop(void) {
  httpServer.handleClient();
    // if there's data available, read a packet
  int packetSize = Udp.parsePacket();
  if (packetSize) {
    Serial.print("Received packet of size ");
    Serial.println(packetSize);
    Serial.print("From ");
    IPAddress remote = Udp.remoteIP();
    for (int i = 0; i < 4; i++) {
      Serial.print(remote[i], DEC);
      if (i < 3) {
        Serial.print(".");
      }
    }
    Serial.print(", port ");
    Serial.println(Udp.remotePort());

    // read the packet into packetBufffer
    Udp.read(packetBuffer, UDP_TX_PACKET_MAX_SIZE);
    Serial.println("Contents:");
    Serial.println(packetBuffer);
  }
	delay(10);
}
```

d-a-v:
All fmt strings in flash
fix #5658

This also allows to avoid warnings and easy mistakes with (no brace):
```
    if (something)
        DEBUGV("blah");
```

*edit*
use newlib unaligned-compatible printf for DEBUGV
fix #5855


cores/esp8266/debug.h
```diff
@@ -5,11 +5,11 @@
 #include <stdint.h>
 
 #ifdef DEBUG_ESP_CORE
-#define DEBUGV(...) ets_printf(__VA_ARGS__)
+#define DEBUGV(fmt, ...) ::printf_P( (PGM_P)PSTR(fmt), ## __VA_ARGS__)
```
earlephilhower: Should this not be the same as others (i.e. `DEBUG_ESP_PORT.printf(`...)?

igrr: ets_printf is safe to call in any context (such as system), but DEBUG_ESP_PORT.printf could yield, which would not work in sys context?

d-a-v: I changed to `::printf_P` so we can use fmt arguments from flash.
I assume `::printf_P` uses `putchar` installed by `ets_install_putc1/putc2` in debug mode like, `ets_printf` does.
(should the others use `::printf_P` instead of `Serial.printf_P` ?)

d-a-v:
use newlib unaligned-compatible printf for DEBUGV
fix #5855

cores/esp8266/debug.h
d-a-v: @igrr Were you suggesting to remove (legacy) `DEBUG_ESP_PORT.printf` and use `DEBUGV` instead ?
About sys context, I believe new newlib's `::vsnprintf()` and fw's `ets_putc()` are safe (no yield).

igrr:
Sorry, I thought that _write_r (which is called by printf) uses
HardwareSerial. Now I see it uses ets_putc, so indeed no issue with calling
printf from sys context.

Issue #5734, opened 2019-02-06
Crash when Debug port is Disabled - HTTPClient destructor issue?


joysfera:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:				[ESP-12]
- Core Version:			[just released 2.5.0]
- Development Env:		[Arduino IDE 1.8.8]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[dout]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[ck]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

In Arduino IDE when `Debug port` is set to `Disabled` the ESP8266 reboots (most probably due to a crash) a short while after fetching JSON data via https, parsing it and displaying it on a LED matrix. However, if I set the `Debug port` to `Serial` then everything works properly. My application does not use any debugging stuff. It does not even use the Serial stuff, does not call `Serial.begin()`, nothing of that.

The `Debug level` is set to `None` all the time so there's no reason why `Debug port` setting should be significant or should affect the application in any way, IMHO.

I cannot debug this because when I enable debugging and thus set the Debug port to Serial the issue disappears :-)

EDIT: here's the minimal code:
```
#include <ESP8266WiFi.h>
#include <ESP8266HTTPClient.h>
#include <WiFiClientSecureBearSSL.h>

const int led = 5;

void setup(void)
{
  pinMode(led, OUTPUT);
  for(byte i=0; i<5; i++) {
    digitalWrite(led, HIGH);
    delay(150);
    digitalWrite(led, LOW);
    delay(200);
  }
  
  WiFi.mode(WIFI_STA);
  // Wait for connection
  while (WiFi.status() != WL_CONNECTED) {
    delay(250);
  }
}

void displayData()
{
[...]

PR #5739, opened 2019-02-08
Wrong HTTPS example fixed. Will save others my #5734 hours of debugging

joysfera:
The provided HTTPS examples are misleading and plain wrong because don't ensure the client object lives longer than the https object, which is required per comment hidden in the source code. Otherwise nasty crashes will occur (#5734 ). I have fixed it using a nested code block and a clear comment.


libraries/ESP8266HTTPClient/examples/BasicHttpsClient/BasicHttpsClient.ino
```diff
@@ -45,32 +45,36 @@ void loop() {
...
-        // HTTP header has been send and Server response header has been handled
-        Serial.printf("[HTTPS] GET... code: %d\n", httpCode);
-
-        // file found at server
-        if (httpCode == HTTP_CODE_OK || httpCode == HTTP_CODE_MOVED_PERMANENTLY) {
-          String payload = https.getString();
-          Serial.println(payload);
+    /* in order to ensure the client object lives the entire time of the HTTPClient
+       the HTTPClient is in its own nested code block */
+    {
```
devyte: HTTPClient doesn't need to be in its own code block, it just needs to be instantiated _after_ client. Destructors are called in reverse order, which ensures https gets destroyed first and client second.
I suggest removing the code block, and updating the comment to explain this.

libraries/ESP8266HTTPClient/examples/StreamHttpsClient/StreamHttpsClient.ino
```diff
@@ -54,56 +54,60 @@ void loop() {
 
     client->setFingerprint(fingerprint);
 
-    HTTPClient https;
+    /* in order to ensure the client object lives the entire time of the HTTPClient
+       the HTTPClient is in its own nested code block */
+    {
```
devyte: Same as above.

libraries/ESP8266HTTPClient/examples/BasicHttpsClient/BasicHttpsClient.ino
joysfera: I don't agree. It's way too dangerous to rely on order of code lines. Took me two days to debug that. It should be totally explicit, IMHO.

d-a-v:
@joysfera The nested block is not the preferred way but the information or warning is relevant. 
Would you update your PR with a comment to inform readers about the requirement ?
Also the last debug line is interesting to have.

joysfera:
I understand you don't want to accept the nested block solution. However, I don't agree with a "solution" based on the order of variable declarations. Adding a comment doesn't really "fix" that.
The whole API is wrong. Possible solution is listed here: https://github.com/esp8266/Arduino/issues/5734#issuecomment-475858786

earlephilhower:
Closing as abandoned with changes requested.

PR #5750, opened 2019-02-10
Fix ssdp

Lan-Hekary:
Hello,
I noticed that the SSDP library always respond to the M-Search Packet with its Device type ..
and does not respond to UUID Search ..
such behavior prevents the other devices from querying the specific device (Such as Windows Network Discovery)
the Properties was returning an error beacuse of this bug
image

I fixed it without breaking compatibility and tested it thoroughly ..


libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -155,12 +156,14 @@ SSDPClass::~SSDPClass() {
...
-
+  
   _pending = false;
+  _st_is_uuid = false;
   if (strcmp(_uuid,"") == 0) {
 	uint32_t chipId = ESP.getChipId();
-	sprintf(_uuid, "38323636-4558-4dda-9188-cda0e6%02x%02x%02x",
-    (uint16_t) ((chipId >> 16) & 0xff),
+	sprintf(_uuid, "%s38323636-4558-4dda-9188-cda0e6%02x%02x%02x",
+  "uuid:",
```
d-a-v: Why don't you include "uuid:" in the format string ?
(you can also put it in flash)
```
sprintf_P(_uuid, F("uuid:38323636-4558-4dda-9188-cda0e6%02x%02x%02x"),
    (uint16_t) ((chipId >> 16) & 0xff), ...
```

Lan-Hekary: I did not think of that actually 😃
Thanks for noticing I will change it now ..
Any other notes ?!

Lan-Hekary:
I was thinking of using Dynamic allocation or Strings to store the parameters instead of Fixed size array ..
I can start on that too ..
But I need to know if you will accept it ..

libraries/ESP8266SSDP/ESP8266SSDP.cpp
d-a-v: Yes, same notice for `SSDPClass::setUUID` :)

d-a-v:
> But I need to know if you will accept it ..

I can't know.
What buffer, the 64 bytes declared in `::_update()`'s stack ?

Lan-Hekary:
no .. I mean 
`char _schemaURL[SSDP_SCHEMA_URL_SIZE];
char _uuid[SSDP_UUID_SIZE];
char _deviceType[SSDP_DEVICE_TYPE_SIZE];
char _friendlyName[SSDP_FRIENDLY_NAME_SIZE];
char _serialNumber[SSDP_SERIAL_NUMBER_SIZE];
char _presentationURL[SSDP_PRESENTATION_URL_SIZE];
char _manufacturer[SSDP_MANUFACTURER_SIZE];
char _manufacturerURL[SSDP_MANUFACTURER_URL_SIZE];
char _modelName[SSDP_MODEL_NAME_SIZE];
char _modelURL[SSDP_MODEL_URL_SIZE];
char _modelNumber[SSDP_MODEL_VERSION_SIZE];`

d-a-v:
> no .. I mean

Yes, maybe in another pull-request so one fix doesn't delay the other

libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -439,7 +454,7 @@ void SSDPClass::setDeviceType(const char *deviceType) {
 }
 
 void SSDPClass::setUUID(const char *uuid) {
-  strlcpy(_uuid, uuid, sizeof(_uuid));  
+  snprintf_P(_uuid, sizeof(_uuid), PSTR("%S%s"), F("uuid:"), uuid);
```
d-a-v: or `snprintf_P(_uuid, sizeof(_uuid), PSTR("uuid:%s"), uuid);` ?

Lan-Hekary: done ..
I really should start noticing these stuff 😃

d-a-v:
Using String will be a good idea, especially since #5690 just merged

Lan-Hekary:
> Using String will be a good idea, especially since #5690 (just merged)

OK.. I will make another pull request after this pull request is merged with the new modifications ..

libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -71,7 +71,7 @@ static const char _ssdp_packet_template[] PROGMEM =
   "%s" // _ssdp_response_template / _ssdp_notify_template
   "CACHE-CONTROL: max-age=%u\r\n" // SSDP_INTERVAL
   "SERVER: Arduino/1.0 UPNP/1.1 %s/%s\r\n" // _modelName, _modelNumber
-  "USN: uuid:%s\r\n" // _uuid
+  "USN: %s\r\n" // _uuid
```
devyte: why is the prefix moved out of the template? it increases permanent heap use by 10 bytes due to the increased _uuid size, and I don't see a good reason for it. I prefer to keep the prefix here.

libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -99,7 +99,7 @@ static const char _ssdp_schema_template[] PROGMEM =
   "<modelURL>%s</modelURL>"
   "<manufacturer>%s</manufacturer>"
   "<manufacturerURL>%s</manufacturerURL>"
-  "<UDN>uuid:%s</UDN>"
+  "<UDN>%s</UDN>"
```
devyte: Same as above comment for USN.

devyte:
If you make a PR with Strings replacing the fixed-size arrays, please try to avoid concatenation, use direct construction or assignment when possible, and use ::reserve().

libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -241,7 +246,7 @@ void SSDPClass::_send(ssdp_method_t method) {
                        _modelNumber,
                        _uuid,
                        (method == NONE) ? "ST" : "NT",
-                       _deviceType,
+                       (_st_is_uuid) ? _uuid : _deviceType,
```
Lan-Hekary: I moved the the prefix `uuid:` to the _uuid to avoid concatenating it in the line 249
alternative proposal would be
`(_st_is_uuid) ? String("uuid:" + String(_uuid)).c_str() : _deviceType,`
but this will introduce a bigger overhead I guess ..

do you have any other thoughts or alternatives I can use ??

libraries/ESP8266SSDP/ESP8266SSDP.cpp
Lan-Hekary: check this review :
https://github.com/esp8266/Arduino/pull/5750#pullrequestreview-205353824

PR #5758, opened 2019-02-13
Move __PRETTY_FUNCTION__ to progmem

earlephilhower:
Match `__FUNCTION__` linking, keep C++ function names out of rodata/heap.


cr1st1p:
Guys, not sure if this was announced or not, but this looks like a rather big change.
On one hand it is good (for those doing logging - it frees some RAM), but on the other hand... you should NOT expect that GCC definition is ONLY used via Arduino's String() class !
Possible wrong cases, where pull request #6368 does nothing:
- allocating some memory (new/malloc), building a string, doing then a strcat(dest, __PRETTY_FUNCTION) (untested)
- doing std:string s(__PRETTY_FUNCTION__) - crashes (real life example :-( )

PR #5763, opened 2019-02-15
Revert to nonos-sdk 2.2.1, new selection menu for generic board only

### LINKED ISSUE #5736: ESP8266WebServer hangs after updating to 2.5.0
### Basic Infos

#### Platform

- Hardware:				ESP8266
- Core Version:			2.5.0
- Operating System:		MacOS

### Settings in IDE

- Module:  			NodeMCU1.0
- Flash Mode:			n/a
- Flash Size: 		4MB/3MB
- Reset Method:		n/a
- Flash Frequency:	n/a
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

After updating to v2.5.0 my sketch can't get a reliable connection to the ESP8266WebServer server. Trying to connect to the ESP from a browser succeeds occasionally, but most of the time the browser hangs and eventually gives up.

I loaded the simple HelloServer example sketch that comes with ESP8266WebServer and it does the same thing. The browser just sits there. Tried various lwIP variants, but none helped. Oddly, turning on all debug options made the problem happen less often.

It seems related to issue #5725, although for me ping requests constantly timeout.

Revert back to v2.4.2 and the problem goes away.

Use the HelloServer example sketch provided with ESP8266WebServer.

### Debug Messages

Problem goes away (or at least gets much better) when all debug options are enabled.

d-a-v:
This commit allows switching SDK firmware:

nonos-sdk-pre-v3 shipped with release 2.5.0 has issues:
- Some boards show erratic behavior (radio connection is quickly lost), with an unknown cause.
  These boards work well with previous nonos-sdk-2.2.1 firmware (#5736)
- Overall performances seem to have decreased (#5513)

This PR restores sdk2.2.1 (as in core-2.4.2).

SDK-pre-3.0 - which has brought long awaited fixes (WiFi sleep modes) - is still available through a menu option available only with generic board.

BREAKING
- new define `-DNONOSDK221=1` or `-DNONOSDK3V0=1`
- for external build systems: new library directory: `tools/sdk/lib/<version>/lib`
- PIO: variable `PIO_FRAMEWORK_ARDUINO_ESPRESSIF_SDK3` is needed for sdk-pre-v3.

Fix #5736


Zeron:
Any chance of a 2.5.1 version with this fix included?

Issue #5769, opened 2019-02-16
esp8266 Arduino package can't be used as referred package


jandrassy:
Arduino IDE 3rd party Hardware specification section ["Referencing another core, variant or tool"](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#referencing-another-core-variant-or-tool) specifies that a boards package can be referred from custom hardware definitions or other boards packages.

esp8266 package can't be used now as referred, because platform.txt file entries use variable {runtime.platform.path}, which points to the folder of selected board's boards.txt file. In case of a referring definition it is not the right folder to access sdk and tools.

PR follows

jandrassy:
the PR did not solve it. the solution was to use `{build.core.path}/../..` used instead of `{runtime.platform.path}` but in `git version` not all platform.txt keys can work with {build.core.path}.

other option is to do the replace as part of release preparation

jandrassy:
@d-a-v David, after the new release I again try to solve the "referred package" issue.  The main problem are paths to the `tools` folder in platform folder (`hardware/esp8266/esp8266/tools`) used in platform.txt as `{runtime.platform.path}/tools/`
I discovered that Arduino IDE creators thought about need of a folder of this kind, but they prescribed a name `system` to it and then it is accessible as `{build.system.path}`. This works good if platform is used as referred package.

So if the `tools` folder would be renamed to `system` and all (note*) `{runtime.platform.path}/tools` would be replaced to `{build.system.path}` we would get closer to make Arduino esp8266 core work as referred.
The `{build.system.path}` works good in git install too.

  note*: there is a new problem in 2.5.1. the
```
tools.esptool.upload.network_pattern="{network_cmd}" "{runtime.platform.path}/tools/espota.py"
```
want evaluate anything else then the problematic `runtime.platform.path`

devyte:
@jandrassy I suspect this isn't an issue with wide impact, so you're the main interested party. You started a proposal with #5770, but then dropped it, and it's not clear to me why. Will you be continuing your work, or should this issue just be closed?

jandrassy:
the maintainers should be interested, because now every manufacturer of a esp8266 board must add the board to this package and can't maintain own.

it is hard to work on this if the platform.txt file changes so much like in last months.

would the maintainers accept renaming of the `tools` folder to `system'? it would solve problems with the path to `tools` because `system` is what Arduino builder expects as name for this kind of folder.

[...]

PR #5770, opened 2019-02-16
platform.txt - to work as referred package

jandrassy:
{build.core.path} used instead of {runtime.platform.path} to work as
referred package for Arduino 3rd-party-Hardware-specification chapter
["Referencing another core, variant or tool"](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#referencing-another-core-variant-or-tool)

solving https://github.com/esp8266/Arduino/issues/5769

note: the keys removed at release are unchanged. referring the package in dev setup doesn't work

comment: This will allow end users to define preferred sets of boards settings ([example](https://github.com/jandrassy/my_boards)) and manufacturers of esp8266 boards could create 'satellite' board packages using core and tools of this main esp8266 package.


jandrassy:
after https://github.com/esp8266/Arduino/commit/9790e1cb7cfa01ac74056fb835e6282af123aba0 it is unclear which lines will stay in platform.txt after release

d-a-v:
I fixed conflicts by hand. You need to check whether it still works as referred package.
You will also need to `git submodule update --init`

jandrassy:
I left `{runtime.platform.path}` in the lines that are removed at release. the `runtime.tools.*` work with `{build.core.path}` too, but the `tools.*` don't.
one of the `runtime.tools.*`, the `signing.py` stayed in release in 2.5.0, so I changed the path for it to `{build.core.path}` 
it is unclear know, I think, which of the `runtime.tools.*` will stay for release, so we can simply use `{build.core.path}` in all `runtime.tools.*`
I will see if I can test it by making a fake release installation. Before I tested it on 2.5.0 release.

jandrassy:
some entries (`tools.^`) don't work in dev 'mode' with `{build.core.path}/../..`, but need it after installation to work as referred package. most prominent problematic entry is
```
tools.esptool.upload.pattern="{cmd}" "{runtime.platform.path}/tools/upload.py" "{runtime.platform.path}/tools/pyserial" "{runtime.platform.path}/tools/esptool/esptool.py" --chip esp8266 --port "{serial.port}" --baud "{upload.speed}" "{upload.verbose}" {upload.erase_cmd} --end  --chip esp8266 --port "{serial.port}" --baud "{upload.speed}" "{upload.verbose}" write_flash 0x0 "{build.path}/{build.project_name}.bin" --end
```
I don't know what to do with it? It looks like it is not a problem of the changes, but I missed something as I tested it before. It looks like I must close this for know.

d-a-v:
Serial upload with windows was broken, and has just been fixed (#5796).
Maybe you can retry with your changes.

jandrassy:
> Serial upload with windows was broken, and has just been fixed (#5796).
> Maybe you can retry with your changes.
I am on Linux. Things worked in git version only not as referred package.

devyte:
@jandrassy please make your proposal in a PR, or update #5770. We can discuss further there.

PR #5785, opened 2019-02-19
emulation on host: Add full UART driver emulation.

everslick:
This PR replaces the high level Serial mock with a more complete UART driver. This way the HardwareSerial works without any modifications. Additionally the driver supports UART0 and UART1 at the same time (UART0 is directed to stdout and UART1 writes to stderr). RX is implemented by switching the terminal into raw non-blocking mode and injecting each key-press directly into the FIFO of UART0. A new command line switch -c was added to ignore CTRL-C and send it via serial as well. The decumentation was updated accordingly.

Reading and setting of GPIOs does only write to stderr, when compiled with D=1. But this is subject to be replaced with a proper GPIO emulation anyway. Reading from GPIO0 now returns 1 instead of 0 because this is most likely a low active input.


d-a-v:
@everslick Can you properly restore the terminal state after the double ^C,
I need to make a `stty sane` to restore `echo`.

Also is there a way to tell `term` to print `\r\n` when only `\n` is sent ?
Or, maybe with an option, tell mock serial to print another `\r` when `\n` is to be printed ?

everslick:
The restore is actually in, but I see it not working here as well. I'll have a look... The \r -> \r\n translation should be easy to add. I will commit something... thanks for the feedback! :-)

PR #5810, opened 2019-02-23
emulation on host: Fix compilation of spiffs_hal.cpp on 64 bit systems

everslick:
This patch is in preparation of a bigger PR that will bring low level SPI flash emulation to the host mocking.

### DIFF
--- a/cores/esp8266/spiffs_hal.cpp
+++ b/cores/esp8266/spiffs_hal.cpp
@@ -122,7 +122,7 @@ int32_t spiffs_hal_write(uint32_t addr, uint32_t size, uint8_t *src) {
 
     if (alignedEnd != alignedBegin) {
         uint32_t* srcLeftover = (uint32_t*) (src + alignedBegin - addr);
-        if (!(srcLeftover & 3)) { // source is already aligned
+        if (!((uintptr_t) srcLeftover & 3)) { // source is already aligned
             if (!ESP.flashWrite(alignedBegin, (uint32_t*) srcLeftover,
                     alignedEnd - alignedBegin)) {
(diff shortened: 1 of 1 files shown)


cores/esp8266/spiffs_hal.cpp
```diff
@@ -122,7 +122,7 @@ int32_t spiffs_hal_write(uint32_t addr, uint32_t size, uint8_t *src) {
 
     if (alignedEnd != alignedBegin) {
         uint32_t* srcLeftover = (uint32_t*) (src + alignedBegin - addr);
-        uint32_t srcAlign = ((uint32_t) srcLeftover) & 3;
+        uint32_t srcAlign = ((uintptr_t) srcLeftover) & 3;
         if (!srcAlign) {
```
d-a-v: because `uint32_t srcAlign = ((uintptr_t)... ` doesn't seem right in 64 bits and this variable is not used elsewhere, wouldn't it be cleaner to 
```
if (!(srcLeftover & 3)) {
    // source is not aligned
    ...
```
?

everslick: I agree! I didn't realize that srcAlign was only used once. good catch!

everslick: hmm, still does not compile, because & is not compatible to a pointer type. so the uintptr_t cast is still needed, I guess.

`if (!((uintptr_t)srcLeftover & 3)) { // source is already aligned`

everslick:
My First thought was to separate this PR from the bigger SPI flash emu PR, but there will be changes to other files in cores/esp8266 as well. So I decided to propose all changes in ONE PR. -> closed.

Issue #5804, opened 2019-02-21
Documentation paths not correct for testing repo in IDE


scottchiefbaker:
I see `esptool` version 2.6.0 has been in the tree since [October](https://github.com/espressif/esptool/commit/7460b2d699784e5debf39228ff27ae89cff68f3c#diff-2e7375ee839ab724de2eec9abc81b312). This version offers a lot of improvements over previous `esptool` versions. Most namely that it supports compression of images on the fly. This greatly increases upload speed, and would really help my workflow.

I've attempted to follow the instructions from the README but I don't think they're correct. On Linux the various packages are stored in `~/.arduino15/packages` not `~/arduino`. 

Following the instructions from the README I end up with `~/.arduino15/packages/hardware/esp8266com/esp8266/` which should probably just be `~/.arduino15/packages/esp8266/hardware`? Before I go too far down the rabbit hole am I on the right track?

scottchiefbaker:
If I install the official release from the IDE I get:

    ~/.arduino15/packages/esp8266/hardware
    ~/.arduino15/packages/esp8266/tools

I'm assuming if I clone this repo it should go in to `hardware` and then I download the tools via the pythong script to `tools`?

oddstr13:
You should clone this repo into your arduino sketch folder, under hardware - not where the boards manager installs the packages.

devyte:
No, you're not on the right track. The instructions are correct, the ~/.arduino15 dir is different. When you install git, the correct place is ~/arduino, or whatever dir you specify in the IDE preferences (I personally have ~/workspace/arduino).

My own setup is as follows (in case anyone reading this is interested in an alternate setup):
* in the above dir, I have the dir structure shown in the instructions: ~/workspace/arduino/hardware/esp8266com
* I forked the repo from the webpage to have my own copy
* I cloned this main repo into ~/workspace/Arduino.esp8266
* I cloned my forked repo into ~/workspace/Arduino.devyte
* I created a symlink inside esp8266com called esp8266 that points to one of the two above.
For official business and testing:
```
cd ~/workspace/arduino/hardware/esp8266com
symlink -s ../../../Arduino.esp8266 esp8266
```
For messing around with my own fork, or making PRs:
```
cd ~/workspace/arduino/hardware/esp8266com
rm esp8266
symlink -s ../../../Arduino.devyte esp8266
```

In other words, I can switch between the two.
[...]

PR #5816, opened 2019-02-24
Update the documentation with clearer paths and info on sub-modules

scottchiefbaker:


scottchiefbaker:
This should clear up some confusion that caused me to open #5804

README.md
```diff
@@ -42,29 +42,34 @@ Documentation: [https://arduino-esp8266.readthedocs.io/en/2.5.0/](https://arduin
...
+```
+
+- Download binary tools (requires Python 2.7)
 ```bash
-cd esp8266/tools
+cd tools/
 python get.py
 ```
-- Restart Arduino
+- Restart the Arduino IDE
```
oddstr13: I would note that opening the boards manager (and letting it download indexes?) gives me the same effect as restarting the IDE (on Windows 10 at least). Same with library manager for libraries.
Seems to force Arduino to do a re-scan of directories.

README.md
```diff
@@ -42,29 +42,34 @@ Documentation: [https://arduino-esp8266.readthedocs.io/en/2.5.0/](https://arduin
...
 ```bash
-cd hardware
-mkdir esp8266com
-cd esp8266com
-git clone https://github.com/esp8266/Arduino.git esp8266
+git clone https://github.com/esp8266/Arduino.git hardware/esp8266com/esp8266/
 ```
-- Download binary tools (you need Python 2.7)
+
+- This repo requires other tools (sub-modules) that we need to fetch
```
d-a-v: I think we don't need that in the documentation
There's currently only one sub-sub-module `tools/sdk/lwip2/builder/lwip2-src/` which is not necessary, because it is precompiled (in `tools/sdk/lib/`).

Or you could state:
> Sub-sub modules are not necessary for a fully running repository. However they can be downloaded with the following command:

scottchiefbaker: To get a functioning toolchain you 100% need the submodules. You don't get `pyserial` or `esptool` without fetching the submodules.

d-a-v: I am talking about sub-**sub**-modules

https://github.com/esp8266/Arduino/pull/5882#issuecomment-473598216

scottchiefbaker: Where are you seeing a reference to sub-sub-modules in the README?

d-a-v: `git clone --recursive` takes everything, we only need one level

```
git clone ...
cd <dir>
git submodule update --init
```

`git clone --recursive` is the equivalent of
```
git clone ...
cd <dir>
git submodule update --init --recursive
```


check #5882 which is merged now

d-a-v:
@scottchiefbaker If you don't agree with current master can you fix this PR (or close and make another one) ?

PR #5821, opened 2019-02-25
Fix repaintable stack calculation

### LINKED ISSUE #5794: cont_repaint_stack incorrectly calculates sp_safe in cont_util.cpp
### Basic Infos

#### Platform

- Hardware:				ESP-12E
- Core Version:			0x951aeffa

### Settings in IDE

- Module:  			Nodemcu 1.0
- Flash Mode:			qio
- Flash Size: 			4MB
- lwip Variant:   		v2 Lower Memory
- Reset Method:		nodemcu
- Flash Frequency:		40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:		115200

### Problem Description

I am very grateful for the work of @igrr and his contribution of the `cont_util` functions. It *appears* though that `cont_repaint_stack(cont_t *cont)` (called by `EspClass::resetFreeContStack()` in `Esp.cpp`) does not behave as documented in the comments of `cont_util.cpp`. 

The code documents that `cont_repaint_stack` should add 64 bytes adjacent to the current stack pointer before painting for safety reasons, however comparing the address pointed to by SP and the resulting address pointed to by `sp_safe` shows that there is much more than 64 bytes. If I'm correct, this results in not enough of the stack memory being painted, and  functions like `cont_get_free_stack(cont_t* cont)` returning an inaccurate stack free amount. 

If I have understood this correctly, this discrepancy is because, in part, the calculation of `sp_safe` doesn't appear to take into account that the operations are pointer arithmetic. I have provided one possible solution, assigned to variable `sp_safe_pos` below, which takes these into consideration. By subtracting the pointer of the head to the stack from the stack pointer, we are left with how many 4-byte entries are needing to be painted, which conveniently translates to the number of stack array entries in `g_pcont` we need to paint. Adding the 64 byte buffer to this is then trivial.

If I have misunderstood this problem, please advise.  

```cpp
#include <cont.h>

void setup() {
  Serial.begin(115200);
  /* Get stack */
  extern cont_t* g_pcont;
  /* use renamed variable to ensure we can copy code from cont_util.h without change */
  cont_t* cont = g_pcont;

  /* Snippet from cont_util.cpp:73-76 */
  register uint32_t *sp asm("a1");
  // Ensure 64 bytes adjacent to the current SP don't get touched to endure
  // we don't accidentally trounce over locals or IRQ temps.
  uint32_t sp_safe = CONT_STACKSIZE/4 - ((sp - &cont->stack[0] - 64)/4);

  /* Compare Addresses */
  Serial.printf("\n\nSP: 0x%08X, sp_safe: %d, &cont->stack[sp_safe]: 0x%08X.\n",sp, sp_safe, &cont->stack[sp_safe]);
  Serial.printf("%d bytes between SP and &cont->stack[sp_safe].\n",(int)sp - (int)&cont->stack[sp_safe]);

  // Solution: 
  // Subtracting the stack pointer address from the stack head pointer address 
  // automatically results in correct stack index. 
  // Because this is pointer arithmetic, we subtract 16 to give us 64 bytes (64/4=16).
  // Add "_pos" so we don't confuse this with an actual memory address. This is an array index.
  uint32_t sp_safe_pos = sp - &cont->stack[0] - 16;

  /* Compare Addresses */
  Serial.printf("\n\nSP: 0x%08X, sp_safe_pos: %d, &cont->stack[sp_safe_pos]: 0x%08X.\n",sp, sp_safe_pos, &cont->stack[sp_safe_pos]);
  Serial.printf("%d bytes between SP and &cont->stack[sp_safe_pos].\n",(int)sp - (int)&cont->stack[sp_safe_pos]);

}

void loop() {
}

```
### Code Output
```
SP: 0x3FFFFF90, sp_safe: 787, &cont->stack[sp_safe]: 0x3FFFFC0C.
900 bytes between SP and &cont->stack[sp_safe].


SP: 0x3FFFFF90, sp_safe_pos: 996, &cont->stack[sp_safe_pos]: 0x3FFFFF50.
64 bytes between SP and &cont->stack[sp_safe_pos].
```

earlephilhower:
Fixes #5794 as found by @mattbradford83

### DIFF
--- a/cores/esp8266/cont_util.cpp
+++ b/cores/esp8266/cont_util.cpp
@@ -74,7 +74,7 @@ void cont_repaint_stack(cont_t *cont)
     // Ensure 64 bytes adjacent to the current SP don't get touched to endure
     // we don't accidentally trounce over locals or IRQ temps.
     // Fill stack with magic values
-    for ( uint32_t *pos = sp - 16;  pos != &cont->stack[0]; pos-- ) 
+    for ( uint32_t *pos = sp - 16; pos >= &cont->stack[0]; pos-- )
     {
        *pos = CONT_STACKGUARD;
     }
--- a/cores/esp8266/cont_util.cpp
+++ b/cores/esp8266/cont_util.cpp
@@ -73,12 +73,10 @@ void cont_repaint_stack(cont_t *cont)
     register uint32_t *sp asm("a1");
     // Ensure 64 bytes adjacent to the current SP don't get touched to endure
     // we don't accidentally trounce over locals or IRQ temps.
-    uint32_t sp_safe = CONT_STACKSIZE/4 - ((sp - &cont->stack[0] - 64)/4);
-
     // Fill stack with magic values
-    for(uint32_t pos = 0; pos < sp_safe; pos++)
+    for ( uint32_t *pos = sp - 16;  pos != &cont->stack[0]; pos-- ) 
     {
-        cont->stack[pos] = CONT_STACKGUARD;
+       *pos = CONT_STACKGUARD;
     }
 }
 
--- a/tools/signing.py
+++ b/tools/signing.py
@@ -27,12 +27,12 @@ def main():
                 val += "#include <pgmspace.h>\n"
                 val += "#define ARDUINO_SIGNING 1\n"
                 val += "static const char signing_pubkey[] PROGMEM = {\n"
-                for i in pub:
-                    val += "0x%02x, \n" % ord(i)
+                for i in bytearray(pub):
+                    val += "0x%02x, \n" % i
                 val = val[:-3]
                 val +="\n};\n"
                 sys.stderr.write("Enabling binary signing\n")
-        except:
+        except IOError:
 # Silence the default case to avoid people thinking something is wrong.
 # Only people who care about signing will know what it means, anyway,
 # and they can check for the positive acknowledgement above.


cores/esp8266/cont_util.cpp
```diff
@@ -73,7 +73,7 @@ void cont_repaint_stack(cont_t *cont)
     register uint32_t *sp asm("a1");
     // Ensure 64 bytes adjacent to the current SP don't get touched to endure
     // we don't accidentally trounce over locals or IRQ temps.
-    uint32_t sp_safe = CONT_STACKSIZE/4 - ((sp - &cont->stack[0] - 64)/4);
+    uint32_t sp_safe = CONT_STACKSIZE/4 - (sp - &cont->stack[0]) - 64/4;
```
devyte: Please check the sign of the arithmetic

earlephilhower: Yeah, mybad.  Just rewrote to be saner and have a quick test:
````
void __attribute__((noinline)) bigstack() {
  char buff[1024];
  memset(buff, 0xee, 1024);
  int sum=0;
  for (int i=0; i<1024; i++) {
    sum += buff[i];
  }
  Serial.printf("sum: %d\n", sum);
}

void setup() {
  Serial.begin(115200);
  Serial.printf("StackFree 1: %d\n", ESP.getFreeContStack());
  ESP.resetFreeContStack();
  Serial.printf("StackFree 2: %d\n", ESP.getFreeContStack());
  bigstack();
  Serial.printf("StackFree 3: %d\n", ESP.getFreeContStack());
  ESP.resetFreeContStack();
  Serial.printf("StackFree 4: %d\n", ESP.getFreeContStack());
}

void loop() {}
````
gives
````
StackFree 1: 3280
StackFree 2: 3988
sum: 243712
StackFree 3: 2256
StackFree 4: 3988
````

PR #5861, opened 2019-03-10
Reimplement SD.h write methods exactly in File

### LINKED ISSUE #5846: Compile error on Linux due to new File::write function
See also this discussion: https://github.com/esp8266/Arduino/commit/b1da9eda467cc935307d553692fdde2e670db258#r32622483

@devyte [asked](https://github.com/esp8266/Arduino/commit/b1da9eda467cc935307d553692fdde2e670db258#r32626017) to open an issue about this.

In pull request https://github.com/esp8266/Arduino/pull/5525 there has been a new function added to the `File` class:
https://github.com/esp8266/Arduino/blob/b1da9eda467cc935307d553692fdde2e670db258/cores/esp8266/FS.h#L82-L83

This function causes compiler issues on Linux.
```
In file included from /home/gijs/GitHub/letscontrolit/ESPEasy/src/ESPEasy.ino:86:0:
/home/gijs/GitHub/letscontrolit/ESPEasy/src/ESPEasyStorage.ino: In function 'String ClearInFile(char*, int, int)':
/home/gijs/GitHub/letscontrolit/ESPEasy/src/ESPEasyStorage.ino:679:30: error: call of overloaded 'write(unsigned int)' is ambiguous
SPIFFS_CHECK(f.write(0), fname);
^
```

In short, this new function makes it impossible to use the existing function `write(uint8_t)`
https://github.com/esp8266/Arduino/blob/b1da9eda467cc935307d553692fdde2e670db258/cores/esp8266/FS.h#L55-L56

earlephilhower:
Replace the individual override with the existing SD.h File's
implementation for all methods of File::write.

Fixes #5846

### DIFF
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -54,6 +54,11 @@ public:
     // Print methods:
     size_t write(uint8_t) override;
     size_t write(const uint8_t *buf, size_t size) override;
+    size_t write(int8_t t) { return write((uint8_t)t); }
+    size_t write(int16_t t) { return write((uint8_t)(t & 0xff)); }
+    size_t write(int32_t t) { return write((uint8_t)(t & 0xff)); }
+    size_t write(uint16_t t) { return write((uint8_t)(t & 0xff)); }
+    size_t write(uint32_t t) { return write((uint8_t)(t & 0xff)); }
 
     // Stream methods:
     int available() override;
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -54,6 +54,7 @@ public:
     // Print methods:
     size_t write(uint8_t) override;
     size_t write(const uint8_t *buf, size_t size) override;
+    // These handle ambiguity for write(0) case
     size_t write(int8_t t) { return write((uint8_t)t); }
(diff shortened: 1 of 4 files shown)


cores/esp8266/FS.h
```diff
@@ -54,6 +54,12 @@ class File : public Stream
     // Print methods:
     size_t write(uint8_t) override;
     size_t write(const uint8_t *buf, size_t size) override;
+    // These handle ambiguity for write(0) case
+    size_t write(int8_t t) { return write((uint8_t)t); }
```
devyte: Do we want these here or in the Print class?
Given that HardwareSerial, which inherits from Stream, and therefore Print, I think these should be put in Print, and HardwareSerial should also be cleaned up, because the write methods there do pretty much the same thing.

earlephilhower: Done for both.  Seems to build fine in Arduino and host tests.

earlephilhower:
I did start to convert the casts to `static_cast` but gave up when I saw the rest of the code in the file (untouched Arduino.cc source) uses the plain casts as well.  It's at least consistent now, if not as nice as we'd like.

PR #5870, opened 2019-03-13
polledTimeout: add option to use CPU count instead of millis()

d-a-v:


cores/esp8266/PolledTimeout.h
```diff
@@ -108,15 +133,25 @@ class timeoutTemplate
...
   
   timeType _timeout;
   timeType _start;
 };
 
-using oneShot = polledTimeout::timeoutTemplate<false>;
-using periodic = polledTimeout::timeoutTemplate<true>;
+using oneShotMs        = polledTimeout::timeoutTemplate<false>;
+using periodicMs       = polledTimeout::timeoutTemplate<true>;
+using oneShotMsYield   = polledTimeout::timeoutTemplate<false, YieldPolicy::YieldOrSkip>;
```
devyte: Careful here, I named the policy YieldOrSkip on purpose due to the delay(0), meaning that the object will yield if it can (CONT) or do nothing otherwise (SYS). If you call the type periodicMs"Yield", it could conflict if ever there is a policy added that does not skip and forcibly yields (calls yield(), which will panic if it's ever called from SYS).

cores/esp8266/PolledTimeout.h
```diff
@@ -47,15 +47,40 @@ struct YieldOrSkip
...
+} //TimePolicy
+
+template <bool PeriodicT, typename YieldPolicyT = YieldPolicy::DoNothing, typename TimePolicyT = TimePolicy::TimeMillis>
 class timeoutTemplate
 {
 public:
-  using timeType = decltype(millis());
+  using timeType = typename TimePolicyT::timeType;
   
   timeoutTemplate(timeType timeout)
```
devyte: Given that you now implemented toMillis, the timeout constructor argument is now restricted to ms units. This merits a Big Phat Comment here.

cores/esp8266/PolledTimeout.h
```diff
@@ -47,15 +47,40 @@ struct YieldOrSkip
 
 } //YieldPolicy
 
+namespace TimePolicy
+{
 
-template <bool PeriodicT, typename YieldPolicyT = YieldPolicy::DoNothing>  
+struct TimeMillis
```
devyte: The TimeMillis and TimeCycle policies have radically different rollover constraints. I think it merits implementing something like a static constexpr timeType maxTime(), or timeMax(), or just max(), and the timeoutTemplate constructor argument should be checked against it.
Up for discussion: if the constructor timeout argument is bigger than max, what should the behavior be?
Because it's an error condition on construction, the usual approach is to throw a C++ Exception, but we don't have them enabled by default yet, so we can't do that.
Maybe a debug msg and setting _timeout to 0 to force construction as expired?

libraries/esp8266/examples/BlinkPolledTimeout/BlinkPolledTimeout.ino
```diff
@@ -39,7 +39,7 @@ void ledToggle() {
 
 
 
-esp8266::polledTimeout::periodic halfPeriod(500); //use fully qualified type and avoid importing all ::esp8266 namespace to the global namespace
+esp8266::polledTimeout::periodicCpu halfPeriod(500); //use fully qualified type and avoid importing all ::esp8266 namespace to the global namespace
```
devyte: Did you change this just to have a cpu type in an example (CI coverage), or was there some other reason?

d-a-v:
> Is there some way to check for an overflow timeout when using CPU cycles?

I'm afraid not.

rollover is 26 seconds @160Mhz. `timeMax()`  is introduced, set to 13s. (or 26 @80Mhz)
comments added.

Misleading type names are removed (YieldOrSkip).
Unclear names `*Cpu` are renamed do `*CycleMs` so it now clearly needs milliseconds.

> Con:
> * Rollover time is drastically reduced from something like several days to a minute or so for 80MHz, and half that for 160MHz.

Yes, comments are added to use it only when it is known to be checked often

> *Neither Pro nor Con:
> * CPU busy waiting will be more or less the same
> * Power usage will be more or less the same

Yes, it is not intended for busy waiting but for periodic testing only. Maybe I should remove `oneShotCycleMs` ?

> * Usage is the same

This is a Pro :)

About performance, `millis()` uses 356 cpu cycles, while the other one only 3 or 4 cpu cycles.
I intend to extend `ScheduleFunctions` with periodic calls. Periodicity will also optionally be checked from inside `yield()` (only from cont stack of course). I could'nt resolve me to use phat `millis()` there, at every yield/loop. CycleMs does the same job about 100 times faster.

cores/esp8266/PolledTimeout.h
devyte: About specifying the timeout in ms:
My original intent was to not force a unit for the given timeout value, especially not ms.
Your intent seems to be higher precision, but the minimum argument value is 1ms, which equates to (I think) 160.000 cycles @160MHz. So you have high precision, but very low resolution.
Are you absolutely certain you want to force the constructor arg to be in ms?
Consider allowing the arg to be in number of cycles. You could then implement a static constexpr fromMillis(), which would then allow usage like this:
```
using oneShotCycles = esp8266::polledTimeout::oneShotCycles;
oneShotCycles timeout(oneShotCycles::fromMillis(500)); //500ms value gets converted to number of cycles and passed to the construction
```
With this, if a user wants the full per-cycle resolution, he can still provide a constructor arg in number of cycles.

d-a-v: That wasn't the primary goal for this PR, I primarily needed `ms` resolution with a lighter test (in terms of cpu load).
But yes that's a fair notice, I'll add it as a third use case.

cores/esp8266/PolledTimeout.h
```diff
@@ -47,16 +47,91 @@ struct YieldOrSkip
 
 } //YieldPolicy
```
devyte: Since you're already playing with the code here, could I ask you to please add the following:
```cpp
template <unsigned long delayMs>
struct YieldAndDelay
{
  static void execute() {delay(delayMs);}
};
```

It essentially means "poll every delayMs" as opposed to the current "poll as fast as possible". The use case is to reduce the busy wait loop CPU time, which reduces power usage. I've seen the use pattern in some places, hence the need.

Example for building a type from this (untested):
```cpp
using periodicFast50Ms = esp8266::polledTimeout::timeoutTemplate<true, YieldPolicy::YieldAndDelay<50>, TimePolicy::TimeFastMillis>;
periodicFast50Ms timeout(500); //poll every 50ms, expire at 500ms
```

cores/esp8266/PolledTimeout.h
```diff
@@ -89,14 +173,24 @@ class timeoutTemplate
   
   bool checkExpired(const timeType t) const
   {
-    return (t - _start) >=  _timeout;
+    // always returns false when _timeout is ::max()
+    return (t - _start) > _timeout;
```
devyte: I understand why you did this, but it means the timeout calc is now off by 1 unit, e.g. if you construct with timeout 10 it won't timeout at 10, but at 11.

devyte: How about this:
```cpp
(t - _start) % TimePolicyT::timeMax() > _timeout;
```

cores/esp8266/PolledTimeout.h
```diff
@@ -45,18 +47,86 @@ struct YieldOrSkip
...
+  using timeType = typename TimePolicyT::timeType;
+
+  static constexpr timeType neverExpires = std::numeric_limits<timeType>::max();
+
+  timeoutTemplate(const timeType userTimeout)
+  {
+    reset(userTimeout);
+  }
 
   bool expired()
```
devyte: If expiredRetrigger() and expiredOneShot() are in IRAM, then expired() and bool() probably should be as well.
BTW, why are any methods in IRAM? What is the use case for calling from an ISR?

cores/esp8266/PolledTimeout.h
devyte: Note for future ref: the ESP8266 does not have a hw div instruction, hence operator % is expensive. Instead, pay the price of an extra flag in memory.

cores/esp8266/PolledTimeout.h
d-a-v: > why are any methods in IRAM? What is the use case for calling from an ISR?

It's about speed, not ISR requirement, so this comment remains true:
```
// "Fast" versions sacrifices time range for improved precision and reduced execution time (by 86%)
// (cpu cycles for ::expired(): 372 (millis()) vs 52 (ESP.getCycleCount()))
```

cores/esp8266/PolledTimeout.h
```diff
@@ -45,19 +47,78 @@ struct YieldOrSkip
...
+template <typename TimeSourceType, unsigned long long second_th, unsigned long long rangeCompensate = 1>
+struct TimeUnit
+{
+  using timeType = typename TimeSourceType::timeType;
+
+  static constexpr timeType user2UnitMultiplierMax = (TimeSourceType::ticksPerSecondMax * rangeCompensate) / second_th;
+  static constexpr timeType user2UnitMultiplier    = (TimeSourceType::ticksPerSecond    * rangeCompensate) / second_th;
+  static constexpr timeType user2UnitDivider       = rangeCompensate;
+  static constexpr timeType timeMax                = (std::numeric_limits<timeType>::max() - 1) / user2UnitMultiplierMax;
+  static constexpr timeType neverExpires           = std::numeric_limits<timeType>::max();
```
devyte: neverExpires and alwaysExpired are not specific to the TimeUnit policy. I suggest removing them from here and defining them directly in in timeoutTemplate

cores/esp8266/PolledTimeout.h
```diff
@@ -45,19 +47,78 @@ struct YieldOrSkip
...
+  // this particular time measurement is intended to be called very often
+  // (every loop, every yield)
+
+  using timeType = decltype(ESP.getCycleCount());
+  static timeType time() {return ESP.getCycleCount();}
+  static constexpr timeType ticksPerSecond    = F_CPU;
+  static constexpr timeType ticksPerSecondMax = 160000000; // Mhz
+};
+
+template <typename TimeSourceType, unsigned long long second_th, unsigned long long rangeCompensate = 1>
```
devyte: I think second_th and rangeCompensate need a Big Phat Comment explaining what they are. 
BTW, if I understand this correctly, rangeCompensate should be > 1 for cases where the TimeUnit is finer than the timeSource ticksPerSecond. In the case of nanos: 1ns < 1/ticksPerSecond => rangeCompensate = 25.
If the previous is true, it seems to me that rangeCompensate could be automagically calculated. In the case of nanos, it's 2*(second_th/TimeSourceType::ticksPerSecond).

PR #5882, opened 2019-03-15
include submodules with git clone --recursive

Juppit:
Since we have some submodules it is recomended to include them while cloning the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Documentation: [https://arduino-esp8266.readthedocs.io/en/2.5.0/](https://arduin
 cd hardware
 mkdir esp8266com
 cd esp8266com
-git clone https://github.com/esp8266/Arduino.git esp8266
+git clone --recursive https://github.com/esp8266/Arduino.git esp8266
 ```
 - Download binary tools (you need Python 2.7)
 ```bash
--- a/README.md
+++ b/README.md
@@ -57,7 +57,9 @@ Documentation: [https://arduino-esp8266.readthedocs.io/en/2.5.0/](https://arduin
 cd hardware
 mkdir esp8266com
 cd esp8266com
-git clone --recursive https://github.com/esp8266/Arduino.git esp8266
+git clone https://github.com/esp8266/Arduino.git esp8266
+cd esp8266
+git submodule update --init
 ```
 - Download binary tools (you need Python 2.7)
 ```bash


d-a-v:
This will include sub-sub modules like full lwIP repo, which is not necessary.
I recommend
```
git clone https://github.com/esp8266/Arduino.git esp8266
git submodule update --init
```
(and not `git submodule update --init --recursive`, just like in `.travis.yml`)

Juppit:
@d-a-v 
My question is: do we know in each case whether sub-sub-modules are **not** needed?
Or is that only in this case?
I urge that the maintainer of the submodule decide what is needed.
So in this particular case, I would ask @d-a-v if he can remove the submodule upstream.
Or we decide here, e.g. via an entry in  [esp8266\ .gitignore](https://github.com/esp8266/Arduino/blob/master/.gitignore)

d-a-v:
> do we know in each case whether sub-sub-modules are not needed?

Currently there are two sub-sub-modules: lwIP and BearSSL.
None of them are necessary for building sketches, because we already have precompiled libs in the core repository.
Even more, submodules "lwip2" and "bearssl adapter" (their parent) are not even needed for the same reason.

>  I would ask @d-a-v if he can remove the submodule upstream.

I don't understand why it is needed.
Goal of sub-(sub-)modules is to rebuild libs from original sources.
We were automatically pulling them but sometimes it fails (because not hosted on github and their respective website are sometimes down, which was breaking CI).

> Or we decide here, e.g. via an entry in esp8266\ .gitignore

I don't know the implication of this, I'd think it would break rebuilding
(`cd tools/sdk/lwip2|bearssl; make install` will update both sub- and sub-sub-module and rebuild)

Juppit:
@d-a-v 
> Currently there are two sub-sub-modules: lwIP and BearSSL.
None of them are necessary for building sketches, because we already have precompiled libs in the core repository.
Even more, submodules "lwip2" and "bearssl adapter" (their parent) are not even needed for the same reason.

This part of [Readme.md](https://github.com/esp8266/Arduino#using-git-version-basic-instructions) provides a standard procedure for creating the full repository from git.
For people only using it by Arduino or Platformio there is no need to do this.

>>    I would ask @d-a-v if he can remove the submodule upstream.

I did not want to go that far with Upstream, I just meant your [repo](https://github.com/d-a-v/esp82xx-nonos-linklayer). :-)

>> Or we decide here, e.g. via an entry in esp8266\ .gitignore

> I don't know the implication of this, I'd think it would break rebuilding
(cd tools/sdk/lwip2|bearssl; make install will update both sub- and sub-sub-module and rebuild)

At this point, I see the same problem when we exclude sub-sub-modules with "git submodule update --init".

Cons: In the specific case, an overhead of a few MByte unused sources
Pros: Avoid unknown implications of future submodule versions

The right decision can actually only be made within the sub-module
specifically by @d-a-v  and @earlephilhower in there repos, but not in the more global recipe in the Readme.

So my opinion is clear: pro "git clone --recursive".

Your knowledge of Github is certainly greater than mine, as I take your various tips for easy patching the github sources in this repo.

d-a-v:
> > This will include sub-sub modules like full lwIP repo, which is not necessary.

> So my opinion is clear: pro "git clone --recursive".
> `git clone --recursive https://github.com/esp8266/Arduino.git esp8266`

Well you are right, it is documentation, I *suggest* adding this for completeness:

```
# mandatory:
git clone https://github.com/esp8266/Arduino.git esp8266
git submodule update --init

# or: with source for precompiled libraries: (fetching more repositories)
git clone --recursive https://github.com/esp8266/Arduino.git esp8266
# same as:
git clone https://github.com/esp8266/Arduino.git esp8266
git submodule update --init --recursive
```

PR #5894, opened 2019-03-19
Fix OTA in AP mode

zacharydrew:
OTA is broken in AP mode because ESP8266mDNS is checking whether the station is connected before processing. Change the logic to WiFi.isConnected() OR WiFi.softAPgetStationNum()>0 fixes the issue.

### DIFF
--- a/libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp
+++ b/libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp
@@ -79,9 +79,10 @@ bool MDNSResponder::_process(bool p_bUserContext) {
         }
     }
     else {
-        bResult = ((WiFi.isConnected()) &&          // Has connection?
-                   (_updateProbeStatus()) &&        // Probing
-                   (_checkServiceQueryCache()));    // Service query cache check
+        bResult = ((WiFi.isConnected() ||               // Either station is connected
+                    WiFi.softAPgetStationNum()>0) &&    // Or AP has stations connected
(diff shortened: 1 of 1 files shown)


d-a-v:
cc @LaborEtArs @hreintke

Lan-Hekary:
This commit is the source of this Issue #5915
I Changed the file back to what it was and It fixed the Issue for me ( Please consider another alternative to fix the Original Issue)
I guess this commit is somehow overloading the ESP, as it does not respond to PINGs or TCP requests ..

devyte:
@zacharydrew are you absolutely sure you reverted _only_ this specific change and _nothing_ else across the entirestreet core you're using?
@LaborEtArs do you see anything wrong with this commit? It looks correct to me.

Lan-Hekary:
I opened wireshark ..
And I noticed that the ESP is flooding the network with a query of itself!!!
That is crazy !!
It's overloading the network .. And this is the source of this issue #5915 ..
the 50ms delay I put there allow the AP network a oom to breathe and process the rest of the requests ..

The Flooding stops once it connects to a Station ..
But it starts again once I disconnect ..
And if there is no station it just keeps flooding the network with queries ..

Lan-Hekary:
The flooding only happens if the wifi mode is AP_STA and there is no connection to a WiFi Station ..
If the Mode changes to AP only .. It works Fine ..
And fixes the Problem of the OTA indeed .. Now I am able to upload to the ESP through Hostname.local ..

devyte:
Let's discuss in the issue, please add this info there.

PR #5897, opened 2019-03-20
Fix String::replace()

### LINKED ISSUE #5883: Some strings seem to have bogus data at the end (core 2.6.0 vs. core 2.4.2)

earlephilhower:
Fixes #5883 and supercedes #5890

The replace() function was using len() while in the middle of buffer
operations.  In SSO mode len() is not stored separately and is a call to
strlen(), which may not be legal if you're in the middle of overwriting
the SSO buffer, as was the case in ::replace when the replacement string
was longer than the find string.  This caused potential garbage at the
end of the string when accessed.  Instead, just cache the length in a
local while doing the operation.

Add in test cases from #5890 as well as some new ones that fail on the
unmodified core.

### DIFF
--- a/tests/host/core/test_string.cpp
+++ b/tests/host/core/test_string.cpp
@@ -383,9 +383,9 @@ TEST_CASE("String SSO handles junk in memory", "[core][String]")
 {
   // We fill the SSO space with garbage then construct an object in it and check consistency
   // This is NOT how you want to use Strings outside of this testing!
-  unsigned char space[16];
+  unsigned char space[64];
   String *s = (String*)space;
-  memset(space, 0xff, 16);
+  memset(space, 0xff, 64);
   new(s) String;
   REQUIRE(*s == "");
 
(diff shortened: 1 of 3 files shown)


earlephilhower:
Darn it.  I did run "make test" and it passed on my box before submitting.  Let me check the code once again, there may be other spots where it's iffy.

tests/host/core/test_string.cpp
```diff
@@ -360,4 +366,65 @@ TEST_CASE("String SSO works", "[core][String]")
...
+void repl(const String& key, const String& val, String& s, boolean useURLencode)
+{
+    s.replace(key, val);
+}
+
+
+TEST_CASE("String SSO handles junk in memory", "[core][String]")
+{
+  // We fill the SSO space with garbage then construct an object in it and check consistency
+  // This is NOT how you want to use Strings outside of this testing!
```
TD-er: Always nice to see your own code commented with some instruction "This is NOT how you want to use...." ;)

PR #5898, opened 2019-03-20
Fix boards-manager install issues on Linux and Windows

### LINKED ISSUE #5881: python.exe": file does not exist
### Basic Infos

#### Platform

- Hardware:					[ESP12]
- Core Version:			[2.6.0-dev-nightly+20190315]
- Development Env:		[Arduino IDE 1.8.8]

### Settings in IDE

- Module:  			[LOLIN(Wemos) D1 mini r2]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [|v2 Lower Memory]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600] 

### Problem Description

Building any sketch I get the error:
`"C:\\Users\\Peter\\AppData\\Local\\Arduino15\\packages\\esp8266\\hardware\\esp8266\\2.6.0-dev-nightly+20190315/tools/python/python.exe": file does not exist`

The file python.exe is not in the tools directory within 2.6.0-dev..., it is in the directory
`C:\Users\Peter\AppData\Local\Arduino15\packages\esp8266\tools\python`

My build steps from Cygwin on Windows 10:

1. clone https://github.com/esp8266/Arduino.git
2. cd esp8266/package
3. ./build_boards_manager_package.sh
4. cp package_esp8266com_index.json from versions/2.6.0-dev... to versions/package_esp8266com_index.json
5. setup the boardmanager URL in the Arduino-IDE to `http://localhost:8000/versions/package_esp8266com_index.json`
6. start a local http-server to delever the package to the Arduino-IDE
7. start the boardmanager from Arduino tools-menu and download the esp8266 version 2.6.0-dev...

One note: to build step 3, there is jq.exe in version 1.6 necessary

earlephilhower:
Fixes #5881

The path to python.exe on Windows needs to change depending on whether
it is installed via the boards manager or GIT (similar to the compiler
paths).  Adjust accordingly.

An empty "python" directory will be created by the boards-manager
installer in Linux to avoid "tool not available" errors, and it will contain a symlink
to the real system Python executable.

Tested under Linux GIT, Linux board-manager, Windows git, and
Windows board-manager.


earlephilhower:
@devyte , @d-a-v this change introduces a "tools/python/python" symlink to "/usr/bin/python" on UNIX-like OSes.

The alternative is to make that symlink a simple wrapper (either a bash file "python $*" or even a python file that calls itself with the passed-in args.

This is required because we need a "python" tool in the package.JSON file.  Its existence causes the Arduino boards.txt to override the {runtime.tools.python.path} var, and that can't be overridden by a boards.txt value (as it is done in GIT builds).

So you need something called "python" in that python tool.tar.gz file, and that's what the IDE is going to call.  The IDE basically forces it when installed via boards-manager.

Issue #5902, opened 2019-03-21
Backport patch from lwip2 to fix lwip1.4 behaviour with MTU<1500


mcspr:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:				ESP-12
- Core Version:			[2019/03/21 git  8961bba]
- Development Env:		Arduino IDE
- Operating System:		Any

### Settings in IDE

- Module:  		Generic ESP8266 Module
- lwip Variant:   **v1.4**
- Flash Mode:			dout
- Flash Size: 		4MB
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200 (serial upload only)

### Problem Description

Using ESPAsyncTCP, ESPAsyncWebServer, AsyncWebSocket, lwip v1.4
It is impossible to receive data-chunk larger than MSS size via websockets if MTU of the client device is different from 1500. I can see data is sent to the router, but fragmenting does not work properly, so router refuses to send the data further. Surprisingly, HTML is sent properly.

I was able to find one related issue:
https://github.com/esp8266/Arduino/issues/2437

This was already fixed in lwip2. Related issue + fix:
http://savannah.nongnu.org/bugs/?46384
http://git.savannah.gnu.org/cgit/lwip.git/commit/src?id=8e8571da6a6771f9d2d82bbd0b5a6c27474ce0fc
Can we apply this patch to the local tree here and rebuild `liblwip_gcc.a` so pre built option works too?

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

Gist of the sketch, to avoid blowing up the issue with long strings:
https://gist.github.com/mcspr/00e95d3ea8d680aeaa04cbe8a707ba8d

### Debug Messages

On Windows machine with MTU of 1492 (meaning, MSS is auto calculated to be 1452)
```
> netsh interface ipv4 set subinterface "WLAN" mtu=1492
```

In browser, using lwip v1.4:
```
[...]

Issue #632, opened 2015-07-30
Ephemeral source port not random


owendelong:
When using the WiFiClient library as follows:

```
#include <ESP8266WiFi.h>
#define MY_SSID "ssid"
#define MY_WFPASSWORD "wpa2_key"

WiFiClient client;

void setup() {
  WiFi.begin(MY_SSID, MY_WFPASSWORD);
  while (WiFi.status() != WL_CONNECTED)
  {
    delay(50);
  }
  client.connect(HOST, HTTPPORT);
}
```

The source port utilized is always "patrolview" or 4097.

This can create a deadly embrace situation if the device is being used to log data and resets itself as a means to try again without delay on certain connection errors. It should be randomized so that a different port is used for each call to connect, even if the device is reset/rebooted between calls.

The problem that can occur is a web server which is in one of the wait states for closing on a previous attempt will see the new attempt to connect as matching the previous one since the SRC_ADDR,SRC_PORT,DST_ADDR,DST_PORT tuple is identical. This results in a retransmission of the previous FIN packet and a reset of the wait timer on the server side. So long as the ESP8266 continues to retry within before the timer expires (often these are ~5 minutes or so), it will be unable to initiate a new connection to the server without being mistaken for the old one.

The simplest solution is to ad code to WiFiClient::connect() that will randomize the local port number.

I will attempt to provide a workable patch shortly.

owendelong:
Additional information, looks like this problem is actually in the tcp_connect function in the underlying lwip library.

igrr:
There is a static method in WiFiClient class, static void
setLocalPortStart(uint16_t port);
Find a random number generator which suits your requirements, seed it with
something (ADC reading perhaps?) and call this function somewhere in your
setup() method before you initialize WiFiClients.

owendelong:
Thanks, Igrr.

I tracked the problem down to this function in the LWIP library:
```
/**
 * Allocate a new local TCP port.
 *
 * @return a new (free) local TCP port number
 */
static u16_t
tcp_new_port(void)
{
  u8_t i;
  u16_t n = 0;
  struct tcp_pcb *pcb;
  
again:
  if (tcp_port++ == TCP_LOCAL_PORT_RANGE_END) {
    tcp_port = TCP_LOCAL_PORT_RANGE_START;
  }
  /* Check all PCB lists. */
  for (i = 0; i < NUM_TCP_PCB_LISTS; i++) {
[...]

PR #5906, opened 2019-03-22
lwIP-v2: new patch to randomize tcp source ports

d-a-v:
ref: https://github.com/d-a-v/esp82xx-nonos-linklayer/issues/31
origin: #5902 
https://github.com/me-no-dev/ESPAsyncTCP/issues/108

Thanks @mcspr !


d-a-v:
Following the links above is instructive.
To summarize:
- currently and from a long time lwIP tcp client connections always uses the same *tcp source port number* right after boot
- this port number is increased  everytime a new one is needed (= new tcp client connection)
  (to be noted, linux has the same increasing behavior)
- when connecting to the same server (right after boot), the triplet (esp-ip-address, source port, destination port) are the same, and may hit remote server list of sockets in time-wait-state (previous connection unproperly closed from the same esp). Consequently the new connection fails when it happens.
- this is happening only when debugging (esp reboots often, in less time than time-wait expiration), so the nasty effect is amplified especially when bugs are being chased
- efforts had been done when espressif's lwIP implementation wasn't open source, with `WiFiClient::setLocalPortStart()` #632 but it must be explicitely called with a different random number at every reboot. Efficient but not ideal.

This PR uses espressif firmware's `r_rand()` everytime a new local source port is needed. A different source port number is now showed by tcpdump right after boot. Source port range and duplication is verified everytime in lwIP's `src/core/tcp.c:tcp_new_port()`. It is implemented as a local patch for upstream lwIP so it is valid not only with `WiFiClient` but also with @me-no-dev's `Async` libraries (they don't use `WiFiClient`).

`WiFiClient::setLocalPortStart()` is still usable with the same effects as before.

TD-er:
I guess this can also lead to issues when a connection attempt is retried, when the ESP misses the acknowledgement of a successful connection attempt?

d-a-v:
> I guess this can also lead to issues when a connection attempt is retried, when the ESP misses the acknowledgement of a successful connection attempt?

If I understand the question, before of after this patch, nothing is changed about TCP connection attempts.
A connecting request is uniquely identified by a quadruplet (source ip & port, destination ip & port). They *remain constant* during all the time of this connection, and missed ACK for SYN are retransmitted on timeout some times (like any other ACKs) before giving up.
Source ports are meaningless (except for identifying an incoming packet and their uniqueness among all tcp connsctions on an host) and supposed to be random. This PR only changes the way the source port is worked out for every *new* connection.

TD-er:
I understand that it does so on a new connection.
I was thinking about whether the current situation could also lead to the issues you try to fix with this patch.
For example, what if the ESP tries to make a connection and it assumes the connection is not successful, for some reason (e.g. missing ack packets)
In the current situation, could it then also lead to these situations?

The use case I have in mind is that I've seen several issues when connecting to a MQTT broker.
It may retry almost indefinitely, but it may succeed when you stop trying for a while and make a new attempt. Another "fix" (which introduces other issues) was to use a new client ID when connecting to the broker.

d-a-v:
> It may retry almost indefinitely, but it may succeed when you stop trying for a while and make a new attempt.

Who is, or how is it retrying,
* your sketch with a new `WiFiClient` for each retry ?
  before or after this PR, a new source port is used
* Let the same `WiFiClient` continuously try (so the timeout is too long) ?
  This behaviour is not addressed nor changed by this PR
* a reboot ?
  After a reboot, before this PR, the same source port was used (leading to an issue described in OP's links and second message)

> but it may succeed when you stop trying for a while and make a new attempt

That looks like a firewall policy. They check for SYN frequency and block the originating IP address for a short (=lucky) or long (=hair-pulling debugging session) period of time. It's an anti-DOS protection.

TD-er:
Hmm, I may have to look at the Domoticz Raspberry Pi image then to see if it uses some kind of filtering like that.  Never thought of that option.
Not sure if it is making a new instance of WiFiClient on each reconnect attempt, but I think it does.

I will look into the firewall option.
If that does rise new questions/issues I will open a new issue and stop polluting your PR :)

PR #5917, opened 2019-03-26
PlatformIO: generate core_version.h when using feature/stage

mcspr:
Original issue: https://github.com/platformio/platform-espressif8266/issues/101
Doc: https://docs.platformio.org/en/latest/platforms/espressif8266.html#using-arduino-framework-with-staging-version

Update platformio-build.py:
- extract version from platform.txt (I see no other definitive source for this info)
- conditionally run makecorever.py script with the same parameters as IDE does 
(however, I am not exactly sure if VerboseAction is a right way to call it, because dependency on git dir sometimes does not work properly)

Update makecorever.py:
- fix ambiguous bytes+string error with p.stdout.readlines() data
- check `if __name__ == "__main__"`, move generation into a function to clearly see git commands
- replace `git --git-dir` with just `git -C`, since we already know that .git dir is there
(unless I am missing some reason for git-dir setting. if it is required, it should probably be `--git-dir=.git --work-tree=. -C ...framework...`)


ivankravets:
Thanks!

PR #5922, opened 2019-03-27
Split IRAM into 2 linker sections to move std::fcn

### LINKED ISSUE #5914: Hspi in slave mode with Wifi enabled produce crash dump
### Basic Infos

#### Platform

- Hardware:					[ESP-8266-wroom2]
- Operating System:		[Ubuntu]

### Settings in IDE

- Flash Size: 		[2MB]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]

### Problem Description
Hi all,

We try to get the spi in slave mode working without crash and we cannot get it. We use 2 esp8266 (one in master and one in slave). We use only the read status command to simplify the setup.  Here the Master code. We use the Example SPI Safe Master Demo and modify it to only use the readStatus method.

```cpp
#include <SPI.h>

class ESPSafeMaster {
  private:
    uint8_t _ss_pin;
    void _pulseSS() {
      digitalWrite(_ss_pin, HIGH);
      delayMicroseconds(5);
      digitalWrite(_ss_pin, LOW);
    }
  public:
    ESPSafeMaster(uint8_t pin): _ss_pin(pin) {}
    void begin() {
      pinMode(_ss_pin, OUTPUT);
      _pulseSS();
    }

    uint32_t readStatus() {
      _pulseSS();
      SPI.transfer(0x04);
      uint32_t status = (SPI.transfer(0) | ((uint32_t)(SPI.transfer(0)) << 8) | ((uint32_t)(SPI.transfer(0)) << 16) | ((uint32_t)(SPI.transfer(0)) << 24));
      _pulseSS();
      return status;
    }

    void writeStatus(uint32_t status) {
      _pulseSS();
      SPI.transfer(0x01);
      SPI.transfer(status & 0xFF);
      SPI.transfer((status >> 8) & 0xFF);
      SPI.transfer((status >> 16) & 0xFF);
      SPI.transfer((status >> 24) & 0xFF);
      _pulseSS();
    }

    void readData(uint8_t * data) {
      _pulseSS();
      SPI.transfer(0x03);
      SPI.transfer(0x00);
      for (uint8_t i = 0; i < 32; i++) {
        data[i] = SPI.transfer(0);
      }
      _pulseSS();
    }

    void writeData(uint8_t * data, size_t len) {
      uint8_t i = 0;
      _pulseSS();
      SPI.transfer(0x02);
      SPI.transfer(0x00);
      while (len-- && i < 32) {
        SPI.transfer(data[i++]);
      }
      while (i++ < 32) {
        SPI.transfer(0);
      }
      _pulseSS();
    }

    String readData() {
      char data[33];
      data[32] = 0;
      readData((uint8_t *)data);
      return String(data);
    }

    void writeData(const char * data) {
      writeData((uint8_t *)data, strlen(data));
    }
};
ESPSafeMaster esp(SS);

void setup() {
  Serial.begin(115200);
  SPI.begin();
  esp.begin();
  delay(1000);
}

uint32_t status;

void loop() {
	esp.readStatus();
}
```
On the slave we use this code. 
```cpp
#include <Arduino.h>
#include <SPISlave.h>
#include <ESP8266WiFi.h>

void setup() {
	Serial.begin(115200);
	Serial.setDebugOutput(true);
	WiFi.mode(WIFI_STA);
	WiFi.begin("MySSID","MyPassword");
        delay(10000); //delay to not crash the ESP. If this delay is not there, the ESP will crash in sys context
	SPISlave.begin();
        SPISlave.setStatus(0xAA);
}

void loop() {
	
}
```
With this configuration, the data of the status is 0x55 (checked with an MSO) and as soon as the WiFi is connected, the data of the status is 0x00. No more 0x55 is returned. 

Does anyone can try the same setup and give feedback ? We have been able to have a reliable communication in SPI mode. 

Jonathan

earlephilhower:
Callbacks need to be placed in IRAM when being called from an IRQ (like
the SPISlave callbacks).

This can be done by hacking the std::functional header and making every
single specialization of the template into an IRAM section, which would
take a lot of space for no benefit in the majority of cases.

The alternate is to specify the single instantiation types/operators
required, but the problem is the flash segment matcher would match them
before the IRAM section was begun, and rules for them would just not be
applied.

Get around this by splitting the IRAM section definition into .text and
.text1.  This is linker syntactic sugar and does not actually change the
on-chip layout.  But it does allow us to put the exception vectors at
the required absolute addresses and add single functions to IRAM.

Fixes #5914 (maybe, may need to add other template types)


dumarjo:
@earlephilhower You will need all the types in RAM.

typedef std::function<void(uint8_t *data, size_t len)> SpiSlaveDataHandler;
typedef std::function<void(uint32_t status)> SpiSlaveStatusHandler;
typedef std::function<void(void)> SpiSlaveSentHandler;

tools/sdk/ld/eagle.app.v6.common.ld.h
```diff
@@ -127,8 +127,8 @@ SECTIONS
     *(.init.literal)
     *(.init)
 
-    /* Special functions/templates that need to be in IRAM (SPI/IRQ callbacks/etc. here */
-    *(.text._ZNKSt8functionIFvvEEclEv)  /* std::function<void ()>::operator()() const */
+    /* all functional callers are placed in IRAM (including SPI/IRQ callbacks/etc) here */
+    *(.text._ZNKSt8functionIF*EE*)  /* std::function<any(...)>::operator()() const */
```
earlephilhower: I think this stops evaluation at the first asterisk.  The regular expression support is very basic in the linker.

d-a-v: I quickly got lost in the big yacc parser.
So to verify I tweaked a test file object,
Before:
```
$ nm /tmp/arduino_build_491121/function.ino.elf|grep _ZNKSt8functio
401000b0 W _ZNKSt8functionIFviEEclEi
40100094 W _ZNKSt8functionIFvvEEclEv
```
Then I edited the object file and replaced `_ZNKSt8functionIFvvEEclEv` by `_ZNKSt8functionIFvvEXclEv` (EE->EX after the first asterisk), then relinked:
```
$ nm /tmp/arduino_build_491121/function.ino.elf|grep _ZNKSt8functio
40100094 W _ZNKSt8functionIFviEEclEi
402010c4 W _ZNKSt8functionIFvvEXclEv
```

Is this a sufficient validation ?

earlephilhower: Yes, that looks fine.

earlephilhower:
DO NOT MERGE.  I just tried the PR on a very simple example and got into an infinite boot loop.  Must've gotten lucky the first time I tried things.

Basically the elf is fine, but the BIN is not pulling in the `.text1` segment for actual image creation.

Need to update elf2bin.py...

earlephilhower:
Latest push actually uploads the app main code and runs on tests I've thrown at it.

devyte:
A warning here: only std::function::operator()() is moved to IRAM. The constructor is not. That means that re-attaching via the functional api from an ISR can still crash, because that will construct a temp std::functional and pass it in, and the constructor could potentially not be in IRAM.
I think it's a case rare enough to not warrant the price in IRAM.

hreintke:
@earlephilhower : 
Don't know whether it is intentional and/or it has any impact but when I was looking into the objdump output checking for functional interrupt I noticed.

From he title of the PR I expected to see both .text and .text1 segments. .text being all as before, .text1 for the specific std::functions.
But it looks like all gets into .text1.

part of objdump before the commit : 
```
40238048 g     F .irom0.text	00000038 periodic_cal
4021dc3c g     F .irom0.text	00000068 ieee80211_pwrsave
40105ff8 g     F .text	00000067 phy_get_bb_freqoffset
401062ec g     F .text	0000002b clockgate_watchdog
40102f28 g     F .text	0000000f lmacIsActive
40101180 g     F .text	00000022 pp_post2
3ffedd54 g     O .bss	00000004 timer_list
40204028 g     F .irom0.text	00000041 _ZN12UpdaterClassC1Ev
40103e34 g     F .text	00000026 lmacMSDUAged
00000220 g       *ABS*	00000000 _memmap_cacheattr_bp_base
402205d8 g     F .irom0.text	00000049 cnx_sta_connect_led_timer_cb
40105e70 g     F .text	00000151 tx_pwr_backoff_old
4022cd3c g     F .irom0.text	00000008 fpm_get_slp_type
40204374  w    F .irom0.text	0000009e _ZN8conststr15parseNthIntegerILj12EEEiRAT__Kcj
402391c8 g     F .irom0.text	0000000b phy_set_powerup_option
```

Same part after the commit

```
40238048 g     F .irom0.text	00000038 periodic_cal
4021dc3c g     F .irom0.text	00000068 ieee80211_pwrsave
40105ff8 g     F .text1	00000067 phy_get_bb_freqoffset
401062ec g     F .text1	0000002b clockgate_watchdog
40102f28 g     F .text1	0000000f lmacIsActive
40101180 g     F .text1	00000022 pp_post2
3ffedd54 g     O .bss	00000004 timer_list
40204028 g     F .irom0.text	00000041 _ZN12UpdaterClassC1Ev
40103e34 g     F .text1	00000026 lmacMSDUAged
00000220 g       *ABS*	00000000 _memmap_cacheattr_bp_base
402205d8 g     F .irom0.text	00000049 cnx_sta_connect_led_timer_cb
40105e70 g     F .text1	00000151 tx_pwr_backoff_old
4022cd3c g     F .irom0.text	00000008 fpm_get_slp_type
40204374  w    F .irom0.text	0000009e _ZN8conststr15parseNthIntegerILj12EEEiRAT__Kcj
402391c8 g     F .irom0.text	0000000b phy_set_powerup_option
```

PR #5937, opened 2019-04-01
sta_config_equal

dontsovcmc:
There is a bug with reconnect to the same WiFi after running Active Point.
Long description:  https://github.com/tzapu/WiFiManager/issues/782

sta_config_equal(conf_compare, conf) doesn't compare all fields.
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L160

This is a reason of failed connection for me. Maybe you can test it for another cases.


dontsovcmc:
How we can prevent future bugs?
two independent commits broke it: 
429f40d32185e7028eaa06cb471d7cef03d4612f  19.02.2018
8ef21ca3aeff76d1265a72ac106999ed9c631a7e 09.10.2018 

I see 2 ways:
1. comment in user_interface.h "please, add fields to 'here' and 'here'..."
2. inialize structures with 0. This way we can use memcmp function and be independent from fields.

d-a-v:
> How we can prevent future bugs?

I'm not sure.

Maybe by adding a comment in `user_interface.h`and/or by constexpr-checking `sizeof(the-structs-used-for-comparison)` against a constant (that has to be changed when more fields appears)

tablatronix:
Thanks @dontsovcmc I did not notice these updates, good catch, and yes this has been an issue in the past also when this struct changes.

I do like the struct integrity checking using a size constraint assert.

d-a-v:
I forgot to mention #5939 (keeping track of reasons)

PR #5941, opened 2019-04-02
add function IRAM address verifier, use it with SPISlave::

d-a-v:
related with https://github.com/esp8266/Arduino/issues/5914#issuecomment-476609107

### DIFF
--- a/cores/esp8266/debug.cpp
+++ b/cores/esp8266/debug.cpp
@@ -23,14 +23,23 @@
 
 void ICACHE_RAM_ATTR hexdump(const void *mem, uint32_t len, uint8_t cols) {
     const uint8_t* src = (const uint8_t*) mem;
-    os_printf("\n[HEXDUMP] Address: 0x%08X len: 0x%X (%d)", (ptrdiff_t)src, len, len);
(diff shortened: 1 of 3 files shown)


cores/esp8266/debug.cpp
```diff
@@ -23,14 +23,23 @@
...
         }
-        os_printf("%02X ", *src);
+        ::printf((PGM_P)F("%02X "), *src);
         src++;
     }
     os_printf("\n");
 }
+
+#ifndef NDEBUG
+void assert_iram (uintptr_t address) {
```
devyte: Arg should be const

libraries/SPISlave/src/SPISlave.cpp
```diff
@@ -93,18 +93,22 @@ void SPISlaveClass::setStatus(uint32_t status)
 }
 void SPISlaveClass::onData(SpiSlaveDataHandler cb)
 {
+    check_function_is_in_iram(cb);
```
devyte: This doesn't look right, the handler arg is received by value. Same for the other cases below.

d-a-v:
It seems function address cannot be extracted with `function<>::target` when RTTI is disabled.

PR #5948, opened 2019-04-04
Hardware SPI BREAKING: unswap mode 2 and 3 for portability

### LINKED ISSUE #2416: SPI Library: SPI Mode 2 and 3 are wrong
First: Thanks a lot for the ESP8266 board package for the Arduino IDE. Great Work.
#### Hardware

Hardware:           Adafruit Feather HUZZAH ESP 8266 
Core Version:       Boardpackage 2.3.0
### Description

Problem description
SPI Mode 2 and 3 are swapped compared to other boards like Uno, Due and 101.
A .ino file written for the Due, which uses SPI Mode 2 or 3 will not work on the ESP8266.
### Settings in IDE

Flash Size:         4MB
CPU Frequency:      80Mhz
Flash Mode:         ?
Flash Frequency:    ?
Upload Using:       USB Serial
Reset Method:       ?
### Sketch

``` cpp
#include <SPI.h>

void setup() {
  SPI.begin();
}

void loop() {
  SPI.beginTransaction(SPISettings(1000000, MSBFIRST, SPI_MODE0));
  SPI.transfer(0x053);
  SPI.endTransaction();

  delayMicroseconds(1);

  SPI.beginTransaction(SPISettings(1000000, MSBFIRST, SPI_MODE1));
  SPI.transfer(0x053);
  SPI.endTransaction();

  delayMicroseconds(1);

  SPI.beginTransaction(SPISettings(1000000, MSBFIRST, SPI_MODE2));
  SPI.transfer(0x053);
  SPI.endTransaction();

  delayMicroseconds(1);

  SPI.beginTransaction(SPISettings(1000000, MSBFIRST, SPI_MODE3));
  SPI.transfer(0x053);
  SPI.endTransaction();

  delayMicroseconds(16*8);  
}
```
### Debug Messages

This picture shows the current (wrong) behavior.
The byte 0x053 is written in all four modes: From left to right mode 0, 1, 2 and 3. The last two transmitts are wrong. See below of correct output.
See also here: http://forum.arduino.cc/index.php?topic=419660.0
esp8266

The common behavior is this for the Due:
due

Or this for the Uno:
uno

BTW: I have seen that this is marked as a todo in the code for this project: Arduino/libraries/SPI/SPI.cpp

Thanks for fixing this.

d-a-v:
use `SPI.restoreEsp8266LegacyMode2And3()` for former behaviour

fix #2416 
close #2418

This [snippet](https://github.com/esp8266/Arduino/blob/f0eb5509a02134aad2158ecc97167e209fa1c098/cores/esp8266/core_esp8266_version.h#L58) can be used to manage this breaking change and portability.


d-a-v:
> I suggest marking the method as deprecated right away.

Per https://github.com/esp8266/Arduino/issues/2416#issuecomment-479856618, I think it would make things not consistent. Marking it deprecated is a way to tell user to change source code. If we want that, I'd just remove this new call (see the number of votes in the comment just before mine).

Keeping it as it is will help esp8266-only trans-core-version projects to live until they are able to make it without older cores.

To summarize:
- I'm ok to remove this call now (I can already hear complaints from behind the 3 years old hill)
- To keep it and mark it deprecated just after next release

(vote -1 to remove, vote +1 to keep it :)

olikraus:
Thanks! :+1:

olikraus:
Hmmm... I might be missing something, but I do not find "SPI.restoreEsp8266LegacyMode2And3() "
When used, it gives a compiler error. I also do not find this function in the latest ESP8266 SPI sources.

olikraus:
Great, everything works as expected for 2.5.1, see https://github.com/olikraus/u8g2/issues/53

PR #5959, opened 2019-04-07
ArduinoOTA: Fix insecure nonce generation

imcatwhocode:
**Issue:** Currently, ArduinoOTA library generates authentication nonce using the value obtained from `micros()` function which is not secure I suppose (correct me if I'm wrong, please).

**Brief PR description**: This PR defines a new private `ArduinoOTAClass::generateNonce()` function which generates random MD5 hash using MCU HWRNG (compatible with ESP8266 / ESP32) or with the `micros()` fallback in case of non-ESP environment. Nonce generation in `ArduinoOTAClass::_onRx` replaced with the call of this function.

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -132,7 +132,7 @@ void ArduinoOTAClass::begin(bool useMDNS) {
   if(!_udp_ota->listen(IP_ADDR_ANY, _port))
     return;
   _udp_ota->onRx(std::bind(&ArduinoOTAClass::_onRx, this));
-  
+
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_MDNS)
   if(_useMDNS) {
     MDNS.begin(_hostname.c_str());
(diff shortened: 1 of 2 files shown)


libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -151,6 +151,40 @@ void ArduinoOTAClass::begin(bool useMDNS) {
...
+ * Generates nonce using PRNG / TRNG if available
+ * @return  Random MD5 string (nonce)
+ */
+String ArduinoOTAClass::generateNonce(void) {
+
+  MD5Builder nonce_md5;
+  nonce_md5.begin();
+
+  // Use ESP8266 PRNG via RANDOM_REG32 register
+  #if defined(ESP8266)
```
devyte: AFAIK, this code isn't shared with ESP32. I just checked, and their code base is currently very different.
Therefore, it doesn't make sense to have a specific case here for ESP32.
The same code change can be proposed to them in their repo.

In addition, I understand that this code is also specific to the ESP8266. Therefore, it doesn't make sense to consider a case where the ESP8266 macro is not defined.

I suggest simplifying the proposal to the specific case of ESP8266.

A side note: I also suggest implementing:
`inline uint32_t esp_random() {return RANDOM_REG32;}`
somewhere.

libraries/ArduinoOTA/ArduinoOTA.h
```diff
@@ -94,6 +94,7 @@ class ArduinoOTAClass
     void _onRx(void);
     int parseInt(void);
     String readStringUntil(char end);
+    String generateNonce(void);
```
devyte: There is no need for this to be a member method. Please move it to be a static function in the .cpp, or put it in an anonymous namespace.

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -151,6 +151,40 @@ void ArduinoOTAClass::begin(bool useMDNS) {
 #endif
 }
 
+/**
+ * Generates nonce using PRNG / TRNG if available
+ * @return  Random MD5 string (nonce)
+ */
+String ArduinoOTAClass::generateNonce(void) {
```
devyte: This method has no dependency on the ArudinoOTA class members or methods. Therefore, it doesn't make sense to make it a class method. That just pollutes the class interface.
Please make it a static function private to this file, or better yet, move it to an anonymous namespace at the beginning of this .cpp.

libraries/ArduinoOTA/ArduinoOTA.cpp
imcatwhocode: @devyte Got it, I'll keep ESP8266 option only. Thank you!

Just found out that ESP8266 Arduino Core has a `secureRandom()` function which is defined in `Arduino.h` and implemented in `WMath.cpp`, so there is a conflict between with the ESP32 `esp_random()`. Maybe it'll be more conventional to implement the `secureRandom()` function in the ESP32, instead of adding `esp_random()` to ESP8266?

earlephilhower:
Closing as multiple changes requested, merge conflicts, and no motion in ~6 months.

PR #5965, opened 2019-04-09
Update EspSoftwareSerial to release 5.0.0.

dok-net:
(Includes half-duplex patch for SerialStress.ino)


dok-net:
I've added defines for RX and TX.

PR #5968, opened 2019-04-09
Fix crash in certain PMEM printf format parsing

earlephilhower:
In some cases the printf implememtation would call an internal puts()
implementation which did not use pgm_read_byte() to access the format
string.  In many operating modes this would work, but in interrupts or
you'd get crashes.

Updated newlib to use pgm_read_byte in that one spot and recompiled.


TD-er:
Do you have any idea how long this bug was present?

d-a-v:
@TD-er since #5728

PR #5982, opened 2019-04-12
Convert ESP8266WebServer* into templatized model

### LINKED ISSUE #4908: ESP8266WebServerSecure.serveStatic() and BearSSL failed with "ERR_CONTENT_LENGTH_MISMATCH"
### Basic Infos

#### Platform

- Hardware:				ESP8266
- Core Version:			e486887

### Settings in IDE

- Module:  			Wemos D1 mini
- Flash Size: 			4MB
- lwip Variant:  		v1.4
- Reset Method:		ck
- Flash Frequency:		40Mhz
- CPU Frequency:		160Mhz
- Upload Using:		SERIAL
- Upload Speed:   		115200

### Problem Description
When using .serveStatic () I always get the error message in the browser "ERR_CONTENT_LENGTH_MISMATCH"
If I use the server.on () function, the file will be sent correctly.
```cpp
server.on("/", [](){
   File file = SPIFFS.open("/index.html", "r");
   server.streamFile(file, "text/html");
   file.close();
});
```
I tried to find the cause
**ESP8266WebServer.cpp**
```cpp
void ESP8266WebServer::_handleRequest() {
  ...
  handled = _currentHandler->handle(*this, _currentMethod, _currentUri);
  ...
}
```
**RequestHandlersImpl.h**
```cpp
bool handle(ESP8266WebServer& server, HTTPMethod requestMethod, String requestUri) override {
  ...
  if (_cache_header.length() != 0)
    server.sendHeader("Cache-Control", _cache_header);

  server.streamFile(f, contentType);  //<-- this function will be called
  return true;
}
```

it is called
**ESP8266WebServer.h**
```cpp
  template<typename T> 
  size_t streamFile(T &file, const String& contentType) {
    _streamFileCore(file.size(), file.name(), contentType);
    return _currentClient.write(file);
  }
```
instead of
**ESP8266WebServerSecure.h**
```cpp
  template<typename T>
  size_t streamFile(T &file, const String& contentType) {
    _streamFileCore(file.size(), file.name(), contentType);
    return _currentClientSecure.write(file);
  }
```

The pointer * this should be refer to BearSSL::ESP8266WebServerSecure server (443)?
I think that's it but I do not know how to fix it.

```cpp

#include <ESP8266WiFi.h>
#include <WiFiClient.h>
#include <ESP8266WebServerSecure.h>
#include <ESP8266mDNS.h>

#include <ESP8266httpUpdate.h>
#include <FS.h>

const char* ssid = "wifi";
const char* password = "1234";

BearSSL::ESP8266WebServerSecure server(443);
static const char serverCert[] PROGMEM = R"EOF(...)EOF";
static const char serverKey[] PROGMEM =  R"EOF(...)EOF";

void setup() { 
  SPIFFS.begin();
  WiFi.begin(ssid, password);

  server.serveStatic("/", SPIFFS, "/index.html");

  server.begin();
}

void loop() {
  server.handleClient();
}

```

### Debug Messages

```
New secure client
method: GET url: / search:
headerName: Host
headerValue: littledroid1
headerName: Connection
headerValue: keep-alive
headerName: Cache-Control
headerValue: max-age=0
headerValue: 1
headerName: User-Agent
headerValue: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.99 Safari/537.36
headerName: Accept
headerValue: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
headerName: Accept-Encoding
headerValue: gzip, deflate, br
headerName: Accept-Language
headerValue: de-DE,de;q=0.9,en-US;q=0.8,en;q=0.7
args:
Request: /
 Arguments:

StaticRequestHandler::handle: request=/ _uri=/
StaticRequestHandler::handle: path=/index.html, isFile=1
```

earlephilhower:
Supercedes #4912

Refactor the three versions of ESP8266WebServer and *WebServerSecure to a
single templated class. Use "using" to enable old, non-templated names to b
used (so no user changes required to compile or run).

Fixes #4908 and clean up the code base a lot.

Basic tests run (the ones in the example code).

No code changes are required in userland except for setting the SSL
certificates which now use a cleaner "getServer()" accessor and lets the
app use the native BearSSL calls on the WiFiClientSecure object.

@devyte should be proud, it removes virtuals...


devyte:
Overall, nice work! I think I cried a little with so many virtuals removed. I see that request handler still has some, though.

I do have some comments about code structure, namespaces and refactoring, but it will take me a bit to go through the whole PR for the specifics.

In the meantime, some comments and questions:
1. Is there any bin size impact? IRAM impact?
2. This change means that the secure webserver is no longer a descendant of the non-secure webserver. As a result, any sketch out there that was implemented using the inheritance may no longer build. The way a user would need to address the change is to migrate any functions that were relying on the inheritance to a template version.
=> Breaking change.
3. As I understand things, the server and client must always match, i.e.: both must either be secure, or both not secure. Is there any scenario where a mix makes sense?
I'm thinking:
WiFiServer returns only WiFiClient
WiFiServerSecure returns only WiFiClientSecure
=> maybe WiFiServer and WiFiServerSecure should have a type member that reflect the WiFiClient type they return. Then, the template parameter here would be unique: WiFiServerType. The client type would then be WiFiServerType::WiFiClientType.
Why do this? it disallows the possibility of declaring a webserver with mismatched server/client types.
4. The templated object(s) should be put in a namespace. The specific types defined for secure/non-secure, i.e.: the using statements, should be outside the namespace in global space and reference the template with the specific type definitions. See polledTimeout for one way to handle this.
5. At a glance I see several methods that aren't really specific to the class they are implemented in, e.g.: urlDecode. It is good practice to move those out of the class into their own .cpp as standalone functions. This is particularly important when dealing with a template class, because each combination of template args will generate a copy of the method, and all copies will essentially be the same code.
6. One thing I have pending is to clean up the using namespace mime, which imports mime into the global space. If the template objects are moved into their own namespace, then mime could be imported there, and global namespace pollution would no longer be an issue.

earlephilhower:
0. RequestHandler needs to be virtual without some major work since we need to keep track of multiple different ones with different implementations per-web-node.  I don't see that changing anytime soon.

1. Bin size is about the same on things I tried.  AdvancedWebServer actually shrinks by 704 bytes vs. master, and it uses 40 bytes less heap RAM as well.  But, in the grand scheme, that's noise.

2. Are you saying an app subclassing WebServerSecure (`class MyServer : public WebServerSecure {}`) won't work?  I think it's a no-op as you can subclass the template-ized just like normal, can't you?  There's the possibility of object slicing, I guess, but the WebServer never returns itself, and only uses WiFiClient/WiFiServer which we don't expect to be subclassed.  At that point, they'd need to make a template-ized instantation to get it going anyway, I think.

3. It's impossible for a WiFiServer to return a WiFiClientSecure, and vice versa.    Let me see if I can use some `typename` magic.

4.  Will do.

5. Agreed, but I was trying to minimized the refactoring involved.  Wouldn't that be a breaking change?  For a template-ized class, I thought all methods were template-type decorated so a naked "WebServer::parseURL()" declaration wouldn't be included as a class member.  Moving the urlDecode to another class/global function would definitely be a breaking change (it's a public member).  I would also suggest that apps will only be using secure or non-secure versions of the object, and not both, so duplicated code bloat is not going to be a common thing.

6. I think we moved mimetable to its own file just because of the PROGMEM segment problems.  I think it can be moved back to the class...

earlephilhower:
Got 3, 4, and 6 all taken care of.

libraries/ESP8266WebServer/src/Parsing-impl.h
```diff
@@ -179,7 +181,7 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
     String plainBuf;
     if (   !isForm
         && // read content into plainBuf
-           (   !readBytesWithTimeout(client, contentLength, plainBuf, HTTP_MAX_POST_WAIT)
+           (   !readBytesWithTimeout<ServerType>(client, contentLength, plainBuf, HTTP_MAX_POST_WAIT)
```
devyte: This is a template function, there should be no need to explicitly specify the template arg due to automatic deduction (simplifies reading):
```cpp
           (   !readBytesWithTimeout(client, contentLength, plainBuf, HTTP_MAX_POST_WAIT)
```

earlephilhower: You'd think there'd be no reason, but the template arg really is needed there to make GCC happy:
````
In file included from /tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer.h:213:0,
                 from /tmp/ci/libraries/DNSServer/examples/CaptivePortalAdvanced/CaptivePortalAdvanced.ino:3:
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing-impl.h: In instantiation of 'bool esp8266webserver::ESP8266WebServerTemplate<ServerType>::_parseRequest(esp8266webserver::ESP8266WebServerTemplate<ServerType>::ClientType&) [with ServerType = WiFiServer; esp8266webserver::ESP8266WebServerTemplate<ServerType>::ClientType = WiFiClient]':
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h:330:41:   required from 'void esp8266webserver::ESP8266WebServerTemplate<ServerType>::handleClient() [with ServerType = WiFiServer]'
/tmp/ci/libraries/DNSServer/examples/CaptivePortalAdvanced/CaptivePortalAdvanced.ino:141:23:   required from here
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing-impl.h:184:89: error: no matching function for call to 'readBytesWithTimeout(esp8266webserver::ESP8266WebServerTemplate<WiFiServer>::ClientType&, uint32_t&, String&, int)'
            (   !readBytesWithTimeout(client, contentLength, plainBuf, HTTP_MAX_POST_WAIT)
                                                                                         ^
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing-impl.h:184:89: note: candidate is:
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing-impl.h:43:13: note: template<class ServerType> bool esp8266webserver::readBytesWithTimeout(typename ServerType::ClientType&, size_t, String&, int)
 static bool readBytesWithTimeout(typename ServerType::ClientType& client, size_t maxLength, String& data, int timeout_ms)
             ^
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing-impl.h:43:13: note:   template argument deduction/substitution failed:
/tmp/ci/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/Parsing-impl.h:184:89: note:   couldn't deduce template parameter 'ServerType'
            (   !readBytesWithTimeout(client, contentLength, plainBuf, HTTP_MAX_POST_WAIT)
                                                                                         ^
````

File: libraries/ESP8266WebServer/src/detail/RequestHandler.h [hunk 2]
```diff
@@ -1,19 +1,20 @@
...
 
+template<typename ServerType>
 class RequestHandler {
 public:
     virtual ~RequestHandler() { }
     virtual bool canHandle(HTTPMethod method, String uri) { (void) method; (void) uri; return false; }
     virtual bool canUpload(String uri) { (void) uri; return false; }
-    virtual bool handle(ESP8266WebServer& server, HTTPMethod requestMethod, String requestUri) { (void) server; (void) requestMethod; (void) requestUri; return false; }
-    virtual void upload(ESP8266WebServer& server, String requestUri, HTTPUpload& upload) { (void) server; (void) requestUri; (void) upload; }
+    virtual bool handle(ESP8266WebServerTemplate<ServerType>& server, HTTPMethod requestMethod, String requestUri) { (void) server; (void) requestMethod; (void) requestUri; return false; }
```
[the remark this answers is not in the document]
earlephilhower: Done.

libraries/ESP8266WebServer/src/Parsing-impl.h
devyte: You're right, it's a dependent type, not worth it

PR #5989, opened 2019-04-15
Make SPIFFS be an integer number of blocks

### LINKED ISSUE #3621: SPIFFS size is not an integer multiple of block size.
For some of the boards defined in boards.txt, the allocation for SPIFFS is not an integer multiple of the SPIFFS block size. 
For 4MB boards, for example, there is 1MB for PROGMEM, 3052 kB for SPIFFS and 20kB for EEPROM. The block size is 8 kB, but 3052 kB is not divisible by 8 kB blocks.
I propose to modify eagle.flash.4m.ld linker file and boards.txt to use an SPIFFS block size of 4 kB, which works fine.

earlephilhower:
Fixes #3621

boards.txt.py simply calculated SPIFFS end and start using flash sizes, but
in cases where an 8K page was used (>512KB SPIFFS), this could leave a
4K half-block left at the end of SPIFFS.

mkspiffs and the SPIFFS code uses integer division to calculate the
maximum block number, so it worked fine in practice and the code simply
ignored the extra, fractional block.

Now actually take block size into account when calculating the end of
SPIFFS, ensuring no fractional blocks are passed in.  Does not result in
data loss on pre-existing SPIFFS filesystems.


tools/boards.txt.py
```diff
@@ -1172,11 +1172,14 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
     else:
         max_upload_size = 1024 * 1024 - reserved
         spiffs_start = (flashsize_kb - spiffs_kb) * 1024
-        if spiffs_kb < 512:
+        if spiffs_kb <= 512:
```
d-a-v: Is this 512k change dangerous ? (per https://github.com/esp8266/Arduino/pull/5412#issuecomment-443743683 and following)

earlephilhower: The existing boards.txt.py has a discrepancy between blocksize in the menu and blocksize in LD for the 512k case.  When I remove the 1-line spiffs integrab block fix, the changed code generates exactly the same LDs but a 1-line different 512k menu entry, where the menu blocksize and LD blocksize match up.

earlephilhower: Err, scratch that.  When I go back to main and only apply the block -> spiffs_blocksize change I see there is a mismatch in the 1M512 section (where the menu and the LD file disagree about blocksize, since `spiffs_blocksize` is used for the menu, while `block` is used for the LD file.

Question is, which should be correct?  I think the menu items are used for the SPIFFS uploader, while the LD entry is used by code running on the core.  So, as I see it, the 1M512 option SPIFFS upload will be corrupted...the host will build the SPIFFS image w/blocksize of 4k, but the core will try and access it with a 8K blocksize....

````
diff --git a/tools/boards.txt.py b/tools/boards.txt.py
index fd960511..f5e7e095 100755
--- a/tools/boards.txt.py
+++ b/tools/boards.txt.py
@@ -1217,13 +1217,13 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
         if spiffs_kb == 0:
             spiffs_start = spiffs_end
             page = 0
-            block = 0
+            spiffs_blocksize = 0
         elif spiffs_kb < 0x80000 / 1024:
             page = 0x100
-            block = 0x1000
+ #           block = 0x1000
         else:
             page = 0x100
-            block = 0x2000
+#            block = 0x2000
 
         print("/* Flash Split for %s chips */" % strsize)
         print("/* sketch @0x%X (~%dKB) (%dB) */" % (spi, (max_upload_size / 1024), max_upload_size))
@@ -1246,7 +1246,7 @@ def flash_map (flashsize_kb, spiffs_kb = 0):
         print("PROVIDE ( _SPIFFS_start = 0x%08X );" % (0x40200000 + spiffs_start))
         print("PROVIDE ( _SPIFFS_end = 0x%08X );" % (0x40200000 + spiffs_end))
         print("PROVIDE ( _SPIFFS_page = 0x%X );" % page)
-        print("PROVIDE ( _SPIFFS_block = 0x%X );" % block)
+        print("PROVIDE ( _SPIFFS_block = 0x%X );" % spiffs_blocksize)
         print("")
         print('INCLUDE "local.eagle.app.v6.common.ld"')
 
diff --git a/tools/esptool b/tools/esptool
--- a/tools/esptool
+++ b/tools/esptool
@@ -1 +1 @@
-Subproject commit 9ad444a6e06e58833d5e6044c1d5f3eb3dd56023
+Subproject commit 9ad444a6e06e58833d5e6044c1d5f3eb3dd56023-dirty
diff --git a/tools/pyserial b/tools/pyserial
--- a/tools/pyserial
+++ b/tools/pyserial
@@ -1 +1 @@
-Subproject commit c54c81d933b847458d465cd77e96cd702ff2e7be
+Subproject commit c54c81d933b847458d465cd77e96cd702ff2e7be-dirty
diff --git a/tools/sdk/ld/eagle.flash.1m512.ld b/tools/sdk/ld/eagle.flash.1m512.ld
index f2df1a07..10e573c7 100644
--- a/tools/sdk/ld/eagle.flash.1m512.ld
+++ b/tools/sdk/ld/eagle.flash.1m512.ld
@@ -17,6 +17,6 @@ MEMORY
 PROVIDE ( _SPIFFS_start = 0x4027B000 );
 PROVIDE ( _SPIFFS_end = 0x402FB000 );
 PROVIDE ( _SPIFFS_page = 0x100 );
-PROVIDE ( _SPIFFS_block = 0x2000 );
+PROVIDE ( _SPIFFS_block = 0x1000 );
 
 INCLUDE "local.eagle.app.v6.common.ld"
````

earlephilhower: Thanks for the notice.  This change was actually a no-op (it's in an if branch that would never be taken), but there was an issue where spiffs_block was changed in the LD which is a no-no (will make existing FS unreadable).  Latest push has no differences in anything but spiffs_end.

TD-er:
Could this also be the reason why I was not able to make a working 16M/14M SPIFFS layout with 32k block size?

earlephilhower:
@TD-er that's pretty unlikely.  Everything just integer divides the size which always rounds down, so the extra blocks at the end of space would never be touched or used.  This specific change should be a no-op functionally, and only affect the "code aesthetics" by keeping all input values consistent.

d-a-v:
@earlephilhower in order to fix conflicts, can you regenerate `boards.txt.py` files ?

earlephilhower:
#6537 will un-breaking this...

PR #5992, opened 2019-04-16
Upgrade to https: serving for JSON, links in docs

### LINKED ISSUE #5480: Serve package index over HTTPS
Why is the board manager [http://arduino.esp8266.com/stable/package_esp8266com_index.json](url) not in https like ESP32 [https://dl.espressif.com/dl/package_esp32_index.json](url) ?
I cannot load ESP8266 in our work office, thus I need to hotspot my mobile to connect and download everytime the json file.

earlephilhower:
Fixes #5480


doc/installing.rst
```diff
@@ -29,7 +29,7 @@ Instructions
    after installation.
 
 You may optionally use *staging* boards manager package link:
-``http://arduino.esp8266.com/staging/package_esp8266com_index.json``.
+``https://arduino.esp8266.com/staging/package_esp8266com_index.json``.
```
igrr: Please note that this link doesn't work anymore (no staging packages)

earlephilhower: Ah, will remove and push.  Wasn't checking that, just the https-ishness of the URLs.

earlephilhower:
@devyte, that's orthogonal to the PR.  The code in the repo doesn't reference those older revs.  The concatenated JSON file kept on the fileserver is the only spot those refs exists.  Dropping bits from it is a separate (manual?) process.

devyte:
I was looking at this PR as the final step in the migration process, i. e. Merge this and the issue auto closes.
If you view it as orthogonal that's fine, but in that case a new issue is needed to track either reducing the json manually or migrating the old stuff to the new server.

earlephilhower:
I'm fine doing that, but I think the SSH keys are hidden in the CI config, so it's not a simple matter of downloading, removing pre 2.3.x, and re-uploading.  It all needs to be scripted as a release or something to make it happen.  Seems like a PITA...am I missing something to make it more straightforward?

Maybe it's a matter of editing the merge release script to make it delete pre XXX versions.  I suppose that's possible, if a little ugly...

d-a-v:
> Maybe it's a matter of editing the merge release script to make it delete pre XXX versions. I suppose that's possible, if a little ugly...

Maybe in a future PR to keep this less ugly we can add a variable containing all releases that are meant to be kept, and a script-function that would filter out all other releases.

earlephilhower:
@devyte, for sanity's sake I've added a drop-versions script and set it to drop the obsolete and unavailable 1.6.5 versions as well as our -betas which have no reason to exist anymore.

PR #5993, opened 2019-04-17
Submodule SoftwareSerial: small footprint reduction

dok-net:
The ICACHE RAM use is optimized and the ISR table size is reduced in size.


dok-net:
@earlephilhower Addressing resource issues due to ISR trampoline tables, any new release of EspSoftwareSerial relies on #6049 getting fixed by #6048 . This allows me to remove the tables completely :-) Releasing EspSoftwareSerial including an interface change for better alignment with ESP32 (think WEMOS D1 mini ESP32 boards that are pin-compatible) is pending this PR :-(

earlephilhower:
@dok-net sure, keep on us if we don't merge in a timely basis.  Right now I think we're still trying to ensure 2.5.1 is working, so merge freeze until that's confirmed.

PR #5995, opened 2019-04-17
ISR: check for address in IRAM

### LINKED ISSUE #5779: interruptFunctional should be placed in IRAM

d-a-v:
closes #5780
closes #5779

### DIFF
--- a/cores/esp8266/core_esp8266_wiring_digital.cpp
+++ b/cores/esp8266/core_esp8266_wiring_digital.cpp
@@ -171,6 +171,16 @@ void ICACHE_RAM_ATTR interrupt_handler(void *arg) {
 extern void cleanupFunctional(void* arg);
 
 extern void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void *arg, int mode) {
+
+  // #5780
+  // https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map
+  if ((uint32_t)userFunc >= 0x40200000)
+  {
+    // ISR not in IRAM
+    ::printf((PGM_P)F("ISR not in IRAM!\r\n"));
+    abort();
+  }
+
   if(pin < 16) {
     ETS_GPIO_INTR_DISABLE();
     interrupt_handler_t *handler = &interrupt_handlers[pin];


dok-net:
Refer to the ESP32 sources' FunctionInterrupt.cpp:
void IRAM_ATTR interruptFunctional(void* arg)

But ESP8266 FunctionalInterrupt.cpp is missing this patch:

--- a/cores/esp8266/FunctionalInterrupt.cpp
+++ b/cores/esp8266/FunctionalInterrupt.cpp
@@ -11,7 +11,7 @@ typedef void (*voidFuncPtrArg)(void*);
 extern "C" void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void*fp , int mode);
 
 
-void interruptFunctional(void* arg)
+void ICACHE_RAM_ATTR interruptFunctional(void* arg)
 {
     ArgStructure* localArg = (ArgStructure*)arg;
 	if (localArg->functionInfo->reqScheduledFunction)


Which leaves the commited IRAM check firing all the time:
ISR not in IRAM!

I'm converting EspSoftwareSerial to use FunctionInterrupt in order to save the ISR table, please patch FunctionalInterrupt.cpp as stated above, if nothing else stands against it and it was just missed.

Issue #5749, opened 2019-02-09
2.5.0: Issues with Arduino API


d-a-v:
> UDP `virtual IPAddress remoteIP() =0;` `virtual IPAddress remoteIP() const =0;`
you can't change a virtual function in an established Arduino core class

https://github.com/esp8266/Arduino/commit/5c4db3acf4353870b6a5663ba06d712d67e6c179#commitcomment-32254962

> other networking libraries Client implementations do not implement `flush` and `stop` with parameter.
"The type 'UIPClient' must implement the inherited pure virtual method 'Client::stop' "

https://github.com/esp8266/Arduino/commit/83a8076db87b77de6a7b500c18b85057e38ed117#commitcomment-32254941

@jandrassy Can you please give more details ?

Also [If your sketch does not compile and returns error](https://learn.adafruit.com/adafruit-feather-huzzah-esp8266/using-arduino-ide) needs to be addressed.

jandrassy:
There are other networking libraries used with esp8266, not only the native WiFi or bundled Ethernet library (why is it there?). If you change the virtual functions of the Arduino networking base classes Client, Server, UDP the libraries don't compile. Function `flush` with parameter is a different virtual function then `flush` without parameter. The default value of the parameter doesn't change it. So now  for example the UIPEthernet library or the standard Ethernet library and perhaps some GPRS libraries don't compile for esp8266 because they don't implement flush, stop, remoteIP with a parameter.

d-a-v:
It will be addressed for next release 2.5.1

PaulStoffregen:
FWIW, there was an unresolved problem with DNS lookups by the Ethernet lib.  Only ESP8266 has this problem.  The exact same code ran perfectly on all other boards, including ESP32.  I never did a deeper investigation into why that failed on ESP8266.  Maybe someday....

If anyone cares about Wiznet-based Ethernet on ESP8266, please test some of the library's examples which use a text-based hostname rather than only IP numbers.

jandrassy:
recommended
https://github.com/arduino/ArduinoCore-API

jandrassy:
> FWIW, there was an unresolved problem with DNS lookups by the Ethernet lib. Only ESP8266 has this problem. The exact same code ran perfectly on all other boards, including ESP32. I never did a deeper investigation into why that failed on ESP8266. Maybe someday....
> 
[...]

PR #5997, opened 2019-04-17
replace old Ethernet library by a submodule ref'ing the updated one

d-a-v:
https://github.com/esp8266/Arduino/issues/5749#issuecomment-482971148


d-a-v:
We don't need it locally because
- we don't use it in examples
- we can't maintain it
- submoduling it will need regular updates following upstream (library manager is here for that)
- we can ship it in releases like we were doing with softwareserial before it was submoduled

To be noted: softwareserial is submoduled because we use it in examples. It is an esp8266 specific library so the author is also supposed to care about this (unlike Ethernet which is more arduino generic).

Comments are welcome.

d-a-v:
> Comments are welcome.

This library is not available from library manager (is it?)

Anyway comments came through the arduino maintainers chat board: We'd better have it locally instead of
* letting users unsuccessfully look for where it was before
* embed it in releases while not being available in git (like SoftwareSerial was)
* point to a known compatible version thanks to submodule git refs

PR #6000, opened 2019-04-19
Fix MFLN probe and allow returning whether MFLN succeeded or not after a connection.

### LINKED ISSUE #5996: probeMaxFragmentLength of BearSSL::WiFiClientSecure is not compatible with OpenSSL supporting Maximum Fragment Length Negotiation extension
### Basic Infos

#### Platform

- Hardware:			        ESP-12
- Core Version:			2c36cfe04df1a712b2afb00820ff64d54bc63588
- Operating System:		MacOS

### Problem Description

The method probeMaxFragmentLength of BearSSL::WiFiClientSecure is not compatible with OpenSSL supporting Maximum Fragment Length Negotiation extension. I've tested on OpenSSL 1.1.1b1 (beta 1) and also the latest OpenSSL 1.1.1 build. The method probeMaxFragmentLength generates very simple ClientHello message which is rejected by the server with the following error:

Client connection from X.X.X.X failed: error:1417A0C1:SSL routines:tls_post_process_client_hello:no shared cipher.

The server drops connection immediately during processing of ClientHello message. OpenSSL 1.1.1 supports usage of  Maximum Fragment Length Negotiation extension properly. If probing is not done but setBufferSizes is used before connection, server accepts extension sent in ClientHello and confirms selected max fragment length in ServerHello.

Use example in libraries/ESP8266WiFi/examples/BearSSL_MaxFragmentLength/BearSSL_MaxFragmentLength.ino against a server backed by OpenSSL 1.1.1 (Mosquitto, Haproxy or Apache). The function fetchMaxFragmentLength in this example will always report that MFLN is not supported even it is working well. If you skip probing, setBufferSizes(512, 512) and examine ClientHello and ServerHello packets, you will find that server accepts MFLN and fragments are not bigger than 512.

earlephilhower:
probeMFLN was failing on some connection attempts to servers which only
supported EC based ciphers because it did not include the proper TLS
handshake extensions to list what kinds of ECs it supported.

Add those to the probeMFLN ClientHello message to make probes pass.

Add client.getMFLNStatus method, returns MFLN state

After a connection it is useful to check whether MFLN negotiation
succeeded.  getMFLNStatus returns a bool (valid only after
client.connect() succeeds, of course) indicating whether the requested
buffer sizes were negotiated successfully.

Fixes #5996


earlephilhower:
Don't merge, I just realized that I need a smarter response parser now that there are multiple extensions being sent over the wire.  Not rocket science, but will take a day or two to sit down and write.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -103,6 +103,11 @@ class WiFiClientSecure : public WiFiClient {
     // Sets the requested buffer size for transmit and receive
     void setBufferSizes(int recv, int xmit);
 
+    // Returns whether MFLN negotiation for the above buffer sizes succeeded (after connection)
+    int getMFLNStatus() {
+      return connected() ? br_ssl_engine_get_mfln_negotiated(_eng) : false;
+    }
```
d-a-v: why not 
```
    bool getMFLNStatus() {
      return connected() && br_ssl_engine_get_mfln_negotiated(_eng);
    }
```

devyte: What @d-a-v said :stuck_out_tongue_closed_eyes:

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -103,6 +103,11 @@ class WiFiClientSecure : public WiFiClient {
     // Sets the requested buffer size for transmit and receive
     void setBufferSizes(int recv, int xmit);
 
+    // Returns whether MFLN negotiation for the above buffer sizes succeeded (after connection)
```
devyte: > after connection

I'm a bit concerned about the error proneness of this. If the mfln was negotiated, and the result is bigger than the set buffer size, is there some assert or something that can be added somewhere? Attempting to use a connection in such a state is clearly a programming error.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
earlephilhower: Maybe I get paid by the trinary?  I'll clean it up.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
earlephilhower: It *may be* an error, but is not guaranteed one.  This went back to the early BearSSL days.  You can set buffers any size you like, and BearSSL will try and negotiate to get the server to fit within your receive (actually, now its min(recv buffer, xmit buffer) and I'm talking w/ @pornin since this seems to be over cautious, but that's another PR)).  BearSSL will never overflow buffers, so it's not a write out-of-bounds problem.    If the server only sends fragments which fit w/in the buffer then even if MFLN does not happen things "just work."

If the server sends a fragment that's larger than the receive buffer, it will be dumped by BSSL since you can't decode partial fragments, and then there will be trouble.

So, even on older TLS 1.2 connections, if you're guaranteed the server will send, say, 1K max, then you can safely set a receive buffer of 2K and not have a problem.

BSSL needs the buffers before connections begin, so it's not possible to resize things after-the-fact.

devyte: My worry is not about overflows et al, but rather about hitting the bad case when establishing a connection, then having fragments silently getting dumped when tryint to use it, then getting a bunch of issues opened about bssl not working.
So what I'm thinking is whether there is some way to completely disallow the bad case: assert, panic, abort, anything to force the user to notice that the set buffer size is smaller than the negotiated size, and hence whether the connection works or not depends on the mood of the nearest plant.

earlephilhower: The problem with a heavy hammer solution like that is that many people cannot update their server to one that supports MFLN, but they can guarantee the maximum message size (i.e. they're in control of the content returned..either MQTT or CGI response).

It's also only an issue when the `setBufferSizes()` call is used.  By default, BearSSL's completely safe and will connect and transfer from anything, at any size.

devyte: Alright, if you think the case is valid for a user, I'll defer to you.

PR #6002, opened 2019-04-20
ISR code must be in ICACHE_RAM - the check was commited in ab125162bf

dok-net:
#5995 hasn't actually merged any PR - the minimum to get actual code working is in this PR.
Caveat: I don't know how to check if ICACHE_RAM_ATTR has the desired effect on lambda expressions, if it does not, there will still be an adverse performance (no real-time behavior!) impact over the standard Arduino attachInterrupt() usage.
Q: why not implement void attachInterruptArg(uint8_t pin, void (*)(void*), void * arg, int mode) like the ESP32 has already?

### DIFF
--- a/cores/esp8266/FunctionalInterrupt.cpp
+++ b/cores/esp8266/FunctionalInterrupt.cpp
@@ -11,7 +11,7 @@ typedef void (*voidFuncPtrArg)(void*);
 extern "C" void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void*fp , int mode);
 
 
-void interruptFunctional(void* arg)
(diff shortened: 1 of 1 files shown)


d-a-v:
> why not implement void attachInterruptArg(uint8_t pin, void ()(void), void * arg, int mode) like the ESP32 has already?

It is the same
```
typedef void (*voidFuncPtr)(void);
extern void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void *arg, int mode) {
```

d-a-v:
> I don't know how to check if ICACHE_RAM_ATTR has the desired effect on lambda expressions

I used that code: https://github.com/esp8266/Arduino/pull/5780#issuecomment-484070364

PR #6006, opened 2019-04-22
New menu option to minimize BSSL ROM with only RSA

### LINKED ISSUE #6005: Enhancement - Allow linking in smaller set of ciphers on BearSSL to save code space
This would be, unfortunately, yet another Arduino menu option where you would be able to define specific cipher groups (say, only RSA or only EC or only ChaCha20) would be available for SSL negotiation.

Make sure there are no references to the unused codes in question, and the linker will leave out large swaths of BearSSL code.

Useful for flash constrained people and folks trying to do OTA in 1MB devices.

earlephilhower:
Adds a menu option and define to limit BearSSL to older RSA connection
options.  This saves ~45K program memory and can speed up connections
since EC, while more secure, is significantly slower on the chip.
The supported ciphers are identical to the ones that axTLS supported.

Fixes #6005


earlephilhower:
@d-a-v, do you have a better solution?  Think this is something that should only be settable from platform_local.txt?

d-a-v:
Not a better solution. Indeed another (worse) solution is to use platform.local.txt (like this example https://github.com/esp8266/Arduino/pull/6003#issuecomment-486401270). It is worse because it is not easily selectable, while not confusing users with tons of options.

So no, to be honest, I prefer the tool menu option (even for #6003).

I wish this menu were more configurable so we can put less obvious choices below a single "uncommon" option.

Issue #4392, opened 2018-02-19
Support for user provided MD5 credentials for ESP8266WebServer authentication


IMAN4K:
It would be great to have an option to provide `ESP8266WebServer` authentication's credentials as `MD5` values instead of clear texts.
This is useful in case, where `<user:pass>` is stored in `MD5` format and is modifiable.
Adding an option switch like this, defaulted to `false`, maybe is sufficient:
`bool authenticate(const char * username, const char * password, bool user_md5 = false);`

earlephilhower:
What threat are you trying to protect against?  Sending MD5 over an unsecured channel is still completely insecure as anyone along the path can just replay the MD5 back to authenticate.  It's just trading one plaintext password for another.  And with rainbow tables, reversing MD5 to get back to plaintext is not a problem for anyone with a computer built this decade.

For security, basic HTTP user:pass over an encrypted channel is a common recommendation.  There is now a WebServerSecure which can pretty much replace the unencrypted one w/a 1-line addition to supply its certificate and key.

IMAN4K:
Agree.
But `MD5` is better than nothing and is common in modem-router authentication mechanisms & not very easy to break!
Also `ESP8266WebServer` simply hash `user:pass` for authentication.
In my case, i have only user:pass as stored `MD5` value, so i should simply be able to provide them to `ESP8266WebServer`

Don't know if `WebServerSecure` is in buggy state or NOT!

earlephilhower:
WebServerSecure's in a pretty good state as far as we know.  I and @d-a-v have done lots of stress testing on the underlying WiFiServerSecure/ClientSecure.  Give it a try, but be aware you need 20KB+ free heap to run a SSL connection.

I'd argue MD5 is perfectly easy to decrypt today.  See https://hashkiller.co.uk/md5-decrypter.aspx for an online version....

PR #6020, opened 2019-04-28
Modified ESP8266WebServer to allow for authentication with H(A1).

overtone1000:
[RFC 2617 4.13](https://tools.ietf.org/html/rfc2617#section-4.13) discusses storing credentials as the username and H(A1) rather than the username and password, but this isn't possible with the current implementation of Digest authentication in the ESP8266WebServer class because the authenticate function takes username and password as arguments. This limitation is addressed in this branch, which contains four modifications:

1. Exposes a function in ESP8266WebServer that performs HTTP Digest authentication with username and H(A1)  as the arguments.
2. Preserves the HTTP authentication function in ESP8266WebServer with username and password as arguments but modifies it to call the new function where appropriate.
3. Adds a public static function to ESP8266WebServer for generating H(A1) from the username, realm, and password.
4. Adds an example of how to use this new feature, which is called HttpHashCredAuth.ino and is in the appropriate folder. This example uses the ESP8266WebServerSecure class to present a web interface where the user can change the HTTP credentials. These are saved in SPIFFS as the username and H(A1). The plain text password is never stored in the program or the file system.


overtone1000:
This branch was tested with the new example as well as with the HttpAdvancedAuth.ino and HttpBasicAuth.ino examples to make sure no existing functionality was disrupted. All three examples compiled and functioned as expected.

overtone1000:
I have only found (#4392) that asked a related question about digest authentication, but this was closed as the question was about the security of the transmission, not about its storage. This issue was quickly closed.

I have found at least two pull requests (#3328 and  #4404) that address a similar issue with basic authentication.

tools/python/python
```diff
@@ -0,0 +1 @@
+/usr/bin/python
```
earlephilhower: Your GIT somehow added this file which is installed from the get.py script.  This needs to be removed.

earlephilhower: Same comment about `placeholder_for_Arduino` addition

libraries/ESP8266WebServer/examples/HttpHashCredAuth/HttpHashCredAuth.ino
```diff
@@ -0,0 +1,258 @@
...
+//Unfortunately it is not possible to have persistent WiFi credentials stored as anything but plain text. Obfuscation would be the only feasible barrier.
+#ifndef STASSID
+#define STASSID "your-ssid"
+#define STAPSK  "your-password"
+#endif
+
+const char* ssid = STASSID;
+const char* wifi_pw = STAPSK;
+
+const String file_credentials=R"(/credentials.txt)"; //SPIFFS file name for the saved credentials
```
earlephilhower: The CI system demands spacing here for Arduino astyle: `... file_credentials = R"(...`

overtone1000:
> especially since raw user/real/passwords are still stored in 8266 flash...ugh

My goal for the example (HttpHashCredAuth.ino) was to show how to avoid this. The credentials are initialized to something generic, but once they are changed through the web interface, only the user name and hash of user/real/password should be in flash. What did I miss?

libraries/ESP8266WebServer/examples/HttpHashCredAuth/HttpHashCredAuth.ino
overtone1000: Thanks! I believe the new commit to this branch addresses all style issues.

tools/python/python
overtone1000: These have been removed from the latest commit.

earlephilhower:
Ah, did not catch you were only writing the hash in SPIFFS.  Looks good to me, we'll see what CI thinks of it.  Thanks!
-EFP3

earlephilhower:
Thanks for the contribution, @overtone1000!  2.5.2 just came out so we can start adding feature updates like this.

overtone1000:
I'm very appreciative of the work you all do on this project. This is my first ever PR, so thank you for helping me!

twinclouds:
Got the following error: 
'class BearSSL::ESP8266WebServerSecure' has no member named 'authenticateDigest'.
How to fix?

overtone1000:
@twinclouds, which version of the core are you using? This has been merged into the master branch but isn't yet part of the current release (2.5.2). If you need this feature now you'll need to use the current git version as described in the readme file in the root of this repositoty.

PR #6027, opened 2019-04-30
use memcpy instead of strcpy in WString.cpp

d-a-v:
I can happen String carries binary data, this simple update allows it.
Also, `memcpy` is more efficient than `strcpy` when size is known.
It's an harmless temporary fix until Webserver's painful `arg("plain")` can be worked out.
@LaborEtArs

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -206,7 +206,7 @@ String & String::copy(const char *cstr, unsigned int length) {
         return *this;
     }
     setLen(length);
-    strcpy(wbuffer(), cstr);
+    memcpy(wbuffer(), cstr, length + 1);
     return *this;
 }
 
@@ -216,7 +216,7 @@ String & String::copy(const __FlashStringHelper *pstr, unsigned int length) {
         return *this;
     }
     setLen(length);
-    strcpy_P(wbuffer(), (PGM_P)pstr);
+    memcpy_P(wbuffer(), (PGM_P)pstr, length + 1);
     return *this;
 }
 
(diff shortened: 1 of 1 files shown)


earlephilhower:
`String`s must not contain 0s, that was one of the assumptions of the SSO optimization (and I believe other bits of the original String class).

SSO uses `strlen` to get the length of a string in SSO (allows 1 more byte in the String object for text), so this won't work as-is.

It would be possible to rewrite SSO using the byte (sizeof(String)-1) as length, and adjusting the maximum SSO string size by 1.

earlephilhower:
Also, random thought:  `memcpy` may be unsafe, would suggest using `memmove` unless we can verify there is no potential for overlapping src/dest.  Had other problems in `String` because of stuff like this.

KlaasjanN:
I may have encountered some issues where the string is empty. In the past this did not seem to cause any issues, but it seems that the switch to memmove is causing problems. Haven't had time to check the underlying code and to do full analysis, but can you please include testing the String length, before starting any memmove operation? 

An example: I just figured out that on 2.5.2 the replace function is not behaving as it was in the past. That is: it leads to a crash, and I suspect that it is due to the string being empty (or perhaps when there is nothing to replace in a valid string?). With the 'old' replace function this didn't happen. 

I'm sorry that I'm not able to provide any further details or sample code right now, but just wanted to contribute my two cents.

earlephilhower:
This patch isn't applied yet, and memmove handles all cases memcpy would, just ensuring data integrity.

If you can find a case that crashes that shouldn't in the current release, @KlaasjanN, we would really appreciate an issue on it.  My own quick testing doesn't show any problems with empty strings:

````
void setup() {
  Serial.begin(115200);
  String a("this is a test");
  a.replace("is", emptyString);
  Serial.printf("'%s'\n", a.c_str());
  a.replace(" ", "");
  Serial.printf("'%s'\n", a.c_str());
  String b;
  b.replace("abcd", "");
  Serial.printf("'%s'\n", b.c_str());
  b.replace("", "");
  Serial.printf("'%s'\n", b.c_str());
}

void loop() {
}
````

Gives
````
'th  a test'
'thatest'
''
''
````
and no crashes.

PR #6035, opened 2019-05-01
StreamString SSO bug

mongozmaki:
With SSO implementation in String, `StreamString::write` generates wrong strings under some circumstances.
Reason is that `String::len()` returns `strlen(sso_buf)` if SSO=true but with newly written data (in `StreamString::write`) the null-termination missing at the time `len()` is called. 
Furthermore, `len()` is called twice which is inefficient if SSO=true.

### Testcode:
```
StreamString s;
s.print('{');
s.print('\"');
s.print(String("message"));
s.print('\"');
Serial.println(s);

Expected: '{"message"'
Actual: '{"message��"'
```


.gitignore
```diff
@@ -22,3 +22,6 @@ boards.local.txt
 *.gcda
 *.o
 *.a
+
```
earlephilhower: Could we remove this file from the commit, please?

earlephilhower: I took the liberty of deleting them manually and will merge once the build is done.

mongozmaki:
Hi!
I added a pull request for the test case.
I was not sure where to put it so I put it in `host/core/test_string.cpp`. I could not test the test case but it should work.

I will submit the same to the ESP32

Issue #5915, opened 2019-03-25
ESP is not Responsive in AP_STA mode while STA is still connecting with OTA Enabled


Lan-Hekary:
-### Basic Infos

- [ X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X ] I have tested that the issue is present in current master branch (aka latest git).
- [X ] I have searched the issue tracker for a similar issue.
- [X ] If there is a stack dump, I have decoded it.
- [ X] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12E(NodeMCU)]
- Core Version:			[7a2e935f]
- Development Env:		[Platformio]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Nodemcu]
- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[80Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600] (serial upload only)

### Problem Description

The ESP in AP_STA mode does not respond to pings or webserver requests while STA is not Connected ( AP not found or Temporary Disconnect from the router )

This happens while the OTA is Enabled only .. It works normally if I turned the OTA off ..

I think it's related to the mDNS Module ..
and I think it's related to this issue as well #5866 
the problem appeared after the last repo update I made ..

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>
#include <WiFiUdp.h>
#include <ArduinoOTA.h>
#include <ESP8266WebServer.h>

#ifndef STASSID
#define STASSID "STA"
#define STAPSK "12345678"
#endif

ESP8266WebServer server(80);

IPAddress apIP(192, 168, 4, 1);
const char* ssid = "AP";
const char* password = "12345678";

void handleRoot() {
  server.send(200, "text/html", "<h1>You are connected</h1>");
}

void setup() {
Serial.begin(115200);
Serial.println("Booting");

WiFi.mode(WIFI_AP_STA);
[...]

PR #6037, opened 2019-05-02
MDNS Callback std:bind reference

hreintke:
MDNS Callback setting needs std::bind parameter by reference instead of by value


hreintke:
@d-a-v : This is not a fix for https://github.com/esp8266/Arduino/issues/5915.
It solves an issue reported by @LaborEtArs in gitter.

Issue #2218, opened 2016-07-03
noInterrupts(), interrupts()


pgScorpio:
#### Hardware

Hardware:       n.a.
Core Version:       2.2.0
### Description

Is there any good reason why we are using xt_rsil(15)/xt_rsil(0) for noInterrupts()/interrupts() ???

Doesn't ets_intr_lock()/ets_intr_unlock() do the same job without the need for storing the result of xt_rsil(...) and calling xt_wsr_ps(..) ???

i use the following class to disable interrupts for some time now without any problems:

```
    class disableInterrupts       //Locks all maskable interrupts while any instance exists...
    {
    protected:
        static uint32_t _count;

    public:
        disableInterrupts()
        {
            ets_intr_lock();
            _count++;
            if (_count == 0) //Overrun ??
            {
                //Over 4 billion locks ??
                //Something is going terribly wrong !!
                //Lock up... (and trigger the watchdog ?)
                ets_intr_unlock();
                while(1);
            }
        }

        ~disableInterrupts()
        {
            if (--_count == 0)
            {
                ets_intr_unlock();
            }
            else if (_count == 0xFFFFFFFF) //Overrun ??
            {
                //Fatal error !!
                //destructor called more often than constructor ????,  _count must be corrupted !
                //Lock up... (and trigger the watchdog ?)
                ets_intr_unlock();
                while(1);
            }
        }
   };

```

igrr:
`ets_intr_lock` increments a counter and sets `PS.INTLEVEL` to 2. `ets_intr_unlock` decrements the counter and sets `PS.INTLEVEL` to 0 if the counter is 0. We have actually used `ets_intr_lock` and `ets_intr_unlock` for a while, in `noInterrupts`/`interrupts` but this got changed because level 3 interrupts were not disabled that way.

Current implementation of `noInterrupts`/`interrupts` doesn't allow nested `noInterrupts`/`interrupts` calls. I have looked through a bunch of libraries and didn't find any use of such nested calls, so overall i think this is okay. But in general, this should be improved.

For the ESP8266, the only values of `PS.INTLEVEL` which make sense are 1, 2, and 3. Every value above 3 works the same way as 3. So in practice it is sufficient to store the value of `PS.INTLEVEL` in the first call to `noInterrupts` and then increment lock count in every successive call to `noInterrupts`. This would allow correct unwinding of `PS.INTLEVEL` values.

[...]

PR #6039, opened 2019-05-02
add regular scheduled functions, now also callable on `yield()`

d-a-v:
added `bool schedule_function_us(std::function<bool(void)> fn, uint32_t repeat_us)`
lambda must return `true` to be not removed from the schedule function list
if `repeat_us` is 0, then the function is called only once.

Legacy `schedule_function()` is preserved

~~Linked list management is simplified~~

This addition allows network drivers like ethernet chips on lwIP to be regularly called
- even if some user code loops on receiving data without getting out from main loop
  (callable from `yield()`)
- without the need to call the driver handling function
  (transparent)

This may be also applicable with common libraries (mDNS, Webserver, )


earlephilhower:
This may be more of a scheduled_functions general question, but what about interrupt safety?  One use I see for this is to put a notice into a queue in an IRQ (instead of actually doing work there).  But you could also have things like repeated ones, now, in the main app.  So isn't there a race condition in the list update stage (or if, say, an IRQ happens while parsing through the existing list)?

d-a-v:
>  but what about interrupt safety?

You are right.
Per https://github.com/esp8266/Arduino/issues/2218#issuecomment-286281984, critical sections are now protected.

earlephilhower:
That looks good and seems like it will preserve the last IRQ level which is what is needed.

I think you also need locking around `get_fn`.  It's not likely (so would be ugly to debug) but you could get an IRQ in the middle of it and a call to schedule_fcn()...which will fall get_fn again and your linked list operations will go wonky.

cores/esp8266/Arduino.h
```diff
@@ -165,6 +165,15 @@ void ets_intr_unlock();
 #define interrupts() xt_rsil(0)
 #define noInterrupts() xt_rsil(15)
 
+#define lockEnterVar(varSave, level) __asm__ __volatile__("rsil %0," __STRINGIFY(level) : "=a" (varSave))
```
earlephilhower: I think the code looks good, but I'm not sure why these `lock*` macros are in `Arduino.h`?  There is no documentation and they're unique to our setup, so I would not expect users to ever see or know how use them.  Maybe `Schedule.cpp` is the right place for them to live?

Also, and maybe this is @devyte getting in my head, would the locking mechanism make sense to be put into some sort of object?  On creation it disables IRQs and on deletion it resets IRQs.  Then you simply instantiate a InterruptLock() in your code block and it automagically does the save/restore when it is created and its lifetime ends.  I know I've seen something similar in other systems...it's a minor niggle and not something to stop the PR, though.

devyte: Oh cool, the inception works! :laughing:
It's a good idea to make it an object that locks in the constructor and unlocks in the destructor. This makes sure that the lock is always released on scope exit, whether via normal way or exception, and it also makes for cleaner code when there are multiple returns due to error conditions or whatever.

d-a-v: This is brilliant !

dok-net:
I've grepped across the source tree, and interrupts are disabled in one form or the other in any of these places:

libraries/EEPROM/EEPROM.cpp:  noInterrupts();
libraries/ESP8266SdFat/src/SdCard/SdioTeensy.cpp:      noInterrupts();
libraries/ESP8266SdFat/src/SpiDriver/DigitalPin.h:    cli();
tools/sdk/lwip2/builder/glue-esp/lwip-esp.c:    ets_intr_lock();
cores/esp8266/core_esp8266_timer.cpp:        uint32_t savedPS = xt_rsil(15); // stop other interrupts
cores/esp8266/core_esp8266_wiring_digital.cpp:      uint32_t savedPS = xt_rsil(15); // stop other interrupts
cores/esp8266/interrupts.h:        _state = xt_rsil(15);

I gather, that timer and GPIO interrupt handlers by design decision ("AVR compatibility" gets mentioned) always run with interrupts disabled, reducing the changes that an ISR gets interrupted to about zero.

I think tools/sdk/lwip2/builder/glue-esp/lwip-esp.c maintains a linked list a lot like the one that's being discussed here, and stands out as having a rather large block of code during which interrupts are disabled.
I intend to take some measurements as to the effect disabling interrupts has on the receive timings of EspSoftwareSerial and report these soon, just be be on the safe side, please mind my word of caution that too much interrupt blocking adversely affects the real-time characteristics of the platform. Also, there is a chance that the data structure being used for the schedule queue is non-optimal for this case, and there might a problem with atomic operations / memory fences even in the case where interrupts get disabled. Is anyone knowledgeable of what the compiler and CPU do to register aliases during interrupts, i.e., do they get written to RAM and restored from RAM upon interrupt entry and return? I doubt that.

Now, late at night, all I can contribute toward a solution is this article, containing supposedly lock-free code for a queue:
https://stackoverflow.com/questions/871234/circular-lock-free-buffer

I will need to look into that for EspSoftwareSerial, the ISR buffer interacts with user code and has all the possibility of issues that I suggest this PR might have.

cores/esp8266/FunctionalInterrupt.cpp
```diff
@@ -3,6 +3,8 @@
 #include "Arduino.h"
 #include <ScheduledFunctions.h>
 
+static ScheduledFunctions* scheduledInterrupts;
```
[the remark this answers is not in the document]
d-a-v: Yes I read it, I'll do that in time or sync with master if your PR is merged first.

d-a-v:
> May I bring the existence of cores/esp8266/interrupts.h to your attention? That actually gets included in Esp.cpp already.

Sure, and thanks for it. I wasn't aware (!)

> About the use of volatile, I have some reserverations - IIRC, memory fences must be handled differently.

This volatile is unnecessary, I forgot to remove it.

> please mind my word of caution that too much interrupt blocking adversely affects the real-time characteristics of the platform

Sure but they are necessary to avoid races.
Lock-free queue structure is very interesting.
Maybe a new issue would have to be opened to discuss about your comment about locking:
- use the best/correct API everywhere
- consider lock-free linked lists
- your results about measurements on the receive timings of EspSoftwareSerial when disabling interrupts

>  Is anyone knowledgeable of what the compiler and CPU do to register aliases during interrupts, i.e., do they get written to RAM and restored from RAM upon interrupt entry and return? I doubt that.

Registers used in a function (or all registers) are pushed on the stack and restored before the "reti".
I've seen that with gcc on other archs. I believe this is always mandatory when dealing with interrupts.

dok-net:
> Registers used in a function (or all registers) are pushed on the stack and restored
That's what I believe and that's the reason to use std::atomic load and store - otherwise all interrupt locking may not be much help. Or am I completely missing something?

earlephilhower:
@dok-net, what specifically are you worried about?

The SDK blob IRQ wrapper stores all the registers in use before calling an IRQ function and then restores then before returning from interrupt.  I don't think there's any particular concern there, it's a simple and common operation.

On function entry in main code you call IRQ-disable as the first state changing operation.  That either finishes uninterrupted, or an IRQ gets called.  That IRQ may call the same function, (and can't be interrupted) so will run to completion changing the linked list).  On return from interrupt the main app's locking code takes completes and it has full access to the updated list (there is no chance of the list being cached incoherently anywhere in the machine) and can do its own work...

dok-net:
@d-a-v  @earlephilhower Is there a quick explanation why linked lists are used? IIRC linked lists are discouraged for use where lock-free programming is an advantage. I just too a perfunctory glance at the code spots and couldn't figure out why linked links are used - if all the reason should be to keep entries in for repeated execution, pop and re-push should do the trick, too. I've implemented a lock-free (well, on ESP32, on single-threaded ESP8266 it disable IRQs after all ;-) ) ring buffer / circular queue for EspSoftwareSerial, which I am still testing. It's multi-producer, single-consumer capable, and could be a nice basis to #6039...

earlephilhower:
@dok-net I didn't do any of the coding on this, but I imagine linked lists were used to minimize heap usage when only a few (or common case: 0) delayed functions are in play.  There's no atomic TAS operations on the chip, though, so even simple things like mutexes aren't doable safely w/o stopping IRQs.  I would not want to guarantee that std::atomic (if available, I don't remember if it compiled or not) works as-expected with interrupts on this chip, actually.

In any case, I'm still trying to understand the concern here.  Logically, I think it's fine.  Are you concerned about performance (i.e. you have a use where you have to add items to the queue at high frequency, etc.?  W/only 32 slots even that's a pretty bounded problem) or something else?

d-a-v:
@dok-net 

> and there might a problem with atomic operations / memory fences even in the case where interrupts get disabled

The only NMI in core is in waveform which I think is orthogonal to ScheduledInterrupt.
So there is no reason an IRQ-disabled portion of code be unsafely interrupted.

dok-net:
@d-a-v the scheduled functions are processed in LIFO order … o.O

d-a-v:
Yes, is that an issue ?

dok-net:
Why would you not think so? Anything to do with timestamps or any message order at all cannot work without proper FIFO, unless totally wasting resources on reordering.

d-a-v:
This linked list is here to remember events to execute once from time to time (when interrupts or core events occur).
It is not designed to embed many/lots of them at once, nor chain them in order when there are more than one. The linked list was FIFO and I simplified it resulting in a LIFO queue (not on purpose, but with this in mind).

It is stated that the API is beta. I haven't thought of a use-case for such execution-in-order with hard dependency between them so I would be glad if you showed one to me.

I can of course revert that in this not-yet-merged PR. That would make it a slightly more complex code.
I remember you were saying the linked-list here and also in lwip2 glue is not efficient, or takes too much time. If you have better proposals, I can hear them of course.

dok-net:
I am very sorry if I sounded harsh, it's just that I have spent hours finalizing a lock-free multi-producer / single-consumer queue and now I was trying to run tests (on x86_64) on both my impl and this linked list, but cannot make the latter work at all, which astonished me - probably it's because there are not atomic stores and loads in this linked list. So unfortunately I cannot compare performance.
I'll send you a PR tonight with my (general) vision of things. Please bear with me.

d-a-v:
Nothing is harsh here !

The linked-list is protected from interrupts, the one in lwip2 is not interrupted. Why do you need atomic stores and load in it / them ?

On top of that, @earlephilhower stated earlier that there's no atomic asm operation (TAS-like) on this chip, so we can't. What would be wrong with disabled interrupts for critical operations ?

hreintke:
@d-a-v : 
One of my usecases of scheduled_functions is to make sure that there are no simultaneous I2C transactions ongoing. -> Every function that initiates an I2C is scheduled,  For that use, the FIFO nature is essential.

devyte:
Use of a simple LIFO without aging or priority can cause items deep in the stack to not be reached for long periods of time, or even ever at all. I don't think it's worth it to implement aging or priority, so I strongly suggest reverting back to FIFO, the cost should be at most one additional ptr in heap.

dok-net:
@d-a-v My starting point is the migration of the scheduler to ESP32, which has two cores and can be configured without core affinity for the "sketch" tasks. Also, I like testing on x86 and writing portable C++ in general, so unless you find issues with my code, that would not be just redundant on ESP8266, but actually make things worse there, I hope we can agree on my motivations. So yes, on ESP8266, I disable interrupts in lieu of mutexes.

d-a-v:
Latest commit restores FIFO as it is in master.
It is currently tested and running well for its purpose (background/hidden ethernet loop on cont stack, these chips can have no exposed interrupt pin like W5100 - although current test can only run W5500 with a currently unused interrupt pin).

This PR aims at be the most possible simple without changing everything internally.
Later improvements like @dok-net 's [proposal](https://github.com/d-a-v/Arduino/pull/5) should preferably be proposed right in the core repository for opened discussion.

tests/host/common/mock.h
```diff
@@ -131,7 +131,7 @@ void mockUDPSwallow (size_t copied, char* ccinbuf, size_t& ccinbufsize);
 class UdpContext;
 void register_udp (int sock, UdpContext* udp = nullptr);
 
-class InterruptLock { };
+//class InterruptLock { };
```
earlephilhower: Minor niggle, can we drop commented dead code like here?

cores/esp8266/Schedule.h
```diff
@@ -6,20 +6,23 @@
 #define SCHEDULED_FN_MAX_COUNT 32
 #define SCHEDULED_FN_INITIAL_COUNT 4
 
-// Warning 
-// This API is not considered stable. 
-// Function signatures will change.
+// This API was not considered stable but is now fossilizing.
```
earlephilhower: s/fossilizing/stabilizing/.  Fossilizing == dead. :)

cores/esp8266/Schedule.h
```diff
@@ -6,20 +6,23 @@
...
+// or `yield` is called,
 // or `run_scheduled_functions` is called.
 // Use std::bind to pass arguments to a function, or call a class member function.
 // Note: there is no mechanism for cancelling scheduled functions.
 // Keep that in mind when binding functions to objects which may have short lifetime.
 // Returns false if the number of scheduled functions exceeds SCHEDULED_FN_MAX_COUNT.
 bool schedule_function(std::function<void(void)> fn);
 
-// Run all scheduled functions. 
+// Run given function every at least <repeat_us> microseconds until it returns false
```
earlephilhower: The actual behavior is a little different.  I would suggest something like:

`````
// Run given function periodically about every <repeat_us> microseconds until it returns false
// Note that it may be more than <repeat_us> microseconds between calls if `yield` is not called
// frequently, and therefore should not be used for timing critical operations.
````

File: cores/esp8266/Schedule.cpp [hunk 6]
```diff
@@ -60,19 +75,45 @@ bool schedule_function(std::function<void(void)> fn)
...
-	sLast  = NULL;
-    while (rFirst) {
-        scheduled_fn_t* item = rFirst;
-        rFirst = item->mNext;
-        item->mFunc();
-        item->mFunc = std::function<void(void)>();
-        recycle_fn(item);
+	scheduled_fn_t* lastRecurring = nullptr;
+	scheduled_fn_t* toCall = sFirst;
+    while (toCall) {
```
earlephilhower: Weird indenting (tab/space mixup?)

cores/esp8266/Schedule.cpp
```diff
@@ -60,19 +75,45 @@ bool schedule_function(std::function<void(void)> fn)
...
-	scheduled_fn_t* rFirst = sFirst;
-	sFirst = NULL;
-	sLast  = NULL;
-    while (rFirst) {
-        scheduled_fn_t* item = rFirst;
-        rFirst = item->mNext;
-        item->mFunc();
-        item->mFunc = std::function<void(void)>();
-        recycle_fn(item);
+	scheduled_fn_t* lastRecurring = nullptr;
```
earlephilhower: I *think* this is IRQ safe, but we need one more pair of eyes.  My reasoning:

Enter `run_scheduled_function`
`toCall = sFirst`
<one more more `while` loops advance and call some things, but not done yet>
IRQ happens and is taken
IRQ calls `schedule_function`
The linked list is *appended* to
IRQ returns
At this point we have a new entry appended to the end of the list, but that's OK
<while continues, going thru the newly added entry at the end, too>

d-a-v: > At this point we have a new entry appended to the end of the list, but that's OK

The new entries may not be executed on that loop if item is (was) the last one and interrupt added more, because `toCall` (= old item's next) would be already nullptr. But indeed that's OK.

cores/esp8266/Schedule.cpp
```diff
@@ -60,19 +75,45 @@ bool schedule_function(std::function<void(void)> fn)
...
-        scheduled_fn_t* item = rFirst;
-        rFirst = item->mNext;
-        item->mFunc();
-        item->mFunc = std::function<void(void)>();
-        recycle_fn(item);
+    scheduled_fn_t* lastRecurring = nullptr;
+    scheduled_fn_t* toCall = sFirst;
+    while (toCall) {
+        scheduled_fn_t* item = toCall;
+        toCall = item->mNext;
```
devyte: toCall = toCall->mNext is better to make advancement more obvious

cores/esp8266/Schedule.cpp
```diff
@@ -60,19 +75,45 @@ bool schedule_function(std::function<void(void)> fn)
...
     return true;
 }
 
+IRAM_ATTR // called from ISR
+bool schedule_function(std::function<void(void)> fn)
+{
+    return schedule_function_us([&fn](){ fn(); return false; }, 0);
+}
+
 void run_scheduled_functions()
```
devyte: This function should have a big phat comment explaining the assumption that ISRs will ever only add to the list and never remove items.

cores/esp8266/Schedule.h
```diff
@@ -6,20 +6,25 @@
...
 // Note: there is no mechanism for cancelling scheduled functions.
 // Keep that in mind when binding functions to objects which may have short lifetime.
 // Returns false if the number of scheduled functions exceeds SCHEDULED_FN_MAX_COUNT.
 bool schedule_function(std::function<void(void)> fn);
 
-// Run all scheduled functions. 
+// Run given function periodically about every <repeat_us> microseconds until it returns false.
+// Note that it may be more than <repeat_us> microseconds between calls if `yield` is not called
+// frequently, and therefore should not be used for timing critical operations.
+bool schedule_function_us(std::function<bool(void)> fn, uint32_t repeat_us);
```
devyte: Are you sure an overload isn't better? i. e. no _us suffix.

cores/esp8266/Schedule.cpp
```diff
@@ -60,19 +75,45 @@ bool schedule_function(std::function<void(void)> fn)
         sLast->mNext = item;
     }
     sLast = item;
+
+    if (repeat_us)
+        item->callNow.reset(repeat_us);
```
devyte: repeat_us is uint32_t. The polledTimeout class used is FastUs. Shouldn't there be a sanity check against the max possible timeout value?

cores/esp8266/Schedule.h
d-a-v: That's a point of view. I like to have units right under my nose so to not make mistakes.
One could say there's no unit in delay(), and that's a shame :)

dok-net:
@d-a-v What happens if a scheduled function/task calls yield() etc. itself? Calling `schedule_function()` recursively seems safe, but what about this?

d-a-v:
> What happens if a scheduled function/task calls yield() etc. itself?

My thinking is these scheduled functions should be thought as interrupt functions (that are artificially shifted out from sys stack).
Only the minimal should be done in them, no `delay` no `yield`.

If it happens we need to be calling other functions that themselves call yield, then we can add a boolean fence set up in `run_schedule_functions` that will be checked in yield family functions.

> I am beginning to think that loop() should be just another (initially) scheduled function

That can be something we think about when we will move from nonos-sdk to rtos-sdk which will happen sometimes soon enough.

> That said, wasn't there a scheduler for adding loops() in place all the time...?

check `loop_task()` and `loop_wrapper()`

dok-net:
@d-a-v please revisit my edited comment above regarding https://github.com/arduino-libraries/Scheduler/

d-a-v:
These schedulers are multitasking schedulers with each its own task.
When we switch to rtos-sdk, we already have that with FreeRTOS pretty much like in arduino-esp32 which you know well because of EspSoftwareSerial compatible with esp32. On esp8266 we have limited ram so maybe it's not a good idea to have a separate task for each task (as opposed to scheduled functions running on the same stack).

cores/esp8266/Schedule.cpp
```diff
@@ -60,19 +75,45 @@ bool schedule_function(std::function<void(void)> fn)
...
+
     return true;
 }
 
+IRAM_ATTR // called from ISR
+bool schedule_function(std::function<void(void)> fn)
+{
+    return schedule_function_us([&fn](){ fn(); return false; }, 0);
```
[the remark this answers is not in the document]
d-a-v: Done in last commit, ethernet tests are still running well.

cores/esp8266/Schedule.cpp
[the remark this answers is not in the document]
d-a-v: Good point.
We don't want to lock while executing the scheduled function themselvses.
One solution is to tag variables with `volatile`.
Another one is to build a local copy of the list while being locked, then unlock and run that list.

dok-net:
@d-a-v I've rebased https://github.com/d-a-v/Arduino/pull/6 - there's still something to do, one critical bug and performance related nullptr and rvalue refs.

cores/esp8266/Schedule.cpp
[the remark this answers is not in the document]
d-a-v: I think I must step back on this.
Locking IRQ is the right thing to do when the value of a variable can be modified in a TAS block.
Even if there are registers holding some variables, they will not be changed while in a locked block because an IRQ won't occur in that block, regardless whether the variable is cached in a register.

d-a-v:
> one critical bug

Which is it ?

dok-net:
Null pointer access in last line of get_fn_unsafe

dok-net:
@d-a-v Big oops, something we've all missed:

```
if (item->callNow)
{
    if (item->mFunc())
    {
         lastRecurring = item;
    }

```

lastRecurring should get updated `if (!item->callNow)` …...

dok-net:
@d-a-v Could we maybe agree to rename toCall as next? I've really a deep expectation in my mind that toCall is the current, to call, item, and it's making the piece of code very hard to reason about.

devyte:
On that case:
toCall => nextItem
item => currentItem or currItem
or something along those lines

d-a-v:
Already updated in #6137

PR #6047, opened 2019-05-04
Alternative to #6003: Fixes and implementation to expose attachInterruptArg in Arduino.h

dok-net:
A C++ std library functional/bind/lambdas based quite clean implementation of the same functionality as in #6003.


dok-net:
@d-a-v : could #6039 please take into account all the work and research I've done on fixing the interrupt handling? Please remember, that at heart, #6038 is the cleanest ISR vs. ICACHE_RAM safe implementation at this present stage I have to offer, the caveat being that the user cannot mix functional and non-functional attaches and detaches without care. #6003 provides this extra ease, at the expense of copied struct definition, dependency inversion between FunctionalInterrupt and core_esp8266_wiring_digital and the extra functional field only for the purpose of supporting FunctionalInterrupt. My take is still to go for clean dependencies over ease of use :-)

d-a-v:
@dok-net Thanks for your work,
Lots of nice improvement-PRs including yours are waiting for the bugfix-one-month-late core-2.5.1 release so they can be properly tested / reviewed.
2.5.1 is imminent now (well, after some neverending recent small but mandatory fixing merged commits).

d-a-v:
@dok-net #6039 is orthogonal to your PRs #6003 #6038 #6047.
What do you mean with " take into account all the work and research" ? Am I stepping on your toes ?
(In the files you modified I fixed a `static` declared in a `.h`, not a real issue. On the other hand I added an `IRAM_ATTR` define which might please your esp8266/esp32 lib :)

dok-net:
Dear @d-a-v , no, stepping on my toes is much to hard a word, I am just a little panicky that #6039 might cause a merge mess for me, I only wanted to point that out. As far a bug fix vs. feature is concerned, I've created issue report #6049 to hopefully motivate sufficiently why this PR is a bug fix and should be in the very next bugfix release.
SoftwareSerial is also pending release reliant on the availability of attachInterruptArg - of course I could back-port, but I'd rather have it all matching up nicely as it does in my own repos' configurations right now. Naturally, I can only try to show that this PR is a relatively minor fix to a real problem without expected side-effects...

d-a-v:
@dok-net 
There are #6048+#6055 and this one #6047.
Does OP states #6047 replace both ? 

Also I understand #5922 should allow functional in IRAM and there are issues about that per @hreintke 's comment https://github.com/esp8266/Arduino/pull/6055#issuecomment-491584558 and it needs to be sorted out.

dok-net:
@d-a-v 
> There are #6048+#6055 and this one #6047.
> Does OP states #6047 replace both ?
Yes, it should work this way (in git).

> Also I understand #5922 should allow functional in IRAM and there are issues about that per @hreintke 's comment #6055 (comment) and it needs to be sorted out.
I couldn't quite wrap my head around the irom.text etc. meddling, and unfortunately the above mentioned comment features incomplete code and I can't figure it quite out either - I would like to trust that the parts of std::function<void()> are either inlined or linker-edited into IRAM as per #5922, and that is sufficient for #6047, unless someone attaches/detaches ISRs from inside executing ISRs, on which opinions vary widely :-)

hreintke:
@dok-net 
I just updated the comment in  #6055 to include a complete sketch. If anything still unclear, just let me know.

dok-net:
@d-a-v @hreintke To keep you informed, with this std::function based ISR attach code, I see sporadic crashes after reset. So yes, there is still something not in order with the IRQ-safety of std::function and friends.

hreintke:
@dok-net @d-a-v : 
I can do further checking on this but am lost in the PR's and associated commits.
Do I need only git master + this PR or is it also dependent on others like the Ticker etc ?

dok-net:
@hreintke I almost can't resist giving an answer that's just as cryptic :-)
No, don't worry, I am glad you care! I am putting a lot of work into keeping the PRs clear to merge with master each by themselves. If this doesn't work now and them, yes, me too, I am dissatisfied and disbelieving how long git and the GitHub PRs exist without proper dependency management for the inter-PR dependencies. Stuffing all into a single branch and thus one PR can't quite be the solution, either.
Anyway, Ticker has nothing to do with #6049, hence this PR stands alone. For differential diagnosis, you can try #6048 plus #6055, which are what I use while this #6047 unfortunately doesn't work.

dok-net:
@hreintke I almost can't resist giving an answer that's just as cryptic :-)
No, don't worry, I am glad you care! I am putting a lot of work into keeping the PRs clear to merge with master each by themselves. If this doesn't work now and then, yes, me too, I am dissatisfied and disbelieving how long git and the GitHub PRs exist without proper dependency management for the inter-PR dependencies. Stuffing all into a single branch and thus one PR can't quite be the solution, either.
Anyway, Ticker has nothing to do with #6049, hence this PR stands alone. For differential diagnosis, you can try #6055 (this includes #6048), which are what I use while this #6047 unfortunately doesn't work.

hreintke:
@dok-net @d-a-v @igrr 
Is there a way that, for debug purposes, we can intercept the "missed cache interrupt" and then check whether we are in  interrupt code. If no -> continue as usual, if yes -> panic() and output somehow the offending function address ?

earlephilhower:
@hreintke, There's no cache miss handler code.  Cache fetches are done completely in HW.  If the HW state machine can't engage the SPI bus to do the read you get a nonrecoverable HW exception at that point and crash hard.  If it can access the bus, it will work since interrupts are just normal codepaths and nothing fancy on these machines.

As a **debug only** method, you might be able to do something fishy to the SPI bus at the start of our interrupt dispatch code and then undo it on return.  Then panics will occur should ISR code attempt to go out to flash.  But that's debug only and it will break SPI, WiFi persistence, and anything else that access flash as a data store.

I think it's probably overkill, though.  Even the ISR-in-RAM check is a bit pedantic since it is possible for an app to never access flash as data store, or an app could disable IRQs before doing so (and killing any IRQ generated stuff like PWM, etc.).

dok-net:
@d-a-v It seems you've some time at your hands today to talk - I've a question: What is the state of the IRAM linker segment manipulations w.r.t. std::function now? I have hacked a fix for EspSoftwareSerial (master HEAD) to use that in the meantime until my bigger PRs get reviewed and merged, and somehow it's working using the FunctionalInterrupts API instead of the C-style attachInterruptArg (in my PRs). Before, it seemed unstable. Also, a standalone test in libraries/FunctionalInterrupt/ that I had seen failing, as I believed due to flash accessing during interrupt service - runs fine. As much as I understood it, std::function<void(void)> are OK, does this depend on the captures in any way?

d-a-v:
> As much as I understood it, std::function<void(void)> are OK, does this depend on the captures in any way?

Since #5922, all functional effective code is moved into IRAM.
Matching signature is *supposed to be*: `std::function<any(...)>::operator()() const`.
In [this comment](https://github.com/esp8266/Arduino/pull/5922/files#r272853465) I didn't check the capture influence to the generated object name. 
You can play with https://demangler.com/ and check if your symbol matches `_ZNKSt8functionIF*EE*`

You can also more easily check from a running sketch whether a lambda is linked IRAM [with this trick](https://github.com/esp8266/Arduino/pull/5780#issuecomment-484070364).

dok-net:
@d-a-v I believe, with the information you provided me, that this PR now contains a safe refactoring of the IRQ code that has the relevant code paths in IRAM. It certainly is a lot cleaner than #6055 .

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -147,31 +134,15 @@ void ICACHE_RAM_ATTR interrupt_handler(void *arg) {
       // to make ISR compatible to Arduino AVR model where interrupts are disabled
       // we disable them before we call the client ISR
       uint32_t savedPS = xt_rsil(15); // stop other interrupts
```
d-a-v: What about using `esp8266:InterruptLock` here ?

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,88 @@
...
+      //}, FALLING); // works on ESP8266
+    };
+    ~Button() {
+      detachInterrupt(PIN);
+    }
+
+#if defined(ESP8266)
+    void ICACHE_RAM_ATTR buttonIsr()
+#elif defined(ESP32)
+    void IRAM_ATTR buttonIsr()
```
d-a-v: maybe, on top of the file, we can use
```
#ifndef IRAM_ATTR
#define IRAM_ATTR ICACHE_RAM_ATTR
#endif
```
that, because:
- this is an example in esp8266 repository
- `IRAM_ATTR` is now defined in esp8266 core master

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,88 @@
...
+    Button(uint8_t reqPin) : PIN(reqPin) {
+      pinMode(PIN, INPUT_PULLUP);
+      // Arduino API:
+      //attachInterruptArg(PIN, [](void* self) {
+      //  static_cast<Button*>(self)->buttonIsr();
+      //}, this, FALLING); // fails on ESP8266: "buttonIsr not in IRAM"
+      //attachInterruptArg(PIN, reinterpret_cast<void(*)(void*)>(&buttonIsr_static), this, FALLING); // works on ESP8266
+      attachInterrupt(PIN, std::bind(&Button::buttonIsr, this), FALLING); // works on ESP8266
+      // FunctionalInterrupt.h API:
+      //attachScheduledInterrupt(PIN, [this](const InterruptInfo & ii) {
```
d-a-v: What about a second example, this one called `InterruptFunctional.ino` and the commented one `ScheduledInterruptFunctional.ino` ?

d-a-v:
A question about your use of `std::move()`,
when using it, argument is supposed to be passed as rvalue reference `type&& arg` no ?
per [this "course"](https://www.cprogramming.com/c++11/rvalue-references-and-move-semantics-in-c++11.html) (which I find worth being read and re-read :)

dok-net:
@d-a-v No, I don't think so, that's an oversimplification. https://en.cppreference.com/w/cpp/utility/move explains this better. What I am trying to express/achieve is that the move constructors/assignments are selected if available.

d-a-v:
Then to be consistent, we'd need to have both
```
void attachScheduledInterrupt(uint8_t pin, 
                              std::function<void(InterruptInfo)> scheduledIntRoutine,
                              int mode)
```
and
```
void attachScheduledInterrupt(uint8_t pin, 
                              std::function<void(InterruptInfo)>&& scheduledIntRoutine,
                              int mode)
```
(both using `std::move()`)
(like they do in STL for example)
(no asking to this here, because if we go that way we'd need to update the whole API)

dok-net:
@d-a-v it's certainly desirable avoiding copying objects all the time for no good reason - but yes, that's a lot of extra work doing it just for consistency. Are you OK with me using it when doing new things step by step, tho?

d-a-v:
> Are you OK with me using it when doing new things step by step, tho?

Of course yes.
Just not because we are on a small MCU should we avoid any cheap kind of optimization :)

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -169,6 +169,13 @@ extern void __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int mode)
   attachInterrupt(pin, std::bind(userFunc), mode);
 }
 
+void set_interrupt_handlers(uint8_t pin, std::function<void()> userFunc, uint8_t mode)
+{
+	interrupt_handler_t* handler = &interrupt_handlers[pin];
+	handler->fn = userFunc;
+	handler->mode = 0;
```
devyte: = mode;

devyte: = mode;

dok-net: aw, minor implementation detail :-) :-) Thanks!!!

libraries/FunctionalInterrupt/src/FunctionalInterrupt.cpp
```diff
@@ -0,0 +1,23 @@
+#include "FunctionalInterrupt.h"
```
devyte: The current filename is misleading, it confused me for a moment. Please change it to ScheduledInterrupts or similar.

libraries/FunctionalInterrupt/src/FunctionalInterrupt.h
```diff
@@ -0,0 +1,18 @@
+#ifndef FUNCTIONALINTERRUPT_H
+#define FUNCTIONALINTERRUPT_H
+
+#include <functional>
+
+// Structures for communication
+
+struct InterruptInfo
```
devyte: I thought from the changes in wiring that this was dropped. Why is it available only for scheduled interrupts, and not for normal interrupts?
I'm thinking maybe it should be implemented in a better way, e.g.: read the values as early as possible in the ISR.

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,75 @@
...
+    uint32_t checkPressed() {
+      if (pressed) {
+        Serial.printf("Button on pin %u has been pressed %u times\n", PIN, numberKeyPresses);
+        pressed = false;
+      }
+      return numberKeyPresses;
+    }
+
+  private:
+    const uint8_t PIN;
```
devyte: I know this is only an example, but let's try to maintain style: please change the naming convention of the members, e.g.: mPin, mNumberKeyPresses, mPressed.

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,75 @@
...
+
+Button* button1;
+Button* button2;
+
+
+void setup() {
+  Serial.begin(115200);
+  Serial.println("FunctionalInterrupt test/example");
+
+  button1 = new Button(BUTTON1);
```
devyte: I suspect you're doing this, because a global instance doesn't work (can't attach in the constructor). Please do one of the following:
- change the class implementation to follow Arduino patterns, e.g.: add begin() and end() methods that do the actual hw init and release.
- implement a default constructor that does nothing, and an assignment constructor that does the same as the current pin constructor. Then, instantiate globally with default, and here do an assignment of a temp object built with the pin constructor
- add a comment here explaining why you're doing a new, that this is ok only in setup because the mem is never released

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,75 @@
...
+#define BUTTON1 D4
+#define BUTTON2 D3
+#else
+#define BUTTON1 2
+#define BUTTON2 0
+#endif
+
+class Button {
+  public:
+    Button(uint8_t reqPin) : PIN(reqPin) {
```
devyte: Minor nitpic: const arg

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,75 @@
...
+    void IRAM_ATTR buttonIsr() {
+      numberKeyPresses += 1;
+      pressed = true;
+    }
+
+    static void IRAM_ATTR buttonIsr_static(Button* const self) {
+      self->buttonIsr();
+    }
+
+    uint32_t checkPressed() {
```
devyte: checkClearPressed(), to be clear about what happens (I was going to suggest const method, then I realized that this isn't just a check, but also a clear of a member)

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,75 @@
...
+#else
+#define BUTTON1 2
+#define BUTTON2 0
+#endif
+
+class Button {
+  public:
+    Button(uint8_t reqPin) : PIN(reqPin) {
+      pinMode(PIN, INPUT_PULLUP);
+      //attachInterruptArg(PIN, reinterpret_cast<void(*)(void*)>(&buttonIsr_static), this, FALLING);
```
devyte: Remove commented line

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
```diff
@@ -0,0 +1,75 @@
...
+    ~Button() {
+      detachInterrupt(PIN);
+    }
+
+    void IRAM_ATTR buttonIsr() {
+      numberKeyPresses += 1;
+      pressed = true;
+    }
+
+    static void IRAM_ATTR buttonIsr_static(Button* const self) {
```
devyte: Remove unused method

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -110,24 +110,10 @@ typedef void (*voidFuncPtrArg)(void*);
...
 typedef struct {
   uint8_t mode;
-  voidFuncPtr fn;
-  void * arg;
-  bool functional;
+  std::function<void()> fn;
 } interrupt_handler_t;
 
-//duplicate from functionalInterrupt.h keep in sync
-typedef struct InterruptInfo {
```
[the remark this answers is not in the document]
dok-net: `This functionality is being dropped.`

I am not sure what you mean - are you referring to my PR removing this from core, or is there a preexisting decision to remove this altother in an upcoming release? "This" referring to the integral timestamp and value capture.

My take was this: the code duplication to quick-fix the include dependency circle is a bug, that needs to be fixed.

libraries/FunctionalInterrupt/src/FunctionalInterrupt.cpp
dok-net: I will happily do so, I feel the same way, just didn't dare do it by myself.

cores/esp8266/core_esp8266_wiring_digital.cpp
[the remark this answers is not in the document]
devyte: What I mean is that this PR seemsnto be dropping the value/micros info functionality for normal interrupts.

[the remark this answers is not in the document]
dok-net: I think you realized by yourself in the meantime, that the "InterruptInfo" is only available to "FunctionalInterrupt"-style ISRs - regular old ISR either have no argument at all, or provide their own "void*" to this effect. So it's not only populated for "Functional"-ISRs at all, but also the other C-style ISRs set via the Arduino.h extern "C" API have no way of getting that structure. Which is think is fine.
On the other hand, for "scheduled ISRs" having the actual timestamp and the value of the GIOP at the time the IRQ occurred is essential.
So, my cleanup, that is, removing the special "scheduled ISR" support from the core and offering the API via the library - your right, it should be renamed - is probably what we both think is the right thing to do, isn't it?

libraries/FunctionalInterrupt/src/FunctionalInterrupt.h
dok-net: Please check my comment above.

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
dok-net: Opted for last in your list.

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
dok-net: decided on m_PIN, m_numberKeyPresses, et cetera - hope that's OK, too.

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
dok-net: Opted for checkResetPressed()

dok-net: Opted for testResetPressed()

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
dok-net: Done

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
dok-net: Done, too

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -110,24 +110,10 @@ typedef void (*voidFuncPtrArg)(void*);
...
-	uint32_t micro;
-} InterruptInfo;
-
-typedef struct {
-	InterruptInfo* interruptInfo;
-	void* functionInfo;
-} ArgStructure;
-
-static interrupt_handler_t interrupt_handlers[16] = { {0, 0, 0, 0}, };
+static interrupt_handler_t interrupt_handlers[16] = { {0, nullptr}, };
```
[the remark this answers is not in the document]
dok-net: No. I knew I had it that way for a reason, CI fails, I didn't remember at first:

/home/travis/arduino_ide/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring_digital.cpp:116:57: error: missing initializer for member 'interrupt_handler_t::fn' [-Werror=missing-field-initializers]

 static interrupt_handler_t interrupt_handlers[16] = { 0 };

libraries/FunctionalInterrupt/examples/Functional/Functional.ino
devyte: Actually, I was wrong, the correct style we have adopted for this is underscore, e.g.: _PIN, _NumberKeyPresses, etc. Apologies for the confusion, I was reading other code in between.

cores/esp8266/core_esp8266_wiring_digital.cpp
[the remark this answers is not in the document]
devyte: Oh this is one of those issues in our gcc version that was only fixed later (apparently fixed in gcc5)

dok-net:
@d-a-v I was asked by @devyte to test and improve the performance of this patch, such that it is no worse for void(*)(void) ISRs over before. I am now seing a major improvement in performance in my test, but I could have implemented a horrible bug :-) Could you please read `core_esp8266_wiring_digital.cpp` over again, with a keen eye out on `interrupt_handler(void*)`? Thanks!

dok-net:
Some test results.
But first, I noticed that in master, `set_interrupt_handlers()` is missing the `ICACHE_RAM_ATTR` attribution - which when done right adds a few bytes to IRAM use.

Master including this fix has following memory use for small test case using void(*)(void) ISRs:

```
Executable segment sizes
IROM   *: 231648          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 28000   \ 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...)
DATA   *: 1272  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 744   ) \ 81920 - constants             (global, static) in RAM\HEAP
BSS    *: 25224 )         - zeroed variables      (global, static) in RAM\HEAP
Program size: 261,664 bytes (used 25% of a 1,044,464 byte maximum) (27.89 secs)
Minimum Memory Usage: 27240 bytes (33% of a 81920 byte maximum)
```

It performs two cascaded IRQs with an avg. latency/CPU_cycles = 961.

Just switching core to this branch, the results are:
```
Executable segment sizes
IROM   *: 231664          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 27960   \ 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...)
DATA   *: 1272  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 752   ) \ 81920 - constants             (global, static) in RAM\HEAP
BSS    *: 25296 )         - zeroed variables      (global, static) in RAM\HEAP
Program size: 261,648 bytes (used 25% of a 1,044,464 byte maximum) (27.34 secs)
Minimum Memory Usage: 27320 bytes (33% of a 81920 byte maximum)
```

Performance is: avg. Latency/CPU_cycles = 891.

Next, changing the test case from plain function pointer to std::function for a void(int) curried with an integer value, this is result:
```
Executable segment sizes
IROM   *: 231952          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 27840   \ 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...)
DATA   *: 1272  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 732   ) \ 81920 - constants             (global, static) in RAM\HEAP 
BSS    *: 25296 )         - zeroed variables      (global, static) in RAM\HEAP
Program size: 261,796 bytes (used 25% of a 1,044,464 byte maximum) (1.85 secs)
Minimum Memory Usage: 27300 bytes (33% of a 81920 byte maximum)
```

Performing at avg. Latency/CPU_cycles = 929.

The sketch:

```
#include <interrupts.h>

uint32_t start;
uint8_t toggle_d7 = HIGH;
uint8_t toggle_d5 = HIGH;

void IRAM_ATTR irq_d8(int x)
{
	toggle_d5 ^= HIGH;
	digitalWrite(D5, toggle_d5);
}

uint32_t sum_delay = 0;
constexpr uint32_t MAXREPEATS = 1000;
uint32_t repeats = 0;
void IRAM_ATTR irq_d6(int x)
{
	auto cpuCycle = ESP.getCycleCount();
	sum_delay += cpuCycle - start;
	++repeats;
	start = cpuCycle;
	if (repeats < MAXREPEATS)
	{
		toggle_d7 ^= HIGH;
		digitalWrite(D7, toggle_d7);
	}
}

void setup()
{
	Serial.begin(115200);
	pinMode(D7, OUTPUT);
	digitalWrite(D7, HIGH);
	pinMode(D8, INPUT_PULLUP);
	pinMode(D5, OUTPUT);
	digitalWrite(D5, toggle_d5);
	pinMode(D6, INPUT_PULLUP);
	attachInterrupt(D8, std::bind(irq_d8, 42), CHANGE);
	attachInterrupt(D6, std::bind(irq_d6, 43), CHANGE);
	start = ESP.getCycleCount();
	toggle_d7 ^= HIGH;
	digitalWrite(D7, toggle_d7);
}

void loop()
{
	if (repeats >= MAXREPEATS)
	{
		{
			esp8266::InterruptLock lock;
			Serial.print("avg. Latency/CPU_cycles = ");
			Serial.println(sum_delay / repeats);
			sum_delay = 0;
			repeats = 0;
			start = ESP.getCycleCount();
			toggle_d7 ^= HIGH;
		}
		digitalWrite(D7, toggle_d7);
	}
}
```

dok-net:
Latest master compared to rebased PR branch.
Master:
```
Executable segment sizes
IROM   *: 234064          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 27580   \ 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...)
DATA   *: 1244  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 1600  ) \ 81920 - constants             (global, static) in RAM\HEAP
BSS    *: 25408 )         - zeroed variables      (global, static) in RAM\HEAP
Program size: 264,488 bytes (used 25% of a 1,044,464 byte maximum) (2.74 secs)
Minimum Memory Usage: 28252 bytes (34% of a 81920 byte maximum)
```
PR:
```
Executable segment sizes
IROM   *: 234080          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 27544   \ 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...)
DATA   *: 1244  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 1608  ) \ 81920 - constants             (global, static) in RAM\HEAP
BSS    *: 25480 )         - zeroed variables      (global, static) in RAM\HEAP
Program size: 264,476 bytes (used 25% of a 1,044,464 byte maximum) (29.70 secs)
Minimum Memory Usage: 28332 bytes (35% of a 81920 byte maximum)
```

Sketch that compiles both in master and PR:
```
#include <interrupts.h>

uint32_t start;
uint8_t toggle_d8 = HIGH;
uint8_t toggle_d6 = HIGH;

void IRAM_ATTR irq_d7(void* x)
{
	toggle_d8 ^= HIGH;
	digitalWrite(D8, toggle_d8);
}

uint32_t sum_delay = 0;
constexpr uint32_t MAXREPEATS = 1000;
uint32_t repeats = 0;
void IRAM_ATTR irq_d5(void* x)
{
	auto cpuCycle = ESP.getCycleCount();
	sum_delay += cpuCycle - start;
	++repeats;
	start = cpuCycle;
	if (repeats < MAXREPEATS)
	{
		toggle_d6 ^= HIGH;
		digitalWrite(D6, toggle_d6);
	}
}

void setup()
{
	Serial.begin(9600);
	pinMode(D8, OUTPUT);
	digitalWrite(D8, toggle_d8);
	pinMode(D7, INPUT_PULLUP);
	pinMode(D6, OUTPUT);
	digitalWrite(D6, toggle_d6);
	pinMode(D5, INPUT_PULLUP);
	attachInterruptArg(D7, irq_d7, (void*)42, CHANGE);
	attachInterruptArg(D5, irq_d5, (void*)43, CHANGE);
	start = ESP.getCycleCount();
	toggle_d6 ^= HIGH;
	digitalWrite(D6, toggle_d6);
}

void loop()
{
	if (repeats >= MAXREPEATS)
	{
		{
			esp8266::InterruptLock lock;
			Serial.print("avg. Latency/us = ");
			Serial.println(1.0 * sum_delay / repeats / ESP.getCpuFreqMHz());
			sum_delay = 0;
			repeats = 0;
			start = ESP.getCycleCount();
			toggle_d6 ^= HIGH;
		}
		digitalWrite(D6, toggle_d6);
	}
}
```

dok-net:
Here's the latest in performance numbers and what the new Delegate class can do for interrupts. See the comments with the measurements below in this working example:
```
#include <interrupts.h>

uint32_t start;
uint8_t toggle_d8 = HIGH;
uint8_t toggle_d6 = HIGH;

void IRAM_ATTR irq_d7(void* x)
{
	toggle_d8 ^= HIGH;
	digitalWrite(D8, toggle_d8);
}

void IRAM_ATTR irq_d7_v()
{
	irq_d7(nullptr);
}

uint32_t sum_delay = 0;
constexpr uint32_t MAXREPEATS = 1000;
uint32_t repeats = 0;
void IRAM_ATTR irq_d5(void* x)
{
	auto cpuCycle = ESP.getCycleCount();
	sum_delay += cpuCycle - start;
	++repeats;
	start = cpuCycle;
	if (repeats < MAXREPEATS)
	{
		toggle_d6 ^= HIGH;
		digitalWrite(D6, toggle_d6);
	}
}

void IRAM_ATTR irq_d5_v()
{
	irq_d5(nullptr);
}

void setup()
{
	Serial.begin(9600);
	pinMode(D8, OUTPUT);
	digitalWrite(D8, toggle_d8);
	pinMode(D7, INPUT_PULLUP);
	pinMode(D6, OUTPUT);
	digitalWrite(D6, toggle_d6);
	pinMode(D5, INPUT_PULLUP);
	
	attachInterrupt(D7, irq_d7_v, CHANGE);
	attachInterrupt(D5, irq_d5_v, CHANGE);
	// master / 2.6.2: avg. Latency/us = 12.29
	// avg. Latency/us = 11.42

	//attachInterruptArg(D7, irq_d7, (void*)42, CHANGE);
	//attachInterruptArg(D5, irq_d5, (void*)43, CHANGE);
	// master / 2.6.2: avg. Latency/us = 12.04
	// avg. Latency/us = 11.14

	//attachInterrupt(D7, { irq_d7, (void*)42 }, CHANGE);
	//attachInterrupt(D5, { irq_d5, (void*)43 }, CHANGE);
	// avg. Latency/us = 11.14

	//attachInterrupt(D7, std::bind(irq_d7, (void*)42), CHANGE);
	//attachInterrupt(D5, std::bind(irq_d5, (void*)43), CHANGE);
	// avg. Latency/us = 11.62

	start = ESP.getCycleCount();
	toggle_d6 ^= HIGH;
	digitalWrite(D6, toggle_d6);
}

void loop()
{
	if (repeats >= MAXREPEATS)
	{
		{
			esp8266::InterruptLock lock;
			Serial.print("avg. Latency/us = ");
			Serial.println(1.0 * sum_delay / repeats / ESP.getCpuFreqMHz());
			sum_delay = 0;
			repeats = 0;
			start = ESP.getCycleCount();
			toggle_d6 ^= HIGH;
		}
		digitalWrite(D6, toggle_d6);
	}
}
```

dok-net:
Update:
I've removed, for the time being, Delegate in favor of plain std::function.
Here are the build results, for a `FunctionalInterrupt.h`, `attachScheduledInterrupt` example:
Master:
```
ICACHE *: 32768           - flash instruction cache
IROM   *: 237528          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 27653   \ 32768 - code in IRAM          (IRAM_ATTR, ISRs...)
DATA   *: 1496  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 1000  ) \ 81920 - constants             (global, static) in RAM\HEAP
BSS    *: 25888 )         - zeroed variables      (global, static) in RAM\HEAP
```
This PR:
```
ICACHE *: 32768           - flash instruction cache
IROM   *: 237720          - code in flash         (default or ICACHE_FLASH_ATTR)
IRAM   *: 27613   \ 32768 - code in IRAM          (IRAM_ATTR, ISRs...)
DATA   *: 1496  )         - initialized variables (global, static) in RAM\HEAP
RODATA *: 992   ) \ 81920 - constants             (global, static) in RAM\HEAP
BSS    *: 25944 )         - zeroed variables      (global, static) in RAM\HEAP
````

**Comment by dok-net** [comment]
Further info, performance of back-to-back IRQs, that is, connecting two pins, and let every change on pin A trigger an interrupt on pin B, where the ISR toggles the state of pin A.

Master @ 80MHz: IRQs/s = 136239
This PR @ 80MHz: IRQs/s = 141242

Sketch:
```
#include <atomic>

constexpr long unsigned MAXIRQS = 200000;

std::atomic<long unsigned> irqCount;
std::atomic<uint32_t> lastTime;
uint32_t prevTime;

IRAM_ATTR void triggered(void*)
{
	if (irqCount.load())
	{
		irqCount.store(irqCount.load() - 1);
		digitalWrite(D4, !digitalRead(D4));
	}
	else
	{
		lastTime.store(ESP.getCycleCount());
	}
}

void setup()
{
	Serial.begin(115200);
	while (!Serial);
	delay(100);
	pinMode(D4, OUTPUT);
	digitalWrite(D4, LOW);
	pinMode(D7, INPUT);
	attachInterruptArg(D7, triggered, nullptr, CHANGE);

	Serial.println("measuring IRQ latency");
	irqCount.store(MAXIRQS);
	prevTime = ESP.getCycleCount();
	digitalWrite(D4, !digitalRead(D4));
}

void loop()
{
	delay(1000);
	if (!irqCount.load())
	{
		Serial.print("IRQs/s = ");
		Serial.println(1000 * MAXIRQS / ((lastTime.load() - prevTime) / ESP.getCpuFreqMHz() / 1000));
		irqCount.store(MAXIRQS);
		prevTime = ESP.getCycleCount();
		digitalWrite(D4, !digitalRead(D4));
	}
}
```

dok-net:
For further motivation, here are the results when substituting `Delegate` back in for `std::function`:
[dok-net:all_in_functional_w_delegate](https://github.com/dok-net/arduino-esp8266/tree/all_in_functional_w_delegate)@80MHz: IRQs/s = 144092

mcspr:
> For further motivation, here are the results when substituting Delegate back in for std::function:
dok-net:all_in_functional_w_delegate@80MHz: IRQs/s = 144092

Notable improvement is removal of if(functional), arg casts, which is mighty good thing for the Core's ISR handler speed.

With the PR though, doesn't std function's function still need to be explicitly marked as IRAM ? What about lambdas?
e.g. `attachInterrupt(PIN, [some, captured, args]() { ... }, CHANGE);` in the setup(), which then looks like `std::_Function_handler<void (), setup::{lambda()#1}>::_M_invoke(std::_Any_data const&)` in the objdump

I've tried some alternative approach at https://github.com/esp8266/Arduino/compare/master...mcspr:isr-why
Looking at objdump (fairly weird) output, std::function<T>:: _M_invoke is part of the std function operator() call tree ([and as described here](https://web.mst.edu/~nmjxv3/articles/function_dispatch.html)), so the branch experiments [by moving it to IRAM via ldscript rule](https://github.com/mcspr/esp8266-Arduino/blob/isr-why/tools/sdk/ld/eagle.app.v6.common.ld.h#L145) (but I'm pretty sure it needs to be something else for the sake of schedule_function / anything else using void() std funcs)

dok-net:
All the compiler limitations with regard to attributes still apply.

PR #6048, opened 2019-05-04
Bugfix: attach interrupt (#6049)

dok-net:
Fixes: #6049 

- BUG FIX: __attachInterruptArg is known in the field, but it's use fails due to incorrect delete of arg pointer via cleanupFunctional. This PR guards against this case.
- Original author commented that ScheduledFunctions.(h|cpp) were added by mistake, and should be removed.
- ICACHE_RAM_ATTR was previously applied to attachInterrupt() / detachInterrupt() functions, but reviewer of #6003 commented that this can be reverted to save some precious IRAM, as attaching / detachInterrupt from inside ISR are corner cases.
- Properly expose attachInterruptArg() via Arduino.h

(Reference: PR #6047 implement all-C++ functional based interrupt service routine handling. Caveat: ICACHE_RAM_ATTR attributes maybe don't quite work on functional instances, like lambdas, yet)

### DIFF
--- a/cores/esp8266/FunctionalInterrupt.cpp
+++ b/cores/esp8266/FunctionalInterrupt.cpp
@@ -47,7 +47,7 @@ void attachInterrupt(uint8_t pin, std::function<void(void)> intRoutine, int mode
 	as->interruptInfo = ii;
 	as->functionInfo = fi;
 
-    __attachInterruptFunctionalArg(pin, (voidFuncPtr)interruptFunctional, as, mode, true);
+	__attachInterruptFunctionalArg(pin, (voidFuncPtr)interruptFunctional, as, mode, true);
 }
 
 void attachScheduledInterrupt(uint8_t pin, std::function<void(InterruptInfo)> scheduledIntRoutine, int mode)
@@ -61,5 +61,5 @@ void attachScheduledInterrupt(uint8_t pin, std::function<void(InterruptInfo)> sc
 	as->interruptInfo = ii;
 	as->functionInfo = fi;
 
-    __attachInterruptFunctionalArg(pin, (voidFuncPtr)interruptFunctional, as, mode, true);
+	__attachInterruptFunctionalArg(pin, (voidFuncPtr)interruptFunctional, as, mode, true);
 }
--- a/cores/esp8266/FunctionalInterrupt.cpp
+++ b/cores/esp8266/FunctionalInterrupt.cpp
@@ -7,7 +7,7 @@ typedef void (*voidFuncPtr)(void);
 typedef void (*voidFuncPtrArg)(void*);
 
 // Helper functions for Functional interrupt routines
-extern "C" void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void*fp , int mode);
+extern "C" void __attachInterruptFunctionalArg(uint8_t pin, voidFuncPtr userFunc, void*fp, int mode, bool functional);
 
 
 void ICACHE_RAM_ATTR interruptFunctional(void* arg)
@@ -47,7 +47,7 @@ void attachInterrupt(uint8_t pin, std::function<void(void)> intRoutine, int mode
 	as->interruptInfo = ii;
 	as->functionInfo = fi;
 
-	__attachInterruptArg (pin, (voidFuncPtr)interruptFunctional, as, mode);
+    __attachInterruptFunctionalArg(pin, (voidFuncPtr)interruptFunctional, as, mode, true);
 }
 
 void attachScheduledInterrupt(uint8_t pin, std::function<void(InterruptInfo)> scheduledIntRoutine, int mode)
@@ -61,5 +61,5 @@ void attachScheduledInterrupt(uint8_t pin, std::function<void(InterruptInfo)> sc
 	as->interruptInfo = ii;
 	as->functionInfo = fi;
 
-	__attachInterruptArg (pin, (voidFuncPtr)interruptFunctional, as, mode);
+    __attachInterruptFunctionalArg(pin, (voidFuncPtr)interruptFunctional, as, mode, true);
 }
--- a/cores/esp8266/core_esp8266_wiring_digital.cpp
+++ b/cores/esp8266/core_esp8266_wiring_digital.cpp
@@ -210,13 +210,7 @@ extern void __attachInterruptArg(uint8_t pin, voidFuncPtrArg userFunc, void* arg
     __attachInterruptFunctionalArg(pin, userFunc, arg, mode, false);
 }
 
-extern void __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int mode)
-{
-    __attachInterruptFunctionalArg(pin, (voidFuncPtrArg)userFunc, 0, mode, false);
-}
-
-extern void ICACHE_RAM_ATTR __detachInterrupt(uint8_t pin)
-{
+extern void ICACHE_RAM_ATTR __detachInterrupt(uint8_t pin) {
     if (pin < 16)
     {
         ETS_GPIO_INTR_DISABLE();
@@ -237,6 +231,11 @@ extern void ICACHE_RAM_ATTR __detachInterrupt(uint8_t pin)
             ETS_GPIO_INTR_ENABLE();
         }
     }
+}
+
+extern void __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int mode)
+{
+    __attachInterruptFunctionalArg(pin, (voidFuncPtrArg)userFunc, 0, mode, false);
 }
 
 void initPins() {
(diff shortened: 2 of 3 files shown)


cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -29,220 +29,301 @@
...
+        if (pin < 16)
+        {
+            ETS_GPIO_INTR_DISABLE();
+            GPC(pin) &= ~(0xF << GPCI);//INT mode disabled
+            GPIEC = (1 << pin); //Clear Interrupt for this pin
+            interrupt_reg &= ~(1 << pin);
+            interrupt_handler_t* handler = &interrupt_handlers[pin];
+            handler->mode = 0;
+            handler->fn = 0;
+            if (handler->functional && handler->arg)
```
dok-net: If handler->arg was previously set via __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void *arg, int mode), the missing functional guard would always call cleanupFunctional even on non-ArgStructure objects, which is a bug.

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -29,220 +29,301 @@
...
+            abort();
+        }
+
+        if (pin < 16)
+        {
+            ETS_GPIO_INTR_DISABLE();
+            interrupt_handler_t* handler = &interrupt_handlers[pin];
+            handler->mode = mode;
+            handler->fn = (voidFuncPtr)userFunc;
+            if (handler->functional && handler->arg)  // Clean when new attach without detach
```
dok-net: If handler->arg was previously set via __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void *arg, int mode), the missing functional guard would always call cleanupFunctional even on non-ArgStructure objects, which is a bug.

dok-net:
@d-a-v quote: "shows lots of changes because of tab/space/style updates" - yes, except that it doesn't :-)
Wrt. #5774, if you trust that 57bdbe1a9522c5091020da85290bb54d26257d7f does exactly that and nothing but, the changes are quite easily digestible.

d-a-v:
> the changes are quite easily digestible

Indeed and I took time to digest them.
It is far more easy to check when style is not changed in a PR so the diff is clear to read.

#5774 is not meant to be used on some files but on all files at once (and will be checked by CI from that moment). In the meantime, doing that in a PR is not nice for reviewers.

dok-net:
@d-a-v I am really aware of that and quite sorry, but please trust me that all these smart code editors I use keep changing layouts all the time and having three proposals in the game, it began to seem to make sense to me to use the styling provided for in the repo sources already. The examples even get 
rejected when they are not astyled - that might be for ESP32, though, things are getting confusing :-)
I hope you will find diff-reading #6049 to #6047 or #6038 to be a more pleasant experience for it.

dok-net:
@d-a-v This bug-fixing PR should be readable now (please use diff settings: hide whitespace changes).

d-a-v:
@dok-net 
I understand this PR is the first of your PR serie. I think it can be merged, however I question myself about your removal of ICACHE_RAM_ATTR from `__detachInterrupt`. I think it is legal for an interrupt handler to detach itself no ?

dok-net:
@d-a-v If deleting
struct FunctionInfo
{
    std::function<void(void)> reqFunction = nullptr;
    std::function<void(InterruptInfo)> reqScheduledFunction = nullptr;
};
is guaranteed to not cause flash operations?  std::function<void(InterruptInfo)> looks suspicious to me.
And I haven't even begun complaining about such long-running operations inside world-stopping ISRs :-)

d-a-v:
You are right, `cleanupFunctional` needs to be in iram too, and probably something like #5922 would need to put functional destructors into iram.
@earlephilhower what do you think ?

> And I haven't even begun complaining about such long-running operations inside world-stopping ISRs :-)

Well it is true in such case where a handler stopping itself needs to destroy everything. 

The other way would be to use a Scheduled function to do that job but then it would have to be managed with a bool in the interrupt handler to check wether the handler is in a destroy state. 

I vote for putting iram back in `detachInterrupt` (not all interrupts have args, let's not break them too) and maybe creating a new "defer detachInterrupt in a Scheduled function" issue to address this case.

dok-net:
I think I reconsider my opinion on supporting detachInterrupt() from inside execution ISRs - does this not lead to object destruction on functional etc. occurring while the same are in the middle of executing? Is that well specified?

dok-net:
@earlephilhower So here I am asking you to speed up acceptance of this PR in order for me to update EspSoftwareSerial. It's getting quite tedious switching between my esp8266 branches, especially due to the SoftwareSerial submodule that keeps changing back and forth all the time due to this dependency.
I firmly believe in adding publicly the `attachInterruptArg()`, plus the bugfixes to `detachInterrupt()` that I have implemented. The chances in this PR are all rather pedestrian.

devyte:
@dok-net please open a gitter account and let me know here. I'll send an invite to discuss directly there.

dok-net:
@devyte I have an account there now.

dok-net:
@d-a-v @devyte At the speed of Schedule merges, I'm getting jealous :-)

> I vote for putting iram back in detachInterrupt (not all interrupts have args, let's not break them too) [...]
I've done this now. This one PR really doesn't change all that much, it just enables me to publish the proper next EspSoftwareSerial and we could go forward with discussing more seriously my other PRs after this.

d-a-v:
I think we can merge. @devyte was requesting a chat, I'll let him merge.

dok-net:
@devyte I've a bug report against EspSoftwareSerial that I expect to fix within 24h, which will bring a new release of the package into focus. Is there anything standing in the way of a positive review and merge of this mostly-pure-bugfix PR? The additional exposing of attachInterruptArg() is just alignment with ESP32 and is also due, I'd like to add.

dok-net:
@devyte I was asked by d-a-v to close the lesser of those PRs fixing the same problem. #6047 supersedes this one, but is much more complex, where this here PR is just a few lines. So please understand that I personally would not close it and hope for an iterative approach - rebasing #6047 on this PR is not challenging.

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -125,11 +127,11 @@ typedef struct {
 	void* functionInfo;
 } ArgStructure;
 
-static interrupt_handler_t interrupt_handlers[16];
+static interrupt_handler_t interrupt_handlers[16] = { {0, 0, 0, 0}, };
```
devyte: How does this init list even build? Shouldn't this be either {0} or {{0}, {0},... x16}?

dok-net: Why, the inner type (4 fields) is initialized by {0, 0, 0, 0}, for the remaining array the usual rules apply. Have you seen compiler warnings that I missed?

cores/esp8266/FunctionalInterrupt.cpp
```diff
@@ -7,7 +7,7 @@ typedef void (*voidFuncPtr)(void);
 typedef void (*voidFuncPtrArg)(void*);
 
 // Helper functions for Functional interrupt routines
-extern "C" void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFunc, void*fp , int mode);
+extern "C" void __attachInterruptFunctionalArg(uint8_t pin, voidFuncPtr userFunc, void*fp, int mode, bool functional);
```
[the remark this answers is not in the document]
dok-net: I don't understand - if we are both looking at the same function, this one is defined and used plus exported in core_esp8266_wiring_digital.cpp, and also also used in FunctionalInterrupt.cpp.

Strangely enough, the versions you've commented are not the latest ones, how is that even possible?

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -183,14 +187,15 @@ extern void ICACHE_RAM_ATTR __attachInterruptArg(uint8_t pin, voidFuncPtr userFu
 
   if(pin < 16) {
     ETS_GPIO_INTR_DISABLE();
-    interrupt_handler_t *handler = &interrupt_handlers[pin];
+    interrupt_handler_t* handler = &interrupt_handlers[pin];
```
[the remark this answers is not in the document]
dok-net: Why GitHub makes you comment on code sections that it by itself recognizes and marks as "Outdated" is beyond me ;-) It wasn't hard to find, though, so thanks for pointing that out, if merged the duplicates into a new function.

cores/esp8266/core_esp8266_wiring_digital.cpp
devyte: Extra comma. I guess the compiler is forgiving here as in other cases.

PR #6056, opened 2019-05-06
Honor timeout in HTTPClient

### LINKED ISSUE #6033: ESP8266 hangs when using HttpClient.getString
Dear team, I have an issue with respect to getString() method. We are using ESP8266 where internet connection is little poor. Hence, after sending the HTTP POST request and when trying to read the response from the server, the ESP sometimes hangs and never comes out of it until hard reset is performed manually. This happens randomly. Kindly help me in resolving this. I am attaching the code below:

```
int httpCode = http.POST(postData);  
if (httpCode == 200 && http.getSize() > 0) {
    String payload = http.getString();    //Get the response payload
}
```

d-a-v:
fix #6033 
(bug is reproduced and fixed)

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -1380,6 +1380,11 @@ int HTTPClient::writeToStreamDataBlock(Stream * stream, int size)
 
                 // read data
                 int bytesRead = _client->readBytes(buff, readBytes);
+                if (!bytesRead)
+                {
+                    DEBUG_HTTPCLIENT("[HTTP-Client][writeToStreamDataBlock] input stream timeout\n");
+                    return HTTPC_ERROR_READ_TIMEOUT;
+                }
 
                 // write it to Stream
                 int bytesWrite = stream->write(buff, bytesRead);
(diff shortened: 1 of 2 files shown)


earlephilhower:
I don't see where the write timeout is going to be different between the versions.  The online `diff` seems to have lost its mind, but manually eyeballing I'm not seeing any new checks.  What did I miss?

d-a-v:
>  What did I miss?

It's not that tricky. Timeout is implemented by `Stream::readBytes()` and only there. 
The issue is the same in `HTTPClient::writeToStreamDataBlock()` and the fixed example.

`readByte()`'s timeout was never triggered because of the use of `available()` which ensured that all requested data would be readable.
The PR avoid the use of `available()` and ask for what's really requested. The timeout mechanism then can work.

Check the example, and this more readable [subcommit](https://github.com/esp8266/Arduino/pull/6056/commits/c6d7aa0dbd06eb22d8e20c417a39698a046fd83e).

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1357,93 +1357,85 @@ int HTTPClient::writeToStreamDataBlock(Stream * stream, int size)
...
 
-                // write it to Stream
-                int bytesWrite = stream->write(buff, bytesRead);
-                bytesWritten += bytesWrite;
+        // read data
+        int bytesRead = _client->readBytes(buff, readBytes);
+        if (!bytesRead)
+        {
+            DEBUG_HTTPCLIENT("[HTTP-Client][writeToStreamDataBlock] input stream timeout\n");
+            free(buff);
```
devyte: Consider using above:
```cpp
std::unique_ptr<uint8_t[]> buff(new uint8_t[buff_size]);
```
above instead of the malloc, and removing the three free(buff) calls

PR #6071, opened 2019-05-11
Moved 'Dropping' print from stdout to stderr in drop_versions.py

Subhajitdas298:
Printing 'Dropping' debug strings to stdout leads to write in package json.
This moved that to stderr stream.
Alternatively it can be removed.

### DIFF
--- a/package/drop_versions.py
+++ b/package/drop_versions.py
@@ -20,7 +20,10 @@ def drop_version(todrop, obj):
     for o in obj:
         version = o['version'].encode('ascii')
         if version == todrop:
-            print("Dropping version {0}".format(todrop))
+            # removing the print, as it is writing to stdout, which in turn writes to json, which makes it invalid
+            # thus shifting to error stream
+            # alternatively, it can be removed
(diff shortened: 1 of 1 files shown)


package/drop_versions.py
```diff
@@ -19,7 +19,9 @@ def drop_version(todrop, obj):
     out = [];
     for o in obj:
         version = o['version'].encode('ascii')
-        if version != todrop:
+        if version == todrop:
+            print("Dropping version {0}".format(todrop), file=sys.stderr)
```
earlephilhower: This, unfortunately, does not work in Python 2.7.  Is there some compatibie way for V2 and V3 to make this happen?  The rest of the scripts are python version agnostic, so I'd rather not make this a V3 only.

Subhajitdas298: Was the script version agnostic **before** application of #6070 ?

earlephilhower: You're right, I think I goofed.  It already had that type of print formatting.

PR #6088, opened 2019-05-13
Fix PIO toolchain by ensuring IRQ vectors in link

### LINKED ISSUE #6087: IRQ vectors not being set in PIO toolchain, causing bootloop w/2.5.x releases
I open the issue here since here https://github.com/earlephilhower/esp-quick-toolchain it is not possible.
It seems how `-Wl,option` is done is not the usual way

> See GCC docs https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html
> -Wl,option
> Pass option as an option to the linker. If option contains commas, it is split into multiple options at the commas. You can use this syntax to pass an argument to the option. For example, -Wl,-Map,output.map passes -Map output.map to the linker. When using the GNU linker, you can also get the same effect with -Wl,-Map=output.map

Issue is discussed
https://github.com/platformio/platform-espressif8266/issues/147#issuecomment-491926073

earlephilhower:
PIO uses GCC in a way that causes the compiler to optimize away the IRQ
vectors (since they're not used in our code itself) in the final build
stage.

We can avoid this by forcing all of libmain to be included in the final
object.  This is a small file, and almost completely related to IRQs, so
space impact is minimal.

Fixes #6087 and
https://github.com/platformio/platform-espressif8266/issues/147

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -51,10 +51,10 @@ compiler.c.flags=-c {compiler.warning_flags} -Os -g -Wpointer-arith -Wno-implici
 compiler.S.cmd=xtensa-lx106-elf-gcc
 compiler.S.flags=-c -g -x assembler-with-cpp -MMD -mlongcalls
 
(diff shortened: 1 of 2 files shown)


mcspr:
Sorry, was AFK. Thank you for testing that.

Note that platformio does not use platform.txt to build, but using SCons that runs platformio-build.py script.

Some quick research at https://stackoverflow.com/questions/22122498/how-do-i-add-whole-archive-linker-option-in-scons advices on modifying LINKFLAGS variable and excluding it from LIBS
https://github.com/esp8266/Arduino/blob/98125f88605cd7e46e9be4e1b3ad0600dd5d2b51/tools/platformio-build.py#L123

earlephilhower:
Actually, since I can't test this I'm going to close it.  If someone using PIO can give it a go, please open a new PR w/the needed changes.

earlephilhower:
Also just discovered that the PIO linking way loses other bits of IRAM so this wasn't a complete fix.

PR #6092, opened 2019-05-14
Re-add original SD FAT info access methods

### LINKED ISSUE #6081: In new SD library several functions are now disabled
### Basic Infos

- [ x] If there is a stack dump, I have decoded it.

#### Platform

- Hardware:					ESP-12
- Core Version:			latest git / 2.5.1
- Operating System:		Windows

### Settings in IDE

- Flash Mode:			dio
- Flash Size: 		4MB
- lwip Variant:  v2 Lower Memory
- Reset Method:		ck
- Flash Frequency:	40Mhz
- CPU Frequency:		160MHz
- Upload Using:		SERIAL
- Upload Speed:   115200

As requested in gitter here are the disabled functions:

```
uint8_t type() {
        return 0;//card.type();
    }
    uint8_t fatType() {
        return 0;//volume.fatType();
    }
    size_t blocksPerCluster() {
        return 0;//volume.blocksPerCluster();
    }
    size_t totalClusters() {
        return 0;//volume.clusterCount();
    }
    size_t totalBlocks() {
        return 0;//(totalClusters() / blocksPerCluster());
    }
    size_t clusterSize() {
        return 0;//blocksPerCluster() * blockSize();
    }
    size_t size() {
        return 0;//(clusterSize() * totalClusters());
    }
```

earlephilhower:
Fixes #6081

The SD rewrite blanked out some of the internal FAT info.. Restore the
function calls and return proper values.


libraries/SD/src/SD.h
```diff
@@ -84,35 +84,39 @@ class SDClass {
...
     }
 
     size_t clusterSize() {
-        return 0;//blocksPerCluster() * blockSize();
+        return blocksPerCluster() * blockSize();
     }
 
     size_t size() {
-        return 0;//(clusterSize() * totalClusters());
+        return (clusterSize() * totalClusters());
```
devyte: it's not reasonable to limit the implementation to 4GB with today's available sizes. In fact, I'm having trouble finding new cards smaller than 4GB to buy where I live. 
I suggest just biting the bullet and making this uint64_t.

earlephilhower: Done

earlephilhower:
Actually, can't change size() to return 64b, it breaks printf() and other existing apps.  Added the size64() (which I think we're going to have for FS)...

libraries/SD/src/SD.h
```diff
@@ -84,35 +84,44 @@ class SDClass {
...
 
     size_t clusterSize() {
-        return 0;//blocksPerCluster() * blockSize();
+        return blocksPerCluster() * blockSize();
     }
 
     size_t size() {
-        return 0;//(clusterSize() * totalClusters());
+        uint64_t sz = size64();
+	return (size_t)std::min( (uint64_t)SIZE_MAX, sz );
```
devyte: This is misleading and is likely to be cause for future issues. Instead, I suggest returning 0 when overflowed (i. e. failed)

devyte: and maybe print out a debug message.

earlephilhower: I'd argue the opposite.  Returning 0 is bound to cause trouble (0 means broken card/etc. with this fcn).  Returning 4GB when cards are >4GB are detected at least gives a size that is less likely to cause trouble.  Think of it as a saturating value as used in audio processing.

A warning printout seems good here, though, agreed.

libraries/SD/src/SD.h
```diff
@@ -84,35 +84,49 @@ class SDClass {
...
     size_t clusterSize() {
-        return 0;//blocksPerCluster() * blockSize();
+        return blocksPerCluster() * blockSize();
     }
 
     size_t size() {
-        return 0;//(clusterSize() * totalClusters());
+        uint64_t sz = size64();
+	if (sz > (uint64_t)SIZE_MAX) {
+            Serial.printf_P(PSTR("WARNING: SD card size overflow (>= 4GB).  Please update source to use size64().  Returning 0.\n"));
```
devyte: Should this Serial be hardcoded for output, or a debug mechanism via a macro or whatever?

earlephilhower:
Debug output shown w/a blank 32GB card:
````
SDK:2.2.1(cfd48f3)/Core:2.5.1-17-g9223b622=20501017/lwIP:STABLE-2_1_2_RELEASE/glue:1.1-7-g82abda3/BearSSL:a143020
Initializing SD card...initialization done.
WARNING: SD card size overflow (31998345216>= 4GB).  Please update source to use size64().
total size: 1933574144
done!
````

File: libraries/SD/src/SD.h [hunk 4]
```diff
@@ -84,35 +84,49 @@ class SDClass {
...
+            return 0;
+        } else {
```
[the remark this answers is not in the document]
earlephilhower: Done

libraries/SD/src/SD.h
earlephilhower: Yes, good point.  Done.

PR #6095, opened 2019-05-14
Support PIO Wl-T and Arduino -T linking properly

### LINKED ISSUE #6087: IRQ vectors not being set in PIO toolchain, causing bootloop w/2.5.x releases

earlephilhower:
The interrupt vectors in IRAM are omitted when there is a PROVIDE
statement in the linker control files when using the PIO method of
-Wl,-T<linkfile>.

Drop the PROVIDES (they're in RAM anyway and not ROM related), and
add the required "-u"s to the PIO build script.

Should have no impact on the Arduino side.

Fixes #6087


Jason2866:
YES!! You nailed it :-)
THX

PR #6104, opened 2019-05-16
Bugfix for stuck in write method of WiFiClient and WiFiClientSecure until the remote peer closed connection

sislakd:
Couple of days I was troubleshooting strange behavior with stability of components built on top of WiFiClient and WiFiClientSecure. Finally, I found the root cause of these issues. From time to time it happened that call of write method get stuck until the remote peer closed connection. It seems that root cause bug is present for quite long time in the code. 

When tcp send buffer is full, ClientContext::_write_from_source increments _send_waiting and switch context to NONOS using esp_yield. If something else call esp_schedule (not _write_some_from_cb method in the same instance of ClientContext), the cycle in  _write_from_source is repeated, send buffer is still full and value of _send_waiting is incremented again (thus from this moment _send_waiting>1). Any successful ack on the relevant connection never call esp_schedule because of condition in _write_some_from_cb where _send_waiting is decremented only if it is equal to 1. 

One example when something else can call esp_schedule method is when there are two or more ClientContext instances (e.g. two client connections). Ack on other client context cause esp_schedule and thus resume of write this client context while there is still no space in tcp send buffer.

The simplest solution is set _send_waiting to 1 instead of its increment. As _send_waiting is one Byte it has no sense to change it to bool.

### DIFF
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -455,7 +455,7 @@ protected:
                 break;
             }
 
-            ++_send_waiting;
+            _send_waiting = 1;
             esp_yield();
         } while(true);
         _send_waiting = 0;
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -437,7 +437,7 @@ protected:
     size_t _write_from_source(DataSource* ds)
     {
         assert(_datasource == nullptr);
-        assert(_send_waiting == 0);
+        assert(!_send_waiting);
         _datasource = ds;
         _written = 0;
         _op_start_time = millis();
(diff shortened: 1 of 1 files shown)


earlephilhower:
To @d-a-v to give this a once-over.

sislakd:
I've updated _send_waiting to be clear bool flag.

earlephilhower:
Thanks!  I'll leave it to @d-a-v to double-check that this only needs to be a flag and not a count (in which case the `==1` in the original code should be a `>= 1`)...

TD-er:
Well, this looks like a very sneaky bug indeed and just by looking at the code, I already imagine a lot of related reported issues.
Will make a test build for a few of my nodes immediately to test it.

@d-a-v if you're aware of more of these 'suspecting' parts of the code, please add some issues about them so they can be looked into.
This change does have the potential to fix a lot of hanging issues.

PR #6107, opened 2019-05-17
Remove deadlock when server is not acking our data

sislakd:
If a server where WiFiClient or WiFiClientSecure is connected stop acking our outgoing tcp packets (network issue, server issue), write attempt can stuck forever (if nothing else is generating esp_schedule). 

The PR propose to change esp_yield in ClientContext::_write_from_source to delay which can be interrupted by esp_schedule or after given timeout expires. In the next cycle _is_timeout() is matched and write attempt ends.

For WiFiClientSecure it took 2x 15 seconds to return from WiFiClientSecure::_write method if the remote peer is not acking out outgoing tcp packets and send buffer is already full. There is _run_until called from flush and the second _run_until called in the loop after which it get out of _write method because of -1. It seems that it would be good to have default timeout consistent with WiFiClient. Thus, PR proposes to reduce timeout from 15 seconds to 5 seconds in WiFiClientSecure after we have successful handshake with the remote peer.


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -456,7 +456,7 @@ class ClientContext
             }
 
             _send_waiting = true;
-            esp_yield();
+            delay(_timeout_ms);
```
d-a-v: This change may decrease bandwidth:
an ACK (= permission to write) could reach in before expiration of this delay.
I'll do some tests to show the impact.

sislakd: I expected to delay function to be returned sooner if esp_schedule() is called. The same patter is used in ClientContext::connect method where there is this comment:

I've tested and it seems that it works, esp_schedule seems to schedule cont task to continue immediately not waiting for full timer expiry.

d-a-v: You seem to be right.
Was this WiFiClient change mandatory and the WiFiClientSecure change alone not sufficient according to you tests ?

earlephilhower: Is it always safe to call `delay()` from here?  Always guaranteed to be the right context?

The other thing is I don't think `delay(XX)` can legally return before the timer.  In all the OSes I've seen, `sleep` or `delay` guarantees a minimum time.  Otherwise, for example, if you were doing a SW PWM or something and a packet came in you'd have a short period.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -1009,6 +1009,9 @@ bool WiFiClientSecure::_connectSSL(const char* hostName) {
   _x509_insecure = nullptr;
   _x509_knownkey = nullptr;
 
+  // reduce timeout after successful handshake to fail fast if server stop accepting our data for whathever reason
+  if (ret) _timeout = 5000;
```
earlephilhower: Seems reasonable here to do this, good catch.

But, given that the object can be reused, it needs to be set at the start of the multiple ::connects and ::connectServers, or on the second pass it will be stuck at the shorter time.

sislakd: I've pushed additional change. All places where TSL context is invalidated calls method _freeSSL. Thus, I've added reset of timeout to 15 seconds there.

libraries/ESP8266WiFi/src/include/ClientContext.h
sislakd: The function delay calls internally esp_yield which check right context using cont_can_yield condition. Thus, I think that we can call delay from any context. ESP8266 implementation seems to allow wakeup application context delay using esp_schedule. 

Currently delay doesn't guarantee that there is wait at full interval. If something use delay it is waken any time there is ack of packet or any other signal on esp_schedule. If you want proper timing, e.g. for generation precise output signal you should use approach like in Software serial where there is a loop based on ESP.getCycleCount():
#define WAIT { while (ESP.getCycleCount()-start < wait); wait += m_bitTime; }

The other option could be to put here delay(1) ms which will perform repeated looping until some packet is accepted and send buffer contains some free space or until connection timeout expires.

sislakd: @d-a-v Without this change esp_yield block execution until there is esp_schedule generated somewhere. In the case of unresponsive server, there are no acks and thus no esp_schedule in application having single WiFiClient connection. With multiple WiFiClient connections, write can be unblocked by ack on the other one but it has to come after that long timeout. In most cases acks will come in max few hundreds of ms (with delayed ACK enabled) and then nothing else will wake the currently blocked write after timeout expires. Other connections will not generate new data because (causing further acks) because application context is blocked here.

earlephilhower:
@sislakd, I've tested it and can say that delay(xxx) is not interrupted on a packet and delay() needs to be undone.

````
#include <ESP8266WiFi.h>
WiFiServer server(80);
void setup() {
Serial.begin(115200);
WiFi.begin(SSID,PASS);
  while (WiFi.status() != WL_CONNECTED) delay(100);
Serial.print("\n\nconnected, ip =");
  server.begin();
}

// the loop function runs over and over again forever
void loop() {
  digitalWrite(LED_BUILTIN, LOW);   // Turn the LED on (Note that LOW is the voltage level
  // but actually the LED is on; this is because
  // it is active low on the ESP-01)
  delay(10000);                      // Wait for a second
  digitalWrite(LED_BUILTIN, HIGH);  // Turn the LED off by making the voltage HIGH
  delay(10000);                      // Wait for two seconds (to demonstrate the active low LED)
}
````

Run the script and you'll see a 10s long high and low period.  While it's going, do a NC/telnet/etc. to the 8266 port 80.  If delay() was interrupted on a packet, the 10S period would be shortened when the connection came in.  It is not, it still is 10s until the LED flips.

So, please undo the delay(timeout) change.

**Comment by sislakd** [comment]
@earlephilhower Have you modified WiFiServer::_accept method to include esp_schedule() in your test? delay call is interrupted only if something call esp_schedule(). The current version of WiFiServer doesnt include esp_schedule(). Thus, you cannot see that behavior.

**Comment by sislakd** [comment]
Revert of delay in _write_from_source method to esp_yield doesn't solve problem of deadlock when there is no ack coming. What will work is just yield() or optimistic_yield causing quick cycling in _write_from_source method allowing timeout to return from the method if the remote peer is not responding.

**Comment by earlephilhower** [comment]
@sislakd, I think maybe I wasn't clear.  I was demonstrating that when you put in a `delay(timeoutms)` in the path, that `delay` will not return before the timeout, even if another packet or ack or whatever comes in.  This is against the current head 968d6fcb0a9a2eab657178c95ee4302d6f470e46.

That's the current behavior and people do depend on it.  delay(100) is around 100ms, and not sometimes 1ms.  SWSerial, others, etc.

I'm not sure what you meant about changing WiFiClient::accept, but if you're talking about something which will globally cause `delay()` to now finish on an event on the TCP stack (and not time >delay) that's breaking things badly for the rest of the ecosystem.

So, what I'm saying is that delay(timeout) can only worsen TCP response since it will go idle no matter what even if an ack or whatever comes in, and that changing delay()'s semantics so that it sometimes returns as a random variable (since the TCP stack is async and anyone can send at any time) isn't going to work.

**Comment by sislakd** [comment]
It seems that you don't understand what I'm trying to explain. I will try to summarize my thinking in details and hopefully it will be much clearer than before.

- The key statement is that delay(ms) function called from application context returns earlier than defined ms interval if and only if function esp_schedule() is called from core context while application context is within delay function during that time.
- The application context is code executed in setup() and loop() functions defined by application. On the other hand core context is everything else (code executed by core, interrupt handlers, system timers, ...).
- Here is excerpt from core_esp8266_wiring.cpp and core_esp8266_main.cpp related to delay, esp_yield and esp_schedule functions:
```
void delay_end(void* arg) {
    (void) arg;
    esp_schedule();
}

void delay(unsigned long ms) {
    if(ms) {
        os_timer_setfn(&delay_timer, (os_timer_func_t*) &delay_end, 0);
        os_timer_arm(&delay_timer, ms, ONCE);
    } else {
        esp_schedule();
    }
    esp_yield();
    if(ms) {
        os_timer_disarm(&delay_timer);
    }
}

extern "C" void esp_yield() {
    if (cont_can_yield(g_pcont)) {
        cont_yield(g_pcont);
    }
}

extern "C" void esp_schedule() {
    ets_post(LOOP_TASK_PRIORITY, 0, 0);
}
```
- The function delay(ms) called with non-zero ms internally configures os_timer (once expired configured delay_end function it is called in core context) and call esp_yield(). Within esp_yield() function it is checked that the function is called from application context (using cont_can_yield) and call cont_yield(...) functions. If you check definition of cont_yield(...) in cont.S, it implements context switch by storing register values, prepare return path back to application context, change stack pointer back to core context, and jump to cont_ret position in cont.S where registers from core context are reloaded and hit ret instruction returning back to core context code execution (the next line after cont_run(...) call within loop_task(...) function). Thus, during call of esp_yield() within delay(ms) function, application context code execution is blocked.
- The way how the execution context code execution is resumed again is given by the call of esp_schedule() function. During this function there is scheduled the next execution of loop_task(...) from core context when possible. Once (some time later), the core context executes function loop_task(...) it calls cont_run(...) function which implements context switch back to application context. The context switch is similar to the one described in the previous bullet but in other direction (from core to application).
- The timed delay is implemented using os_timer where the function delay_end(...) is called after timer expires. The function delay_end just call esp_schedule() function described in the previous bullet.
- At this stage it should be clear that delay(ms) can return due to either expired os_timer (which calls internally esp_schedule()) or call of esp_schedule() from different place. Thus, if there is no esp_schedule() function call anywhere else delay(ms) implements timed sleep returning close to defined time interval. However, if you search Arduino code for usage of esp_schedule() function there are many places and some of them are called while code is executed in core context (these are those which can truncate delay and ends it before given interval elapses).
- Now, it is crucial to know when such situation happens. Just existence of possible esp_schedule() calls doesn't mean that they are always executed. There are many conditions when this can happen.

Now back to your simple test application:
- Within your loop() function you call delay(ms) and turn ON and OFF LED to see premature return from delay(ms). That's fine.
- You expect to have delay(ms) call be interrupted by WiFiServer upon new connection attempt made to the listening tcp.
- I'm saying that in your test application there is nothing (really nothing) calling esp_schedule() in the core context except delay_end(...) called from os_timer. Thus, running this application is not affected by whatever number of connection attempts to WiFiServer in your application.
- If you would like to really see the interrupted return from delay(ms) in your test application, you have to modify the core code to make esp_schedule() call from core context upon external event (e.g. your connection attempt to WiFiServer). The simplest way to achieve this is to add call of esp_schedule() function in the beginning of WiFiServer::_accept  method. This method is called from lwIP as registered connection accept event handler. I'm not going to describe internals of lwIP to describe that this event handler is called from tcp_input (which is called from WiFi frame processing via ip_input processing). The key is that WiFiServer::_accept method is called from core context and thus execution of your delay(ms) will be interrupted upon your connection attempt to WiFiServer then.

Now back to this PR:
- I'm proposing to use delay(ms) function (instead of esp_yield()) within loop in ClientContext::_write_from_source(...) method. 
- You think that it will worsen TCP response time because incoming ack will not interrupt delay(ms) until given timeout elapses.
- Thats not true. If you don't see that from my long explanation above, you can take it from different perspective. The current esp_yield() in ClientContext::_write_from_source(...) method is interrupted by call esp_schedule in ClientContext::_acked(...) method. And under good connection and other peer condition this works. Thus, it must work also using delay(ms) function instead of esp_yield(). As described above delay(ms) uses esp_yield() anyway.

Why this PR is proposed then:
- If network or remote peer is not working well and our going tcp packets are not acked, the tcp send queue gets filled until there is no more space to accommodate data from ClientContext and ClientContext:: _write_from_source(...) method sets _send_waiting to true and call esp_yield() function. This blocks application context execution. With application not using any other libraries calling esp_yield() from the core context, there is no place where esp_yield() is resuming blocked execution of application context. Thus, the code is stuck in esp_yield() forever in ClientContext:: _write_from_source(...). There are no more loops and thus it is not possible to timeout at all.
- Even if you have multiple WiFiClient and/or WiFiClientSecure objects, no other can call esp_yield(). There is only one application context which is stuck in esp_yield() in one ClientContext instance. Only this one instance has field _send_waiting set to true. All other have it false. Thus, even there is ClientContext::_acked(...) called in different ClientContext instance (because there is ACK on that connection), no esp_schedule() is called (given by the condition in ClientContext::_write_some_from_cb()) as there is not fulfilled condition having  _send_waiting==true.
- Thus, proposed usage of delay(ms) instead of esp_yield() in ClientContext:: _write_from_source(...) method will unblock either upon freeing some space in tcp send buffer (because of esp_schedule() in ClientContext::_acked(...)) or after elapsing given timeout.
- Timed based end of delay(ms) is crucial to allow loop continue, detect timeout and return back to WiFiClient or WiFiClientSecure to handle further this case.

Let me know whether this explanation makes sense to you and you understand why this PR is important to fix ClientContext implementation.

Sorry for the long comment, but I don't know how much details from other code you know in details.

earlephilhower:
@sislakd, thanks for your detailed explanation.  The LWIP and packet interface is one place I've thankfully not had to dive into, and I think I understand where you're going here now.

I was talking w /@devyte about this as well and I can see that the pattern of "delay(timeout)" with a "esp_schedule()" to short-circuit it is used in the ::connect method.  Your last bit is clear and seems fine now that I get it.
> Thus, proposed usage of delay(ms) instead of esp_yield() in ClientContext:: _write_from_source(...) method will unblock either upon freeing some space in tcp send buffer (because of esp_schedule() in ClientContext::_acked(...)) or after elapsing given timeout.

I appreciate your recent PRs, but I do hope you understand that I'm only trying to be extremely careful with such a critical bit of the code and not trying to discourage you!

earlephilhower:
To @d-a-v for final signoff.

d-a-v:
@sislakd 

> Let me know whether this explanation makes sense to you and you understand why this PR is important to fix ClientContext implementation.

> Sorry for the long comment, but I don't know how much details from other code you know in details.

After @earlephilhower 's question, your explanation is crystal clear.
I would have liked some comments around the delay() change stating something like "will be interrupted by esp_schedule on next received ack".

You are mastering the nonos model of this core, also network management internals.
Thanks for this second PR.

As a side note, I'm a bit concerned about how _timeout is managed. It is defined by Stream:: (1000 by default), possibly modified by user and overwritten during SSL handshake (to 15000 with good reasons) and now per this PR restored to 5000. I'll leave @earlephilhower think of it since this is very specific to BearSSL client.

earlephilhower:
`_timeout` is much longer because SSL handshake is very slow.  AXTls started it and BearSSL just took it over.

sislakd:
I've added comment to introduced delay to be understandable while reading the code.

PR #6117, opened 2019-05-19
Improve timeout in Updater.

### LINKED ISSUE #6113: Improvement to timeout in Updater.cpp
Recently, when doing a deep dive into behavior in slow network environments, I can across the code in Updater.cpp that handles a timeout while an update is being streamed. Effectively, it will wait for 100ms after a timeout and then try once more. If this second attempt fails, the update fails. I would like to propose a change to this mechanism that makes the download succeed in even the slowest networks. 

This change uses a call to millis() to track the start time, reseting this start time every time data is received. After 60 seconds of not receiving more data, it will timeout.

The new function is below with the new variable called "timeout":

```
size_t UpdaterClass::writeStream(Stream &data) {
    size_t written = 0;
    size_t toRead = 0;
    if(hasError() || !isRunning())
        return 0;

    if(!_verifyHeader(data.peek())) {
#ifdef DEBUG_UPDATER
        printError(DEBUG_UPDATER);
#endif
        _reset();
        return 0;
    }
    unsigned long timeout = millis();
    if (_progress_callback) {
        _progress_callback(0, _size);
    }
    if(_ledPin != -1) {
        pinMode(_ledPin, OUTPUT);
    }

    while(remaining()) {
        if(_ledPin != -1) {
            digitalWrite(_ledPin, _ledOn); // Switch LED on
        }
        size_t bytesToRead = _bufferSize - _bufferLen;
        if(bytesToRead > remaining()) {
            bytesToRead = remaining();
        }
        toRead = data.readBytes(_buffer + _bufferLen,  bytesToRead);
        if(toRead == 0) { //Timeout
          if (millis() - timeout > 60000) {
            _currentAddress = (_startAddress + _size);
            _setError(UPDATE_ERROR_STREAM);
            _reset();
            return written;
          }
          delay(100);
        } else {
          timeout = millis();
        }
        if(_ledPin != -1) {
            digitalWrite(_ledPin, !_ledOn); // Switch LED off
        }
        _bufferLen += toRead;
        if((_bufferLen == remaining() || _bufferLen == _bufferSize) && !_writeBuffer())
            return written;
        written += toRead;
        if(_progress_callback) {
            _progress_callback(progress(), _size);
        }
        yield();
    }
    if(_progress_callback) {
        _progress_callback(progress(), _size);
    }
    return written;
}
```
Also, I read another thread recently with comments thanking you all for the incredible work you do on this project. One response was it is not common to get thanks, so here is another. THANK YOU!!

PurpleAir:
I added an optional timeout parameter in the writeStream function of the Updater. This makes the update more reliable (it finishes) on slow networks.

*edit*
fixes #6113


cores/esp8266/Updater.cpp
```diff
@@ -463,14 +464,15 @@ size_t UpdaterClass::writeStream(Stream &data) {
...
         if(toRead == 0) { //Timeout
-            delay(100);
-            toRead = data.readBytes(_buffer + _bufferLen, bytesToRead);
-            if(toRead == 0) { //Timeout
-                _currentAddress = (_startAddress + _size);
-                _setError(UPDATE_ERROR_STREAM);
-                _reset();
-                return written;
-            }
+          if (millis() - timeout > streamTimeout) {
```
TD-er: Is this valid when `millis()` just overflows between calls?

devyte: As I said, pkease reimplement the timeout in terms of the polledTimeout class.

adrionics: I do believe it should be fine during rollovers. https://arduino.stackexchange.com/questions/12587/how-can-i-handle-the-millis-rollover
Is that right?

adrionics:
Can anyone help with suggesting how to use the polledTimeout class properly?

d-a-v:
in [examples](https://github.com/esp8266/Arduino/blob/master/libraries/esp8266/examples/BlinkPolledTimeout/BlinkPolledTimeout.ino)

adrionics:
Thanks, I will update the pull request when I get the time.

adrionics:
As requested, polledTimeout implemented.

d-a-v:
@devyte It is true saying `Stream::readBytes()` already implements a timeout, opposed to non-arduino `StreamInheritance::read(buf,len)` which immediately returns when no more data is available (implemented in HardwareSerial and WiFiClient at least).

There are several locations in the core or our libraries where  a retry is done before giving up. All of them do the same thing: move data from a Stream instance to a Print insntance (or local buffer). Fixing that would need to be done all at one by using a proper generic API or function template that allows transferring a Stream instance to a Print instance. I may have a something with no intermediate buffer but it is still an alpha state and I will propose something at some point.

@adrionics you may try with `data.setTimeout(streamTimeout)` or even better set `data`'s timeout before starting the updater without changing `writeStream()` signature, *and* remove the retry.

devyte:
> set data's timeout before starting the updater

that is exactly what I was thinking.

adrionics:
"data" is a WiFiClient and it is using the default timeout. I found that by setting a timeout on data, it seems to make things less reliable. I tried setting a timeout before connecting to the host, but that makes the client wait too long to request a retry (I think?) and so if anything, you need a different timeout before connecting to the host and after connecting(?). Through experimenting, I found it works better to not mess with the WiFiClient's setTimeout() and just wait long enough for it to get data. I any way, the way it was done before was much worse. I found it never is able to complete the update on some devices with low signal strength. 

I am of course willing to follow your advice and retry the experiment, so if you have specifics on the setTimeout as related to WiFiClient, I will try again.

d-a-v:
@adrionics 

Default timeout is 1000ms in Stream. 
I don't understand why you would need a shorter timeout for other operations than in updater. Can you elaborate "I found that by setting a timeout on data, it seems to make things less reliable" if you set a timeout bigger ?

If really a bigger timeout is needed in Updater \<\<this update is really needed give it best luck>>, what about adding a `Stream::getTimeout()`, saving it in `::writeStream()`'s beginning then setting it to twice its value (more? a big constant?), and restoring it in the end of `::writeStream()` (still removing the retry) ?

adrionics:
There are two timeouts. A connect timeout and a stream timeout. setTimeout seems to change both though? I would not want a different timeout for other things. I was setting a timeout of 8 seconds - and tried 12 for the WiFiClient before connecting (it's default is 5). I then tried setting timeout on the stream of around 2 or 3 seconds.

Correct me if I am wrong though: The stream timeout is a timeout that after it expires, the stream will retry so if you have too long timeout set on that, it will wait for nothing and not help getting the data. That is why, if the stream timeout is lower, and the read is retried more times it works. Would that be right?

d-a-v:
>  setTimeout seems to change both though? 

Yes, ClientContext uses WiFiClient/Stream timeout in `::connect()`. And yes WiFiClient overrides timeout to 5000.

> The stream timeout is a timeout that after it expires, the stream will retry 

Stream itself does not retry. If the delay between two separate receives is greater than timeout, it returns.
If timeout was longer, then there'd be no need to externally retry (like it does now).
That's why I propose to increase timeout in `::writeStream()` (and restore it after), *or* increase your timeout once for all (onto `data`) to something big enough for the updater Stream/client.

devyte:
@adrionics What is the status here? did you test based on @d-a-v 's comments?

adrionics:
> @adrionics What is the status here? did you test based on @d-a-v 's comments?

Sorry, I did not have time to spend on this and used the modifications as posted already with excellent performance or reliability. I can revisit in the future, but it can be hard to test or verify on thousands of devices as has already been done now. I think it is better than it has ever been in terms of successful updates across the base with the changes described.

devyte:
@d-a-v does this PR still make sense in view of your Stream:send implementation? Or would a different Stream-ing approach be possible/better?

d-a-v (approved):
The `stream::send` approach should indeed be considered but that could be in another PR.
This PR apparently solves some updater API issues.
I wonder if @PurpleAir is still using it.
(merge conflict is fixed)

PurpleAir:
> The `stream::send` approach should indeed be considered but that could be in another PR.
> This PR apparently solves some updater API issues.
> I wonder if @PurpleAir is still using it.
> (merge conflict is fixed)

Thank you, yes we are.

devyte (approved):

PR #6124, opened 2019-05-20
Split carriage return's character from result

mbm60:


devyte:
Previous comment is correct, doing this in Stream would break things that later rely on looking for the cr. 
Closing

PR #6129, opened 2019-05-22
The use of bind in Ticker.h is prone to type inference failure

dok-net:
During development of a different Schedule.h queue, it turns out that the use of std::bind has nasty issues with ambiguous type inference. The use of lambdas is generally encouraged in current C++, so this PR refactors the two spots where std::bind is used.


libraries/ESP8266mDNS/src/LEAmDNS.cpp
```diff
@@ -1053,14 +1053,14 @@ const char* MDNSResponder::answerTxts(const MDNSResponder::hMDNSServiceQuery p_h
...
+    m_HostProbeInformation.m_fnHostProbeResultCallback = p_fnCallback;
     
     return true;
 }
 
 bool MDNSResponder::setHostProbeResultCallback(MDNSHostProbeFn1 pfn) {
-	using namespace std::placeholders;
-	return setHostProbeResultCallback(std::bind(pfn, std::ref(*this), _1, _2));
+    using namespace std::placeholders;
+    return setHostProbeResultCallback([obj=std::ref(*this), pfn](const char* _1, bool _2) { pfn(obj, _1, _2); });
```
devyte: What is obj? It's not clear at a glance. Prefer a more descriptive name, e.g. MDNSref or whatever.

libraries/ESP8266mDNS/src/LEAmDNS.cpp
```diff
@@ -1088,8 +1088,8 @@ bool MDNSResponder::setServiceProbeResultCallback(const MDNSResponder::hMDNSServ
 
 bool MDNSResponder::setServiceProbeResultCallback(const MDNSResponder::hMDNSService p_hService,
                                                   MDNSResponder::MDNSServiceProbeFn1 p_fnCallback) {
-	using namespace std::placeholders;
-	return setServiceProbeResultCallback(p_hService, std::bind(p_fnCallback, std::ref(*this), _1, _2, _3));
+    using namespace std::placeholders;
+    return setServiceProbeResultCallback(p_hService, [obj=std::ref(*this), p_fnCallback](const char* _1, const hMDNSService _2, bool _3) { p_fnCallback(obj, _1, _2, _3); });
```
devyte: What is obj? Prefer a more descriptive name.

dok-net:
The error message when std::bind is used looks like this:

```
Ticker.cpp:33: In file included from
Ticker.h: In member function void Ticker::attach_scheduled(float, Ticker::callback_function_t)
 
Ticker.h: 46:55: error: no matching function for call to 'bind(<unresolved overloaded function type>, Ticker::callback_function_t&)
   attach(seconds,std*: bind(schedule_function, callback))
Ticker.h:46: note  candidates are
 
Ticker.h:28: In file included from
Ticker.cpp:33: from
functional:1655: note  template<class _Func, class ... _BoundArgs> typename std  _Bind_helper<std  __or_<std  is_integral<typename std  decay<_Tp>  type>, std  is_enum<typename std  decay<_Tp>  type> >  value, _Func, _BoundArgs ...>  type std  bind(_Func&&, _BoundArgs&& ...)
   bind(_Func&& __f, _BoundArgs&&... __args)
functional:1655: note    template argument deduction\substitution failed

```

Too bad it doesn't actually list the candidates, I deduce that what's giving the compiler hiccups here is the choice of rValue reference and const reference signatures of schedule_function in Schedule.h, which goes like this:

```
bool schedule_function(std::function<void(void)>&& fn);
bool schedule_function(const std::function<void(void)>& fn);

```

libraries/ESP8266mDNS/src/LEAmDNS.cpp
dok-net: Changing according to signature
```
typedef std::function<void(MDNSResponder& resp,
								   const char* p_pcServiceName,
								   const hMDNSService p_hMDNSService,
								   bool p_bProbeResult)> MDNSServiceProbeFn1;
```

libraries/ESP8266mDNS/src/LEAmDNS.cpp
dok-net: Compared to bind it's no less clear... *smile* Changing according to signature
```
typedef std::function<void(MDNSResponder& resp,
								   const char* p_pcDomainName,
								   bool p_bProbeResult)> MDNSHostProbeFn1
```

PR #6139, opened 2019-05-24
Refactor linked list to lock-free ring buffer

dok-net:
This amends #6039 to use a circular queue that is lock-free between producer/consumer and is tested on ESP8266, ESP32 and x86_64.


d-a-v:
Do you have other use-cases for this ring buffer ?

dok-net:
@d-a-v The queue is used in EspSoftwareSerial.

dok-net:
@d-a-v @devyte I've prepared a test sketch that stresses the queue implementation of the Scheduler.
Over 100000 scheduled tasks, in chunks of 30 [20] per ticker, the circular_queue-based scheduler requires just over 3.3s [5s] real time.
For the same workload, the current linked-list finishes after more than 43s [10s] real time.

In terms of quality, the circular queue suffers a performance degredation (24s real time) when the chunk size is increased to 50, this is expected behavior due to the many failed and repeated attempts to insert into the well exhausted queue of size 32.
Even at 35 schedules-per-ticker, though, the linked-list fails disastrously (Soft WDT reset).

Please check for yourself and comment if I've egregiously misses something:

```
// QueueTest.cpp : This file contains the 'main' function. Program execution begins and ends there.
//

#include <Ticker.h>
#include "PolledTimeout.h"

constexpr long unsigned MSGCNT = 100000;
constexpr long unsigned MSGSPERTICKER = 30;

int pending = 1;

long unsigned msgNo = 0;
long unsigned nextVal[] = { 0 };

uint32_t startMillis;

bool checkMessage(long unsigned msg)
{
	const auto id = msg & 0x1f;
	const auto val = msg >> 5;

	if (id < 0 || id >= 1) {
		String str = "received bad thread id = ";
		str += int(id);
		Serial.println(str);
		return true;
	}
	if (val == MSGCNT - 1) {
		--pending;
		return false;
	}

	if (val != nextVal[id])
	{
		if (val < nextVal[id]) {
			String str = "value in reverse, id = ";
			str += int(id);
			str += ", val = ";
			str += int(val);
			Serial.println(str);
			nextVal[id] = val + 1;
		}
		else {
			String str = "values skipped, id = ";
			str += int(id);
			str += ", lost = ";
			str += int(val - nextVal[id]);
			Serial.println(str);
			nextVal[id] = val + 1;
		}
	}
	else
	{
		nextVal[id] = val + 1;
	}
	return true;
}

Ticker ticker;

ICACHE_RAM_ATTR void tickerCallback(int)
{
	for (int i = 0; i < MSGSPERTICKER; ++i) {
		if (msgNo >= MSGCNT) return;
		auto msg = (msgNo << 5) + 0;

		if (schedule_function([msg]()
			{
				if (!checkMessage(msg)) { ticker.detach(); }
			}))
		{
			++msgNo;
		}
	}
}

void setup()
{
	Serial.begin(115200);
	delay(1000);
	Serial.println("scheduler test");
	startMillis = millis();

	if (schedule_function_us([]() { Serial.print("scheduler"); /*delay(2000);*/ Serial.println(" alive"); yield(); return true; }, 5000000))
	{
		Serial.println("scheduled 'scheduler alive' function");
	}
	else
	{
		Serial.println("could not schedule 'scheduler alive' function");
	}

	schedule_function_us([]() { Serial.print(".."); return true; }, 500000);

	ticker.attach_ms<int>(1, tickerCallback, 0);
}

esp8266::polledTimeout::periodicFastUs loopAlive(5000000);

void loop()
{
	if (pending == 0)
	{
		const auto val = nextVal[0];
		if (val != MSGCNT - 1)
		{
			String str = "received bad last message, id = ";
			str += 0;
			str += ", val = ";
			str += int(val);
			Serial.println(str);
		}

		Serial.print("done after ");
		Serial.print(millis() - startMillis);
		Serial.println("ms");

		pending = -1;
	}

	if (loopAlive) Serial.println("loop alive");
}

// x86_64:
// linked-list: 129340ms CPU usage,  57s wall clock
// linked-list:  93610ms CPU usage,  39s wall clock
// linked-list: 214984ms CPU usage, 100s wall clock
// circular q:   66317ms CPU usage,  30s wall clock
// circular q:   67909ms CPU usage,  32s wall clock
// circular q:   55780ms CPU usage,  25s wall clock
// linked-list: 193866ms CPU usage,  90s wall clock
// circular q:   70780ms CPU usage,  32s wall clock

// ESP8266:
// linked-list:  43904ms wall clock
// circular q:    3354ms wall clock

```

d-a-v:
Thanks for your work and your proofs/tests.
I will try it after the exposed API #6158 is approved by everyone.
I think you will need to add some comments in the class.
Since it is supposed to be used in this core and also in espsoftwareserial used by esp32 too, did you consider to put it in a separate submoduled repository so everybody can benefit from bugfixes from a single location ?

dok-net:
@d-a-v git submodules... but have a point here, certainly.
I've merged and fixed your notinyield branch with the circular queue, if you're interested this is in https://github.com/dok-net/Arduino/tree/notinyield

dok-net:
@d-a-v I've implemented the Ticker-based timing here - this eliminates the use of PolledTimeout except for the 100ms period yielding.
- Are you dependent on microsecond resolution? The Ticker by default is in millisecond resolution mode and IIRC works at 5ms minimum delay.
- The repeat value in this implementation is not continues, but always begins counting at the time the function was scheduled and returns to the scheduler - this adds to the imprecision...
- Could you again give this implementation a run in your environment, please (if the above listed constraints don't forbid it straight off)? Porting the Ticker use to the linked-list version of Schedule is a no-brainer, I believe, for comparison purposes.

dok-net:
> @dok-net
> How is ticker going to call a function in cont stack while an external library is busy waiting on network ?

The ESP8266 already has convenience functions that call `schedule_function()` - thus letting the user code run in CONT, not SYS. I've adapted the `schedule_function_us()` in Schedule.(h|cpp) to use Ticker, but in such fashion that eventually the execution of the scheduled code happens when `run_scheduled_functions()` kicks in - in CONT. Please check my questions above.
As a working example, here's my test code adapted to your `yield()` loop:

```
#include <atomic>
#include <Schedule.h>

void setup()
{
	Serial.begin(115200);
	delay(1000);
	Serial.println("scheduler test");

	if (schedule_function_us([]() { Serial.print(".."); return true; }, 500000))
	{
		Serial.println("scheduled 'alive'");
	}
	else
	{
		Serial.println("could not schedule 'alive'");
	}
}

void loop()
{
	uint32_t deadline = millis() + 10000;
	Serial.println("external library now busy");
	while (deadline - millis() > 0) yield();
	Serial.println("external library finally gets out");
}

```

dok-net:
@d-a-v You might like the latest commit https://github.com/esp8266/Arduino/pull/6139/commits/5a8584af9c6a752c56b4ff95a5a340f1e6a6d3dc : now the PolledTimeout is used for delays/periods up to 5000ms, where polling is quite sane, I guess, whereas for longer delays above the 5000ms minimum of Ticker, the latter is used, alleviating the scheduler from polling tasks that will not be runnable for quite a few cycles.

dok-net:
@d-a-v With `SCHEDULED_FN_MAX_COUNT` == 32, the Ticker heap now consumes an additional 1280 bytes RAM. Would you think a lower number of slots for Ticker objects can do the job? These Tickers are used only for time-scheduled functions at 5000µs period or longer...

dok-net:
Note: feature branches rebase against master, never in ongoing work merge master. Master eventually merges feature branches, does not rebase on feature branches (why should it).
Experience: Learn about git rerere before starting the first rebase. Merging master into a feature branch brings master's change history in that branch, makes the feature branch unmanageable, and almost always reverts master to some mixed old state when the feature branch is accepted and merged.

libraries/Schedule/src/circular_queue/circular_queue.h
```diff
@@ -0,0 +1,438 @@
...
+#define ICACHE_RAM_ATTR
+#define IRAM_ATTR
+#endif
+
+/*!
+    @brief	Instance class for a single-producer, single-consumer circular queue / ring buffer (FIFO).
+            This implementation is lock-free between producer and consumer for the available(), peek(),
+            pop(), and push() type functions.
+*/
+template< typename T > class circular_queue
```
devyte: This file is very difficult to understand. I was trying to look at the class interface, and there is no way to tell at a glance how it's meant to be used. 
I suspect the monolithic nature of the file is convenience for porting, but here we need maintainability. 
I suggest:
1. splitting the single consumer and multiconsumer classes into their own specific files
2. Splitting the template class method implementations into separate .h (usually blah_impl.h naming convention) 
3. including the above headers into this circular_queue.h header

devyte: Also, please put the template <> test on its own line rather than precursor to class blah

dok-net:
Rebase - done.
Refactoring circular_queue - tbd next.

libraries/Schedule/src/circular_queue/circular_queue.h
dok-net: Thank you so much for your input, splitting up the two classes into their own source files is rather obvious, and I forgot doing it. It's a clear maintenance improvement.
For the other requests, I have a strong opinion on those. The template<> prefix occurs once in each file, and makes it rather more expressive in my reading - I would not want to change that, given that the resulting line lengths are not an issue, either.
Second, I am generally opposed to creating a plethora of single source files, which I find extremely hard to deal with in any editor, when there is no overriding reason to do so. While a separate interface would definitely be helpful, I see so reasonably easy way of doing so in non-virtual C++ classes - for lack of the interface keyword. If there's a pattern to do so that I missed, please help me out. Introducing the usual pure-virtual abstract base class in code for an MCU in my mind is out of the question.
So, while removing the inline-definitions of the member function into another source file (*_impl.h), I would find code editing and comprehension a pain. YMMV (respestively does vary), no hard feeling, but could you find a way to agree to keeping this to my preference?

At the end of the day, I was hoping to have the "Schedule" code an own library that can be plugged, you've seen #6182 prepare for that.

devyte: What are you talking about with virtualness, there is no virtualness here.

The current circular_queue.h is 281 lines. Splitting it will increase it and probably bump it over 300. As a rule of thumb, it's better to have multiple smaller files than one big one. 
For a normal class, the usual implementation is to put the class declaration in the .h and the method definitions in the .cpp. Of course this isn't possible for a template class, so the next best thing is to put the template class declaration in one .h, and the template method definitions in another *_impl.h which then gets #included at the end of the definition one.
Despite the length, I can live with one .h for both declaration and definition here, but the minimum is to split the class declaration from the definition to make the class interface readable at all.
About template and class on different lines, we (maintainers) have put forth large effort to clean up style and layout throughout the core, and maintain consistency thereafter. The request is merely to change to this:
```cpp
template <typename arg>
class circular_queue
{
...
};
```
Please comply with that.

libraries/Schedule/src/circular_queue/circular_queue.h
```diff
@@ -0,0 +1,281 @@
...
+
+        m_outPos.store((outPos + size) % m_bufSize, std::memory_order_release);
+        return size;
+    }
+
+    /*!
+        @brief	Iterate over and remove each available element from queue,
+                calling back fun with an rvalue reference of every single element.
+    */
+    void for_each(std::function<void(T&&)> fun)
```
devyte: Instead of polluting the class with an iteration method, how about defining an iterator class and making it compatible with STL algorithms? That would decouple the iteration logic from the class itself, which is at it should be.

devyte: E.g.: add begin(), end(), iterator class, const_iterator class, etc. Then the user could do e.g.: `for(auto t : myq) {fun(t);}` or whatever. I would really prefer such compatibility, I was even thinking why not implement a circular queue in terms of an existing STL container, but I'm ok with a custom implementation if it's compatible.

libraries/Schedule/src/circular_queue/circular_queue_mp.h
```diff
@@ -0,0 +1,184 @@
+/*
```
devyte: Same comments as for sp class.

libraries/Schedule/src/circular_queue/circular_queue.h
dok-net: I agree, so I have moved the definitions of each member function that has more than 4-5 lines of code outside the class definition. I hope you like this, too.

libraries/Schedule/src/circular_queue/circular_queue.h
dok-net: Uh oh - I am not sure that this can really be the same logic as the `for_each()`, more specifically the `for_each_requeue()`. There might a way to do this in STL style, but that looks like a major piece of overhaul, and research. As I am actively thinking about doing more special stuff with interaction and lambda IOC, I'd rather no be bound by compatibility constraints (just now). There are some performance problems that the linked list doesn't suffer, I would rather try and invent something to get closer in performance to that, next.

libraries/Schedule/src/circular_queue/circular_queue_mp.h
dok-net: Same opinion on my side as above. Maybe later?

dok-net:
@devyte Could you please help me with this CI problem? I don't understand a thing :-)

dok-net:
Preparation for this library is now in other, simpler PRs. This library is not strictly needed, as the functionality exists in Schedule. Portability to ESP32 makes it preferable to take this to a standalone library. Closing PR. Thanks for the review!

PR #6154, opened 2019-05-27
Add FS::info64 call for fielsystems > 4GB

### LINKED ISSUE #6082: wrapper SDFS.info(FSInfo& info)  is overflowed for SD Cards > 4GB
### Basic Infos

- [ x] If there is a stack dump, I have decoded it.

#### Platform

- Hardware:					ESP-12
- Core Version:			latest git / 2.5.1
- Operating System:		Windows

### Settings in IDE

- Flash Mode:			dio
- Flash Size: 		4MB
- lwip Variant:  v2 Lower Memory
- Reset Method:		ck
- Flash Frequency:	40Mhz
- CPU Frequency:		160MHz
- Upload Using:		SERIAL
- Upload Speed:   115200

Because storage is using `size_t` instead of `uint64_t` so size over 4GB are incorrects
`size_t` is ok for SPIFFS, which is the original usage for `info(FSInfo& info)`, but now if used on SD in must be `uint64_t` for SD, but this may lead to some others compatibilities issues if used on SPIFFS, because lot of functions  currently use `size_t`  and several others does not support `uint64_t`.

FYR On ESP32 the struct  usage originaly done to add additional parameters in transparent way was dropped to use independant functions instead, each one using proper type :
SPIFFS:   
```
class SPIFFSFS : public FS
{
...
    size_t totalBytes();
    size_t usedBytes();
```

For SDFS:   
```
class SDFS : 
....
sdcard_type_t cardType();
uint64_t cardSize();
uint64_t totalBytes();
uint64_t usedBytes();
```

Sorry I do not have solution to offer because this issue need a mitigation between compatibility and features.
But I hope it give clear base description to help for a decision.

earlephilhower:
Fixes #6082

Add an info64() call which returns used and total sizes as 64 bit
quantities.  A default wrapper that just copies the 32-bit values is
included for LittleFS/SPIFFS which can't hit those capacities.


libraries/SDFS/src/SDFS.h
```diff
@@ -113,8 +113,22 @@ class SDFSImpl : public FSImpl
...
+    bool info(FSInfo& info) override {
+        FSInfo64 i;
+        if (!info64(i)) {
+            return false;
+        }
+        info.blockSize     = i.blockSize;
+        info.pageSize      = i.pageSize;
+        info.maxOpenFiles  = i.maxOpenFiles;
+        info.maxPathLength = i.maxPathLength;
+        info.totalBytes    = (size_t)i.totalBytes;
```
devyte: I would assert here if the uint64 value can't be contained in 32bits, with a medsage that the sd card capacity is larger than supported. 
Alternatively, at least print out some debug msg or something if the is encountered.

cores/esp8266/FSImpl.h
```diff
@@ -75,6 +76,20 @@ class FSImpl {
     virtual void end() = 0;
     virtual bool format() = 0;
     virtual bool info(FSInfo& info) = 0;
+    // For seamless backward compatibilty, supply default which just copies form the 32bit version
+    virtual bool info64(FSInfo64& info64) {
```
devyte: This isn't a good idea. If this ever gets called, it will silently do the wrong thing for capacity > 4GB and result in a bug. 
I suggest a pure virtual, or asserting.

libraries/SDFS/src/SDFS.h
d-a-v: assert would break backward compatibility (the bug was already there)
What about a print (in no debug mode) and return max(32bits) ?

cores/esp8266/FSImpl.h
earlephilhower: The alternative here is to replicate the exact functionality in both SPIFFS and LittleFS.  I think it might be better to have the one copy rather than the two.  All other FSes will implement the 64-bit (SD, FTP-FS, SAMBA, etc.) version unless they only support <4GB which seems unlikely.

earlephilhower: Yeah, fair enough.  A wayward cast could cause odd behavior, and who knows what's living in Arduino-land.  Done.

libraries/SDFS/src/SDFS.h
earlephilhower: Done in latest push.

earlephilhower:
While it may be possible to do a multiGB LittleFS or SPIFFS, I think they'd both fall over and die right now because of the assumptions they're baked into things.  Like defining sizes in terms or int32 everywhere.  If some decoupling occurs (it could today, just shim the flash_write to do SPI writes to an external chip) we can definitely revisit this, but my guess is they'd need to rewrite LFS/SPIFFS on-chip format in an incompatible way so would end up making a "new" FS anyway...

PR #6155, opened 2019-05-27
Make SSO support \0s, use memmove, add test

earlephilhower:
Supercedes #6027

Make SSO more generic by keeping track of its length explicitly,
allowing for embedded \0s to exist in the String (just like the non-SSO
ones).

Use memmove/memcpy_P when we know the length of a string to save CPU
time.

Add tests to inject \0s in a String to ensure it is still working as
designed.


cores/esp8266/WString.h
```diff
@@ -250,27 +250,29 @@ class String {
...
-
-        // SSO is handled by checking the last byte of sso_buff.
-        // When not in SSO mode, that byte is set to 0xff, while when in SSO mode it is always 0x00 (so it can serve as the string terminator as well as a flag)
-        // This allows strings up up to 12 (11 + \0 termination) without any extra space.
-        enum { SSOSIZE = sizeof(struct _ptr) + 4 }; // Characters to allocate space for SSO, must be 12 or more
+        // This allows strings up up to 11 (10 + \0 termination) without any extra space.
+        enum { SSOSIZE = sizeof(struct _ptr) + 4 - 1 }; // Characters to allocate space for SSO, must be 12 or more
+        struct _sso {
+            char     buff[SSOSIZE];
+            unsigned len  : 7;
```
devyte: Shouldn't this be unsigned char?

cores/esp8266/WString.h
```diff
@@ -250,27 +250,29 @@ class String {
...
+            unsigned flag : 1;
+        };
```
devyte: Prefer a more descriptive name instead of "flag", e. g. "isSSO" or "caseSSO"

cores/esp8266/WString.h
```diff
@@ -250,27 +250,29 @@ class String {
...
         enum { CAPACITY_MAX = 65535 }; // If size of capacity changed, be sure to update this enum
```
devyte: A bit of a nitpick, but consider defining this value in terms of the type of _ptr. cap,. That way if cap ever changes this changes automatically.
Alternatively, please update the comment that the max means the max value that can be held in cap, which is if type uint16_t.
Why? it took me a bit to re-understand the reasons behind this.

cores/esp8266/WString.h
```diff
@@ -250,27 +250,29 @@ class String {
...
-        inline bool sso() const { return sso_buf[SSOSIZE - 1] == 0; }
-        inline unsigned int len() const { return sso() ? strlen(sso_buf) : ptr.len; }
-        inline unsigned int capacity() const { return sso() ? SSOSIZE - 1 : ptr.cap; }
-        inline void setSSO(bool sso) { sso_buf[SSOSIZE - 1] = sso ? 0x00 : 0xff; }
-        inline void setLen(int len) { if (!sso()) ptr.len = len; }
-        inline void setCapacity(int cap) { if (!sso()) ptr.cap = cap; }
-	inline void setBuffer(char *buff) { if (!sso()) ptr.buff = buff; }
+        inline bool isSSO() const { return sso.flag; }
+        inline unsigned int len() const { return isSSO() ? sso.len : ptr.len; }
+        inline unsigned int capacity() const { return isSSO() ? (unsigned int)SSOSIZE - 1 : ptr.cap; }
```
devyte: Please add comment that capacity means number of chars not including the null term.

Questions:
1. if we no longer care about zeroes in the string, shouldn't we treat the null term as just another byte? Or is that already the case, and here it's a matter of definition for capacity? 
2. Does ptr.cap mean the same, i. e. does NOT include the null term? Or is the - 1 missing in that case?

cores/esp8266/WString.h
earlephilhower: Talking about len or buff?  `buff` should be `char` as that's the non-SSO type.  `len` is a bitfield so `unsigned` or `signed` will do the trick.

cores/esp8266/WString.h
earlephilhower: It seemed straightforward to me since the only way to get to this var is `sso.flag`.  I'll make it into `sso.isSSO`...

cores/esp8266/WString.h
earlephilhower: "size of capacity" is odd.  Didn't seem that way when I wrote it before.  Will reword.

cores/esp8266/WString.h
earlephilhower: Comment added, but it's always been kind of funky this way (old Arduino code)
1) Nul term still added at end so things like `::c_str()` can work properly.  It's just that internally it's just like a Pascal string with `\0` being as valid a character as `a`
2) capacity() is the space available for chars not including the required ending NUL, yes.  On SSO, that's 1 less than the buffer, on non-SSO that's set to 1 less than the malloc'd size.

cores/esp8266/WString.h
devyte: I mean unsigned char (1 byte)  as opposed to just unsigned (4 bytes)

Issue #6143, opened 2019-05-26
BearSSL::WiFiClientSecure::connect() throws Exception


braiden:
### Basic Infos

- [X ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ X] I have tested that the issue is present in current master branch (aka latest git).
- [ X] I have searched the issue tracker for a similar issue.
- [ X] If there is a stack dump, I have decoded it.
- [ X] I have filled out all fields below.

#### Platform

- Hardware:				ESP8266
- Core Version:			2d9253e46c41ef6949759ec028a356917d289831
- Development Env:		Platformio
- Operating System:		Debian Testing

### Settings in IDE

- Module:  			Wemos D1 mini
- Flash Mode:			dout
- Flash Size: 		4MB
- lwip Variant:        -DLWIP_OPEN_SRC -DLWIP_FEATURES=1 -DLWIP_IPV6=0 
- Reset Method:		nodemcu
- Flash Frequency:	80Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   1M

### Problem Description

BearSSL::WiFiClientSecure::connect() throws Exception while connecting. The issue happens about 3/10. So the sketch below fails in a few seconds. Both a hostname of "mqtt.2030.ltsapis.goog" and a call to configTime() are required or the bug doesn't happen (or happens much much less often.)

Also, an exception is not always thrown, sometimes connect() completes successfully but memory elsewhere not owned by wifi client is corrupted. I could provide another sketch demonstrating this, but thought the exception more useful.

The sketch omits X509List, same issue happens with valid certs.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <ESP8266WiFi.h>
#include <WiFiClientSecure.h>
#include <time.h>

WiFiClientSecure *client;

void setup() {
  Serial.begin(115200);

  WiFi.mode(WIFI_STA);
  WiFi.hostname("test");
  WiFi.begin("redacted", "redacted");
  while (WiFi.status() != WL_CONNECTED) delay(10);
  Serial.println("WiFi Connected");

  configTime(0, 0, "pool.ntp.org");
  while (time(nullptr) < 1514764800) delay(10);
  Serial.println("NTP Time Synced");

[...]

PR #6156, opened 2019-05-27
Add basic canary check to BSSL stack thunk

earlephilhower:
On return from a BSSL call, check that the last element of the stack is
still untouched.  If it is modified, print an error and abort().

Will catch problems like #6143 many times with an informative error
message instead of corrupting the heap and having a random crash
sometime later.

### DIFF
--- a/cores/esp8266/StackThunk.cpp
+++ b/cores/esp8266/StackThunk.cpp
@@ -36,7 +36,7 @@ uint32_t *stack_thunk_top = NULL;
 uint32_t *stack_thunk_save = NULL;  /* Saved A1 while in BearSSL */
 uint32_t stack_thunk_refcnt = 0;
 
-#define _stackSize (5750/4)
+#define _stackSize (5748/4)
 #define _stackPaint 0xdeadbeef
 
 /* Add a reference, and allocate the stack if necessary */
@@ -122,6 +122,13 @@ void stack_thunk_dump_stack()
     pos += 4;
   }
   ets_printf("<<<stack<<<\n");
+}
+
+/* Called when the stack overflow is detected by a thunk.  Main memory is corrupted at this point.  Do not return. */
+void stack_thunk_fatal_overflow()
+{
+    ets_printf("FATAL ERROR: BSSL stack overflow\n");
+    abort();
 }
 
 };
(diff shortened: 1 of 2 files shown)


dok-net:
Dear @earlephilhower!
I am building a cooperative scheduler [https://github.com/dok-net/CoopTask/], and stack space is an issue there, too. I've found this StackThunk implementation, and I'd like to refactor and use it. I've got two questions:

- On the ESP8266, does the stack grow downward as usual? I think StackThunk has stack_top and stack_bottom backward, stack_bottom should be the base, have the highest address, and stack_top should be the lower address, in terms of malloc the address that's returned by that. I guess the naming is consistent right now, but it would be entirely confusing if I'm right, would you agree?

cont_stack_start and cont_stack_end on the other hand denominate the reserve memory area, for that, the naming is OK.

- Could the StackThunk implementation be refactored into a class, such that it becomes re-usable for other libraries that need their own additional stack space? I imagine creating a singleton instance of it in the BearSSL *.cpp would be easy to do, and fixing core_esp8266_postmortem.cpp to deal with that can't be an unsurmountable task either.

If you have no objections and would be willing to merge such a change "immediately", I might give it a try.

earlephilhower:
For a scheduler, I'd not use the current thunk code.  It's really special purpose and has assumptions about parameter passing (i.e. nothing overflows the parameter regs so no stack copying either way, no preservation of processor state, etc.).

dok-net:
I've found a pure C++ way of giving the tasks each their own stack on the heap, which works on ESP8266 and ESP32. I will test this on an 8bit Arduino, too.
I am surprised that nobody has built cooperative multitasking on the ESP8266, using the cont.h/.S etc. functionality, which seems to go almost all the way toward that goal already. I am in a discussion with the ESP32 maintainer regarding cooperative multitasking on that SOC, where he is convinced that the FreeRTOS tasks should be used instead. Yes, I am willing to believe that RTOS task scheduling is faster that that of, say, Windows or Linux, but given the complexity of concurrent programming, and all that work Microsoft have done for their TPM async/await programming is not without reason, I am confident that our small devices stand to take good advantage of cooperative multitasking. Take delays and infinite loops, they look really much simpler than if you try to express that in the single-loop() model, while there is none of the atomicity and synchronization problems that befall preemptive MT.

Possibly, even some of the current driver work, for which the Schedule class was revamped to supported recurrent and timed functions, could use such a scheduler for much more expressive code?

PR #6162, opened 2019-05-28
Move all SSL constants into PROGMEM, free 1.2KB

earlephilhower:
Rewrite all the integer math operations with const input parameters to
use PROGMEM properly (pgm_read_xx or memcpy_P), and move all the EC
order and generators and SHA OIDs to PROGMEM.

This frees around 1.2KB of heap for any SSL applications with no
perceptible speed impact.

Also delete unneeded objects from the bearssl.a library to shrink the
GIT repo size.

**This runs everything I've tried without incident, but needs community testing to validate I've covered all the constant use cases**


earlephilhower:
@mcspr @sislakd, you should give this a try if you're still doing SSL development.

mcspr:
No problems so far using HTTPS updater.
More heap as advertised too.

earlephilhower:
@TD-er, did you give this PR a test with your app?  Anything to report?

TD-er:
@earlephilhower Nope, haven't tested it, since we currently have no SSL code in use.
It was simply taking too much resources, but this may help.

So I cannot say it will break previous implementations, since we simply have none (yet).

PR #6165, opened 2019-05-30
Added BR_OPT_NO_RENEGOTIATION flag to forbid TLS renegociation

s-hadinger:
Since PR #6065 x509 memory structure are dropped after succesful handshake. This is ok as long as there is no TLS renegociation, otherwise it will crash.

Adding this flag to avoid any TLS renego, as described here: https://bearssl.org/x509.html

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
@@ -827,6 +827,7 @@ extern "C" {
     uint16_t suites[cipher_cnt];
     memcpy_P(suites, cipher_list, cipher_cnt * sizeof(cipher_list[0]));
     br_ssl_client_zero(cc);
+    br_ssl_engine_add_flags(&cc->eng, BR_OPT_NO_RENEGOTIATION);  // forbid SSL renegociation, as we free the Private Key after handshake
     br_ssl_engine_set_versions(&cc->eng, BR_TLS10, BR_TLS12);
     br_ssl_engine_set_suites(&cc->eng, suites, (sizeof suites) / (sizeof suites[0]));
     br_ssl_client_set_default_rsapub(cc);


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -827,6 +827,7 @@ extern "C" {
     uint16_t suites[cipher_cnt];
     memcpy_P(suites, cipher_list, cipher_cnt * sizeof(cipher_list[0]));
     br_ssl_client_zero(cc);
+    br_ssl_engine_add_flags(&cc->eng, BR_OPT_NO_RENEGOTIATION);  // forbid SSL renegociation, as we free the Private Key after handshake
```
earlephilhower: That's a good catch. I'm not quite sure whether it's the right solution or if we should undo the dropping of X509 after connection.

After looking at the RFC I see this could cause two problem in two ways (and dropping X509 would need to be undone):
1. When more 2^64 messages are sent and the sequence num needs to overflow.  This one we can probably ignore.
2. Servers are free to reject connections with this option sent in the HELLO.  Unfortunately I don't know how prevalent this would be.

I'm inclined to say no. 2 happens very infrequently and go with this fix.  If we get feedback otherwise we can undo this and the original drop-x509-after-handshake patch.

s-hadinger:
I think it's safe to reject renegociation. This is what Bearssl mentions:

> As described above, the BR_OPT_NO_RENEGOTIATION flag can be used to prohibit renegotiations. Note that TLS-1.3 bans renegotiations, and when HTTP/2 is used with TLS-1.2, renegotiations are not permitted either except at the very beginning of the connection.

Btw, thanks a lot for all the great work you did on TLS and BearSSL.

This is off-topic, I have been working on a trimmed down version for Tasmota and connection to AWS IoT. The goal was to reduce even more the memory and code footprint. 

AWS IoT requires client certificate. I reduced to a single cipher: `BR_TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`, which means a lot less code (SH256 only, AES GCM/CTR only, ECDHE/RSA only, P256 EC curve only, max RSA key 2048 and max EC key 256, EC private key). I'm currently consuming only 6.0k, and an additional 6.1k of heap during handshake. I'm not sure how this could be reverted to a more general version.

earlephilhower:
@s-hadinger, If AWS IoT connections are well defined, we can see about adding a cipher option in the menu (where now it's basic and all) for this. The main work is done, you could add the menu in the boards.py and the defines can be used in WiFiClientSecureBSSL.c to trim down even further.

PR #6173, opened 2019-06-01
Update CaptivePortal.ino

MisterD66:
Android devices have a feature that opens a captive portal right after connecting to an accesspoint. to trigger this feature the target IP Adress cant be in the 192.168.x.x range.

After testing this on three devices and all of them work with the new ip adress and not the old i propose to change it

Tested on:
Xiamoi Redmi Note 4 (Android 7)
Samsung Galaxy A7  (Android 8)
Amazon Fire 7 (Android 5)

### DIFF
--- a/libraries/DNSServer/examples/CaptivePortal/CaptivePortal.ino
+++ b/libraries/DNSServer/examples/CaptivePortal/CaptivePortal.ino
@@ -3,7 +3,7 @@
 #include <ESP8266WebServer.h>
 
 const byte DNS_PORT = 53;
-IPAddress apIP(192, 168, 1, 1);
+IPAddress apIP(172, 217, 28, 1);
 DNSServer dnsServer;
 ESP8266WebServer webServer(80);
 


d-a-v:
Why don't you use official [private network](https://en.wikipedia.org/wiki/Private_network) address range instead of 172.217.28/24 which is public (belonging to google at the moment) ?

MisterD66:
because android only automatically opens the captive portal promt if the google adress it is checking to see if internet connection is availabel is not direkted to a private ip adress.

i did not discover this but i confiremd it with all devices available to me.

i found this problem here:
https://github.com/espressif/arduino-esp32/issues/1037#issuecomment-383264214

d-a-v:
@MisterD66 thanks for the useful link.
Please watch duplicate #5529.

d-a-v:
closed by #5529

Misiu:
@d-a-v https://github.com/esp8266/Arduino/pull/5529 is merged,  so this PR can be closed.

Issue #6152, opened 2019-05-27
HTTPClient HTTP 1.1 persistence issue


jpboucher:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-8266-wroom2]
- Core Version:			[latest git hash]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Size: 		[2MB (1MB SPIFFS)]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description
HTTPClient's Persistence support does not seem to follow the RFC 2068 standard for HTTP 1.1 but seems to follow HTTP 1.0 instead.

The condition to prevent disconnection is inside the HTTPClient::disconnect() method itself:
https://github.com/esp8266/Arduino/blob/a4ae8564f7efc51f93521f3088eb025c34e69f09/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L431
Because the disconnect method is called at the end of HTTPClient::writeToStream() and because writeToStream() is called by HTTPClient:getString(), you can only keep persistence if you set the **__reuse_** and **__canReuse_** flags.
The only way to set the _reuse flag is by calling HTTPClient::setReuse() and the only way to set the _canReuse flag is by receiving "Connection: Keep-Alive". (See HTTPClient::handleHeaderResponse()) Here lies the issue.

In HTTP 1.1, both the client and server are to assume that the connection will be persistent unless specified otherwise. The client needing the server to send "Connection: Keep-Alive", to keep the connection alive, is a behavior only found in HTTP 1.0 when persistence was not the default. 

This means that a HTTP 1.1 server would not respond with a Keep-Alive if it was informed that the client is in HTTP 1.1.

[...]

Issue #6149, opened 2019-05-27
http.GET fails every other call


pgooch:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] I have filled out all fields below.

#### Platform

- Hardware:				[other]
- Core Version:			[latest git hash or date]
- Development Env:		[Arduino IDE]
- Operating System:		[MacOS]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[dior]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

When I try and call against the DD-WRT router firmware with an http authorization username/password it will give me a -2 "send header failed" every other call. This does not happen when it is called from other devices. It also is not related to the time between calls, as it is always every other be it once per second or once per half second.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <Arduino.h>

v/* All the imports and setup for the wifi */
#include <Arduino.h>
#include <ESP8266WiFi.h>
// Imports
#include <ESP8266WiFiMulti.h>
#include <ESP8266HTTPClient.h>
#include <WiFiClient.h>
ESP8266WiFiMulti WiFiMulti;
WiFiClient client;
HTTPClient http;

// Settings 
const int time_between_updates = 500; // time between bandwidth checks in ms

// Variables we will be using
char last_router_return = 'Wait for it...';
int last_update = -1001;

void setup() {
  // Start the serial at 115200 baurd
  Serial.begin(115200);

  // Connect to the wifi
  WiFi.mode(WIFI_STA);
  WiFiMulti.addAP("SSID", "PASSWORD");
}

void loop() {
  int now = millis();
  if(now>=last_update+time_between_updates){
    last_update = now;
    get_bandwidth_usage();
  }else{
    Serial.print(millis());
    Serial.print(" ");
    Serial.println(last_update+time_between_updates);
    delay(100);
  }
}

[...]

PR #6176, opened 2019-06-02
Bugfix/esp8266 http client

Jeroen88:
Fix HTTP1.1 persistence issue found by @jpboucher solving issue #6152. With HTTP1.1 default behavior now is to reuse the connection, which is compliant with specs.


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1266,6 +1268,7 @@ int HTTPClient::handleHeaderResponse()
 
             if(headerLine.startsWith("HTTP/1.")) {
                 _returnCode = headerLine.substring(9, headerLine.indexOf(' ', 9)).toInt();
+                _canReuse = (_returnCode != '0');
```
earlephilhower: You're checking the return code, not the HTTP version to set canReuse.  Shouldn't you just be looking at header[7] and if it's != 0 you can reuse?

jpboucher: It would indeed feel less "hacky" and easier to understand.

jpboucher:
With this code _canReuse will be set to false in HTTP 1.0. Which is fine as it will only be set to true if keep-alive is received in the Connection token in HTTP 1.0. 

I have more of a problem with the _reuse flag being set to true in the header file as it will be applied for both HTTP 1.1 and 1.0. 

To fix that we would probably need to set _reuse to false when useHTTP10() is called and _useHTTP10 is set to true. And set it back to true if _useHTTP10 is set to false.

Jeroen88:
@earlephilhower you were right about the header parsing. This is corrected in this commit.

@jpboucher the connection is only kept alive if _reuse and _canReuse both are true, At the beginning of HTTPClient::handleHeaderResponse() _canReuse is initialized to !_useHTTP10, and later the response header is checked for 'Connection: Keep-Alive', so this is right

jpboucher:
If you look at HTTPClient::sendHeader(const char * type) you will see that the _reuse flag is checked for sending keep-alive. If _reuse is true by default you would be sending Keep-alive in HTTP 1.0 even if you never called setReuse(). You are informing the server that you are reusing the socket even if you are not. This will cause issues.

Jeroen88:
@jpboucher I think you are right, I will check this later today and correct it

jpboucher:
Which line?

Jeroen88:
@jpboucher In ::sendHeader()

jpboucher:
@Jeroen88 No, this would not work. You want to send Keep-alive in HTTP 1.0 but only if setReuse(true) was called. To keep it simple, the default value of _reuse should depend on the HTTP version being used. Which is why I think the best course of action is to change it's state inside useHTTP10(). It's alright that the default value in the header file is True since by default we are in HTTP1.1.

Jeroen88:
@jpboucher thnx for your insights, I followed your suggestion :)

Jeroen88:
> MklittleFS is still included in the PR. That's actually a .gitignore bug and I'll put a PR in to fix it.
> 
> Please double-check that header parsing, I think you're going to set canreuse to true even on HTTP 1.0 connections.

@earlephilhower I do not know how to remove it from the PR, can you do that for me?

earlephilhower:
@Jerone88, I can't edit your repo where the PR actually lives, but it should not be hard for you.  I think `git rm` on it will remove it, then you can repush.  You might need to redo `get.py` if git wants toactually delete the dir.

earlephilhower:
@Jeroen88 Sorry for name typo before.   I can't edit your repo where the PR actually lives, but it should not be hard for you. I think git rm on it will remove it, then you can repush. You might need to redo get.py if git wants to actually delete the dir.

Jeroen88:
@earlephilhower I can't get it done. I pull from the remote master to local master. Git says: everything is up-to-date. I push to my remote branch and then I checkout my bugfix/ESP8266HTTPClient branch. Even if I do not delete the extra file, when I push the branch to remote, I get:

`
 ! [rejected]          bugfix/ESP8266HTTPClient -> bugfix/ESP8266HTTPClient (fetch first)
error: failed to push some refs to 'https://github.com/Jeroen88/Arduino.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
`

Any suggestions?

jpboucher:
At first glance it seems that your remote bugfix branch evolved independently from your local branch somehow. You should try out git kraken (or any other git GUI) to seek differences between your remote and local branch. Have you tried pulling from you remote bugfix branch before pushing?

jpboucher:
Another nitpick that would probably help the API is to add a lock on the _reuse variable when setReuse() is called. I don't think it's mandatory but I think it would be weird for users to call setReuse() only for the flag to be unexpectedly changed anyways when useHTTP10() is called.

Jeroen88:
@jpboucher thnx, with git kraken it was possible to solve the branch conflict :)

Jeroen88:
@jpboucher Could you suggest some lines of code?

Jeroen88:
@jpboucher could you suggest some code for the reuse lock you suggested? Than maybe we can finish this PR.

jpboucher:
I will do so as soon as I have time. It's nothing complicated though. Anything that will make it so that the _reuse variable does not change anymore when changing HTTP mode when the user decides to call setReuse() will do the job.

jpboucher:
I think we should consider the pull request finished. Changing from HTTP 1.0 to HTTP 1.1 or vice-versa run-time sounds like a very rare use-case after all. 

If someone does end up doing that they are going to have to call setReuse() every time they change the HTTP version if they don't want to follow the default behavior. I think this is acceptable.

Only feedback will tell wether or not this is problematic.

Jeroen88:
@devyte @earlephilhower as far as I am concerned this could be merged :)

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1265,6 +1268,7 @@ int HTTPClient::handleHeaderResponse()
             DEBUG_HTTPCLIENT("[HTTP-Client][handleHeaderResponse] RX: '%s'\n", headerLine.c_str());
 
             if(headerLine.startsWith("HTTP/1.")) {
+                _canReuse = (headerLine[sizeof "HTTP/1."] != '0');
```
d-a-v: (sizeof "HTTP/1.") == 8, is it what is expected ?

Jeroen88: (sizeof "HTTP/1.") must be **7**, although I did not check it. I can add the if (_canReuse) but I am not shure what should be done if HTTP/1.0 is requested and HTTP/1.1 is offered.

d-a-v: It is 8 (last '\0' is always part of a `"string"`)
If HTTP/1.0 is requested, I think we should just comply and set reuse to false.

Jeroen88: @d-a-v You are right, I update the PR next week

Jeroen88:
@d-a-v @jpboucher I finally got time to correct the PR. Everything okay now?

Jeroen88:
@d-a-v @jpboucher No still not good. If I POST data to my server, HTTP/1.1, with setReuse(true), I get every other call an HTTPC_ERROR_SEND_HEADER_FAILED error, because ::connected() is false.

When I set ::reuse(false) verything works as expected

Any ideas?

Jeroen88:
@d-a-v and @jpboucher I think it does work with setReuse(true). Could you test it?

d-a-v:
@Jeoren88 @jpboucher Is there a MCVE or a simple way to reproduce the issue ?

Jeroen88:
@d-a-v @jpboucher I have no real issue at the moment, but I suspect the HTTPClient behavior is not correct. Let me explain:

My use case is repeatedly POSTing to an InfluxDB. My version of InfluxDB sends a response header 'Connection: Close'. Data is repeatedly received by influx, however: the response header is imho **not handled** by the HTTPClient! My working use case is just luck!

HTTP/1.1 without request and without response headers 'Connection: ' works fine now, I think. But for both HTTP/1.0 and HTTP/1.1 with request and/or response headers there are a few cases:
- request Keep-Alive, response Keep-Alive or no response, I think this works
- request Keep-Alive, response Close, I think this is not handled. I think ::disconnect() should be called in or after ::handleHeaderResponse
- request Close, response Close or no response, I think this is also not handled and ::disconnect() should also be called
- request Close, response Keep-Alive, I think, if this can occur, this would be a protocol error. I think this should be logged. The connection should be closed?

If the above is true, ::connect() [here](https://github.com/esp8266/Arduino/blob/76cda9bdbccd00c02d69119fea9dee34f05cbd9a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L1141) should also check && (!_reuse || !_canReuse)

Other than the name of the function suggests, ::disconnect() does not always disconnect. If _reuse && _canReuse the connection is not closed. This is correct behavior, but in combination with the function name this might be misleading (like the missing check in ::connect() mentioned above).

However, this is pure guessing of what would be logical. I do not have enough protocol knowledge to be sure, nor do I have test cases for all situations (HTTP/1.0, HTTP/1.1, with or without request and/or response headers, with request 'Close' and response 'Keep-Alive'.

So my questions are:
- Do you agree with the above changes? I could adapt the PR although it might take some time because I am busy
- I have no clue yet how to test it. This seems a risk to me. Any ideas?

d-a-v:
@Jeroen88 Would you like to have a look to #6332 (if that is related) ?

Jeroen88:
@d-a-v might very well be related. I will have a look later

earlephilhower:
Yeah, we've got 3 or 4 open issues where HTTP connections fail every other attempt due to the bug @Jeroen88's trying to fix here.  You'll close a lot of bugs when this gets going. :)

Jeroen88:
@dmdt, @d-a-v, @earlephilhower, @pgooch, latest commit solves issues #6332, #6149 and possibly some more as @earlephilhower suggested.
The problem was that _payload was not reset, causing ::getString() to return the _previously_ fetched payload, in stead of calling ::writeToStream() to read the _current_ payload (and to handle the connection properly)

d-a-v:
I confirm #6149 and #6332 are fixed with this PR. Thanks @Jeroen88 !

d-a-v:
With the examples, I tried faking HTTP/1.0 (by forcing `_canReuse` to false when reading `HTTP/1.`) and #6332 does not work anymore.
This is because of `bool disconnect(preserve_client = false)`:
It can be fixed by either calling `disconnect(true)` (instead of `disconnect()`), or by changing default to `true` (= API break, but I wonder why doing that by default?).

Also, I wonder about the logic of 
```
                 if(headerName.equalsIgnoreCase("Connection")) {
                    _canReuse = headerValue.equalsIgnoreCase("keep-alive");
                 }
```

if we are in `HTTP/1.0` and `keep-alive` is met, what to do ?
if we are in `HTTP/1.1` and `keep-alive` is not met, what to do ?
... googling around ...
Also worth reading: https://stackoverflow.com/questions/11443669/how-to-interpret-connection-keep-alive-close

Maybe we should do
```
                 if(headerName.equalsIgnoreCase("Connection")) {
                   _canReuse &= headerValue.indexOf("close") < 0;
                 }
```

Jeroen88:
@d-a-v I tried to fake HTTP/1.0 as well, in the same way as you did, but everything keeps working. Setting _canReuse = false still keeps sending HTTP/1.1 in the request, might that cause the difference in behavior? I use a NodeJS server that sends a response header Connection: close.


```
[HTTP-Client][begin] url: the url
[HTTP-Client][begin] host: the host port: 443 url: the uri
[HTTP-Client][sendRequest] type: 'GET' redirCount: 0
[HTTP-Client] connected to the host:443
[HTTP-Client] sending request header
-----
GET the uri HTTP/1.1
Host: the host
User-Agent: ESP8266HTTPClient
Connection: close
Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0
Content-Length: 0

-----
[HTTP-Client][handleHeaderResponse] RX: 'X-Powered-By: Express'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Type: application/json; charset=utf-8'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Length: 574'
[HTTP-Client][handleHeaderResponse] RX: 'ETag: W/"23e-DB5ulCBY347Z4j19qxT6CsFyi84"'
[HTTP-Client][handleHeaderResponse] RX: 'Date: Mon, 05 Aug 2019 09:01:40 GMT'
[HTTP-Client][handleHeaderResponse] RX: 'Connection: close'
[HTTP-Client][handleHeaderResponse] size: 574
Code: 200
[HTTP-Client][writeToStreamDataBlock] connection closed or file end (written: 574).
[HTTP-Client][end] tcp is closed
Body: {"config_id":"1" etc...}
[HTTP-Client][sendRequest] type: 'GET' redirCount: 0
[HTTP-Client] connected to the host:443
[HTTP-Client] sending request header
-----
GET the uri HTTP/1.1
Host: the host
User-Agent: ESP8266HTTPClient
Connection: close
Accept-Encoding: identity;q=1,chunked;q=0.1,*;q=0
Content-Length: 0

-----
[HTTP-Client][handleHeaderResponse] RX: 'X-Powered-By: Express'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Type: application/json; charset=utf-8'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Length: 574'
[HTTP-Client][handleHeaderResponse] RX: 'ETag: W/"23e-DB5ulCBY347Z4j19qxT6CsFyi84"'
[HTTP-Client][handleHeaderResponse] RX: 'Date: Mon, 05 Aug 2019 09:01:42 GMT'
[HTTP-Client][handleHeaderResponse] RX: 'Connection: close'
[HTTP-Client][handleHeaderResponse] size: 574
Code: 200
[HTTP-Client][writeToStreamDataBlock] connection closed or file end (written: 574).
[HTTP-Client][end] tcp is closed
Body: {"config_id":"1", etc..}

```

I think the ::disconnect() parameter defaults to false, because maybe the client was developed for HTTP/1.0? _reuse was false by default too, I guess because of the same reason?

Th server might send Connection: close, Connection: keep-alive or Connection: keep-alive, close. In the first case _canReuse should be false, but in the latter two cases it should be true. This leads to:

```
                 if(headerName.equalsIgnoreCase("Connection")) {
                   _canReuse &= headerValue.indexOf("keep-alive") < 0;
                 }

```

or am I missing something??

Jeroen88:
@d-a-v if I use (without the _reuse hack)

```
    https.useHTTP10();
    https.setReuse(true);
```
A Connection: keep-alive is send and a Connection: keep-alive is received. The connection is kept alive accordingly.

If I use instead:

```
    https.useHTTP10();
    https.setReuse(false); // this is already the default for useHTTP10()...
```
A Connection: close is send and a Connection: close is received. The connection is closed accordingly.

So HTTP/1.0 seems to work correctly.

I am surprised though that Express seems to totally ignore the HTTP/1.0 request, because it responses with a HTTP/1.1:

```
[HTTP-Client][handleHeaderResponse] RX: 'HTTP/1.1 200 OK
```

I expected a HTTP/1.0 response for backward compatibility.

Should we put more effort in HTTP/1.0 compatibility? Is it still around?

dmdt:
@Jeroen88 I could not able to make code work anyway... May be that because I manually edited files. When I tried clone repo through platformio - it crashed and now can not display platform tab at all.

Jeroen88:
@dmdt Could you try the PR without modifications?

dmdt:
@Jeroen88 Tried branch 37b62a4, still not working.
```
[env:nodemcuv2]
platform = espressif8266
platform_packages = framework-arduinoespressif8266 @ https://github.com/Jeroen88/Arduino.git
```

Jeroen88:
Could you post your version of HTTPClient::handleHeaderResponse()?

dmdt:
``` cpp
int HTTPClient::handleHeaderResponse()
{

    if(!connected()) {
        return HTTPC_ERROR_NOT_CONNECTED;
    }

    clear();

    _canReuse = !_useHTTP10;


    String transferEncoding;
    _transferEncoding = HTTPC_TE_IDENTITY;
    unsigned long lastDataTime = millis();

    while(connected()) {
        size_t len = _client->available();
        if(len > 0) {
            String headerLine = _client->readStringUntil('\n');
            headerLine.trim(); // remove \r

            lastDataTime = millis();

            DEBUG_HTTPCLIENT("[HTTP-Client][handleHeaderResponse] RX: '%s'\n", headerLine.c_str());

            if(headerLine.startsWith("HTTP/1.")) {
                  if(_canReuse) {
                    _canReuse = (headerLine[sizeof "HTTP/1." - 1] != '0');
                }
                _returnCode = headerLine.substring(9, headerLine.indexOf(' ', 9)).toInt();
            } else if(headerLine.indexOf(':')) {
                String headerName = headerLine.substring(0, headerLine.indexOf(':'));
                String headerValue = headerLine.substring(headerLine.indexOf(':') + 1);
                headerValue.trim();

                if(headerName.equalsIgnoreCase("Content-Length")) {
                    _size = headerValue.toInt();
                }

                if(headerName.equalsIgnoreCase("Connection")) {
                    _canReuse = headerValue.equalsIgnoreCase("keep-alive");
                }

                if(headerName.equalsIgnoreCase("Transfer-Encoding")) {
                    transferEncoding = headerValue;
                }

                if(headerName.equalsIgnoreCase("Location")) {
                    _location = headerValue;
                }

                for(size_t i = 0; i < _headerKeysCount; i++) {
                    if(_currentHeaders[i].key.equalsIgnoreCase(headerName)) {
                        if (_currentHeaders[i].value != "") {
                            // Existing value, append this one with a comma
                            _currentHeaders[i].value += "," + headerValue;
                        } else {
                            _currentHeaders[i].value = headerValue;
                        }
                        break; // We found a match, stop looking
                    }
                }
            }

            if(headerLine == "") {
                DEBUG_HTTPCLIENT("[HTTP-Client][handleHeaderResponse] code: %d\n", _returnCode);

                if(_size > 0) {
                    DEBUG_HTTPCLIENT("[HTTP-Client][handleHeaderResponse] size: %d\n", _size);
                }

                if(transferEncoding.length() > 0) {
                    DEBUG_HTTPCLIENT("[HTTP-Client][handleHeaderResponse] Transfer-Encoding: %s\n", transferEncoding.c_str());
                    if(transferEncoding.equalsIgnoreCase("chunked")) {
                        _transferEncoding = HTTPC_TE_CHUNKED;
                    } else {
                        return HTTPC_ERROR_ENCODING;
                    }
                } else {
                    _transferEncoding = HTTPC_TE_IDENTITY;
                }

                if(_returnCode) {
                    return _returnCode;
                } else {
                    DEBUG_HTTPCLIENT("[HTTP-Client][handleHeaderResponse] Remote host is not an HTTP Server!");
                    return HTTPC_ERROR_NO_HTTP_SERVER;
                }
            }

        } else {
            if((millis() - lastDataTime) > _tcpTimeout) {
                return HTTPC_ERROR_READ_TIMEOUT;
            }
            delay(0);
        }
    }

    return HTTPC_ERROR_CONNECTION_LOST;
}

```

Jeroen88:
@dmdt, that is the correct version. What exactly is not working? Could you enable debug level TTP_CLIENT and post the result?

d-a-v:
My test:
```
             if(headerLine.startsWith("HTTP/1.")) {
                 if(_canReuse) {
                     _canReuse = (headerLine[sizeof "HTTP/1." - 1] != '0');
+                    _canReuse=false;
                 }
```
(force detection of HTTP/1.0)

Log:
```
12:09:55.274 -> [HTTP-Client][handleHeaderResponse] RX: 'Connection: Keep-Alive'
12:09:55.274 -> [HTTP-Client][handleHeaderResponse] RX: ''
12:09:55.274 -> [HTTP-Client][handleHeaderResponse] code: 200
12:09:55.274 -> [HTTP-Client][handleHeaderResponse] size: 2
12:09:55.274 -> Code: 200
12:09:55.274 -> :c0 1, 242
12:09:55.274 -> [HTTP-Client][writeToStreamDataBlock] connection closed or file end (written: 2).
12:09:55.274 -> [HTTP-Client][end] tcp stop
12:09:55.274 -> :close
12:09:55.274 -> Body: 6
12:09:55.274 -> 
12:09:57.264 -> [HTTP-Client][sendRequest] type: 'GET' redirCount: 0
12:09:57.264 -> [HTTP-Client] connect: HTTPClient::begin was not called or returned error
12:09:57.264 -> [HTTP-Client][returnError] error(-1): connection refused
12:09:57.297 -> Code: -1
```

This is fixed by changing default value of `disconnect()`

> I expected a HTTP/1.0 response for backward compatibility.

True, but we should try to work in any case

> Should we put more effort in HTTP/1.0 compatibility? Is it still around? If even Express ignores it...

I don't know, in doubt I think we should.

> Am I missing something??

My proposal was "we cannot reuse if `close` is not present in the connection header line", which is indeed a non-sense.

Yours means "we cannot reuse if `keep-alive` is not present in the connection header line"
```
        if(headerName.equalsIgnoreCase("Connection")) {
            _canReuse &= headerValue.indexOf("keep-alive") < 0;
        }
```

After reading w3c spec (field may not be present), what about:
```
        if(headerName.equalsIgnoreCase("Connection")) {
            if (headerValue.indexOf("close") >= 0)
                _canReuse = false;
        }
```
which means "we cannot reuse if `close` is present in the connection header line" .

dmdt:
@Jeroen88 By the some reason pio cloned repo without changes, and couldn't compile firmware with enabled HTTP_CLIENT_DEBUG. But with manual changes and after few requests I get -11 response code  and following -1. As good as on my test sketch and on the bigger sketch. It seems like some buffer overflow, because on lesser sketch it appears later. If you got all working fine - may be fail on my side. Try #6332 MCVE sketch working some time (5-7 min) with differ receiving data.
```
[DEBUG]Connecting to WiFi...
[DEBUG]Update initialized.
[DEBUG]Composed update link: https://api.prsk.pw/update_center/update_handler.php
[DEBUG]Update error (-101): Server Did Not Report Size
[DEBUG]Update ended.
HeHe
[DEBUG]Final API request link:
https://api
[DEBUG]Attempting connect to API.
[DEBUG]Connect to API successful.
[DEBUG]Request sent.
[DEBUG]Old response:
[DEBUG]Recieved response: mode?a=16711680&b=1
[DEBUG]Parsed a as: 16711680
[DEBUG]Parsed b as: 1
[DEBUG]Parsed color: 16711680
[DEBUG]Parsed mode: 1
[DEBUG]Mode 1.
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=1
[DEBUG]Recieved response: mode?a=16711680&b=1
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=1
[DEBUG]Recieved response: mode?a=16711680&b=1
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=1
[DEBUG]Recieved response: mode?a=16711680&b=3
[DEBUG]Parsed a as: 16711680
[DEBUG]Parsed b as: 3
[DEBUG]Parsed color: 16711680
[DEBUG]Parsed mode: 3
[DEBUG]Mode 3.
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=3
[DEBUG]Recieved response: mode?a=16711680&b=3
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=3
[DEBUG]Recieved response: mode?a=16711680&b=3
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=3
[DEBUG]Recieved response: mode?a=16711680&b=2
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=2
[DEBUG]Recieved response: mode?a=16711680&b=2
[DEBUG]Parsed a as: 16711680
[DEBUG]Parsed b as: 2
[DEBUG]Parsed color: 16711680
[DEBUG]Parsed mode: 2
[DEBUG]Mode 2.
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=2
[DEBUG]Recieved response: mode?a=16711680&b=2
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=2
[DEBUG]Recieved response: mode?a=16711680&b=4
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=4
[DEBUG]Recieved response: mode?a=16711680&b=4
[DEBUG]Parsed a as: 16711680
[DEBUG]Parsed b as: 4
[DEBUG]Parsed color: 16711680
[DEBUG]Parsed mode: 4
[DEBUG]Mode 4.
[DEBUG]Request sent.
[DEBUG]Old response: mode?a=16711680&b=4
[DEBUG]Recieved response: mode?a=16711680&b=4
[DEBUG]Request sent.
[DEBUG]Request error. Response code: -11
[DEBUG]Request sent.
[DEBUG]Request error. Response code: -1
And so on after that.
```

Jeroen88:
@d-a-v I did exactly the same test. Stangely we get different results...

Would you rather change the default to true than calling ::disconnect(true)?

My interpretation of https://stackoverflow.com/questions/11443669/how-to-interpret-connection-keep-alive-close is, with Connection: keep-alive, close, the client may choose. I would choose to keep the connection alive, why would you close it? In what I can find in W3C documentation is only an explicit Connection: close, I can find any reference to Connection: keep-alive, close...

Jeroen88:
@dmdt Response -11 means a read time out, see in ESP8266HTTPClient.h:

```
#define HTTPC_ERROR_READ_TIMEOUT        (-11)
```
I think this has nothing to do with keeping the connection alive, would you?

I might be that a time out is not handled correctly, but this would be another issue. I am not sure if ESP8266HTTPClient should handle the time out. You could handle it at application level (if you get the HTTPC_ERROR_READ_TIMEOUT response call ::end() and next call ::begin() again?

dmdt:
> @dmdt Response -11 means a read time out, see in ESP8266HTTPClient.h:
> ```#define HTTPC_ERROR_READ_TIMEOUT        (-11)```
> I think this has nothing to do with keeping the connection alive, would you?

Yeah, I saw. Will try figure out what's wrong.

> NB, my test (using your example .ino) ran for several hours without any problems.

It strange... You test it on static response data? Then I leave esp with static response, may be thing in this.

Jeroen88:
@dmdt What do you mean by 'static response'? My url responded with the same data every time, but I made sure that is was read every time (using some debug println's)

d-a-v:
@Jeroen88 

About `::disconnect(bool preserveClient)`,
Ideally, default value should be `true`, why would we need to call `http.begin` everytime ?
But that'd be an API break (`::disconnect()` is not used in examples though).
To prevent a silent API break, I'd remove the default value, and use true (= preserve client) everywhere in HTTPCLient.

> I would choose to keep the connection alive, why would you close it?

I propose to close it only when it is required (`if (headerValue.indexOf("close") >= 0) _canReuse = false;`)

Jeroen88:
@d-a-v in::end() there is a disconnect() too, this should be disconnect(false), right? The other one should be disconnect(true).

I am still not sure of your interpretation versus mine, especially because keep-alive is the default for HTTP/1.1. But I will follow your suggestion :)

d-a-v:
> in::end() there is a disconnect() too, this should be disconnect(false), right?

I think that too. `end()` semantic is when one want to call `begin()` again.

> I am still not sure of your interpretation versus mine, especially because keep-alive is the default for HTTP/1.1. But I will follow your suggestion :)

You must not follow if you don't agree, because it's your PR.
Do you agree with: 
```
        if(headerName.equalsIgnoreCase("Connection")) {
            if (headerValue.indexOf("close") >= 0 &&
                headerValue.indexOf("keep-alive") < 0) {
                _canReuse = false;
            }
        }
```

Jeroen88:
@d-a-v, @dmdt, I get a similar error like @dmdt after a while:

```
[HTTP-Client][handleHeaderResponse] RX: 'X-Powered-By: Express'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Type: application/json; charset=utf-8'
[HTTP-Client][handleHeaderResponse] RX: 'Content-Length: 574'
[HTTP-Client][handleHeaderResponse] RX: 'ETag: W/"23e-DB5ulCBY347Z4j19qxT6CsFyi84"'
[HTTP-Client][handleHeaderResponse] RX: 'Date: Mon, 05 Aug 2019 16:12:11 GMT'
[HTTP-Client][handleHeaderResponse] RX: 'Connection: close'
[HTTP-Client][handleHeaderResponse] size: 574
Code: 200
[HTTP-Client][writeToStreamDataBlock] connection closed or file end (written: 574).
[HTTP-Client][end] tcp stop
Body: {"config_id":"1" ... etc}
[HTTP-Client][sendRequest] type: 'GET' redirCount: 0
[HTTP-Client] connect: HTTPClient::begin was not called or returned error
[HTTP-Client][returnError] error(-1): connection refused
Code: -1
Body: {"config_id":"1" ... etc}
[HTTP-Client][sendRequest] type: 'GET' redirCount: 0
[HTTP-Client] connect: HTTPClient::begin was not called or returned error
[HTTP-Client][returnError] error(-1): connection refused
```

Should this be handled by the client? Or should this be handled by the application? I assume that calling ::end() and then ::begin() will restore the connection, but I did not try it yet...

The logging comes from a HTTP/1.1 example with setReuse (false)

d-a-v:
Isn't that fixed with disconnect(true) ?

Jeroen88:
@d-a-v Nope, this is with the code that I just commited

Jeroen88:
@d-a-v ::geString() still returns the previously fetched payload. ::clear() should be called earlier in the process?

d-a-v:
There is a missing `true` L994. Without it I still have the error when forcing `_canReuse` to false.

Jeroen88:
@d-a-v 

> Do you agree with:

        if(headerName.equalsIgnoreCase("Connection")) {
            if (headerValue.indexOf("close") >= 0 &&
                headerValue.indexOf("keep-alive") < 0) {
                _canReuse = false;
            }
        }

Yes, I do agree :)

Missing 'true' added.

I think I found another bug, in ::handleHeaderResponse() _canReuse should be initialized to _reuse and not to !_useHTTP10. In this way 
```
    https.useHTTP10();
    https.setReuse(true);
```
works correctly.

I lost some work during pushing and pulling. I think I recreated everything, could you confirm? I did test HTTP/1.0 and HTTP/1.1, both with and without reusing the connection, but not the '_canReuse = false'-hack, and everything looks fine.

Jeroen88:
@d-a-v What about getting an http code < 0 indicating an error? I am running a test now, waiting for an error, and the call ::end() and ::begin() to restore the connection, to see if this works.

I think it should be handled in the .ino, not in the library, do you agree?

d-a-v:
I get no more error with `_canReuse=false` faking http/1.0 :+1:

> What about getting an http code < 0 indicating an error?

I think I don't understand the question:

- all `::sendRequest` return values are <0 on error, and
- http code [are normalized](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)

> I think it should be handled in the .ino, not in the library, do you agree?

Yes, error handling should be done in sketch, and demonstrated in examples.

Jeroen88:
@d-a-v That was my question, error handling should be done in the sketch :)

Thanks for the cooperation on this one!

Jeroen88:
@dmdt, @d-a-v, a sketch is running for more than 10 hours now, GETting data every 2 seconds without any error

Jeroen88:
@d-a-v Shouldn't we keep the default parameter for ::disconnect()? Every call from the library is with parameter now, and keeping the default parameter does not break any sketch.

d-a-v:
> Shouldn't we keep the default parameter for ::disconnect()?

I think it will give this PR a faster chance to get merged :)

Jeroen88:
@d-a-v Done!

Jeroen88:
@d-a-v I think I found another flaw in the ESP8266HTTPClient:

::getString() calls ::writeToStream() and at the end of this last function ::disconnect(true) is called. I think after every call to ::connect() a call to::disconnect() should be made. However, if ::sendRequest() returns an error, my sketch does not make a call to ::getString() (because it tests for a return code < 0) and then  ::disconnect() is not called. Also, if the server returns 204, no content, getString() returns an error, because the client has already disconnected, and there is no data available.

Both errors can easily be corrected, but lead to the requirement to always call ::getString() (even if you know there is no content or if you e.g. read the result directly using ::getStreamPtr()) because otherwise the connection is not handled properly. 

I do not like the above solution (although I am running tests on it now)

Another option would be to introduce a new function, e.g. ::stop() or any other name that should always be called after finishing reading the response, or require ::disconnect(true) to be called always. But the latter suggests that the client is disconnected while it might be reused...

I do not like this solution either.

Any ideas?

d-a-v:
@Jeroen88 
Can we simply add more `::disconnect(true)` on `204` or failing `::sendRequest()`, and/or internally read+cancel content (if needed, maybe with the help of a new `bool` set to true when user has read content, in `::sendRequest()`) when user has not previously read content / called `::getString()` ?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -563,6 +573,7 @@ void HTTPClient::setRedirectLimit(uint16_t limit)
 void HTTPClient::useHTTP10(bool useHTTP10)
 {
     _useHTTP10 = useHTTP10;
+    _reuse = !useHTTP10;
```
earlephilhower: With this method and `::setReuse()` it is possible to set `useHTTP10 = true` and `_reuse=true`.  That is the user shooting themselves in the foot, which is not a high prio for me.

At some point might want to go back and see if `useHTTP10 === !reuse` (I think so after reading the code briefly) and then drop one or the other variable if that's true.  But for this PR, I'm OK.

d-a-v:
@Jeroen88 can you open a new issue with MCVE about your last message ?

Jeroen88:
@d-a-v I will later this week

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
Jeroen88: @earlephilhower for HTTP/1.0 the default is *not* reusing the connection, the default for HTTP/1.1 is reusing. However reusing is allowed also on HTTP/1.0. In this case the client must send a Connection: Keep-Alive request header. The server may respond with a Connection: Keep-Alive in which case the connection can be reused. In HTTP/1.1 the client may send a connection: Keep-Alive header, but this is not obligatory. The connection is supposed to be reused unless the *server* sends a connection: close response header.
This is exactly what ESP8266HTTPClient does

PR #6055, opened 2019-05-06
Completes #6048: let FunctionalInterrupt.h support only IRAM-safe scheduled ISRs

dok-net:
Makes for a much cleaner cores/esp8266/core_esp8266_wiring_digital.cpp:
- no special "functional" predicate
- no reverse "duplicate from functionalInterrupt.h keep in sync" copied-code dependency

PR #6182, opened 2019-06-05
Pluggable Scheduler: Make entry points into sequential cont/user scheduler weak functions

dok-net:
For sketches that don't use libraries depending on a user scheduler, this PR helps eliminate the Scheduler from the binaries, saving memory. To complete this, the Scheduler would have to be moved from core to the libraries.
Also, plugging different schedulers becomes more obvious to do.

Questions:
- is there a way to make the concrete Scheduler used in Ticker or LEAmDNS easily pluggable?
- are there different preferences for naming` loop_completed()` and  `yield_completed()`?


dok-net:
Next step would be to remove #include "Schedule.h" from core_esp8266_main.cpp; move FunctionalInterrupt.(h|cpp) and Schedule.(h|cpp) into libraries.

Blink.ino before:
Program size: 257,740 bytes (used 25% of a 1,044,464 byte maximum) (1.99 secs)
Minimum Memory Usage: 26548 bytes (32% of a 81920 byte maximum)

After:
Program size: 257,356 bytes (used 25% of a 1,044,464 byte maximum) (22.41 secs)
Minimum Memory Usage: 26532 bytes (32% of a 81920 byte maximum)

@d-a-v @devyte Please review?

dok-net:
@hreintke What do you think? Could you please check the FunctionalInterrupt/library.properties?

dok-net:
@d-a-v @earlephilhower @devyte This is 3 weeks old, and I am sorry, had gotten out of control somehow, being a PR containing a lot of unrelated changes.
I've cleaned up now to the minimum changeset, and I would like to ask you again to review and consider merging, possibly ahead of #6214 , which I personally am having a bit of a hard time figuring out what the rationale is, nevermind.

dok-net:
@d-a-v #6214 , why o' why...
The explanation is quite comprehensive yet. I disagree with the wording "long running operations or yield() or delay() are NOT WISE in the lambda." - should be either a yes or a no, nothing indeterminate.
I think it's a mistake not allowing one-shot scheduled functions to run on yield now - an async programming style yields much slower or un-responsive sketches this way. And this just because yields are not allowed? One-shot functions don't have to be slower - are the scheduled interrupts not hampered now for no good reason at all?

d-a-v:
> I think it's a mistake not allowing one-shot scheduled functions to run on yield now - an async programming style produces much slower or un-responsive sketches this way. 

I agree. I did what the majority told me.

In some other comment it was clearly stated that scheduled functions should be allowed to run `delay()` and `yield()`, plus the fifo semantic, plus once per loop. I did my homework. This doesn't fit my (ethernet) needs so I added the recurrent version.

You can however use recurrent scheduled function like you do with scheduled function with
- `return false`
- `::alwaysExpires` polledTimeout parameter.
- without fifo semantic

dok-net:
@d-a-v I'll see what I can do to the circular-queue based scheduler. First, I'd like to see the EspSoftwareSerial issue (#6227) through, the attachInterrupt fix/improvement is important (#6047 does away with #6055 and #6048 in one step), and getting this here pluggability into core would be awesome.

dok-net:
@d-a-v 

> You can however use recurrent scheduled function like you do with scheduled function with
> 
> return false
> ::alwaysExpires polledTimeout parameter.
> without fifo semantic

Oh wait a minute - is this a suggestion for scheduled ISRs? Doing those without FIFO semantics is completely wrong.

d-a-v:
It's not a suggestion for scheduled ISR, there are conditions if you wish to do that, and one of them is "*without fifo semantic*".

> Oh wait a minute - is this a suggestion for scheduled ISRs? Doing those without FIFO semantics is completely wrong.

Now if you think scheduled ISR should be called also from `yield`, fifo semantic needs to be added to recurrent scheduled functions, which will increase overhead, which was an argument against the previous implementation.

dok-net:
@d-a-v What's in master now doesn't work, I am sorry:

```

wdt reset
tail 8
chksum 0x2d
csum 0x2d
vb20cfcf5
~ld

```

Just running `ScheduledFunctional.ino` , then clicking button 2:

```

#ifndef IRAM_ATTR
#define IRAM_ATTR ICACHE_RAM_ATTR
#endif

#if defined(ESP32)
#define BUTTON1 16
#define BUTTON2 17
#elif defined(ARDUINO_ESP8266_WEMOS_D1MINI)
#define BUTTON1 D4
#define BUTTON2 D3
#else
#define BUTTON1 2
#define BUTTON2 0
#endif

class Button {
  public:
    Button(uint8_t reqPin) : PIN(reqPin) {
      pinMode(PIN, INPUT_PULLUP);
      attachScheduledInterrupt(PIN, [this](const InterruptInfo & ii) {
        Serial.print("Pin ");
        Serial.println(ii.pin);
        buttonIsr();
      }, FALLING); // works on ESP8266
    };
    ~Button() {
      detachInterrupt(PIN);
    }

    void IRAM_ATTR buttonIsr() {
      numberKeyPresses += 1;
      pressed = true;
    }

    static void IRAM_ATTR buttonIsr_static(Button* const self) {
      self->buttonIsr();
    }

    uint32_t checkPressed() {
      if (pressed) {
        Serial.printf("Button on pin %u has been pressed %u times\n", PIN, numberKeyPresses);
        pressed = false;
      }
      return numberKeyPresses;
    }

  private:
    const uint8_t PIN;
    volatile uint32_t numberKeyPresses = 0;
    volatile bool pressed = false;
};

Button* button1;
Button* button2;


void setup() {
  Serial.begin(115200);
  Serial.println("FunctionalInterrupt test/example");

  button1 = new Button(BUTTON1);
  button2 = new Button(BUTTON2);

  Serial.println("setup() complete");
}

void loop() {
  button1->checkPressed();
  if (nullptr != button2 && 10 < button2->checkPressed()) {
    delete button2;
    button2 = nullptr;
  }
}

```

d-a-v:
> What's in master now doesn't work, I am sorry.

That one was hairy. It will be fixed by #6233 (I hadn't retested after change from malloc to new).

dok-net:
@d-a-v Can you help me? Where's the discussion that led to implementing two separate lists for immediate and timed scheduled functions? Was there any proof of the efficiency of this?

d-a-v:
There are two many pull request of yours, and we can't read and understand them all while fixing other parts of the core (everyday brings a new orthogonal issue to reproduce and fix). We also try to improve what we have, and think of the future with nonosk EOLing, while having a real life with kids, wife, work and IRL issues to fix too. Little place for resting.

Please come with a single pull request, not several versions constantly keeping CI busy, and please summarize what's improved in terms of timing, flash occupation, complexity, and benefits. When we have a proper explanation easy to read, with examples simple to execute and demonstrating the goals, then we may be able to read your huge PR and discuss about it.

> Where's the discussion that led to implementing two separate lists for immediate and timed scheduled functions?

The loop dealing with recurrent and regular scheduled function was too complex. I restored the old fifo structure, and added another trivial one for recurrent functions. There is no proof. Some discussions stated there was too much overhead because of the requirements (once per loop for legacy scheduled function, yield and delay allowed, but at every yield for the regular scheduled functions). I thought that too. The current implementation is nearly exactly the former scheduled functions implementation. Recurrent functions have been added. So it is not complex anymore.

dok-net:
Kind of an offensive comment. Well, whatever. Is there a commit message tag that will spare the CI server from running, like ***NO_CI*** or something?
I (obviously) disagree with you - nobody told me in nearly two months time if and how there was a way to keep PRs (or something identical in spirit) open for discussion without having them open in here - which is the point of open-source development on GitHub, I might add.
At the heart of it, there is one bug-fix for the crashing `detachInterrupt` , that I have documented to my content. Then second, for lack of interaction, I had what's basically two different branches for a scheduler that doesn't do weird stuff for what I have in mind (triggering non-timing critical work right out of IRQ handlers without thrashing in `loop()`).
I can't help it if your issues outweigh mine, but another person's priorities may be different altogether, again.
As far as I am concerned, I am trying to get the best out of EspSoftwareSerial, and have some fun doing it. That said, from my POV it's rather me who's offering help, not vice-versa. If you don't like comments and suggestions, I may have to look elsewhere, this hidden cost quickly outweighs the cheap hardware price.
Your request to stop committing / rebasing branches in MY repository equates a request to let them rot in the dust. I am sorry, Dave, I can't do that (tm). Retracting a PR is a option, but then, what's the point of it anymore.
Splitting up PRs would increase they number and hide their intent from review. This Github online review tool is painful for anything more than simple bugfixes, I agree.
With any indication of whether the minimalistic C-style bug-fix #6048 or the complete C++ overhaul in #6047 is preferred, I would be more that happy to drop the one less likely to get merged - majority opinion counting in this. I restate my opinion, `__attachInterruptArg()` is out in the wild and must work for non-"FunctionalInterrupt.h" uses.

Revisiting #6139 , insult comes to injury: I've long done what you demand, given a reproducable sketch, given rationale, given timings. Iterative improvements are bound to create a longer commit history and discussion thread, there's nothing that can be done about that. Asking me to go back over and over to prove the same point get kind of boring for me. Before I start writing another test procedure, please let me know one thing that's not documented in the schedule.h right now: is it supposed to be legal to schedule functions from inside a function running off the scheduler?

d-a-v:
@dok-net 

> Kind of an offensive comment

This is my first. It is like *lots* of yours, most recent example is [just above](https://github.com/esp8266/Arduino/pull/6182#issuecomment-505429227).

> that will spare the CI server from running, like NO_CI or something?
> Your request to stop committing / rebasing branches in MY repository equates a request to let them rot in the dust.

No. Please forget about what I said about CI. sorry. You have the right to push. About that, please stop rebasing (if you can). The github button "squash and merge" does the final rebasing.

> At the heart of it, there is one bug-fix for the crashing detachInterrupt , that I have documented to my content.

Maybe. Your posts are just TL;DR. I am lost with all your PRs. I guess I am not alone. Like with two solutions for the same problem, long PRs with lots of commits and comments asking for attention. Which one to read first ? The answer is none. Finally #6047 is noticed because I catched a line where you said if is better than another one of yours on the same subject. Please close those you consider worse.

>  I had what's basically two different branches for a scheduler that doesn't do weird stuff for what I have in mind 

Nobody says your code is weird. I said we just can't give you all the attention you need. When I came to this repo with lwip2, it took months before it got attention even though it solved quite a number of bugs. It took months for you to be heard in SoftwareSerial and you even came to us about that, us with no power on this repo. I speak in my name and I find this too much. I am trying to explain without offense how I receive your comments like the one above. And still I keep answering to you where others would just ignore these comments, or you.  You are free to take my comments as you like, and keep on the same line, or try to think about that.

>  I am trying to get the best out of EspSoftwareSerial, and have some fun doing it

On that we agree. But what if someone is constantly coming to you telling you stuff like `Doing 
so is completely wrong.`, or `there is too much overhead`, or things looking like "nobody is listening to me" ? Fun goes away.

> from my POV it's rather me who's offering help, not vice-versa

For example, what is that ? What are you trying to say ?

> this hidden cost quickly outweighs the cheap hardware price

And that ? **Fun** has a money price now ?
I am aware a lot of people earn money with their application based on esp8266-arduino, maybe including SoftwareSerial that is shipped with it. I personally didn't receive any cent for that (apart from the IPv6 bounty the donator insisted for me to take it - and thanks for that I got some beers :)

> Revisiting #6139 , insult comes to injury: I've long done what you demand, given a reproducable sketch, given rationale, given timings.

Like I said. Too many PRs and TL;DR kill the interesting information.

> Before I start writing another test procedure, please let me know one thing that's not documented in the schedule.h right now: is it supposed to be legal to schedule functions from inside a function running off the scheduler?

You are right, it is not documented. `schedule_function()` can be called from an ISR or a scheduled function. I could have used that for recurrent functions, but I find that removing a function to always reinsert it is too much overhead.

libraries/Schedule/library.properties
```diff
@@ -1,7 +1,7 @@
 name=Schedule
 version=1.0
-author=Earle F. Philhower, III <earlephilhower@yahoo.com>
-maintainer=Earle F. Philhower, III <earlephilhower@yahoo.com>
+author=David Gauchard <gauchard@laas.fr>
+maintainer=David Gauchard <gauchard@laas.fr>
```
d-a-v: @igrr Ivan Grokhotkov (ivan@esp8266.com) is the original author.

dok-net:
Withdrawing this PR - existing Scheduler in core has received patches, works nicely, belongs in core.

PR #6187, opened 2019-06-06
Obey the BASIC_SSL request for TLS servers

earlephilhower:
When in basic SSL mode, don't include ciphers for EC for the server by
replacinf the BearSSL "support everything" init call with our own
version which obeys the requested BASIC_SSL.

Saves ~47KB on TLS servers, about the same as basic SSL saves for
clients.

Also saves ~80bytes of heap due to some static local variables being removed.


earlephilhower:
Just tested connecting to a server with and without, gives the expected ~46KB difference in program size and properly connects using allowed ciphers.

PR #6026, opened 2019-04-30
fix switching to static address with lwip2

d-a-v:
For some reason, ip address is not propagated in a visible way for lwip2
when switching to static address (wifi.config()) *after* wifi.begin().

This patch calls lwip-v1.4's netif_set_addr() with the new ip address to set
all things up, just like it is done and right when wifi.begin() is called
after wifi.config().

Also tested when IPv6 is enabled.

fixes #5839
fixes #6024

PR #6194, opened 2019-06-10
lwip2: (re)fix setting static ip address

### LINKED ISSUE #6191: Activating static IP makes WiFi.status() return WL_CONNECTED
### Basic Infos

#### Platform

- Hardware:					ESP-12
- Core Version:			Latest Git as of 2019-06-10
- Operating System:		Windows

### Settings in IDE

- Module:  			NodeMCU 1.0
- Flash Mode:			dio
- Flash Size: 		4MB
- Reset Method:		nodemcu
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

The changes made in PR #6026 by @d-a-v makes `WiFi.status() == WL_CONNECTED` become true whenever one activates static IP via `WiFi.config`. This was not the case before the change, and it also seems strange that the status would be connected even when no connection has been made, so it is likely a bug. Calling `WiFi.disconnect()` after static IP activation makes `WiFi.status()` return 0 (`WL_IDLE_STATUS`), as was the case before the PR change. 

```cpp

#include <ESP8266WiFi.h>

void activateStaticIP()
{
  // WiFi mode does not matter 
  WiFi.mode(WIFI_STA);
  //WiFi.mode(WIFI_AP_STA); 
  
  WiFi.config(IPAddress(192,168,4,22), IPAddress(192,168,4,1), IPAddress(255,255,255,0)); 
}

void setup() {

  WiFi.persistent(false);
  
  Serial.begin(115200);
  delay(50); // Wait for Serial.

  Serial.println();
  Serial.println();

  WiFi.disconnect();
    
  activateStaticIP();
}

void loop() {
  
  Serial.print("status pre " + String(WiFi.status()));
  if(WiFi.status() == WL_CONNECTED)
    Serial.println(", this means WiFi.status() == WL_CONNECTED");
  else
    Serial.println();
    
  WiFi.disconnect(); // Adding this disconnect seems to fix WiFi status.
  Serial.println("status post " + String(WiFi.status()));

  delay(3000);
}
```

d-a-v:
#6026 was not tested in every cases.
This PR hopefully fixes a missing one
(trigger address change by calling lwIP-v1.4 api *only when ip is already set by other mean (generally dhcp)*)
fixes #6191

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -309,6 +309,13 @@ bool ESP8266WiFiSTAClass::config(IPAddress local_ip, IPAddress arg1, IPAddress a
     return false;
   }
 
+#if LWIP_VERSION_MAJOR != 1 && !CORE_MOCK
+  // get current->previous IP address
+  // (check below)
+  struct ip_info previp;
+  wifi_get_ip_info(STATION_IF, &previp);
+#endif
+
   struct ip_info info;
   info.ip.addr = local_ip.v4();
   info.gw.addr = gateway.v4();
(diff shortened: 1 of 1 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -309,6 +309,13 @@ bool ESP8266WiFiSTAClass::config(IPAddress local_ip, IPAddress arg1, IPAddress a
     return false;
   }
 
+#if LWIP_VERSION_MAJOR != 1 && !CORE_MOCK
```
devyte: Please move this block down and merge with the one at the bottom. There doesn't seem to be a need to have a separate block up here.

d-a-v: I can't. `wifi_get_ip_info()` would get `wifi_set_ip_info()`'s result (called just below).

devyte: Ah yes you're right.

PR #6195, opened 2019-06-11
fix random(min,max) in WMath.cpp

a7md0:
Update WMath.cpp function
`long random(long howsmall, long howbig)`
`secureRandom(long howsmall, long howbig)`
The previous code is inclusive of argument howsmall, but exclusive of argument howbig. E.g. => random(1, 3) will return only [1 or 2].


d-a-v:
This breaks [arduino API](https://www.arduino.cc/reference/en/language/functions/random-numbers/random/): min is inclusive, max is exclusive

devyte:
Thanks for the PR, but previous comment is correct, this can't go in.

PR #6196, opened 2019-06-11
[ESP8266WiFiSTA] Add get broadcast IP address functionality

a7md0:
Calculate the broadcast IP address using the subnet mask and default gateway IP address.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -499,6 +499,23 @@ IPAddress ESP8266WiFiSTAClass::gatewayIP() {
 }
 
 /**
+ * Get the broadcast ip address.
+ * @return IPAddress broadcastIP
+ */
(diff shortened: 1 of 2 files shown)


d-a-v (approved):

devyte (changes requested):
This looks like generic network calcs. I think it is therefore also applicable to the AP interface, in whichncasenit shouldn't be in the STA class.

a7md0:
> This looks like generic network calcs. I think it is therefore also applicable to the AP interface, in whichncasenit shouldn't be in the STA class.

It would be great if you redirected me to the "generic network class" to add this code.

d-a-v:
I think what @devyte meant was to move your function into `ESP8266WiFiGeneric.cpp` with a parameter (the interface index `STATION_IF` or `SOFTAP_IF`), and add a call to this function from both ESP8266WiFiSTA and ESP8266WiFiAP with the matching interface index.

TyIsI:
I believe this one has been obsoleted by #7899?

PR #6212, opened 2019-06-20
delay() refactoring

emelianov:
Refactoring delay to allow to run esp_schedule() and esp_yield()
periodically (not only on delay ends).
This will be helpfull in implementation of Ethernet dirver.

### DIFF
--- a/cores/esp8266/core_esp8266_wiring.cpp
+++ b/cores/esp8266/core_esp8266_wiring.cpp
@@ -44,15 +44,17 @@ void delay_end(void* arg) {
 }
 
 void delay(unsigned long ms) {
-    if(ms) {
-        os_timer_setfn(&delay_timer, (os_timer_func_t*) &delay_end, 0);
-        os_timer_arm(&delay_timer, ms, ONCE);
+    if (ms) {
+        unsigned long delay_start = millis();
+        while(millis() - delay_start < ms) {
+            os_timer_setfn(&delay_timer, (os_timer_func_t*) &delay_end, 0);
+            os_timer_arm(&delay_timer, 1, ONCE);
+            esp_yield();
+            os_timer_disarm(&delay_timer);
+        }
     } else {
         esp_schedule();
-    }
-    esp_yield();
-    if(ms) {
-        os_timer_disarm(&delay_timer);
+        esp_yield();
     }
 }
 


d-a-v:
[Per this detailed explanation](https://github.com/esp8266/Arduino/pull/6107#issuecomment-493656391), `delay()` is *sometimes* supposed to be interrupted by `esp_schedule()` before its end (like the timeout in `ClientContext`).

This is used in some places in the core, so we **also need to**
- rename current `delay()` to `interruptible_delay()`
- identify the `delays()` supposed to be interrupted and use `interruptible_delay()` instead
- be aware of all potential side effects

However this PR is very nice for recurrent scheduled functions to be called when some libraries or sketches call `delay(some-long-delay);`

cores/esp8266/Arduino.h
```diff
@@ -209,6 +209,7 @@ unsigned long millis(void);
 unsigned long micros(void);
 uint64_t micros64(void);
 void delay(unsigned long);
+void interruptable_delay(unsigned long ms);
```
d-a-v: `interruptible_delay()` ?

emelianov: [Sorry, pushed changes with mistype before verify]
I've walk through delay() calls in esp8266/Arduino and found only two suspected delays both in mDNS related stuff. All other delays (in libraries code) defined by constants and not longer than 300mS (so they out of possible problems scope).

d-a-v:
This delay is supposed to be interrupted by `esp_schedule()` from `wifi_dns_found_callback()`:
https://github.com/esp8266/Arduino/blob/653f58e209f6a036da78f600d8082c488403d65a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp#L568

libraries/ESP8266mDNS/src/LEAmDNS.cpp
```diff
@@ -640,7 +640,7 @@ uint32_t MDNSResponder::queryService(const char* p_pcService,
         if (_sendMDNSServiceQuery(*pServiceQuery)) {
             // Wait for answers to arrive
             DEBUG_EX_INFO(DEBUG_OUTPUT.printf_P(PSTR("[MDNSResponder] queryService: Waiting %u ms for answers...\n"), p_u16Timeout););
-            delay(p_u16Timeout);
```
d-a-v: @LaborEtArs @hreintke 
Is this delay supposed to be interrupted by `esp_schedule()` ?
(I don't think so)

Ideally, should there be instead a timer started here and that would be tested later ?

emelianov: After some time of thinking it's seems to me that we should leave generic delay() in mDNS code. As delay is not required to be interrupted -- it should lasting to get responses from [unknown count of] multiple mDNS actors.

cores/esp8266/core_esp8266_wiring.cpp
```diff
@@ -56,6 +56,21 @@ void delay(unsigned long ms) {
     }
 }
 
+void delay(unsigned long ms) {
+    if (ms) {
+        unsigned long delay_start = millis();
+        while(millis() - delay_start < ms) {
```
devyte: Please reimplement this in terms of polledTimeout, see examples.

emelianov:
Other delays (in ClientContext) updated during #6213

cores/esp8266/core_esp8266_wiring.cpp
```diff
@@ -56,6 +57,22 @@ void delay(unsigned long ms) {
     }
 }
 
+void delay(unsigned long ms) {
+    if (ms) {
+        using esp8266::polledTimeout::oneShotMs;
+        oneShotMs waitDone(ms);
+        while(!waitDone) {
+            os_timer_setfn(&delay_timer, (os_timer_func_t*) &delay_end, 0);
+            os_timer_arm(&delay_timer, 1, ONCE);
```
emelianov: To reduce power consumption during delay cycles i've left os_timer_arm and no used something like
```cpp
using oneShotYieldMs = polledTimeout::timeoutTemplate<false, YieldPolicy::YieldOrSkip>;
oneShotYieldMs waitDone(ms);
while (!waitDone);
```

dok-net:
@devyte Please, with all due respect, this PR right now is a horrible idea from my POV, given that it would, depending on how exactly it's fixed, either terribly break or at least work against the (cooperative multi-loop() multitasking/multithreading library CoopTask](https://github.com/dok-net/CoopTask).

I cannot be sure to claim that CoopTask is sufficiently evolved to handle what #6213 needs (believing it does so fine is another thing ;-) ), but letting other code run during ``delay()``s is precisely one of the use cases for cooperative multitasking. I'm sorry for the pitch, but you can probably empathize that I'd rather have my implemention get more use rather than being locked out of the ESP8266 core libraries by ongoing work in the opposite direction. An alternate delay-like ``delay_interruptible()`` must not happen, please.

d-a-v:
@dok-net `interruptible_delay()` can be a weak function that you'd override in your environment.

dok-net:
@d-a-v @devyte I think it's not as simple as that, because it's not the delays, but the breaking from the delays that is troubling:
```c++
libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp:657:    esp_schedule(); // resume the hostByName function
libraries/ESP8266WiFi/src/ESP8266WiFiScan.cpp:326:        esp_schedule();
libraries/ESP8266WiFi/src/ESP8266WiFiSTA-WPS.cpp:110:    esp_schedule(); // resume the beginWPSConfig function
libraries/ESP8266WiFi/src/include/ClientContext.h:438:            esp_schedule(); // break current delay()
libraries/ESP8266WiFi/src/include/ClientContext.h:539:            esp_schedule(); // break current delay()
libraries/ESP8266WiFi/src/include/ClientContext.h:615:            esp_schedule(); // break current delay()
```
Once there are multiple concurrent delays in multitasking, the concept of "current delay()" no longer applies. Add on top that the esp_scheduler follows only a minimal 2 task design, sys and cont, and there is no off-the-shelf solution to this problem.
Right now,
```c++
// Following delay will be interrupted by connect callback
for (decltype(_timeout_ms) i = 0; _connect_pending && i < _timeout_ms; i++) {
    // Give scheduled functions a chance to run (e.g. Ethernet uses recurrent)
    delay(1);
}
```
would work without the ``esp_schedule()`` calls to cancel the 1ms delay immediately, just using ``_connect_pending = false;`` and letting the remaining part of the 1ms expire.
Luckily, also right now, it seems to me that ``esp_schedule()`` is not causing harm regardless that ``cont_yield()`` is overwritten by CoopTask. To be continued...

d-a-v:
I think you don't understand @dok-net 

We can modify the delay(1) loop that you cite above to call a function that does the same job, and make it weak.

So you can redefine it with whatever you think is good for your library. You know we can adapt, and we often do, so every use case has its chance to live and work.

We can do that for both function delay and delay_interruptible, both can be weak. So you don't need to be so much concerned about that. Ethernet is not yet merged and this PR will probably happen afterwards.

dok-net:
@d-a-v Each time there is code that expects ``esp_schedule()`` a.k.a. ``ets_post()`` to cause a ``delay()`` to return prematurely, weak bindings of ``delay()`` and friends don't matter. ``esp_schedule()`` can't know which of many concurrent delays needs to be shortcut, and the whole work-stealing multitasking breaks down :-(

dok-net:
I'd like to point out that the changes in this PR are contradictory and there's a lack of separation of concern. The change in ESP8266WiFiGeneric.cpp has nothing to do with the delay()/interruptable_delay() changes, as the use of delay(1) is simply made a bit less efficient due to the oneShotMs loop guard.

Also it's interruptible (with an I).

infrafast:
Dear all, I am in early stage to use ethernet driver in my esp8266 and it appears that I faced  some issue with the name resolution which takes much longer than via WIFI. Seems that the delay used in the "hostByName" function could be one of the cause. I propose to change it to a non blocking delay using the following patch:

diff --git a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cppba07e07 100644
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
@@ -632,7 +632,8 @@ int ESP8266WiFiGenericClass::hostByName(const char*
aHostname, IPAddress& aResul
          aResult = IPAddress(&addr);
      } else if(err == ERR_INPROGRESS) {
          _dns_lookup_pending = true;
-        delay(timeout_ms);
+        for (int i = 0; i < timeout_ms && !aResult.isSet(); i++)
+            delay(1);
          // will resume on timeout or when wifi_dns_found_callback fires
          _dns_lookup_pending = false;
          // will return here when dns_found_callback fires

MassiPi:
> Dear all, I am in early stage to use ethernet driver in my esp8266 and it appears that I faced some issue with the name resolution which takes much longer than via WIFI. Seems that the delay used in the "hostByName" function could be one of the cause. I propose to change it to a non blocking delay using the following patch:

Hello @d-a-v ,
sad to see this specific modification was delayed to 3.1, since this totally breaks the ethernet usability..
Is it possible to expect this pull request to be implemented before 3.1 (which now is 3%), or should i keep manually modifying locally the source files?
Thanks a lot

d-a-v:
@MassiPi I'm afraid we'll have to wait a bit more,
but you are right, now that new/experimental ethernet drivers are in the core, these delays have to be changed.
3.0.2 should be - I hope - the last regression-bugfix release after 3.0.0.

PR #6213, opened 2019-06-20
TCP connect and send delay fix

emelianov:
Implement early exit as connection established or data already sent.
(In my tests previous implementation was exiting only on timeout expired)


libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -456,8 +458,10 @@ class ClientContext
             }
 
             _send_waiting = true;
-            // This delay will be interrupted by esp_schedule on next received ack
-            delay(_timeout_ms);
+            for (decltype(_timeout_ms) i = 0; _send_waiting && i < _timeout_ms; i++) {
+               // let a chance to recurrent scheduled functions (ethernet)
```
d-a-v: I would also keep the old comments because delays can and will still be interrupted by `esp_schedule` and it needs to be clear in this code.
```
// This delay will be interrupted by esp_schedule on next received ack
```

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -603,6 +609,7 @@ class ClientContext
         (void) pcb;
         assert(pcb == _pcb);
         assert(_connect_pending);
+        _connect_pending = 0;
```
d-a-v: Thanks!

d-a-v:
reference for reviewers: https://github.com/d-a-v/W5500lwIP/issues/6#issuecomment-503522520

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -130,8 +130,11 @@ class ClientContext
         }
         _connect_pending = 1;
         _op_start_time = millis();
-        // This delay will be interrupted by esp_schedule in the connect callback
-        delay(_timeout_ms);
+        // Following delay will be interrupted by connect callback
+        for (decltype(_timeout_ms) i = 0; _connect_pending && i < _timeout_ms; i++) {
```
devyte: This looks like the same logic in polledTimeout. @d-a-v does it make sense to replace usage here?

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -130,8 +130,11 @@ class ClientContext
...
+               // let a chance to recurrent scheduled functions (ethernet)
```
devyte: Wording: Give scheduled functions a chance to run (e.g. Ethernet uses recurrent)

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -456,8 +459,11 @@ class ClientContext
             }
 
             _send_waiting = true;
-            // This delay will be interrupted by esp_schedule on next received ack
-            delay(_timeout_ms);
+            // Following delay will be interrupted by on next received ack
+            for (decltype(_timeout_ms) i = 0; _send_waiting && i < _timeout_ms; i++) {
```
devyte: Same as above: this looks like a polledTimeout

libraries/ESP8266WiFi/src/include/ClientContext.h
d-a-v: The original `delay()` is supposed to be interrupted by `esp_schedule()` anytime (after the original included `yield()`)
In this PR it is divided into small delays but the desired effect is the same. We want to continue as soon as a packet is received (which calls `esp_schedule()`) without latency.
(afaik PolledTimeout is not interruptible)

devyte: I was thinking of PolledTimeout with different policies than we currently have. But maybe it's not worth implementing for these cases.

ascillato:
Hi,

@emelianov @d-a-v @devyte @earlephilhower 

This PR makes the pubsubclient library to hung (device unresponsive) for 5 seconds if the MQTT broker is down and it is trying to reconnect.

Reverting just this commit makes the STAGE core to be fast again as core 2.5.2 while using MQTT.

I have just made the PR #6440 in order to fix this hung issue. Thanks.

PR #6214, opened 2019-06-20
 remove scheduled functions complexity overhead, change recurrent api

d-a-v:
*edit*:
* restore simplicity of scheduled function management
* no overhead when recurrent scheduled function are not used
* separate API

### DIFF
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -147,6 +147,7 @@ void run_scheduled_recurrent_functions ()
         // fence is like a mutex but as we are never called from ISR,
         // locking is useless here. Leaving comment for reference.
         //esp8266::InterruptLock lockAllInterruptsInThisScope;
+
         if (fence)
             // prevent recursive calls from yield()
             // (even if they are not allowed)
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -84,7 +84,7 @@ bool schedule_recurrent_function_us (const std::function<bool(void)>& fn, uint32
 {
     assert(repeat_us < decltype(recurrent_fn_t::callNow)::neverExpires); //~26800000us (26.8s)
 
-    InterruptLock lockAllInterruptsInThisScope;
+    esp8266::InterruptLock lockAllInterruptsInThisScope;
 
     recurrent_fn_t* item = new recurrent_fn_t(repeat_us);
     if (!item)
@@ -112,7 +112,7 @@ void run_scheduled_functions ()
         sFirst->mFunc();
 
         {
-            InterruptLock lockAllInterruptsInThisScope;
+            esp8266::InterruptLock lockAllInterruptsInThisScope;
 
             auto to_recycle = sFirst;
             sFirst = sFirst->mNext;
@@ -163,7 +163,7 @@ void run_scheduled_recurrent_functions ()
         if (current->callNow && !current->mFunc())
         {
             // remove function from stack
-            InterruptLock lockAllInterruptsInThisScope;
+            esp8266::InterruptLock lockAllInterruptsInThisScope;
 
             auto to_ditch = current;
 
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -58,7 +58,7 @@ static void recycle_fn_unsafe (scheduled_fn_t* fn)
 }
 
 IRAM_ATTR // (not only) called from ISR
-bool schedule_function (std::function<void(void)>&& fn)
+bool schedule_function (const std::function<void(void)>& fn)
 {
     InterruptLock lockAllInterruptsInThisScope;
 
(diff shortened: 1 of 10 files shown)


cores/esp8266/Schedule.cpp
```diff
@@ -6,25 +6,31 @@
...
+};
+
+static recurrent_fn_t* rFirst = nullptr; // fifo not needed
+
 IRAM_ATTR // called from ISR
-static scheduled_fn_t* get_fn_unsafe()
+static scheduled_fn_t* get_fn_unsafe ()
 {
     scheduled_fn_t* result = nullptr;
     // try to get an item from unused items list
```
earlephilhower: Very minor, but maybe the comment should be moved above the function?  get_fn_unsafe is a bit non-obvious to me.  Something like

````
// Returns a pointer to an unused sched_fn_t, or if none are available allocates a new one.
````

File: cores/esp8266/Schedule.cpp [hunk 2]
```diff
@@ -33,38 +39,33 @@ static scheduled_fn_t* get_fn_unsafe()
         result = sUnused;
         sUnused = sUnused->mNext;
         result->mNext = nullptr;
-        result->callNow.reset(esp8266::polledTimeout::periodicFastUs::alwaysExpired);
     }
     // if no unused items, and count not too high, allocate a new one
     else if (sCount < SCHEDULED_FN_MAX_COUNT)
     {
         result = new scheduled_fn_t;
```
earlephilhower: You may want to replace `new` with `malloc` only because `new` could throw an exception and this is being executed in an interrupts and it'll just crash at that point, I think.

cores/esp8266/Schedule.cpp
```diff
@@ -76,90 +77,99 @@ bool schedule_function_us(std::function<bool(void)>&& fn, uint32_t repeat_us, sc
...
     // Scheduled functions are removed only from this function, and
     // its purpose is that it is never called from an interrupt
     // (always on cont stack).
 
+    if (!rFirst)
+        return;
+
     static bool fence = false;
     {
         InterruptLock lockAllInterruptsInThisScope;
```
earlephilhower: I'm not seeing the need for IRQ lock here.  this is a local static variable and the only assignments are below, no?

cores/esp8266/Schedule.cpp
```diff
@@ -76,90 +77,99 @@ bool schedule_function_us(std::function<bool(void)>&& fn, uint32_t repeat_us, sc
...
-                    sFirst = sFirst->mNext;
-                else if (lastRecurring)
-                    lastRecurring->mNext = toCall->mNext;
+    recurrent_fn_t* prev = nullptr;
+    recurrent_fn_t* current = rFirst;
 
-                if (sLast == toCall)
-                    sLast = lastRecurring;
+    while (current)
+        if (current->callNow && !current->mFunc())
```
earlephilhower: I know the entire while() is a single very long if{} else{}, but could we add a bracket to the block since it is so long?

cores/esp8266/Schedule.cpp
```diff
@@ -76,90 +77,99 @@ bool schedule_function_us(std::function<bool(void)>&& fn, uint32_t repeat_us, sc
...
+        {
+            // remove function from stack
+            InterruptLock lockAllInterruptsInThisScope;
 
-                recycle_fn_unsafe(toCall);
-            }
+            auto to_ditch = current;
+            if (prev)
+                prev->mNext = current = current->mNext;
```
earlephilhower: Valid C, but maybe split this into two separate assignments so as to be explicit?

cores/esp8266/Schedule.cpp
```diff
@@ -76,90 +77,99 @@ bool schedule_function_us(std::function<bool(void)>&& fn, uint32_t repeat_us, sc
...
+            else
+                current = rFirst = rFirst->mNext;
```
earlephilhower: Ibid or ditto

earlephilhower:
LGTM.

cores/esp8266/Schedule.cpp
```diff
@@ -76,88 +80,110 @@ bool schedule_function_us(std::function<bool(void)>&& fn, uint32_t repeat_us, sc
...
-}
+    assert(repeat_us < decltype(recurrent_fn_t::callNow)::neverExpires); //~26800000us (26.8s)
 
-IRAM_ATTR // called from ISR
-bool schedule_function(std::function<void(void)>&& fn, schedule_e policy)
-{
-    return schedule_function_us([fn]() { fn(); return false; }, 0, policy);
+    esp8266::InterruptLock lockAllInterruptsInThisScope;
+
+    recurrent_fn_t* item = new recurrent_fn_t(repeat_us);
```
devyte: @d-a-v Elsewhere you changed new to malloc, but here new is still used. Is that on purpose?

d-a-v: Yes, on purpose, because `schedule_recurrent_function_us()` is never called from sys context.

dok-net:
About malloc() vs. new - is this about exceptions when there is no space on heap, or are there other reasons? After all, just reinterpreting uninitialized memory as std::function<void(void)> (the type of member scheduled_fn_t::mFunc) is undefined in the sense of the C++ type system.

devyte:
@dok-net thanks, the nothrow fix is already done

PR #6215, opened 2019-06-21
Unrelated includes, but necessary Arduino.h missing

dok-net:
The interrupts.h header includes a few headers that are not strictly needed, but the xt_rsil xt_wsr_ps are in Arduino.h, and that was not included.
Master compiles fine after cleanup.

For instance in EspSoftwareSerial, the unfixed interrupts.h forced inclusion of Arduino.h before interrupts.h, that's confusing to the human reader, this PR fixes that hidden prerequisite.


dok-net:
@d-a-v could you merge #6217 just as quickly, please?

PR #6217, opened 2019-06-21
interrupts.h in tests/host/common is empty

dok-net:


d-a-v:
`xt_rsil()` and `xt_wsr_ps()` are both emulated so the core file `core/esp8266/interrupts.h` should compile just fine in test/mock directories.

Well it is right this file is empty in `test/host/common/`, per this PR's title. I just removed it from my tree and run `make CI` in `test/host` and all seems fine. Can you try it ? Can we just remove it from the tree ?

dok-net:
@d-a-v If https://github.com/esp8266/Arduino/pull/6182 builds fine, I guess it's OK to remove it - does that mean the real `interrupts.h` gets used, which in turn is OK because `xt_rsil()` and friend are emulated elsewhere?

d-a-v:
>  the real interrupts.h gets used 

yes

>  because xt_rsil() and friend are emulated elsewhere?

[yes](https://github.com/esp8266/Arduino/blob/d93c7e06d7f1472e899d1bc1c0bef51a92e46485/tests/host/common/Arduino.h#L146)

dok-net:
@d-a-v #6182 passed all build tests, so it is as you said, removing the mock interrupts.h entirely is the easiest fix. I'm modifying this PR to do exactly that.

dok-net:
@d-a-v #6182 passed all build tests, so it is as you said, removing the mock interrupts.h entirely is the easiest fix. I modified this PR to do exactly that.

d-a-v:
Did you remove the file in this PR?
It seems it is still there

dok-net:
I removed it, it's not in my working directory anymore, and the "Files changed" says "empty file removed".

PR #6219, opened 2019-06-22
EspSoftwareSerial release 5.2.1

dok-net:
Using the std::bind workaround via FunctionInterrupt while attachInterruptArg is pending.


dok-net:
@earlephilhower After finding a one-liner way of back-porting from my attachInterruptArg PR(s), this is finally the long awaited release 5.2.1 of EspSoftwareSerial. I've you have a moment of time, could you please confirm to me how your basic tests perform? I'd expect noticeably improved performance, respectively lower error rates. I am quite curious what the handful of users that have issues with the ESP8266 and SoftwareSerial will be saying.

Issue #5711, opened 2019-02-02
mdns is not working STA and AP mode


sonukushwah221:
Hello 

I want to use mDns service with AP and STA mode but it's working fine in the AP mode but it's not working in the STA mode. please share the suggestions and solutions how we can access this mdns in both mode.

codebeat-nl:
1. Wrong question title, at least one mode works as stated in your question;
2. Are you sure you not confusing both? 

You need a bonjour (mDNS) service on the network (STA=Station mode, part of the network) and with AP (Access Point, own network) there is no bonjour service available so that's why it is not working. This is not an error.

d-a-v:
There's a possibility where mDNS can be used in AP mode, where a computer running bonjour or avahi is connected. ESP broadcasts/multicasts its mDNS name and the computer's daemon should be able to resolve this name.
We need to check or reproduce that issue.
- is mDNS broadcasting to 255.255.255.255 or multicasting ?
- is lwIP sending such broadcast/multicast packets on every up/enabled interfaces (ap & sta) ?

devyte:
@d-a-v I could be wrong, but I suspect that @sonukushwah221 request is to be able to use mdns in STA_AP mode on both interfaces at the same time. I haven't tested with the @LaborEtArs implementation, but at least the legacy class would work ether in STA mode, or in AP mode, or only on the STA interface when in STA_AP mode.

LaborEtArs:
Currently LEAmDNS will work in STATION OR AP mode, not in both. If both are activated, the AP netif is used as the multicast interface.

flav1972:
For me is not working in AP mode. When STA or both are enabled it is working.

d-a-v:
Is it possible for you to try PR #6224 ?

arantius:
I'm commenting on this open bug after finding at least #2602 and #6290 (already resolved)


1. Open Examples > ESP8266mDNS > mDNS_Web_Server
2. Type in SSID/PSK values.
3. Flash/upload.
4. mDNS works, device is discovered correctly (from LAN client, via STA address).
5. Change to `WiFi.mode(WIFI_AP_STA);` (and change the name so it's clear which version is running, by the mDNS answer).
6. Flash/upload.
7. mDNS fails, name never resolves.


[...]

Issue #5866, opened 2019-03-11
I can't use OTA update when ESP is an Access Point


OladapoAjala:
### Basic Infos
- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[latest git hash or date]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Nodemcu]
- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA]

### Problem Description
I can't perform Arduino OTA update on an ESP serving as the access point on my network. The IDE fails to detect the network port when my laptop is connected to the network created by my ESP module. OTA works fine when I set the ESP as a station and connect it to another network with my laptop connected to that network also.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>
#include <WiFiUdp.h>
#include <ArduinoOTA.h>

#ifndef STASSID
#define STASSID "my-ssid"
#define STAPSK  "my-password"
#endif

IPAddress apIP(192, 168, 1, 1);
const char* ssid = STASSID;
const char* password = STAPSK;

void setup() {
  Serial.begin(115200);
  Serial.println("Booting");
  
  WiFi.mode(WIFI_AP);
  WiFi.softAPConfig(apIP, apIP, IPAddress(255, 255, 255, 0));
  WiFi.softAP(ssid, password);

  ArduinoOTA.begin();
  Serial.println("Ready");
  Serial.print("IP address: ");
  Serial.println(WiFi.localIP());
}

void loop() {
  ArduinoOTA.handle();
}

bigFin:
[...]

PR #6224, opened 2019-06-24
mDNS: restriction to a single interface

d-a-v:
Default interface is STA (or AP if available and STA is unavailable).
An interface can also be specified in `::begin()` by its IP address.
MDNS will not cross interfaces (there is currently no notion of "bridged interfaces")

Multiple instances should be working, this is not tested in this commit.

### DIFF
--- a/cores/esp8266/AddrList.h
+++ b/cores/esp8266/AddrList.h
@@ -127,6 +127,7 @@ struct netifWrapper
     const char* ifhostname () const { return _netif->hostname?: emptyString.c_str(); }
     const char* ifmac () const      { return (const char*)_netif->hwaddr; }
     int ifnumber () const           { return _netif->num; }
+    bool ifUp () const              { return !!(_netif->flags & NETIF_FLAG_UP); }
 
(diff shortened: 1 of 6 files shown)


d-a-v:
@LaborEtArs @hreintke 

Should fix #5711

Lan-Hekary:
The Issue #5915 is resolved by this PR 
Also .. this kinda of Eliminate the Solution in this PR #5894
but sadly .. the original issue in #5866 is Back ..
esp8266.local .. works only in WIFI_AP_STA mode or WIFI_STA mode ... But NOT in WIFI_AP mode ..

ramiws:
> Default interface is STA (or AP if available and STA is unavailable).
> An interface can also be specified in `::begin()` by its IP address.
> MDNS will not cross interfaces (there is currently no notion of "bridged interfaces")
> 
> Multiple instances should be working, this is not tested in this commit.

 Multiple instances are not working... so we end up with no way for AP and STA working at the same time.

d-a-v:
@ramiws check #7213

Issue #6223, opened 2019-06-24
Conflicting interruptlock between core and dht library


supersjimmie:
I installed a fresh Arduino IDE, and added the latest esp8266 core library from git.
Then I added the DHT library, which seems to have a conflict with the core?

```
In file included from C:\Users\me\Documents\Arduino\IoT\IoT.ino:182:0:

C:\Users\me\Documents\arduino-1.8.9\libraries\DHT-sensor-library/DHT.h:67:7: error: redefinition of 'class InterruptLock'

 class InterruptLock {

       ^

In file included from C:\Users\me\Documents\arduino-1.8.9\hardware\esp8266com\esp8266\libraries\SoftwareSerial\src/circular_queue/circular_queue.h:28:0,

                 from C:\Users\me\Documents\arduino-1.8.9\hardware\esp8266com\esp8266\libraries\SoftwareSerial\src/SoftwareSerial.h:27,

                 from C:\Users\me\Documents\Arduino\IoT\IoT.ino:157:

C:\Users\me\Documents\arduino-1.8.9\hardware\esp8266com\esp8266\cores\esp8266/interrupts.h:19:7: error: previous definition of 'class InterruptLock'

 class InterruptLock {

       ^

```

d-a-v:
Our class name exists for long time but is only recently used
@dok-net (SoftwareSerial) will comment

I think we can rename this generic name in our core
depending on what other maintainers think of it.

In the meantime, you can manually/locally rename InterruptLock in DHT.h.
If it's an esp8266 specific library, you can also tell the author to check if he could use the core version of this class.

dok-net:
@d-a-v @supersjimmie There's a PR in Adafruits repo now: https://github.com/adafruit/DHT-sensor-library/pull/135

dok-net:
@supersjimmie Are you getting a working combination by default download methods already now? The closing integration of public release 5.2.2 via ESP8266 Arduino submodule libraries/SoftwareSerial is still awaiting, and without trying it all by myself, I expect the name conflict to continue until then?

devyte:
Closing because #6225 is merged.

PR #6225, opened 2019-06-24
Put InterruptLock (from interrupts.h) into namespace esp8266

dok-net:
Fix now and future collisions with 3rd party Arduino libraries (currently Adafruit DHT has this issue).


dok-net:
@d-a-v @supersjimmie This is the fix for #6223 done in ESP8266 Arduino core.

PR #6231, opened 2019-06-26
Expand gitignore to cover files created by Visual Micro

dok-net:
The Arduino IDE for Visual Studio in part uses the standard MS Visual Studio project layout.


dok-net:
@d-a-v Thank you, should have submitted this much sooner :-)

PR #6251, opened 2019-07-03
using new (std::nothrow) instead of malloc

d-a-v:


mcspr:
OT: Should this be documented? I only recently discovered that espasyncwebserver doing `some = new ... ; if (some)` is not the same thing as basic malloc on current Core; Nove not used updated toolchain very much because of platformio shipping the old one.
Might also be surprising for Arduino in general:
- AVR https://github.com/arduino/Arduino/pull/108 is ongoing
https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/new.cpp
- SAMD https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/new.cpp uses the same override

d-a-v:
It is useless when exceptions are disabled in the toolchain. 
But we have them as an option.

mcspr:
Yes, but my main question is about the null check routines. It does not trigger exception, but board reset still happens when internal malloc fails in low memory situations.

*async web server being the case where there are flow objects are constructed with new and checked for NULL afterwards. which spectacularly fails with 4KB-large context:sys stack trace*

PR #6261, opened 2019-07-06
MDNS: fix random crash on startup

### LINKED ISSUE #6463: MDNS Not working properly after this commit 9009b8
### Basic Infos

#### Platform

- Hardware:			         ESP-12
- Core Version:			85f1ea7c78a583b7a414f016a71d14c1c91099d5
- Operating System:		MacOS

### Settings in IDE

- Module:  			NodeMcu 1.0
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v2 Lower Higher Bandwidth]
- Upload Using:		SERIAL

### Problem Description
MDNS isn't working continuously after this commit https://github.com/esp8266/Arduino/commit/f9009b8a5e141235d3380cc6c6c06223008a1099. 
Problem Scenario -
After calling mdns.begin() once, it says ok. Now if a new device which is running a service discovery application, Joins the wifi network, it cannot find the .local service and cannot connect to the esp node using .local hostname. 
I am attaching a screenshot from my service discovery browser, which publishes a remove event for this esp node. 

This was working fine previous to this commit, i have verified that. 

```cpp

/*
  ESP8266 mDNS responder sample

  This is an example of an HTTP server that is accessible
  via http://esp8266.local URL thanks to mDNS responder.

  Instructions:
  - Update WiFi SSID and password as necessary.
  - Flash the sketch to the ESP8266 board
  - Install host software:
    - For Linux, install Avahi (http://avahi.org/).
    - For Windows, install Bonjour (http://www.apple.com/support/bonjour/).
    - For Mac OSX and iOS support is built in through Bonjour already.
  - Point your browser to http://esp8266.local, you should see a response.

*/


#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>
#include <WiFiClient.h>

#ifndef STASSID
#define STASSID "MYSSID"
#define STAPSK  "MYPASS"
#endif

unsigned long last_done = millis();
const char* ssid = STASSID;

// TCP server at port 80 will respond to HTTP requests
WiFiServer server(80);

void setup(void) {
  Serial.begin(115200);
  delayInit(2);

  // Connect to WiFi network
  WiFi.mode(WIFI_STA);
  connectToWifi();

  // Set up mDNS responder:
  // - first argument is the domain name, in this example
  //   the fully-qualified domain name is "esp8266.local"
  // - second argument is the IP address to advertise
  //   we send our IP address on the WiFi network
  if (!MDNS.begin("esp8266_special_xx")) {
    Serial.println("Error setting up MDNS responder!");
    while (1) {
      delay(1000);
    }
  }
  Serial.println("mDNS responder started");

  // Start TCP (HTTP) server
  server.begin();
  Serial.println("TCP server started");

  // Add service to MDNS-SD
  MDNS.addService("http", "tcp", 80);
}

void loop(void) {
  MDNS.update();

  // Check if a client has connected
  WiFiClient client = server.available();
  if (!client) {
    return;
  }
  Serial.println("");
  Serial.println("New client");

  // Wait for data from client to become available
  while (client.connected() && !client.available()) {
    delay(1);
  }

  // Read the first line of HTTP request
  String req = client.readStringUntil('\r');

  // First line of HTTP request looks like "GET /path HTTP/1.1"
  // Retrieve the "/path" part by finding the spaces
  int addr_start = req.indexOf(' ');
  int addr_end = req.indexOf(' ', addr_start + 1);
  if (addr_start == -1 || addr_end == -1) {
    Serial.print("Invalid request: ");
    Serial.println(req);
    return;
  }
  req = req.substring(addr_start + 1, addr_end);
  Serial.print("Request: ");
  Serial.println(req);
  client.flush();

  String s;
  if (req == "/") {
    IPAddress ip = WiFi.localIP();
    String ipStr = String(ip[0]) + '.' + String(ip[1]) + '.' + String(ip[2]) + '.' + String(ip[3]);
    s = "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\n\r\n<!DOCTYPE HTML>\r\n<html>Hello from ESP8266 at ";
    s += ipStr;
    s += "</html>\r\n\r\n";
    Serial.println("Sending 200");
  } else {
    s = "HTTP/1.1 404 Not Found\r\n\r\n";
    Serial.println("Sending 404");
  }
  client.print(s);

  Serial.println("Done with client");
}

```

d-a-v:
MDNS:
- AP address is used by default when STA is also present 
- enforce selected interface
- igmp: force selected interface

Arduino menu:
- mDNS debug option

*edit*: 
Fixes #6463


libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp
```diff
@@ -79,10 +79,8 @@ bool MDNSResponder::_process(bool p_bUserContext) {
         }
     }
     else {
-        bResult = ((WiFi.isConnected() ||               // Either station is connected
-                    WiFi.softAPgetStationNum()>0) &&    // Or AP has stations connected
-                   (_updateProbeStatus()) &&            // Probing
-                   (_checkServiceQueryCache()));        // Service query cache check
+        bResult = _updateProbeStatus() &&           // Probing
```
devyte: The first part of the condition in the current code is supposed to mean "there's somebody to send to", i. e. : if the station is connected to an ap, then there's at least one receiver OR if there's someone connected to the softap there's at least one receiver.
i think that removing that part of the condition makes MDNS operate needlessly.

d-a-v: It indeed does make mDNS operate even when there is no connection.
However mDNS is now supposed to be interface agnostic, bound to an IPAddress defining an interface instead of trying to find out which interface is up or to answer to. The only place where AP or STA should be dealt with is when the IP address from `begin()` is not set and a default interface has to be selected to keep with backward compatibility.

The first test `WiFi.isConnected()` checks wether STA is up, the second one counts the number of WiFi clients connected to the AP without really checking if AP is up. 

I propose to merge this PR asap so one can debug #5866 (@LaborEtArs and @hreintke will hopefully help on this) (and UDP is fixed), and I will propose a new function to check on interface status with an IP address passed as argument because we lack&need this functionality.

PR #6267, opened 2019-07-06
WIP - Move toolchain to GCC 7.2

earlephilhower:
Use at your own risk, but it compiles and run what I've thrown at it so far.  Needs to be examined much deeper to make sure things are still linked in the right spot, not duplicated, etc.


earlephilhower:
You also need to clear the DIST directory of tarballs before running this as the mkspiffs binary does not have the same checksum as the prior build, but is the same git (so same filename).

earlephilhower:
@d-a-v, I seem to be stuck between IPv6 and LWIP1.4.   `igmp_joingroup` takes either a `ip_addr_t` or a `ip4_addr_t` depending on the build, while `IP4_ADDR_ANY4` is a `u32`.

So it seems like the common code snippet `igmp_joingroup(IP4_ADDR_ANY4, ip_2_ip4(&multicast_addr))` can't be reconciled.

If I cast to `(ip_addr_t*)` it works in IPv4 for LWIP1.4 and 2, but fails in IPv6 (since there `ip_addr_t` can hold an IPv6 address).  If I cast to `ip4_addr_t*` then I can't build outside of IPv6.

Any ideas?  Is there any helper type with the same name that uses `ip_addr_t` when in v4 mode but `ip4_addr_t` in v6 mode?

d-a-v:
@earlephilhower here is the IPv6 fix (which is in fact an lwIP-1.4 IPv4 fix)
```
diff --git a/cores/esp8266/IPAddress.h b/cores/esp8266/IPAddress.h
index f4ed5a06..f78b0227 100644
--- a/cores/esp8266/IPAddress.h
+++ b/cores/esp8266/IPAddress.h
@@ -36,7 +36,7 @@
 #define IP_SET_TYPE_VAL(x,y) do { (void)0; } while (0)
 #define IP_ANY_TYPE (&ip_addr_any)
 #define IP4_ADDR_ANY IPADDR_ANY
-#define IP4_ADDR_ANY4 IPADDR_ANY
+#define IP4_ADDR_ANY4 IP_ADDR_ANY
 #define IPADDR4_INIT(x) { x }
 #define CONST /* nothing: lwIP-v1 does not use const */
 #define ip4_addr_netcmp ip_addr_netcmp
diff --git a/libraries/ESP8266SdFat b/libraries/ESP8266SdFat
--- a/libraries/ESP8266SdFat
+++ b/libraries/ESP8266SdFat
@@ -1 +1 @@
-Subproject commit 6326c71ff1f6294756b6e348edaf723bfe7be3e7
+Subproject commit 6326c71ff1f6294756b6e348edaf723bfe7be3e7-dirty
diff --git a/libraries/ESP8266mDNS/src/ESP8266mDNS_Legacy.cpp b/libraries/ESP8266mDNS/src/ESP8266mDNS_Legacy.cpp
index af39bff6..80f55ad8 100644
--- a/libraries/ESP8266mDNS/src/ESP8266mDNS_Legacy.cpp
+++ b/libraries/ESP8266mDNS/src/ESP8266mDNS_Legacy.cpp
@@ -198,7 +198,7 @@ bool MDNSResponder::_listen() {
 
     IPAddress mdns(MDNS_MULTICAST_ADDR);
 
-    if (igmp_joingroup((ip_addr_t *)IP4_ADDR_ANY4, mdns)!= ERR_OK) {
+    if (igmp_joingroup(IP4_ADDR_ANY4, mdns)!= ERR_OK) {
       return false;
     }
 
diff --git a/libraries/ESP8266mDNS/src/LEAmDNS_Helpers.cpp b/libraries/ESP8266mDNS/src/LEAmDNS_Helpers.cpp
index af3e13a2..d99a57f2 100644
--- a/libraries/ESP8266mDNS/src/LEAmDNS_Helpers.cpp
+++ b/libraries/ESP8266mDNS/src/LEAmDNS_Helpers.cpp
@@ -199,7 +199,7 @@ bool MDNSResponder::_allocUDPContext(void) {
     //TODO: set multicast address (lwip_joingroup() is IPv4 only at the time of writing)
     multicast_addr.addr = DNS_MQUERY_IPV6_GROUP_INIT;
 #endif
-    if (ERR_OK == igmp_joingroup((ip_addr_t*)IP4_ADDR_ANY4, ip_2_ip4(&multicast_addr))) {
+    if (ERR_OK == igmp_joingroup(IP4_ADDR_ANY4, ip_2_ip4(&multicast_addr))) {
         m_pUDPContext = new UdpContext;
         m_pUDPContext->ref();
```

earlephilhower:
We're all set, only binary replacement is needed to move to GCC 7.2 at some point in the future.

earlephilhower:
Abandoning in favor of going full-ahead to GCC 9.1.  The toolchain can be rebuilt via make in the future is we find some issues w/9.1 for the 3.x release.

PR #6272, opened 2019-07-08
Switch default FW to "2.2.2-dev(38a443e)" (menu:2.2.1+100)

d-a-v:
This is a convenience PR
* fix a LittleFS bug: avoid crash at boot when FS size is 0
* debug message in FS when implementation cannot start
* change defaults in arduino IDE menus

in details, new **default** options:

* ESP-NOW is working
* flash size:
    * 1M for generic board (instead of 512K)
    * FS: maximal not-0 size with maximal size for sketch+OTA


libraries/LittleFS/src/LittleFS.h
```diff
@@ -70,9 +70,9 @@ class LittleFSImpl : public FSImpl
         _lfs_cfg.read_size = 64;
         _lfs_cfg.prog_size = 64;
         _lfs_cfg.block_size =  _blockSize;
-        _lfs_cfg.block_count = _size / _blockSize;
-	_lfs_cfg.block_cycles = 16; // TODO - need better explanation
-	_lfs_cfg.cache_size = 64;
+        _lfs_cfg.block_count =_blockSize? _size / _blockSize: 0;
+        _lfs_cfg.block_cycles = 16; // TODO - need better explanation
```
earlephilhower: Space to tab problems?

tools/boards.txt.py
```diff
@@ -68,7 +68,7 @@
             'crystalfreq_menu',
             'flashfreq_menu',
             'flashmode_menu',
-            '512K', '1M', '2M', '4M', '8M', '16M',
+            '1M', '2M', '4M', '8M', '16M', '512K',
```
earlephilhower: I see the 512K moved to then end of lists in other spots, is this on purpose?

tools/boards.txt.py
```diff
@@ -542,7 +542,7 @@
             'flashfreq_40',
             '4M',
             ],
-        'serial': '57',
+      # 'serial': '57',
```
earlephilhower: Why disabling the hardcoded serial speeds?  If it's 921 it will silently use 400k.  If it's 56K, there might be a HW reason for it on those specific boards....

earlephilhower: (like a really, really cheap USB<->serial chip, or high resistance/capacitance on the lines due to something else like LEDs sharing the pin, etc.)

tools/boards.txt.py
```diff
@@ -1573,7 +1573,7 @@ def usage (name,ret):
 # entry point
 
 lwip = 2
-default_speed = '115'
+default_speed = '921'
```
earlephilhower: I'm a little worried here about the speed change, see above comment.

libraries/LittleFS/src/LittleFS.h
d-a-v: yes

tools/boards.txt.py
d-a-v: Yes, so the default flash space for esp8266 generic board is 1MB.
512K boards are obsolete and not sold anymore, also abandonned / not usable in nonos-sdk-v3.
We still have them anyway, but no more as default.

tools/boards.txt.py
d-a-v: That's true. I was [wrongly](https://github.com/espressif/esptool#baud-rate) thinking esptool would try at lower rates.
I will revert this.

tools/boards.txt.py
d-a-v: Speed change is now reverted.

ascillato:
@d-a-v @earlephilhower 

We have tested this new version of the core with SDK on Tasmota (on several devices at once) and it is working very good. Great work :+1: Congratulations.

Several Tasmota users were stuck at 2.3.0 core due was the most stable with some old wifi routers. Now this new version seems to be the new stable core for all types of routers!!! No more mqtt socket disconnections!

Are you planning to release a new version any time soon? That would be great!

Thanks a lot for your hard work

d-a-v:
> were stuck at 2.3.0 

Thanks, very good to hear you are finally able to use latests versions !

> Are you planning to release a new version any time soon? That would be great!

Currently 2.6.0 is planned for [September, 2](https://github.com/esp8266/Arduino/milestones).

Issue #6246, opened 2019-07-01
Calls made to malloc/free will enable interrupts on return


mhightower83:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:					ESP-12
- Core Version:			5a47cab77d396afeb842c9161327f9f4dca7c833
- Development Env:		Arduino IDE
- Operating System:		Ubuntu

### Settings in IDE

- Module:  			Adafruit Feather HUZZAH ESP8266
- Flash Mode:			qio
- Flash Size: 		4MB
- lwip Variant:   v2 Lower Memory
- Reset Method:		nodemcu
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200 (serial upload only)

### Problem Description

To simulate an IRQ handler calling malloc I used `xt_rsil()` to set INTLEVEL.
when malloc/free are called with an elevated INTLEVEL, INTLEVEL is set to 0 on return.

I noticed that `ets_intr_lock()` and `ets_intr_unlock()` are being used to handle critical section for umm_malloc. `ets_intr_unlock()` does not restore the previous INTLEVEL; it does an `rsil a2, 0`. 
Ref: https://github.com/esp8266/Arduino/issues/2218#issuecomment-230256371 Thus causing the immediate problem. 

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
#include <ets_sys.h>
#include <interrupts.h>

#ifndef xt_rsr_ps
#define xt_rsr_ps()  (__extension__({uint32_t state2; __asm__ __volatile__("rsr.ps %0" : "=a" (state2)); state2;}))
#endif

/* 
 * A printf routine that can safely print with interrupts disabled
 * and does no "malloc lib" calls. 
 * 
 * The default printf will do a malloc call, spoiling the test.
 */
 
// ROM _putc1, ignores CRs and sends CR/LF for LF, newline.
// Always returns character sent.
[...]

PR #6274, opened 2019-07-09
proposed umm_malloc improvements

mhightower83:
Addresses: https://github.com/esp8266/Arduino/issues/6246#issue-462531331

  Correct critical section with interrupt level preserving and nest support
  alternative. Replace ets_intr_lock()/ets_intr_unlock() with uint32_t
  oldValue=xt_rsil(3)/xt_wrs(oldValue). Added UMM_CRITICAL_DECL macro to define
  storage for current state. Expanded UMM_CRITICAL_... to  use unique
  identifiers. This helpt facilitate gather function specific  timing
  information.

  Replace printf with something that is ROM or IRAM based so that a printf
  that occurs during an ISR malloc/new does not cause a crash. To avoid any
  reentry issue it should also avoid doing malloc lib calls.

  Refactor realloc to avoid memcpy/memmove while in critical section. This is
  only effective when realloc is called with interrupts enabled. The copy
  process alone can take over 10us (when copying more than ~498 bytes with a
  80MHz CPU clock). It would be good practice for an ISR to avoid realloc.
  Note, while doing this might initially sound scary, this appears to be very
  stable. It ran on my troublesome sketch for over 3 weeks until I got back from
  vacation and  flashed an update. Troublesome sketch - runs ESPAsyncTCP, with
  modified fauxmo emulation for 10 devices. It receives lost of Network traffic
  related to uPnP scans, which includes lots of TCP connects disconnects RSTs
  related to uPnP discovery.

  I have clocked umm_info critical lock time taking as much as 180us. A common
  use for the umm_info call is to get the free heap result. It is common
  to try and closely monitor free heap as a method to detect memory leaks.
  This may result in frequent calls to umm_info. There has not been a clear
  test case that shows an issue yet; however, I and others think they are or
  have had crashes related to this.

  I have added code that adjusts the running free heap number from _umm_malloc,
  _umm_realloc, and _umm_free. Removing the need to do a long interrupts
  disabled calculation via _umm_info.

  Build optional, min/max time measurements for locks held while in info,
  malloc, realloc, and free. Also, maintain a count of how many times each is
  called with INTLEVEL set.


cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -519,8 +540,92 @@ extern int umm_last_fail_alloc_size;
...
+  // Preprocessor and compiler together will optimize away the if.
+  if (strcmp("Serial1", VALUE(DEBUG_ESP_PORT)) == 0) {
+    uart_buff_switch(1U);
+  } else {
+    uart_buff_switch(0U);
+  }
+#else
+  uart_buff_switch(0U); // Side effect, clears RX FIFO
+#endif
+  size_t str_len = ets_strlen(fmt);
```
earlephilhower: The format strings live in progmem, so ets_strlen() is going to bomb with an access error, I believe.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1223,6 +1338,9 @@ void ICACHE_FLASH_ATTR umm_init( void ) {
     /* index of the latest `umm_block` */
     const unsigned short int block_last = UMM_NUMBLOCKS - 1;
 
+    /* init heapFreeBlocks */
+    ummHeapFreeBlocks = block_last;
```
earlephilhower: Yay, O(1) here we are!

cores/esp8266/umm_malloc/umm_malloc.cpp
earlephilhower: The format strings live in progmem, so ets_strlen() is going to bomb with an access error, I believe.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1286,15 +1405,18 @@ static void _umm_free( void *ptr ) {
...
    */
 
-  /* Protect the critical section... */
-  UMM_CRITICAL_ENTRY();
-
   /* Figure out which block we're in. Note the use of truncated division... */
 
   c = (((char *)ptr)-(char *)(&(umm_heap[0])))/sizeof(umm_block);
 
+  /* Protect the critical section... */
```
earlephilhower: Maybe move this down to 1416 since the debug print isn't critical?

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1462,26 +1586,31 @@ static void *_umm_malloc( size_t size ) {
...
+    ummHeapFreeBlocks -= blocks;
+
   } else {
     /* Out of memory */
 
     DBG_LOG_DEBUG(  "Can't allocate %5d blocks\n", blocks );
 
     /* Release the critical section... */
-    UMM_CRITICAL_EXIT();
+    UMM_CRITICAL_EXIT(id_malloc);
```
earlephilhower: Same comment as above, maybe move back before the debug.  In this case, it's probably a fatal app error anyway, so not a big deal.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1600,6 +1751,13 @@ static void *_umm_realloc( void *ptr, size_t size ) {
 
     c = umm_assimilate_down(c, 0);
 
+    /*
+     * For the ESP8266 interrupts should not be off for more than 10us.
+     * An unprotect/protect around memmove should be safe to do here.
+     * All variables used are on the stack.
+     */
+    UMM_CRITICAL_EXIT(id_realloc);
+
```
earlephilhower: Only gotcha here is that the cached ummHeapFreeBlocks will be inconsistent should it be read during the memmove.  Probably not worth worrying about.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: It has been working for me. Before using ets_strlen I checked one of the assembly listings on the net for the ROM. It starts by verifying the pointer alignment. If it is aligned by 4, it drops down to the fast path of reading a word at a time, masking a byte out at a time and checking for 0. So unless I misread we should be safe as long as we start with a PROGMEM string that is aligned by 4. I just rechecked `tools/sdk/libc/xtensa-lx106-elf/include/sys/pgmspace.h` 
```
#define PSTR(s)            (__extension__({static const char __c[] __attribute__((__aligned__(4))) PROGMEM = (s); &__c[0];}))
```
I think this is relatively new. Let me know if I have errored in my thinking.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: I agree.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: This may be a style issue. I moved it before the `Out of memory` comment. So that the debug print is closest to the comment.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: Yea, I thought about that. I am thinking it is not a critical error and it saves a few bytes of IRAM by not addressing a (minor?) transient calculation error.

mhightower83:
Uhh, sorry, I may not have understood correctly, what I was asked there on that merge branch question?
Please disregard.

cores/esp8266/umm_malloc/umm_malloc.cpp
earlephilhower: You're correct there.  I added that alignment in 2.5.0 (IIRC) because I rewrote the _P routines in the std. lib to be much faster w/aligned data.  I didn't know ets_strlen had that optimization.  I guess that's a lucky coincidence on our part!

earlephilhower: Could you add a note to that effect, just for our editication?  There may be other, more memory constrained branches somewhere looking at our code w/o that alignment.  It'll crash for them, but at least they'd get a note in the source why it worked for is?

cores/esp8266/umm_malloc/umm_malloc.cpp
earlephilhower: The name is fine by me.  I've seen camelCase, PascalNames, under_scores, and ijk in the codebase.  The IRQ unsafeness, fine by me too.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: After further thought, I am thinking this comment before ets_strlen would be good:
```
  /* 
    To use ets_strlen() and ets_memcpy() safely with PROGMEM, flash storage,
    the PROGMEM address must be word (4 bytes) aligned. The destination
    address for ets_memcpy must also be word-aligned. We also round the
    buf_len up to the nearest word boundary. So that all transfers will be
    whole words. 
  */
```

mhightower83:
@earlephilhower - oh you got that already. I didn't think I had finished?
Sorry, I am really new to using GitHub with these updates using push/pull, etc to handle follow-ups.   

When I a have an established pull request from say branch pr3 on my fork of arduino/esp8266 to master branch on https://github.com/esp8266 and I the push an update to my branch pr3 on GitHub, am I done? 

No need to redo the push to https://github.com/esp8266?

Did I need to do the merge of https://github.com/esp8266 master back to my pr3 on GitHub?

devyte:
@mhightower83 you seem to be doing this correctly:
- work in a branch in ykur forknof this repo
- commit locally in your workspace
- push to the branch in your repo once done
- visit github (either yourbfork or here) and push the green button to make the PR
- If you need to update, just lush tonyour branch in your reoi, and the PR gets automagically updated
- we can update the branch from the web interface. This can cause your local workspace to become unsynced. In that case you have to do special stuff to pull latest changes and merge them into your local workspace.

mhightower83:
@devyte Thanks for the info, I get it now.

mhightower83:
I think we may need to add memory barrier to `xt_rsil()`. 
Based on the comments I saw here:

From `ESP8266_RTOS_SDK/components/esp8266/include/xtensa/xtruntime.h`

```cpp
/*  NOTE: these asm macros don't modify memory, but they are marked
 *  as such to act as memory access barriers to the compiler because
 *  these macros are sometimes used to delineate critical sections;
 *  function calls are natural barriers (the compiler does not know
 *  whether a function modifies memory) unless declared to be inlined.  */
# define XTOS_SET_INTLEVEL(intlevel)            ({ unsigned __tmp; \
                        __asm__ __volatile__(   "rsil   %0, " XTSTR(intlevel) "\n" \
                                                : "=a" (__tmp) : : "memory" ); \
                        __tmp;})
```
In my first unseen fix to umm_malloc I used `XTOS_SET_MIN_INTLEVEL` similar to the above; however, 
based on @devyte comment:
> ... The core currently doesn't support masking a specific interrupt level. That doesn't mean it can't be implemented, only that at this time locking is implemented as all or nothing.

I switched it over to use `xt_rsil()`/`xt_wsr_ps()` for the proposed fix. 

I went back to try and understand better, the significance of the "memory" parameter. Then I realized that the current `xt_rsil()` in
https://github.com/esp8266/Arduino/blob/cd6cf984ec2b61667e419db2d2c82a1a2fad3c52/cores/esp8266/Arduino.h#L162
might need it.

I wanted to see the results in assembly. So I did a quick compile to .S of this sample function with the old `xt_rsil()` and one with "memory" specified:
```cpp
#include <Arduino.h>

#if 0
#undef xt_rsil
#define xt_rsil(level) (__extension__({uint32_t state; __asm__ __volatile__("rsil %0," __STRINGIFY(level) : "=a" (state)::"memory"); state;}))
#endif

int just4Fun(int *a, int *b) {
  uint32_t old_ps;
  int result = *a;
  old_ps = xt_rsil(15);
  *b = *a;
  xt_wsr_ps(old_ps);
  return result;
}
```
Using `xt_rsil()` without "memory" fence, trimmed .S results:
```S
_Z8just4FunPiS_:
        l32i.n  a2, a2, 0
        rsil a4,15
        s32i.n  a2, a3, 0
        wsr a4,ps; isync
        ret.n
```
Using `xt_rsil()` with "memory" fence, trimmed .S results:
```S
_Z8just4FunPiS_:
        l32i.n  a4, a2, 0
        rsil a5,15
        l32i.n  a2, a2, 0
        s32i.n  a2, a3, 0
        wsr a5,ps; isync
        mov.n   a2, a4
        ret.n
```

Notice the change. 
 * The first example reads ` *a` outside the critical area. Then saves that value to `*b` while in the critical area. 
 * The second example reads a copy of `*a` outside the critical area. However, while in the critical area it reads `*a` again before saving it to `*b`.

mhightower83:
The change to xt_rsil() needs to be copied into to umm_malloc_cfg.h. Is there a better way to handle this? At the moment, when Esp-frag.cpp builds, a warning is issued so you know the mismatch exists. Which is probably better than renaming the macro locally in umm_malloc_cfg.h and not know it mismatches.

I also have some comment corrections.

@devyte If I do a push will it create any disturbance with your review?

devyte:
Push away :) I tend to do multiple review passes anyways  because I don't always see everything at first

earlephilhower:
@devyte, it looks like @mhightower83 is next in line for merge, once he gets the change he was talking about included.

mhightower83:
@earlephilhower  The changes I described were there when you merged master into pr3, 1 hour ago. @devyte indicated he had a lot of comments. Based on that, I'll need to do at least one more push after he reviews.

d-a-v:
> The change to xt_rsil() needs to be copied into to umm_malloc_cfg.h. Is there a better way to handle this?

Yes. Please move, in this PR, `xt_rsil&co` definitions out from `Arduino.h`, say, to already existing `cores/esp8266/core_esp8266_features.h`.

Including `Arduino.h` just to get these macros available is painful because of include file dependencies, in `umm` and also everywhere else.

mhightower83:
@d-a-v  Looks like xt_rsil/xt_wsr_ps move trips on the stuff in the Travis ./common/mock.h file.

cores/esp8266/core_esp8266_features.h
```diff
@@ -33,7 +33,30 @@
...
+//    uint32_t savedPS = xt_rsil(1); // this routine will allow level 2 and above
+//    // do work here
+//    xt_wsr_ps(savedPS); // restore the state
+//}
+//
+// level (0-15), interrupts of the given level and above will be active
+// level 15 will disable ALL interrupts,
+// level 0 will enable ALL interrupts,
+//
+#define xt_rsil(level) (__extension__({uint32_t state; __asm__ __volatile__("rsil %0," __STRINGIFY(level) : "=a" (state) :: "memory"); state;}))
```
d-a-v: About, host tests or emulation on host, I think this is sufficient around ps state and cycle count:
```
#ifndef CORE_MOCK
...
#endif // not CORE_MOCK
```

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -498,9 +498,30 @@
 #include "umm_malloc.h"
 
 #include "umm_malloc_cfg.h"   /* user-dependent */
+#if defined(DEBUG_ESP_PORT)
```
devyte: This is specific to the ESP. It would not be accepted into umm upstream. I suggest moving this elsewhere, e.g. the config file.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -498,9 +498,30 @@
...
+#include <assert.h>
+#endif
 
 extern "C" {
 
+#ifdef UMM_CRITICAL_PERIOD_ANALYZE
```
devyte: There is currently no assurance that these performance functions will be accepted upstream. We can propose them anyways, but it could take a while for the code to be reviewed and tested.
Instead, I suggest moving into a umm performance.h|cpp files, including the .h in the config file, and adjusting the build as/if necessary for the extra .cpp.
The above is possible thanks to the use of the DECL, ENTRY and EXIT macro hooks.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -498,9 +498,30 @@
...
+static struct _UMM_TIME_STATS time_stats = {
```
devyte: Suggest moving this static global to another .cpp, as previously explained.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -498,9 +498,30 @@
...
+  {0xFFFFFFFF, 0U, 0U, 0U},
+  {0xFFFFFFFF, 0U, 0U, 0U},
+  {0xFFFFFFFF, 0U, 0U, 0U},
+  {0xFFFFFFFF, 0U, 0U, 0U} };
+
+bool ICACHE_FLASH_ATTR get_umm_get_perf_data(struct _UMM_TIME_STATS *p, size_t size) {
```
devyte: Suggest moving this function to another .cpp, as previously explained.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -519,8 +540,102 @@ extern int umm_last_fail_alloc_size;
...
+  _umm_realloc, and _umm_free. Removing the need to do a long interrupts
+  disabled calculation via umm_info.
+
+  Build optional, min/max time measurements for locks held while in info,
+  malloc, realloc, and free. Also, maintains a count of how many times each is
+  called with INTLEVEL set.
+
+ */
+
+#if defined(DEBUG_ESP_PORT) || defined(DEBUG_ESP_ISR)
```
devyte: Suggest moving this entire if defined block to another .cpp, as previously explained.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -646,6 +761,10 @@ unsigned short int umm_numblocks = 0;
 #define UMM_PFREE(b)  (UMM_BLOCK(b).body.free.prev)
 #define UMM_DATA(b)   (UMM_BLOCK(b).body.data)
 
+/*
```
devyte: This should likely be pursued upstream at the umm repo.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1084,6 +1205,10 @@ void ICACHE_FLASH_ATTR *umm_info( void *ptr, int force ) {
       ummHeapInfo.maxFreeContiguousBlocks = curBlocks;
     }
   }
+#if defined(DEBUG_ESP_PORT)
```
devyte: This is specific to the ESP, and won't be accepted as-is upstream.I'm not sure what would be the best approach, maybe discussing asserts with the original author should be done first.

mhightower83: The asserts were for initial verification of the new method for tracking freeheap. It compared the new with the original method. I think it can now be removed. As well as the include <assert.h>

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: As described later this can be deleted.

cores/esp8266/umm_malloc/umm_malloc_cfg.h
```diff
@@ -110,8 +111,79 @@ extern char _heap_start[];
...
+ *
+ * Build option to collect timing usage data on critical section usage in
+ * functions: info, malloc, realloc. Collects MIN, MAX, and number of time
+ * IRQs were disabled at request time. Note, for realloc MAX disabled time
+ * will not include the time from calling malloc and/or free.
+ * Examine code for specifics on what info is available and how to access.
+*/
+// #define UMM_CRITICAL_PERIOD_ANALYZE
+
+#if !defined(UMM_CRITICAL_PERIOD_ANALYZE)
```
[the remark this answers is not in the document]
mhightower83: okay

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: Okay

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: okay

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: okay

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: okay

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: I made this comment in case somebody understood better what was going on. It doesn't look like it is being built. Don't know if it is some kind of offline test code. I wondered if maybe it predated the addition of the critical section. I added a note in case someday somebody tries to turn it on.

devyte: I understand. My point is that we don't own this code, it is a copy taken from the repo of the original author, so making a comment in our copy doesn't help, unless the author were to abandon the code and we were to take ownership of our copy. 
The comment should be made at the author's repo instead, so that it can be addressed.
All of this doesn't mean remove this, just that there is additional action that needs to be taken.

mhightower83:
Oops, forgot to comment out `#define UMM_CRITICAL_PERIOD_ANALYZE` I'll fix it on the next round.

cores/esp8266/umm_malloc/umm_performance.cpp
```diff
@@ -0,0 +1,78 @@
...
+  uart_buff_switch(0U); // Side effect, clears RX FIFO
+#endif
+  /*
+    To use ets_strlen() and ets_memcpy() safely with PROGMEM, flash storage,
+    the PROGMEM address must be word (4 bytes) aligned. The destination
+    address for ets_memcpy must also be word-aligned. We also round the
+    buf_len up to the nearest word boundary. So that all transfers will be
+    whole words.
+  */
+  size_t str_len = ets_strlen(fmt);
```
devyte: @earlephilhower is there any code from your progmem functions that could be reused here?

cores/esp8266/umm_malloc/umm_performance.cpp
```diff
@@ -0,0 +1,78 @@
...
+  Printing from the malloc routines is tricky. Since a lot of library calls
+  will want to do malloc.
+
+  Objective:  To be able to print "last gasp" diagnostic messages
+  when interrupts are disabled and w/o availability of heap resources.
+*/
+
+// ROM _putc1, ignores CRs and sends CR/LF for LF, newline.
+// Always returns character sent.
+int constexpr (*_rom_putc1)(int) = (int (*)(int))0x40001dcc;
```
devyte: @earlephilhower what do you think of moving this to a .ld sceipt, like the others [here](https://github.com/esp8266/Arduino/blob/master/tools/sdk/ld/eagle.rom.addr.v6.ld)?

earlephilhower: A `PROVIDES` line in https://github.com/esp8266/Arduino/blob/master/tools/sdk/ld/eagle.rom.addr.v6.ld is probably the best place, yes.  There would also need to be a line in a header somewhere since this is an "undocumented" ROM function.

cores/esp8266/umm_malloc/umm_performance.cpp
earlephilhower: No, he's actually fine here since PSTRs are now 4-byte aligned and the ROM uses an optimized 4-char at a time algorithm.

mhightower83:
Since we touched `get_cycle_count()` here, I want to revisit it.
Should `esp_get_cycle_count()` have a _memory fence_? 
The original core definition didn't have one. My use case didn't need it; however, I am undecided whether it should have one. 
Maybe unrelated, @earlephilhower indicates an issue here that was resolved by optimization change.
https://github.com/esp8266/Arduino/blob/2b9a69b42f2268e1f2752ef345f988c62b17ec9a/cores/esp8266/core_esp8266_waveform.cpp#L149-L155

devyte:
Pending now: check the include chain

cores/esp8266/umm_malloc/umm_performance.cpp
devyte: We can do that in a separate PR.

earlephilhower:
> Since we touched `get_cycle_count()` here, I want to revisit it.
> Should `esp_get_cycle_count()` have a _memory fence_?
> The original core definition didn't have one. My use case didn't need it; however, I am undecided whether it should have one.
> Maybe unrelated, @earlephilhower indicates an issue here that was resolved by optimization change. 
> https://github.com/esp8266/Arduino/blob/2b9a69b42f2268e1f2752ef345f988c62b17ec9a/cores/esp8266/core_esp8266_waveform.cpp#L149-L155

I don't have a good pipeline understanding, but for the use in this spot I would say it doesn't need the fence.  Reasoning is that, assuming mem fence just stalls the ID pipeline until writes complete, then we're talking about 1-10 80MHz counts noise returned from the function.  For the Waveform, that's plenty good precision and I'd rather use 1-10 fewer cycles for other work than waiting to get a 100.000% accurate result.  Practically, there's not been any issue with the current config.

There was also discussion about the global version of the macro.  A simple `#define get_cycle_count __asm__ (... : : : )` could replace it and avoid any issues about inline/linker section compatibility.

TD-er:
> I don't have a good pipeline understanding, but for the use in this spot I would say it doesn't need the fence. Reasoning is that, assuming mem fence just stalls the ID pipeline until writes complete

A memory fence is not some synchronization, but only telling the compiler not to perform out-of-order optimizations of code crossing that fence.
Since the call to that function is inline, I can understand the question of whether to use a memory fence. Although I do not have a clear answer to it ;)

If that's not what you meant, I am very curious to what you did mean, since then I didn't understand your remark :)

mhightower83:
I believe a memory barrier can have a hardware component (a bigger issue for multi-core); however, for the humble ESP8266, I don't believe this is a concern. I was thinking in terms of compile-time memory barriers. After a lot of searching, this is the best article I have seen (so far) on the topic, [Memory Ordering at Compile Time](https://preshing.com/20120625/memory-ordering-at-compile-time/). 

For a deeper discussion which gets into more multiprocessor/caching issues, this article appears to cover it well (TL:DR all of it) [LINUX KERNEL MEMORY BARRIERS](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/memory-barriers.txt?id=HEAD).

earlephilhower:
Got it, @TD-er and @mhightower83.  I come from CPU design and for me "memory barrier" == a HW interlock (memw/etc.) to ensure the load/store pipeline is in a clean state.  A GCC memory barrier is what you meant, I get it now.

If you think it makes a difference, a PR on the waveform generator's always welcome.  Given the organization of the code (it grabs and compares the cycle count before doing any state-changing) in the waveform generator I'm not sure how GCC could make a representation that would cause an issue, but there's no harm in checking.  I am not using it for a busy-loop wait, and when it was being developed I was regularly checking the performance for microsecond-range pulses and didn't notice anything.  A quick check of the generated assembly w/ and w/o would be instructive.

TD-er:
Well, as far as I know, inline functions are just "copied" into the code where the function is called.
This also implies the compiler might see it fit to optimize the code as seen in the entire function where it is being called from.
If my understanding of this is incorrect please tell me.

earlephilhower:
Sure, like cut-n-pasting like you said.  CSE and other optimizations can happen then.  My statement was that you can't move an assignment past a conditional dependent on it (we're not the Itanic :) ).  The main IRQ goes like:  `now=get-cycle-count; forall active-waveform-pins: if (now>trigger[pin]) do-toggle-stuff`...

We're probably getting down a rathole not related to UMM here, and this is already a long thread.  Any update @devyte?  I don't understand the concern (infinite loop of includes or something?)

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: I didn't think about the case of an ISR doing a malloc during this period and didn't realize the assimilate up and down could capture near all of the memory available. (I saw only 56 bytes remaining in my case when I was trying to capture a heap low watermark) Looks to me of the three places I tried to optimize realloc for lock time only one is viable.  :(

earlephilhower:
@devyte, I think you're up again on this one.  Don't want it to fade off the first page of PRs...

cores/esp8266/umm_malloc/umm_performance.h
```diff
@@ -0,0 +1,98 @@
...
+
+    xt_wsr_ps(*saved_ps);
+}
+#endif
+
+#if defined(DEBUG_ESP_PORT) || defined(DEBUG_ESP_ISR)
+int _isr_safe_printf_P(const char *fmt, ...) __attribute__((format(printf, 1, 2)));
+#endif
+
+// Use _UMM_MALLOC_CPP to limit access to umm_malloc.cpp module specific
```
earlephilhower: Lines 81-90 should probably go in `umm_malloc.cpp`, with the `#define _umm_malloc_cpp` removed from the file, too.  I believe that's what's throwing @devyte off.  Since you're just trying to make `printf` safe inside that one file, there's no reason to export it here under `ifdef` protection...

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -495,6 +495,7 @@
 #include <string.h>
 #include <pgmspace.h>
 
+#define _UMM_MALLOC_CPP
```
earlephilhower: Drop this define, as mentioned elsewhere.

cores/esp8266/umm_malloc/umm_performance.h
earlephilhower: Alternatively, these lines w/o the ifdef bracketing could go in the UMM private `umm_malloc_cfg.h` file, like the old `#define printf()->PSTR()`

devyte: 2nd choice sounds better. We're trying to avoid adding unnecessary code to files coming from upstream like umm_malloc.cpp to reduce divergence.

mhightower83: > Alternatively, these lines w/o the ifdef bracketing could go in the UMM private `umm_malloc_cfg.h` file, like the old `#define printf()->PSTR()`

Actually, the old `#define printf` was in `umm_malloc.cpp` moving all of the print stuff out seem to create the complications.

Also, the UMM private file `umm_malloc_cfg.h` turned out not to be so private. It is included in `umm_malloc.h` which in turn created issues when compiling `Esp.cpp` and maybe more I don't remember.

I made the changes as you requested compiles fine and a quick run with umm_info() works.
And I have done a git push.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: Done.

cores/esp8266/umm_malloc/umm_performance.h
mhightower83: > 2nd choice sounds better. We're trying to avoid adding unnecessary code to files coming from upstream like umm_malloc.cpp to reduce divergence.

I have been looking at the upstream version. It makes better use of a macro `DBGLOG_FUNCTION` for doing printf; however, there are a few places that are still prinf. Finishing the replacement of all printf's with `DBGLOG_FUNCTION` will make things a lot easier. Then there is no problem with `#define` creating conflicting printf definitions. But that is in the future.

mhightower83:
travis failed. @earlephilhower any thoughts on what is needed here:
```bash
Installing APT Packages
35611.72s$ travis_apt_get_update
3570.45s$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install g++-7 gcc-7
358Reading package lists...
359Building dependency tree...
360Reading state information...
361E: Unable to locate package g++-7
362E: Couldn't find any package by regex 'g++-7'
363E: Unable to locate package gcc-7
apt-get.diagnostics
364apt-get install failed
469The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install g++-7 gcc-7" failed and exited with 100 during .
```

earlephilhower:
Looks like a transient Travis.ci problem.  I've restarted the job.

mhightower83:
https://github.com/esp8266/Arduino/issues/6366#issuecomment-519983744
> * WiFi disconnected/in promiscuous mode, no heap fragmenter - umm_malloc calls panic() because one of it's structures fails the integrity check - mask wrong @ umm_malloc.cpp:736.

@devyte @d-a-v @earlephilhower there are some diagnostic areas of the _umm_malloc_ code that I am noticing and have concerns with:
1. UMM_INTEGRITY_CHECK
  * doesn't use critical sections when scanning the heap and neither does the upstream version. The issue I see here is the heap entries changing by an ISR doing an _alloc_ call, while integrity is running.
  * if/when critical sections are added this may push past the 10us interrupts disable guideline.
  * currently this requires editing files to enable, not a menu option which makes it less likely to be used. Therefore less likely to cause someone a problem.
2. UMM_POISON (UMM_POSION_CHECK)
 * Two functions are enabled.
    1. CHECK_POISON_ALL_BLOCKS() - A full heap check of all active allocations.
       * doesn't use critical sections when scanning the heap and neither does the upstream version
       * if/when critical sections are added this may push past the 10us disabled IRQs guideline.
       * Is enabled when debug serial is selected. (I didn't realize this one was enabled here.)
    2. GET_UNPOISONED() - Individually check the current allocation presented through a `realloc()` or `free()`.
       * This doesn't need a critical section, because our allocation cannot change w/o our action.
       * Is enabled when debug serial is selected.
       * I think this is a good option. It has lower overhead less likely to exceed 10us.
3. CHECK_POISON_ALL_BLOCKS() is done before INTEGRITY_CHECK() I think it should be reversed. INTEGRITY_CHECK() verifies that links are good. I think it would be better to do that before CHECK_POISON_ALL_BLOCKS() goes confidently walking all the links checking for memory oversteps. 
4. Using the _newlib_ `printf()` function may be a problem if the heap is damaged and it does a malloc to handle any printing. Also, an issue is if the error occurs from an ISR call, `printf()` is in flash.

At this time I currently have no measurements on the time to perform INTEGRITY_CHECK() or CHECK_POISON_ALL_BLOCKS() I guess I'll add that to my TODO list.

I am thinking that only the GET_UNPOISONED() option should be enabled when just selecting debug serial port. The CHECK_POISON_ALL_BLOCKS() should take more work to turn on. Some of these more aggressive diagnostics could start creating problems, especially when critical sections are added.

Sigh, I am never going to be though with this PR :face_with_head_bandage:

devyte:
CC @rhempel for @mhightower83 's concerns

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1600,6 +1731,29 @@ static void *_umm_realloc( void *ptr, size_t size ) {
...
 
+    /*
+     * Currently all or most of the heap has been grabbed. Do an early split of
+     * allocation down to the amount needed to do a successful move operation.
+     * This will allow an alloc/new from a ISR to succeed while a memmove is
+     * running.
+     */
+    if( (UMM_NBLOCK(c) - c) > prelimBlockSize ) {
+        umm_make_new_block( c, prelimBlockSize, 0, 0 );
+        _umm_free( (void *)&UMM_DATA(c+prelimBlockSize) );
```
devyte: Isn't this call to umm free() inside the umm realloc critical section? That would mean there is still nesting present.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1610,9 +1764,12 @@ static void *_umm_realloc( void *ptr, size_t size ) {
     /* And don't forget to adjust the pointer to the new block location! */
 
     ptr    = (void *)&UMM_DATA(c);
+
+    /* Now resume critical section... */
+    UMM_CRITICAL_ENTRY(id_realloc);
   }
```
devyte: what if the above memmove gets interrupted by an ISR that happens to write to the mem chunk being moved?

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -1610,9 +1764,12 @@ static void *_umm_realloc( void *ptr, size_t size ) {
...
 
-  /* Now calculate the block size again...and we'll have three cases */
+  // Now calculate the block size again...and we'll have three cases /
```
devyte: We're compiling this as. cpp, but upstream is still. c. Please keep original comment style.

mhightower83: Huh, Don't know how that happened. And I missed it with all the diffs that I do to remove accidents.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: As a shared resource that memory allocation may become corrupted. 

An ISR that chooses to do `realloc` or `free` has to implement some locking method to ensure exclusive access when doing these operations, Similar to what a multithreaded APP would have to do with a shared resource.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: Yes, I didn't like that either; however, if we unlock outside the critical section, and we take an interrupt. Then, that ISR could try and malloc memory and get an OOM result because at this moment we have claimed all the free memory above and below us. Which might be most of the memory that is left, depending on fragmentation. So I had to move the `free` just before the critical section exit.

And of course, the improved critical section supports nesting.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: An option that could be added, would be to define two new critical section macros something like:
```cpp
 #ifdef UMM_LIGHTWEIGHT_CPU
 #define UMM_CRITICAL_SUSPEND(tag) UMM_CRITICAL_EXIT(tag)
 #define UMM_CRITICAL_RESUME(tag) UMM_CRITICAL_ENTRY(tag)
 #else
 #define UMM_CRITICAL_SUSPEND(tag)
 #define UMM_CRITICAL_RESUME(tag)
 #endif
```
These would replace the newer `UMM_CRITICAL_EXIT` and `UMM_CRITICAL_ENTRY` that were added to enable IRQs for memcpys. This would allow for easy backward compatibility.

mhightower83: Hmm, a worse scenario would be the foreground APP is writing to memory and the writing process is interrupted and the ISR realloc(s) space and the space the foreground is using goes away or is then owned by a different ISR due to two IRQs back to back each using realloc/malloc.  When the call returns the foreground code is writing on someone else's memory. I think this gets back to the foreground and ISRs have to coordinate shared resources through some form of lock. -- Oh how I miss `interlockedExchange()`.

cores/esp8266/umm_malloc/umm_malloc.cpp
devyte: I don't think it's necessary to move the critical section locks to remove the nesting, but rather refactor _umm_free into two functions: a clre function that does no locking or arg checking, and a wrapper function that checks the args and locks before calling the core and unlocks afterwards. Then, this call here would be replaced with the core call, i. e. no locking and hence no nested critical sections.

PR #6287, opened 2019-07-11
Prepare signing before sketch is compiled

qistoph:
This makes the right value available in ARDUINO_SIGNING in the sketch.

The order of compilation makes the actual value of the define `ARUDINO_SINGING` inconsistently available to users. It might come in handy to check from user code if signature validation is enabled. This PR changes the moment in the build process the `Updater_Signing.h` file is updated.

Old order of build process:
1. The core contains a place holder file "Updater_Signing.h" with `#define ARDUINO_SIGNING 0`, this file is always readily available.
2. **Compile sketch**
3. Compile libraries
4. **Enabling binary signing** - generates "Updater_Signing.h" with `#define ARDUINO_SIGNING 1`.
5. Compile core

If the sketch is build and no cache is available, `ARDUINO_SIGNING` will be `0` during sketch compilation (step 2). However, if the sketch is build again and a cache of the core is available, `ARDUINO_SIGNING` will be `1`. This makes the resulting binary inconsistent, even though nothing's changed in the code.

Order after this PR:
1. The core contains a place holder file "Updater_Signing.h" with `#define ARDUINO_SIGNING 0`, this file is always readily available.
2. **Enabling binary signing** - generates "Updater_Signing.h" with `#define ARDUINO_SIGNING 1`.
3. **Compile sketch**
4. Compile libraries
5. Compile core

Now the `ARDUINO_SIGNING` is already set to `1` in step 2, before compilation of the sketch.

[Example sketch demonstrating ARDUINO_SIGNING usage](https://github.com/esp8266/Arduino/files/3382829/ESP8266_PR_Sample.txt).

**Essential part of the example**
```
  server.on("/", HTTP_GET, [&]() {
    String content;
    content += F("<html><body>");
    content += F("Signing is ");
    content += (ARDUINO_SIGNING ? F("enabled") : F("disabled"));
    content += F("<br/>");

    if(ARDUINO_SIGNING) {
      content += F("<a href=\"/pubkey\">Public key</a>");
    }

    content += F("</body></html>");
    server.send(200, PSTR("text/html"), content);
  });

  #if ARDUINO_SIGNING
  server.on("/pubkey", HTTP_GET, [&]() {
    server.send_P(200, PSTR("text/plain"), signing_pubkey, sizeof(signing_pubkey));
  });
  #endif
```

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -83,8 +83,8 @@ compiler.elf2hex.extra_flags=
 
 ## generate file with git version number
 ## needs bash, git, and echo
-recipe.hooks.sketch.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
-recipe.hooks.core.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.makecorever}" --build_path "{build.path}" --platform_path "{runtime.platform.path}" --version "unix-{version}"
+recipe.hooks.sketch.prebuild.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
+recipe.hooks.core.prebuild.pattern="{runtime.tools.python.path}/python" "{runtime.tools.makecorever}" --build_path "{build.path}" --platform_path "{runtime.platform.path}" --version "unix-{version}"
 
 ## Build the app.ld linker file
 recipe.hooks.linking.prelink.1.pattern="{compiler.path}{compiler.c.cmd}" -CC -E -P {build.vtable_flags} "{runtime.platform.path}/tools/sdk/ld/eagle.app.v6.common.ld.h" -o "{build.path}/local.eagle.app.v6.common.ld"
--- a/platform.txt
+++ b/platform.txt
@@ -83,8 +83,8 @@ compiler.elf2hex.extra_flags=
 
 ## generate file with git version number
 ## needs bash, git, and echo
-recipe.hooks.core.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
-recipe.hooks.core.prebuild.2.pattern="{runtime.tools.python.path}/python" "{runtime.tools.makecorever}" --build_path "{build.path}" --platform_path "{runtime.platform.path}" --version "unix-{version}"
+recipe.hooks.sketch.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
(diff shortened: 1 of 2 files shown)


platform.txt
```diff
@@ -83,8 +83,8 @@ compiler.elf2hex.extra_flags=
 
 ## generate file with git version number
 ## needs bash, git, and echo
-recipe.hooks.core.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
-recipe.hooks.core.prebuild.2.pattern="{runtime.tools.python.path}/python" "{runtime.tools.makecorever}" --build_path "{build.path}" --platform_path "{runtime.platform.path}" --version "unix-{version}"
+recipe.hooks.sketch.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
+recipe.hooks.core.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.makecorever}" --build_path "{build.path}" --platform_path "{runtime.platform.path}" --version "unix-{version}"
```
earlephilhower: This breaks the packager script.  For build-boards-manager, we remove this line by name since the version will be pre-gen's in the Arduino tarball.

platform.txt
```diff
@@ -83,8 +83,8 @@ compiler.elf2hex.extra_flags=
 
 ## generate file with git version number
 ## needs bash, git, and echo
-recipe.hooks.core.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
-recipe.hooks.core.prebuild.2.pattern="{runtime.tools.python.path}/python" "{runtime.tools.makecorever}" --build_path "{build.path}" --platform_path "{runtime.platform.path}" --version "unix-{version}"
+recipe.hooks.sketch.prebuild.1.pattern="{runtime.tools.python.path}/python" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
```
earlephilhower: No need for a .1. here if there's only 1 step in sketch.prebuild.  That'll also help us get closer to working w/the native Arduino plugin on Eclipse/.

platform.txt
earlephilhower: Same comment about the .1. not being needed, as above.

qistoph: Both fixed. PR branch is updated.

Could I have easily spotted this myself, e.g. by running some test scripts?

platform.txt
qistoph: Fixed, updated PR branch.

platform.txt
earlephilhower: Not easily, unfortunately.  The `package/build_boards_manager.sh` script is run on releases to generate the uploaded tarball.  While the script w/o any changes would still produce a *legal* platform.txt, it would try and regenerate the GIT version and fail whenever anything was built.

You'd have to run the script, make a json, upload it and your tarball to a private webserver, add that JSON url to your boards manager in the IDE, then install it (after removing/hiding your git copy).

PR #4482, opened 2018-03-09
Raise an exception when assert fails

igrr:
- Raise an exception when assert fails
- Move GDB stub hooks into a separate file, provide header for it
- Use syscall instruction raise user mode exception
- Remove unused code in postmortem.c

Fixes #4480

PR #6288, opened 2019-07-11
fix raise_exception()

### LINKED ISSUE #6283: panic() called from ISR causes Hardware WDT
### Basic Infos

- [x] I have tested that the issue is present in the current master branch (aka latest git).

#### Platform

- Hardware:					**ESP-12**
- Core Version:			[5a47cab77d396afeb842c9161327f9f4dca7c833]
- Development Env:		**Arduino IDE**
- Operating System:		**Ubuntu**

### Settings in IDE

- Module:  			**Adafruit HUZZAH**
- Flash Mode:			**qio**
- Flash Size: 		**4MB**
- lwip Variant:   **v2 Lower Memory**
- Reset Method:		**nodemcu**
- Flash Frequency:	**40Mhz**
- CPU Frequency:		**80Mhz**
- Upload Using:		**SERIAL**
- Upload Speed:   **115200** (serial upload only)

### Problem Description

I have found, that there are some actions that an ISR can take, that can unexpectedly result in a "Hardware WDT". Calls made from an ISR to panic(), abort(), assert(), or the __unhandled_exception() will result in: no message, no stack trace, just: 
```

wdt reset
```

Obviously, it would be best to avoid these in our ISRs. However, it might not be obvious we are doing these calls from an ISR. Take the case of running a build with just "Debug: Serial" selected in the IDE.
This will enable the `UMM_POISON`, build flag. If an ISR steps on the poison, then later does a `free()`, a message is printed (with non-IRAM based `printf()` function) then an attempt is made within `umm_malloc.cpp` to call `panic()`. When I tested this the umm poison message printed; however, the `panic()` details never appeared. Alter a long period, the "Hardware WDT" event appeared.

What I think I am seeing is, calls from an ISR that reach a panic(), abort(), assert(), or the __unhandled_exception() will eventually arrive at  `raise_exception()` in module `core_postmortem.cpp`. Then onto the inline assembly for "syscall". After a long period of silence, I see the "Hardware WDT" event. My test method at that time was less than ideal because I was using a printf function that was not in IRAM. However, it did print all the way up to the "syscall", before the WDT appeared and nothing at __wrap_system_restart_local(). 

It is interesting to find that there are causes for the "Hardware WDT", that are not caused by being stuck in a loop. 

In Arduino Core 2.5.0 and up, a failed "new" operator without `(std::nothrow)` in an ISR will call `abort()` resulting in WDT. I am concerned that there might be a lot more cases out there. 

This one is unconfirmed however looks like a good one to verify:

https://github.com/esp8266/Arduino/blob/7c6701512f44d27b998bd59119631f645dc37dba/cores/esp8266/core_esp8266_main.cpp#L104-L115

https://github.com/esp8266/Arduino/blob/7c6701512f44d27b998bd59119631f645dc37dba/cores/esp8266/cont_util.cpp#L65-L68

It looks like a call from an ISR, that reaches yield(), would call panic() and the message would never be seen.

I could say more but I'll wait and see what you say, just in case I made a wrong turn in my assessment of this issue.

```cpp

#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <interrupts.h>
#include <assert.h>

#define CC4US_DELAY(us)   ((us) * clockCyclesPerMicrosecond())

void ICACHE_RAM_ATTR timer0_handler(void) {
  timer0_detachInterrupt();
  // From an ISR context these calls will fail with a hardware WDT
  // No event related information or stack trace is printed.
  panic();
  // abort();
  // assert(1==0);
}

void setup(void) {
  Serial.begin(115200);
  delay(20);
  WiFi.mode(WIFI_OFF);
  Serial.printf("\n\nUp and running.\n");
  {
    AutoInterruptLock(15);
    timer0_isr_init();
    timer0_attachInterrupt(&timer0_handler);
    timer0_write(ESP.getCycleCount() + CC4US_DELAY(150));
  }
}

void loop() {
}

```

### Debug Messages

```
Up and running.

wdt reset
tail 8
chksum 0x2d
csum 0x2d
v5a47cab7
~ld
```

d-a-v:
This is a proposal 

~It seems this has no *visible* effect:~
```
__asm__ __volatile__ ("syscall"); // triggers gdb or does nothing
```

On `panic`/`abort`/`assert` called from user context, WDT happens due to the following `while(1);` and finally produces an exception with the right message (after the WDT reason and delay).
On ISR context, only the WDT reason is visible.

This commit permits to avoid WDT and show the right message from both user or ISR context.

Before #4482, `*((int*)0)=0` was used to trigger an hardware exception (similar to the WDT exception we have now, also showing an additional message unrelated to the real cause).

On real hardware exception, `raise_exception()` is not called.

fixes #6283 (thanks @mhightower83)

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -52,6 +52,13 @@ static void uart0_write_char_d(char c);
 static void uart1_write_char_d(char c);
 static void print_stack(uint32_t start, uint32_t end);
 
+// using numbers different from "REASON_" in user_interface.h (=0..6)
+enum rst_reason_sw
+{
+    REASON_USER_SWEXCEPTION_RST = 254
+};
+static int s_user_reset_reason = REASON_DEFAULT_RST;
+
 // From UMM, the last caller of a malloc/realloc/calloc which failed:
 extern void *umm_last_fail_alloc_addr;
 extern int umm_last_fail_alloc_size;
@@ -82,29 +89,24 @@ static void ets_printf_P(const char *str, ...) {
     }
 }
 
-#define USER_REASON_NONE 255
-#define USER_REASON_SWEXCEPTION 254
-// using numbers different from "REASON_" in user_interface.h 0..6
-static int user_reset_reason = USER_REASON_NONE;
-
 void __wrap_system_restart_local() {
     register uint32_t sp asm("a1");
     uint32_t sp_dump = sp;
 
     struct rst_info rst_info;
     memset(&rst_info, 0, sizeof(rst_info));
-    if (user_reset_reason == USER_REASON_NONE)
+    if (s_user_reset_reason == REASON_DEFAULT_RST)
     {
         system_rtc_mem_read(0, &rst_info, sizeof(rst_info));
         if (rst_info.reason != REASON_SOFT_WDT_RST &&
             rst_info.reason != REASON_EXCEPTION_RST &&
             rst_info.reason != REASON_WDT_RST)
         {
-            return;
+            rst_info.reason = REASON_DEFAULT_RST;
         }
     }
     else
-        rst_info.reason = user_reset_reason;
+        rst_info.reason = s_user_reset_reason;
 
     // TODO:  ets_install_putc1 definition is wrong in ets_sys.h, need cast
     ets_install_putc1((void *)&uart_write_char_d);
@@ -128,6 +130,9 @@ void __wrap_system_restart_local() {
     }
     else if (rst_info.reason == REASON_SOFT_WDT_RST) {
         ets_printf_P(PSTR("\nSoft WDT reset\n"));
+    }
+    else {
+        ets_printf_P(PSTR("\nGeneric Soft Reset\n"));
     }
 
     uint32_t cont_stack_start = (uint32_t) &(g_pcont->stack);
@@ -226,7 +231,7 @@ static void raise_exception() {
     if (gdb_present())
         __asm__ __volatile__ ("syscall"); // triggers GDB when enabled
 
-    user_reset_reason = USER_REASON_SWEXCEPTION;
+    s_user_reset_reason = REASON_USER_SWEXCEPTION_RST;
     ets_printf_P(PSTR("\nUser exception (panic/abort/assert)"));
     __wrap_system_restart_local();
 
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -132,7 +132,7 @@ void __wrap_system_restart_local() {
         ets_printf_P(PSTR("\nSoft WDT reset\n"));
     }
     else {
-        ets_printf_P(PSTR("\nGeneric Soft Reset\n"));
+        ets_printf_P(PSTR("\nGeneric Reset\n"));
     }
 
     uint32_t cont_stack_start = (uint32_t) &(g_pcont->stack);
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -230,6 +230,10 @@ static void uart1_write_char_d(char c) {
 
 static void raise_exception() {
     __asm__ __volatile__ ("syscall");
+    ets_printf_P(PSTR("exception from IRQ\n"));
+workaround=7; // last REASON_EXT_SYS_RST      = 6
+    __wrap_system_restart_local();
+    ets_printf_P(PSTR("beh\n"));
     while (1); // never reached, needed to satisfy "noreturn" attribute
 }
 
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -82,7 +82,10 @@ static void ets_printf_P(const char *str, ...) {
     }
 }
 
-static int workaround = 0;
+#define FAKE_REASON_NONE 255
+#define FAKE_REASON_USER  254
+static int fake_rst_reason = FAKE_REASON_NONE;
+
 void __wrap_system_restart_local() {
     register uint32_t sp asm("a1");
     uint32_t sp_dump = sp;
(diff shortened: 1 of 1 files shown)


d-a-v:
This is tested with gdb:

- user exception in user context:
gdb doesn't show the exact line,
but that kind of exception is anyway displayed on console (panic/abort/assert)
- hardware exception in user context - ok
- user exception in sys context - ok
- hardware exception in sys context - ok

So there is room for improvement on gdb side (I'm not a specialist)

earlephilhower:
I think that's exactly as it was for GDB.  On assert/panic/etc. you get a breakpoint in the raise_exception() call and need to "up" one or two levels to get to the actual line of code (but it's still there in the stack so it's fully functional).

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -82,10 +82,15 @@ static void ets_printf_P(const char *str, ...) {
     }
 }
 
+#define FAKE_REASON_NONE 255
```
earlephilhower: Maybe silly question, but why not just a "bool swException" or something like that?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -82,10 +82,15 @@ static void ets_printf_P(const char *str, ...) {
...
+#define FAKE_REASON_USER 254
+static int fake_rst_reason = FAKE_REASON_NONE;
+
 void __wrap_system_restart_local() {
     register uint32_t sp asm("a1");
     uint32_t sp_dump = sp;
 
+#if 0
```
earlephilhower: We should drop these lines completely.  Git will preserve the old info if needed.

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -222,7 +233,24 @@ static void uart1_write_char_d(char c) {
 }
 
 static void raise_exception() {
-    __asm__ __volatile__ ("syscall");
+#if 0
+
```
earlephilhower: We should drop these lines completely.  Git will preserve the old info if needed.

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -222,7 +233,24 @@ static void uart1_write_char_d(char c) {
...
+    // works but also showing
+    // "Fatal exception 29(StoreProhibitedCause)"
+    *((char*)0) = 0;
+
+#else
+
+    if (gdb_present())
+        //*((char*)0) = 0;
```
earlephilhower: Ditto

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -94,16 +99,22 @@ void __wrap_system_restart_local() {
...
-        rst_info.reason != REASON_WDT_RST)
+    if (fake_rst_reason == FAKE_REASON_NONE)
     {
-        return;
+        system_rtc_mem_read(0, &rst_info, sizeof(rst_info));
+        if (rst_info.reason != REASON_SOFT_WDT_RST &&
+            rst_info.reason != REASON_EXCEPTION_RST &&
+            rst_info.reason != REASON_WDT_RST)
+        {
+            return;
```
mhightower83: Since you are in this code I thought I would raise this question `return`.

I have been through this module a few times and I don't understand how a `return` would work from `system_restart_local()`? Is there anything to return to? 

I searched for quite a while for info on `system_restart_local` but I could never find anything official. However, I did find an old pseudo C interpretation of disassembled ~v1 SDK. After taking a second look at the pseudo C code, `system_restart_local()` appears to be called from either a scheduled timer callback 100ms after `system_restart` is called or from Soft WDT or some system startup failure. I would expect more paths but I don't see them yet. That's all the info I could find and it is probably outdated for what we have today.

Here is an outline of the logic flow I am reading:

`system_restart()` (FLASH)
  * shutdown wifi interface
  * shutdown wifi timer callbacks
  * Start a callback timer to run `system_restart_local()` once, after 100ms delay.
  * returns to caller
    * SDK notes: "... do not call other functions after calling this API"

`system_restart_local()` (FLASH) - Runs from a timer callback or called directly from a Soft WDT call or Startup, maybe some more indirect paths!
  * Handle some strange stuff
  * Grab a copy of restart info from RTC memory.
  * Keep restart info reason REASON_EXCEPTION_RST or REASON_SOFT_WDT_RST for all others:
      * zero restart info
      * set reason to REASON_SOFT_RESTART.
      * Update RTC.
  * Call `system_restart_hook(&rst_inf)`
  * wait for uart TX fifo's to empty
  * Enter lock, do some hardware stuff
  * call `system_restart_core()` - ps.intlevel still at 3

`system_restart_core()` (IRAM) - This is the end of the line
  * Wait for SPI flashchip to idle down
  * Disable read cache.
  * pass control to reset vector

I have not done any experimentation with this one, so I cannot say much more. I am not exactly sure how to generate an event to take the path toward that `return`. Maybe something for a later investigation.

earlephilhower: system_restart_local is called by the system boot logic in the SDK libmain.a file.  I don't have the exact logic sequence, but it is expected to return.  The SDK will then handle the IRAM loading, clearing the BSS, copying RODATA, etc. and start the app.  Don't have a clear picture, but since a) the stack is still good enough to backtrack and b) the static globals haven't been cleared, we can intuit the above.

The way things work here is we use GCC to rename our function "system_restart_local" and then change the libmain.a's fcn to __real_... Basically, @igrr found a cool way to shim in this debug code before the stack is corrupted or the RTC clock memory (storing the exception info) is cleared.

The return you hilighted *maybe* should be   `{ __real_system_restart_local(); return; }` but, if it was a return there before it's been there a long time without any issue so I'd be loathe to open it up w/o deeper understanding.

What are you looking for, specifically?  I'm a little worried this could be pulling you away from the real problem you're trying to solve...

devyte: I agree with @earlephilhower: maybe it shouldn't be a return, but a call to __real_system_restart_local(). That was my first thought when trying to understand the issue described.
But my second thought was exactly the same as his: if that is true, then the issue has been present for a very long time, which begs the question: why hasn't a problem been found before? Or maybe there has, but we haven't realized it?
I think it would make sense to discuss this return statement in detail elsewhere. We shouldn't hijack this PR. @mhightower83 If you're willing to pursue this, could you please open a new issue and explain the suspected problem there? Maybe link this discussion, or quote it there. That issue could then be used to track further investigation as well as experimentation.

cores/esp8266/core_esp8266_postmortem.cpp
devyte: I was thinking maybe we should differentiate between panic, abort, maybe other cases, with just one static global, i.e.: replace s_abort_called and this user reset reason with a single enum.

earlephilhower:
It should work w/C++ exceptions, should they get fixed at some point.  Exceptions don't involve this bit at all unless they're uncaught, at which point panic() is called and an exception reason printed (__unhandled_exception).

PR #6292, opened 2019-07-12
esptool.py: disable 9600bauds in menu for flash upload serial speed

d-a-v:
Users can be mistaken by some board documentation stating to use 9600bauds.
Flash upload speed at 9600bauds doesn't work with esptool.py.
([ref](https://github.com/esp8266/Arduino/issues/6110#issuecomment-510067085))

(+ minor cosmetics update in host emulation makefile)


d-a-v:
coverage is disabled in the host emulation part only. CI host test still has it.

PR #6293, opened 2019-07-12
Create empty method for String class

ruggi99:


earlephilhower:
@d-a-v, this brings to mind the constant emptyString.  Your thoughts?

jjsuwa:
@d-a-v wrote:
> There was an aborted effort #5546 from @jjsuwa to bring global `emptyString` right into this class.
> It it were to happen again, its name would be `String::empty`.
> Thus I'd propose do rename the bool function in this PR to `String::isEmpty()`, what do you think ?

Of course me too, `String::empty()` appears for me to be a straightforward misunderstanding of "truncation the string to 0-length".

d-a-v:
@ruggi99 
Can you please rename the new `String::empty()` to `String::clear()`,
so we can reserve `String::empty` for another PR (not this one) in which global `emptyString` would be renamed to `String::empty` (with changes for all references to `emptyString` through this repository).

ruggi99:
I've done. I think `setLen(0)` is enough to clear a string, isn't it?

jjsuwa:
@d-a-v wrote:
> so we can reserve `String::empty` for another PR (not this one) in which global `emptyString` would be renamed to `String::empty` (with changes for all references to `emptyString` through this repository).

No excessive consideration about the PR, it's merely proposal :)

ruggi99:
> No excessive consideration about the PR, it's merely proposal :)

I think this is a good idea

d-a-v:
> I've done. I think setLen(0) is enough to clear a string, isn't it?

It marks it as empty without any kind of reallocation, I guess this is what to be expected for this call.

> No excessive consideration about the PR, it's merely proposal :)

Why has it been closed ?

jjsuwa:
> > No excessive consideration about the PR, it's merely proposal :)

> Why has it been closed ?

It was premature I guessed.

PR #6294, opened 2019-07-13
Upgrade to GCC 10.1 toolchain

### LINKED ISSUE #3352: Multiline raw Strings not parsed in Macro
### Basic Infos

The esp8266 toolkit uses gcc 4.8.2 an thus it fails to parse a multiline raw String when it is passed as an argument to a macro. The same sketch as below will build without errors when using the original Arduino toolkit (when switching to AVR / Arduino Hardware in the Arduino IDE).

This was fixed in gcc 4.9 : 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57824

### Sketch

```cpp

#include <Arduino.h>

#define S(s) s

const char y[] = S(R"(
abc
)");

void setup() {
}

void loop() {
}

```

### Error Message

```
sketch_jun13a:3: error: unterminated raw string
 const char y[] = S(R"(
                    ^
exit status 1
unterminated raw string
```

### Resolution?

Is there a reason on why you are using the 4.8.2 version of gcc? Max Filippov seems to provide newer versions on https://github.com/jcmvbkbc/linux-xtensa

So would it be possible to use a newer gcc version? Or would one have to "backfix" the old gcc for some reason?

earlephilhower:
All examples I've tested run fine.  GDB is working.  No warnings.

Fixes #3352


earlephilhower:
Lots of parenthesis warning, but basic examples like "Blink.ino" and "BearSSL_Validation.ino" run fine as far as I can see now.  Still optimization on exceptions and removing the warnings which are causing the plain Arduino CI builds to fail.

earlephilhower:
@d-a-v, could you check the AddrList and IPAddr changes?

d-a-v:
Comparing a sketch with `.text1` fitting issue, it appears symbol `.divdi3` takes much space in iram
```
      0x3df /path/to/libgcc.a(_divdi3.o)
```
This symbol is not here with gcc4.8.
Apart from that, I also run `BearSSLValidation` and it works great here too.
That's a great leap for espkind !

earlephilhower:
Thanks, @d-a-v.  I haven't been looking at the iram yet, still trying to get rid of the warnings.  Are you OK with the changes to IP classes here?  https://github.com/esp8266/Arduino/pull/6294/commits/3afe9e3f0387d60b2038281679e521be8b89578b

earlephilhower:
The latest push has added back "-mforcel32" as a GCC option, and patches that allow exception processing to process in-flash `eh_tables` without any HW misaligned exception handler or issues, and a fix to std::regex where it was crashing due to accessing `__ctype__` directly (bytewise, but ctype is now in flash)...

Jason2866:
Tried to compile project Tasmota with. Does compile (some warnings....) resulting firmware does work :-)
Drawback new GCC 9.1 does use even more of the barely iram.

earlephilhower:
@Jason2866, thanks for the update.

Can you attach the ELFs with 4.8 (i.e. normal) and 9.1 gccs?  It's possible to see where the growth is assuming the function names are identical.  @d-a-v found one large function that needed to be stripped, and there may be others.

earlephilhower:
There were 2 FP functions that should have been stripped out, @Jason2866.  With that fixed, I see darn near identical IRAM usage on my big test script:
````
91/SecureBearSSLUpdater.ino.map:                0x0000000040106df8                _text_end = ABSOLUTE (.)
48/SecureBearSSLUpdater.ino.map:                0x0000000040106dd0                _text_end = ABSOLUTE (.)
````
That's df8-dd0 = only 40 bytes larger in total usage, GCC 9.1 vs GCC 4.8.

**Comment by Jason2866** [comment]
@earlephilhower  The 2 elf and map files (gcc4.8 and gcc 9.1)
[Gcc91_vs_gcc48_elf.zip](https://github.com/esp8266/Arduino/files/3416545/Gcc91_vs_gcc48_elf.zip)

**Comment by earlephilhower** [comment]
@Jason2866 looked at the ELFs and the only difference were the 2 new functions that I just removed from the libgcc.a.  Downloading and running the latest version of the PR (need to do a tools/get.py again) will remove that and get you to within a few bytes of 4.8 size in IRAM.

**Comment by Jason2866** [comment]
Cant test at the moment. Far away from my Linux machine. Have only a Windows 10 Pc here...

**Comment by earlephilhower** [comment]
No worries or rush.  `<famous_last_words>`I'm pretty sure that was the last major difference and now it's just a matter of sitting on our hands until 3.0.`</famous_last_words>`

**Comment by earlephilhower** [comment]
Note that #6472 needs to be applied here, as well, once merged.

**Comment by devyte** [comment]
Conflicts

**Comment by Niek** [comment]
For everyone willing to test this in PlatformIO, here are some instructions:
```bash
cd ~/.platformio/packages

# Clone GCC9 branch from @earlephilhower 
git clone https://github.com/earlephilhower/Arduino.git framework-arduinoespressif8266@3
cat <<EOF > framework-arduinoespressif8266@3/package.json
{
    "name": "framework-arduinoespressif8266",
    "description": "Arduino Wiring-based Framework (ESP8266 Core)",
    "url": "https://github.com/esp8266/Arduino",
    "version": "3.0.0"
}
EOF

# Install latest GCC9 toolchain (instructions are for macOS, for other OSes download the right package and change the system value)
curl -sL https://github.com/earlephilhower/esp-quick-toolchain/releases/download/3.0.0-gnu2/x86_64-apple-darwin14.xtensa-lx106-elf-dd9f9a2.1569802152.tar.gz | tar xvz
mv xtensa-lx106-elf toolchain-xtensa@3
cat <<EOF > toolchain-xtensa@3/package.json
{
    "name": "toolchain-xtensa",
    "system": [
        "darwin_x86_64",
        "darwin_i386"
    ],
    "url": "https://github.com/earlephilhower/esp-quick-toolchain",
    "version": "3.0.0"
}
EOF
```
Then, add in your platformio.ini: ``platform_packages = framework-arduinoespressif8266@3, toolchain-xtensa@3``

I'm getting some linking errors though:
```c++
~/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/env/libfbb/libESP8266WiFi.a(ESP8266WiFiGeneric.cpp.o):(.text._ZNSt23_Sp_counted_ptr_inplaceI22WiFiEventHandlerOpaqueSaIS0_ELN9__gnu_cxx12_Lock_policyE0EE14_M_get_deleterERKSt9type_info[std::_Sp_counted_ptr_inplace<WiFiEventHandlerOpaque, std::allocator<WiFiEventHandlerOpaque>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)]+0x4): undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
~/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/env/libfbb/libESP8266WiFi.a(ESP8266WiFiGeneric.cpp.o): in function `std::_Sp_counted_ptr_inplace<WiFiEventHandlerOpaque, std::allocator<WiFiEventHandlerOpaque>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)':
~/.platformio/packages/toolchain-xtensa@3/xtensa-lx106-elf/include/c++/9.2.0/bits/shared_ptr_base.h:587: undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
~/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/env/libfbb/libESP8266WiFi.a(WiFiClientSecureBearSSL.cpp.o): in function `std::_Sp_counted_ptr_inplace<BearSSL::br_x509_insecure_context, std::allocator<BearSSL::br_x509_insecure_context>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)':
~/.platformio/packages/toolchain-xtensa@3/xtensa-lx106-elf/include/c++/9.2.0/bits/shared_ptr_base.h:583: undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
~/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/env/libfbb/libESP8266WiFi.a(WiFiClientSecureBearSSL.cpp.o): in function `std::_Sp_counted_ptr_inplace<br_x509_knownkey_context, std::allocator<br_x509_knownkey_context>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)':
~/.platformio/packages/toolchain-xtensa@3/xtensa-lx106-elf/include/c++/9.2.0/bits/shared_ptr_base.h:575: undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
~/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/env/libfbb/libESP8266WiFi.a(WiFiClientSecureBearSSL.cpp.o): in function `std::_Sp_counted_ptr_inplace<br_x509_minimal_context, std::allocator<br_x509_minimal_context>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)':
~/.platformio/packages/toolchain-xtensa@3/xtensa-lx106-elf/include/c++/9.2.0/bits/shared_ptr_base.h:583: undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
~/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/env/libfbb/libESP8266WiFi.a(WiFiClientSecureBearSSL.cpp.o):~/.platformio/packages/toolchain-xtensa@3/xtensa-lx106-elf/include/c++/9.2.0/bits/shared_ptr_base.h:583: more undefined references to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)' follow
```

earlephilhower:
@Niek can you post some small bit of code that shows this error on link?  The CI using arduino-builder is compiling apps with BearSSL just fine.

It also might be a cached pre-built library you've got somewhere in PIO config.  The C++ std:: library could well have changed internal representations for templates like this, and while a full compile from scratch would work fine, any attempts to link to gcc 4.8 compiled code could barf.

Niek:
@earlephilhower I thought it would be a cached lib as well, but this is after removing the package cache and with a fresh new environment. I'll try to come up with a minimal sample to show the issue.

Jason2866:
@earlephilhower  Tried 3.0.0 with platformio. Compiles without any error
```
Compiling .pioenvs\sonoff\FrameworkArduino\umm_malloc\umm_malloc.cpp.o
Archiving .pioenvs\sonoff\libFrameworkArduino.a
Linking .pioenvs\sonoff\firmware.elf
Retrieving maximum program size .pioenvs\sonoff\firmware.elf
Checking size .pioenvs\sonoff\firmware.elf
Building .pioenvs\sonoff\firmware.bin
Creating BIN file ".pioenvs\sonoff\firmware.bin" using ".pioenvs\sonoff\firmware.elf"
Memory Usage -> http://bit.ly/pio-memory-usage
DATA:    [======    ]  55.5% (used 45500 bytes from 81920 bytes)
PROGRAM: [=====     ]  49.2% (used 504168 bytes from 1023984 bytes)
======================================== [SUCCESS] Took 196.46 seconds ========================================

Environment      Status    Duration
---------------  --------  ------------
sonoff           SUCCESS   00:03:16.464
```

Niek:
@earlephilhower this is a minimal example to show the issue:

platformio.ini:
```ini
[platformio]
src_dir = .

[env:test]
platform = https://github.com/platformio/platform-espressif8266#feature/stage
platform_packages = framework-arduinoespressif8266@3, toolchain-xtensa@3
framework = arduino
board = esp01_1m
```
Put the BearSSL validation example in the same dir: https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino

You will get the linking errors when running ``platformio run``.

mcspr:
It could be mixing up toolchain / package versions? I don't think it correctly resolves platform_packages = ...
```cpp
#include <Arduino.h>
#include <memory>

struct test_t {
    void print() {
        Serial.println("something");
    }
};

void setup() {
    Serial.begin(115200);
    auto test1 = std::make_shared<test_t>();                                                                                                                                test1->print();                                                                                                                                                         auto test2 = std::shared_ptr<test_t>(new test_t());                                                                                                                     test2->print();                                                                                                                                                     }                                                                                                                                                                                                                                                                                                                                               void loop() {                                                                                                                                                                                                                                                                                                                                   }
```

If I try to build this with gcc-9 using the current master it would fail with the same error related to the  make_shared_tag. But everything works if I checkout into the gcc9.1 branch at the ~/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910
```
$ git describe --all
heads/master
$ pio run -v 
...
PACKAGES: toolchain-xtensa 3, framework-arduinoespressif8266 6944b6f1 (git+https://github.com/esp8266/Arduino.git), tool-esptool 1.413.0 (4.13), tool-esptoolpy 1.20600.
0 (2.6.0)
...
/home/mcspr/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/d1_mini/src/sketch.ino.cpp.o:(.text._ZNSt23_Sp_counted_ptr_inplaceI6test_tSaIS0_ELN9__gnu_cxx12_Lock_policyE0EE14_M_get_deleterERKSt9type_info[std::_Sp_counted_ptr_inplace<test_t, std::allocator<test_t>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)]+0x4): undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'         /home/mcspr/.platformio/packages/toolchain-xtensa@3/bin/../lib/gcc/xtensa-lx106-elf/9.2.0/../../../../xtensa-lx106-elf/bin/ld: .pio/build/d1_mini/src/sketch.ino.cpp.o: in function `std::_Sp_counted_ptr_inplace<test_t, std::allocator<test_t>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)':
sketch.ino.cpp:(.text._ZNSt23_Sp_counted_ptr_inplaceI6test_tSaIS0_ELN9__gnu_cxx12_Lock_policyE0EE14_M_get_deleterERKSt9type_info[std::_Sp_counted_ptr_inplace<test_t, std::allocator<test_t>, (__gnu_cxx::_Lock_policy)0>::_M_get_deleter(std::type_info const&)]+0x1a): undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
```

Jason2866:
@Niek  Tried your described way to setup. Result, i did not got a correct build environment.
With by "hand" copied files in a already existing config and changing the .json files did work.

earlephilhower:
The BearSSL_validation.ino stuff we supply (libraries and sources) definitely do build using 9.2, so this has to be a PIO issue somewhere.

Since I'm not a PIO user, thanks for your help @Jason2866 and @mcspr!

Niek:
My bad... forgot the -b gcc9.1 in the checkout, thanks for the pointer @mcspr. Building works fine now.

earlephilhower:
Great.  The move to GCC 9.2 was needed because GCC 9.1 had some issues compiling for MINGX and OSX that were cleared.  It's only a bugfix from my read of the changelog, so nothing but goodness.

devyte:
Conflicts here again

mikee47:
Hi, we're integrating support for your new toolchains into [Sming](https://github.com/SmingHub/Sming/pull/1825) and looking really good so far. Thank you for all the hard work.

One issue so far is that the new compiler appears to generate more verbose code in IRAM. Is that something you've come across?

earlephilhower:
@mikee47, thanks for being the guinea pigs! :)

For the larger IRAM sizes, what is most helpful is getting an `objdump -t` of the same app using old and new toolchains.  That lets us see if something from gcc proper is being brought in (that wasn't before).  I'm pretty sure I got rid of any redundancies (things already present in ROM), but you might have turned up something.

For the template and section interaction, sorry, but I think that's a GCC issue and not much can be done about it.  I just pull what the GNU org supplies and they might have changed how templates are expanded under the hood...

mikee47:
@earlephilhower Thanks for the tips. It's definitely the optimiser. Recompiling at `-O3` brings the size right down, does a better job than `-Os` so worth noting for IRAM:

https://github.com/SmingHub/Sming/pull/1825#issuecomment-560428095

The code is templated but I suspect the reason for this is more likely to be code complexity. I'll try to pin it down to specific optimisation setting(s).

mikee47:
Couple of linker script .irom0.text output section filters, might be worth considering for Arduino:

`*(.rodata.__func__*)` - __FUNCTION__ usage within C code, e.g. `__func__$4470`
`*(.rodata._ZZ*__c)` - PSTR() usage in templated code

It would be better I think if PSTR() used something more descriptive than `__c` for it's static variable, e.g. `__pstr`, would you consider updating pgmspace.h?

earlephilhower:
@mikee47 I've just started a full release build (-gnu4) with I *think* all the changes you'd mentioned.  Give it about a hour and then you can try pointing to that new release and give it a go.  We had changes spread around in lots of places, so I *hope* it's all good.

@d-a-v, @devyte as part of this I tried w/C++17 and the only issues w/the core are `register` which is deprecated and the macro redefinition of `abs()` in `Arduino.h`.

This PR is still at gnu++14, but maybe gnu++17 should be considered?

devyte:
The abs() macro needs to be cleaned up anyways the same as was done with min/max.
The register can just be dropped in our core.

mikee47:
Note that `strlen` is still present in newlib...

With Sming, found a few extra libgcc routines lurking in IRAM, saves me another 800 bytes moving them to flash. I can't see them filtered out in Arduino so worth a check:

```
    *libgcc.a:_modsi3.o(.literal .text)
    *libgcc.a:_divsf3.o(.literal .text)
    *libgcc.a:_fixsfsi.o(.literal .text)
    *libgcc.a:_cmpdf2.o(.literal .text)
    *libgcc.a:_cmpsf2.o(.literal .text)
```

earlephilhower:
For the two that are already in ROM, I've got another clean build going.  They were taken out of 4.8, but I lost them somewhere before 9.2.

I don't think we're going to move all the soft-FP routines not in ROM into IRAM.  They're just too big and haven't been there before.  IRAM for us is pretty much the biggest issue, not nearly enough...

mikee47:
> I don't think we're going to move all the soft-FP routines not in ROM into IRAM.

Eeek! No, I was suggesting the reverse - those soft-fp routines are present in `libgcc.a` and end up in IRAM as there's no linker map filter for them.

Just to confirm the issue, I've installed arduino from GIT develop and compiled `test_newlib`:

```
IROM   : 242932          - code in flash         (default or ICACHE_FLASH_ATTR) 
IRAM   : 27504   / 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...) 
```

Then updated `eagle.app.v6.common.ld.h`, adding these lines to .irom0.text output section:

```
    *libgcc.a:_divsf3.o(.literal .text)
    *libgcc.a:_fixsfsi.o(.literal .text)
    *libgcc.a:_cmpdf2.o(.literal .text)
    *libgcc.a:_cmpsf2.o(.literal .text)
```

We now get this:

```
IROM   : 243828          - code in flash         (default or ICACHE_FLASH_ATTR) 
IRAM   : 26464   / 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...) 
```

earlephilhower:
Got it, @mikee47 .  Latest -gnu5 and this push should have those taken care of as well.

earlephilhower:
Looks like `using std::abs` blows up the C library headers.  Will drop that change and just leave it undefined in Arduino.h which will pull in the C lib version by default.

earlephilhower:
Fixes #6299

Jason2866:
Great work. Compiled Tasmota with your latest changes and ESP8266 Quick Toolchain for 3.0.0-gnu11(windows). Just a few warnings. The compiled firmware does work -:)

earlephilhower:
@Jason2866, I can add a package.json pretty simply as part of the tarball process.  However, I'm not really sure what to put in it.  Where is the format specified?  I assume it's Platform.IO specific file, but when I check the PIO stuff I don't see anything with the `system` key like in the example you linked.

Jason2866:
@earlephilhower 
```{
    "description": "xtensa-gcc",
    "name": "toolchain-xtensa",
    "system": [
        "windows",
        "windows_amd64",
        "windows_x86"
    ],
    "url": "https://github.com/earlephilhower/esp-quick-toolchain",
    "version": "3.00100.150520"
} ```
The first part doesnt need a change. "System" i think does explain itself.
"version": "3.00100.150520"  `3` is build chain main version and `00100` describes it as version 1.0
`150520` is the date in format `ddmmyy` of the build date of the version

**Comment by earlephilhower** [comment]
@Jason2866 , please look at https://github.com/earlephilhower/esp-quick-toolchain/releases/tag/6.6.6-testonly and let me know if the package.json files work.  I only build the linux versions as my server's at high load today.

If the format's good, I can add the files to the standard automated builds.  If not, let me know what needs to be fixed to clean it up.

**Comment by earlephilhower** [comment]
Also, wouldn't it make more sense to do YYMMDD on the timestamp?  DDMMYY won't sort alphabetically or numerically properly...

**Comment by Jason2866** [comment]
Great, thx!! Does work. Just tested with Gitpod (Online Platformio IDE using Linux)
`toolchain-xtensa @ 5.01010.210520 has been successfully installed!`
Good idea for changing timestap order.
If you would include in your builds it would be really great. Thx in advance!!

**Comment by earlephilhower** [comment]
Thanks for the update, @Jason2866.  I've added the code to the makefile so the next rebuild will include it.  I will not blow away the temp 6.6.6 release.

**Comment by mcspr** [comment]
I just noticed an issue building brzo i2c lib. As described above, I used the PIO + package.json and tried to build an example code
https://github.com/pasko-zh/brzo_i2c/blob/master/examples/HTU21D/HTU21D.ino

```
xtensa-lx106-elf-gcc -o ".pio/build/d1_mini/lib8be/Brzo I2C_ID335/brzo_i2c.c.o" -c -std=gnu99 -Wpointer-arith -Wno-implicit-function-declaration -Wl,-EL -fno-inline-functions -nostdlib -Os -mlongcalls -mtext-section-literals -falign-functions=4 -U__STRICT_ANSI__ -ffunction-sections -fdata-sections -fno-exceptions -Wall -DPLATFORMIO=40305 -DESP8266 -DARDUINO_ARCH_ESP8266 -DARDUINO_ESP8266_WEMOS_D1MINI -DF_CPU=80000000L -D__ets__ -DICACHE_FLASH -DARDUINO=10805 -DARDUINO_BOARD=\"PLATFORMIO_D1_MINI\" -DFLASHMODE_DIO -DLWIP_OPEN_SRC -DNONOSDK22x_190703=1 -DTCP_MSS=536 -DLWIP_FEATURES=1 -DLWIP_IPV6=0 -DVTABLES_IN_FLASH "-I/home/max/.platformio/lib/Brzo I2C_ID335/src" -I.pio/build/d1_mini/core -I/home/max/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/tools/sdk/include -I/home/max/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/tools/sdk/libc/xtensa-lx106-elf/include -I/home/max/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266 -I/home/max/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/tools/sdk/lwip2/include -I/home/max/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/variants/d1_mini "/home/max/.platformio/lib/Brzo I2C_ID335/src/brzo_i2c.c"
/home/max/.platformio/lib/Brzo I2C_ID335/src/brzo_i2c.c: In function 'brzo_i2c_write':
/home/max/.platformio/lib/Brzo I2C_ID335/src/brzo_i2c.c:72:2: error: cannot find a register in class 'RL_REGS' while reloading 'asm'
   72 |  asm volatile (
      |  ^~~
/home/jhdtl/.platformio/lib/Brzo I2C_ID335/src/brzo_i2c.c:72:2: error: 'asm' operand has impossible constraints
*** [.pio/build/d1_mini/lib8be/Brzo I2C_ID335/brzo_i2c.c.o] Error 1
```

Only other mention is 4 years old, https://github.com/jcmvbkbc/gcc-xtensa/issues/7, but changing `-O{0,1,2,3,s}` does not have any effect.

**Comment by devyte** [comment]
@mcspr The brzo lib is licensed under GPL3, which is incompatible with the closed source libs in this repo, except for personal projects that are not distributed to others (that rules out commercial applications). The author of brzo has refused to work with us to resolve this, so we can't justify spending time or effort on addressing issues related to that code. I suggest bringing it up in that repo and working with the author there.

**Comment by earlephilhower** [comment]
IIRC, `asm` is no longer a valid keyword.  Try `__asm__` instead.

**Comment by mcspr** [comment]
> I suggest bringing it up in that repo and working with the author there.

Sure, done.

I just wonder what exactly might have changed between 4.8.2 and 10.1.0, brzo builds just fine with the current release toolchain version.
Just poking around the asm (as I have almost zero experience with that, barring reading it + objdump decode), the issue seems to be with the amount of registers set at the bottom of the `asm` block. Removing a random one seems to fix the cryptic error gcc gives out.

> IIRC, `asm` is no longer a valid keyword. Try `__asm__` instead.

I was wondering about that... https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc/Alternate-Keywords.html#Alternate-Keywords, it does seem to only matter with -std=c99? I only see that when changing from gnu99 - `error: 'asm' undeclared (first use in this function)` - but it does work by default.

**Comment by earlephilhower** [comment]
> I was wondering about that... https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc/Alternate-Keywords.html#Alternate-Keywords, it does seem to only matter with -std=c99? I only see that when changing from gnu99 - error: 'asm' undeclared (first use in this function) - but it does work by default.

We're building with `-std=c17` and `-std=gnu++17` so that may be the cause then.  We're hoping to make use of the new C++17 stuff, so I think you'll need to do the `asm`->`__asm__` rename.  Nothing inside the asm block should need to change, that I can think of.

**Comment by mcspr** [comment]
> We're building with `-std=c17` and `-std=gnu++17`

That explains it, see `git grep -- '-std'`. PIO does not use platform.txt (given how much this come up I wish it did, somehow).

Also check out this entry from [`-std=` option docs:](https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc/C-Dialect-Options.html)
> `gnu17` "GNU dialect of ISO C17. This is the default for C code."

Changing `__asm__` does not fix anything btw, as mentioned, since the error complains about the asm 'constrain' ("r", "+r", etc.? reading from https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Extended-Asm)

**Comment by earlephilhower** [comment]
Ah, sorry then.  Looks like there will be some porting.  Check the register classes in the MD, it may have been renamed in mainline GCC from whatever the old Espressif builds were using.    https://github.com/gcc-mirror/gcc/tree/master/gcc/config/xtensa  

It may be as simple as renaming to whatever GCC moved to, which at first glance looks to br `AR_REGS`.

File: cores/esp8266/Arduino.h [hunk 1]
```diff
@@ -127,14 +127,8 @@ void timer0_isr_init(void);
 void timer0_attachInterrupt(timercallback userFunc);
 void timer0_detachInterrupt(void);
 
-// undefine stdlib's abs if encountered
-#ifdef abs
-#undef abs
-#endif
-
-#define abs(x) ((x)>0?(x):-(x))
+// Use stdlib abs() and round() to avoid issues with the C++ libraries
```
mcspr, line 130: *Sidenote:* since this is milestoned 3.0.0, can we push this change separately for 2.7.2?
Just noticed after switching back to master and also esp32 discrepancy https://github.com/espressif/arduino-esp32/commit/2f249edb8e2627421be87c43167a187e55ed7db7

btw I think `using abs;` is missing too?

earlephilhower: @d-a-v can comment more, as I think he tried this before and got nothing but grief from users of 3rd party libs about the change, but I think we had compatibility issues.  `std::abs` is very fussy about the operand types exactly matching, whereas the macro "just works".

If you'd really like it for 2.7.2, please open a separate issue to get his feedback.  My gut says it's too deep a change for a 3-dot release, but we'll see.

As for the missing `using std::abs;`, let me double check and update if needed.  Thx!

earlephilhower: Thx, fixed.

mcspr: I think I goofed up researching the issue. As full disclosure, [I was trying out building the test/host framework out of the tree, so I end up including mock Arduino.h....](https://github.com/mcspr/rpnlib/blob/master/examples/host/CMakeLists.txt) Which has quite different contents :)
Something to also note when running test suites!

earlephilhower: OK, if there's a host mock problem do open a bug on that so we can fix it, too!

earlephilhower:
I'll have to undo the `using std::abs`.  The problem is if you include `<stdlib.h>` it already includes the `extern int abs(int x);` for the standard lib.  Having the `using std::abs` and this `extern` definition causes GCC to barf on code with errors like:
````
/home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/SPI/SPI.cpp: In member function 'void SPIClass::setFrequency(uint32_t)':
/home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/SPI/SPI.cpp:263:38: error: call of overloaded 'abs(uint32_t)' is ambiguous
  263 |                 if(abs(freq - calFreq) < abs(freq - bestFreq)) {
      |                                      ^
In file included from /home/travis/arduino_ide/hardware/esp8266com/esp8266/tools/sdk/libc/xtensa-lx106-elf/include/stdlib.h:11,
                 from /home/travis/arduino_ide/hardware/esp8266com/esp8266/cores/esp8266/Arduino.h:27,
                 from /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/SPI/SPI.h:24,
                 from /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/SPI/SPI.cpp:22:

/home/travis/arduino_ide/hardware/esp8266com/esp8266/tools/sdk/libc/xtensa-lx106-elf/include/stdlib.h:72:12: note: candidate: 'int abs(int)'
   72 | int _EXFUN(abs,(int));
      |            ^~~
/home/travis/arduino_ide/hardware/esp8266com/esp8266/tools/sdk/libc/xtensa-lx106-elf/include/_ansi.h:65:30: note: in definition of macro '_EXFUN'
   65 | #define _EXFUN(name, proto)  name proto
      |                              ^~~~

In file included from /home/travis/build/esp8266/Arduino/tools/xtensa-lx106-elf/xtensa-lx106-elf/include/c++/10.1.0/cstdlib:77,
                 from /home/travis/build/esp8266/Arduino/tools/xtensa-lx106-elf/xtensa-lx106-elf/include/c++/10.1.0/bits/stl_algo.h:59,
                 from /home/travis/build/esp8266/Arduino/tools/xtensa-lx106-elf/xtensa-lx106-elf/include/c++/10.1.0/algorithm:62,
                 from /home/travis/arduino_ide/hardware/esp8266com/esp8266/cores/esp8266/Arduino.h:231,
                 from /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/SPI/SPI.h:24,
                 from /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/SPI/SPI.cpp:22:
/home/travis/build/esp8266/Arduino/tools/xtensa-lx106-elf/xtensa-lx106-elf/include/c++/10.1.0/bits/std_abs.h:79:3: note: candidate: 'constexpr long double std::abs(long double)'
   79 |   abs(long double __x)
      |   ^~~
/home/travis/build/esp8266/Arduino/tools/xtensa-lx106-elf/xtensa-lx106-elf/include/c++/10.1.0/bits/std_abs.h:75:3: note: candidate: 'constexpr float std::abs(float)'
   75 |   abs(float __x)
      |   ^~~
.....
````

So, back to the "You use `abs`, you get stdlib's `abs` and not a macro from 1979.  If you want float, use `fabs` or `std::abs` in your code."  This will probably still cause grief later on, but we'll work through it...

**Comment by mcspr** [comment]
SPI implementation is wrong though? It tries to use abs with unsigned, quoting
https://gcc.gnu.org/gcc-7/porting_to.html#cmath
https://gcc.gnu.org/gcc-6/porting_to.html#overloaded-abs
> Additionally, calling abs with an argument of unsigned type is now ill-formed after inclusion of any standard abs overload."

https://github.com/esp8266/Arduino/blob/e70092ca3e7a974f268521d85833ac58187f4649/libraries/SPI/SPI.cpp#L201
https://github.com/esp8266/Arduino/blob/e70092ca3e7a974f268521d85833ac58187f4649/libraries/SPI/SPI.cpp#L234
```cpp
#include <type_traits>
#include <cstdint>

int main() {
    int32_t a;
    uint32_t b;
    static_assert(std::is_same<uint32_t, decltype(b - a)>::value, "promotion");
}
~  
```

platform.txt
```diff
@@ -53,7 +53,7 @@ compiler.libc.path={runtime.platform.path}/tools/sdk/libc/xtensa-lx106-elf
 compiler.cpreprocessor.flags=-D__ets__ -DICACHE_FLASH -U__STRICT_ANSI__ "-I{compiler.sdk.path}/include" "-I{compiler.sdk.path}/{build.lwip_include}" "-I{compiler.libc.path}/include" "-I{build.path}/core"
 
 compiler.c.cmd=xtensa-lx106-elf-gcc
-compiler.c.flags=-c {compiler.warning_flags} -Os -g -Wpointer-arith -Wno-implicit-function-declaration -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals -falign-functions=4 -MMD -std=gnu99 -ffunction-sections -fdata-sections {build.exception_flags} {build.sslflags}
+compiler.c.flags=-c {compiler.warning_flags} -std=c17 -Os -g -Wpointer-arith -Wno-implicit-function-declaration -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals -falign-functions=4 -MMD -std=gnu99 -ffunction-sections -fdata-sections {build.exception_flags} {build.sslflags}
```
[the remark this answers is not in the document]
earlephilhower: Thanks, I believe that should now be fixed.  I can't test platform.io on my own (never was able to gett it running on either Linux or Windows), so I'll keep an eye on the CI system and see if something gets broken.

earlephilhower:
> SPI implementation is wrong though? It tries to use abs with unsigned, quoting
> https://gcc.gnu.org/gcc-7/porting_to.html#cmath
> https://gcc.gnu.org/gcc-6/porting_to.html#overloaded-abs
> > Additionally, calling abs with an argument of unsigned type is now ill-formed after inclusion of any standard abs overload."
> https://github.com/esp8266/Arduino/blob/e70092ca3e7a974f268521d85833ac58187f4649/libraries/SPI/SPI.cpp#L201
> https://github.com/esp8266/Arduino/blob/e70092ca3e7a974f268521d85833ac58187f4649/libraries/SPI/SPI.cpp#L234

I think you might have misread the code.  It's doing `abs(int32_t)` == `abs(signed type)` so all is well, no?

Plus, right now, `abs()` is the POSIX `abs()` https://pubs.opengroup.org/onlinepubs/009695399/functions/abs.html
(as a side note, do *NOT* do a Google search for `man abs`....completely different!  😆)

platform.txt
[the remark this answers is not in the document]
mcspr: as-is CI will fail, toolchain package version is still 4.8.2
tests/platformio.sh does some tricks to point to the Travis clone instead of trying to fetch things, so it needs a similiar thing that will point it to the toolchain, since it is hardcoded in the 'platform' files and platformIO's esp8266 'platform' is described outside of this repo.

I'd guess safe approach would be to
- get.py
- copy existing ~/.platformio/packages/toolchain-xtensa/package.json
- remove existing toolchain dir, replace with what get.py did get
- move package.json back
- pio ci, pio run should use it

mcspr:
> if(abs(freq - calFreq) < abs(freq - bestFreq)) {

using std::abs; forces it do try overload for both, failing for both at the same time. It just did not have any options earlier.

What I am saying though, abs is pointless if `freq - calFreq` and `freq - bestFreq` expressions result is already unsigned?

platform.txt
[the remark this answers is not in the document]
earlephilhower: Wow, looks like I have to get PIO running in a VM to figure out how to automate that in the test script.  Will be a bit of work, but I'll give it a try.

Jason2866:
If you rebuild your toolchain with the json file (you made the changes already)
you just need to add to platformio.ini
```
platform                  = espressif8266@2.5.1
platform_packages         = toolchain-xtensa @ https://github.com/earlephilhower/esp-quick-toolchain/releases/download/new_build_xxx/x86_64-linux-gnu.xtensa-lx106-elf-38230f5.210520.tar.gz
```
to use your toolchain

mcspr:
`pio ci` does not use platformio.ini (at least by default, there is `-c / --project-conf` option), why I initially thought about silently changing toolchain-dir.

It generates temporary .ini for each example, so we can use [`-O "key = value"`](https://docs.platformio.org/en/latest/core/userguide/cmd_ci.html#cmdoption-platformio-ci-o):
```
$ pio ci -O "toolchain-xtensa @ https://github.com/earlephilhower/esp-quick-toolchain/releases/download/new_build_xxx/x86_64-linux-gnu.xtensa-lx106-elf-38230f5.210520.tar.gz" <the rest>
```
So the change would be to parse ../package/package_esp8266com_index.template.json / import get.py and get the download link?

Jason2866:
@mcspr  ahh, yes CI. Never did a lot with.

earlephilhower:
@mcspr  and @Jason2866 , I've hacked in the toolchain update in the PIO CI tests.  Can you take a gander and see if there's something else you would suggest doing vs just overwriting the toolchain as the `tests/platformio.sh` does now.

Gizmodo:
@d-a-v When do you plan to merge PR? I can't wait for C++17

devyte:
You can check the milestone for issues and PRs on the right. This is targeted for 3.0.0. That means it will be merged after 2.7.2, which is very close at hand.

d-a-v:
@Gizmodo it is included in the [alpha release](https://d-a-v.github.io)

Jason2866:
@Gizmodo  
I did a PlatformIO setup for testing the Project Tasmota with GCC10.1
You cant use it with the entry in Platformio.ini 
`platform                  = https://github.com/Jason2866/platform-espressif8266.git#new_gcc`
You can take a look at:  https://github.com/Jason2866/Tasmota/tree/Gcc10.1

hreintke:
@earlephilhower 

There is a difference between the current an gcc9 toolchain in the .d files that are created.
These .d files are used by make in the build process.

Current : 
```
T1.cpp.o: ..\T1.cpp ..\T1.h C:\temp\ArduinoC11\cores\esp8266/Arduino.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/stdlib.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/machine/ieeefp.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/_ansi.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/newlib.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/sys/config.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/sys/features.h \
 C:/temp/ArduinoC11/tools/sdk/libc/xtensa-lx106-elf/include/xtensa/config/core-isa.h \
```
gcc9
```
T1.cpp.o: ..\T1.cpp ..\T1.h C\:\temp\ArduinoC17\cores\esp8266/Arduino.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/stdlib.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/machine/ieeefp.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/_ansi.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/newlib.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/sys/config.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/sys/features.h \
 C\:/temp/ArduinoC17/tools/sdk/libc/xtensa-lx106-elf/include/xtensa/config/core-isa.h \
```
There is an additional `\` after the first `C`

In my eclipse sloeber development environment that results in not being able to do incremental builds.

```
           Finished prerequisites of target file '..\T1.h'.
          No recipe for '..\T1.h' and no prerequisites actually changed.
          No need to remake target '..\T1.h'.
          Considering target file 'C\:\temp\ArduinoC17\cores\esp8266/Arduino.h'.
           File 'C\:\temp\ArduinoC17\cores\esp8266/Arduino.h' does not exist.
           Finished prerequisites of target file 'C\:\temp\ArduinoC17\cores\esp8266/Arduino.h'.
          Must remake target 'C\:\temp\ArduinoC17\cores\esp8266/Arduino.h'.
make: *** No rule to make target 'C\:\temp\ArduinoC17\cores\esp8266/Arduino.h', needed by 'T1.cpp.o'.  Stop.
"C:/temp/Sloeber/arduinoPlugin/tools/make/make --debug=verbose all" terminated with exit code 2. Build might be incomplete.
```

Is there a way to revert to the old format ?

earlephilhower:
That's pretty strange.  It's not a dependency format so much as an escaping of paths issue.  I doubt there's anything you can do to make GCC not escape the colon in makefiles.

What make version do you have installed?  Looking thru gnu.org I can see in gcc 4.x has an `unescape_char(...,':')` in the depfiles parser, but that is not present in the 3.x branch.  No guarantees this fixes your problem, but it does look very promising.

hreintke:
Sloeber eclipse has make 3.8.2 in it's distribution. 
In contact with him now probably moving to 4.2.1 which doesn't have problems with the new format.

Thanks for pointing in the right direction.

earlephilhower:
Here goes nothing...

mhightower83:
FWIW: The optimization does appear to have been improved. Code that constructs a slightly obscured divide by 0, like some examples do to show the exception processing, may now report an HWDT event instead. In my case, the 10.1 compiler figured it out and replaced it with a `break 1, 15`. My workaround was to add a noinline to my local divide a / b function that I call to generate the exception.

PR #6300, opened 2019-07-14
Add a dump of received FP and CERT when in debug mode

earlephilhower:
To simplify BearSSL debugging, print the received FP (when it doesn't
match the expected) and the binary certificate (always), when in debug
mode.


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -676,7 +683,24 @@ extern "C" {
     char res[20];
     br_sha1_out(&xc->sha1_cert, res);
     if (xc->match_fingerprint && memcmp(res, xc->match_fingerprint, sizeof(res))) {
+#ifdef DEBUG_ESP_SSL
       DEBUG_BSSL("insecure_end_chain: Received cert FP doesn't match\n");
+      char buff[3 * sizeof(res) + 5];
+      buff[0] = 0;
+      for (size_t i=0; i<sizeof(res); i++) {
+        char hex[6];
```
devyte: Why 6? What is the max possible output of sprintf below? What about nullterm? Shouldn't snprintf and strncat be used instead?

I suspect the max output of sprintf is 4 bytes: 2 hex digits a space and a nullterm. Despite knowing the max size, it's considered best to move towards de "n" versions of the classic functions.

earlephilhower: Updated in latest push.

earlephilhower:
Ah, crap.  `strncat` is one of those funny functions that I remember MSFT warning us about.  The `n` there is the max # of elements of the **src**, not the total length.  Let me update PR with `strlcat` which does what a sane person would want it to do...

PR #6301, opened 2019-07-14
Added memory fence to xt_rsil(). Without this the compiler may use

mhightower83:
memory references loaded to registers before the fence, in computation
within the fence. These values could have changed before xt_rsil()
(critical section start) was called.

addresses: https://github.com/esp8266/Arduino/pull/6274#issuecomment-511173751


mhightower83:
Fixes https://github.com/esp8266/Arduino/issues/6302#issue-467875288

PR #6307, opened 2019-07-16
Fix for scheduling recurrent functions while inside scheduled function

dok-net:
Dear @d-a-v,
I hope to have come up with a relatively small fix to Schedule that allows scheduling from inside already scheduled functions. I am using this with my CoopTask scheduling, without this patch, the sketch crashed, with it, it performs correctly. I don't see any reason for side-effects, like poorer performance. Memory use is minimally increased by the use of a recurrent_fn_t instance for rFirst instead of the previous pointer, but the logic looked cleaner to me when I began the patch, and I'd rather not change it back if there's no compelling reason.
Could you please review this?

### DIFF
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -27,7 +27,7 @@ struct recurrent_fn_t
     recurrent_fn_t (esp8266::polledTimeout::periodicFastUs interval): callNow(interval) { }
 };
 
-static recurrent_fn_t* rFirst = nullptr; // fifo not needed
+static recurrent_fn_t rFirst(0); // fifo not needed
 
 // Returns a pointer to an unused sched_fn_t,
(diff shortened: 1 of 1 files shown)


dok-net:
Just food for thought, after concerning myself with #6212. Is there really any power-saving on the ESP8266 by leaving "cont", like delay() does with the timer and yielding to sys? I was under the impression that all the power-saving modes come with some limition, such that the user sketch must decide what it wants to do. On the other hand, leaving "cont" like in delay() with a millisecond resolution, but using microsecond times in Schedule.h, is that sensible? I see between 20-30µs cycle times in loop in tests of my coop scheduler, which doesn't block the whole user context for delays in separate tasks.

If so, maybe the recurrent scheduling loop could check for the delay until the next timed function and "sleep"?

dok-net:
Superseded by #6485

Issue #6269, opened 2019-07-07
new operator w/o (std::nothrow) aborts, when "Exceptions: disabled"


mhightower83:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in the current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					**ESP-12**
- Core Version:			[76cda9bdbccd00c02d69119fea9dee34f05cbd9a]
- Development Env:		**Arduino IDE**
- Operating System:		**Ubuntu**

### Settings in IDE

- Module:  			**Adafruit HUZZAH**
- Flash Mode:			**qio**
- Flash Size: 		**4MB**
- lwip Variant:   **v2 Lower Memory**
- Reset Method:		**nodemcu**
- Flash Frequency:	**40Mhz**
- CPU Frequency:		**80Mhz**
- Upload Using:		**SERIAL**
- Upload Speed:   **115200**(serial upload only)

- Exceptions:   **Disabled**

### Problem Description

With "Exceptions: **Disabled**" selected from the Arduino IDE.
A build using the _new_ operator without `(std::nothrow)` causes a crash, when out of memory.
Fails with "Abort called" and stack trace.
No opportunity to test for NULL. 

This test sketch also fails with Arduino ESP8266 Core 2.5.0; however, it works (returns NULL) in Core 2.4.2.

And before rerunning each test, I did a `git pull` against the branch and ran `./get.py` from the `tools` directory.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

void setup() {
  Serial.begin(115200);
  delay(20);
  Serial.printf("\nUp and running ...\n");
//  char *p = new (std::nothrow) char[128000];  // This will return NULL
  char *p = new char[128000];  //This fails with "Abort called" and stack trace

  if (p == NULL) {
    Serial.printf("\n\"new\" failed as it should! PASS!\n");
  } else {
    Serial.printf("\n\"new\" worked, it should not have!, FAIL!\n");
  }
}

void loop() {
}

```

### Debug Messages

```
Up and running ...

Abort called

[...]

PR #6309, opened 2019-07-17
operator new: no exception without need for std::nothrow

d-a-v:
Overloads `operator new` so it can return `nullptr` instead of throwing an exception without need for `(std::nothrow)`.

Requirement: `<Arduino.h>` must be included prior to any c++ header file

addresses #6269 @mcspr @mhightower83

### DIFF
--- a/cores/esp8266/Arduino.h
+++ b/cores/esp8266/Arduino.h
@@ -33,24 +33,32 @@ extern "C" {
 #include <string.h>
 #include <math.h>
 
-#if defined(__cplusplus) && !defined(__cpp_exceptions)
+#if defined(__cplusplus) \
+ && !defined(__cpp_exceptions) \
+ && !defined(_NEWHACK) \
+ && !defined(CORE_MOCK)
 // this is a hack working with gcc4.8
 // Overload operator new so it can return nullptr instead of throwing an
 // exception without (std::nothrow)`
-// Requirement (or the constructor is called even on nullptr)
+// Requirements: (or the constructor is called even on nullptr)
 // - "noexcept" is needed
 // - "#include <new>" must not be done *before* these definitions
+#define _NEWHACK
+#ifdef _NEW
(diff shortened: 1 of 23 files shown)


earlephilhower:
Interesting solution to the problem.

The standard libraries do not ever check the return value of new() != nullptr, they use a `catch` or they expect their code to stop and your upper lever code to have that catch().  There has been more than one bug where we find std::shared_ptr or some other compound type (which has elements that are new()ed) crashes with a stack trace after doing a new and trying to use the returned object.

Theoretically, it seems like this would let the std:: libs abort() on the failed new (since they should use the base new() and it will throw), and let user code use the broken nothrow new().

cores/esp8266/Arduino.h
```diff
@@ -33,6 +33,25 @@ extern "C" {
...
+// this is a hack working with gcc4.8
+// Overload operator new so it can return nullptr instead of throwing an
+// exception without (std::nothrow)`
+// Requirement (or the constructor is called even on nullptr)
+// - "noexcept" is needed
+// - "#include <new>" must not be done *before* these definitions
+#include <bits/c++config.h>
+extern "C++" inline void* operator new (std::size_t size) noexcept
+{
+    return malloc(size);
```
earlephilhower: Please check out the old new() and new[]() override, it sets the last-failed-alloc address which this needs to do as well:
https://github.com/esp8266/Arduino/blob/bb28d4a3927758b0b345579792cdeebbc9e2e6a3/cores/esp8266/abi.cpp#L31-L49

d-a-v:
This is a hack.

Remove `noexcept` or move `#include <new>` before these definitions, and constructors will be called no matter what.

earlephilhower:
For reference the prior new override was handled here:
https://github.com/esp8266/Arduino/blob/2.4.2/cores/esp8266/abi.cpp

PR #6320, opened 2019-07-19
Expose AP list clearing method

dav1901:
Expose AP list clearing method in ESP8266WiFiMulti so that it would be possible to clear the list without calling the destructor and also expose the private functions and remove their useless wrappers.


earlephilhower:
Thanks for the PR, but we already have #4107 is up for 2.6.0 possible inclusion with the AP list clearing API.

The additional changes you've done will, unfortunately, cause errors with users who have been using these calls, and so would make this a breaking change.

Issue #4639, opened 2018-04-12
Incorrect expansion of ,##__VA_ARGS__ in macro


wmacevoy:
### Basic Infos

- [ X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ?] I have tested that the issue is present in current master branch (aka latest git).
         Latest build from README docs build failed (Mac OS 10.12.6) on python get.py with:
         > cd esp8266/tools
         > python get.py
         IOError: [Errno socket error] [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:590)
- [ X] I have searched the issue tracker for a similar issue.
- [ X] If there is a stack dump, I have decoded it.
- [X ] I have filled out all fields below.

#### Platform

- Hardware:				Generic ESP8266 device
- Core Version:			2.4.1 - could not build from git
- Development Env:		Arduino IDE 1.8.5
- Operating System:		MacOS

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			qio
- Flash Size: 		512MB
- lwip Variant:   v2mss536
- Reset Method:		ck
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		na compile time error
- Upload Speed:   na compile time error

### Problem Description

g++ supports variadic macros which cleanly reduces to 0 arguments with the notation

```
#define va_macro(...) va(begin, ##__VA_ARGS__, end)
```
The `,##__VA_ARGS__`  notation expands to nothing if no arguments are given (the leading comma is removed), while the comma remains for 1 or more arguments.

This works correctly on uno, mega and esp32 boards, but fails to expand correctly on esp8266.  This notation is important for variable parameter macros.  Here is a sketch that compiles under uno, mega, esp32, but has a compile-time-error for esp8266, with log files for uno, mega, esp32 (correct) and esp8266 (compile time error):

https://github.com/mmurdoch/arduinounit/tree/pr65/va_args

```cpp

#define va_macro(...) va_sub(-9,##__VA_ARGS__,9)

// Unimportant, just to have something the va_macro can compile to
void va_sub(int first, int last) { }
void va_sub(int first, int va1, int last) { }
void va_sub(int first, int va1, int va2, int last) {}

[...]

PR #6333, opened 2019-07-23
exceptions: optionally enforce c++ standards

d-a-v:
#6312 restored "arduino `new`" 's right to return `nullptr`

This PR ***optionally*** allows to enforce C++ standards by making `new` call `abort` in case of OOM.

It also introduces an *optional* allocator to allow a clean behaviour in case of oom.

check maintainers' effort to [document `new` without exceptions](https://github.com/d-a-v/Arduino/blob/new0/doc/reference.rst#c) .


d-a-v:
Benefits brought by the new allocator: [example output](https://github.com/d-a-v/Arduino/blob/fdaa27c86a570210ff4b96f6b562912ab400e845/libraries/esp8266/examples/arduino_new/arduino_new.ino#L95)

TD-er:
> infinite loop of the new discussion.

That's a true statement in itself :)

d-a-v:
Thanks @earlephilhower for reminding #4639 (allowing the last commit)

PR #6338, opened 2019-07-24
Update WiFiServer docs for ::write(all clients)

### LINKED ISSUE #2743: WiFiServer::write() write to all clients is not implemented
### Basic Infos
So I tried searching if this issue is already present but due to the mass it could be that I overlooked it, so apologies if this is a duplicate report.

### Description
The WiFiServer currently doesn't support writing to all clients. Instead one has to keep track of the client instances and do that manually which leads to a bit more ugly code.

See: [WiFiServer.cpp#L135](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/WiFiServer.cpp#L135)

I would file a pull request but since I am no C++ expert, the list code is hard to read.
In general I would recommend refactoring the code to be in line with the [Arduino's EthernetServer](https://github.com/arduino-libraries/Ethernet/blob/master/src/EthernetServer.cpp).

The key difference is that it handles the client array internally, hiding it from the user.
This means that `available` always returns the first readable client instead of only returning new connections like `WiFiServer` does.

I agree that this is a design decision and that it would break exisiting code, so this is soley a proposal, but since I had to rewrite some code depending on the Platform (Arduino Mega vs ESP it would be awesome to have that consistent.

earlephilhower:
Fixes #5116
Fixes #2743

The Arduino WiFiClient object allows sending the same packet to all
connected clients of a WiFiServer.  In their implementation it may make
sense, but on the 8266 with things like SSL it doesn't.

Update the docs to note that WiFiServer::write() is a no-op, and that
the app should use the WiFiCliebnt::write() on all connected clients as
appropriate.


jandrassy:
why implement Arduino Server class at all? there is no use case where a server object is useful as Server.

d-a-v:
@jandrassy 
WiFiServer is needed to create new WiFiClient connections associated to a server port (like http/80).
There is a `::write()` method which is supposed to send data to all currently connected client, that we don't implement.
Maybe I didn't understand your comment.

jandrassy:
I recently made a generic ArduinoOTA library for AVR, SAMD, nRF5, esp8266, esp32 for many different networking libraries (for example to upload to esp8266 over Ethernet). The code of the library interacts with client object as Client class. And there is a server to receive the binary, but there was no use for Server base class. it has only `begin()`, but can't have `available()` because the return type of `available()` is the class derived from Client class. So there is no real use of the Server class as base class. I used C++ template to work with actual server class ([ArduinoOTA.h](https://github.com/jandrassy/ArduinoOTA/blob/master/src/ArduinoOTA.h))

Now I write a new WiFiEsp library (Arduino WiFi API over AT commands of connected esp8266). I decided not to implement the WiFiServer as Server in this library. Then I don't need to add dummy `write(b)` and can add optional parameters to `begin()`. And all use cases are still supported, except of the never used "write to all clients" feature.

TL;DR Arduino API Server class is useles

Issue #5987, opened 2019-04-15
Append to file on SPIFFS does not return actual bytes writen


TD-er:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[ESP82xx Core 2.6.0-dev, NONOS SDK 2.2.2-dev(c0eb301), LWIP: 2.1.2 PUYA support]
- Development Env:		[Platformio]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA|SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

I know that SPIFFS may get fragmented, therefore I have a check to see if the `write` to a file (append mode) on SPIFFS may return less than the number of bytes I want to write to it.
However, this check was never hit in my sketch.
I now have a check for file size before and after the write and this does show the bytes are not appended to my file when the SPIFFS is out of free sectors to write to.

```c++
        size_t filesize = fw.size();
        int bytesWriten = fw.write(&RTC_cache_data[0], RTC_cache.writePos);
        if (bytesWriten < RTC_cache.writePos || fw.size() == filesize) {
          String log = F("RTC  : error writing file. Size before: ");
          log += filesize;
          log += F(" after: ");
          log += fw.size();
          log += F(" writen: ");
          log += bytesWriten;
          addLog(LOG_LEVEL_ERROR, log);
          fw.close();
          GarbageCollection();
          writeerror = true;
          return false;
        }
```

N.B. `fw` is a `File fw;`

This does result in errors like:
```
RTC  : error writing file. Size before: 720 after: 720 written: 240
```

I guess the `write` function should return the actual number of bytes written.
[...]

PR #6340, opened 2019-07-24
Add a FS::check() optional method

### LINKED ISSUE #2634: Add SPIFFS.check()  to implement checking capabilities.
In sensors log with casual power loss may happen that file-system stop working (especially remove()),
there is a check implementation in SPIFFS library that  should  exposed in Arduino.

earlephilhower:
Fixes #2634

Expose any low-level filesystem check operations for users, and add
documentation on this and the gc() methods.


doc/filesystem.rst
```diff
@@ -402,6 +402,29 @@ block size - ``pageSize`` — filesystem logical page size - ``maxOpenFiles``
...
+gc
+~~
+
+.. code:: cpp
+
+    SPIFFS.gc()
+
+Only implemented in SPIFFS.  Performs a quick garbage collection operation on SPIFFS,
+possibly making writes perform faster in the future.  Not normally needed by applications
+as SPIFFS will take care of things itself on writes.
```
TD-er: Not only perform faster, but even allow for a new block allocation which could fail otherwise.

What it does:
It does search for blocks which have all occupied sectors mapped to files which are marked for deletion.
If such a block is found, it will erase that block.

earlephilhower: Wow, that sounds like a bug in SPIFFS.  Shouldn't it be keeping track of used but no longer needed blocks and erase them as needed?  Garbage collection is normally something handled behind the scenes (Java, SSDs, etc.) and not something users need to explicitly worry about.

TD-er: I don't know if it is a bug in SPIFFS or me just trying to use SPIFFS in a way it was never intended to :)
My use case:
I try to append to a file (until some size to prevent file handling becoming too slow) and if appending fails I will throw away the oldest file of that type.
Some sort of circular buffer in files on SPIFFS.
If I do not call the GC on a highly fragmented filesystem, I may need to throw away a lot more files before I find a block to write to.
To my understanding, SPIFFS only runs a limited number of loops to call the GC and it then only will check a limited number of blocks.
A next call to the gc will check the next N blocks etc.
So it may be that a number of calls to gc yields a different result compared to only calling it once.
SPIFFS does do its administration on a bit lower priority, so if you call for a delete the space is not freed immediately.
This is why I added the call to gc from the higher level wrapper.
Calling the (SPIFFS) gc function with a number other than 0 may even try to re-allocate sectors if there are max. N sectors not marked to be deleted in a single block.
But since that's a bit tricky (and also remarkably slow on a fragmented filesystem), I did not add the integer parameter. Also the parameter is really SPIFFS specific and some other gc function on another filesystem may need a completely different integer value.

Also there is still the bug that the actual write function may return a success which later appears to be unsuccessful and you can only check it by checking the file size (if appending)

earlephilhower: That's pretty bogus, reason enough to go to LittleFS. :)    What wording would you suggest here?  Fixing SPIFFS is a bit larger than I'd planned with this PR...

TD-er: It was by no means my intention to make this PR into a universal fix for SPIFFS :)
Just some extension to the documentation.
Like I said, my use case is probably not how SPIFFS was intended to be used.
And for the bug of `write` not returning the actual amount of data written (in some cases) is already [another issue report](https://github.com/esp8266/Arduino/issues/5987). (which I maybe should move/copy to the SPIFFS repo instead of here)

earlephilhower: Looking at the **comment** here, it seems like this would be a SPIFFS buf:
https://github.com/pellepl/spiffs/blob/ec68ba8208d7550860e4e78299d58a529b88bf85/src/spiffs.h#L640-L642 .

Have you found that doing a _gc_fast() actually fixes anything?  If so, we can note it here.  OTW, I'm not really sure what to say for _gc().  "It *might* help if SPIFFS is reporting out of space errors on a non-full filesystem." ?

TD-er: > Have you found that doing a _gc_fast() actually fixes anything? If so, we can note it here. OTW, I'm not really sure what to say for _gc(). "It _might_ help if SPIFFS is reporting out of space errors on a non-full filesystem." ?

I do use it on my nodes that collect data using my "cache controller" in ESPeasy.
It is used there right after deleting a file or when writing fails. (and at boot)
Without calling the gc, the process sometimes got stuck, probably because there were no 2 erased blocks available anymore.

Also on nodes with only 128k SPIFFS it can happen that settings files cannot be changed or become corrupted (64k file, updated in chunks).
After adding the call to the garbage collection at boot, I have not seen it happen anymore.
So at least that issue was made a lot harder to reproduce :)

earlephilhower: OK, got it.  I've added a high-level note to that effect and a link to this discussion for users who need to delve further.  Thanks for the info!

Issue #6348, opened 2019-07-26
espota.py does not receive `OK` when Signed Update is used


thangktran:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware: [ESP-12]
- Core Version: [2.5.2]
- Development Env: [Arduino IDE]
- Operating System: [Ubuntu]

### Problem Description
When Signed Update is used, the `*bin.signed` is uploaded normally. However when the upload is done, `espota.py` did not receive any `OK` ACK and timeouted.

```
/home/ubuntu/.arduino15/packages/esp8266/tools/python/3.7.2-post1/python /home/ubuntu/.arduino15/packages/esp8266/hardware/esp8266/2.5.2/tools/espota.py -i 192.168.0.101 -p 49867 --auth=************ -f /tmp/arduino_build_503180/sketch_jul09a.ino.bin 
Signed Update is used. /tmp/arduino_build_503180/sketch_jul09a.ino.bin.signed will be uploaded instead.
Authenticating...OK
Uploading..................................................................................................................................................................................................................................................
11:38:33 [ERROR]: No Result!
```
The ESP-12 did however receive this update and execute the new update normally.

I believe the bug is somewhere in the Updater module that makes `Update.end()` return `false`
hence
https://github.com/esp8266/Arduino/blob/824a83347d206bdc875b2fa46df42422dcfbe713/libraries/ArduinoOTA/ArduinoOTA.cpp#L330-L333

I'm currently investigating this bug.

earlephilhower:
Looking at the Updater base class, it seems like the uboot is only ever told to do something at the very end:  
https://github.com/esp8266/Arduino/blob/824a83347d206bdc875b2fa46df42422dcfbe713/cores/esp8266/Updater.cpp#L266-L284
That path can only ever return true to the ArduinoOTA library.

[...]

PR #6351, opened 2019-07-27
libraries: fixed no "OK" ACK for Signed Update

thangktran:
Fixes: https://github.com/esp8266/Arduino/issues/6348


earlephilhower:
In the worst case, I think we can use this.  But let's see if @d-a-v has any better ideas for making sure the packets are all gone before reboot than a timeout.

Issue #6346, opened 2019-07-25
WiFi Station Mode Confusion


louislewis2:
Hi,

I have read the documentation around WiFi, I have also read numerous issues and could not work out from that what the "right" thing is, that I need to do.

I am using my own configuration file for storing WiFi credentials, along with many other system configuration values. I have my own process, of being able to change these from a configuration section. The units do deep sleep, and upon awakening they reconnect to the WiFi network configured. 

What is the best way to call into the exposed api's for this kind of setup?

Do I need to call WiFi.persistant(false) after each boot?
I do at the moment use WiFi.begin("ssid","password") to connect to the network saved in configuration, is this right or is there a better way?

I am also seeing strange occurrences which can be best described by this issue. Where regardless that the WiFi network has not changed, and the units are using the correct credentials, that they at some point simply stop connecting to the network. Only doing a full power cycle restores it.

I also see that sometimes the WiFi returns that the password is incorrect, at this point I remove the credentials from configuration and then requires the user to enter new credentials, however this happens at multiple locations, where the stations have never had their password changed?

https://github.com/esp8266/Arduino/issues/5527

I have more than 50 of these devices, all showing the same issues.
Any help, guidance on the matter will be greatly appreciated.

TD-er:
> Do I need to call WiFi.persistant(false) after each boot?

As far as I know, yes that's needed to make sure WiFi related information is not stored in flash and thus wear out that sector quite fast.

> I do at the moment use WiFi.begin("ssid","password") to connect to the network saved in configuration, is this right or is there a better way?

Not sure if it's considered to be better, but there is also an option to include the last known channel and BSSID (MAC address of AP)
This allows for faster reconnect, but it may fail if the AP has changed channel or in some setups with a repeater (and thus same SSID) where the option without given channel and BSSID may be able to connect.
Another issue which may be fixed with giving the last known channel and BSSID is that the node may start 'hopping' from AP to AP if it can see both the repeater and AP.

> I am also seeing strange occurrences which can be best described by this issue. Where regardless that the WiFi network has not changed, and the units are using the correct credentials, that they at some point simply stop connecting to the network. Only doing a full power cycle restores it.

I do see the same behavior every now and then.
In my project I have a setting to let the node perform a reboot after N failed connect attempts (connecting to a host, not AP) This does at least make the nodes available again.
I have tried several things to fix this. For example:
- Destruct and recreate WiFi object
[...]

Issue #6266, opened 2019-07-06
[WD reboot] Reconnect to WiFi AP cause WD reboot


TD-er:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12|other]
- Core Version:			[core 2.5.2 and core 2.6.0 latest]
- Development Env:		[Platformio|]
- Operating System:		[Windows|Ubuntu|]

### Settings in IDE

- Module:  			[Generic ESP8266 Module|]
- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory|Higher Bandwidth]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

This has been discussed also in other issues [like here](https://github.com/esp8266/Arduino/issues/5527#issuecomment-508260107), but since it does appear to be something different and also to keep the information in a single issue I thought it would be better to make a separate issue for it.

The problem is, for over a year now, I get lots of reports of WD reboots which are really hard to reproduce.
I now have found a way to reproduce it on my nodes using the disconnect feature of my (MikroTik) AP and narrowed it down to somewhere in the code used to (re)connect to WiFi.

There may still be several issues at stake here, and for sure it will not be the only reason for WD reboots, but I believe it is responsible for lots of them.

What I found is that the WD-reboots do happen right at the moment the WiFi connection transitions from "connected" to "Got IP".
The same transition does happen when static IP is being used.
This can happen at the first connect attempt or at any other reconnect, but it doesn't happen always.

The standard WiFi connection cycle can be seen as several stages:

- Connect to AP
- Authenticate connection => When successful, event "connected" is fired.
- Setup network configuration => When successful, event "got IP" is fired.

Every now and then the last step does halt the system for some reason. Not even the loop() function is called then and thus a WD-reboot.
[...]

PR #5703, opened 2019-02-01
stop lwIP dhcp client when WiFi goes off.

d-a-v:
fix #5667 

Bug is reproduced
This patch stops dhcp client when WiFi goes off,
dhcp client was anyway started by firmware when WiFi goes on.

Thanks @dalbert2 

*edit: bug no more appearing*

Issue #5667, opened 2019-01-25
Exception 0 in LWIP2 on DHCP renew with WIFI_OFF


dalbert2:
### Basic Infos

- [X ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [X ] I have searched the issue tracker for a similar issue.
- [X ] If there is a stack dump, I have decoded it.
- [X ] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[Core:2_4_2 SDK:2.2.1(cfd48f3) LWIP:2.0.3]
- Development Env:		[Platformio]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description
Exception 0: interaction between application context printf and interrupt context printf

Detailed problem description goes here.

I'm experiencing periodic Exception 0s and the stack trace shows it is happening during vsnprintf in the LWIP stack when it's handling a DHCP timeout/renew (probably when WiFi is off).  The exception address is always part of a printf format string from the application context.  The particular address/string isn't the same in each exception, but the stack trace always ends at _vsnprintf_r and the exception address is always part of an application Serial.printf_P format string.  In the example provided below, the exception address is 0x2C75253D which is ASCII "=%u," which only exists at one place in the code (see instruction below).

I first guessed that this might be a case of the app context stack overflowing into the ISR stack and tried increasing CONT_STACKSIZE to 6K but the exception still takes place.  Moreover, the stack dump isn't anywhere near 4K (it's 768 bytes deep in the stack dump below)...so it must be something else (how big is the interrupt stack?)

All of the application code is running in the application context (no application ISRs).  Any ideas would be much appreciated.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

[...]

PR #6356, opened 2019-07-29
Experimental: add new WiFi (pseudo) modes: WIFI_SHUTDOWN & WIFI_RESUME

d-a-v:
*edit*: This is an experimental addition to the API

Adds new WiFi (pseudo) modes: WIFI_SHUTDOWN & WIFI_RESUME,
also:
- an example
- restore WiFi.onWiFiModeChange()
- add crc32() in core
- cancel #5703 because of https://github.com/esp8266/Arduino/issues/5667#issuecomment-512180419 (related to all the issues referenced below)

references:
#6346 
#6266 
#6057 #5527

This is an attempt to put right into the core all the necessary code to wifi-sleep and deep-sleep so there is only one place to debug for everybody.
I am personally not able to reproduce all the WDT issues so even after hypothetical merge, we will need to bring fixes found by all the community, like  https://github.com/esp8266/Arduino/issues/6266#issuecomment-515803290 and https://github.com/esp8266/Arduino/issues/6172#issuecomment-500457407 (only citing them).

This PR does not change legacy behavior when using the regular API (well.. not supposed to :]).


libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -386,22 +394,36 @@ bool ESP8266WiFiGenericClass::getPersistent(){
...
+
+    if (wifi_fpm_get_sleep_type() != NONE_SLEEP_T) {
+        // wifi may have been put asleep by ESP8266WiFiGenericClass::preinitWiFiOff
+        wifi_fpm_do_wakeup();
+        wifi_fpm_close();
+    }
+
     if(_persistent){
         if(wifi_get_opmode() == (uint8) m && wifi_get_opmode_default() == (uint8) m){
+            saveState(state, m, _persistent);
```
TD-er: I don't get it.
If the mode is already the same as the one you want to set, then save the state?

libraries/ESP8266WiFi/src/include/WiFiState.h
```diff
@@ -0,0 +1,23 @@
...
+    struct
+    {
+        bool persistent;
+        WiFiMode_t mode;
+        uint8_t channel;
+        station_config fwconfig;
+        ip_info ip;
+        ip_addr_t dns[2];
+        ip_addr_t ntp[2];
+    } state;
```
TD-er: Ordering members from big to small will often lead to smaller struct sizes.
Also the members are not initialized in the constructor.
For example channel is only set if there was a successful connect and/or the active mode was WIFI_STA when shutdown and fwconfig is only set when in STA mode at shutdown.
But these are being used at wifi.begin.

I have not looked in much detail to see if the settings are somewhere read from flash, but I can imagine that would result in very hard to reproduce results.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
d-a-v: This is right.
saveState should be used only with shutdown (write all structure members) or resume mode (real). Fixed.

libraries/ESP8266WiFi/src/include/WiFiState.h
d-a-v: Members are initialized in shutdown only (now). The whole structure is valid only when crc is matching, and crc is calculated when all arguments are set.

libraries/ESP8266WiFi/src/ESP8266WiFiType.h
```diff
@@ -58,9 +59,10 @@ typedef enum WiFiEvent
     WIFI_EVENT_SOFTAPMODE_STACONNECTED,
     WIFI_EVENT_SOFTAPMODE_STADISCONNECTED,
     WIFI_EVENT_SOFTAPMODE_PROBEREQRECVED,
+    WIFI_EVENT_MODE_CHANGE,
```
devyte: Maybe swap these to keep the softapmode together

d-a-v: The order needs to be the same as [SDK defines](https://github.com/esp8266/Arduino/blob/0399bb1fee4b95bfb8f9007f561e4f7060eedcdb/tools/sdk/include/user_interface.h#L466)'s

cores/esp8266/crc32.cpp
```diff
@@ -0,0 +1,23 @@
+
+#include "coredecls.h"
+
+// taken from esp8266/examples/RTCUserMemory/RTCUserMemory.ino
+uint32_t crc32 (const void* data, size_t length)
```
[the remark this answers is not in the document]
d-a-v: There is no difference. But eboot is a standalone code and only uses SDK.
Sharing common source code with eboot should be addressed from another issue.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -88,6 +90,11 @@ class ESP8266WiFiGenericClass {
         bool forceSleepBegin(uint32 sleepUs = 0);
         bool forceSleepWake();
 
+        bool shutdown (uint32 sleepUs = 0, WiFiState* stateSave = nullptr);
```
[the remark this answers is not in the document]
d-a-v: Commented and moved to protected

PR #6371, opened 2019-08-01
Refactor HardwareSerial.cpp for consistent use of PolledTimeout

dok-net:
After quick review comment, both of the two places in HardwareSerial.cpp shall use PolledTimeout - fixed by this PR.


devyte:
Thanks @dok-net @d-a-v

PR #6372, opened 2019-08-01
Make PolledTimeout a library

dok-net:
This potentially reduces the BSP footprint, and more significantly make PolledTimeout porting to ESP32 and post-NONOS SDK ESP8266 running on FreeRTOS more straightforward.


libraries/PolledTimeout/library.properties
```diff
@@ -0,0 +1,9 @@
+name=PolledTimeout
+version=1.0
+author=
```
d-a-v: That would be @devyte "Develo <deveyes@gmail.com>"

PR #6374, opened 2019-08-01
fix _min and _max macros

d-a-v:


jandrassy:
here is some discussion above this kind of improvements of macros
https://groups.google.com/a/arduino.cc/forum/#!topic/developers/QRw4xJni75M

d-a-v:
I proposed there the same fixes with little echo.
This original definition hurts my eyes every time I see it.

PR #6378, opened 2019-08-03
Move all scripts and documentation to Python3

### LINKED ISSUE #6376: Migrate all scripts, symlinks, and installed tools to Python3
Python 2 has less than 5 months left to live (or at least, until all support ends)
https://pythonclock.org/

#6364 had an issue with a symlink to Python 2 being installed in the tools directory, which is one part but not all.

We need to:
. Update all script refs from `/usr/bin/python` to `/usr/bin/python3`
. Update the tools python-wrapper.tgz to point to python3
. Update the Windows standalone installation to python3
. Update platform.txt/etc. to ensure it's using the right python verison
. Validate things still work.

It's not as simple as changing scripts to `#/usr/bin/env python`, either, as that only works if there is a Python2 installation somewhere.  Newer OS releases may not install Python2 by default.

earlephilhower:
Python 2 EOL is Jan 1, 2020.  Migrate scripts to run under Python 3.

Under Windows, we're already running Python 3.7, by dumb luck.  The
oddness is that the Windows standalone executable for Python 3 is called
"python" whereas under UNIX-like OSes it's called "python3" with
"python" always referring to the Python 2 executable.  The ZIP needs to
be updated to include a Python3.exe (copy of Python.exe) so that we can
use the same command lines under Linux and Windows, and to preserve my
sanity.

Fixes #6376


earlephilhower:
https://www.ncsc.gov.uk/blog-post/time-to-shed-python-2

mikekgr:
Dear Sirs, Dear @earlephilhower ,
I did all the git master updates yesterday. In my case, that I use Windows and the Arduino 1.8.8 IDE, the simple blink.ino compilation gave me this error, ( although I have got and run the recent get.py ) :
exec: "E:\ARDUINO_ESP_DEV\arduino-IDE\hardware\esp8266com\esp8266/tools/python3/python3": file does not exist
Error compiling for board Generic ESP8266 Module.

After having a quick look, it seems that the directory that holds the python executables need to be renamed from "python" to "python3" in order to restore the normal operation. ( Alternative the calling function should try the path with "python" that is the standard directory name as it is now ).
I did this rename and I can verify the correct working compilation after that.

Thanks and Best Regards,
Mike Kranidis

PR #6398, opened 2019-08-08
Fix reverse dependency core Updater -> library ESP8266WiFi

dok-net:
Per review comments from @earlephilhower, see https://github.com/esp8266/Arduino/pull/6393#issuecomment-519386409 ff.


dok-net:
@earlephilhower The unnamed namespace is C++ for internal linkage, `static `in C. I'm paranoid and use both :-)

libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
```diff
@@ -20,6 +20,7 @@
   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include "BearSSLHelpers.h"
```
earlephilhower: Why the move of this include before the others?

libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
```diff
@@ -901,3 +905,17 @@ make_stack_thunk(br_ssl_engine_sendrec_buf);
...
+
+#if ARDUINO_SIGNING
+namespace {
+  static BearSSL::PublicKey signPubKey(signing_pubkey);
+  static BearSSL::HashSHA256 __hash;
+  static BearSSL::SigningVerifier __sign(&signPubKey);
+};
+
+namespace esp8266 {
+  UpdaterHashClass& hash = __hash;
```
earlephilhower: I'll leave the "access yyyy as unnamed namespace __xxxx" stuff up to @devyte to comment on (for me, I'm indifferent to the separation).

However, can we make these var names something more descriptive?  `updaterHash` instead of plain `hash`?  Since this is the only use of the vars, would make it cleaner when we go back and change things again in a couple years. :)

libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
dok-net: I use this idiom as good practice to ensure that headers are self contained - I don't want other people having to find out the hard way that some header dependencies must be manually resolved each time.

libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
dok-net: Sure, but in the interest of minimal changesets, I haven't touched the external names - except the namespace esp8266, a symbol called hash was just too much of an invitation to name clash disaster :-)

dok-net: Done:
hash -> updaterSigningHash
sign -> updaterSigningVerifier

PR #6408, opened 2019-08-10
Fix <sys/syslimits.h> for <climits>

nomis:
It's not possible to compile a file containing only `#include <climits>`
because `ARG_MAX` and `PATH_MAX` are defined by `<limits.h>` and then
`<sys/syslimits.h>` attempts to redefine them.

### DIFF
--- a/tools/sdk/libc/xtensa-lx106-elf/include/sys/syslimits.h
+++ b/tools/sdk/libc/xtensa-lx106-elf/include/sys/syslimits.h
@@ -37,7 +37,9 @@
 #ifndef _SYS_SYSLIMITS_H_
 #define _SYS_SYSLIMITS_H_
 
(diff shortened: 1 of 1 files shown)


nomis:
For some reason `espressif8266@2.2.2` is ok, but `espressif8266@2.0.4` isn't.

The release structure of `esp8266/Arduino` within `earlephilhower/esp-quick-toolchain` within `platformio/platform-espressif8266` is not helpful...

nomis:
Including only `<climits>` is fixed somehow in `espressif8266@2.2.2`, but if you then include `<sys/param.h>` it fails again:

```
#include <climits>
#include <sys/param.h>
```

I need this to be able to compile [protocolbuffers/protobuf](https://github.com/protocolbuffers/protobuf).

PR #6414, opened 2019-08-12
Prevent rewriting core_version.h if content unchanged

dok-net:
The PR compares old and new and suppresses file writing if nothing changes.
This stops builds from oscillating between 5s and 20s, when there is NO change to the core and "core libraries".


dok-net:
@d-a-v According to history, you're the commiter of previous changes to this file, could you please review?

PR #6419, opened 2019-08-13
Fix APConfig DHCP IP start in debug

mbm60:
Debug tells the " [APConfig] DHCP IP start: 192.168.1.100 " but the real start ip is 192.168.1.2


mbm60:
Closes because of #6422.

Issue #6031, opened 2019-04-30
Static DHCP Lease not working


mbm60:
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

### Platform

- Hardware:			ESP-12
- Core Version:			2.5.0
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			Dout
- Flash Size: 		        4MB
- lwip Variant:                    v2 Lower Memory
- Reset Method:		ck
- Flash Frequency:	        40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		        SERIAL
- Upload Speed:                115200

### Problem Description

unknown device must get ip 192.168.0.102 but never get that ip Sometimes get IP 192.168.0.100 and sometimes 192.168.0.2

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
#include <ESP8266WiFi.h>

const char *ssid = "ESPap";
const char *password = "thereisnospoon";

IPAddress apIP(192, 168, 0, 1);

void setup() {
  struct station_info *stat_info;
  String address;
  uint8_t i;

  uint8 mac_CAM[6] = { 0x00, 0x0C, 0x43, 0x01, 0x60, 0x15 };
  uint8 mac_PC[6] = { 0xb4, 0x52, 0x7e, 0x9a, 0x19, 0xa5 };

  Serial.begin(74880);
  Serial.println();
  Serial.println("Configuring access point...");

  WiFi.persistent(false);
  WiFi.mode(WIFI_AP);
  WiFi.softAPConfig(apIP, apIP, IPAddress(255, 255, 255, 0));

  wifi_softap_add_dhcps_lease(mac_CAM);  // always 192.168.0.100
  wifi_softap_add_dhcps_lease(mac_PC);   // always 192.168.0.101

  WiFi.softAP(ssid, password);
  
  Serial.print("AP IP address: ");
  Serial.println(WiFi.softAPIP());
  Serial.println("HTTP server started");
[...]

PR #6421, opened 2019-08-14
Added ability to set start and end ip for softAP

mbm60:
Base on #3562 and discussed in #6031


mbm60:
Closes for unwanted fork deletion.

PR #6422, opened 2019-08-14
Add ability to set start and end ip for softAP

mbm60:
Base on #3562 and discussed in #6031


jandrassy:
```    
dhcp_start = local_ip;
dhcp_start[3] += 99;
```
someone could use .200 as local IP

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -214,10 +221,24 @@ bool ESP8266WiFiAPClass::softAPConfig(IPAddress local_ip, IPAddress gateway, IPA
...
     info.gw.addr = gateway.v4();
     info.netmask.addr = subnet.v4();
-
+	
+    uint32_t softap_ip = htonl(info.ip.addr);
+    uint32_t start_ip = htonl(dhcp_start.v4());
+    uint32_t end_ip = htonl(dhcp_end.v4());
+    softap_ip >>= 8;
+    if (( start_ip >> 8 != softap_ip ) || ( end_ip >> 8 != softap_ip )) {
+        DEBUG_WIFI("[APConfig] dhcp_start or dhcp_end isn't in range of local_ip Address!\n");
```
d-a-v: This test is not right. v4 addresses need to be shifted by the number of zeroed bits in the netmask address. It is indeed 8 with 255.255.255.0 but since netmask is not hardcoded, "8" neither should be.
There are other attempts for a proper API for network calculations like #6196 .

mbm60: You are right. But this part of the code compares the start and the end IP with the local IP to be in the same range. And I copied it from the esp-dhcpserver.c file. for example :
If the local IP is 192.168.4.1, the start IP and end IP must be in range 192.168.4 .
The comparison with the netmask you are looking for is done elsewhere.

mbm60:
@jandrassy
> someone could use .200 as local IP

This part of the code applies the default value and if someone needs another start and end IP, must be set wiith --> softAPConfig(local_ip, gateway, subnet, dhcp_start, dhcp_end)

jandrassy:
> @jandrassy
> 
> > someone could use .200 as local IP
> 
> This part of the code applies the default value and if someone needs another start and end IP, must be set wiith --> softAPConfig(local_ip, gateway, subnet, dhcp_start, dhcp_end)

with .200 as IP of the AP, the default dhcp start IP would be .300

mbm60:
@jandrassy
> with .200 as IP of the AP, the default dhcp start IP would be .299

It was in the old code and I just used it again.

`    IPAddress ip = local_ip;
    ip[3] += 99;
    dhcp_lease.start_ip.addr = ip.v4();`

mbm60:
Maybe i should close this PR because of no feedback.

jandrassy:
and the default start IP and end IP calculation? if the end IP calculation overflows the range will have start IP larger then end IP. is it OK for DHCP server?

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
mbm60: I understand what you mean and you are absolutely right. But if the subnetmask is e.g 255.255.0.0 and the dhcp_start or dhcp_end is e.g 192.168.5.10, the "[APConfig] wifi_set_ip_info failed" error occurs in debug. Because of the following lines in the "esp-dhcpserver.c" file :

Capture

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -37,7 +37,12 @@ extern "C" {
 
 #include "debug.h"
 
-
+#ifndef htonl
+#define htonl(x) ( ((x)<<24 & 0xFF000000UL) | \
```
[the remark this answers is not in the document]
mbm60: Yes it is in the "lwip/def.h" file

libraries/ESP8266WiFi/examples/StaticLease/StaticLease.ino
```diff
@@ -75,26 +73,29 @@ void setup() {
...
   /* Configure AP with IP = 192.168.0.1 / Gateway = 192.168.0.1 / Subnet = 255.255.255.0
      if you specify the ESP8266's IP-address with 192.168.0.1, the function softAPConfig() sets the DHCP-range as 192.168.0.100 - 192.168.0.200
   */
   WiFi.softAPConfig(apIP, apIP, IPAddress(255, 255, 255, 0));
-  /* Setup your static leases.
 
-     As it depend from your first address, and need to be done BEFORE any request from client,
-     this need to be specified after WiFi.softAPConfig() and before WiFi.softAP().
+  /* Start Access Point. You can remove the password parameter if you want the AP to be open. */
+  WiFi.softAP(ssid, password);
```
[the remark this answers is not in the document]
mbm60: Yes but I did test and if it is before it won't work

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
d-a-v: Because you are right saying it is dependent on dhcp server,
I propose you test and exit with an error when `netmask != IPAddress(255,255,255,0)`.

mbm60: @d-a-v Thanks for your consideration.
Since the "DHCPS_MAX_LEASE" is limited to 100, we don't need more than 254 IPs in a range. And if we were to limit the "subnet mask" to 255.255.255.0, why would there be an option to set the "subnet mask"? And wouldn't it be better to implement "softAPConfig" without it as follows?
```cpp
bool ESP8266WiFiAPClass::softAPConfig(IPAddress local_ip, IPAddress gateway,  IPAddress dhcp_start, IPAddress dhcp_end) {

IPAddress subnet = (255, 255, 255, 0);
info.netmask.addr = subnet.v4();
```

Issue #2820, opened 2017-01-02
BUG: The SPI library does not handle HW SC for other bit lengths than 8.


tclarsson:
BUG: The SPI library does not handle HW SC for other bit lengths than 8.
For example: There is a bug in the routine "transfer16".

Hardware:			WEMOS ESP8266

If one uses transfer16 for 16 bit SPI transfer using HW CS, the SS/CS pin goes high after each 8 bit transfer. the 16bit word is therefore transferred as 2 x 8 bit bytes.

For the below to work:
  SPI.setHwCs(true);  //Use (automatic) HW SS/CS  (GPIO 15 , D8)
  SPI.setBitOrder(MSBFIRST);
  SPI.transfer16(0x1234);

I had to write my own alternative "_transfer16" routine:

uint16_t _bs(uint16_t data) {
  return(((data&0xFF)<<8)|((data>>8)&0xFF));
}

void _setDataBits(uint16_t bits) {
    const uint32_t mask = ~((SPIMMOSI << SPILMOSI) | (SPIMMISO << SPILMISO));
    bits--;
    SPI1U1 = ((SPI1U1 & mask) | ((bits << SPILMOSI) | (bits << SPILMISO)));
}

uint16_t _transfer16(uint16_t data) {
    if(!(SPI1C & (SPICWBO | SPICRBO))) data=_bs(data);
    while(SPI1CMD & SPIBUSY) {}
    _setDataBits(16);
    SPI1W0 = data;
    SPI1CMD |= SPIBUSY;
    while(SPI1CMD & SPIBUSY) {}
    data=SPI1W0;
    if(!(SPI1C & (SPICWBO | SPICRBO))) data=_bs(data);
    return (data);
}

devyte:
@tclarsson is this issue still valid in latest git?

UlliBien:
The bug is still there (Arduino/ESP8266, version 2.5.2). write16() does it correctly. I am not familiar with github and unfortunately can not make the change myself.

Change code to:
uint16_t SPIClass::transfer16(uint16_t data) {
  while (SPI1CMD & SPIBUSY) {}
  // Set to 16Bits transfer
  setDataBits(16);
  // Change byte order for MSBFIRST
  if (!(SPI1C & (SPICWBO | SPICRBO))){
    // MSBFIRST Byte first
    SPI1W0 = (data >> 8) | (data << 8);
  }
  else {
    // LSBFIRST Byte first
    SPI1W0 = data;
  }
[...]

Issue #6417, opened 2019-08-13
SPIClass::transfer(void *buf, uint16_t count) transfers with wrong block structure


UlliBien:
concerns spi.cpp, lines 309 to 329

The current algorithm of void SPIClass::transfer(void *buf, uint16_t count) is
-  repeat until buf is 32bits-aligned: transmit a single bytes using SPIClass::transfer(uint8_t data), increment buf and decrements count. This could occur up to 3 times.
- transmits multiple of 4 bytes via SPIClass::transferBytes(const uint8_t * out, uint8_t * in, uint32_t size). transferBytes() splits data into 64 byte blocks.
- transmit the remaining bytes (up to 3) as single byte transfers using SPIClass::transfer(uint8_t data)

If you use hardware CS there is the following problem. Mention, you *must* use hardware CS if you use the altenate pin configuration (hspi overlap):

After each transfer CS goes HIGH and LOW again. So, using transfer(void *buf, uint16_t count) you dont have a single block transfer. There are lot of tranfers with different sizes. In worst case: 3 single byte transfers, some 64 Byte transfers, 3 single Byte transfers.

You one can not avoid that this happens after every 64 bytes, because the IO buffer is only that long. But up to 64 bytes only a single block should be transferred.

The call sequence is:
```
transfer(void *buf, uint16_t count)      (establishes 32 bit aligment)
 |
 +--> transferBytes
      |
      +--> transferBytes_    (establishes 32 bit aligment too)
	        |
		    +--> transferBytesAligned_
```
You see, there is no need for method "transfer" to make a 32 bit alignment. This is done by "transferBytes_" (with underline at the end) anyway later. The algorithm of "transferBytes_" is much better. It transfers in 64 Byte blocks.

Please replace the code if "transfer" to a simple call to "transferBytes":
```
void SPIClass::transfer(void *buf, uint16_t count) {
  transferBytes(reinterpret_cast<uint8_t*>(buf), nullptr, count);
}
```

thangktran:
i'll take a look at this issue

earlephilhower:
@UlliBien, is the current behavior with multiple chip selects actually illegal per the spec, or is it something that's just not nice to see on a scope?  I'm asking because things like the SD library now use the SPI library for larger transfers, and it doesn't seem to be having any issue.

UlliBien:
[...]

PR #6423, opened 2019-08-15
libraries: fixed SPI

thangktran:
to handle other bit lengths rather than 8.
Thanks to @UlliBien for the fix suggestion.

Fixes: https://github.com/esp8266/Arduino/issues/2820

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -307,26 +307,32 @@ uint8_t SPIClass::transfer(uint8_t data) {
 }
 
 uint16_t SPIClass::transfer16(uint16_t data) {
-    union {
-            uint16_t val;
-            struct {
-                    uint8_t lsb;
-                    uint8_t msb;
-            };
-    } in, out;
-    in.val = data;
+    while (SPI1CMD & SPIBUSY) {}
+    // Set to 16Bits transfer
+    setDataBits(16);
 
-    if((SPI1C & (SPICWBO | SPICRBO))) {
-        //LSBFIRST
-        out.lsb = transfer(in.lsb);
-        out.msb = transfer(in.msb);
-    } else {
-        //MSBFIRST
-        out.msb = transfer(in.msb);
-        out.lsb = transfer(in.lsb);
-    }
-    return out.val;
+    const bool isBigEndian = !(SPI1C & (SPICWBO | SPICRBO));
+
+    // Change byte order for MSBFIRST
+    if (isBigEndian)
+        SPI1W0 = (data >> 8) | (data << 8);
+    else
+        SPI1W0 = data;
+
+    // Start the transmission
+    SPI1CMD |= SPIBUSY;
+    // Wait for transmission to complete
+    while (SPI1CMD & SPIBUSY) {}
+    data = SPI1W0;
+
+    // Change byte order for MSBFIRST
+    if (isBigEndian)
+        return (data >> 8) | (data << 8);
+
+    // If !isBigEndian, return the normal data.
+    return data;
 }
+
 
 void SPIClass::transfer(void *buf, uint16_t count) {
     uint8_t *cbuf = reinterpret_cast<uint8_t*>(buf);


thangktran:
> Thanks, @thangktran. I can't test this, but it seems fine logically.
> 
> The problem originated because we had a `write16()` call and a `transfer16()` call, both of which did the same thing (except for the value return).
> 
> I think it makes sense here to remove the `write16()` method guts and replace it with a call to `(void) transfer16()` (the `(void)` to be explicit that we're ignoring the returned value).
> 
> This way we only have one version of the actual work code, which seems like good code hygiene. I haven't gone thru the rest, but if there's a `writeByte()` and a `transferByte()`, the same comment applies...

Thank you for your input @earlephilhower . 
I just scanned through the code and find out that there are lots of code duplication between `transfer*()` method and `write*()` method.
I would open an Issue an prepare a PR to do these refactoring since they're not related to this bugfix branch/commit.
What is your opinion?

thangktran:
Is this PR read-to-be-merged? I need it for https://github.com/esp8266/Arduino/issues/6425

libraries/SPI/SPI.cpp
```diff
@@ -307,27 +307,32 @@ uint8_t SPIClass::transfer(uint8_t data) {
...
-        out.msb = transfer(in.msb);
-    } else {
-        //MSBFIRST
-        out.msb = transfer(in.msb);
-        out.lsb = transfer(in.lsb);
-    }
-    return out.val;
+    while (SPI1CMD & SPIBUSY) {}
+    // Set to 16Bits transfer
+    setDataBits(16);
```
devyte: Disclaimer: not an SPI expert or even a user. 

This change is forcing 16bit transfer. I think the "16" in transfer16() refers to the data type taken as arg, i. e. uint16_t, and not the transfer bit length. With this change it is no longer possible to send a 16 bit value as 2 8bit values with configured byte ordering, unless the user does 2 method calls. 
If my previous comments are correct, this is essentially changing the method semantics.
Do all SPI devices support 16bit transfers in this new way? 
Does it make sense to switch back and forth between 16 and 8 bit length modes for one particular device?
Does the current code not make sense? 
Is anyone out there using the current transfer16()?

libraries/SPI/SPI.cpp
```diff
@@ -307,27 +307,32 @@ uint8_t SPIClass::transfer(uint8_t data) {
...
+    // Start the transmission
+    SPI1CMD |= SPIBUSY;
+    // Wait for transmission to complete
+    while (SPI1CMD & SPIBUSY) {}
+    data = SPI1W0;
+
+    // Change byte order for MSBFIRST
+    if (isBigEndian)
+        return (data >> 8) | (data << 8);
+    else
```
devyte: Please remove else and leave the return statement at the end.

UlliBien:
@devyte:

> This change is forcing 16bit transfer. I think the "16" in transfer16() refers to the data type taken as arg, i. e. uint16_t, and not the transfer bit length. With this change it is no longer possible to send a 16 bit value as 2 8bit values with configured byte ordering, unless the user does 2 method calls.

That's what intended! transfer16() should transfer 16 bits (a word) with bit order for the complete word. If you want to transfer two bytes call transfer() twice in that byte order you want. 
There are devices that need 16 Bit transfers. RFM12 e.g. And if you want to transfer a 16 or 32 bit address you have to transfer the bits as a unit and not as a set of bytes.

Without the change you get:
CS singal controlled by the user:
```
  digitalWrite(CS, LOW); // CS goes low
  transfer16(x);         // 16 bit transfer without change of CS
  digitalWrite(CS, LOW); // CS goes high
```
Everything works as expeted.

Now let the hardware control the CS signal:
```
setHwCs(true);
...
transfer16(x); // CS goes low
               // 8 Bit transfer
               // CS goes high
               // CS goes low
               // 8 Bit transfer
               // CS goes high
```
Do you see the difference?

libraries/SPI/SPI.cpp
thangktran: @devyte we already have a discussion regarding this problem here: https://github.com/esp8266/Arduino/issues/6417#issuecomment-522224896
And as @UlliBien explained below.

devyte:
The change affects any app that currently uses transfer16(), and transfer16() has been around for a while. My main concern is that this PR introduces a change in behavior which could affect compatibility. 

Receiving an arg uint16_t is one thing, and transmitting a 16bit value is something completely unrelated. You can have a transfer16() that receives a uint16_t argument, and transmit it either as 2 individual bytes or as a 16bit value, depending on the setDataBits() setting, or on some member bool in the SPI class. 

Is the current code illegal SPI? Does it not work? Is there s device with which it does work? 

You said:
> There are devices that need 16 Bit transfers

I understand that. But what about the opposite? Do _all_ devices understand 16bit transfers? Is there a device that does not understand 16bit transfers, but to which you would like to send a 16bit value as 2 8bit transfers?
If the answer is yes, then the current code is also valid and shouldn't be dropped. As I said, my view is that receiving a uint16_t as function argument and sending a 16bit transfer are two different independent things. In that case, I would think either a new function is needed, or there should be a new setting (i. e. a bool) thst cbooses between the old and new code. 

To be clear, I'm not saying such a bool should be added and be done with it. I'm saying it is not clear to me that the current code should be dropped in favor of the proposed code, because so far I haven't seen a convincing argument that the current code is invalid. 

You said:
> The change does not effect old code that does not use the hardware control

So then that means that the behavior for hw vs. sw CS is not the same with this change. Enable hw CS, and transfer16 toggles CS between bytes. Does it make sense to have such different behavior? I would think that if I change the CS pin used in my circuit, and switch to sw CS control, the transmission bits within the SPI functions would remain the same! Then again, as I said, I'm not an SPI expert.

devyte:
I'm still not convinced that the current code is invalid. However, I also don't want to hold up your SPI work*.
How about this: I merge this PR as it is now, and if somebody complains about broken 16bit transfers I will @ you to step in and coordinate fixing whatever is needed? 

*The comments in the discussion that was pointed at are valid. If the hw supports more than one CS, an enhancement proposal would be welcome. Comments in the header would help understand the SPI class methods better and svoid having to look at the cpp. Precalculation. Etc.

earlephilhower:
I'm not able to find a "definitive" spec for it, but from what I've seem at TI and uChip, CS is asserted for the length of the transfer for all lengths of comms:
https://wiki.analog.com/resources/eval/sdp/sdp-b/peripherals/spi

CS deassertion looks like it is supposed to reset the slave's state machine, but all I've been able to find are non-definitive verilog (read: some guy on the internet's code) so that's not really canon.

thangktran:
> I'm still not convinced that the current code is invalid. However, I also don't want to hold up your SPI work*.
> How about this: I merge this PR as it is now, and if somebody complains about broken 16bit transfers I will @ you to step in and coordinate fixing whatever is needed?
> 
> *The comments in the discussion that was pointed at are valid. If the hw supports more than one CS, an enhancement proposal would be welcome. Comments in the header would help understand the SPI class methods better and avoid having to look at the cpp. Precalculation. Etc.

@devyte just to be clear. So we have `transfer16()` and `write16()` is on purpose due to the way they behave and not code duplication?

thangktran:
> Would it not make sense that this is a "transaction" setting? So SPISettings should be extended to include `dataWidth`?

@Makuna This is a good idea. Thank you.

@devyte what is your opinion about @Makuna's approach?

devyte:
Looking at the SPISettings class members, a _dataWidth addition looks right at home there. Whatever else, I think that should be done, preferably in its own PR. 

That still leaves this PR, where the question is which approach to use for write/transfer. I see the following options (please correct me if I'm wrong):
1. Overload methods. The arg type defines how many bytes are to be transferred, the method postfix defines the transfer width. Example : transfer16(uint32_t arg) would transfer 4 bytes as 2 successive 16bit transfers. 
2. A class flag or mode setter, which would change the transfer width for all calls after that. Example:  SPI.setTransferBits(blah).
3. An additional arg to the current methods, similar to msb. Example: SPI.transfer16(val, true, width16).

UlliBien:
Nobody needs such things! The current implementation is far away from the Arduino standards. Don't make it worse.
Please reead SPI specs and look how SPI devices work.

Makuna:
I strongly believe that all three options are incorrect.  I think this pull is going in the wrong direction to fix the issue.  Please correct me if I get this wrong (I may have missed something).  The API has two issues that are in conflict.  One is that Esp specific issue of the pointer being on 8bit/16bit/32bit boundary.  The other a SPI issue of sending "transactions" (SPI specific term here) of different bit width.  The SPI API is using SpiSettings is used to configure the "transaction"; and I believe this should be adhered to.   There should be no other way to set the bit width of a transaction.  This then leaves the Esp specific "16" methods to be left alone to be used to signify the source data size.

devyte:
@UlliBien Arduino is not a standard, it's a reference. Please be aware of the difference. 

@Makuna from looking at the Arduino reference api docs, I see the following:
1. transactions don't support 16bit width
2. transfer() means take an 8bit value and transfer in 8bit width
3. transfer16() means take a 16bit value and transfer in 16bit width

There is no support for other cases. To me, that looks like an incomplete api. It certainly looks like we already support a  broader range of cases.

devyte:
>  One is that Esp specific issue of the pointer being on 8bit/16bit/32bit boundary

That isn't quite right. The discussion is about the argument type passed to the transferXX() functions vs. the XX in tje function name. Currently there is this:
```cpp
transfer(uint8_t arg); //single byte arg, 8bit transfer width
transfer16(uint16_t arg); //16bit arg, 2x 8bit transfer width
```

The [Arduino reference for SPI](https://www.arduino.cc/en/Reference/SPITransfer) says nothing of 16bit transfer width, only of a 16bit val to send/receive. 

My overloaded method proposal is this:

```cpp
transfer(uint8_t arg); //single byte arg, 8bit transfer width 
transfer(uint16_t arg); //16bit arg, 2x 8bit transfer width 
transfer16(uint8_t arg); //single byte arg, 16bit transfer width
transfer16(uint16_t arg); //16bit arg, 16bit transfer width
```

Functionally, this contains our current functionality plus the proposal in this PR, and covers the Arduino reference ambiguity and all, i. e. it's a superset. 

> The other a SPI issue of sending "transactions"

As I said, the Arduino reference says nothing about 16bit width for transactions (unless I missed something). Still, I think the suggestion of adding the setting as a member in SPISettings is sound. Again, it would be a superset of the Arduino reference.

Makuna:
Thank you both for making the issues clear to me.  

I still feel this is the wrong direction, the nuance of passing a byte versus 16 bit could be too easy to miss the size difference and the nuance of what it means is not obvious when looking at the API.  This screams bad API design.  This is why I made the suggestion I did.
  
While often we find the Arduino "reference" vague and not well thought out,  I feel this would not make it better; or even this API better.  It feels like it would be harder to maintain and will require more future work.  Let alone more people asking questions on the gitter channel and forums ;-)

thangktran:
> Thank you both for making the issues clear to me.
> 
> I still feel this is the wrong direction, the nuance of passing a byte versus 16 bit could be too easy to miss the size difference and the nuance of what it means is not obvious when looking at the API. This screams bad API design. This is why I made the suggestion I did.
> 
> While often we find the Arduino "reference" vague and not well thought out, I feel this would not make it better; or even this API better. It feels like it would be harder to maintain and will require more future work. Let alone more people asking questions on the gitter channel and forums ;-)

I agree with you.
I think that if the user wants to send a 16bit data as 2 single-8-bit transfer, they should split the 16-bit data themself and call `transfer(uint8_t arg)` or `write(uint8_t arg)` twice. 
I would propose that we keep `transfer(uint8_t arg)` and `transfer16(uint16_t arg)` for backward-compatibility and i would change it that they call `write(uint8_t arg)` and `write16(uint16_t arg)` internally for correct behaviour. What do you think @devyte @UlliBien ?

Issue #6296, opened 2019-07-13
Flashing issues with esptool (2.5.1 and 2.5.2 relases with esptool)


mcer12:
Esptool erases memory and attempts to reset the device via RTS before upload which won't work if the device doesn't have a reset button or auto-reset circuitry. Devices like Sonoff plugs and others can't be flashed with this. The reset between memory wipe and actual upload should be removed.

earlephilhower:
@mcer12, I don't think any of the maintainers have such a device.  Would you be able to submit a PR with your proposed change?

f5soh:
Hi,
I encountered the same issue with a ESP-01 and basic usb serial + programming button since esptool is used.
The "Only Sketch" works fine but others options erasing or partially erasing flash fails while starting binary upload.

In platform.txt i added "--before no_reset_no_sync" before the write_flash command: 

```
tools.esptool.upload.pattern="{cmd}" "{runtime.platform.path}/tools/upload.py" --chip esp8266 --port "{serial.port}" --baud "{upload.speed}" "{upload.verbose}" {upload.erase_cmd} --end --chip esp8266 --port "{serial.port}" --baud "{upload.speed}" "{upload.verbose}" --before no_reset_no_sync write_flash 0x0 "{build.path}/{build.project_name}.bin" --end
```
and replaced 
`generic.menu.wipe.none.upload.erase_cmd=version`
with 
`generic.menu.wipe.none.upload.erase_cmd=flash_id`
into the board.txt so the firt command do something with serial without erase_flash or erase_region.

This workaround allow the "Flash erase: > Only Sketch / Sketch + Wifi settings / All flash contents" 
 3 options to work in my case, not sure what happens for boards with a functionnal RTS reset.

mcer12:
Glad somebody picket this up, thanks @f5soh . The erase should stay there otherwise you will basically just skip the erase step or do I get this wrong?

f5soh:
I'm not sure what do you mean, with the workaround above all the options works here:
1 - Only sketch upload (without erase or partial erase, as previously)
2 - Partial erase (wifi settings) + sketch upload
3 - Full erase + sketch upload

While adding "--before no_reset_no_sync" to the second part of command (used for upload) in platform.txt, the 2 and 3 case works but case 1 do not work any more.

To make the simple sketch upload (aka erase=none) to work again, the first command (used for erase options) is "flash_id" instead of "version" which just output esptool version without talking with serial.

**Only sketch** output:
[...]

PR #6429, opened 2019-08-19
Update esptool reset method

f5soh:
Related to #6296 

This is my first PR here, so be patient :)

Since 2.5.1 the upload tool moved to esptool.py, reset methods are deprecated and all boards use in fact the default_reset with dtr.

First commit update the reset methods available: 
- _none_ becomes as 'no dtr' in menu,
- _nodemcu, dtrset_  as 'dtr',
- _ck_ (fixed) temporary handled like 'none' (not sure about this), 
- new reset method 'no dtr, no_sync' in extra menu

Second commit is a workaround because the flash erase and write_flash is done in two commands and allow the two consecutive commands using 'no dtr, no_sync' without a reset circuit on board.
This workaround will not be needed if esptool.py handle the erase and write in one command.


earlephilhower:
For a first PR, you did pretty well!  Normally the boards.txt or an astyle problem hits users.

earlephilhower:
I think @d-a-v is the expert on this, but he's on vacation for another week, so do please be patient.

f5soh:
No issue, this PR still need some work and more information about how the ck method can be supported with esptool.py

Will be rebased when ready for merge

tools/boards.txt.py
```diff
@@ -914,20 +914,18 @@
     ####################### menu.resetmethod
 
     'resetmethod_menu': collections.OrderedDict([
-        ( '.menu.ResetMethod.ck', 'ck' ),
-        ( '.menu.ResetMethod.ck.upload.resetmethod', 'ck' ),
-        ( '.menu.ResetMethod.nodemcu', 'nodemcu' ),
-        ( '.menu.ResetMethod.nodemcu.upload.resetmethod', 'nodemcu' ),
+        ( '.menu.ResetMethod.dtr', 'dtr' ),
+        ( '.menu.ResetMethod.dtr.upload.resetmethod', '--before default_reset --after hard_reset' ),
+        ( '.menu.ResetMethod.nodtr', 'no dtr' ),
```
d-a-v: I would mention the old names with something like this:
* `dtr (aka nodemcu)`
* `no dtr (aka ck)`

so users are not lost

f5soh: Good idea, added changes.

earlephilhower:
@f5soh, you'll need to pull in master to fix the compile problems the CI system is having.  For some reason I can't do it for you from the website just now.

f5soh:
Seems the build.py need update for tests: https://github.com/esp8266/Arduino/blob/901410fa00b29c8f73aa9e05dca44832d464190f/tools/build.py#L55

`Error resolving FQBN: getting build properties for board esp8266com:esp8266:generic: invalid value 'nodemcu' for option 'ResetMethod'`

earlephilhower:
Ah, yeah.  You'll need to fix that in your PR, too.  There was also just 2 nights ago an update to the Arduino.cc executable nightly build which changes the FQBN format.  See #6461 .  Double whammy.

d-a-v:
@f5soh Would you mind renaming back the internal fqbn part to what it currently is (`nodemcu` and not `dtr`, `ck` and not `no-dtr`) - It is OK to leave the displayed name though.
Mainly because of annoyance following such change, like https://github.com/esp8266/Arduino/issues/5572#issuecomment-451074011 (and included links, also the number of references to it show how many people are affected by such change).

tools/boards.txt.py
```diff
@@ -1075,7 +1077,7 @@
 
     'flash_erase_menu': collections.OrderedDict([
         ( '.menu.wipe.none', 'Only Sketch' ),
-        ( '.menu.wipe.none.upload.erase_cmd', 'version' ),
+        ( '.menu.wipe.none.upload.erase_cmd', 'flash_id' ),
```
d-a-v: I am concerned about this change. 
`flash_id` asks for a connection to the esp, while `version` is a no-op which is what we need in this case.

f5soh: Take a board without reset circuit and version as first default command as you want :
- "Only sketch" works only with the "no dtr" reset method
- "Flash erase" works only with the "no dtr, no sync" reset method or second programming step will fail.

Do we really want the user change the reset method every time the erasing flash option is changed ?

With flash_id, both "Only sketch" and "Flash erase" works with "no dtr, no sync" reset method.

As described, this is a workaround needed since the esptool do the erasing and programming in two consecutive commands.  May need also the "no dtr, no sync" reset method for all "ck" boards.

sampetrosyan:
Does anyone know what's wrong here? after I upload the code successfully, below that it says

Leaving...

Hard resetting via RTS pin..

devyte:
What is pending here?

d-a-v:
I'd merge but release is too soon, feedback from users with issues is welcome.
I tested it with nodemcu reset method and nothing seems broken.
If anyone with an esp-01 progammer could also test it, that's be great.
@mcer12 @denisfou @IllyaMoskvin @ranger81

d-a-v:
@f5soh 
You'd need to update your local master, then on your branch locally remove boards.txt, merge master and regenerate boards.txt

Jason2866:
@d-a-v  There are more users using NodeMCU and wemos minis than users using ESP-01 devices with special programmer(s).  There is **no** standard ESP-01 programmer. I have two different types... I have no change in behaviour of this two. One works as before (reset). 
The other did not reset the device before this PR nor with this PR 

IMHO: Merging this PR will solve more problems, than it maybe! generates

d-a-v:
@Jason2866 This PR is not mergeable as-is and I don't know why.
@f5soh I made a [subPR](https://github.com/f5soh/Arduino/pull/1) to see how it goes.

Erik84750:
The issue I described in version 2.5.2 does not work when ESP8266 uses auto-reset #6631 is that when using the ESP8266 in conjunction with auto-reset circuitry (using DTR to set the ESP8266 in programming mode, and automatic restoring to normal mode after program download) there needs to be a hard (power-off power-on) reset to run the ESP8266 in normal mode.
Unless I am wrong, versions 2.5.1 and higher do not acknowledge the use of auto-reset circuitry, therefor requiring users to do a hard reset instead of automatically restarting in normal mode.

jandrassy:
> 
> 
> The issue I described in version 2.5.2 does not work when ESP8266 uses auto-reset #6631 is that when using the ESP8266 in conjunction with auto-reset circuitry (using DTR to set the ESP8266 in programming mode, and automatic restoring to normal mode after program download) there needs to be a hard (power-off power-on) reset to run the ESP8266 in normal mode.
> Unless I am wrong, versions 2.5.1 and higher do not acknowledge the use of auto-reset circuitry, therefor requiring users to do a hard reset instead of automatically restarting in normal mode.

and why don't you try the latest git which has the fix?

Erik84750:
Oh I am so sorry, of course I will try it, and asap! I am not sufficiently familiar yet with the workings of Github, my apologies for undue inconvenience.

PR #6440, opened 2019-08-23
Revert TCP connect and send delay fix (#6213)

ascillato:
The PR#6213 makes the pubsubclient library to hung (device unresponsive) for 5 seconds if the MQTT broker is down and it is trying to reconnect.

Reverting just this commit makes the STAGE core to be fast again as core 2.5.2 while using MQTT.


d-a-v:
This patch is useful for ethernet.
I must reproduce the issue and find the reason why pubsubclient is frozen for 5 seconds.

d-a-v:
Can you try with #6454 ?

Issue #6434, opened 2019-08-22
WiFiClientSecure::available() blocks on dropped connection when TX buffer full


johnm545:
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [ ] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:				ESP-12E
- Core Version:			2.5.2
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			NodeMCU
- Flash Mode:			[qio|dio|other]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   v2 Lower Memory
- Reset Method:		nodemcu
- Flash Frequency:	[40Mhz]
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

I've come across some unwanted behaviour in the WiFiClientSecure::available() and _run_until methods.  This becomes apparent when sending MQTT messages of a few hundred bytes at relatively slow intervals (minutes) without regular keep-alive pings, and the connection drops somewhere between the access point and the server (i.e. wifi stays up but internet goes down).  

The issue is as follows: when messages are sent but not ACK-ed, the TCP buffers begin to fill.  When there is no longer enough space for a complete message, the write attempt will block for 10 seconds, but still report success.  No problem yet, because it remains buffered.  

At this point, subsequent calls to WiFiClientSecure::available() will block for 5-10 seconds, because the available() method calls _run_until, which then attempts to re-write the bytes  remaining in the bearssl engine's buffer to the underlying WiFiClient.  This will occour endlessly until a further write attempt is made from the application via WiFiClientSecure::write; at this point the connection will be closed after write timeout.

Example and debug log below.  I am using the new ArduinoMqttClient, which I much prefer over the old PubSubClient, but this issue will occour with any WiFiClientSecure connection

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp


#include <ESP8266WiFi.h>
#include <ArduinoMqttClient.h>

const char* msg = R"(To be, or not to be, that is the question:
[...]

PR #6449, opened 2019-08-26
Fix WiFiClientSecure::available() blocking on dropped connections

johnm545:
When a secure socket connection is dropped somewhere between the access point and the server, write attempts eventually fill the WiFiClient's tx buffer.  At this point, calls to WiFiClientSecure::available() block for 5 (default) seconds.  See here for full description and example: https://github.com/esp8266/Arduino/issues/6434

This happens because WiFiClientSecure::available() calls _run_until(BR_SSL_RECVAPP, false), and this also triggers a blocking write attempt if there is still record data hanging around from previous write attempts (i.e. state & BR_SSL_SENDREC == true). The fix is to avoid writing to the WiFiClient more data than will fit in its tx buffer when the blocking flag == false, by first checking WiFiClient::availableForWrite().  This also works fine when availableForWrite() == 0 (WiFiClient::write just returns 0 instantly if the size argument is 0).

I have also removed the if (blocking) guard from the optimistic_yield call here.  A yield is required to avoid WDT reset if available() is called in a busy loop (as is common).  The WiFiClient::available() method includes a yield, so including one in the WiFiClientSecure method is consistent.


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -461,8 +458,19 @@ int WiFiClientSecure::_run_until(unsigned target, bool blocking) {
...
       buf = br_ssl_engine_sendrec_buf(_eng, &len);
+      availForWrite = WiFiClient::availableForWrite();
+      
+      if (!blocking && len > availForWrite) {
+        /* 
+           writes on WiFiClient will block if len > availableForWrite()
+           this is needed to prevent available() calls from blocking
+           on dropped connections 
+        */
+        len = availForWrite;
```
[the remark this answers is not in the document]
johnm545: In that case WiFiClient::write just returns 0 instantly:
```cpp
size_t WiFiClient::write(const uint8_t *buf, size_t size)
{
    if (!_client || !size)
    {
        return 0;
    }
    _client->setTimeout(_timeout);
    return _client->write(buf, size);
}
```

Previously if would return 0 after timeout.
_run_until then returns -1

There is (and was) a minor issue here: because _run_until first tries to send record data, and returns -1 if it can't be sent, we never get to test other states.  It's possible (I think) that there is actually app data to receive (i.e. state & BR_SSL_RECVAPP), or data in the WiFiClient input buffer to be processed, if it wasn't read out before write attempts fill the tx buffer on a dropped connection.

The return -1 here could probably be removed, and the "no_work = 0; continue;" statements moved up into the "if (wlen > 0) block", then the loop allowed to run through the other states before breaking at the bottom, but it would need a lot more testing.

As it stands the "if (wlen > 0)" on line 485 is actually redundant, because we only get there if wlen > 0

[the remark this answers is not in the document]
d-a-v: When `availForWrite`==0, -1 is returned from `WiFiClientSecure::_run_until`.
Should it retry until timeout ?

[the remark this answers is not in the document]
johnm545: IMO, no, not when blocking==false. 

Turns out the complex ordering of comparisons in _run_until is only needed in half-duplex mode.  In full duplex it can be simplified, as in this [example:](https://www.bearssl.org/gitweb/?p=BearSSL;a=blob;f=tools/sslio.c;h=e784d07877b98f395edb4348cc6ec7dec02f8928;hb=f3ac21cb00646600475b02ecd2c592ee36e417da#l104)

I've been experimenting with getting rid of _run_until entirely due to a few issues I've had with it (this one, https://github.com/esp8266/Arduino/issues/6464, and the sometimes-double or triple timeout).  The underlying IO can be broken out into _readRecordData and _writeRecordData functions, which to me seems simpler and more natural.  My experimental version is here for the curious:

https://github.com/johnm545/Arduino/blob/rewrite-brssl-io/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp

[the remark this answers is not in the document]
devyte: That last sounds like something for @earlephilhower

johnm545:
I've just pushed a fix for #6464 here too, as it's also a dropped connection issue and changes are in the same method

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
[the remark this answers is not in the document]
earlephilhower: Looks really cool, @johnm545!  Please do think about polishing up to another PR once you've got client and server tests running.  One less way of having a false/artificial deadlock would be great, and fixing a hard-to-repro bug.  Good stuff if it's kosher!

Client testing can be run on the host, so maybe it's worthwhile simulating the network failures/hiccups you're seeing on the real system there.  Sort of a fault injection framework...

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -437,12 +437,16 @@ int WiFiClientSecure::_run_until(unsigned target, bool blocking) {
     DEBUG_BSSL("_run_until: Not connected\n");
     return -1;
   }
-  for (int no_work = 0; blocking || no_work < 2;) {
-    if (blocking) {
-      // Only for blocking operations can we afford to yield()
-      optimistic_yield(100);
+  
+  unsigned long startMillis = millis();
```
earlephilhower: This timeout really should use a `polledTimeout`.  See the examples for info.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -437,12 +437,16 @@ int WiFiClientSecure::_run_until(unsigned target, bool blocking) {
...
+  for (int no_work = 0; blocking || no_work < 2;) {    
+    optimistic_yield(100);
+    
+    if (millis() - startMillis > 30000) {
```
earlephilhower: Here the timeout (or in the polledTimeout constructor) really should be `_timeout` and not a hard coded value.

I'm also not sure how you would get 30s (or `_timeout`) of delay in only a couple of loops of this.  Seems like if you're hard-hung due to a disconnect somewhere, the call that's taking those many seconds isn't returning control back to your main no_work loop, no?

johnm545: This timeout is a fix for #6464, which is a little different to the original issue #6434 - hopefully I didn't cause confusion by putting it in this PR.  

_run_until has a default argument of blocking=true in the header file.  When blocking==true the loop simplifies to a while(true): no_work is ignored and it can potentially run forever.  This happens if _wait_for_handshake is waiting for input - the underlying ClientContext::read is non-blocking and returns a size_t, which WiFiClient::read then casts to an int, so it will never return -1. When there's no input to read, WiFiClient::read returns 0 and no_work is incremented, but no_work is ignored when blocking==true.  And the sendapp state will not be reached until the handshake completes, so the loop can spin forever on a half-open socket.  

Connecting to servers that use complex keys can take 5-10 seconds, so it's quite possible for the connection to drop at this stage leaving a half-open socket.

The timeout here needs to be longer than _timeout, because there are multiple read/write operations involved in the SSL handshake, plus lots of number crunching which can take several seconds.  Hard-coding is IMO OK here because (a) it's addressing an edge case and (b) the function of the hard-coded number is clear from the code, but 2*_timeout or 3*_timeout would also work.  Currently _timeout is set to 15000 during the connection, then back to 5000 after the handshake, but interestingly the handshake still works perfectly fine with a much shorter _timeout, because that only applies to *individual* IO attempts.

Hope that makes it clearer - the logs in #6464 demonstrate the loop running for many cycles.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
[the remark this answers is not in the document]
johnm545: Thanks, I'm running this branch in my projects, but have yet to get the tests setup.  Testing on the host is a good idea - it should be possible to write some host tests that simulate connection drops etc.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
d-a-v: @johnm545 FYI [polledTimeout example](https://github.com/esp8266/Arduino/tree/master/libraries/esp8266/examples/BlinkPolledTimeout)

johnm545: IMO the conventional timeout is just as clear.  Is there another reason to prefer the polledTimeout? Looks like it will result in more variables on the stack, and ram is precious when establishing secure connections.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
johnm545: On second thought, and after some testing, you are right: using _timeout here makes more sense, provided the timeout is long enough.  The current 15 second value is enough.  I have run this against the test_BearSSL on-device test suite and it works fine.  

Interestingly the setting of _timeout to 15000 in the current (master branch) code seems to be unneeded, because it only applies to individual IO attempts on the underlying connection.  For interest's sake I ran the tests with _timeout set to 5000 instead of 15000 in WiFiClientSecureBearSSL, and everything still works, even when the handshake takes much longer than 5 seconds:

``` 
BearSSL - Maximum stack usage < 5600 bytes @ ecc384.badssl.org
test_BearSSL
Time Taken: 12.615107s
Stdout

16
>>>>>bs_test_start file="test_BearSSL.ino" line=158 name="BearSSL - Maximum stack usage < 5600 bytes @ ecc384.badssl.org" desc="[bearssl]"
ecc384.badssl.com: Trying: ecc384.badssl.com:443...Connected!
-------
HTTP/1.1 200 OK
-------
Stack: 5508



MAX THUNK STACK USAGE: 5508
>>>>>bs_test_end line=0 result=1 checks=1 failed_checks=0

Stderr
``` 

On a side note I'm getting occasional stack overflows on the eec256 test (on master branch and this one), it's uncomfortably close to the 5600 byte limit.

earlephilhower: We're actually allocating 5748 bytes for the stack.  A previous user found cases where ~5700 bytes were used.  Unfortunately, there's no static way of figuring out stack max usage so these numbers are experimentally determined.  Plus, I believe IRQs will use this stack when they happen during SSL processing, so the problem isn't even circumscribed in SSL code.

https://github.com/esp8266/Arduino/blob/8dd068eb40b4214f0ba81413d8888ae825e1d0f4/cores/esp8266/StackThunk.cpp#L39

We also now have a stack canary, so on stack overflow the return to main code will notice and abort (assuming the overflow doesn't cause BSSL code itself to crash).

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
earlephilhower: The BSSL stack is already paid for (allocated on first BSSL object creation), so an extra 4 bytes there is cost-free.

We're trying to reduce technical debt in the code, because many folks have trouble implementing timeouts.  So as we update sections where timeouts are implemented (not many in the core, but a few in the examples), we really try to keep a consistent usage.

johnm545: Makes sense - have pushed another commit with polledTimeout using _timeout in constructor.  This works fine against my MCVE and against the BearSSL device test suite.  I will have a crack at setting up a test when time permits.  I think the best way would be an on-device test, with a little python server running on the host that just drops with connection without closing the socket.

PR #6457, opened 2019-08-27
Bugfix/esp8266 http client

Jeroen88:
See PR #6176, this PR was merged while there was still an issue to solve.

With this PR:
- ::getString() no more returns an error but returns an empty String() if the server sends a '204 no content'
- if a connection is being reused and this connection gets lost (this can be detected because e.g. ::GET() returns a value less than zero), the connection can be restored by calling ::end() followed by ::begin(). An example of this usage is added as ReuseConnectionV2.ino

This behavior of this example is tested with both a WiFiClient and a (BearSSL) WiFiClientSecure. The lost connection was simulated with switching off the router and switching it on again, and with stopping my HTTP(S) server and restarting it. In the first case a connection refused was reported, in the second case a connection lost. Switching the router back on or restarting the server caused the program to resume showing the ::GET() payload as expected.


Jeroen88:
@d-a-v as promised a new PR. 

Could you take a look at ::disconnect() [here](https://github.com/esp8266/Arduino/blob/62e6282a846fd1f42c362fcf634f6cf6decb6750/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L451) and [here](https://github.com/esp8266/Arduino/blob/62e6282a846fd1f42c362fcf634f6cf6decb6750/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L459)?
I think the two reset()'s should only be done if !preserveClient and should be added to the second block of code as well. What do you think?

Jeroen88:
Last commit fixes style check Travis build error. Other Travis failed tests are not caused by my PR

Jeroen88:
@d-a-v Have you got any idea why the Travis build is failing?

Jeroen88:
@earlephilhower Thnx for your reaction Earle!

Jeroen88:
@earlephilhower I do not know where it comes from, but I will try to get it out. Could you have a look at this question:

> @d-a-v as promised a new PR.
> 
> Could you take a look at ::disconnect() [here](https://github.com/esp8266/Arduino/blob/62e6282a846fd1f42c362fcf634f6cf6decb6750/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L451) and [here](https://github.com/esp8266/Arduino/blob/62e6282a846fd1f42c362fcf634f6cf6decb6750/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L459)?
> I think the two reset()'s should only be done if !preserveClient and should be added to the second block of code as well. What do you think?

Jeroen88:
@earlephilhower please help with git: now I got rid of mklittlefs but got lwip2/builder instead...

earlephilhower:
You're in git hell, I think.  I'd roil back commits to before you did a push containing lwip and try again.

git reset --hard 85d836b4943a97212d8621b17c4f7065d4e182e8  # Go back to GH merge a while back
git rm <path to the mklittlefs.exe you've still got here in this hash>  # get rid of mkfs
git pull upstream master   # merge w/upstream
git submodule update  # will pull in the lwip change in the master branch which you missed and remove from the commit later
git commit -a
git push --force

Jeroen88:
Close because of git problems. Created a new PR #6476

Issue #6467, opened 2019-08-29
Windows build busted with missing \Python3\ directory


earlephilhower:
Opening for @Mikekgr.

The Python3 merge broke Windows due to a renaming of the installed python directory.

See for full details:  https://github.com/esp8266/Arduino/pull/6378#issuecomment-526284744

Either the python.zip file for Windows needs to be moved on the toolchain dir to \python3\ (simplest), or the platform.txt needs to not use {python} as the subdir.

earlephilhower:
@mikekgr, please give #6472 a try.  You will need to re-execute `get.py` and it will download a properly named python3-xxx.zip file and install into the appropriate spot.  No renaming should be needed.

mikekgr:
Dear @earlephilhower , I happily confirm that it is working fine now. Many many thanks !

earlephilhower:
Merged and confirmed.  Closing.

PR #6472, opened 2019-08-30
Fix the python3 directory so scripts work on Win32

earlephilhower:


earlephilhower:
Fixes #6467 

I'm trying to create clean VMs to ensure a fresh install works under both Windows and Linux.  From what I can see, I'm not sure how the Linux side was working since it was installing the python3 symlink under `tools/python/python3` and platform.txt called for `tools/python3/python3`.  Very strange...

earlephilhower:
Verified a build on install into a clean Win32 and Ubuntu 18.04 VM.  Good to go.

dok-net:
@earlephilhower Please update the gitignore entry, too.

PR #6476, opened 2019-09-01
bugfix2/ESP8266HTTPClient

Jeroen88:
Because of git problems, start from a new fork and create a new PR. This was PR #6457


Jeroen88:
See PR #6176 and PR6457, this PR was merged while there was still an issue to solve.

With this PR:

    ::getString() no more returns an error but returns an empty String() if the server sends a '204 no content'
    if a connection is being reused and this connection gets lost (this can be detected because e.g. ::GET() returns a value less than zero), the connection can be restored by calling ::end() followed by ::begin(). An example of this usage is added as ReuseConnectionV2.ino

This behavior of this example is tested with both a WiFiClient and a (BearSSL) WiFiClientSecure. The lost connection was simulated with switching off the router and switching it on again, and with stopping my HTTP(S) server and restarting it. In the first case a connection refused was reported, in the second case a connection lost. Switching the router back on or restarting the server caused the program to resume showing the ::GET() payload as expected.

Jeroen88:
@d-a-v and @earlephilhower 

Could you take a look at ::disconnect() [here](https://github.com/esp8266/Arduino/blob/bf1f1076a8e4b47bb5eb0d712e300b06c1a3c3e2/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L452) and [here](https://github.com/esp8266/Arduino/blob/bf1f1076a8e4b47bb5eb0d712e300b06c1a3c3e2/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L459)?
I think the two reset()'s should only be done if !preserveClient and should be added to the second block of code as well. What do you think?

d-a-v:
@Jeroen88 
Can you replace the former example by the new one ?

Jeroen88:
@d-a-v I think I do not understand your request? Which example do you want to be replaced by what other example?

d-a-v:
reuseConnectionV2 should replace reuseConnection.ino no ?

Jeroen88:
@d-a-v I just checked the differences (because I forgot :)). They are minimal: in v2 ::begin() is called in setup() in stead of in the loop(), which is logical, because the connection is reused. Also, ::end(); ::begin() is called in the loop if an error occurred. Assumed is that the connection is lost and should be reinitialized.

So in short: the examples are different but v2 could replace the original example. However: I lost this branch and I am no git hero. Could you do it?

d-a-v:
@Jeroen88 I sent an email to your git address yesterday

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1301,6 +1309,7 @@ int HTTPClient::handleHeaderResponse()
                     _canReuse = (headerLine[sizeof "HTTP/1." - 1] != '0');
                 }
                 _returnCode = headerLine.substring(9, headerLine.indexOf(' ', 9)).toInt();
+                _canReuse = (_returnCode != '0');
```
earlephilhower: Few problems here.

_canReuse sholuld only be cleared, not set by this (i.e. if its already false due to HTTP 1.0).

_returnCode is already an integer, not a char.

Also, *maybe* _returnCode !=0 and _returnCode < 500 (i.e. does a server error mean no connection reuse?  I don't know the answer yet.)

I think that results in:  `_canReuse = _canReuse && (_returnCode != 0)`

earlephilhower:
Ping @Jeroen88.  Would you be able to see about the merge conflict and the review?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
d-a-v: applied in conflict resolution

d-a-v:
@Jeroen88 can you give a look to your updated PR ?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1334,22 +1342,20 @@ int HTTPClient::handleHeaderResponse()
...
             if (headerLine.startsWith(F("HTTP/1."))) {
-                if (_canReuse) {
-                    _canReuse = (headerLine[sizeof "HTTP/1." - 1] != '0');
-                }
+
+                _canReuse = _canReuse && (headerLine[sizeof "HTTP/1." - 1] != '0');
                 _returnCode = headerLine.substring(9, headerLine.indexOf(' ', 9)).toInt();
-                continue;
-            }
+                _canReuse = _canReuse && (_returnCode != '0');
```
earlephilhower, line 1356: `_returnCode` is already parsed to an integer in line 1355, so don't use the character comparison

If you wanted to disallow connection reuse on a server error (playing it safe) I would suggest:
`_canReuse = _canReuse && (_returnCode > 0) && (_returnCode <500);`

Otherwise, the original review comment would be requested:
`_canReuse = _canReuse && (_returnCode != 0);`

libraries/ESP8266HTTPClient/examples/ReuseConnectionV2/ReuseConnectionV2.ino
```diff
@@ -0,0 +1,80 @@
...
+  // allow reuse (if server supports it)
+  http.setReuse(true);
+
+
+  http.begin(client, "http://jigsaw.w3.org/HTTP/connection.html");
+  //http.begin(client, "jigsaw.w3.org", 80, "/HTTP/connection.html");
+}
+
+void loop() {
+  for (int i = 0; i < 10; i++) {
```
devyte, line 52: This is a potentially long for() which could block the loop for a long time. In this very specific sketch that may work, but there is a long list of cases where it's not ok to do that, and examples should show good practices. It's a bit like a game: you don't want to do too much in one frame because the fps goes down and the game freezes for the duration.
Instead, I suggest moving the counter i to be global, then on every pass through the loop() increment, then once done  with the 10 tests just skip everything. No forever loop either.

libraries/ESP8266HTTPClient/examples/ReuseConnectionV2/ReuseConnectionV2.ino
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    Serial.println("\n\n\nWait 5 second...\n");
+    delay(5000);
+  }
+
+  http.end();
+
+  Serial.println("Done testing, now wait forever");
+  for (;;) {
```
devyte, line 77: Forever loops shouldn't be put in examples, they tend to spawn bad code that leads to crashes and issues.

earlephilhower:
I think the HTTP/1.0 stuck open errors after a 500 (and probably in other cases) are basic issues with the way HTTPClient handles the TCP connection.

HTTPClient never actually closes the TCP connection on its own.  It will leave the TCP connection open unless you explicitly do a `getString` which makes a `StreamString` and stuffs it with the HTTP server response, at which point the HTTP server itself will close the connection.  

If you check the HTTP error code and find failure, unless you do a `getString` and throw it away, it won't disconnect.

A possible solution would be to do an explicit disconnect when `!_canReuse` at the beginning of `HTTPClient::connect()` before it does its read and dump of old HTTP server data at line 1209 :
 
https://github.com/esp8266/Arduino/blob/c4e0cbec6efdf63463f976e12b89689c428228f5/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L1207-L1220

libraries/ESP8266HTTPClient/examples/ReuseConnectionV2/ReuseConnectionV2.ino
earlephilhower: Done.

PR #6484, opened 2019-09-04
Ets intr lock nest

ChocolateFrogsNuts:
Replace the SDK's use of ets_intr_lock / ets_intr_unlock with nestable versions.

This is intended for use in addition to using xt_rsil / xt_wsr_ps in the Arduino code and other more correct approaches where they are possible.


d-a-v:
Are the modifications you brought to fw libs scriptable ?
Can you provide this script in the PR for when we update the sdk ?
(like `fix_sdk_libs.sh`)

tools/sdk/lib/fix_sdk_libs.sh
```diff
@@ -16,3 +16,10 @@ xtensa-lx106-elf-objcopy --redefine-sym default_hostname=wifi_station_default_ho
 xtensa-lx106-elf-objcopy --redefine-sym default_hostname=wifi_station_default_hostname eagle_lwip_if.o 
 xtensa-lx106-elf-ar r libmain.a eagle_lwip_if.o user_interface.o 
 rm -f eagle_lwip_if.o user_interface.o
+
+# Replace use of ROM ets_intr_(un)lock with nestable ets_intr_(un)lock_nest
+for f in libmain.a libpp.a libnet80211.a; do
```
earlephilhower: If we're changing every instance of this, the simplest and most direct way, w/o faffing about with the lib.a files (and making sure that it stays stuck when we upgrade SDK revs), this isn't necessary.

Go back to the original versions of these libs and remove the `PROVIDES` line in the `/tools/sdk/ld/*rom.ld` file for `ets_intr_lock` and `ets_intr_unlock` and then it will link against your version directly.

ChocolateFrogsNuts: As far as I can see it links against functions in libs in preference to rom functions, so all it needs is naming the funcs the same as the original rom ones. Builds without problems and doesn't need modifications to the libs.
I can remove the PROVIDES as well if required, but it works without it.

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -60,6 +60,13 @@ static os_event_t s_loop_queue[LOOP_QUEUE_SIZE];
 /* Used to implement optimistic_yield */
 static uint32_t s_micros_at_task_start;
 
+/* For ets_intr_lock_nest / ets_intr_unlock_nest
+ * Max nesting seen by SDK so far is 2.
+ */
+#define ETS_INTR_LOCK_NEST_MAX 7
+uint16_t ets_intr_lock_stack[ETS_INTR_LOCK_NEST_MAX];
```
earlephilhower: Probably want these new vars as `static` here, just to match the rest of the locals.

ChocolateFrogsNuts:
The ets_post wrapper appears to have resolved the arp / network deafness issue - at least my test unit is 58 hours up and reachable where I was seeing it go deaf after around 25-30 hours operation.
Testing continues, but it looks like it's worth getting a few more people verifying the result.

cores/esp8266/core_esp8266_main.cpp
ChocolateFrogsNuts: yep, good point. I'll do it once we get some positive reports back.
I've been accessing them from some of my other test code, but once that's no longer needed they can go static.

d-a-v:
If that's true, this is a real major improvement !

What about
- some comments like "// save PS state before calling rom ets_post, workaround for silicium bug",
- restoring and commenting removed `provides` and telling the reason

This should be merged asap !

d-a-v:
IRAM_ATTR is needed for this PR to work on my side:
(otherwise I get wdt or bootloop)

```
diff --git a/cores/esp8266/core_esp8266_main.cpp b/cores/esp8266/core_esp8266_main.cpp
index ad3c3522..0ddff2fb 100644
--- a/cores/esp8266/core_esp8266_main.cpp
+++ b/cores/esp8266/core_esp8266_main.cpp
@@ -128,14 +128,18 @@ extern "C" void optimistic_yield(uint32_t interval_us) {
     }
 }
 
-extern "C" void ets_intr_lock() {
+extern "C"
+IRAM_ATTR
+void ets_intr_lock() {
   if (ets_intr_lock_stack_ptr < ETS_INTR_LOCK_NEST_MAX)
      ets_intr_lock_stack[ets_intr_lock_stack_ptr++] = xt_rsil(3);
   else
      xt_rsil(3);
 }
 
-extern "C" void ets_intr_unlock() {
+extern "C"
+IRAM_ATTR
+void ets_intr_unlock() {
   if (ets_intr_lock_stack_ptr > 0)
      xt_wsr_ps(ets_intr_lock_stack[--ets_intr_lock_stack_ptr]);
   else
@@ -145,7 +149,9 @@ extern "C" void ets_intr_unlock() {
 
 extern "C" bool ets_post_rom(uint8 prio, ETSSignal sig, ETSParam par);
 
-extern "C" bool ets_post(uint8 prio, ETSSignal sig, ETSParam par) {
+extern "C"
+IRAM_ATTR
+bool ets_post(uint8 prio, ETSSignal sig, ETSParam par) {
   uint32_t saved;
   asm volatile ("rsr %0,ps":"=a" (saved));
   bool rc=ets_post_rom(prio, sig, par);
diff --git a/tools/sdk/ld/eagle.rom.addr.v6.ld b/tools/sdk/ld/eagle.rom.addr.v6.ld
index 215c3d07..97d6081d 100644
--- a/tools/sdk/ld/eagle.rom.addr.v6.ld
+++ b/tools/sdk/ld/eagle.rom.addr.v6.ld
@@ -119,6 +119,10 @@ PROVIDE ( ets_install_external_printf = 0x40002450 );
 PROVIDE ( ets_install_putc1 = 0x4000242c );
 PROVIDE ( ets_install_putc2 = 0x4000248c );
 PROVIDE ( ets_install_uart_printf = 0x40002438 );
+/* permanently hide reimplemented ets_intr_*lock(), see #6484
+PROVIDE ( ets_intr_lock = 0x40000f74 );
+PROVIDE ( ets_intr_unlock = 0x40000f80 );*/
+*/
 PROVIDE ( ets_isr_attach = 0x40000f88 );
 PROVIDE ( ets_isr_mask = 0x40000f98 );
 PROVIDE ( ets_isr_unmask = 0x40000fa8 );
@@ -126,6 +130,9 @@ PROVIDE ( ets_memcmp = 0x400018d4 );
 PROVIDE ( ets_memcpy = 0x400018b4 );
 PROVIDE ( ets_memmove = 0x400018c4 );
 PROVIDE ( ets_memset = 0x400018a4 );
+/* renamed to ets_post_rom(), see #6484
+PROVIDE ( ets_post = 0x40000e24 );
+*/
 PROVIDE ( ets_post_rom = 0x40000e24 );
 PROVIDE ( ets_printf = 0x400024cc );
 PROVIDE ( ets_putc = 0x40002be8 );
```

TD-er:
With the IRAM_ATTR my node does boot and seems to run fine.
Without it, the node entered a reboot loop.

TD-er:
Just a short update here.
Running for 4 hours, with sending ARP disabled and also the communication from the ESP node is set to a minimum.
It is still responding quickly and does not seem to be missing UDP packets from other nodes.
Will now increase the likelihood for missing packets by enabling "eco mode" (calling delay() when nothing has to be done)

d-a-v:
With `WIFI_LIGHT_SLEEP` ?

TD-er:
@d-a-v If that's the default behavior then yes.
I do not set any WiFi sleep mode.

The only WiFi sleep related code I use is this:
```c++
```

And that one is now **not** being called, so it is just running the default sleep mode.

TD-er:
Just some quick update.
Running on 2 nodes now, both with my so called "eco" mode enabled.
Running just fine and also the timing stats do seem to show less extremes which were mostly caused by timeouts connecting to a host.
I will leave them running now without querying them for a number of hours to see what happens.
But for now I'm quite positive about this fix.

cziter15:
Me too, running two nodes with modem sleep option DTIM=3.

I've had problems with connection stability on LWIP2 before this fix, while LWIP 1.4 was working longer w/o disconnection from MQTT and without router disassociation due to inactivity. Testing LWIP2 Higher Bandwidth right now.

TD-er:
My nodes were still reacting fast after the night, so at least it is not breaking anything :)
The nodes with this PR are not running Gratuitous ARP, so that's a good thing.

ChocolateFrogsNuts:
Awesome!  My original test device is at 95 hours now and still going strong.
I'll change the new vars to static and add some comments to ets_post as suggested above.
I'll push that in the morning - by then the test will be past the 100 hour mark.

cziter15:
25 hours passed, no disconnection from MQTT on LWIP2. Never reached this before :)
It means that it improves #2330 , don't saying 'fixes' yet.

d-a-v:
> don't saying 'fixes' yet

Let's dare

@ChocolateFrogsNuts This is thrilling

ChocolateFrogsNuts:
> > don't saying 'fixes' yet
> 
> Let's dare
> 
> @ChocolateFrogsNuts This is thrilling

I'll say - a couple of weeks ago I was beginning to think I should abandon the ESP platform because it wasn't reliable - I've only been using it a couple of months.

With this fix it's looking rock solid, at least for what I want to do.

Just added some comments and made the couple of vars static.
I'm pretty happy with it now.

d-a-v:
Then please remove this "WIP" :)

:rocket: ***rock solid*** :rocket:

(I had the same reasoning with lwIP)

devyte:
Fixes #2330

devyte:
Ref: tools uploaded to this repo:
https://github.com/ChocolateFrogsNuts/ESP-DebugTools

d-a-v:
I'd like to mention that following this fix, a board that was never working suddenly resurrected.
Namely, it is a sonoff basic. I had ordered two of them, one was working and not the other. I thought of an hardware malfunction. But it was not. This fix is golden. Thanks again!

(symptom: Losing wifi after the first connected second. Never able to reconnect until a reset)

cziter15:
Yeah ;) My boards connection time is now more than one week. No disconnections anymore! On LWIP2, before this fix, it was dropping after few hours.

PR #6485, opened 2019-09-04
Wakeup delayed scheduling

dok-net:
I have a use case for the ESP8266 core schedule class that needs the capability to let recurrent, delayed functions run immediately on the next iteration of the scheduler without taking into account a remaining delay or repeat timeout.
This PR [originally added] an optional parameter `const std::atomic<bool>* wakeupToken` to `schedule_recurrent_function_us()`. This was changed, after review comments, into a `std::function<bool(void)>` alarm polling callback.
The remaining delay or the logic by which the scheduled function return value determines whether to keep scheduling for execution or to be removed from scheduling is unaltered by the alarm.

[CoopTask release 2.0.0](https://github.com/dok-net/CoopTask/releases/tag/2.0.0) depends on this PR for it's new Semaphore with milliseconds timeout.

### DIFF
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -29,7 +29,8 @@ struct recurrent_fn_t
     recurrent_fn_t(esp8266::polledTimeout::periodicFastUs interval) : callNow(interval) { }
 };
 
-static recurrent_fn_t rFirst(0); // fifo not needed
+static recurrent_fn_t rFirst(0);
+static recurrent_fn_t* rLast = &rFirst;
 
 // Returns a pointer to an unused sched_fn_t,
 // or if none are available allocates a new one,
@@ -98,13 +99,14 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn, uint32_
         return false;
 
     item->mFunc = fn;
+    item->mNext = nullptr;
     item->wakeupToken = wakeupToken;
     item->wakeupTokenCmp = wakeupToken && wakeupToken->load();
 
     esp8266::InterruptLock lockAllInterruptsInThisScope;
 
-    item->mNext = rFirst.mNext;
-    rFirst.mNext = item;
+    rLast->mNext = item;
+    rLast = item;
 
     return true;
 }
@@ -124,9 +126,9 @@ void run_scheduled_functions()
             sFirst = sFirst->mNext;
             if (!sFirst)
                 sLast = nullptr;
-			recycle_fn_unsafe(to_recycle);
-		}
-    	
+            recycle_fn_unsafe(to_recycle);
+        }
+
         if (yieldNow)
         {
             // because scheduled function are allowed to last:
@@ -147,7 +149,7 @@ void run_scheduled_recurrent_functions()
     // its purpose is that it is never called from an interrupt
     // (always on cont stack).
 
-    recurrent_fn_t* current = rFirst.mNext;
+    auto current = rFirst.mNext;
     if (!current)
         return;
 
@@ -164,10 +166,14 @@ void run_scheduled_recurrent_functions()
         fence = true;
     }
 
-    recurrent_fn_t* prev = &rFirst;
+    auto prev = &rFirst;
+    // prevent scheduling of new functions during this run
+    auto stop = rLast;
 
+    bool done;
     do
     {
+        done = current == stop;
         const bool wakeupToken = current->wakeupToken && current->wakeupToken->load();
         const bool wakeup = current->wakeupTokenCmp != wakeupToken;
         if (wakeup) current->wakeupTokenCmp = wakeupToken;
@@ -180,11 +186,10 @@ void run_scheduled_recurrent_functions()
 
             auto to_ditch = current;
 
-            // current function recursively scheduled something
-            if (prev->mNext != current) prev = prev->mNext;
-
             current = current->mNext;
             prev->mNext = current;
+            // removing rLast
+            if (!current) rLast = prev;
 
             delete(to_ditch);
         }
@@ -201,7 +206,7 @@ void run_scheduled_recurrent_functions()
             esp_schedule();
             cont_yield(g_pcont);
         }
-    } while (current);
+    } while (current && !done);
 
     fence = false;
 }
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -24,7 +24,7 @@ struct recurrent_fn_t
     recurrent_fn_t* mNext = nullptr;
     mRecFuncT mFunc;
     esp8266::polledTimeout::periodicFastUs callNow;
-    std::function<bool()> alarm = nullptr;
+    std::function<bool(void)> alarm = nullptr;
     recurrent_fn_t(esp8266::polledTimeout::periodicFastUs interval) : callNow(interval) { }
 };
 
@@ -86,7 +86,7 @@ bool schedule_function (const std::function<void(void)>& fn)
 }
 
 bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
-    uint32_t repeat_us, std::function<bool()> alarm)
+    uint32_t repeat_us, std::function<bool(void)> alarm)
 {
     assert(repeat_us < decltype(recurrent_fn_t::callNow)::neverExpires); //~26800000us (26.8s)
 
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -27,7 +27,7 @@ struct recurrent_fn_t
     recurrent_fn_t (esp8266::polledTimeout::periodicFastUs interval): callNow(interval) { }
 };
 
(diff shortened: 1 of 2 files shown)


dok-net:
Caveat: this PR implies and contains #6307 - both are unconditional requirements for using the core schedule class from CoopTasks.

dok-net:
This PR completely supersedes #6307 now - dropping #6307.

cores/esp8266/Schedule.cpp
```diff
@@ -80,22 +86,27 @@ bool schedule_function (const std::function<void(void)>& fn)
...
+    if (!fn)
+        return false;
 
     recurrent_fn_t* item = new (std::nothrow) recurrent_fn_t(repeat_us);
     if (!item)
         return false;
 
     item->mFunc = fn;
+    item->mNext = nullptr;
+    item->wakeupToken = wakeupToken;
```
devyte: What happens in this assignment? I'm not quite sure, but cppreference says at the top that std::atomic is not copyable.

cores/esp8266/Schedule.cpp
```diff
@@ -152,28 +166,31 @@ void run_scheduled_recurrent_functions ()
...
-                current = current->mNext;
-                prev->mNext = current;
-            }
-            else
-            {
-                rFirst = rFirst->mNext;
-                current = rFirst;
-            }
+            // removing rLast
+            if (rLast == current) rLast = prev;
```
devyte: No single-line if's please. Put the if body on its own line.

cores/esp8266/Schedule.cpp
```diff
@@ -152,28 +166,31 @@ void run_scheduled_recurrent_functions ()
...
 
-    while (current)
+    bool done;
+    do
     {
-        if (current->callNow && !current->mFunc())
+        done = current == stop;
+        const bool wakeupToken = current->wakeupToken && current->wakeupToken->load();
+        const bool wakeup = current->wakeupTokenCmp != wakeupToken;
+        if (wakeup) current->wakeupTokenCmp = wakeupToken;
```
devyte: No single-line ifs please. Put the if body on its own line.

cores/esp8266/Schedule.cpp
```diff
@@ -152,28 +166,31 @@ void run_scheduled_recurrent_functions ()
...
+    // prevent scheduling of new functions during this run
+    auto stop = rLast;
 
-    while (current)
+    bool done;
+    do
     {
-        if (current->callNow && !current->mFunc())
+        done = current == stop;
+        const bool wakeupToken = current->wakeupToken && current->wakeupToken->load();
```
devyte: I don't understand how this logic (i.e.: surrounding code, not just this single line) relates to the explanation in the .h. Please add an explanation describing how the wakeup/wakeupToken/etc pieces interact.

cores/esp8266/Schedule.cpp
```diff
@@ -24,10 +24,13 @@ struct recurrent_fn_t
     recurrent_fn_t* mNext = nullptr;
     mRecFuncT mFunc;
     esp8266::polledTimeout::periodicFastUs callNow;
-    recurrent_fn_t (esp8266::polledTimeout::periodicFastUs interval): callNow(interval) { }
+    const std::atomic<bool>* wakeupToken = nullptr;
```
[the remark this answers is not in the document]
earlephilhower: Not sure what the point is, either, as there is no multiprocessing and 32b writes are by definition atomic here.  It adds a `memw` ASM instruction around uses, but other than that it seems to be a no-op on the 8266.

cores/esp8266/Schedule.cpp
earlephilhower: Here he;s just copying a pointer to the std::atomic, so that should be legal.

cores/esp8266/Schedule.cpp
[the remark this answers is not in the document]
dok-net: @devyte @earlephilhower I hope I can settle this with a rather synthetic-looking example, but essentially, it's all about sequential consistency and whether data gets written to memory, and is therefore accessible to ISRs, or whether it's not.
Synthetic it is in this sense, that there are no function calls, that can force the compiler to emit register computations to RAM, but again, that's the point, in an ISR I may very well need the values of objects any time, not depending on something outside of my code-writing control.

Please, let's first establish as a fact that the use of `volatile` is not a option (undefined, poorly defined, inappropriate, whatever ;-) )

Take this code:

```
#include <atomic>

#define ATOMIC

#ifdef ATOMIC
std::atomic<int> a(0);
std::atomic<int> b(0);
std::atomic<int> c(0);
#else
int a(0);
int b(0);
int c(0);
#endif

void IRAM_ATTR isr(void * arg)
{
#ifdef ATOMIC
	Serial.printf("ISR - a = %i, b = %i, c = %i\n", a.load(), b.load(), c.load());
#else
	Serial.printf("ISR - a = %i, b = %i, c = %i\n", a, b, c);
#endif
}

// The setup() function runs once each time the micro-controller starts
void setup()
{
#ifdef ESP8266
	Serial.begin(74880);
#else
	Serial.begin(115200);
#endif
	Serial.println("Sequential consistency test");
	pinMode(D3, INPUT_PULLUP);
	attachInterruptArg(D3, isr, nullptr, FALLING);
}

// Add the main program code into the continuous loop() function
void loop()
{
#ifdef ATOMIC
	while (a.load() < 1000000)
	{
		auto d = b.load();
		a.store(a.load() + 1);
		b.store(b.load() + 1);
		a.store(a.load() + 1);
		c.store(c.load() + a.load() - d);
		b.store(b.load() + 1);
	}
	a.store(a.load() - b.load());
	b.store(b.load() - c.load());
	c.store(0);
#else
	while (a < 1000000)
	{
		auto d = b;
		a = a + 1;
		b = b + 1;
		a = a + 1;
		c = c + a - d;
		b = b + 1;
	}
	a = a - b;
	b = b - c;
	c = 0;
#endif
}

```

With `#define ATOMIC`, hitting the button prints nice expected lines like these:
```
ISR - a = 493858, b = 493857, c = 493858
ISR - a = 493858, b = 493857, c = 493858
ISR - a = 861129, b = 861128, c = 861128
ISR - a = 861129, b = 861128, c = 861128
ISR - a = 17491, b = 17490, c = 17490
ISR - a = 917356, b = 917355, c = 917354
ISR - a = 770572, b = 770571, c = 770570
ISR - a = 42625, b = 42624, c = 42624
ISR - a = 264746, b = 264745, c = 264744
ISR - a = 666557, b = 666557, c = 666556
```

With `#undef ATOMIC`, all that the ISR ever sees is:
`ISR - a = 0, b = 0, c = 0`

That's why I am using `std::atomic` in my lock-free programming for good measure even on the single-core ESP8266.

cores/esp8266/Schedule.cpp
dok-net: I hope the latest commit clears this up sufficiently, if you don't think so, please just let me know. Were you asking for an additional comment in the cpp, or is it the one the h file?

cores/esp8266/Schedule.cpp
dok-net: Std::atomic & is out, std::function callback is in. Appears to be faster, even.

cores/esp8266/Schedule.cpp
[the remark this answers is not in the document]
dok-net: While this is still an interesting field of debate and research, I've opted to replace the reference to atomic bool logic with a std::function callback interface. Please re-review.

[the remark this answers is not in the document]
dok-net: Resolving this comment because std::atomic is no longer in there.

cores/esp8266/Schedule.cpp
dok-net: Done.

cores/esp8266/Schedule.cpp
```diff
@@ -80,27 +85,37 @@ bool schedule_function (const std::function<void(void)>& fn)
...
-    esp8266::InterruptLock lockAllInterruptsInThisScope;
+    if (!fn)
+        return false;
 
     recurrent_fn_t* item = new (std::nothrow) recurrent_fn_t(repeat_us);
     if (!item)
         return false;
 
     item->mFunc = fn;
+    item->alarm = std::move(alarm);
```
d-a-v: I believe `std::move()` is only needed when var is a rvalue reference and useless otherwise. Here alarm is passed by value (not constref, not rvalue-ref)
We use in this file constrefs for all functionals, why not use one for alarm at least for consistency  ?

d-a-v:
Out of curiosity, why is there a need for a FIFO when functions are called pseudo-asynchronously with potentially different timings ?

dok-net:
I commented this in the comment 2 months ago:

> Ordered, more predictable, scheduling. Before, it had different ordering compared to
> [...] sequential calls from loop().

This affected test results with the use of CoopTasks (in this context, think of these as preemptable scheduled functions, stackful coroutines).

PR #6489, opened 2019-09-06
Update UART selection for Boot ROM ets_putc, when debug port is selected.

mhightower83:
Add code to select the UART for Boot ROM `ets_putc` which is used by
`::printf`, `ets_printf_P` in `core_esp_postmortem.cpp` and others.
This only applies to a debug build where Debug Port has been selected.

`ets_putc` is a wrapper for `uart_tx_one_char`. `uart_tx_one_char` uses
the element `buff_uart_no` in `UartDev` (A structure in the data area of the
Boot ROM) to select UART0 or UART1. uart_buff_switch is used to set
that entry.

The structure for `UartDev` can be found in `uart.h` from the
ESP8266_NONOS_SDK. As best I can tell the Boot ROM always
defaults to UART0.


mhightower83:
Sorry - I need to rethink the placement of this call. I am now thinking it should be in function `uart_set_debug()` file `uart.cpp`

PR #6494, opened 2019-09-08
Add HardwareSerial::updateBaudRate(unsigned long baud) 

Jeroen88:
Add HardwareSerial::updateBaudRate(unsigned long baud) to change the baudrate after ::begin() was called.

### DIFF
--- a/cores/esp8266/HardwareSerial.cpp
+++ b/cores/esp8266/HardwareSerial.cpp
@@ -60,6 +60,15 @@ void HardwareSerial::end()
     _uart = NULL;
 }
 
+void HardwareSerial::updateBaudRate(unsigned long baud)
+{
+    if(!_uart) {
+        return;
+    }
+
+    uart_set_baudrate(_uart, baud);    
+}
+
 size_t HardwareSerial::setRxBufferSize(size_t size){
     if(_uart) {
         _rx_size = uart_resize_rx_buffer(_uart, size);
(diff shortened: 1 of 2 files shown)


Jeroen88:
@d-a-v or @earlephilhower, any chance this gets merged soon? :)

cores/esp8266/HardwareSerial.cpp
```diff
@@ -133,8 +142,8 @@ unsigned long HardwareSerial::detectBaudrate(time_t timeoutMillis)
     return detectedBaudrate;
 }
 
-size_t HardwareSerial::readBytes(char* buffer, size_t size)
-{
+size_t HardwareSerial::readBytes(char* buffer, size_t size)
```
earlephilhower: Is this a linefeed change here?  I think this whole file is DOS \r\n formatted, but your patch only has \UNIX n.  For now, could you make sure you match the linefeeds of the rest of the file (yes, I hate DOS linefeeds, too, but we need to be consistent in any particular file).

Jeroen88: I really do not know. I did not touch these lines. Could you please check it?

d-a-v: It is indeed a crlf change.
And after checking, this PR restores consistency which is not here in current master.
(The Allman process shall have a try-again session)

PR #6496, opened 2019-09-09
Minimize header use, move Ticker function definitions into cpp file

dok-net:
This is a corresponding port of ESP32 review results.


dok-net:
@devyte Are you certain that float multiplication is even inlined? I find that it's hard to know which case to optimize for. As you suggested, I am building a test that calls each attach/attach_ms/...scheduled/once/once_ms/...scheduled function once. Plus another test that does this, but for two ticker objects. From master to this PR, for one object, the ROM usage grows, for two objects, it shrinks. Inlining vs. header and single def in cpp doesn't help either. It's about inlining apparently saving flash for a single invocation, but naturally using more when called multiple times, vs. the linker eliminating code that's never called.
I guess I'd like to stick to the ESP32 review result, that demanded to move as much as possible from the header into cpp. *sigh*

devyte:
@dok-net I see you reverted the move from h to cpp. Did you compare the bin size lf your current proposal vs. unifying the float mult in a single attach method as is currently the case?

dok-net:
@devyte My comment in the ESP32 PR was more informative: [better than 204 bytes saved by inclining in header](https://github.com/espressif/arduino-esp32/pull/2849#issuecomment-539584322)

As to your precise question, I don't see the issue you're hinting at, as far as I can tell from objdump either, the multiplication is not unrolled/inlined, therefore it's rather insignificant in how many places a `1000UL * x` or `1000000UL * x` expression exists. This is from extensive trying out all sorted of combinations of inline vs. cpp impl, nested functions vs. copy&paste. The version I have in this PR is the only one that turned out well.

That said, I don't expect any single sketch to use all modes of the Ticker at once, YMMV if you do. But for using two Ticker instances, one with attach_ms, the other with scheduled_once, this is what I base my findings on and that's what this PR is quite good at.

devyte:
@dok-net the relevant code is in effect only when using attach() with a float arg.

dok-net:
@devyte Naturally, I've tested all sort of combinations, also on ESP32, where `attach_ms()` scales to µs (`1000ULL *`) :-)

dok-net:
@devyte CI is broken - not by me (?)

earlephilhower:
The pleasures of DevOps and needing sources from 15 different projects.  The Python version in the Win VM was changed, and installed into a different dir than before.  Fixed now, hopefully, permanently.

dok-net:
@devyte It's not on purpose, git doesn't recognize it as a move, the internal probably is to large. Could you do a local diff by hand, please?

Issue #905, opened 2015-10-18
OTA that survives power failure


plerup:
Maybe I'm missing something but I looks to me like if we have a power failure during the OTA update in eboot.c (ACTION_COPY_RAW) we end up with a non functional system.

Couldn't the data structure "eboot_command" be stored in flash instead and not be reset until ACTION_COPY_RAW operation has been successfully completed?

BR
/Peter

igrr:
It's certainly possible to pass bootloader arguments via flash. This will require allocating an additional sector for eboot_command (like it is done in other bootloaders).
By the way, another point of failure is when the actual image is written. If the power goes out between the moment when the first sector (which contains the bootloader itself) is erased and when it is written, system will not boot.

plerup:
Couldn't we have (optional) skip of updating the bootloader sector?

igrr:
Sure, such option may be implemented in Updater class.

plerup:
I could give this a try. Just not sure how to allocate the extra sector though. Never touched base with the linker before... Have to explore a bit first.

igrr:
The tricky part is to allocate this sector when users asks for it, so that flash space is not wasted by default. This could be an additional option in Tools menu or something like that.

plerup:
I guess the file system is not possible to load and use in the boot-loader

me-no-dev:
I'm thinking of something else... since there are sectors at particular locations that are used by the SDK to store settings and such, maybe we can use those? They are a full 4K blocks but data in them is usually not more than 128B, so using the end of one of them can maybe give you the space for storing that data and not mess with EEPROM or the linkers.

davisonja:
I'm a big fan of updates that survive power-failures (in as much as they don't render the device unusable).
I've got a revision of master that stores a command in flash, falling back to RTC if there's no valid flash block available, or there's no command (so we gracefully fall back to current behaviour).

Aesthetically I like the idea of stealing a 4k page at the start of flash - so eboot and its config are all kept together. On a module with 4M this is no real issue, but I realise that this is more of a concern in smaller ones so it does really need to be optional.
Having an optional page at the start makes life more complicated, so for dev I've opted to use a page at the end, modifying the .ld file(s) to reduce the SPIFFS space by 4k. While this works relatively nicely, there appears to be some use of the space at the end of flash - does anyone have a pointer to concrete documentation on which pages are used?
[...]

PR #6505, opened 2019-09-11
Support FS update in two steps

Androbin:
Problem: I am updating my sketch and FS over HTTPS and keep the certificates in FS like in the [httpUpdateSecure.ino](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266httpUpdate/examples/httpUpdateSecure/httpUpdateSecure.ino) example. If updating the FS fails (e.g. power failure during download) and the certificates get damaged, my customers might not be able to update and recover anymore. 

Solution: Support the same two step process that is used to update the sketch.

A step towards #905. Hopefully the option to move from RTC to flash will be merged soon also.


d-a-v:
where or when is `UPDATE_IN_TWO_STEPS` supposed to be defined ?
Is this PR dependant of #6533 ?

Androbin:
@d-a-v It's an independent effort. I'm using `#ifdef`, so not defining it is the point.
As long as this flag is not defined (by the user), it will default to the current behavior.

Androbin:
@d-a-v Friendly ping

d-a-v:
Can you elaborate on how this is supposed to be used when `UPDATE_IN_TWO_STEPS` is defined ?

earlephilhower:
The FS is presently updated right in Updater.cpp during the upload, and can end up in an inconsistent state should there be a dropout or other problem during update.

It's not like the app where the new copy is only uploaded if the entire upload and MD5 match.  What this looks to do is treat the FS the same way as the program, and upload to spare area and only ask the bootloader to overwrite it iff the upload is a success.

d-a-v:
I understand `UPDATE_IN_TWO_STEPS` is only used for FS, so can this macro be renamed to make this code more easily readable ?

earlephilhower:
@Androbin can you see to @d-a-v's request?  Maybe `ATOMIC_FS_UPDATE`or `UPDATE_FS_IN_TWO_STEPS`?

Also, now that I think about it, a short note somewhere in the docs/ota_updates about this define (since it seems unintuitive to some people) would be appropriate.

PR #6512, opened 2019-09-13
prepare alpha channel release builder

d-a-v:
tag '0.0.1' in the repository should trigger the build of a special '-0.0.1' release visible in arduino board manager.
This is intended to provide a regular snapshot of git head to arduino board manager.
This way of releasing alpha versions is a workaround to https://github.com/arduino/Arduino/issues/6012

closes #6511 

*edit*: https://twitter.com/mbanzi/status/1173532857857400832

### DIFF
--- a/package/build_boards_manager_package.sh
+++ b/package/build_boards_manager_package.sh
@@ -3,25 +3,37 @@
 
 #set -x
 
-# Extract next version from platform.txt
-next=`sed -n -E 's/version=([0-9.]+)/\1/p' ../platform.txt`
+ver=`git describe --tag`
+visiblever=$ver
+if [ "$ver" = 0.0.1 ]; then
 
-# Figure out how will the package be called
-ver=`git describe --exact-match`
-if [ $? -ne 0 ]; then
-    # not tagged version; generate nightly package
-    date_str=`date +"%Y%m%d"`
-    is_nightly=1
-    plain_ver="${next}-nightly"
-    ver="${plain_ver}+${date_str}"
-else
+    git tag -d 0.0.1
+    ver=`git describe --tag HEAD`
     plain_ver=$ver
+
+else
+
+    # Extract next version from platform.txt
+    next=`sed -n -E 's/version=([0-9.]+)/\1/p' ../platform.txt`
+
+    # Figure out how will the package be called
+    ver=`git describe --exact-match`
+    if [ $? -ne 0 ]; then
+        # not tagged version; generate nightly package
+        date_str=`date +"%Y%m%d"`
+        is_nightly=1
+        plain_ver="${next}-nightly"
+        ver="${plain_ver}+${date_str}"
+    else
+        plain_ver=$ver
+    fi
+    visiblever=$ver
 fi
 
 set -e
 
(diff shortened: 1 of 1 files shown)


d-a-v:
Releases are created when tagging the main (this) repository ([doc](https://github.com/esp8266/Arduino/tree/master/package))

With this PR, tagging with '0.0.1' will generate a release ("esp8266-0.0.1.zip") and update the package.json file to show a new version '0.0.1' that will be displayed by the arduino IDE in the bottom of the list of available versions (so users cannot accidentally update to this version).

Once the alpha release is generated, the tag '0.0.1' has to be removed from the main repository.


What I tested so far:

* generate an alpha release 0.0.1 by tagging my local repo with '0.0.1' and starting the release script
* put the generated files (.zip and .json) on my local webserver
  (URL to 0.0.1.zip file is changed in the json file to point to my webserver)
* configure the Arduino IDE board manager URL to my local webserver
* install this release 0.0.1 with board manager (sketch compile, upload, run)
* uninstall, reinstall 0.0.1, nothing is downloaded, everything is cached, still working
* make a random commit, regenerate alpha 0.0.1 (points above)
* Arduino IDE board manager sees nothing, so 0.0.1 is not updated from IDE
* uninstall 0.0.1 from board manager
* reinstall 0.0.1, this time the new release is downloaded because crc in the package.json file does not match the 0.0.1.zip in arduino cache (the old 0.0.1 is definitely lost)
* the new '0.0.1' release works and is indeed updated

And the most important, no version accumulation in arduino cache or installed version, which is the real goal of this operation: a workaround to arduino/Arduino#6012 which is what prevents to provide a "nightlies" release channel.

One caveat: if a user has an alpha version working, then remove/update later with a new version with bad issues, there is no path back to the previous one, it is definitely lost in the limbs of the past.

Now, it has to be tested on the real field.

Review for this process is welcome from @igrr @me-no-dev and current maintainers.

d-a-v:
Alpha release demo, check https://d-a-v.github.io/

Jason2866:
Install went smooth.

pfeerick:
> One caveat: if a user has an alpha version working, then remove/update later with a new version with bad issues, there is no path back to the previous one, it is definitely lost in the limbs of the past.

Perhaps a manual save of the package file on the local system _BEFORE_ uninstalling/reinstalling to refresh to the current nightly would be the workaround? The user could then unpack that in their hardware folder? I know it's self-defeating as far as 'easy for the end user'... cause you might as well just run the git version at that point... but still... it kinda... could of... should work? :laughing:

d-a-v:
> save of the package file (or esp8266 folder) on the local system BEFORE 

Totally.
But that would be a <<< complex >>> thing to do, similar the the complexity level for:

> 'cause you might as well just run the git version at that point... 

And I/we'd not store the 40MB zip file for every generated alpha.

BTW testing is at https://d-a-v.github.io

pfeerick:
zfs? Who needs zfs... just use btrfs + timeshift already!! :-P 

Didn't say it was a bright idea now, did I? :laughing: :rofl:

PR #6514, opened 2019-09-13
Add memmove_P, use it in String to ensure F() safety

### LINKED ISSUE #6384: feature request: memmove_P
This follows https://github.com/esp8266/Arduino/pull/6368#issuecomment-517816819
and would bring a better fix to wstring.cpp

earlephilhower:
memmove_P is now in libc, so use it to allow WString to handle F()
pointers without errors.

Supercedes #6368

Fixes #6384


earlephilhower:
Ran the failing test in https://github.com/esp8266/Arduino/issues/6367, and it passes with this update, too.  So no regression that I can see.

PR #6533, opened 2019-09-18
Merge eboot changes from davisonja

Androbin:
Solves #905
@davisonja @igrr


d-a-v:
Nice !
Can you merge your changes from `parseBoards.py` with `tools/boards.txt.py` ?

davisonja:
@Androbin awesome, if that's taken care of all the conflicts it should almost work! :)

@d-a-v `parseBoards.py` was a first step in an alternative approach to `boards.txt.py`; you don't want any of it.

The key goal that `parseBoards.py` was part of is to amend the `ld` files to include slightly different layout of things in flash, and to expose addresses for eboot to use. The relevant part of one of the files is https://github.com/esp8266/Arduino/pull/6533/files#diff-54af750862005b1a7d50aab347930ae1R17 - it's basically the same for each, but some of the numbers change due to the varying sizes of flash that are defined.

Along with the change to `eagle.app.v6.common.ld.h` we end up with a flash page reserved for the eboot commands, just after the "sketch area" and just before the "spiffs area", and the start address of that page stored in the very start of flash.

The net result is that we lose a page's worth of codespace for the sketch area, but will surive power outages since the commands are stored there. There's some minimal flash-page-wear-levelling which can be tweaked (I had it at something small like 16, with a view to the rest of the page being potentially used for something else that should persist, like a signed update key).

My schedule is still proving chaotic, but was aiming to look at this tomorrow 😁 

(I've not explored the travis fails yet)

davisonja:
@d-a-v @Androbin - looking at the travis errors I suspect it's simpler if I do this from my base.

I've pushed a new branch that has most of the code changes in it, and will patch up what's left - it's based off a pull from master that's (as of now) an hour or so old, so it's fairly up-to-date.

Most of the actual changes for my OTA stuff are trivial - the full merge of HEAD from my master fork is cluttered with other merges from upstream. Hopefully get most of it sorted today, and lodge a PR.

Androbin:
Closing in favor of #6538

PR #6534, opened 2019-09-19
segment size printout cosmetics

d-a-v:
*(updated)*
Adding hints into segment size printout:
```
Executable segment sizes:
IROM   : 276348          - code in flash         (default or ICACHE_FLASH_ATTR) 
IRAM   : 28904   / 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...) 
DATA   : 1284  )         - initialized variables (global, static) in RAM/HEAP 
RODATA : 3064  ) / 81920 - constants             (global, static) in RAM/HEAP 
BSS    : 26040 )         - zeroed variables      (global, static) in RAM/HEAP 
```

### DIFF
--- a/tools/sizes.py
+++ b/tools/sizes.py
@@ -25,11 +25,11 @@ import sys
 
 def get_segment_hints():
     hints = {}
-    hints['IROM'] = '      code in flash (default or ICACHE_FLASH_ATTR)'
-    hints['IRAM'] = '/32KB code in IRAM (ICACHE_RAM_ATTR, ISRs...)'
-    hints['DATA'] = '/80KB initialized variables (global, static) in RAM/HEAP'
-    hints['RODATA'] = '/80KB constants (global, static) in RAM/HEAP'
-    hints['BSS'] = '/80KB uninitialized variables ("zeroed": global, static) in RAM/HEAP'
+    hints['IROM'] = '         - code in flash         (default or ICACHE_FLASH_ATTR)'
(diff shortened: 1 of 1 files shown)


tools/sizes.py
```diff
@@ -23,6 +23,15 @@
 import subprocess
 import sys
 
+def get_segment_hints():
+    hints = {}
+    hints['IROM'] = '      code in flash (default or ICACHE_FLASH_ATTR)'
+    hints['IRAM'] = '/32KB code in IRAM (ICACHE_RAM_ATTR, ISRs...)'
```
earlephilhower: The LHS is in raw bytes, so I suggest the RHS should also be in bytes ("28444 out of 32768" is easier to understand for folks who don't know what 32*1024 is)

tools/sizes.py
```diff
@@ -23,6 +23,15 @@
...
+    hints['DATA'] = '/80KB initialized variables (global, static) in RAM/HEAP'
+    hints['RODATA'] = '/80KB constants (global, static) in RAM/HEAP'
+    hints['BSS'] = '/80KB uninitialized variables ("zeroed": global, static) in RAM/HEAP'
```
earlephilhower: These three are kind of funny.  As printed, it's incorrect.  You have 80K for all three summed, not 80K per.  Not sure how to present that.  Maybe 1 more line that's the "DRAM : %d out of 81333 bytes)" %  DATA+RODATA+BSS or something?

PR #6536, opened 2019-09-19
ESP8266WiFi examples - 'traditional' WiFiWebClient example

JAndrassy:
some commits last year modified the WiFiClient example to a non HTTP example.

The (WiFi)WebClient example started in Arduino Ethernet library and most Arduino networking libraries have some modification of it. I modified the version from WiFiNINA library for ESP8266WiFi and used my favourite test address, the Arduino ASCII logo because it is readable on Serial Monitor.

### DIFF
--- a/libraries/ESP8266WiFi/examples/WiFiWebClient/WiFiWebClient.ino
+++ b/libraries/ESP8266WiFi/examples/WiFiWebClient/WiFiWebClient.ino
@@ -1,0 +1,77 @@
+/*
+  Web client
+
+  This sketch connects to a website (http://arduino.cc)
+  using the WiFi module.
+
+  created 13 July 2010
+  by dlf (Metodo2 srl)
+  modified 31 May 2012
+  by Tom Igoe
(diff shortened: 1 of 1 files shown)


d-a-v:
If you are using linux or WSL, you can run the script `tests/restyle.sh` which will solve the CI 'style check' issue.
Also, like in all other examples, you can use this for SSID and PSK:
```
#ifndef STASSID
#define STASSID "your-ssid"
#define STAPSK  "your-password"
#endif

const char* ssid = STASSID;
const char* pass = STAPSK;
```

jandrassy:
@d-a-v, I changed the ssid and pass definition and run the restyle (it changed only spaces in the header comment)

jandrassy:
and the HTTPSRequest example?

earlephilhower:
Good point!  Probably worth reworking to a TLS web service and not HTTP.  Or using the HTTP requester with a HTTPS client, but I *think* we already have one of those.

PR #6537, opened 2019-09-19
Restore EEPROM address to prior released location

### LINKED ISSUE #6531: EEPROM location was changed
In the thirst place I have update to v2.5.2 and it was working ok, but when I have downloaded the newest branch from GIT, then I have discovered problem with EEPROM. 
In the newest branch EEPROM location was changed.
My problem is that I have put authentication key there. Now I have to have access to it.

EEPROM function is working fine. In v2.52 and with the newest. It is just not the same part of the flash memory.

How can I have access to old EEPROM location with the newest Arduino framework from GIT?

ESP-12S with 4MB, - 4M2M=4MB (FS:2MB OTA:~1019KB)

earlephilhower:
When the FS_END was adjusted to end on a full block (i.e. rounded down)
to avoid filesystem issues, but _FS_end was changed.  The EEPROM library
used _FS_end to implicitly calculate the start of the EEPROM data, so
this means after the _FS_end fix, EEPROM data written with prior
releases would "disappear."

Avoid the issue by explicitly calculating the EEPROM start location in
the linker, using the same formula as prior release.

Fixes #6531


mcspr:
Hey
If adding another symbol is a way to go, perhaps we can also add `_EEPROM_end`?

Back in april when #5989 was proposed, I also patched to my [home-brew ld script generator for ESPurna](https://github.com/xoseperez/espurna/blob/dev/dist/ldscript_gen.py) with the same fs_end calculation, but have not noticed any EEPROM issues because my 4m1m script reserved 16K for EEPROM (for wear-leveling writes) 🤷‍♂

earlephilhower:
@mcspr there's no real clean way to get what you're trying to do.  The way EEPROM works is to erase and write starting from the EEPROM beginning, and it erases/rewrites the sector on each commit, so there is no way to do wear leveling, w/o going to a new EEPROM setup (which wouldn't be compat with existing setups, sadly) that iterated over a set of sectors...

mcspr:
True, but that is exactly what is happening for that specific firmware because of the small shim library working on top of the eeprom logic - https://github.com/xoseperez/eeprom_rotate - where it calculates free space between sdk and fs_end and then uses a different sector for each write.

Given that right now there is also interest to create bootloader config sector (#6538), maybe there needs to be a consistent approach to add sort-of partitions

earlephilhower:
Partitions are part of the 3.0.0 stuff (one of the reasons for the move to esptool), and on the ESP32 are defined by Espressif.  I'm not following their 8266 SDK RTOS 3.0, but that's better left for another issue.

I can add an _EEPROM_end, sure, but for compatibility it can only ever be `_EEPROM_end = _EEPROM_start + 4096`, so for now I'm not seeing the utility.  It would also need some minor rework to the lib itself, because the lib now assume a 1-sector erase/write.

For now I'd like to have this just be a bugfix ASAP for 2.6.  We can work out the best way of getting your setup working in another issue (or track in @davisonja 's PR) or new PR since EEPROM wear leveling is a breaking change...

mcspr:
Btw, EEPROM at runtime specifically I think it does not even need to know where fs is and just count [-20KB (-5 sectors) from the ESP::getFlashChipSize()](https://github.com/xoseperez/eeprom_rotate/blob/377935a85f005d49b66de569204cec86246999c1/src/EEPROM_Rotate.cpp#L63)?

There is no specific problem with the PR, only that without `-DNO_GLOBAL_EEPROM` my builds would fail with `undefined reference` err because I need to keep custom .ld script up-to-date. But fix is pretty easy, just adding new symbol or INCLUDE the up-steam .ld script somewhere in the custom one. Perhaps I am also overthinking it and should use a different approach altogether instead of moving fs_end address down (as the only reason of that is to simplify sector calculations and use last one as an base point and counting down by a specified index)

It was not meant to derail the discussion by mentioning Rotate library, but I have not thought about adding it here before. I could try to update the existing class to optionally support it, hopefully without breaking too much.

earlephilhower:
Great.  Maybe it would make sense to put your suggestion in an issue and we can mark it for 3.0 consideration.  That way we won't forget!

d-a-v:
This rotate library was proposed here as a PR and instead moved to an arduino library #4493
[doc](https://arduino-esp8266.readthedocs.io/en/latest/libraries.html#improved-eeprom-library-for-esp-esp-eeprom) (should be moved to the 'Other libraries' paragraph below?)
https://github.com/jwrw/ESP_EEPROM

mcspr:
One quirk is that both want extern _SPIFFS_end, which is replaced with _FS_end.
https://github.com/xoseperez/eeprom_rotate/issues/6 had mentioned this some time ago, but I forgot about it...
I hot-fixed it via the .ld scripts too, but maybe there should be `PROVIDE ( _SPIFFS_... = _FS_... )` for each one in the local.eagle.app.v6.common.ld to avoid breaking such libs?

davisonja:
I'd have thought when _SPIFFS_ got replaced it happened everywhere. A
provide would seem like simple patch

PR #6541, opened 2019-09-20
Fix UDP send to IPv6 link local addresses

nomis:
lwIP's tcp/udp_connect() and tcp/udp_bind() functions automatically
set the zone if it is required but missing, but udp_connect() is not
used so this doesn't happen.

Explicitly set the zone to the default network interface if it is
required for the type of address being used. Otherwise there is no
zone set and packets to a link local destination don't go anywhere.


d-a-v:
Thanks !
What about a comment with your explanation ?
Did you try to simply call `udp_connect()` instead ?

nomis:
I've just tried using `udp_connect()` and lwIP isn't selecting the zone for a link local address because it tries to route to it and no source address has been provided... so it looks like TCP to link-local address won't currently work either.

d-a-v:
I had been testing TCP with `fe80:` and I remember it worked. Did you try ?

nomis:
No, I've not tested TCP at all. Did you set or bind a source address explicitly?

If it works without specifying a source address then it would be reasonable to apply this fix so that UDP works too.

d-a-v:
> No, I've not tested TCP at all. Did you set or bind a source address explicitly?

No.
I'd have to retry testing. You can too: take any tcp/http client example and change the address to your link-local IPv6 server address.

I assume this issue would exist only for client not server, but it can also be tried with server examples. Under linux, link-local server addresses must be extended with `%intfname` like `fe80::%eth0`.

devyte:
What is the status here?

devyte:
@nomis ping

d-a-v:
This change is also necessary for TCP (credit will be given).
Merging for UDP.

PR #6544, opened 2019-09-21
Add time to filesystem API

earlephilhower:
Support the ESP32 File::getLastWrite() call and setting the time on
all filesystems automatically (assuming the system clock has
been set properly and time(NULL) returns the proper time!).

Adds Dir::fileTime() to get the time of a file being listed, similar to
Dir::fileName() and Dir::fileSize().

Adds ::setTimeCallback(time_t (*cb)()) to File, Dir, and FS, allowing
users to override the default timestamp on a per-file, directory, or
filesystem basis. By default, a simple callback returning time(nullptr)
is implemented.

LittleFS uses the 't' attribute and should be backwards compatible.

SD/SDFS work and include wrappers for obsolete SdFat timestamp callbacks
using the MSDOS time.

This PR does not update SPIFFS, due to compatability concerns and a
possible massive rewrite which would make it possible to determine if an
old-style ot metadata enabled FS is present at mount time.

Includes an updated SD/listfiles and LittleFS_time example.

Replaces #6315


earlephilhower:
By removing the possibly data-corrupting changes required for SPIFFS time addition (see PR #6315), this PR should be able to be merged at 0 risk and pave the way for a future SPIFFS enhancement to include dynamic metadata and others (possibly w/the 3.0 partition table giving us the FS info as opposed to us manually poisoning things).

cores/esp8266/FS.cpp
```diff
@@ -315,7 +350,9 @@ File FS::open(const char* path, const char* mode) {
         DEBUGV("FS::open: invalid mode `%s`\r\n", mode);
         return File();
     }
-    return File(_impl->open(path, om, am), this);
+    auto f = File(_impl->open(path, om, am), this);
```
devyte:
```cpp
File  f(_impl->open(path, om, am), this);
```
?

cores/esp8266/FS.cpp
```diff
@@ -192,7 +205,9 @@ File Dir::openFile(const char* mode) {
         return File();
     }
 
-    return File(_impl->openFile(om, am), _baseFS);
+    auto f = File(_impl->openFile(om, am), _baseFS);
```
devyte:
```cpp
File  f(_impl->open(path, om, am), this);
```
?

cores/esp8266/FS.cpp
```diff
@@ -334,7 +371,9 @@ Dir FS::openDir(const char* path) {
         return Dir();
     }
     DirImplPtr p = _impl->openDir(path);
-    return Dir(p, this);
+    auto d = Dir(p, this);
```
devyte:
```cpp
Dir d(p, this);
```
?

cores/esp8266/FS.h
```diff
@@ -179,17 +189,26 @@ class FSConfig
 class SPIFFSConfig : public FSConfig
 {
 public:
-    SPIFFSConfig(bool autoFormat = true) {
+    SPIFFSConfig(bool autoFormat = true, bool enableTime = true) {
```
devyte:
```cpp
class FSConfig
{
public:
  enum fsid { FSId = 0x00000000 };
  FSConfig(bool autoFormat = true, fsid type = FSId)
    : _autoFormat(autoFormat), _type(type) {
  ...
  }
...
};

class SPIFFSConfig : public FSConfig
{
public:
  enum fsid { FSId = 0x53504946 };
  SPIFFSConfig(bool autoFormat = true, bool enableTime = true) 
    : FSConfig(autoFormat, FSId), _enableTime(enableTime) {
  ...
  }
...
};
```

devyte:
I understand the FS is using local time stamps. If so, how about using global time instead, like real platforms? That would disallow problems with changing timezone, changing DST, updating time manually, etc.

cores/esp8266/FS.h
earlephilhower: Actually, you found a bug not related to style here.  I should remove `enableTime` completely here, since this PR does not include the (massive) SPIFFS changes to allow it to actually store times.

earlephilhower:
> I understand the FS is using local time stamps. If so, how about using global time instead, like real platforms? That would disallow problems with changing timezone, changing DST, updating time manually, etc.

As we discussed online, this is what we're stuck with for SDFS/SD, since FAT doesn't use UTC, only local time.  It is possible, however, for users to override the setTimeCallback() to return UTC if they wish for LittleFS (or even FAT) and documented in the File section.

earlephilhower:
I couldn't find a good way to use the enum for FSId (each class' enum was a unique type so there were ugly casts), so I settled on a `static constextr uint32_t` member in each with the ID.  I believe that's the proper way to do it for something like this, but I'm open to suggestions on other ways.

hellochenwang:
where is the function time(NULL) referred in FS.h?

d-a-v:
In `tools/xtensa-lx106-elf/xtensa-lx106-elf/include/time.h`
(this is [newlib](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/time.h;h=3031590b441c5d2341dafacdb1db3bd467233e81;hb=HEAD#l59))

PR #6550, opened 2019-09-23
Add typedef for putc1, fn_putc1_t.

mhightower83:
Replaced relevent usage of `(void *)` with `fn_putc1_t`.
Correct usage of `ets_putc()`, returning 0, in libc_replacement.cpp
This PR assumes PR https://github.com/esp8266/Arduino/pull/6489#issue-315018841 has merged and removes `uart_buff_switch` from `umm_performance.cpp` in this PR.
Updated method of defining `_rom_putc1` to be more acceptable (I hope) to the new compiler.


cores/esp8266/umm_malloc/umm_performance.cpp
```diff
@@ -40,24 +40,16 @@ bool ICACHE_FLASH_ATTR get_umm_get_perf_data(struct _UMM_TIME_STATS *p, size_t s
...
-// Always returns character sent.
-int constexpr (*_rom_putc1)(int) = (int (*)(int))0x40001dcc;
-void uart_buff_switch(uint8_t);
+/*
+ * ROM builtin "print character function" _putc1, ignores CRs and sends CR/LF
+ * for LF, newline. This function is used internally by ets_uart_printf. It is
+ * also the function that gets installed by ets_uart_install_printf through a
+ * call to ets_install_putc1.
+ */
+#define _rom_putc1 ((fp_putc_t)0x40001dcc)
```
d-a-v: What about adding this in the PROVIDE list ?

mhightower83: That would be a good way to handle it. Now I feel like it needs a better name. Since `ets_install_uart_printf()` and `ets_uart_printf()` both use this internal function, I am leaning toward `_uart_putc1` as a better name. I hesitate to name it `ets_uart_putc1` since it is not formally exposed. Any thoughts?

devyte: I don't know if there is any naming rule in place, so I don't particularly mind any name. However, there is already a `ets_putc` in the provide list, so the difference between them should be explained, e.g. as a comment.

PR #6552, opened 2019-09-24
Add 20/26MHz Flash frequencies for slow/cheap flash chips on the Generic ESP Board

ChocolateFrogsNuts:
A number of non-genuine boards exist mainly from flea-bay sellers that
use under-sized and/or low quality flash chips which can not handle
a 40MHz FlashFreq properly.

This patch adds slower flash frequencies to the menu for generic ESP boards
so that these cheap knock-offs can be run in a stable manner, hopefully saving
some people a few headaches and keeping these boards out of landfill.

see #6366 final solution.


Jason2866:
Imho bad cheap flash chips will not become reliable with tricks.
The should simply avoided. Strange errors will occur and Arduino core will be blamed
So the bin is the best place for this crap

devyte:
@Jason2866 there's a reason that the generic board options allow selection of the flash frequency.

ChocolateFrogsNuts:
@Jason2866 agreed that if the chip is in fact bad, it won't become good and it belongs in the bin.

Where this is useful is when the chip is storing the data correctly and reliably, but during the read cycle is not getting the data stable on the output fast enough for a 40mhz clock, resulting in intermittent bad reads.
The lower clock does in fact work in these cases - which I have seen with XM branded flash chips.
It is the difference between "why does this board crash randomly" and "this board is rock solid".

Jason2866:
@d-a-v  I replace flash chips (Puya...). But most user wont do ;-)

PR #6560, opened 2019-09-27
Add interrupt section to docs

### LINKED ISSUE #6428: Document realloc/free being not interrupt safe
### Basic Infos

#### Platform

All

### Settings in IDE

All

### Problem Description

With #6274, several critical section flaws were uncovered. Among them is the fact that it is not safe to use realloc() or free() from within an ISR, except maybe under very specific conditions. 

At this time, there doesn't appear to be a way to make them safe, so the recommendation is: don't use them in an ISR, unless you know very well what you're doing.
"Don't use them" includes adding elements to mem-managed objects with contiguous memory, such as String, std::string and std::vector, as well as any user-implemented objects that could internally manage an array or similar. 

I think use of malloc is ok from an ISR. Therefore, adding elements to mem-managed objects with non-contiguous memory should be ok, e. g. std::list, std::forward_list, std::map, etc.

Use of malloc et al from ISRs or signal handlers is not supported even in full OSs like Linux (malloc is thread safe but not interrupt safe), so this non-safety shouldn't be news to programmers. Even so, the above needs to be documented for full clarity. This issue is meant to track that.

earlephilhower:
Adds a section on interrupts to the docs and lists the restrictions
    and warnings about running long tasks.

    Fixes #6428

### DIFF
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -9,18 +9,33 @@ and have several limitations:
 
 * Interrupt callback functions must be in IRAM, because the flash may be
   in the middle of other operations when they occur.  Do this by adding
-  the ``ICACHE_RAM_ATTR`` attribute on the function definition.
- 
-* Interrupts should not perform delay() or any long-running (>1ms) task.
-  WiFi and other portiong of the code can become unstable if interrupts
+  the ``ICACHE_RAM_ATTR`` attribute on the function definition.  If this
+  attribute is not present, the sketch will crash when it attempts to
+  ``attachInterrupt`` with an error message.  
+
+.. code:: cpp
+
+    ICACHE_RAM_ATTR void gpio_change_handler(void *data) {...
+
+* Interrupts must not call ``delay()`` or ``yield()``, or call any routines
+  which internally use ``delay()`` or ``yield()`` either.
+  
+* Long-running (>1ms) tasks in interrupts will cause instabilty or crashes.
+  WiFi and other portions of the core can become unstable if interrupts
   are blocked by a long-running interrupt.  If you have much to do, you can
   set a volatile global flag that your main ``loop()`` can check each pass
-  or use a scheduled function (which will be called by the OS when it is
-  safe, and not at IRQ level) to do the work.
+  or use a scheduled function (which will be called outside of the interrupt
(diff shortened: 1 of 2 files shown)


doc/reference.rst
```diff
@@ -1,6 +1,27 @@
...
+ 
+* Interrupts should not perform delay() or any long-running (>1ms) task.
+  WiFi and other portiong of the code can become unstable if interrupts
+  are blocked by a long-running interrupt.  If you have much to do, you can
+  set a volatile global flag that your main ``loop()`` can check each pass
+  or use a scheduled function (which will be called by the OS when it is
+  safe, and not at IRQ level) to do the work.
+
+* Memory operations can be dangerous and avoided in interrupts.  Calls to
+  ``new`` or ``malloc`` should be minimized, and calls to ``realloc`` and
```
devyte: minimized due to possibly long running time

doc/reference.rst
```diff
@@ -1,6 +1,27 @@
...
+  ``free`` must NEVER be used.
+
```
devyte: Of note: memory-managed objects that use contiguous memory, like String, std::string, std::vector, etc, must be used with care. Adding elements, e. g. ::push_back(), can trigger a realloc under the hood, which is forbidden within an ISR. 
Non-contiguous memory managed objects, like std::list, std::set, std::map, etc, don't trigger realloc() under the hood and should therefore be ok.

earlephilhower:
Updated w/your proofreading and requests.

> I believe there were some comments about slme of this already in section "why does my esp crash". It might be a good idea to cross check against that.

The FAQ does have a section on Interrupts, too, but I though it would be better to make the IRQ restrictions front-and-center *before* things crash, too.  The contents are pretty close and don't contradict each other.

If you're really against it, I can drop the section in the reference.rst and just add a the bit about memory allocation in the FAQ bit.

devyte:
Oh, I'm completely in favor of this. Users have missed the current section enough times that I think an additional specific section would be beneficial, even if the info is duplicated.
What I meant was to check for contradictions, which you already did, and missed info in the other section that could be added here, e. g. timing constraints.

PR #6566, opened 2019-09-29
Add a CRC32 over progmem and ESP.checkFlashCRC

### LINKED ISSUE #4165: feature request: CRC for  program memory
I do occasionally struggle with failing flash chips and are implementing a CRC32. From 0x400000 to 0x410000-1 which is the SPI flash mapped to the memory map. 
That part I got working, however, I need a matching, link-time generated CRC in memory. 

Question: 
1. is there a mechanism which does that already? (Would be awesome :-)
2. is there a linker option to place a (const) variable (my CRC) at a specific location in the memory, e.g. 0x410000-4 ?

earlephilhower:
Automatically embed a CRC32 of the program memory (including bootloader
but excluding any filesystems) in all images in unused space in the
bootloader block.

Add a call, ESP.checkFlashCRC() which returns false if the calculated
CRC doesn't match the one stored in the image (i.e. flash corruption).

Fixes #4165


cores/esp8266/Esp.cpp
```diff
@@ -439,6 +440,19 @@ bool EspClass::checkFlashConfig(bool needsEquals) {
     return false;
 }
 
+bool EspClass::checkFlashCRC() {
+    uint32_t flashsize = *((uint32_t*)0x40200ff8);
+    uint32_t flashcrc = *((uint32_t*)0x40200ffc);
```
d-a-v: > in unused space in the bootloader block.

With `--size 4088 --crc 4092`, how are theses addresses known ?

earlephilhower: Magic constants, didn't expect to sneak 'em through.  :)  `0x4020000 + 4088 == 0x40200ff8`, etc.  It's stealing the last 8 bytes of the 3K eboot block in the BIN.  Because we need flash absolute addresses in the BIN (it's post-link and just a blob) I can't think of a good way of marking them during the process to be less magical.  Maybe an export in eboot.h or something...

Having Python3 always available on all OSes makes a lot of things like this easy, compared to the tricks @igrr and others were thinking about when the request came in.

earlephilhower: Added comments to this as well as an example showing its use (self-corrupts and then self-corrects).

platform.txt
```diff
@@ -117,8 +118,9 @@ recipe.objcopy.eep.pattern=
 
 ## Create hex
 recipe.objcopy.hex.1.pattern="{runtime.tools.python3.path}/python3" "{runtime.tools.elf2bin}" --eboot "{runtime.tools.eboot}" --app "{build.path}/{build.project_name}.elf" --flash_mode {build.flash_mode} --flash_freq {build.flash_freq} --flash_size {build.flash_size} --path "{runtime.tools.xtensa-lx106-elf-gcc.path}/bin" --out "{build.path}/{build.project_name}.bin"
-recipe.objcopy.hex.2.pattern="{runtime.tools.python3.path}/python3" "{runtime.tools.signing}" --mode sign --privatekey "{build.source.path}/private.key" --bin "{build.path}/{build.project_name}.bin" --out "{build.path}/{build.project_name}.bin.signed" --legacy "{build.path}/{build.project_name}.bin.legacy_sig"
-recipe.objcopy.hex.3.pattern="{runtime.tools.python3.path}/python3" "{runtime.tools.sizes}" --elf "{build.path}/{build.project_name}.elf" --path "{runtime.tools.xtensa-lx106-elf-gcc.path}/bin"
+recipe.objcopy.hex.2.pattern="{runtime.tools.python3.path}/python3" "{runtime.tools.crc}" --bin "{build.path}/{build.project_name}.bin" --size 4088 --crc 4092
```
mcspr: If this is not optional, why not embed the script into the elf2bin.py?

PR is also missing PlatformIO integration because someone would need to add that script call. And while I was trying to come up with that, I am having real trouble reasoning about the dependency chain when final .bin is not final after all :/ crc32bin can have `--output` param, but that even more code

earlephilhower: Fair enough.  I was thinking of the UNIX philosophy (little utils which do exactly one thing).  Since systemd seems to have taken over the world, I guess I'll have to get with the times. :)  

I'll see about combining this and moving the crc32 to always be ROM safe in an upcoming push to the PR.

dok-net:
@earlephilhower Is this supposed to work out of the box just yet? Like pulling master, restarting the IDE, rebuilding my sketch.

I get on one board:
```
esptool.py v2.8
Serial port COM3
Connecting....
Chip is ESP8266EX
Features: WiFi
Crystal is 26MHz
MAC: […]
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Auto-detected Flash size: 16MB
Flash params set to 0x0290
Compressed 278784 bytes to 205050...
Writing at 0x00000000... (7 %)
[…]
Writing at 0x00030000... (100 %)
Wrote 278784 bytes (205050 compressed) at 0x00000000 in 18.9 seconds (effective 118.2 kbit/s)...
	The upload process has finished.
Hash of data verified.
Leaving...
Hard resetting via RTS pin...
```
```
Repeater example for EspSoftwareSerial
CRC check: ERROR
Vcc = 3020V
```
But, on another board:
```
esptool.py v2.8
Serial port COM5
Connecting....
Chip is ESP8266EX
Features: WiFi
Crystal is 26MHz
MAC: […]
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Auto-detected Flash size: 4MB
Compressed 278784 bytes to 205049...
Writing at 0x00000000... (7 %)
[…]
Writing at 0x00030000... (100 %)
Wrote 278784 bytes (205049 compressed) at 0x00000000 in 18.1 seconds (effective 123.3 kbit/s)...
	The upload process has finished.
Hash of data verified.
Leaving...
Hard resetting via RTS pin...
```
```
Repeater example for EspSoftwareSerial
CRC check: OK
Vcc = 3041V
```
Immediately noticeable is the 1 byte difference in compressed size.

tools/elf2bin.py
```diff
@@ -30,6 +30,9 @@
 ffreqb = { '40': 0, '26': 1, '20': 2, '80': 15 }
 fsizeb = { '512K': 0, '256K': 1, '1M': 2, '2M': 3, '4M': 4, '8M': 8, '16M': 9 }
 
+crcsize_offset = 4088
+crcval_offset = 4092
```
mcspr: Hm. I think it is off by one error, starting index should be 4087?

dok-net:
@mcspr Build+upload with `crcsize_offset = 4087` crashes on boot.

dok-net:
@earlephilhower Ah, OK, I never knew that **I HAVE TO** select the precise right board w.r.t. flash size. Building for LOLIN D1 mini Pro (16MB flash settings in config) it's fine:
```
Repeater example for EspSoftwareSerial
CRC check: OK
Vcc = 3018V
```
Please allow one remark: could you be more precise in the readme on the use of
`ADC_MODE(ADC_VCC);` - my first quite naïve take on "at the top of your sketch" was to place it inside `setup()` ;-)

tools/elf2bin.py
dok-net: @mcspr Build+upload with crcsize_offset = 4087 crashes on boot.

mcspr: Right, nvm :)

earlephilhower:
@dok-net, your comment explains it!  If you send the wrong flash size, the Updater will *change one byte silently* to match what's already set.  So the CRC check did succeed and found an error...what was in flash was not what you sent in!  :)

This was related to another bug (already fixed) where the Updater was *always* changing 1 byte of an uploaded image, even when it was a FS.

Signing guarantees that you created the bin being uploaded, and is a 1-time check on upload only.

CRC checks dynamically (when you call it) for changes to the flash vs. when you generated the bitstream.  Random bit flips, flash page failures, etc.

PR #6569, opened 2019-09-30
Remove duplicated sha1 implementation (Fixes #6568)

### LINKED ISSUE #6568: Large code size impact when using sha1 functions
### Basic Infos

#### Platform

- Hardware:				[ESP-12]
- Core Version:			2.5.2
- Operating System:		MacOS

### Problem Description

Using this code in a sketch:

```
  String s = "abc";
  return sha1(s);
```

causes the resulting firmware bin to grow by about 6kb of code size. this is predominantly due to the function SHA1Transform:

```
1076073460 00000269 T SHA1_Final
1076072872 00000467 t SHA1ProcessMessageBlock
1076153740 00004595 T SHA1Transform
```

The root cause for that is ./libraries/Hash/src/sha1/sha1.c which is a large-machine optimized SHA1 implementation using unrolled loops and other optimisations for speed that are likely extremely unuseful for esp8266. 

Interestingly in a typical sketch build there is also a 2nd sha1 implementation (at least) included from bearssl, which provides a similar api but with a much smaller footprint:

```
1076042492 00000049 T br_sha1_init
1076042440 00000051 T br_sha1_set_state
1076043468 00000123 T br_sha1_update
1076043592 00000235 T br_sha1_out
1076042584 00000881 T br_sha1_round
```

A very naive replacement of the lib/sha1 with the corresponding implementation from bearssl gives the expected footprint (code size) reduction, but it isn't clear to me if that is acceptable policy wise (dependency on bearssl in core hash lib). the impact  in typical Arduino sketch builds using sha1() and https related functions is that we have (at least) two sha1 implemenations in the resulting binary, and one of them being very excessive in size.

dirkmueller:
The Hash library had its own copy of a loop-unrolled sha1 implementation
adding a large code footprint for no good reason, as there are several
sha1 implementations already in tree (one in NONOS-SDK as well as one
in bearssl). Switching to the bearssl one is straightforward and removes
about 3kb of code size overhead.

Also cleanup some obvious inefficiencies (copy by value, string
summing, no reservation causing repeated reallocations) in the
implementation.


libraries/Hash/src/Hash.cpp
```diff
@@ -78,20 +75,21 @@ void sha1(const char * data, uint32_t size, uint8_t hash[20]) {
...
 
-void sha1(String data, uint8_t hash[20]) {
+void sha1(const String& data, uint8_t hash[20]) {
     sha1(data.c_str(), data.length(), hash);
 }
 
 String sha1(uint8_t* data, uint32_t size) {
     uint8_t hash[20];
-    String hashStr = "";
+    String hashStr((const char*)nullptr);
```
earlephilhower: Minor stylistic request:  This is equivalent to the cleaner `String hashStr;`  Could we use that here instead.

libraries/Hash/src/Hash.cpp
```diff
@@ -78,20 +75,21 @@ void sha1(const char * data, uint32_t size, uint8_t hash[20]) {
...
+    hashStr.reserve(20 * 2 + 1);
 
     sha1(&data[0], size, &hash[0]);
 
     for(uint16_t i = 0; i < 20; i++) {
         String hex = String(hash[i], HEX);
-        if(hex.length() < 2) {
-            hex = "0" + hex;
+        if(hash[i] < 0x10) {
```
earlephilhower: I know you just did a quick change, but the original was written in a weird way.  Can we make it a little cleaner?
````
for(uint16_t i = 0; i < 20; i++) {
    char hex[3];
    snprintf(hex, sizeof(hex), "%02x", hash[i]);
    hashStr += hex;
}
````

File: libraries/Hash/src/Hash.cpp [hunk 3]
```diff
@@ -78,20 +75,21 @@ void sha1(const char * data, uint32_t size, uint8_t hash[20]) {
     sha1((uint8_t *) data, size, hash);
 }
 
-void sha1(String data, uint8_t hash[20]) {
+void sha1(const String& data, uint8_t hash[20]) {
```
earlephilhower: Good optimization.

libraries/Hash/src/Hash.h
```diff
@@ -31,12 +31,12 @@ void sha1(uint8_t * data, uint32_t size, uint8_t hash[20]);
 void sha1(char * data, uint32_t size, uint8_t hash[20]);
 void sha1(const uint8_t * data, uint32_t size, uint8_t hash[20]);
 void sha1(const char * data, uint32_t size, uint8_t hash[20]);
-void sha1(String data, uint8_t hash[20]);
+void sha1(const String& data, uint8_t hash[20]);
```
earlephilhower: Can you add the `const` modifier to the `uint8_t *` prototype on line 30 and line 36 as well?  They don't copy anything, but they also are constant input parameters anyway.

libraries/Hash/src/Hash.cpp
dirkmueller: It is not equivalent. See the explanation in https://cpp4arduino.com/2018/11/21/eight-tips-to-use-the-string-class-efficiently.html - I would only repeat what is echoed there.

I was thinking about suggesting a fix for that in wstring but that would very likely be extremely disruptive for existing users.

libraries/Hash/src/Hash.cpp
dirkmueller: I wonder why we wouldn't do a single sprintf call then using 20 parameters instead of 20 in a loop?

libraries/Hash/src/Hash.h
dirkmueller: Will do

libraries/Hash/src/Hash.cpp
earlephilhower: Code size will be larger, probably.  And, less typing or chance of typo. :)

libraries/Hash/src/Hash.cpp
earlephilhower: Actually, it is. :)  Check WString's history, I rewrote the thing a while back to use small string optimization.  Strings of 0-11 chars take 0 heap now.

dirkmueller: Yes, thanks for reminding me on that part, I do remember having seen that but still at the risk of being borderline pedantic:

```
String::String(const char *cstr) {
    init();
    if(cstr)
        copy(cstr, strlen(cstr));
```

There is still a difference here, as the default constructor default argument for cstr makes  the "if (cstr)" condition true, which means it will wander off into strlen and ::copy. both are not very inefficient, but with the nullptr / reserve we're avoiding this extra code path. I agree it is a much smaller optimisation than it used to be, so thanks for doing the SSO optimisation..

My comment was about making the default argument nullptr not "" but that would mean all casual users of string would have to learn about the difference between null and empty and I think thats not going to be appreciated by the majority.

I can do the change as you asked for, but mine is still (marginally) more efficient..

dirkmueller: Also from looking at assembly:

```assembler```
402157fc:       fffc31                  l32r    a3, 402157ec <sha1(unsigned char const*, unsigned int, unsigned char*)+0x40>
...
40215809:       0ca785                  call0   40222284 <String::String(char const*)>
```

vs

```assembler
402157f8:       030c                    movi.n  a3, 0
..
40215805:       0ca785                  call0   40222280 <String::String(char const*)>
``` 

the initialisation to 0 (nullptr) is one byte shorter.

libraries/Hash/src/Hash.cpp
dirkmueller: WFM

earlephilhower:
Argh, looks like Arduino.cc updated their nightly builder and it's now SEGV'ing.  DevOps and CI are great fun and a source of perpetual employment. :(

libraries/Hash/src/Hash.cpp
earlephilhower: Well, you did save 3KB so a couple bytes shouldn't be a biggie. :)  Anyway, not worth sweating over and fine as is.

devyte: I suggest proposing a fix anyways.

dirkmueller:
@devyte the proposed change is definitely not making it faster, as it removes loop unrolling which is known to improve performance. However, there are two considerations to make: any sketch using sha1 in 2019 for business critical (performance) logic is probably at their own fault, and really shouldn't do that anymore given that sha1 is no longer "secure" and b) this is not adding a new implementation, it just uses the one from bearssl. Any bugfix/improvements made to sha1 performance in bearssl (where it actually if anywhere at all still matters) will automatically translate to an improvement here. The main purpose of this change is not to improve performance but to remove repetition. 

That being said, you probably want numbers anyway, so I made some strawman benchmark. all of the ones are the variant generating String() return values (e.g. a hexified sha1 hash). there is probably considerable overhead on that part alone. 

```
using sha(const char*, size, shasum) (80 MHz):

micros for abc: 632 (+56.1%)
micros for 2048 a's: 1096 (+32.5%)

using sha1(String) and 80 MHz:

micros for "abc": 604 (+50%)
micros for 2048 a's: 1623 (+32.1%)

using sha1(const char*, size) (80 Mhz):

micros for abc: 619 (+49.8%)
micros for 2048 a's: 1626 (+42.5%)

using sha1(const char*, size) (160mhz):

micros for abc: 344 (+42.7%)
micros for 2048 a's: 835 (+39.8%)
```

conclusions: 
* This is entirely cpu bound, increasing freq to 160MHz very proportionally translates into a speedup. 
* The optimisations in the hexifed-string-creation part of the sha1() function variants offset the cost of the new function somewhat, by reducing the slowdown by roughly 10%
* The actual SHA1 calculation slowdown is between 56.1% (for single block/very short input) and 32.5% for long input. I don't know what the typical input data length is though to say what the average expected slowdown is, but anyway it is still the same magnitude of runtime. 

So the numbers are surprising, because it is not clear why single-block(short) input should be 56% slower while longer input (where loop unrolling etc would be helpful) is only 32.5%. it is entirely possible that there is some performance tuning to make in bearssl's implementation for very short input, in any way this change actually makes that tuning effort worthwhile because we no longer duplicate the implementation twice. 

So on the other side we're saving a few kilobytes (!) of flash. which for all projects that I was involved in was worth a lot more than speed.

earlephilhower:
I would say that large block runs suffer from cache issues with the unrolled original code, if not directly in the sha1 implementation than in other routines that are going on in parallel.  So IPC will be lower than a smaller routine.

The much smaller block puts less pressure on the cache, so while it may be take X% more instructions executed, the average IPC has gone up due to lower cache misses.

In any case, neither runtime seems to be that significant given the code savings.

I should probably pull the Updater SHA256 into this class at some time, from BearSSLHelpers.  I honestly didn't even know this "library" (of only a single hash routine?!) even existed.

Issue #5528, opened 2018-12-19
Double I2C read in one transaction skips a clock pulse


maarten-pennings:
I was writing a clock stretch injector for https://github.com/esp8266/Arduino/issues/5340 and then I stumbled on another error in the driver file core_esp8266_si2c.c "Software I2C library for esp8266" function `twi_readFrom`.

The problem occurs when issuing two reads in one I2C transaction (i.e. with a repeated START in between). In this scenario, **one clock pulse is not generated** by the master: the clock pulse for the 9th bit (the ack bit) just before the repeated start. I tested this with the latest release 2_5_0_BETA2 (but it also 
fails on 2.4.2). 

Unfortunately, I don't know how to write a single ESP8266 program that detects this. Instead, I took an arbitrary I2C slave, and hooked a logic analyzer to the I2C bus

This is the test code running on the ESP8266 (see full sketch attached at the end):
```
  // From slave, read 2 bytes, then repeated START
  int res1=Wire.requestFrom(SLAVE_ADDR,(size_t)2,false); // No STOP yet!
 
  // From slave, read 2 bytes, then generate STOP condition
  int res2=Wire.requestFrom(SLAVE_ADDR,(size_t)2,true);  
```

and this logic analyzer capture shows the problem:

![missingacknack](https://user-images.githubusercontent.com/30388931/50230966-01267080-03ae-11e9-9208-35c95c5e23da.png)

I tried to fix the code, the following seems to work (but I have not thoroughly tested it):

![suggestedfix-missingacknack](https://user-images.githubusercontent.com/30388931/50231270-b0fbde00-03ae-11e9-9dcc-f2e8e1232c1c.png)

This leads to the following capture of the logic analyser:

![fixedacknack](https://user-images.githubusercontent.com/30388931/50231368-ee606b80-03ae-11e9-93ac-f52006a67b5d.png)

Here is the test sketch:
[ESP8266-i2c2xread.zip](https://github.com/esp8266/Arduino/files/2695557/ESP8266-i2c2xread.zip)

devyte:
@maarten-pennings Thank you for the analysis and for offering a fix!
I believe the current maintainers aren't involved with I2C, so help is needed to test this.

maarten-pennings:
I'm not yet familiar enough with the code, so I'm not sure if the fix works in all cases.
This was my attempt to locate the place in the code where something is wrong.

maarten-pennings:
[...]

PR #6579, opened 2019-10-01
Double I2C read in one transaction skips a clock pulse (#5528)

TD-er:

But I wondered if there should be an extra delay between the sequence of pulses and the extra added clock "valey". See my comment in the code.

Have not tested it myself, this is just the long awaited PR.

### DIFF
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -197,13 +197,16 @@ static bool twi_write_start(void) {
   return true;
 }
 
+static void ICACHE_RAM_ATTR twi_wait_clockStretchLimit() {
+  uint32_t t=0; while(SCL_READ()==0 && (t++)<twi_clockStretchLimit); // twi_clockStretchLimit
+}
+
 static bool twi_write_stop(void){
-  uint32_t i = 0;
   SCL_LOW();
   SDA_LOW();
   twi_delay(twi_dcount);
   SCL_HIGH();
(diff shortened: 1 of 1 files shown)


Jason2866:
@TD-er  Have you seen https://github.com/esp8266/Arduino/pull/6326 ?
We use this PR for our core pre2.6 Tasmota. We have until now NO feedback with issues with.
Could you adapt you PR to this. This I2C saves a lot of iRam!

TD-er:
Have seen it and indeed it may be useful to base this PR on that one.
I will have a look at it to see how hard it is to do the same based on that branch.

TD-er:
@Jason2866 See PR #6592

earlephilhower:
Minor niggle, but in English it is spelled "valley" (two ells).

> According to this documentation on I2C clock stretching it is not allowed to have some slave keep the clock low during transmission of a byte, only after an ack.

NXP (spinoff from Philips IIRC) says that clock stretching can be on a bit-by-bit basis, and that's the way it's (theoretically) done w/the existing I2C code in the repo:  https://www.nxp.com/docs/en/user-guide/UM10204.pdf

TD-er:
@earlephilhower In the other PR https://github.com/esp8266/Arduino/pull/6592 I had also changed it into "valley". Noticed the same strange notation :)

It makes more sense to allow clock stretching per bit, so the device can make sure it will be able to handle the stream.
But I wonder if it will then also be able to read if something is addressed to him/her/it or that it may stretch all bits on the bus even when not addressed to that device.
That would be bad for performance of the whole bus and maybe that's the reason it should not stretch the clock during transfers.

TD-er:
Superseded by #6654

PR #6581, opened 2019-10-01
Base64::encode : const correctness / String by reference passing

dirkmueller:
This is another instance in the core library where we pass in read-only
parameters as pass-by-value, where in the case of String() that
is inefficient as it involves copy-constructor/temp string creations.

### DIFF
--- a/cores/esp8266/base64.cpp
+++ b/cores/esp8266/base64.cpp
@@ -31,11 +31,11 @@ extern "C" {
 
 /**
  * convert input data to base64
- * @param data uint8_t *
+ * @param data const uint8_t *
  * @param length size_t
  * @return String
  */
-String base64::encode(uint8_t * data, size_t length, bool doNewLines) {
+String base64::encode(const uint8_t * data, size_t length, bool doNewLines) {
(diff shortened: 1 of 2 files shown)


dirkmueller:
@earlephilhower Inoticed the same, but I failed so far to "use" the String buffer for this. Only trick I know is with std::move().

earlephilhower:
I think you'd need to make `class base64` a friend to get access to the `String::wbuffer()` (aka writable buffer pointer) method (and String::reserve() the proper length first, of course).  Otherwise some sort of new `String::assumeBuffer()` or something would need to be added.  Either would engender more debate than this PR, so I'd worry about it later.

earlephilhower:
@dirkmueller I realized you can actually do inline base64 encoding directly into the string w/o any add'l String accessor methods.  Just pass in the String and instead of writing bytes to the array, just += the chars as they're generated.  Easy-peasy, and 1/2 the RAM needed.  But that's another PR. :)

devyte:
Don't forget to reserve() beforehand.

PR #6583, opened 2019-10-02
Pass String by const reference [3.0]

dirkmueller:
Passing String by value means a full copy-constructor/temporary
string creation which is slightly inefficient. Avoid that
by using const references.


devyte:
The reason they're virtual is to allow a user to implement his own RequestHandler objects.
Target must be v3.

earlephilhower:
Actually, @devyte, while what you say is *possible*, IMO they're virtual because we need to store a pointer to a generic RequestHandler in the webserver (i.e. internal considerations, not custom overloads). 

Overloaded classes would suffer object slicing if we stored a pointer to the base class in the WebServer object, so these functions need to be virtual (even though I was able to rewrite the WebServer class to be a template w/o virtuals).

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -95,7 +95,9 @@ class StaticRequestHandler : public RequestHandler<ServerType> {
         return true;
     }
 
-    bool handle(WebServerType& server, HTTPMethod requestMethod, String requestUri) override {
+    bool handle(WebServerType& server, HTTPMethod requestMethod, const String& __requestUri) override {
+        String requestUri(__requestUri);
+
```
d-a-v, line 100: This String copy can be handled differently in a further fixing commit.

PR #6585, opened 2019-10-02
Feature/travis jobs tmp

himanshugupta30062:
using nodemcu is better than using ESP8266 module


d-a-v:
nodemcu the hardware ?
It is an esp8266 module.

nodemcu the firmware ?
If you don't do or don't like Arduino, C++ or C, then yes another solution is to use Lua.

PR #6587, opened 2019-10-02
Block not only uart interrupts

### LINKED ISSUE #6564: Exception (0) in SoftAP when SPIslave used
### Basic Infos

#### Platform

- Hardware:					[ESP8285 device]
- Core Version:			
[SDK:2.2.1(cfd48f3)/Core:2.5.2=20502000/lwIP:STABLE-2_1_2_RELEASE/glue:1.1-7-g82abda3/BearSSL:a143020]

### Settings in IDE

- Module:  			[Generic ESP8285 Module]
- Flash Mode:			[other]
- Flash Size: 		[1MB]
- Reset Method:		[ck]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description
I am using SPISlave functionality and WiFiManager lib to config wifi connection.  External MCU makes some SPI requests to ESP. All works ok, except that moment, when I am trying to do any SPI request while ESP in STA+AP mode (while it switches to this mode, I believe). I found same issue, but with UART, several years ago: #1073. This "protection from UART interrupts" I found here: https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp#L429
Soon I will try to add `ETS_SPI_INTR_DISABLE()` and enable too and write here about results, but this issue is about another possible problem:
**Maybe we need to block all other unsafe interrupts in same critical sections, not only UART?** There also a plenty of another interrupts, which may cause same issues: https://github.com/esp8266/Arduino/blob/master/tools/sdk/include/ets_sys.h#L128

saloid:
Fixes #6564

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
@@ -139,13 +139,13 @@ bool ESP8266WiFiAPClass::softAP(const char* ssid, const char* passphrase, int ch
 
     if(!softap_config_equal(conf, conf_compare)) {
 
-        ETS_UART_INTR_DISABLE();
+        ETS_INTR_LOCK();
         if(WiFi._persistent) {
             ret = wifi_softap_set_config(&conf);
         } else {
             ret = wifi_softap_set_config_current(&conf);
         }
-        ETS_UART_INTR_ENABLE();
+        ETS_INTR_UNLOCK();
 
         if(!ret) {
             DEBUG_WIFI("[AP] set_config failed!\n");
@@ -290,13 +290,13 @@ bool ESP8266WiFiAPClass::softAPdisconnect(bool wifioff) {
     *conf.ssid = 0;
     *conf.password = 0;
     conf.authmode = AUTH_OPEN;
-    ETS_UART_INTR_DISABLE();
+    ETS_INTR_LOCK();
     if(WiFi._persistent) {
         ret = wifi_softap_set_config(&conf);
     } else {
         ret = wifi_softap_set_config_current(&conf);
     }
-    ETS_UART_INTR_ENABLE();
+    ETS_INTR_UNLOCK();
 
     if(!ret) {
         DEBUG_WIFI("[APdisconnect] set_config failed!\n");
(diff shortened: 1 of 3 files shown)


saloid:
I should use it in this way?:
```cpp
InterruptLock *lock=new InterruptLock;

// time critical section

delete lock;
```

d-a-v:
Easier than that:
```
// block to protect
{ 
    esp8266::InterruptLock lockAllInterruptsInThisScope;

    // code to protect
    ....
}
```
check this [example](https://github.com/esp8266/Arduino/blob/fb2cbe36a1c162baebc47e0a65210eeabb3eae31/cores/esp8266/Schedule.cpp#L65)

saloid:
In some places interrupts unlocks before scope ends. I think in this places better to use `delete`.

d-a-v:
new and delete are slow, increase heap fragmentation and useless in this case.
what places are you thinking of ?

saloid:
I have tested more:
with `ETS_INTR_LOCK`, `InterruptLock` or even with `AutoInterruptLock` with different (tried all 15) levels i still getting exception 0 inside `WiFi.begin` while requesting SPI status (and no exception if no SPI transfer happens at this time). But if I use `ETS_SPI_INTR_DISABLE` - all works fine.
Maybe in this case (to cover another possible interrupts) I need to write similar class to `InterruptLock`, but instead interrupt level store bit mask (`ets_isr_mask`)?

devyte:
@saloid said:
> In some places interrupts unlocks before scope ends

Please don't work on this over core v2.5.2. There are recent fixes related to lock/unlock that were merged recently and aren't released. What you said sounds like that. 
Instead, please make absolutely sure you're testing and working on latest git.

saloid:
Still getting same exceptions with the latest git version.

> @saloid said:
> 
> > In some places interrupts unlocks before scope ends
> 
> Please don't work on this over core v2.5.2. There are recent fixes related to lock/unlock that were merged recently and aren't released. What you said sounds like that.
> Instead, please make absolutely sure you're testing and working on latest git.

PR #6588, opened 2019-10-02
Put longer string literals into PROGMEM

dirkmueller:


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -115,6 +115,7 @@ class BearSSLTraits : public TransportTraits
 HTTPClient::HTTPClient()
 {
     _client = nullptr;
+    _userAgent = F("ESP8266HTTPClient");
```
devyte: Please use member construction here instead of assignment, like this:
```cpp
HTTPClient::HTTPClient() :
    _userAgent(F("blah")), _client(nullptr) 
{
 #if blah
... 
#endif
} 
```

The member construction order should be the same as the member declaration order.

dirkmueller: done.

PR #6593, opened 2019-10-03
Fix pgm_read_float_unaligned macro

### LINKED ISSUE #6590: pgm_read_float alignment issue when reading from an array of floats
### Basic Infos

#### Platform

- Hardware:					[ESP-12E]
- Core Version:			[2.5.2 (may-20-2019, installed via board manager]

### Settings in IDE

- Flash Mode:			[dout (compatible)]
- Flash Size: 		[4MB (3M SPIFFS)]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [512000] (serial upload only)

### Problem Description

My MmlMusic library reads tone frequencies as float values from a PROGMEM array. It worked fine for ESP8266 core version 2.5.0, but failed in cores version 2.5.1 and 2.5.2 
Using the defines mentioned in issue #5628 fixed the problem, but I guess it's better to use pgm_read_float_aligned() as mentioned in pull #5692 . The default define uses the _unaligned version, but as I noticed then the floats are not properly read from the PROGMEM array.
This issue is about compatibility. The Arduino library works fine on ATmega and ATtiny MCU's.
See the sketch below for more detailed info.

```cpp

#include <Arduino.h>

const float aftTest[] = {1.23, 2.34, 3.45};
const float aftTestPGM[] PROGMEM = {1.23, 2.34, 3.45};

void setup() {
  float ftTest=0.0;
  Serial.begin(115200);  
  Serial.println("\nTesting pgm_read_float");
  Serial.print("mem array:");
  ftTest=aftTest[1];
  Serial.println(ftTest);       // should print 2.34
// uncomment to fix pgm_read_float issue
//  #define pgm_read_float(addr)            pgm_read_float_aligned(addr)
  ftTest = pgm_read_float(&aftTestPGM[1]);
  Serial.print("pgm array:");
  Serial.println(ftTest);       // should print 2.34, but prints silly large number
}

void loop() {
}

```

### Debug Messages

```
Testing pgm_read_float (unfixed)
mem array:2.34
pgm array:1075167872.00
```

```
Testing pgm_read_float (fixed)
mem array:2.34
pgm array:2.34
```

earlephilhower:
Fixes #6590

The ASM block that implements the read-uint32-unaligned returns a
uint32_t.  The old code was doing a cast like `(float)(uint32_t ret)
which actually goes and creates a new float of the positive uint value
(approx, of course due to exponent and sign bits) which is not correct.

C and C++ don't have a concise way to convert the bits in a register
from int to float interpretation, so avoid the whole issue by making a
new function which uses the same ASM block as the read-uint32-unaligned,
just make the destination and return values as floats.

### DIFF
--- a/tools/sdk/libc/xtensa-lx106-elf/include/sys/pgmspace.h
+++ b/tools/sdk/libc/xtensa-lx106-elf/include/sys/pgmspace.h
@@ -80,6 +80,14 @@ static inline uint16_t pgm_read_word_inlined(const void* addr) {
   return (uint16_t) res;    /* This masks the lower half-word from the returned word */
 }
 
+/* Can't legally cast bits of uint32_t to a float w/o conversion or std::memcpy, which is inefficient. */
+/* The ASM block doesn't care the type, so just pass in what C thinks is a float and return in custom fcn. */
+static inline float pgm_read_float_unaligned(const void *addr) {
+  register float res;
+  pgm_read_with_offset(addr, res);
+  return res;
+}
+
 #define pgm_read_byte(addr)                pgm_read_byte_inlined(addr)
 #define pgm_read_word_aligned(addr)        pgm_read_word_inlined(addr)
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


earlephilhower:
Good catch, @maxint-rd!

PR #6594, opened 2019-10-03
Demonstrate how to use built-in SNTP client

davefiddes:
This change fixes a duplication in the sntp.h Arduino header which prevents use of sntp_getserver() and sntp_setserver() from being used with LwIP v2. It also include a simple example demonstrating how the SNTP client can be used to configure system time. SNTP time server(s) can be configured manually or via DHCP and/or DHCPv6.

### DIFF
--- a/libraries/ESP8266WiFi/examples/SNTP/SNTP.ino
+++ b/libraries/ESP8266WiFi/examples/SNTP/SNTP.ino
@@ -1,0 +1,140 @@
+
+/*
+  arduino SNTP example
+  released to public domain
+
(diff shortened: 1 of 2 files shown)


jandrassy:
https://github.com/esp8266/Arduino/blob/master/libraries/esp8266/examples/NTP-TZ-DST/NTP-TZ-DST.ino

davefiddes:
Interesting. I'd managed to miss that example somehow. It's demonstrating/testing a quite different part of the API and thus hasn't uncovered the bugs that my code fixes. I'll have a look at extending it.

PR #6598, opened 2019-10-03
Order class fields by size to avoid alignment padding

dirkmueller:
Mixing bools with other types that have higher alignment
requirement are not a good idea - they either require unaligned
access or padding bytes.


dirkmueller:
@devyte I understand the complexity, however embedded structs always get a struct padding based on the alignment requirements of the largest type inside, there is nothing that you can do about that other than being clever about the struct padding itself.

dirkmueller:
@devyte I should maybe update the commit message. this has been developed with -Werror=padded to get hints and manual resolution (there are too many errors to get everything building with that option).

PR #6600, opened 2019-10-04
use a scheduled function for settimeofday_cb

d-a-v:
fixes https://gitter.im/esp8266/Arduino?at=5d977042940b4c2fc074b949

### DIFF
--- a/cores/esp8266/coredecls.h
+++ b/cores/esp8266/coredecls.h
@@ -26,8 +26,11 @@ uint32_t crc32 (const void* data, size_t length, uint32_t crc = 0xffffffff);
 }
 
 #include <functional>
-void settimeofday_cb (std::function<void()>&& cb);
-void settimeofday_cb (const std::function<void()>& cb);
+
+using functionalVoidVoid = std::function<void()>;
+
+void settimeofday_cb (functionalVoidVoid&& cb);
+void settimeofday_cb (const functionalVoidVoid& cb);
 
 #endif
 
--- a/cores/esp8266/coredecls.h
+++ b/cores/esp8266/coredecls.h
@@ -27,10 +27,10 @@ uint32_t crc32 (const void* data, size_t length, uint32_t crc = 0xffffffff);
 
 #include <functional>
 
-using functionalVoidVoid = std::function<void()>;
+using TrivialCB = std::function<void()>;
 
-void settimeofday_cb (functionalVoidVoid&& cb);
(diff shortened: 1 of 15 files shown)


cores/esp8266/sntp-lwip2.cpp
```diff
@@ -42,16 +42,22 @@
 #include <osapi.h>
 #include <os_type.h>
 #include "coredecls.h"
+#include "Schedule.h"
 
-extern "C" {
+std::function<void()> _settimeofday_cb;
```
devyte: Please use anonymous namespace or static

cores/esp8266/coredecls.h
```diff
@@ -17,14 +17,18 @@ extern bool timeshift64_is_set;
...
 void disable_extra4k_at_link_time (void) __attribute__((noinline));
 
 uint32_t sqrt32 (uint32_t n);
 uint32_t crc32 (const void* data, size_t length, uint32_t crc = 0xffffffff);
 
 #ifdef __cplusplus
 }
+
+#include <functional>
+void settimeofday_cb (std::function<void()>&& cb);
```
devyte: Considering the number of repetitions I see, I think the arg type here should get a type alias, e.g.:
```cpp
using timeOfDayCB = std::function<void()>;
void setTimeOfDayCB (timeOfDayCB &&cb);
```

devyte: Also, please camelCase.

d-a-v: `settimeofday()` is a libc function.
Using `setTimeOfDayCB()` makes it quite unrelated with the system function.

cores/esp8266/coredecls.h
```diff
@@ -17,14 +17,21 @@ extern bool timeshift64_is_set;
...
 
 uint32_t sqrt32 (uint32_t n);
 uint32_t crc32 (const void* data, size_t length, uint32_t crc = 0xffffffff);
 
 #ifdef __cplusplus
 }
+
+#include <functional>
+
+using functionalVoidVoid = std::function<void()>;
```
devyte: This is rather confusing. 
How about just functionalVoid? Alternatively, a function that takes no args is called a nullary function. A function that doesn't return a value is called a void function. How about:
-nullaryVoid
-functionalNullaryVoid
-functionalVoidNullary
-some variation of the above

PR #6607, opened 2019-10-05
Cleanup base64::encode functions

dirkmueller:
The implementation choice here using libb64 is generally good as it
is a relatively fast implementation, however the adaptation to
use PROGMEM for the translation function was a bad choice, as reading
randomly PROGMEM with byte-wide access is very very very slow.

Doing a naive if-snake is between 10% and 50% faster and uses less
flash (about 120 bytes less) and also for reasons I don't understand
8 bytes less data RAM (maybe the removal of static?).


cores/esp8266/base64.cpp
```diff
@@ -39,33 +38,21 @@ String base64::encode(const uint8_t * data, size_t length, bool doNewLines) {
...
-        }
-        else 
-        {
-            base64_init_encodestate_nonewlines(&_state);
-        }
-        int len = base64_encode_block((const char *) &data[0], length, &buffer[0], &_state);
-        len = base64_encode_blockend((buffer + len), &_state);
+    String base64;
+    if (!base64.reserve(size)) {
+        return String("-FAIL-");
```
devyte: If here you do:
```cpp
base64 = "-FAIL-";
return base64;
```

and then you change the last return statement to:
```cpp
return base64;
```

then the compiler will be enabled for NRVO, aka copy elision, so no copies at all.

cores/esp8266/base64.h
```diff
@@ -1,39 +1,39 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ */
+
+#ifndef CORE_BASE64_H_
+#define CORE_BASE64_H_
+
+namespace base64 {
```
devyte: This may break current applications. It is the Arduino way to instantiate objects, even empty ones, before use. Current applications that have something like
```cpp
base64 somename;
...
somename.encode(blah);
```
anywhere in their code won't compile anymore. In contrast, the current implementation allows both instantiated and non-instantiated usage. 

If you feel strongly about this change, we can keep it, but it is a breaking change. In that case, we can target this entire PR for v3, or this breaking part of the PR can be split into a separate PR to be targeted for v3, and we only merge the other performance enhancement now. It's your choice.

cores/esp8266/libb64/cencode.cpp
```diff
@@ -23,10 +22,15 @@ void base64_init_encodestate_nonewlines(base64_encodestate* state_in){
...
-  if (value_in > 63) return '=';
-  return pgm_read_byte( &encoding[(int)value_in] );
+static inline char base64_encode_value(const char n) {
+  char r;
+
+  if (n < 26) r = n + 'A';
+  else if (n < 26 + 26) r = n - 26 + 'a';
+  else if (n < 26 + 26 + 10 ) r = n - 26 - 26 + '0';
+  else if (n == 62 ) r = '+';
+  else r = '/';
```
devyte: No single-line if-elses. Please put the bodies on their own lines.

cores/esp8266/libb64/cencode.cpp
```diff
@@ -23,10 +22,15 @@ void base64_init_encodestate_nonewlines(base64_encodestate* state_in){
   state_in->stepsnewline = -1;
 }
 
-char base64_encode_value(char value_in){
-  static const char encoding[] PROGMEM = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
-  if (value_in > 63) return '=';
-  return pgm_read_byte( &encoding[(int)value_in] );
+static inline char base64_encode_value(const char n) {
```
devyte: This function was public, and now it's private to the TU. That is a breaking change. Any application or library that currently makes use of it won't compile anymore.

cores/esp8266/libb64/cencode.h
```diff
@@ -33,8 +33,6 @@ typedef struct {
 void base64_init_encodestate(base64_encodestate* state_in);
 void base64_init_encodestate_nonewlines(base64_encodestate* state_in);
 
-char base64_encode_value(char value_in);
```
devyte: Removal of public function is a breaking change, see other comment.

cores/esp8266/base64.cpp
dirkmueller: yes, but I do have to call c_str() to do a deep copy because otherwise len() is set incorrectly to the initial value of 0 (it is not modified by the writeable begin()) . 

this is the part where we'd have to extend the public api of WString or add friend access.

cores/esp8266/base64.h
dirkmueller: thanks for the explanation, finally understood the reasoning behind it. reverted this part.

cores/esp8266/libb64/cencode.cpp
dirkmueller: well, is it really so that the libb64/* part is part of the public API? felt like a internal implementation accidentally leaking.

anyway, I reverted this part. although I would argue that libb64 is an implementation detail worth hiding..

cores/esp8266/libb64/cencode.h
dirkmueller: reverted

dirkmueller:
I'll play with NRVO a bit, but need to rebuild my test setup first before pushing the change.

cores/esp8266/libb64/cencode.cpp
dirkmueller: changed, is this how you prefer it?

earlephilhower:
Could we just use the ROM base64 functions with a tiny wrapper, instead?

````
earle@server:~/Arduino/hardware/esp8266com/esp8266/bootloaders/eboot$ grep base64 ../../tools/sdk/ld/eagle.rom.addr.v6.ld 
PROVIDE ( base64_decode = 0x40009648 );
PROVIDE ( base64_encode = 0x400094fc );
````

**Comment by dirkmueller** [comment]
I've seen that as well, but I couldn't find a header definition with the calling convention or any reference to it in the espressif nonos sdk documentation. Any idea where to find this? 

And can we remove libb64/ ? Otherwise this won't remove code size.. there was conflicting information about whether or not it is part of the api guarantees.

**Comment by earlephilhower** [comment]
Looking at the disassembly, I think it's safe to ignore my suggestion, @dirkmueller .  I see it does some math and uses the ROM malloc implementation, which is definitely a no-no for our apps.

File: cores/esp8266/base64.cpp (hunk 1 above)
Comment (dirkmueller) [reply cores/esp8266/base64.cpp]: changed by using friend access

**Comment by dirkmueller** [comment]
@devyte please reevaluate the 3.0 tagging. I think this can be merged any time now.

File: cores/esp8266/base64.cpp (hunk 1 above)
Comment (devyte) [reply cores/esp8266/base64.cpp]: friend access is a step in the wrong direction: it pollutes String with knowledge about base64, which has nothing to do with it. Please revert.
I don't follow your explanation about calling c_str(), wrong len, and deep copy. Please provide further details.

File: cores/esp8266/libb64/cencode.cpp (hunk 3 above)
Comment (devyte) [reply cores/esp8266/libb64/cencode.cpp]: No, I actually meant:
```cpp
if(blah1)
  bleh1;
else if (blah2)
  bleh2;
...
```

Braces are ok. In general, the preferred style for core code is Allman, while the Arduino-facing code (e.g.: examples) is Arduino style, which is a K&R variant.

cores/esp8266/base64.cpp
```diff
@@ -39,33 +38,20 @@ String base64::encode(const uint8_t * data, size_t length, bool doNewLines) {
     // base64 needs more size then the source data, use cencode.h macros
     size_t size = ((doNewLines ? base64_encode_expected_len(length)
                                : base64_encode_expected_len_nonewlines(length)) + 1);
-    char * buffer = (char *) malloc(size);
-    if(buffer) {
+    String base64("-FAIL-");
```
devyte: I just realized this local name shadows the class name. It would be wise to change it.

devyte: If you init like this, you're doing a double work internally: first copy the const char "_FAIL_" in, then change that when it's not used. Please reconsider my previous comment: init with default constructor, then assign "-FAIL-" on failure and return, or keep going. As I said, it also allows NRVO to work, which means no object copies on return.

cores/esp8266/WString.h
```diff
@@ -32,6 +32,9 @@
 // result objects are assumed to be writable by subsequent concatenations.
 class StringSumHelper;
 
+// class for base64 handling, needs access to setLen()
```
earlephilhower: This isn't going to pass muster, sorry.

On a practical level, we share this file with ESP32.  On an architectural level, having String and base64 mucking about each others' innards feels like a bad thing to do.  I understand where you're coming from, but there's got to be a better way.

cores/esp8266/base64.cpp
```diff
@@ -39,33 +38,20 @@ String base64::encode(const uint8_t * data, size_t length, bool doNewLines) {
...
+        else {
             base64_init_encodestate_nonewlines(&_state);
         }
-        int len = base64_encode_block((const char *) &data[0], length, &buffer[0], &_state);
-        len = base64_encode_blockend((buffer + len), &_state);
-
-        String base64 = String(buffer);
-        free(buffer);
-        return base64;
+        int len = base64_encode_block((const char *) &data[0], length, (char*)base64.begin(), &_state);
```
earlephilhower: Instead of this, can we have base64_encode_block take a String instead and use the `String += <char>` way of setting it.  That keeps base64 out of String's innards, while only requiring the single allocation of the block w/String::reserve.

cores/esp8266/base64.cpp
dirkmueller: I've removed the friend access. the issue is that String() is handling a string with 3 attributes: a buffer, a  capacity and a length. the capacity is the size of the memory allocation of the buffer, and the length ist he actual length of the current string stored (e.g. this is an optimisation instead of having to call strlen(buffer) all the time. 

What I did was to increase capacity of the buffer to the desired size, and then pass the buffer to the libb64 functions as the target output buffer. This works fine except that nothing ever modifies the "length" attribute of the string, and for that I need to be able to call setLen() (which is protected, hence friend). 

The only other option is to "append" a char buffer, which is less efficient - it calls strlen() on the buffer.

cores/esp8266/libb64/cencode.cpp
dirkmueller: done Allman without braces ;-)

cores/esp8266/WString.h
dirkmueller: I agree with you, I also think the friend access is a mistake. did another variant now.

cores/esp8266/base64.cpp
dirkmueller: well, thats the question. in the discussion earlier it was said that libb64/ is part of the API and can't be changed, so I don't want to create yet another wrapper around it. 

if we stop considering libb64/ being part of the API then I'd be happy to fold this into base64.cpp and clean it up and make it use String() directly.

cores/esp8266/base64.cpp
dirkmueller: done.

dirkmueller:
@devyte please also reconsider the "target 3.0.0" labeling here, there is no API change anymore, so this should be suitable for any release.

devyte:
Done.

cores/esp8266/base64.cpp
```diff
@@ -35,37 +34,43 @@ extern "C" {
...
-}
 
-/**
- * convert input data to base64
- * @param text const String&
- * @return String
- */
-String base64::encode(const String& text, bool doNewLines) {
-    return base64::encode((const uint8_t *) text.c_str(), text.length(), doNewLines);
+    return base64;
```
earlephilhower: @devyte, the C++ guru, is this the right way to minimize the copying of return values?  Copy elimination or whatever it's called?  We want to hand over the pre-existing String object (created in this method's stack) instead of doing a 2nd malloc/copy.

devyte: Single return statement, looks good for nrvo.

PR #6609, opened 2019-10-06
Add simple RLE compression for ArduinoOTA

### LINKED ISSUE #4277: Add run-length encoding for compressing OTA SPIFFS?
#### Hardware
Hardware:			ESP-12F
Core Version:      	2.4.0

### Description

Would it be possible to add a very simple run-length encoding scheme to compress SPIFFS file as they are transferred to the ESP8266 ? Because the SPIFFS are generally limited to a few options (e.g. 1MB or 3MB) users are very likely to have unused space in their SPIFFS image, which is easily compressible runs of 0xFF.

Alternatively, transfer pages of SPIFFS data that have data in them (uncompressed), but embed a code to indicate to the ESP8266 that "all following pages are to be written as empty" ?

In my tests it takes 51 secs to transfer a 1MB SPIFFS image OTA to an ESP8266 over a strong WiFi signal, or about 20KB/sec.

(My workaround for now is defining a small SPIFFS area (e.g. 64KB) in my boards.txt that will upload in just a few seconds.)

earlephilhower:
When uploading large, sparse filesystems, there are often many repeated,
empty sectors.  These take a long time to transmit, even over WiFi.

Implement a basic RLE compression in espota.py (and only use it if it
actually saves upload size), and a simple decompressor in ArduinoOTA.

The actual flash update image is decompressed as received to the staging
location, so the bootloader/etc. do not need to be aware of it (i.e. a
1MB filesystem still takes 1MB of flash in the update area, even if it
only took 100KB to upload it).

Fixes #4277

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -30,10 +30,114 @@ extern "C" {
 #endif
 #endif
 
+
+/* This class is only used in OTA and implements a dumb, low-memory decompression engine */
+class RLEDecompressor : public Stream {
+public:
+  RLEDecompressor(WiFiClient client) {
+    _client = client;
+    _blockLen = 0;
+    _blockIdx = 0;
+    _block = new uint8_t[128];
+  }
+
+  virtual ~RLEDecompressor() {
+    delete[] _block;
+  }
+
+  virtual int read() {
+    int ret = -1;  // Default to EOF
+    if (_blockLen == _blockIdx) {
+        _refill();
+    }
+    if (_blockIdx < _blockLen) {
+        ret = _block[_blockIdx++];
+    }
+    return ret;
+  }
+
+  virtual int peek() {
+    return -1; // Not implemented, not needed for Updater
+  }
+
+  size_t read(uint8_t*a, size_t&b) { return readBytes((char*)a, b); }
+
+  virtual size_t readBytes(char *buffer, size_t length) {
+    if (_blockLen == _blockIdx) {
+      _refill();
+    }
+    int toRead = std::min((int)(_blockLen - _blockIdx), (int)length);
+    memcpy(buffer, _block + _blockIdx, toRead);
+    _blockIdx += toRead;
+    return toRead;
+  }
+    
+  virtual size_t write(uint8_t b) { return _client.write(b); }
+
+  virtual int available() {
+    if (_blockLen == _blockIdx) {
+      _refill();
+    }
+    return _blockLen - _blockIdx;
+  }
+
+
+private:
+  bool _refill() {
+    int c = -1;
+    while (_client.connected() && (c < 0)) {
+      c = _client.read();
+      yield();
+    }
+    if (c < 0) {
+       return false;
+    }
+    if (c < 128) {
+      int l = -1;
+      while (_client.connected() && (l < 0)) {
+        l = _client.read();
+        yield();
+      }
+      if (l < 0) {
+         return false;
+      }
+      memset(_block, l, c);
+      _blockLen = c;
+      _blockIdx = 0;
+      return true;
+    } else {
+      c = c - 128;
+      _blockLen = c;
+      _blockIdx = 0;
+      while (_client.connected() && c) {
+        int ret = _client.readBytes(_block + _blockIdx, c);
+        if (ret > 0) {
+          c -= ret;
+          _blockIdx += ret;
+        } else {
+          _blockIdx = _blockLen = 0;
+          return false;
+        }
+      }
+      _blockIdx = 0;
+      return true;
+    }
+  }
+
+private:
+  WiFiClient _client;
+  int        _blockLeft;
+  int        _blockIdx;
+  int        _blockLen;
+  uint8_t    *_block;
+};
+
+
 ArduinoOTAClass::ArduinoOTAClass()
 : _port(0)
 , _udp_ota(0)
 , _initialized(false)
+, _useCompression(false)
 , _rebootOnSuccess(true)
 , _useMDNS(true)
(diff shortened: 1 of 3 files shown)


earlephilhower:
Forgot to note in the commit, this change transparently supports uploading to older ArduinoOTA servers by adding the "I can compress" flag at the end of the invite, in a section the older OTA simply ignores.

earlephilhower:
> 1. The RLE algorithm is implemented entangled with Stream, and the class even has a WiFiClient inside. It is therefore usable only in this specific scenario, and I think it would be very difficult to extend for usage in other scenarios....snip

It's a `Stream` because it does `stream`ing decompression.  That's the neat trick, really.  You cannot use an interface like base-64 since that requires in-memory space for the complete decompressed unit (aka 1MB in this case).  The `RLEStream` class could be lifted out and made to use a generic `Stream` input (i.e. a pure `Stream` filter), which would make it usable anywhere you wanted.  Now it's like a UNIX pipe, you stick a WiFiClient (should be a `Stream*) and anything that eats a `Stream` can use it.

The decompression can also be pushed down into Updater (maybe a special header to ID it), but then all the MD5 calculations would need to be adjusted to the RLE compressed bin.

If minigz also gets ported, it too should have a `Stream` interface.  However I have doubts about RAM usage given (and this seems an algorithmic requirement): https://zlib.net/zlib_tech.html
>      inflate memory usage (bytes) = (1 << windowBits) + 1440*2*sizeof(int) 
That's 20KB + the window size (8KB or whatever).  So it would need to be done while in eboot (and therefore minigz code image needs to fit in the ~2kb of free eboot space).  Having 28KB+ free while doing OTA is possible (not with HTTPS), but I doubt it's common...

> 2. The implementation is specific to ArduinoOTA-espota.py. It doesn't allow use with e.g.: OTA over webserver or via httpclient or other means. In addition, the compression code is contained within espota.py...snip

Fair enough.  We don't really import local modules into the existing Python code, but the actual compression bit is a 20 line function that can migrate to its own file simply enough.  

> 4. I had outlined the requirements for bin compression in this comment, where point 2 references RLE. This implementation doesn't address any of the cases there.

Re the 1st point there, no compression guarantees it'll actually be able to shrink data (your point about uploading larger images than free flash), so best-effort is all you can hope for there.  RLE does speed uploads, a lot, for sparse FSes, so the 2nd point is taken care of.  It also doesn't slow down anything since it checks compressed size < raw before sending compressed.  Bin size increase for RLE is about as small as you're likely to get, 3rd point.  As Meatloaf said, "Two out of three ain't bad." :)

> 5. <basically put decompress in eboot, store compressed in update region> 

That's a good idea, and I can almost guarantee a simple RLE decompress will fit in the eboot sector.


I hear your concern, though, so let's consider this a strawman thrown out there and see where it goes.

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -30,10 +30,114 @@ extern "C" {
...
+    _blockLen = 0;
+    _blockIdx = 0;
+    _block = new uint8_t[128];
+  }
+
+  virtual ~RLEDecompressor() {
+    delete[] _block;
+  }
+
+  virtual int read() {
```
dirkmueller: add 'override' ?

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -30,10 +30,114 @@ extern "C" {
...
+    if (_blockLen == _blockIdx) {
+        _refill();
+    }
+    if (_blockIdx < _blockLen) {
+        ret = _block[_blockIdx++];
+    }
+    return ret;
+  }
+
+  virtual int peek() {
```
dirkmueller: add 'override' ?

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -30,10 +30,114 @@ extern "C" {
...
+    return -1; // Not implemented, not needed for Updater
+  }
+
+  size_t read(uint8_t*a, size_t&b) { return readBytes((char*)a, b); }
+
+  virtual size_t readBytes(char *buffer, size_t length) {
```
dirkmueller: add 'override' ?

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -30,10 +30,114 @@ extern "C" {
...
+    }
+    int toRead = std::min((int)(_blockLen - _blockIdx), (int)length);
+    memcpy(buffer, _block + _blockIdx, toRead);
+    _blockIdx += toRead;
+    return toRead;
+  }
+    
+  virtual size_t write(uint8_t b) { return _client.write(b); }
+
+  virtual int available() {
```
dirkmueller: add 'override' ?

dirkmueller:
What's the code size overhead ? I like the feature, but ArduinoOTA is one of the classes that many will use, and the footprint increases in the last few releases didn't make it easy to upgrade.

earlephilhower:
@dirkmueller it adds 976 bytes to the ArduinoOTA code and ~150 bytes of heap when used.  RLE was used to be a simple and low-mem way of compressing FS updates.  This doesn't really shrink apps, so they'd just use uncompressed uploads w/the logic in espota.

dirkmueller:
Thanks for the quick reply. Imho the 976 bytes addition is quite a lot. Can we make it a compile time option? The current way it is implemented means I think that it will always be included.

earlephilhower:
Closing this for now.  I should be able to fit this into eboot with room to spare (so 0 impact on program size), and push a new PR that way.  Avoid the whole issue of other modes not being able to utilize it, too.

PR #6611, opened 2019-10-06
Expand NTP-TZ-DST example to list SNTP servers

davefiddes:
This change expands the NTP-TZ-DST example to list the active
SNTP servers. It lists the IP address server name (if used) and
reachability in ntpq style octal format.

Tests:
 - Build against all LwIP stack configurations and correct
   operation with a single SNTP server
 - Vary connectivity to an SNTP server and check reachability
   updates
 - Configure 3 SNTP servers by name
 - With a patched LwIP stack that supports simultaneous DHCP
   and DHCPv6 SNTP servers check that all server details are
   listed and update correctly during operation


davefiddes:
Thanks David. This change works fine with or without my other fixes.

PR #6613, opened 2019-10-07
MDNS: fix legacy unicast responses

mcspr:
When trying to resolve mdns hostname using lwip built-in resolver (`WiFi.hostByName("something.local")`), LeaMDNS responder does not construct a response that lwip expects. While it does preserve the request ID here:
https://github.com/esp8266/Arduino/blob/48ca7cfce3de99f1d906e95938a67895e73f7f4a/libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp#L248-L250
This ID is never set when preparing the message:
```
[MDNSResponder] _parseQuery: Legacy query from local host 10.0.0.22, id 30!
...
[MDNSResponder] _prepareMDNSMessage: ID:0 QR:1 OP:0 AA:1 TC:0 RD:0 RA:0 R:0 QD:1 AN:1 NS:0 AR:0
...
```
(debugging message could be improved though)

### DIFF
--- a/libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp
+++ b/libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp
@@ -245,7 +245,7 @@ bool MDNSResponder::_parseQuery(const MDNSResponder::stcMDNS_MsgHeader& p_MsgHea
                          ((wifi_get_ip_info(STATION_IF, &IPInfo_Local)) &&
                           (ip4_addr_netcmp(&IPInfo_Remote.ip, &IPInfo_Local.ip, &IPInfo_Local.netmask))))) { // Remote IP in STATION's subnet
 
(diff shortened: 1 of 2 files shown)


mcspr:
For posterity's sake, small example with `dig` failing:
```cpp
#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>

#define HOSTNAME "something"

#ifndef USE_SETTINGS
#define WIFI_SSID "SSID"
#define WIFI_PASSWORD "PASSWORD"
#else
#include "settings.h"
#endif

void setup() {

    Serial.begin(115200);
    Serial.println();
    Serial.println();

    delay(1000);

    Serial.printf("> MDNS test\n");
    Serial.printf("> hostname=%s ssid=%s password=%s\n",
        HOSTNAME, WIFI_SSID, WIFI_PASSWORD
    );

    WiFi.persistent(false);
    WiFi.mode(WIFI_STA);
    WiFi.begin(WIFI_SSID, WIFI_PASSWORD);
        delay(500);
        Serial.print(".");
    }

    Serial.println();
    Serial.println("> connected")

    MDNS.begin(HOSTNAME);

}

void loop() {
    MDNS.update();
}
```
```bash
$ dig -p 5353 @224.0.0.251 something.local
;; Warning: ID mismatch: expected ID 12327, got 0
```

Issue #6559, opened 2019-09-27
XMC Flash chip output drive support


ChocolateFrogsNuts:
### Basic Infos
- [ X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X ] I have tested that the issue is present in current master branch (aka latest git).
- [X ] I have searched the issue tracker for a similar issue.
- [X ] If there is a stack dump, I have decoded it.
- [X ] I have filled out all fields below.

#### Platform

- Hardware:			other - XMC Flash chips attached to any ESP8266
- Core Version:			[-]
- Development Env:		[Arduino IDE]
- Operating System:		[N/A]

### Settings in IDE

- Module:  			[Wemos D1 mini]
- Flash Mode:			[qio]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz|160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [460800] (serial upload only)

### Problem Description

XMC Flash chips used on a number of boards (most notably some WEMOS D1 mini) have a power saving feature that defaults to 75% drive on their outputs. This can result in unstable operation at 40MHz and above in some circumstances.

While #6552 allows lowering the flash frequency to achieve stable operation, it is possible to run the chips at the full 80MHz with 100% drive by setting bits 5 and 6 of SR3 with 
esptool.py write_flash_status --bytes 3 0x600000
Unfortunately this only lasts for the current power cycle.

Following on from the discussion started at the end of #6366, we now attempt to develop some code that will set SR3:5,6 in XMC flash chips when they are detected...

ChocolateFrogsNuts:
Ok, got code that sets SR3 at https://github.com/ChocolateFrogsNuts/ESP-DebugTools/blob/master/DebugTools/flash.ino
but for some reason it's not able to read SR3 yet... once that's sorted I can organize a PR with some neater/stripped down code suitable for a startup check.

ChocolateFrogsNuts:
Update: got it reading the flash status registers properly. 
[...]

PR #6619, opened 2019-10-07
Puyafix

d-a-v:
`PUYA_SUPPORT` was not correctly enabled by default.

It was originally defined to 0 by default in `Esp.h` and later to 1 in `Arduino.h` (!).
Its definition is now only in `Esp.cpp`.

@TD-er should the definition of `PUYA_SUPPORT` be visible if it is not user-defined ?

### DIFF
--- a/cores/esp8266/Arduino.h
+++ b/cores/esp8266/Arduino.h
@@ -285,10 +285,6 @@ void configTime(const char* tz, const char* server1,
 
 #include "pins_arduino.h"
 
-#ifndef PUYA_SUPPORT
-#define PUYA_SUPPORT 1
-#endif
-
 #endif
 
 #ifdef DEBUG_ESP_OOM
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -18,7 +18,7 @@
  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  */
 
-#include "Arduino.h"
+#include "Esp.h"
 #include "flash_utils.h"
 #include "eboot_command.h"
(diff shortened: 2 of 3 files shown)


TD-er:
I do use it to create a descriptive string for the core libraries and give some warning if support is not enabled and a puya device is found.
So it would be a nice to have if it is "publicly" visible.
But you already have it wrapped in #ifndef statements, so I can define it in the platformio.ini file and thus still have it publicly available.
Only thing is, if someone has made a build of its own it may not show the errors.

Maybe you could add a public function for "non supported flash found" ?

d-a-v:
> Maybe you could add a public function for "non supported flash found" ?

Could you please elaborate ?

TD-er:
Well, in the CPP file you now have the define for PUYA support.
So that's not "public", which is fine.
But to be able to check whether or not PUYA support is enabled, it could check for the existence of a puya chip and whether it has PUYA support enabled.
Maybe later there will be other patches for other non standard flash chips, so I thought it would be a bad idea to have this check function named after "puya", so maybe a more generic function name to check whether we may expect issues with the current firmware.

`bool nonSupportedFlashChipFound();`
Or something like that.

d-a-v:
it would be interesting to have a flash config/state bitfield-like API.
With #6559 we would be able to know whether the chip is
- unsupported (like puya, reporting when fix os not enabled)
- underclocked (can run faster, like XMC, reporting when not enabled)

TD-er:
That would be great indeed.
Such information can really help in narrowing down the cause of reported issues.
What is left then is to do the correct interpretation of those bits :)   (e.g., is underclocked more stable? I would expect it to be)

PR #6653, opened 2019-10-19
Fix inconsistent CRLF/LF combinations

dirkmueller:
These files had either CRLF, LFCR or LF or a mixture of it.
Consistently format them as git text files (unix style) with
trailing whitespace removed.


devyte:
Thanks!

PR #6666, opened 2019-10-25
Reapply fix for redundant FS size check

everslick:
My fix for a redundant FS size check was merged, (https://github.com/esp8266/Arduino/commit/bde1ce0e012321dff7d57900c920fb6314e7c6ca#diff-6690102111beb33824c743666a874d7c) but later reverted (https://github.com/esp8266/Arduino/commit/a389a995fb12459819e33970ec80695f1eaecc58#diff-6690102111beb33824c743666a874d7c) I guess by accident.


d-a-v:
This requirement is now ensured by these tests for all FSes:
https://github.com/esp8266/Arduino/blob/1e17dddd895883806c9bfd3dd7b7042aa813d94f/tools/boards.txt.py#L1207
https://github.com/esp8266/Arduino/blob/1e17dddd895883806c9bfd3dd7b7042aa813d94f/tools/boards.txt.py#L1248

However anyone could use custom definitions or ldscripts so this test is still valid.
The test will prevent the sketch to fail in case of issue, but it is silent (cc @devyte)
I don't think we can use a `static_assert()` because these constants are not known at compile time (but only at link time).
Should we emit a debug message when `start < end` ?
Should we add the same test for littleFS ?
Or should we make the test in some python script at around link time ?

everslick:
IMHO, that test should not exist in the first place. The API allows the FS to be placed in flash wherever the user wants. The _FS_START and _FS_END symbols are there to provide sensible default values. If the user does not override them AND start and end point to the same address, then `size` will be zero and the FS will fail to mount with a reasonable error message. I don't see the harm in not having the test at all.

PR #6680, opened 2019-10-28
more lwIP physical interfaces

### LINKED ISSUE #1263: Access DHCP options in STA mode?
Can I access the additional DHCP options that my DHCP server is supplying to client ?

Things like sys log server can be provided to the clients and I'd really like to be able to access them.

Thanks !

### LINKED ISSUE #3277: Ethernet and SSL/TLS?
I'm beginning to experiment with the Ethernet support in the library, and I am curious if there is any way to leverage the SSL/TLS work that was done for WiFi over the Ethernet client connections?

Right now, there does not seem to be any support in the examples ... but can this be done?

### LINKED ISSUE #5695: ESP8266 WiFi Client does not honor DHCP host-name option setting when using DHCP
### Basic Infos

- [ x I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.

#### Platform

- Hardware:			ModeMCU v3 LoLin Wemos.cc ESP8266
- Core Version:			ESP8266 WiFi 2.5.0-beta3

### Settings in IDE

- Module:  			Nodemcu
- Flash Mode:			other
- Flash Size: 		        4MB
- lwip Variant:                   v2
- Reset Method:		nodemcu
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:                115200

### Problem Description

DHCP ip address acquired by device as expected with WiFi configuration. However DHCP host-name option is not honored, the automatic host name is reported by the WiFi.hostname() call, not the name assigned in the DHCP reservation on DHCP server for said MAC address of device.

---

DHCP Reservation Example... Note the host-name option value of "Node01"
```
#--------------------
# NodeMCU...
#--------------------

host Node01 {
        hardware ethernet 68:C6:3A:D6:59:C7;
        fixed-address 192.168.1.50;
        option host-name "Node01";
}
```
---
```
#include <ESP8266WiFi.h>

CONST int   BAUDRATE = 115200;
CONST int   HALFSECOND = 500;
CONST char* DESCRIPTOR = "Dachshund Digital_optout_nomap";
CONST char* PASSPHRASE = "D@chshund Digit@1";

//
struct InformationStructure {
  //

  String    theMediaAccessControl;
  String    theNetwork;
  String    theKey;
  String    theId;
  IPAddress theIP;
  String    theName;
  int       theStrength;
};

void setup() {
  struct InformationStructure theInformation;

  Serial.begin(BAUDRATE);
  while (!Serial);
  //Serial.setDebugOutput(false);

  // WiFi Initialization...
  Serial.println("WiFi Initialization");

  WiFi.mode(WIFI_STA);
  WiFi.begin(DESCRIPTOR, PASSPHRASE);

  // WiFi Connection...
  Serial.print("WiFi Connection");
  
    // WiFi Connection Wait...
    delay(HALFSECOND);
    Serial.print(".");
  }              
  
  //
  theInformation.theMediaAccessControl = WiFi.macAddress();
  theInformation.theName = WiFi.hostname();
              theInformation.theKey = WiFi.psk();
              theInformation.theId = WiFi.BSSIDstr();
              theInformation.theStrength = WiFi.RSSI();
              theInformation.theIP = WiFi.localIP();
              theInformation.theNetwork = WiFi.SSID();
              
              Serial.printf("NodeMCU (ESP8266) Web Server<br>Media Access Control %s\nHost Name %s\nKey \'%s'\\nBasic Service Set Id %s\nSignal Strength %d (dBm)\nIP Address %s\nNetwork \'%s\'\n",
                theInformation.theMediaAccessControl.c_str(), theInformation.theName.c_str(), theInformation.theKey.c_str(), theInformation.theId.c_str(), theInformation.theStrength, theInformation.theIP.toString().c_str(), theInformation.theNetwork.c_str());
              
// Host name is auto generated, DHCP option value appears to be ignored.  Host name from DHCP should be 'Node01' not ESP_D659C7.

}

void loop() {

}
```

### Debug Messages
```
SDK:3.0.0-dev(c0f7b44)/Core:2.5.0-beta2=20499902/lwIP:STABLE-2_1_2_RELEASE/glue:1.0-4-gc434c6f/BearSSL:2398cc6
WiFi Initialization
scandone
WiFi Connectionscandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 2
cnt 

connected with xxx channel 2
dhcp client start...
ip:192.168.1.50,mask:255.255.255.0,gw:192.168.1.251
.NodeMCU (ESP8266) Web Server<br>Media Access Control 68:C6:3A:D6:59:C7
Host Name ESP_D659C7
Key 'xxx'\nBasic Service Set Id C8:D7:19:0F:4C:6F
Signal Strength -41 (dBm)
IP Address 192.168.1.50
Network 'Dachshund Digital_optout_nomap'
pm open,type:2 0
```

d-a-v:
19-10-28:

* Ethernet: W5100, W5500 and ENC28J60, thanks to quality drivers from @njh repositories
https://github.com/njh/EtherSia/tree/master/src/enc28j60.cpp
https://github.com/njh/W5100MacRaw
https://github.com/njh/W5500MacRaw
SPI needs to be initialized before using these drivers.

* PPP server with example

* DHCP server moved out from lwip2 directories is now a per-interface instanciable class
  (and more easily improvable)

* Every services / libraries should work with ethernet 
  ~~- mDNS is [a work in progress](https://github.com/Labor-Et-Ars/ESP8266-Arduino/tree/LEAmDNS2x)~~

* More on history in https://github.com/d-a-v/W5500lwIP

*edit*

Fixes #1263
Fixes #5695 
Fixes #3277


Misiu:
@d-a-v maybe this is a silly question, but will this enable us to use esp8266 without WIFI enabled and without any WIFI code? I'd like to use only the Ethernet connection. Unfortunately, when using WIFI without deep sleep the chip is getting hot. I hope that with disabled WIFI the chip will have a much cooler temperature. I think that wifi related code is taking much space and by not using it some space can be saved. Please correct me if I'm wrong.

devyte:
> without WIFI enabled

I  theory, yes. 
> without any WIFI code

No, wifi code and libs still get pulled in. This would be nice, it would free up a lot of mem (estimated at 16-20KB additional free heap), but getting there is very very far away. Building without wifi is a rather huge undertaking, because it means pulling apart several closed source libs to separate the wifi bits, and then redoing the bootup code and also redoing the build system to pull those bits in only when wifi is used. A lot of what I just said isn't known im detail because of the closed nature of Espressif's libs. 
The idea has been present for a while, but is currently shelved due to higher priorities. 

To everyone reading this, if anyone wants to help in that direction, please contact me in our gitter channel.

Jason2866:
There will be a Tasmota version for wired usage

d-a-v:
CI succeeded, sorry for the noise :]

xBlue0:
Are there still problems with this PR? I really need the support for PPP and I used this branch as a starting point, apart from some problems that i managed to solve this seems to be stable, so I was wondering why this is still not merged and if some help to move forward is needed.

d-a-v:
This will be merged after release 2.7.0 (on which we are actively working), and be released in 3.0.0.

Thanks for your interest, and your feedback.

Are you using PPP as client or server ?
What is connected on the other side ?

xBlue0:
Hi, thanks for your answer.
I'm using PPP protocol in order to connect the ESP8266 (client) to a micro placed on a sensor (server); it is a peculiar use case, since the micro is highly constrained and it uses a weird way to establish connection:

* esp8266 send 'CLIENT' string through serial
* micro answer 'CLIENTSERVER' string through serial

---> then PPP setup starts without any kind of authentication, something similar to 'communication cable between two computers' made by Microsoft.

At this point, the two parts communicate using TCP/UDP over PPP.

Your works are great btw, I got a lot of trubles going through official Espressif's ESP8266 RTOS SDK, when i found out something about PPP was made for the Arduino core I was very happy (this SDK offers also WebServer and WebSocket that are essential in this project) :smile:

Pablo2048:
Hi,
>     * esp8266 send 'CLIENT' string through serial
> 
>     * micro answer 'CLIENTSERVER' string through serial
> 
This looks like SLIP not an PPP to me...

d-a-v:
In both case, there is usually a clear/readable dialog between the two peers to agree on when starting speaking that protocol (like when we were using `expect` with ATDT hayes commands back in internet on PSTN days - maybe still used somewhere I'm not aware about that).

PPP is a SLIP evolution

(is there any french in the audience :)

xBlue0:
Hi,

> This looks like SLIP not an PPP to me...

I ensure I'm using PPP not SLIP, I'm not an expert btw so it's probable I don't deeply know these protocols.
In fact, since I have to be very quick in reading PPP packets from UART, how could I implement a mechanism based on interrupt instead of reading them periodically? Is this a long work?
It's kinda the most difficult part about the work I'm doing, since the rest of it is working with sockets&co, and I could really use an help about how to do it right :grinning: 

> (is there any french in the audience :)

I'm Italian btw :it:

libraries/lwIP_PPP/src/PPPServer.cpp
```diff
@@ -0,0 +1,191 @@
...
+    case PPPERR_DEVICE:            /* Invalid I/O device for PPP. */
+        ets_printf("ppp_link_status_cb: PPPERR_DEVICE\n");
+        break;
+
+    case PPPERR_ALLOC:             /* Unable to allocate resources. */
+        ets_printf("ppp_link_status_cb: PPPERR_ALLOC\n");
+        break;
+
+    case PPPERR_USER:              /* User interrupt. */
+        ets_printf("ppp_link_status_cb: PPPERR_USER\n");
```
xBlue0: It can be useful here to remove the netif in order to give the user the possibility to start connection again without errors, like this:
```
netif_remove(&static_cast<PPPServer*>(ctx)->_netif);
```

Same could be done for connect errors.

ssshake:
I'm pre-apologizing for adding noise to this PR. But I came across this because I'm looking for a way to host a PPP server on an esp8266 NodeMCU. Is this code leading me down the right path?

Recently I build a serial modem emulator for use on my old computers. I use this for connecting to BBS's. However I would like to extend this to be able to handle full on dial up networking. I currently use my modem to telnet to my linux machine and then from there start up PPPd. 

I came here looking for a way to host PPP on the ESP itself and not having to have a linux server to supply this.

d-a-v:
> host a PPP server on an esp8266 NodeMCU.

Yes, in the example included in this PR,
your linux PC is a PPP client, the ESP is the PPP server
(wifi/ethernet must be switched off on linux during the test)

https://github.com/d-a-v/Arduino/blob/ethernet/libraries/lwIP_PPP/examples/PPPServer/PPPServer.ino

d-a-v:
@xBlue0 
> how could I implement a mechanism based on interrupt instead of reading them periodically?

I think we can use scheduled functions to do that. It is not implemented yet.
Please open a new issue for such a feature request.

libraries/lwIP_PPP/src/PPPServer.cpp
xBlue0: I forgot that `netif` is pulled out from `ppp_netif(pcb)` so it is possible another way:
```
netif_remove(nif);
```

FIAV1:
Hi, sorry to make noise again in this PR. I wanted to know if there's any news about this PR since version 2.7 is ready to be released :smile: 
My current problem is that i can't use mDNS using this PR, the rest is working like a charm :+1: 
Thanks for the amazing work btw!

d-a-v:
mDNS will be fixed for all interfaces hopefully quite soon.
In the meantime please report here what and how you've been trying.

FIAV1:
Hi, sorry for late reply, i didn't receive any in my mail!

> mDNS will be fixed for all interfaces hopefully quite soon.
> In the meantime please report here what and how you've been trying.

I just tried to add mDNS like the basic example, but it doesn't work, and i didn't try anything else because:

> - mDNS not yet because [changes are coming](https://github.com/esp8266/Arduino/compare/master...LaborEtArs:LEAmDNS2?expand=1)

So i'm just waiting for it... in the meantime i forked this branch and the link layer for lwIP2 because i needed to change some parameters to PPP stack (like setting the PPP_MAXIDLEFLAG to 0), since (if i didn't get this wrong :laughing:) this pr is gonna take long before being merged.
Again, thanks a lot for the hard work!

d-a-v:
I hope it won't be that long.
mDNS PR is already here too.
Can you tell why or in which application PPP_MAXIDLEFLAG default value is not better than 0 ?

FIAV1:
Hope that too!
Btw, I needed to set PPP_MAXIDLEFLAG to 0 because the PPP stack implementation has been made from scratch for a micro of the factory for which I'm working for, so it has a custom way to communicate (like this flag, since the escape char is always used to understand the packet beginnig/end)

> mDNS PR is already here too.

You mean that it is now possible to use mDNS here too? I tried to use it a long ago, so maybe I lost some commits in the meantime 😅

d-a-v:
@davidk88 ([ref](https://github.com/esp8266/Arduino/issues/7454#issuecomment-658110021))

> I'm more interested on PPP client for communicationg with modem and sharing internet via ESP8266's AP.

PPP client is also available, there's no example yet in this PR.

cores/esp8266/lwIPDhcpServer-NonOS.cpp
```diff
@@ -0,0 +1,68 @@
...
+ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+// STARTS/STOPS DHCP SERVER ON WIFI AP INTERFACE
+// these functions must exists as-is with "C" interface,
+// nonos-sdk calls them at boot time and later
+
+#include <lwip/init.h> // LWIP_VERSION
+
+#if LWIP_VERSION_MAJOR != 1
```
Misiu, line 28: @d-a-v  the code for old lwip can be removed, ref: https://github.com/esp8266/Arduino/pull/7436

BTW: There will be only one LWIP in 3.0? Just asking to be aware of upcoming changes :)

d-a-v: Not as easy as code removal. Old lwIP include files are needed to compile the other side of the glue.
I'm moving these files into lwip2 repository.
Yes, lwIP1.4 will not be available in esp8266 Arduino core v3.

davidk88:
Files for PPP auth are missing. Is auth not supported?

d-a-v:
They are in tools/sdk/lwip2/builder/lwip-source/include/lwip/
Copy the needed files to tools/sdk/lwip2/include and give me the list so I can add them in the PR

cores/esp8266/lwIPDhcpServer-NonOS.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <lwip/netif.h>
+#include "lwIPDhcpServer.h"
+
+extern netif netif_git[2];
+
+// global DHCP instance for softAP interface
+DhcpServer dhcpSoftAP(&netif_git[SOFTAP_IF]);
```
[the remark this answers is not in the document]
d-a-v: Yes. Currently nonos-sdk makes a call to softAP at boot time.
This PR brings this management to the core out from lwip2.
After that we can try and see how the initial useless start could be faked-answered.

cores/esp8266/lwIPIntf.h
```diff
@@ -0,0 +1,28 @@
+
+#ifndef _LWIPINTF_H
+#define _LWIPINTF_H
```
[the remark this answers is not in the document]
d-a-v: `lwIPIntfDev` is the device, inheriting from raw device `RawDevice`, and `lwIPIntf`. That one contains common methods for any interface, including the arduino reordering helper because any interface can be a dhcp server (except STA because we want to). This class contains `hostname()` too.

FIAV1:
After a while, I had to upgrade my version of this branch in order to make OTA work (with signed bin and Updater class).
Now though, my PPP client isn't working anymore: it receives packets but the output function is not sending anything. Any idea?

d-a-v:
@FIAV1 can you list your changes that allowed you to use PPP-client ?
Let's push these changes in this PR so it is easier to reproduce and debug.

FIAV1:
@d-a-v I can put here the code for PPPClient, since if i push my code here, also binaries of lwip rebuild and some other changes will be published and this can lead to main code corruption.

PPPClient.cpp:
```cpp
#include <PPPClient.h>

using namespace std;
using namespace std::placeholders;

PPPClient::PPPClient(HardwareSerial *sio)
{
  _ppp = pppos_create(&_netif, PPPClient::output_cb_s, PPPClient::link_status_cb_s, this);

  if (_ppp)
  {
    _buffer = (uint8_t *)malloc(BUFFER_SIZE);
    ppp_set_default(_ppp);
    ppp_set_usepeerdns(_ppp, true);
  }
  else
  {
    _buffer = NULL;
    _ppp = NULL;
  }
}

PPPClient::~PPPClient()
{
  if (_ppp)
  {
    ppp_close(_ppp, 1);
    ppp_free(_ppp);
  }

  if (_buffer)
  {
    free(_buffer);
    _buffer = NULL;
  }
}

u32_t PPPClient::output_cb_s(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx)
{
  (void)pcb;

  int written = static_cast<PPPClient *>(ctx)->_sio->write(data, len);

  DEBUG_ESP_PPP_VERBOSE("written %s: %.*s", PPPClient::formatBytes(written).c_str(), written, (char *)data);
  if (data[0] == '~' && data[written - 1] == '~')
  {
    DEBUG_ESP_PPP_VERBOSE("packet written is well formed");
  }
  else
  {
    DEBUG_ESP_PPP_VERBOSE("packet written is bad formed");
  }

  return written;
}

void PPPClient::link_status_cb_s(ppp_pcb *pcb, int err_code, void *ctx)
{
  (void)pcb;

  PPPClient *PPPClient = static_cast<PPPClient *>(ctx);

  if (err_code == PPPERR_NONE)
  {
    PPPClient->_onStartCb(IPAddress(netif_ip4_addr(&PPPClient->_netif)), IPAddress(netif_ip4_gw(&PPPClient->_netif)));

    DEBUG_ESP_PPP("link status: %s", PPPClient->getLinkStatusString(err_code).c_str());
    DEBUG_ESP_PPP_VERBOSE("\tclient ip: %s", ip4addr_ntoa(netif_ip4_addr(&PPPClient->_netif)));
    DEBUG_ESP_PPP_VERBOSE("\tserver ip: %s", ip4addr_ntoa(netif_ip4_gw(&PPPClient->_netif)));
    DEBUG_ESP_PPP_VERBOSE("\tnetmask: %s", ip4addr_ntoa(netif_ip4_netmask(&PPPClient->_netif)));
    DEBUG_ESP_PPP_VERBOSE("\tdns1: %s", ipaddr_ntoa(dns_getserver(0)));
    DEBUG_ESP_PPP_VERBOSE("\tdns2: %s", ipaddr_ntoa(dns_getserver(1)));
  }
  else
  {
    PPPClient->_enabled = false;
    PPPClient->_onStopCb();

    DEBUG_ESP_PPP("link status: %s", PPPClient->getLinkStatusString(err_code).c_str());
  }
}

bool PPPClient::handlePackets()
{
  if (_enabled && _sio->available())
  {
    size_t read = _sio->readBytes(_buffer, BUFFER_SIZE);

    DEBUG_ESP_PPP_VERBOSE("read %s", PPPClient::formatBytes(read).c_str(), read);

    if (_buffer[0] == '~' && _buffer[read - 1] == '~')
    {
      DEBUG_ESP_PPP_VERBOSE("packet read is well formed");
    }
    else
    {
      DEBUG_ESP_PPP_VERBOSE("packet read is bad formed");
    }

    pppos_input(_ppp, _buffer, read);
  }
}

bool PPPClient::isRunning()
{
  if (_ppp)
  {
    return _ppp->phase != PPP_PHASE_DEAD;
  }

  return false;
}

bool PPPClient::begin()
{
  if (_ppp)
  {
    DEBUG_ESP_PPP("ppp over serial has been created");

    _enabled = true;

    bool scheduled = schedule_recurrent_function_us(bind(&PPPClient::handlePackets, this), 1000);

    if (scheduled)
    {
      DEBUG_ESP_PPP("packet handler scheduled");

      err_t error = ppp_connect(_ppp, 0);

      if (error == ERR_OK)
      {
        DEBUG_ESP_PPP("connect success: %s", PPPClient::getErrorString(error).c_str());
        return true;
      }
      else
      {
        _enabled = false;
        DEBUG_ESP_PPP("connect fail: %s", PPPClient::getErrorString(error).c_str());
      }
    }
    else
    {
      _enabled = false;
      DEBUG_ESP_PPP("packet handler not scheduled");
    }
  }
  else
  {
    DEBUG_ESP_PPP("ppp over serial has not been created");
  }

  return false;
}

bool PPPClient::end()
{
  u8_t nocarrier = 0;
  err_t error = ppp_close(_ppp, nocarrier);

  if (error == ERR_OK)
  {
    DEBUG_ESP_PPP("close success: %s", PPPClient::getErrorString(error).c_str());
    return true;
  }

  DEBUG_ESP_PPP("close fail: %s", PPPClient::getErrorString(error).c_str());
  return false;
}

// Utils

String PPPClient::formatBytes(size_t bytes)
{
  if (bytes < 1024)
  {
    return String(bytes) + "B";
  }
  else if (bytes < (1024 * 1024))
  {
    return String(bytes / 1024.0) + "KB";
  }
  else if (bytes < (1024 * 1024 * 1024))
  {
    return String(bytes / 1024.0 / 1024.0) + "MB";
  }

  return "N/D";
}

String PPPClient::getErrorString(err_t error)
{
  switch (error)
  {
  case ERR_OK:
    return "No error, everything OK";
  case ERR_MEM:
    return "Out of memory error";
  case ERR_BUF:
    return "Buffer error";
  case ERR_TIMEOUT:
    return "Timeout";
  case ERR_RTE:
    return "Routing problem";
  case ERR_INPROGRESS:
    return "Operation in progress";
  case ERR_VAL:
    return "Illegal value";
  case ERR_WOULDBLOCK:
    return "Operation would block";
  case ERR_USE:
    return "Address in use";
  case ERR_ALREADY:
    return "Already connecting";
  case ERR_ISCONN:
    return "Conn already established";
  case ERR_CONN:
    return "Not connected";
  case ERR_IF:
    return "Low-level netif error";
  case ERR_ABRT:
    return "Connection aborted";
  case ERR_RST:
    return "Connection reset";
  case ERR_CLSD:
    return "Connection closed";
  case ERR_ARG:
    return "Illegal argument";
  default:
    return "Unknown error";
  }
}

String PPPClient::getLinkStatusString(int err_code)
{
  switch (err_code)
  {
  case PPPERR_NONE:
    "No error";
    break;
  case PPPERR_PARAM:
    "Invalid parameter";
    break;
  case PPPERR_OPEN:
    "Unable to open PPP session";
    break;
  case PPPERR_DEVICE:
    "Invalid I/O device for PPP";
    break;
  case PPPERR_ALLOC:
    "Unable to allocate resources";
    break;
  case PPPERR_USER:
    "User interrupt";
    break;
  case PPPERR_CONNECT:
    "Connection lost";
    break;
  case PPPERR_AUTHFAIL:
    "Failed authentication challenge";
    break;
  case PPPERR_PROTOCOL:
    "Failed to meet protocol";
    break;
  case PPPERR_PEERDEAD:
    "Connection timeout";
    break;
  case PPPERR_IDLETIMEOUT:
    "Idle timeout";
    break;
  case PPPERR_CONNECTTIME:
    "Max connect time reached";
    break;
  case PPPERR_LOOPBACK:
    "Loopback detected";
    break;
  default:
    "Unknown error code " + String(err_code);
    break;
  }
}
```

PPPClient.h
```c
#ifndef __PPPCLIENT_H
#define __PPPCLIENT_H

#pragma once

#include <Arduino.h>
#include <IPAddress.h>
#include <lwip/netif.h>
#include <netif/ppp/ppp.h>
#include <netif/ppp/pppos.h>
#include <DebugPrintMacros.h>

#define BUFFER_SIZE 128

typedef std::function<void(IPAddress clientIP, IPAddress serverIP)> PppStartHandler;
typedef std::function<void()> PppStopHandler;

class PPPClient
{
private:
  bool _enabled;
  HardwareSerial *_sio;
  ppp_pcb *_ppp;
  netif _netif;
  uint8_t *_buffer;

  PppStartHandler _onStartCb;
  PppStopHandler _onStopCb;

  static u32_t output_cb_s(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx);
  static void link_status_cb_s(ppp_pcb *pcb, int err_code, void *ctx);
  bool handlePackets();
  void connect();
  void disconnect();

  // Utils
  String formatBytes(size_t bytes);
  String getErrorString(err_t error);
  String getLinkStatusString(int err_code);

public:
  PPPClient(HardwareSerial *sio);
  ~PPPClient();

  void onStart(PppStartHandler onStartCb)
  {
    _onStartCb = onStartCb;
  }

  void onStop(PppStopHandler onStopCb)
  {
    _onStopCb = onStopCb;
  }

  bool isRunning();

  bool begin();
  bool end();
};

#endif // __PPPCLIENT_H
```

PrintDebugMacros.h
```c
#ifndef __DEBUGPRINTMACROS_H
#define __DEBUGPRINTMACROS_H

#define DEBUG_LEVEL_BASIC 1
#define DEBUG_LEVEL_VERBOSE 2

#if defined(DEBUG_ESP_PORT) && !defined(DEBUG_DEFAULT_P)
#define DEBUG_DEFAULT_P(module, format, ...) DEBUG_ESP_PORT.printf_P(PSTR("\r\n" module " " format), ##__VA_ARGS__)
#endif

#ifndef DEBUG_DEFAULT_P
#define DEBUG_DEFAULT_P(...) \
  do                         \
  {                          \
    (void)0;                 \
  } while (false)
#endif

#if defined(DEBUG_CUSTOM_PPP) && !defined(PPP_CUSTOM_DEBUG)
#if DEBUG_CUSTOM_PPP == DEBUG_LEVEL_VERBOSE
#define PPP_CUSTOM_DEBUG_VERBOSE(format, ...) DEBUG_DEFAULT_P("[PPP]", format, ##__VA_ARGS__)
#endif
#define PPP_CUSTOM_DEBUG(format, ...) DEBUG_DEFAULT_P("[PPP]", format, ##__VA_ARGS__)
#endif

#ifndef DEBUG_CUSTOM_PPP
#define PPP_CUSTOM_DEBUG(...) \
  do                          \
  {                           \
    void(0);                  \
  } while (false)
#endif
#if !defined(DEBUG_CUSTOM_PPP) || defined(DEBUG_CUSTOM_PPP) && DEBUG_CUSTOM_PPP != DEBUG_LEVEL_VERBOSE
#define PPP_CUSTOM_DEBUG_VERBOSE(...) \
  do                                  \
  {                                   \
    void(0);                          \
  } while (false)
#endif

#endif // __DEBUGPRINTMACROS_H
```

FIAV1:
> @FIAV1 can you list your changes that allowed you to use PPP-client ?
> Let's push these changes in this PR so it is easier to reproduce and debug.

I deleted the code posted last time, since I found out it is full of bugs (sorry for that). Btw, PPP client is now working again. I just modified some buffers size and all worked correctly.

d-a-v:
@davidk88 @FIAV1 
Can you please tell what PPP files are missing / need to be imported for 
- PPP auth
- PPP client

FIAV1:
@d-a-v for ppp client no other files are needed; I don't use auth, so I don't know if other files are required for such functionality.

davidk88:
@d-a-v  Hello, back then I also got PPP client working without auth, so I didn't look which files are needed.

cores/esp8266/LwipIntfDev.h
```diff
@@ -0,0 +1,336 @@
...
+#include <lwip/netif.h>
+#include <lwip/etharp.h>
+#include <lwip/dhcp.h>
+#include <lwip/apps/sntp.h>
+
+#include <user_interface.h>	// wifi_get_macaddr()
+
+#include "SPI.h"
+#include "Schedule.h"
+#include "lwIPIntf.h"
```
jbaudoux, line 20: framework-arduinoespressif8266/cores/esp8266/LwipIntfDev.h:20:10: fatal error: LwIPIntf.h: No such file or directory

```suggestion
#include "LwipIntf.h"
```

d-a-v: Thanks, I will update this PR to latest master.

d-a-v: Fix pushed, thanks @jbaudoux

cores/esp8266/LwipIntfDev.h
```diff
@@ -0,0 +1,386 @@
...
+    IPAddress    gatewayIP() const
+    {
+        return IPAddress(ip4_addr_get_u32(ip_2_ip4(&_netif.gw)));
+    }
+
+    void setDefault();
+
+    bool connected()
+    {
+        return !!ip4_addr_get_u32(ip_2_ip4(&_netif.ip_addr));
```
jbaudoux, line 70: While testing, I saw this always returns true once it has connected once and acquired an IP address. So this does not really reflect the fact that the connection is active like you have with the WiFi.  Even if you unplug the cable, it still returns true. This prevents to re-route to WiFi when ethernet is down.
Can't we replace this by checking the socket status by calling getSn_SR() ?

infrafast: Hello, I observed the same, did you find a solution to that problem ?
In my case, this just return if the RJ45 was phyiscally plugged once even if you didn't get an IP address, it would return true.
Feedback appreciated

d-a-v: Can you open a new issue for a better tracking ?

d-a-v: This is fixed by #8784

Issue #6679, opened 2019-10-28
[Feature Req] Auto-detect SPIFFS/partition layout


TD-er:
For ESPEasy I have a lot of builds defined.
The number of builds can be more than halved by not having separate builds for 1M, 2M, 4M etc.
Lately I also added test builds for 4M2M layout (4M flash, 2M SPIFFS) and when rolled out to other build definitions the number of build environments will increase even more.

Also the chance of loading the wrong image on a node does increase with all these options.

Would it be possible to have the sketch detect the flash size and detect the location of the SPIFFS, instead of hard-coded in the build?

Then I can add a number of best practice offsets for different setups and chose the best one based on the found hardware when a SPIFFS has to be created (or LittleFS) and when there is one already present just use that one.

This will remove all the flash size related builds and make it just about selecting the processor + set of plugins you want.

If this is already possible, where can I find information on how to set it?

I guess the core libraries also need some fixed offsets for things like WiFi settings and the EPROM area.  So these should then also be auto-probed and offer some function call to query their locations.

d-a-v:
Very interesting

Flash size: `Esp.h`: `uint32_t EspClass::getFlashChipRealSize(void)`
Locations: the [logic in python boards.txt builder](https://github.com/esp8266/Arduino/blob/6e51ef0cc8a17706b8981999bfb64e1b039726ee/tools/boards.txt.py#L1178) can be duplicated
SPIFFS/LittleFS: a quick scan for FS signatures can be done in the possible locations computed by the above logic.

WiFi settings addresses is independently computed by nonos-sdk-fw's by a similar way. There's nothing to do here. The only symbols that we use and defined by flash chip size / menu configuration are put to `tools/sdk/ld/eagle.flash.*`:
```
PROVIDE ( _FS_start = 0x402BB000 );
PROVIDE ( _FS_end = 0x402FB000 );
PROVIDE ( _FS_page = 0x100 );
PROVIDE ( _FS_block = 0x1000 );
PROVIDE ( _EEPROM_start = 0x402fb000 );
```

TD-er:
Those WiFi settings addresses, are they computed at compile time, or run time?
If at run time, then we don't need to set the flash size at compile time for those.
If at compile time, then we still have to differentiate between flash sizes, but may be a bit more flexible in sketch/SPIFFS/LittleFS layout.

d-a-v:
[...]

Issue #2321, opened 2016-07-23
Auto size spiffs file system. Universal binary creation. 


mmiscool:
I would really like to have an option to produce a generic binary that can be flashed to any device and auto size the spiffs file system to the max size.

Enhancement request.

igrr:
If the device already has smaller SPIFFS partition flashed (e.g. someone was using a 1M binary on 4M flash), would you like the partition to be preserved? Or is it okay to just create new partition which fits actual flash size?

mmiscool:
I would like it to max out the size of the file system automatically.

igrr:
I.e. it has to overwrite older small partition, correct?

mmiscool:
Yes. Wipe out the old smaller one. That would be great.

mmiscool:
Somewhat related. Don't know if it is possible. I have run in to trouble compiling for 1 megabyte modules. The file system takes up half of the module and the firmware the other half. Unfortunately my firmware is now over the allowable size limit. Would be nice if this allowed for the size of spiffs system to shrink either tightly using space unused or reserved by the firmware or to size to the next step down for rigid size of the spiffs.

mmiscool:
Any word on this becoming a possibility?

devyte:
From what I can tell, this requires some rather deep and fundamental changes to code, ld scripts, and IDE menu.
Removing milestone for now.

d-a-v:
This will be fixed by #6690.
Pushing to v3.0.1.

PR #6690, opened 2019-10-29
flash-size agnostic builds

d-a-v:
(*edit 2021/12/01*)

This PR adds a new option in the flash size configuration options list called
`Mapping defined by Hardware and Sketch`.

This option is useful because:
- Users need not to care about flash configuration
  (If the option is selected. It is not currently the default)
- A binary can indifferently be flashed over 512k or 16MB flash chips, or esp8285 if `dout` is selected
  - FS and EEPROM location will automatically match legacy addresses as if flash size was correctly selected
  - Three configurations are available, [they reflects standard / current menus flash mapping](https://github.com/esp8266/Arduino/blob/6be7adfa27f326c90042ceb488097a408a97610c/tools/boards.txt.py#L1517-L1551)
    - FLASH_MAP_OTA_FS (default if `FLASH_MAP_SETUP_CONFIG()` is not used)
    - FLASH_MAP_MAX_FS
    - FLASH_MAP_NO_FS

    Sketch can use FLASH_MAP_SETUP_CONFIG(config) to [select a configuration](https://github.com/d-a-v/Arduino/blob/nosizeconf/cores/esp8266/FlashMap.h), for example: `FLASH_MAP_SETUP_CONFIG(FLASH_MAP_MAX_FS)`
   - Sketch can also define another configuration by providing a custom mapping structure.

- No more FS overwrite / reformatting when bad configuration is selected and flashed

- Binary releases for projects can see their number of builds reduced

When not enabled, this PR makes no change. and flash size and configuration needs to be correctly selected as before (with the same symbols).

Documentation update:
- [filesystem](https://github.com/d-a-v/Arduino/blob/nosizeconf/doc/filesystem.rst#flash-layout)
- [faq](https://github.com/d-a-v/Arduino/blob/nosizeconf/doc/faq/readme.rst#how-to-resolve-undefined-reference-to-flashinit-error-)

When the option is enabled, the magic word at physical address 0x0 in flash which contains "software" flash size is ignored; its reading is disabled and is replaced by reading the "chip vendor id" (`ESP.getFlashChipRealSize()`).

Fixes: #6679
Fixes: #2321
Closes: #6806 via #6813


d-a-v:
@TD-er for a start I ran the same binary DOUT lib-eeprom read/write sketch on esp8285(1M) and esp8266(4M) with good results.

TD-er:
So, you've already started with it :)
The earliest test I can do with it is tomorrow evening. (I also have 2M modules)
Have got to finish something else now and will be away most of tomorrow.

TD-er:
I have not yet tested it, but looked through the code.
The `flashInit` function is called from the core main function and does look at the flash size to pick a layout.
Can I later set it again to another layout at runtime?
Or is it possible to set the magic byte to define some alternate setting and then reboot to load the alternate setting?

The reason I'm asking is that I cannot support a single binary for 4M flash with either 1 MB SPIFFS (default for all nodes out there running my software) or 2 MB SPIFFS if it is only looking at the flash size.
So either I have to have my own detection algorithm to see where the SPIFFS is located and change at runtime, or only try once to find any existing 1M SPIFFS and if it isn't there switch it to the 4M2M option, write that to the magic byte and reboot.
If it still cannot find a valid SPIFFS, then format it as a 2M SPIFFS.,

N.B. the reason I have not tested it yet, is mainly because of the amount of work needed to get a PR ready within PlatformIO. I also just made a feature request for it: https://github.com/platformio/platformio-core/issues/3239
In the retrieved packages directory of PIO, the git config is not complete or correct.
I do see the pull requests on origin, but when I want to checkout based on a PR, I get an error like "missing delta bases"

d-a-v:
> The flashInit function is called from the core main function and does look at the flash size to pick a layout.

That is on purpose.
The parameter (what you call layout) is in fact an array of layouts indexed by the flash chip size.
There are three hardcoded layout sets, but you can use a self-defined one.

> Can I later set it again to another layout at runtime?

FSes and EEPROM should be closed before doing that, but yes I don't see any reason against that.

d-a-v:
As @earlephilhower said elsewhere, we cannot autodetect SPIFFS because there is no signature. I don't know about LittleFS.
In either case the information about location and size can anyway be written somewhere and allow to build the right descriptor.

TD-er:
For my own purpose, I could also use the EPROM part to store something with a very specific signature to help detect the used offset.
It does probably have 2 possible locations at most for any given flash size.

d-a-v:
[this file](https://github.com/esp8266/Arduino/pull/6690/files#diff-59b990f2f1cf275b3c87cd0d7b112d41) contains all the addresses computed by boards.txt.py.
I think the easier path is to use the example provided in OP then change FS addresses:
```
... (later)
close_all_FSes(); // or not
FS_start = myFS_start;
FS_stop = myFS_end;
SPIFFS.begin()...
```

cores/esp8266/FlashMap.h
```diff
@@ -0,0 +1,51 @@
...
+
+#define FLASH_MAP_MAX_FS \
+    { \
+        { 0x402fb000, 0x4027b000, 0x402fb000, 0x2000, 0x100, 1024 }, \
+        { 0x403fb000, 0x40300000, 0x403fa000, 0x2000, 0x100, 2048 }, \
+        { 0x405fb000, 0x40500000, 0x405fa000, 0x2000, 0x100, 4096 }, \
+        { 0x409fb000, 0x40300000, 0x409fa000, 0x2000, 0x100, 8192 }, \
+        { 0x411fb000, 0x40300000, 0x411fa000, 0x2000, 0x100, 16384 }, \
+        { 0x4027b000, 0x4025b000, 0x4027b000, 0x1000, 0x100, 512 }, \
+    }
```
TD-er, line 38: Can these be marked with some comments per line, to see what they actually define here?

tools/boards.txt.py
```diff
@@ -1406,44 +1420,93 @@ def all_flash_map ():
...
     f1m.update( flash_map(    1024,     160 ))
     f1m.update( flash_map(    1024,     192 ))
     f1m.update( flash_map(    1024,     256 ))
-    f1m.update( flash_map(    1024,     512 ))
-    f1m.update( flash_map(    1024))
+    f1m.update( flash_map(    1024,     512, 'MAX_FS' ))
+    f1m.update( flash_map(    1024,       0, 'NO_FS'  ))
 
-    f2m.update( flash_map(  2*1024,      64 ))
+    f2m.update( flash_map(  2*1024,      64, 'OTA_FS' ))
```
TD-er, line 1437: Personally, I would like to see a bit more FS with the OTA_FS for 2M flash.
256k still leaves roughly 850k of sketch + OTA.
But 64k makes almost no sense for a file system as you still need to have some blocks free (at least with SPIFFS) to make it work.

d-a-v: 256K for 2MB flash chips it is. It is a fair choice (leaving 1.75MB for sketch + gzipped OTA).
Please head to OP, which has now links to updated documentation.

cores/esp8266/FlashMap.h
d-a-v: comments are added

d-a-v:
@TD-er Did you make further tests with this PR ?

TD-er:
I finally came to building ESPEasy with this (sorry that it took so long)
I ran into these build issues:

```
C:\users\gijsn\.platformio\packages\framework-arduinoespressif8266\libraries\ESP8266HTTPUpdateServer\src/ESP8266HTTPUpdateServer-impl.h: In lambda function:
C:\users\gijsn\.platformio\packages\framework-arduinoespressif8266\libraries\ESP8266HTTPUpdateServer\src/ESP8266HTTPUpdateServer-impl.h:95:38: error: '_FS_end' was not declared in this scope; did you mean 'FS_end'?
   95 |           size_t fsSize = ((size_t) &_FS_end - (size_t) &_FS_start);
      |                                      ^~~~~~~
      |                                      FS_end
C:\users\gijsn\.platformio\packages\framework-arduinoespressif8266\libraries\ESP8266HTTPUpdateServer\src/ESP8266HTTPUpdateServer-impl.h:95:58: error: '_FS_start' was not declared in this scope; did you mean 'FS_start'?
   95 |           size_t fsSize = ((size_t) &_FS_end - (size_t) &_FS_start);
      |                                                          ^~~~~~~~~
      |                                                          FS_start

```

d-a-v:
Thanks for testing !
I must have missed this change. Update is pushed.

TD-er:
Thanks!
I was looking into the code and noticed you only can have a single set for each flash size.
I'm looking into `spiffs_probe` for some code wrapper for me to use 1M SPIFFS on 4M flash if there is a file system present, but if it doesn't have a FS, then it should use the 2M SPIFFS.
But it looks like `SPIFFS_USE_MAGIC && SPIFFS_USE_MAGIC_LENGTH` is not set, so I doubt it is even possible to probe an existing FS as there doesn't seem to be a magic entry present. Am I correct?

d-a-v:
> I doubt it is even possible to probe an existing FS as there doesn't seem to be a magic entry present. Am I correct?

This PR does not change anything about `SPIFFS_USE_MAGIC` and `SPIFFS_USE_MAGIC_LENGTH`. The default proposed sizes match the ones that are currently selectable by defines (or by arduino IDE menu entries). If you are able to autodetect the beginning of the FS and initialize it with the proper addresses, then you can use the NO_FS define.

Otherwise In the 4MB version, the MAX-FS conf has 3MB FS, and the OTA+FS conf has 2MB FS.
While user could also propose custom mapping parameters, would you need a `OTA_SMALLER_FS` that'd propose 1 more unused MB in the flashmap (that would leave 1MB FS with 4MB flash chips)?

TD-er:
I already changed the mapping for the 1M definition like this:
```c++
#ifdef AUTOFLASHSIZE
#define FLASH_MAP_OTA_FS_ESPEASY \
    { \
        { 0x402fb000, 0x402DB000, 0x402FB000, 0x1000, 0x100, 1024 }, \
        { 0x403fb000, 0x403c0000, 0x403fb000, 0x1000, 0x100, 2048 }, \
        { 0x405fb000, 0x40400000, 0x405fa000, 0x2000, 0x100, 4096 }, \
        { 0x409fb000, 0x40400000, 0x409fa000, 0x2000, 0x100, 8192 }, \
        { 0x411fb000, 0x40400000, 0x411fa000, 0x2000, 0x100, 16384 }, \
        { 0x4027b000, 0x40273000, 0x4027b000, 0x1000, 0x100, 512 }, \
    }

FLASHMAPCONFIG(FLASH_MAP_OTA_FS_ESPEASY)
#endif
```

But the problem is, the once default value for 4M flash was 1M SPIFFS and thus that's what is still used on a lot of nodes running ESPEasy out there.
But so far I have not been able to properly detect an existing file system, so that makes it hard to make a "1 build fits all" for existing users and be able to move to the more sensible 4M flash/2M SPIFFS layout.
What I meant with the SPIFFS_probe call is that it does search for a magic marker, but that function is not callable without setting other defines. Meaning the magic marker is currently not present in current builds out there so there is not possibility to check for it...
Only work-around could be to enable it for new builds and check for it on the 2M marker and if it is use that one or else default to the old behavior.

Another way could be to try and mount it and see if a specific file is present, but I am not entirely sure if it doesn't try to 'fix' incorrect file systems thus damaging one if I attempt to 'probe' the incorrect one.
Either way, it is going to be 'messy'.

d-a-v:
There is a mount parameter that prevents from formatting when mount fail.

For a try, I added for a try the SMALLER_FS option this way (showing only 4MB chips)
```
    #                      flash(KB) spiffs(KB) confname(C)
    f4m.update( flash_map(  4*1024,  2*1024, 'OTA_FS' ))
    f4m.update( flash_map(  4*1024,    1024, 'OTA_SFS' ))     <= OTA_SFS name added 
    f4m.update( flash_map(  4*1024,  3*1024, 'MAX_FS' ))
    f4m.update( flash_map(  4*1024,       0, 'NO_FS'  ))
```
it gives (autogenerated using the once default value)
```
#define FLASH_MAP_OTA_SFS \
    { \
        { 0x405fb000, 0x40500000, 0x405fa000, 0x2000, 0x100, 4096 }, \
    }
```
Values are a little different than yours.

But I think there's some better way.

Autodetecting FSes beginning in flash would be indeed a good idea but it is beyond this PR.
What you could do is to store the FS address/size in some private structure and reuse them at next reboot including after next flash.  When not found, the OTA_FS conf could be used to format the FS.

mcspr (approved):
As mentioned in the chats, still wip but it might be much nicer to edit it in the tree instead of amending the PR ad-infinitum

Only two sore points
- we still have flash layout as symbols & defines, I wonder if we should move closer to something 'iterable' so we see all of things at once as [begin, end] ranges. at least of the sake of the reader, not necessary to be closer to partition schemes of the updated sdks
- boards.txt.py python could do better :>

cores/esp8266/Updater.cpp
```diff
@@ -24,8 +24,7 @@ extern "C" {
     #include "user_interface.h"
 }
 
-extern "C" uint32_t _FS_start;
-extern "C" uint32_t _FS_end;
+#include <flash_hal.h>
```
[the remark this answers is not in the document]
d-a-v: This one needs to be reverted because
- mock version of `flash_hal.h` needs to be loaded first, to redefine `FS_start` and `FS_end`

This is only needed in sources files
- located in `cores/esp8266/`
- which make use of `FS_start` or `FS_end`

PR #6699, opened 2019-11-02
Make SPIFFS and LittleFS stay out of link when not needed

### LINKED ISSUE #6691: LittleFS library linked when not used

d-a-v:
- define two weak functions defaulting to no-op
- redefine them to do something useful when either spiffs or littlefs are used/linked

This comes after a discussion with @earlephilhower 
(who will be too busy for the next days to come to continue on this)

replaces and closes #6697
fixes #6691 

Tested with all 4 uncommented `ets_printf` from this PR,
calling the two weak functions in `setup()`, and calling or not `SPIFFS.begin()` / `LittleFS.begin()`.


TD-er:
I've looked at the code of this PR and I do think I get how it should work.
I'm not entirely sure about why it will not link the strong function implementations, but that's more lack of knowledge on my part I guess.

Do you already have any idea on how much smaller a SPIFFS-only (or LittleFS-only) build is with this fix?

d-a-v:
SPIFFS is 32KB and LittleFS 25KB as far as I remember.

The trick in this PR and WPS is that there are `.cpp.o` files always linked in which we put the weakref functions. Other `.cpp.o` are indeed compiled but may not be included because their symbols are not needed (no dependencies). Once a dependency is forced in a linker unit (ie. call to `someInstance.begin()`, all of its symbols come to the final binary, and weak functions are overridden at that time if their new implementation are brought with.

TD-er:
But why doesn't it try to link in the .o file if it does find the strong function implementation in there (and nothing else is needed)
Is it purely because the weak link already is resolved by something that is guaranteed to be already processed? How is that guaranteed?

d-a-v:
> But why doesn't it try to link in the .o file if it does find the strong function implementation in there (and nothing else is needed)

I don't know that for sure.

> Is it purely because the weak link already is resolved by something that is guaranteed to be already processed? How is that guaranteed?

What I can say is that `close_all_fs()` needs to be linked, and it brings the weak definitions with it.

devyte:
@TD-er it links either both, in which case the weak is discarded due to the presence of the strong, or just the weak, in which case the weak is used (and optimized out because it's an empty function).

TD-er:
> @TD-er it links either both, in which case the weak is discarded due to the presence of the strong, or just the weak, in which case the weak is used (and optimized out because it's an empty function).

That's something I know, but what I don't know is how it is guaranteed the weak one is found first before the strong one, when the strong one is not needed.
The strong one will also be compiled, but the other functions in the .o file will not be needed when not called. So either linking the strong one or the weak one will resolve the link requirements, so I can imagine it depends on what order the function will be found by the linker.
Or am I missing something here?

d-a-v:
When a strong function is present, the weak function is used when
- both weak and strong are in a library (`.a`)
- the weak function library is listed *before* the strong function library
- no other symbol from the strong object file is imported by linker

Example:
```
decls.h
===========

void func1 ();
void bring_strong_to_me();

main.cc
===========

#include "decls.h"

int main ()
{
#ifdef bringIt
    bring_strong_to_me();
#endif
    func1();
    
    return 0;
}
strong.cc
===========

#include <stdio.h>
#include "decls.h"

void bring_strong_to_me ()
{
}

void func1 ()
{
    printf("strong\n");
}
weak.cc
===========

#include <stdio.h>
#include "decls.h"

void func1 () __attribute__((weak));
void func1 ()
{
    printf("weak\n");
}
```

Run & results:
```
#!/bin/sh
set -e
CXX='g++ -Wall -Wextra -Werror'
for i in weak strong; do
    g++ -Wall -Wextra -Werror -c $i.cc -o $i.o
    ar rcs lib$i.a $i.o
done
set -x
$CXX           main.cc     strong.o weak.o -o func && ./func # -> strong
$CXX           main.cc     weak.o strong.o -o func && ./func # -> strong
$CXX           main.cc -L. -lstrong -lweak -o func && ./func # -> strong
$CXX           main.cc -L. -lweak -lstrong -o func && ./func # -> weak
$CXX -DbringIt main.cc -L. -lweak -lstrong -o func && ./func # -> strong
```

TD-er:
Thanks.
That's a very clear example.

d-a-v:
Have you seen a decrease in binary size with this PR ?
If not, maybe the linking order in PIO has to be examined.

TD-er:
Have not yet tried this PR, since I have not been near a compile PC today.

TD-er:
I did test it and it does save about 20k in build size (using SPIFFS)
Just updated one of my nodes and it works just fine.

cores/esp8266/spiffs_api.cpp
```diff
@@ -127,6 +127,12 @@ bool isSpiffsFilenameValid(const char* name)
 
 }; // namespace
 
+extern "C" void spiffs_weak_end(void)
```
devyte: Function name is a bit misleading: this function is not weak. 
I was going to suggest just using `spiffs_end()` and `littlefs_end()` across the board, but I find it does make sense to have the name reflect the weak symbol trick. 

What do you think of something like:
`spiffs_end_weaklink()`
?

libraries/LittleFS/src/LittleFS.cpp
```diff
@@ -181,6 +181,12 @@ int LittleFSImpl::lfs_flash_sync(const struct lfs_config *c) {
 
 }; // namespace
 
+extern "C" void littlefs_weak_end(void)
+{
+    //ets_printf("debug: not weak littlefs end\n");
+    LittleFS.end();
```
devyte: what happens when FS_NO_GLOBALS is defined? NO GLOBAL_INSTANCES? etc
what happens when FS_NO_GLOBALS etc  is defined and there is a manual instance of a FS?

cores/esp8266/spiffs_api.cpp
TD-er: It is for internal call only right?
Then it does indeed need some reference to the weak trick.

libraries/LittleFS/src/LittleFS.cpp
TD-er: I think it would make sense to have the call to `end()` placed in the destructor.

d-a-v: @devyte thanks. override is now only enabled with global instance is defined.
@TD-er littlefs already has a destructor, spiffs now calls `end()` on destructor

cores/esp8266/spiffs_api.cpp
d-a-v: Functions are renamed, indeed `weak` should not appear in their name.
A link to the snippet above is added in source code.

devyte:
@TD-er please retest. If ok, this can be merged now for 2.6.0.

TD-er:
> @TD-er please retest. If ok, this can be merged now for 2.6.0.

I will be back home Thursday evening so I cannot test it before.

Jason2866:
We dont use SPIFFS nor LittleFS in Tasmota

Jason2866:
A short test sketch with SPIFFS worked. Only SPIFFS linked
image

Jason2866:
**Time to release 2.6.0 :-)** Tasmota does work reliable with latest master from tonight.
We have best experience using SDK22x_191024 with Tasmota.
With SDK22x_190703 there are a few devices (not many!) which have wifi reconnects
This devices do work too when using SDK22x_191024!

d-a-v:
We are going to release very soon now, probably this week.
Thanks for the report on the Nonos SDK version.
I'm now wondering if default fw should be 1910 or 1907.

PR #6720, opened 2019-11-07
Add wait loop at the end of mode, refactor can_yield

### LINKED ISSUE #4372: wifi_set_opmode_current return true but failing?
#### Hardware
Hardware:			ESP-12?
Core Version:      	       staging

### Description

Let me know if this sounds familiar or like a known issue or something
I am having an issue with mode, I am setting mode and it returns true, but the mode does not actually change

I will try to produce a sketch, it is very odd, does anyone know how I can debug this further, from my understanding this sounds like an SDK bug (or memory bounds issue?), but maybe I am missing something or we are missing some crucial check before wifi_set_opmode_current().

I am (essentially) doing this

```
persistent(false);
ret = WiFi.mode(WIFI_AP_STA);
ret = WiFi.mode(WIFI_STA);
...
ret = WiFi.mode(WIFI_AP_STA);
ret TRUE
getMode() is WIFI_STA !!!!!!
```

which is essentially

```
wifi_set_opmode_current(3) == true
wifi_set_opmode_current(1) == true
wifi_set_opmode_current(3) == true
wifi_get_opmode = 1 <--- aahhhhhhhhhhhhhhh
// wifi_set_opmode_current() != wifi_get_opmode(), but returns true
```
lots of boom here (enableSTA, enableAP all fail catastrophically
since there is no sanity checks here, since they return true must be true, 
so set config struct blows up )

Could this be a memory issue, where is this memory ?

### Settings in IDE

Module:  			?Generic ESP8266 Module?
Flash Size: 		?4MB/1MB?
CPU Frequency:		?80Mhz?
Flash Mode:			?qio?
Flash Frequency:	?40Mhz?
Upload Using:		?OTA / SERIAL?
Reset Method:		?ck / nodemcu?

devyte:
Fixes #4372


devyte:
This is messed up, I can't retrigger the build. Closing.

PR #6722, opened 2019-11-07
EspSoftwareSerial release 5.4.0 with performance/error rate improvement in TX

dok-net:
This is the culmination of reviewing my IRQ PR with @devyte.


devyte:
@dok-net the check failures are due to build errors. Can you please fix?

PR #6724, opened 2019-11-07
FW: use NONOS-SDK branch 2.2.x from 2019-10-24 by default

d-a-v:
former one is available when using generic board configuration


Jason2866:
@devyte 100%.  Always good to have options to choose from.

d-a-v:
Did I understand the latest one by default for all boards ?

@Jason2866 are you ready for testing this within the next few hours/days ?

Jason2866:
Just installed SDK22x_191105 and compiled Tasmota with.
3 devices are running now with. Will post first results tomorrow (it is 11.42 pm here) 
image

d-a-v:
@Jason2866 Thanks !
It appears the 191105 version will be an option (but still available).

Jason2866:
4 devices: 13 hours uptime. Not a single wifi or mqtt reconnect

d-a-v:
@Jason2866 You will need to use `PIO_FRAMEWORK_ARDUINO_ESPRESSIF_SDK22x_191105` in your PIO configuration from now with current core-stage/git-version and next release (which is, I believe, happening before Monday).

Jason2866:
Thx, for the info. It is already there
image

ascillato:
I have also tested SDK22x_191105 and didn't have any issue so far.

ascillato:
> I have also tested SDK22x_191105 and didn't have any issue so far.

After a massive update of all my devices to core 2.6.1 and several days of testings, I found the following related to SDKs:

* SDK22x_190703
Uses 2k flash less than SDK22x_191105
No wifi/mqtt disconnections on devices with RSSI >35 (the same as core 2.3.0)

* SDK22x_191105
Uses 2k flash more than SDK22x_190703
No wifi/mqtt disconnections on RSSI >55
On RSSI=(35-55) it performs really bad

I tested taking out as many interfering factors as possible. Tested and replicated with different **wifi routers** (MikroTik, TP-Link and Linksys), with different **weather conditions** (dry, wet, wind, hot, cold, nice), different **ESP8266 boards** (itead sonoff basic, sonoff pow R2, sonoff dual, NodeMCU -with proper power source-, MagicHome, Esp01, between others), correct tool-chain and same firmware in all devices (to rule out compilation issues - old toolchain produces wifi issues on SDK22x_191105)

So, as a result, Core 2.6.1 SDK22x_190703 performs better than any other core/SDK and should be the default.

Issue #6714, opened 2019-11-05
Compile error on FS.h: missing time.h import


Niek:
See https://github.com/esp8266/Arduino/commit/72dd589599e0b7d2e208eafeeca01a83b2d4cb94#r35810441: that commit breaks compilation for me because a ``<time.h>`` include is missing:
```c++
[...]

PR #6730, opened 2019-11-08
Add missing time.h include, use relative include path on time.h includes

Niek:
Added missing time.h include. The path is relative (instead of simply using ``<time.h>``) so it doesn't conflict with Time.h of the Arduino Time library. See https://github.com/esp8266/Arduino/issues/6714#issuecomment-551554120 for details.

### DIFF
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -23,6 +23,7 @@
 
 #include <memory>
 #include <Arduino.h>
+#include <../include/time.h>
 
 class SDClass;
 
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -23,7 +23,7 @@
 
 #include <memory>
 #include <Arduino.h>
-#include <../include/time.h>
+#include <../include/time.h> // See issue #6714
 
 class SDClass;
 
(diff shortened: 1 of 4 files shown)


d-a-v:
> You want to change all time.h includes?

[Yes ! ](https://github.com/esp8266/Arduino/issues/6714#issuecomment-551751387)

Niek:
> > You want to change all time.h includes?
> 
> [Yes ! ](https://github.com/esp8266/Arduino/issues/6714#issuecomment-551751387)

Done, except for the examples - since those don't use external libraries anyways I don't think it makes sense to change those (CMIIW).

d-a-v:
> CMIIW

If you have this library, windows or macOS, and if you can compile&run these examples, then it's fine.
Is the initial include in `FS.h` necessary when the others are modified ?

Niek:
> If you have this library, windows or macOS, and if you can compile&run these examples, then it's fine.
That's the case, unless you specifically link it against the Time library (which for an example would be weird).

> Is the initial include in FS.h necessary when the others are modified ?
In my opinion FS.h should have a time.h include, because it calls time(). You can't really rely on the parent code to include the header file.

PR #6743, opened 2019-11-10
Puts to rest the confusion about the builtin leds for nodemcu boards

### LINKED ISSUE #4715: Wrong #define LED_BUILTIN for NodeMCU board
### Basic Infos
   ESP8266 Arduino Core 2.4.1 has a mistake (?) #define LED_BUILTIN
for NodeMCU board  to pin gpio16 , then when using BLINK example ,
it can't blink the LED_BUILTIN

https://github.com/esp8266/Arduino/blob/master/variants/nodemcu/pins_arduino.h#L35

It should be defined to pin gpio 2  ?

#### Platform

- NodeMCU 1.0 (ESP12F)
- Core version 2.4.1
- ArduinoIDE
- Windows10

### Settings in IDE

- NodeMCU
- 4MB
- lwip V2 & V1.4

devyte:
Given that most nodemcu boards are based on the ESP12 boards, the definition of the builtin led is now changed to 2.
In addition, for those nodemcu boards that have an additional LED on the board connected to gpio16, an additional define is added called LED_BUILTIN_AUX.
Fixes #4715

### DIFF
--- a/variants/nodemcu/pins_arduino.h
+++ b/variants/nodemcu/pins_arduino.h
@@ -32,7 +32,8 @@
 static const uint8_t SDA = PIN_WIRE_SDA;
 static const uint8_t SCL = PIN_WIRE_SCL;
 
-#define LED_BUILTIN 16
+#define LED_BUILTIN 2
+#define LED_BUILTIN_AUX 16
 
 static const uint8_t D0   = 16;
 static const uint8_t D1   = 5;


dsyleixa:
I have to disagree:
as stated, LED_BUILTIN for ESP-12E nodeMCU is GPIO 16 == D0, and this is true for all of my 6 ESP-12E Boards!
By GPIO 2 nothing happens, neither LED is flashing.

PR #6754, opened 2019-11-11
Add String::concat(char*, len) to allow non null-term strings

### LINKED ISSUE #5061: request to modify String::concat
unsigned char String::concat(const char *cstr, unsigned int length) {
    unsigned int newlen = len + length;
    if(!cstr)
        return 0;
    if(length == 0)
        return 1;
    if(!reserve(newlen))
        return 0;
    strcpy(buffer + len, cstr);
    len = newlen;
    return 1;
}

### Basic Infos

#### Platform

- Hardware:				[All ESPxx]
- Core Version:			[latest]
- Development Env:		[All]
- Operating System:		[All]

### Problem Description

The String (Wstring.cpp) class has no method for concat() from a non-null terminated string.  This is annoying, and inffecient, when processing payloads from MQTT messages, which are not-null terminated.

This existing Concat(cstr,length) method would work with minor modifications, but it is protected. I am proposing modifying it as shown below and making it public.

unsigned char String::concat(const char *cstr, unsigned int length) {
    unsigned int newlen = len + length;
    if(!cstr)
        return 0;
    if(length == 0)
        return 1;
    if(!reserve(newlen))
        return 0;

--   strcpy(buffer + len, cstr);   // Would fail if cstr is not NULL terminated
++  memcpy(buffer + len, cstr,length);
++ buffer[newlen] = 0; // insert NULL terminator
 
   len = newlen;
    return 1;
}

```cpp

#include <Arduino.h>
#include "WString.h"

void setup() {
   char bigBuff[1000];

   int messageSize= ReadMQTT_Payload(bigBuff);
   
    String someString;

// How you have to do it now.  concat called for each character
   string.reserve(messageSize)
   for (int i=0; i < messageSize; i++)
      someString.concat(bigBuff[i];

// How you could do it with proposal.  Single call to concat
   someString.concat(bigBuff,messageSize);
}

void loop() {

}

```

### Debug Messages

```
Debug messages go here
```

earlephilhower:
Fixes #5061

Adds a concat(const char *data, int len) method which allows arbitrary
sequences of data (including ones w/embedded \0s) to be appended to a
String.  May be useful for certain MQTT operations.

Adds sanity test for the feature to host suite

### DIFF
--- a/tests/host/core/test_string.cpp
+++ b/tests/host/core/test_string.cpp
@@ -135,13 +135,15 @@ TEST_CASE("String concantenation", "[core][String]")
     const char buff[] = "abcdefg";
     String n;
     n = "1234567890"; // Make it a SSO string, fill with non-0 data
-    n = "1"; // Overwrite [1] with 0, but leave old jump in SSO space still
+    n = "1"; // Overwrite [1] with 0, but leave old junk in SSO space still
     n.concat(buff, 3);
     REQUIRE(n == "1abc"); // Ensure the trailing 0 is always present even w/this funky concat
-    for (int i=0; i<20; i++) n.concat(buff, 1); // Add 20 'a's to go from SSO to normal string
+    for (int i=0; i<20; i++)
+        n.concat(buff, 1); // Add 20 'a's to go from SSO to normal string
     REQUIRE(n == "1abcaaaaaaaaaaaaaaaaaaaa");
     n = "";
-    for (int i=0; i<=5; i++) n.concat(buff, i);
+    for (int i=0; i<=5; i++)
+        n.concat(buff, i);
     REQUIRE(n == "aababcabcdabcde");
     n.concat(buff, 0); // And check no add'n
     REQUIRE(n == "aababcabcdabcde");
--- a/tests/host/core/test_string.cpp
+++ b/tests/host/core/test_string.cpp
@@ -131,6 +131,20 @@ TEST_CASE("String concantenation", "[core][String]")
     REQUIRE(str == "-100");
     str = String((long)-100, 10);
     REQUIRE(str == "-100");
+    // Non-zero-terminated array concatenation
+    const char buff[] = "abcdefg";
+    String n;
+    n = "1234567890"; // Make it a SSO string, fill with non-0 data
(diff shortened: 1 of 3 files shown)


tests/host/core/test_string.cpp
```diff
@@ -131,6 +131,20 @@ TEST_CASE("String concantenation", "[core][String]")
     REQUIRE(str == "-100");
     str = String((long)-100, 10);
     REQUIRE(str == "-100");
+    // Non-zero-terminated array concatenation
+    const char buff[] = "abcdefg";
+    String n;
+    n = "1234567890"; // Make it a SSO string, fill with non-0 data
+    n = "1"; // Overwrite [1] with 0, but leave old jump in SSO space still
```
devyte: junk?

tests/host/core/test_string.cpp
```diff
@@ -131,6 +131,20 @@ TEST_CASE("String concantenation", "[core][String]")
...
+    n.concat(buff, 3);
+    REQUIRE(n == "1abc"); // Ensure the trailing 0 is always present even w/this funky concat
+    for (int i=0; i<20; i++) n.concat(buff, 1); // Add 20 'a's to go from SSO to normal string
```
devyte: for body on new line pls

tests/host/core/test_string.cpp
```diff
@@ -131,6 +131,20 @@ TEST_CASE("String concantenation", "[core][String]")
...
+    REQUIRE(n == "1abcaaaaaaaaaaaaaaaaaaaa");
+    n = "";
+    for (int i=0; i<=5; i++) n.concat(buff, i);
```
devyte: for body on new line pls

tests/host/core/test_string.cpp
earlephilhower: Actually, no.  Idea is to fill SSO array[0...10] with non-0 values, to catch if we don't 0-terminate properly in the concat(char*, int) method.  Otherwise if there's a leftover 0 in the SSO array and we don't add one in concat() it may not actually be detected.

earlephilhower: I did see the typo while fixing the others and fixed it.  D'oh.  I'm still jet lagged.

PR #6757, opened 2019-11-11
Update to LittleFS 2.1.3

earlephilhower:


earlephilhower:
Pushed to 2.7.0  I assume 2.6.1 is only the immediate bugfix and don't want to delay things.  The changelog or LFS 2.1.3 doesn't seem to be a big deal (and host tests pass, anyway, in either case).

PR #6759, opened 2019-11-12
WString explicit converters to reduce Flash size

s-hadinger:
Tasmota uses a lot of `String` with constructs like below:

```
str->replace("aa", "bb);
str->replace(F("cc"), String(num));
str->startsWith(F("prefix"));
str->endsWith(F("suffix"));
```

Currently for any non `String` argument, the compiler silently instantiates a new `String`object and deletes it after the call. This generates more than 1k of not very useful code.

This PR adds explicit methods to deal with common arguments `const unsigned char *` and `const __FlashStringHelper *`. It does not change the fact that a `String` object is created then deleted, it just reduces redundant code.

With this change, we see a 1k code size reduction with Tasmota.

### DIFF
--- a/cores/esp8266/WString.h
+++ b/cores/esp8266/WString.h
@@ -202,8 +202,20 @@ class String {
         unsigned char equalsIgnoreCase(const String &s) const;
         unsigned char equalsConstantTime(const String &s) const;
         unsigned char startsWith(const String &prefix) const;
+        unsigned char startsWith(const char * prefix) {
+            this->startsWith(String(prefix));
+        }
+        unsigned char startsWith(const __FlashStringHelper * prefix) {
+            this->startsWith(String(prefix));
+        }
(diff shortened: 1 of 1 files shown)


cores/esp8266/WString.h
```diff
@@ -202,8 +202,20 @@ class String {
         unsigned char equalsIgnoreCase(const String &s) const;
         unsigned char equalsConstantTime(const String &s) const;
         unsigned char startsWith(const String &prefix) const;
+        unsigned char startsWith(const char * prefix) {
```
d-a-v: `const`s are missing before '`{`'

s-hadinger:
@d-a-v Indeed, I added the missing `const` for startsWith() and endsWith(). As you mentioned, the added methods should not be inlined, that's the point.

Jason2866:
Could you add to milestone 2.6.1 or better merge before? :-)

PR #6762, opened 2019-11-12
Add missing "LittleFS.begin()" in LittleFS_Timestamp example.

mbm60:
Add missing first LittleFS.begin() call.


libraries/LittleFS/examples/LittleFS_Timestamp/LittleFS_Timestamp.ino
```diff
@@ -152,21 +152,25 @@ void setup() {
   getLocalTime(&tmstruct, 5000);
   Serial.printf("\nNow is : %d-%02d-%02d %02d:%02d:%02d\n", (tmstruct.tm_year) + 1900, (tmstruct.tm_mon) + 1, tmstruct.tm_mday, tmstruct.tm_hour, tmstruct.tm_min, tmstruct.tm_sec);
   Serial.println("");
-
-  Serial.printf("Formatting LittleFS filesystem\n");
+  Serial.println("Mount LittleFS");
+  if (!LittleFS.begin()) {
```
earlephilhower: LittleFS.begin() should be between lines 161-162.  Beginning and then formatting doesn't make sense, as @devyte mentioned.

There is a bug, though, and that begin should be there at 161-162 after the format() to actually mount things.

mbm60: Yes, you are right.

PR #6765, opened 2019-11-13
a single esptool command for erasing and flashing

### LINKED ISSUE #6755: Flashing fails with Arduino IDE 1.8.9 / Linux Mint
### Basic Infos

#### Platform

- Hardware:					[ESP-12]
- Core Version:			2.6.0
- Development Env:		[Arduino IDE|Platformio|Make|other]
- Operating System:		[Windows|Ubuntu|MacOS]

### Settings in IDE

- Module:  			 Wemos D1 mini  
- Flash Mode:			serial
- Flash Size: 		4MB
- lwip Variant:  v2
- Reset Method:		 
- Flash Frequency:	[40Mhz]
- CPU Frequency:		80
- Upload Using:		[OTA|SERIAL]
- Upload Speed:   57600 (serial upload only)

### Problem Description
flashing fails with serial cable for core 2.6.0 toolchain, while it is working for 2.4.2.

This is core 2.4.2 flashing output
```
Uploading 680992 bytes from /tmp/arduino_build_270612/ESPEasy.ino.bin to flash at 0x00000000
	erasing flash
	size: 0a6420 address: 000000
	first_sector_index: 0
	total_sector_count: 167
	head_sector_count: 16
	adjusted_sector_count: 151
	erase_size: 097000
	espcomm_send_command: sending command header
	espcomm_send_command: sending command payload
	setting timeout 15000
	setting timeout 1000
	espcomm_send_command: receiving 2 bytes of data
	writing flash
................................................................................ [ 12% ]
................................................................................ [ 24% ]
................................................................................ [ 36% ]
................................................................................ [ 48% ]
................................................................................ [ 60% ]
................................................................................ [ 72% ]
................................................................................ [ 84% ]
................................................................................ [ 96% ]
..........................                                                       [ 100% ]
starting app without reboot
	espcomm_send_command: sending command header
	espcomm_send_command: sending command payload
	espcomm_send_command: receiving 2 bytes of data
closing bootloader

```

After successful flash do: 
- go to board manager, scroll downn (twice) 
- pick core 2.6.0, hit install, 
- restart IDE
- power cycle Wemos D1
- upload (fails), output is: 

```
Der Sketch verwendet 714452 Bytes (68%) des Programmspeicherplatzes. Das Maximum sind 1044464 Bytes.
Globale Variablen verwenden 43224 Bytes (52%) des dynamischen Speichers, 38696 Bytes für lokale Variablen verbleiben. Das Maximum sind 81920 Bytes.
/home/john/.arduino15/packages/esp8266/tools/python3/3.7.2-post1/python3 /home/john/.arduino15/packages/esp8266/hardware/esp8266/2.6.0/tools/upload.py --chip esp8266 --port /dev/ttyUSB0 --baud 57600 flash_id --end --chip esp8266 --port /dev/ttyUSB0 --baud 57600 --before default_reset --after hard_reset write_flash 0x0 /tmp/arduino_build_906128/ESPEasy.ino.bin --end 
esptool.py v2.7
Serial port /dev/ttyUSB0
Connecting....
Chip is ESP8266EX
Features: WiFi
Crystal is 26MHz
MAC: 5c:cf:7f:39:b7:b0
Uploading stub...
Running stub...
Stub running...
Manufacturer: e0
Device: 4016
Detected flash size: 4MB
Hard resetting via RTS pin...
esptool.py v2.7
Serial port /dev/ttyUSB0
Connecting........_____....._____....._____....._____....._____....._____....._____
  File "/home/john/.arduino15/packages/esp8266/hardware/esp8266/2.6.0/tools/upload.py", line 25, in <module>
    esptool.main(fakeargs)
  File "/home/john/.arduino15/packages/esp8266/hardware/esp8266/2.6.0/tools/esptool/esptool.py", line 2841, in main
    esp.connect(args.before)
  File "/home/john/.arduino15/packages/esp8266/hardware/esp8266/2.6.0/tools/esptool/esptool.py", line 483, in connect
    raise FatalError('Failed to connect to %s: %s' % (self.CHIP_NAME, last_error))
esptool.FatalError: Failed to connect to ESP8266: Timed out waiting for packet header
esptool.FatalError: Failed to connect to ESP8266: Timed out waiting for packet header

```

- additional note: during upload attempt the internal LED flashes occasionally. 
- timeout appears much faster / LED flashes faster when high upload speeds are selected.

d-a-v:
Per https://github.com/esp8266/Arduino/issues/6755#issuecomment-553208688
(thanks @projectgus for the hints)

esptool:
* a single esptool command for erasing (all or a region) and flashing
* update to esptool-2.8

fixes #6755


d-a-v:
@ChazTuna @janzieg  @s0170071 @f5soh 
Can you please try this pull request
(I also updated the [alpha release channel](https://d-a-v.github.io) with it)

f5soh:
> Can you please try this pull request

esptool v2.8 looks good, tested basic programmer without dtr and Nodemcu v0.9 : full erase, partial erase and only sketch

A little improvement will be output scrolling while writing.

```
esptool.py cmdline: ['--chip', 'esp8266', '--port', '/dev/ttyUSB0', '--baud', '115200', '--before', 'no_reset', '--after', 'soft_reset', 'write_flash', '0x0', '/tmp/arduino_build_628729/Wifi_Angle_meter_mma8452_esp01.ino.bin']
esptool.py v2.8
Serial port /dev/ttyUSB0
Connecting....
Chip is ESP8266EX
Features: WiFi
Crystal is 26MHz
MAC: 84:f3:eb:b4:e1:72
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Auto-detected Flash size: 1MB
Compressed 405648 bytes to 265513...

Writing at 0x00000000... (5 %)
Writing at 0x00004000... (11 %)
Writing at 0x00008000... (17 %)
Writing at 0x0000c000... (23 %)
Writing at 0x00010000... (29 %)
Writing at 0x00014000... (35 %)
Writing at 0x00018000... (41 %)
Writing at 0x0001c000... (47 %)
Writing at 0x00020000... (52 %)
Writing at 0x00024000... (58 %)
Writing at 0x00028000... (64 %)
Writing at 0x0002c000... (70 %)
Writing at 0x00030000... (76 %)
Writing at 0x00034000... (82 %)
Writing at 0x00038000... (88 %)
Writing at 0x0003c000... (94 %)
Writing at 0x00040000... (100 %)
Wrote 405648 bytes (265513 compressed) at 0x00000000 in 33.2 seconds (effective 97.8 kbit/s)...
Hash of data verified.

Leaving...
Soft resetting...
```

d-a-v:
> A little improvement will be output scrolling while writing.

That, I thought it was only an effect on my setup (linux, xcfe) but it appears to be true even on windows.
I think that is the way the arduino IDE console interprets the `\r` character

tools/upload.py
```diff
@@ -16,20 +17,67 @@
     sys.path.insert(0, toolspath + "/esptool") # Add esptool dir to search path
     import esptool # If this fails, we can't continue and will bomb below
 except:
-    sys.stderr.write("Error in command line, need pyserial path as 1st arg and esptool path as 2nd.\n")
+    sys.stderr.write("pyserial or esptool directories not found next to this upload.py tool.\n")
```
earlephilhower: Good catch on the error message.

tools/upload.py
```diff
@@ -16,20 +17,67 @@
...
+    if thisarg == "erase_flash":
+        write_option = '--erase-all'
+        thisarg = ''
+
+    if thisarg == 'erase_region':
+        erase_addr = sys.argv.pop(0)
+        erase_len = sys.argv.pop(0)
+        thisarg = ''
+
+    if thisarg == 'ignore_this_arg':
```
earlephilhower: I believe you can simply pass an empty value (0-length argv[] value) in boards.txt.py instead of 'ignore_this_arg' and be a little cleaner by removing this if block.  Your len(thisarg) will be false in this case and you'll be fine.  Or, depending on how Arduino pastes things together, the 0-len arg might actually be dropped completely when calling upload.py.

tools/upload.py
```diff
@@ -16,20 +17,67 @@
...
+if len(write_option):
+    cmdline = cmdline + [write_option]
+cmdline = cmdline + ['0x0', binary]
+
+erase_file = ''
+if len(erase_addr):
+    # generate temporary empty (0xff) file
+    eraser = tempfile.mkstemp()
+    erase_file = eraser[1]
+    print(eraser)
```
earlephilhower: Debug info left over

tools/upload.py
```diff
@@ -16,20 +17,67 @@
...
+    # fill it
+    data = bytearray(512)
+    for i in range(512):
+        data[i] = 0xff
+    for i in range(int((int(erase_len, 0) + 511)/512)):
+        os.write(eraser[0], data)
+    os.close(eraser[0])
```
earlephilhower: Jumped through a lot of hoops here to save a few bytes.  Why not
````
data = bytearray(erase_len)
for i in range(erase_len):
    data[i] = 0xff
````
?

File: tools/upload.py [hunk 5]
```diff
@@ -16,20 +17,67 @@
...
+    cmdline = cmdline + [ erase_addr, erase_file ]
+
+print('esptool.py cmdline:', cmdline)
```
earlephilhower: Ditto on debug info

earlephilhower:
> 
> 
> > A little improvement will be output scrolling while writing.
> 
> That, I thought it was only an effect on my setup (linux, xcfe) but it appears to be true even on windows.
> I think that is the way the arduino IDE console interprets the `\r` character

I think it's all on the IDE here (as in we can't do anything on our end), but maybe if we disabled any buffering on stdout/stderr in the Python before calling esptool it might do something?

Adding `-u` after the python executable to the command line in platform.txt would do that.  Whether it would affect esptool or the IDE remains to be seen.

tools/upload.py
mcspr: Or maybe `[0xff] * 512`, if the goal is to write in chunks? No loops required.

earlephilhower:
Just tested it, and `-u` has no effect on the IDE.  So the scrolling issue isn't something we can adjust.

tools/upload.py
earlephilhower: Ah, yes.  Python has as many ways of expressing an algorithm as PERL. :)

earlephilhower:
````
eraser = tempfile.mkstemp()
erase_file = eraser[1]
os.write(eraser[0], bytearray([255] * erase_len))
os.close(eraser[0])
````

jandrassy:
is the upload.py still necessary? can't esptool command line be directly configured in platform.txt?

earlephilhower:
Not the way it's being done here.  We simulate the erase_region bit by writing 0xffs (i.e. erased) to the block, and need to generate a file to make that happen (hence the Python code above)

jandrassy:
> Not the way it's being done here. We simulate the erase_region bit by writing 0xffs (i.e. erased) to the block, and need to generate a file to make that happen (hence the Python code above)

Espressif recommends blank.bin

f5soh:
> I think it's all on the IDE here (as in we can't do anything on our end), but maybe if we disabled any buffering on stdout/stderr in the Python before calling esptool it might do something?

Previous esptool.exe was using [\n](https://github.com/igrr/esptool-ck/blob/ee33240587d1ef8768a0edfb415b9d13a3be00c5/espcomm/espcomm.c#L439) instead of \r

tools/upload.py
d-a-v: Thanks !

d-a-v:
@jandrassy our temp file if filled with 0xff just like espressif's blank.bin is.
Thanks to python we don't need to store it.

earlephilhower:
> 
> 
> > I think it's all on the IDE here (as in we can't do anything on our end), but maybe if we disabled any buffering on stdout/stderr in the Python before calling esptool it might do something?
> 
> Previous esptool.exe was using [\n](https://github.com/igrr/esptool-ck/blob/ee33240587d1ef8768a0edfb415b9d13a3be00c5/espcomm/espcomm.c#L439) instead of \r

Ah, it might make sense to put a PR in esptool.py repo to check if `isatty()` and if not (the case when run from the IDE) use normal endlines `\n`...

d-a-v:
It will be easier when/if we have another `write_flash` option like `--erase_region` along with `--erase_flash`

ChazTuna:
> @ChazTuna @janzieg  @s0170071 @f5soh
> Can you please try this pull request
> (I also updated the alpha release channel with it)

New board version 2.6.1 with fixes now works.. Thank you :)

PR #6766, opened 2019-11-13
minor: examples: reorder functions for declarations prior to call

d-a-v:
This is harmless for arduino, but it helps with emulation on host or PIO
[ref](https://gitter.im/esp8266/Arduino?at=5dcbecd9ea7d147cb36faee1)

### DIFF
--- a/libraries/ESP8266WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
+++ b/libraries/ESP8266WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
@@ -95,8 +95,9 @@ void handleNotFound() {
 }
 
 void drawGraph() {
-  String out = "";
-  char temp[100];
+  String out;
+  out.reserve(2600);
+  char temp[70];
   out += "<svg xmlns=\"http://www.w3.org/2000/svg\" version=\"1.1\" width=\"400\" height=\"150\">\n";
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
```diff
@@ -94,6 +94,24 @@ void handleNotFound() {
   digitalWrite(led, 0);
 }
 
+void drawGraph() {
+  String out = "";
```
devyte: no empty string assign

devyte: also, this series of concats is very bad practice. Let's add a call to reserve(), i think 2550 should be enough to contain everything (approximate, I eyeballed it).

libraries/ESP8266WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
```diff
@@ -94,6 +94,24 @@ void handleNotFound() {
...
+  char temp[100];
```
devyte: This is high, I count 63bytes worst case. 70 should be enough.

PR #5252, opened 2018-10-16
fix parseArgument #2

d-a-v:
#5230's twin brother using functors on @devyte's proposal
less available flash (-164B), more available ram (+28B)

PR #6768, opened 2019-11-13
webserver: do not count "plain" in argument list

### LINKED ISSUE #6435: server.send send unexpected argument migrating from 2.4.x to 2.5.2
### Basic Infos

#### Platform

- Hardware:			Amica NudeMCU
- Core Version:			2.5.2 downloaded by Arduino board manager
- Operating System:		Windows

### Settings in IDE

attached picture
setting-

### Problem Description

I have updated the Arduino IDE to 1.8.9 and the core to 2.5.2 using the board manager ( not the git).
Compiling my project that was working and running very stable with the core 2.4.x now I'm facing some issue related to the number of argument returned by the http communication.
In particular I see an extra argument named **plain** received by the AP, that was not present and transferred before.
So expected number of arguments (correct in the last compile with 2.4.x) in this example was 3, instead with the new core I get 4.
extraarg

And this is making me crazy because I have many pages and not sure what is happening.

Maybe it is a stupid error of mine but having it in conjunction to the switch to this new core, I try to write you for a check.
Thanks!

```cpp



#include <ESP8266WebServer.h>
#include <ESP8266WiFi.h>

#define BLUELED 02 

WiFiClient client;
ESP8266WebServer server(80);

void setup() 
{

	/*****************************/
	Serial.begin(115200);
	Serial.setDebugOutput(true);
	delay(10);

	Serial.println("Start! ....");

	Serial.println(F("Access point mode selected"));
	WiFi.mode(WIFI_AP);
	WiFi.softAPConfig(IPAddress(192,168,1,1),IPAddress(192,168,1,1),IPAddress(255,255,255,0));
	WiFi.softAP("MyAPN","12345678");
	Serial.println(F("Access point mode started"));
	IPAddress myIP = WiFi.softAPIP();
	Serial.print(F("AP IP address: "));
	Serial.print(myIP);
	Serial.println(F(" - PWD:12345678"));

	server.on("/", handleRoot);
	server.on("/setting",handleSetting);
	// Start the server
	server.begin(); //server.begin(sysSetting.httpPort);   //
}


void loop() 
{
	server.handleClient();
	digitalWrite(BLUELED,(digitalRead(BLUELED)?LOW:HIGH));
	delay(100);
	return;
}



/******************************************************************************
* Function: debugArgum
*
* Print to the serial the received arguments of the client/server http comm
*
* Parameters:                
*             header and argum set to 'y' means print to serial            
* Return:                 
*             none            
******************************************************************************/
void debugHeaderArgum(char header, char argu)
{
  String argum = "";
  int cnt;
  if(header =='y')
  {
    for (cnt = 0; cnt < server.headers(); cnt++) 
    {
      argum = "HeadArg: " +(String)cnt + " - " + server.headerName(cnt) + " - " + server.header(cnt);
      Serial.println(argum);
    }
  }
  if(argu =='y')
  {
    for (cnt = 0; cnt < server.args(); cnt++) 
    {
      argum = "Arg: " +(String)cnt + " - " + server.argName(cnt) + " - " + server.arg(cnt);
      Serial.println(argum);
    } 
  }
}

void handleRoot() 
{
	String argum = "";
	int cnt = 0;

	debugHeaderArgum('n','n'); //show header and argument

	Serial.println(F("APN - param page"));
	argum = F("HTTP/1.1 301 OK\r\nLocation: /setting\r\nCache-Control: no-cache\r\n\r\n");
	server.sendContent(argum);
	return;
}

/******************************************************************************
* Function: moduleInfo
*
* general module info usable in serial port and http.
*
* Parameters:                
*             httpmode: is true the {t] will be replaced with <DIV> and {-t} with </DIV> otherwise with \r\n             
* Return:                 
*             string containing the info            
******************************************************************************/
String moduleInfo(bool httpMode)
{
	byte mac[6];
	char tmpBuff[100];
	String retString = "\r\n{t}xxx Module: " +(String)ESP.getChipId() + "{-t}\r\n{t}Module ID: 10 - test {-t}\r\n";
	
    WiFi.macAddress(mac);
	sprintf(tmpBuff,"{t}MAC Address: %02X:%02X:%02X:%02X:%02X:%02X{-t}\r\n",mac[5],mac[4],mac[3],mac[2],mac[1],mac[0]);
	retString += (String)tmpBuff + "{t}Firmware version: x.x {-t}\r\n";
   
    // print the received signal strength:
    long rssi = WiFi.RSSI();
    retString += "{t}Signal strength (RSSI): " + (String)rssi + "dBm{-t}\r\n";
    
    uint32_t realSize = ESP.getFlashChipRealSize();
    uint32_t ideSize = ESP.getFlashChipSize();
    FlashMode_t ideMode = ESP.getFlashChipMode();
    sprintf(tmpBuff,"{t}Flash: Real size: %u - IDE Size: %u - IDE speed: %u{-t}\r\n", realSize, ideSize, ESP.getFlashChipSpeed());
	retString += (String)tmpBuff ;
    sprintf(tmpBuff,"{t}Flash IDE mode:  %s{-t}\r\n", (ideMode == FM_QIO ? "QIO" : ideMode == FM_QOUT ? "QOUT" : ideMode == FM_DIO ? "DIO" : ideMode == FM_DOUT ? "DOUT" : "UNKNOWN"));
	retString += (String)tmpBuff ;
	if(httpMode)
	{
		retString.replace("{t}","<DIV>");
		retString.replace("{-t}","</DIV>");
	}
	else
	{
		retString.replace("{t}","");
		retString.replace("{-t}","");
	}
	return retString;	
}



/******************************************************************************
* Function: handleSetting
*
* Handle the connection to the configuration page used in APN and normal operation
*
* Parameters:                
*             none            
* Return:                 
*             none            
******************************************************************************/
void handleSetting()
{
	
	String decodeDst = "";
	server.sendHeader("Cache-Control", "no-cache, no-store, must-revalidate");
	server.sendHeader("Pragma", "no-cache");
	server.sendHeader("Expires", "-1");
	server.sendHeader("Host", "192.168.1.1");
	server.setContentLength(CONTENT_LENGTH_UNKNOWN);
	
	String webApn = F("<!DOCTYPE HTML>\r\n<html>\r\n<head>\r\n<title>Dommy</title>\r\n");                      
	webApn += F("<style type=\"text/css\">\r\n body {background: white; color: black; text-align: center} ");
	webApn += F(" form { color: blue; text-align: center}\r\n div {color: black; text-align: left}\r\n");
	webApn += F(" h1 { color: green; font: 36px Helvetica, Arial, sans-serif;}\r\n</style>\r\n</head>\r\n");

	debugHeaderArgum('n','y');
	
	if(server.args()== 3)
	{

		Serial.println(server.arg("ARG1"));

		Serial.println(server.arg("ARG2"));

		Serial.println(server.arg("ARG3"));

		ESP.wdtFeed(); // to ensure the wdt doesn't trigger
		webApn += F("<body><DIV></DIV><DIV>Module updated.</DIV>\r\n<DIV>It is restarting.</DIV>");
		webApn += "<DIV> <a href='http://www.xxxxxxx.yyy:8080'> Return to the main page </a> </DIV></body>\r\n</html>\r\n"; 
		//server.send(200, "text/html",webApn);
		server.send(200, "text/html", ""); 
		server.sendContent(webApn);
		server.sendContent("");
		delay(500);
		ESP.restart();
		return;
	}
	else if(server.args()> 3)
	{
		Serial.println("Wrong arg number!!");
	}
 // String header = "HTTP/1.1 200 OK\r\nContent-Type: text/html Cache-Control: no-cache\r\n\r\n";
 //   server.sendContent(header);
    

  webApn += F("<body>\r\n<form name='Setting' method='POST'>\r\n");      
  webApn += F("<table width='60%' border='0' cellspacing='5' cellpadding='5'>\r\n");
  webApn += F("<tr><td width='10%' valign='center'><h1>xxx Module Config</h1></td></tr>\r\n");
 
  webApn += "<tr><td width='10%' valign='center'>" + moduleInfo(true) + "</td></tr></table>\r\n";
  webApn += F("<table width='60%' border='0' cellspacing='5' cellpadding='5'>\r\n"); 
  webApn += F("<tr><td valign='center'>ARG1</td></tr>\r\n");  
  webApn += F("<tr><td valign='center'><input type='text' name='ARG1' size='2' maxlength='2' value='1'></td></tr>\r\n");  
  webApn += F("<tr><td valign='center'> ARG2 (no space but '_')</td></tr>\r\n");
  webApn += F("<tr><td valign='center'> \r\n<input type='text' name='ARG2' size='20' maxlength='20' value='2'><br></td>\r\n</tr>\r\n");
  
  webApn += F("<tr>\r\n<td valign='center'>ARG3 - max 25\r\n</td>\r\n</tr>\r\n");
  webApn += F("<tr>\r\n<td valign='center'>\r\n<input type='text' name='ARG3' size='25' maxlength='25' value='test'><br></td>\r\n</tr>\r\n");
   
  webApn += F("<tr>\r\n<td valign='center'>\r\n<input type=submit value='Update'>\r\n</td>\r\n</tr>\r\n</table>\r\n</form>\r\n</body>\r\n</html>"); 
  server.send(200, "text/html",webApn);
}
```

### Debug Messages

```
Debug messages go here
```

### LINKED ISSUE #7106: ESP8266WebServer args() result to high?
Hi,

I'd like to understand if the following is a feature or bug:

I pass a few post parameters to a URL to set my WiFi configuration. These are:

- WiFiSSID = a
- WiFiPasswd = b
- DevName = c
- useDHCP =  true

I assumed if I get the number of arguments via httpserver.args() it would return "4", but it returns "5".

The content of the last argument is:
`plain=WiFiSSID=a&WiFiPasswd=b&DevName=c&useDHCP=true`

In the documentation I found:
```
arg - get request argument value, use arg("plain") to get POST body
argName - get request argument name
args - get arguments count
```

So is it expected that the post body gets counted in the args() integer?

best regards

Stefan

d-a-v:
`"plain"` is implicit and is not a real url parameter.
This PR will hide it from the argument counter, but `arg("plain")` is still available.

fixes #6435 

*edit*:

fixes #7106

#5252 added key `"plain"` to the list of arguments. Unfortunately as a side effect, this key is also always counted in the argument list. This PR removes this effect.


devyte:
I think this could break things. If we're going to break things,we likely should go all the way and rethink this feature.
Some thoughts:
- The "plain" case is currently meant for holding all args in raw form without parsing
- We can remove the current "plain" case from the counter (like in this PR)
- We can make sure the current "plain" arg doesn't show up in any iteration (hold the raw arg string elsewhere, not together with the parsed args)
- Implement a getter that returns the raw arg string

d-a-v:
> The "plain" case 

.. is the Content = the Payload of the http message.

d-a-v:
> I think this could break things.

Things are already and currently broken and this PR fixes the situation. 
Like I said, "plain" was previously artificially added (by me so solve another issue) but must not be counted in the url arguments.
- your point 1 is wrong. `arg("plain")` holds payload.
- point 2, of course
- point 3 can be done in another PR. When args are iterated, code will check for iterated name anyway
- point 4 can be done in another PR.

Both point 3 and point 4 are not urgent. Users who need it can already use `arg("plain")` when relevant. On the other hand, when testing the number of args is relevant (like in our examples and #7182), then the fix proposed by this PR is needed.
I think it should be merged asap because at least our examples rely on this.

d-a-v:
I forgot to mention that "plain" is always the last arg, so iterating on args counter will anyway never encounter the "plain" arg, like before #5252.

devyte:
@d-a-v what does args() (number of arguments) return when there were no arguments provided
- prior to this PR
- with this PR
?

d-a-v:
I was previously thinking that I added `arg("plain")` in the webserver.
It turns out I didn't, it was there before. I discovered about this artificial argument  when I fixed some bug some time ago (I had to read the code several time, maybe too many times so I finally believed I made this unusual hack).
This arg `plain` is added on post requests.
When it is added it is always last. So with or without it, there is no change on argument order.
The only consequence I know about it is the reported number of arguments, that this PR fixes.
I originally thought this PR was planned for 2.7.0 but when I realized the issue was there from long time ago I let milestone set to 3.0.0 because fixing it *might* be a breaking change (which I don't really believe but one never knows).

It is anyway **always not safe to rely on numbers** and it is preferred to check presence of a variable name, then get its value, so nothing is dependent on arg counter.

The only reason I can see about reading positional arguments is when one iterate over them to bravely parse a request, by looping from 0 to count-1, checking the name, and if it is recognized, getting its value to make something with it. In that case `plain` should be ignored (or reported?).

@devyte @vdeconinck

This has to be double checked

> What does args() (number of arguments) return when there were no arguments provided

before : 0 (not POST) or 1 (POST)
after: 0 (always)

> What does arg(0) (argument at index 0) return when the first arg provided has no value

`emptyString` in all cases

> What does arg(0) (argument at index 0) return when no args were provided

before: when it is a POST it can return the request content (== `arg("plain")`) which is irrelevant in that case (`emptyString` otherwise)
after: `emptyString` (even when count == 0 which would make an illegal call)

vdeconinck:
Hi, @d-a-v , thanks for the explanation.
Just to make things clear, when you say "it is preferred to check presence of a variable name, then get its value, so nothing is dependent on arg counter.", I guess you mean something along the lines of 
```
if (server.hasArg("dir")) {
  String dir = server.arg("dir");
  ...
}
```
But what is your take on the following construct from a risk/performance/cleanliness point of view:
```
String dir = server.arg("dir");
if (!dir.isEmpty()) {  
  ...
}
```

d-a-v:
@vdeconinck 
```
String dir = server.arg("dir");
if (!dir.isEmpty()) {  
  ...
}
```
Sorry for the late response. Yes, that's better because there's only one search.

Sparatutto:
Hi all, 
I write here because it seems to be the last thread about "plain".
I had to fix a minor issue on my code and then recompile the project (last 3.0.2 release).
But I still see "plain" counted by server.args() as per #6435.
Meanwhile I got that it has been removed from the count along the released version.

I tried to check for a step back on this but not found.
Can you confirm please which strategy you will keep here?
Do you confirm it will count or not?
 
arg

PR #6769, opened 2019-11-13
scheduled functions can schedule functions without oo loops

### LINKED ISSUE #6764: using schedule_function() within scheduled function prevents loop from running
#### Platform

- Hardware:				Wemos D1
- Core Version:			Latest git
- Development Env:		Sloeber ide
- Operating System:		Windows

### Problem Description

When using schedule_function() within a scheduled function, the run_scheduled_functions() will never return,
This has been working before, I do not know yet which of the scheduled functions PR removed the functionality.

```cpp

#include "Arduino.h"
#include "Schedule.h"

void cp()
{
	static int c = 0;
	if ((c++ % 10000) == 0 ) Serial.printf(".");
	schedule_function(cp);
}

void setup()
{
	Serial.begin(115200);
	schedule_function(cp);
}

void loop()
{
	static int l = 0;
	if ((l++ % 10000) == 0 ) Serial.printf(":");
}

d-a-v:
fixes #6764

### DIFF
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -117,10 +117,18 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
 
 void run_scheduled_functions()
 {
+    if (!sFirst)
+        return;
+
     esp8266::polledTimeout::periodicFastMs yieldNow(100); // yield every 100ms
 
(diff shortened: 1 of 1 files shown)


d-a-v:
[closing](https://github.com/esp8266/Arduino/issues/6764#issuecomment-553426192)

PR #6780, opened 2019-11-15
Comments for target and source sites of async scheduling

dok-net:
Where async event scheduling uses ``delay()``/``esp_yield()`` and ``esp_schedule()``, use same wording and properly identify source / target site pairs.
This is a purely cosmetic PR.
This PR helps identify these sites and aids in (my) discussion about the effects this pattern has on the
compatibility to (AVR) Arduino-style overridable weak-binding yield()/__yield() based 3rd party scheduling [libraries](https://github.com/dok-net/CoopTask).


dok-net:
@d-a-v oh great, thanks, that makes it much easier presenting a fix/change to the code spots. PR coming right next. Thanks again :-)

dok-net:
Continued in #6782

Issue #6777, opened 2019-11-15
Pull (#6765) breaks arduino-esp8266fs-plugin functionality


gnu-emu:
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[latest]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Wemos D1 mini]
- Flash Mode:			[?]
- Flash Size: 		[4MB]
- lwip Variant:   [v1.4|v2 Lower Memory]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

Hi,
First time as active github contributor here so please forgive if not everything is correct.

The change in upload.py  breaks arduino-esp8266fs-plugin functoinality. 

The plugin esp8266fs.jar executes:

....upload.py --chip esp8266 --port /dev/ttyUSB0 --baud 921600 write_flash 0x200000/tmp/something.bin --end

... so that esptool gives out:

esptool write_flash: error: argument <address> <filename>: Must be pairs of an address and the binary filename to write there

The problem seems to be that the change in upload.py.

cmdline in "esptool.main(cmdline)" becomes:

--chip', 'esp8266', '--port', '/dev/ttyUSB0', '--baud', '460800', 'write_flash', '0x200000', '--end', '0x0', '/tmp/something.bin'

... so that '--end' is passed to esptool and '0x0' is passed aditionally but should not.

Ref: #6765

devyte:
It's not correct, you ignored the issue template which requires certain information. Please edit your post above, add the template, follow the instructions in it, and fill out the required info.

devyte:
@earlephilhower I think a new plugin version is needed

gnu-emu:
[...]

PR #6781, opened 2019-11-15
Update upload.py

gnu-emu:
May another solution to:
https://github.com/esp8266/Arduino/issues/6777

### DIFF
--- a/tools/upload.py
+++ b/tools/upload.py
@@ -20,70 +20,84 @@ except:
     sys.stderr.write("pyserial or esptool directories not found next to this upload.py tool.\n")
     sys.exit(1)
 
-fakeargs = [];
 cmdline = []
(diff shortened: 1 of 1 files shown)


gnu-emu:
How about that ^

gnu-emu:
Only eating the '--end' is not working because  ['write_flash',  '0x0'] is appended and the binary is moved to the end. Also if another unknown plugin should use upload.py with two '--end' in one command esptool would not execute twice or (n-) times. As I'm a python beginner I did not know nesting if..else statements is a bad thing. What could happen while using the outer if-else (other than an ugly diff)?

gnu-emu:
Did not see your solution before. Thanks to your solution I learned about the if..elif statement :)

PR #6782, opened 2019-11-15
Change async esp_yield() / esp_schedule() "stop the Arduino world" metaphor sites

dok-net:
In order to let Arduino-like scheduling libraries, that shim yield() and/or delay(), make use of CPU cycles that are available while async callbacks are awaited, the CONT task must not be completely stopped.
This PR changes the way async callbacks are handled in ESP8266WiFi, currently the only code that works with NonOS async callbacks in ESP8266 Arduino core and core libraries according to Unix grep.


dok-net:
FYI: Alters part of #6107. Conflicts/supersedes #6212.

dok-net:
@devyte Delay(0) vs yield() in master, actual code path unrolled, illegal (panic) paths pruned:

delay(0) =>
```
esp_schedule();
cont_yield(g_pcont);
run_scheduled_recurrent_functions(); 
```
yield() =>
```
esp_schedule();
cont_yield(g_pcont);
run_scheduled_recurrent_functions();
```

dok-net:
``esp_schedule()`` can be un-removed in the CBs if there is no ill effect in calling it on non-suspended CONT, just to make the change smaller - reverting to delay(x) (x > 0) though turns this PR on its head.

dok-net:
Before you drive in, let me digest the input you have given me wrt power and performance. On a hunch, I think I have an idea involving just a rename and weak bindings, that leave behavior unchanged without shims, AND let's me recover those coveted cycles for MT when needed. I'll let you know. Thanks for the valuable feedback.

devyte:
No hurry, this is one of those cases where extremely careful consideration and thought is needed, or you risk breaking/worsening very subtle things that eventually blow things up in seemingly unrelated places. I suggest taking your time, doing broad and careful testing, and reporting your findings/progress back here for further discussion.

dok-net:
With the latest commit, this PR is supposed to introduce basically no difference in behavior over master, when the weak bindings are not touched.
One pattern is required and implemented, though.
For any ``delay()`` that can be resumed by ``esp_schedule()``, a guard loop is required, like so:
```
//delay(timeout_ms);
auto op_start_time = millis();
// will continue on timeout or when wifi_dns_found_callback fires
while (millis() - op_start_time < timeout_ms && _dns_lookup_pending) {
    esp_delay(timeout_ms);
}
```
Use ``esp_delay(1)`` in order to "give scheduled functions a chance to run (e.g. Ethernet uses recurrent)".

For any ``esp_yield()`` that is resumed by ``esp_schedule()``, a guard is also required, for example:
```
//esp_yield();
_wps_config_pending = true;
// will continue when wifi_wps_status_cb fires
while (_wps_config_pending) {
    esp_yield();
}
_wps_config_pending = false;
```

dok-net:
@devyte I was just reminded by a warning I've placed in CoopTask that this PR is sorely needed to recover cycles for MT that are currently locked up in a few busy loops or such.
Please, give this another benign review, and never mind CoopTask, predominantly check how it doesn't break anything, in places adds a layer of protection, and how it introduces a pattern rather than modify a few places to my liking :-)

dok-net:
The use of `Delegate` was removed from this PR to make it more acceptable, without that additional risk.

dok-net:
With commit https://github.com/esp8266/Arduino/pull/6782/commits/79d987fa187917aac2a39730f2881deb0843d6d7 WiFi Multi now uses `esp_delay`. As stated in the commit message and the source code, there exists the opportunity to reduce thrashing. @devyte @d-a-v Any ideas if instead of busy waiting (0ms) `intvl_ms` could be set to some higher value that allow the MCU to maybe even save some power? IHMO, it depends on how often the Scheduler.h routines are supposed to run there.

PR #6784, opened 2019-11-15
Version bump for package.json

ascillato:


TD-er:
Maybe ArduinoIDE also uses these JSON files in their library manager? (not sure, just wondering)

devyte:
Ping. Is this for PIO?

PR #6787, opened 2019-11-16
Update SdFat to not undefine F() macro

### LINKED ISSUE #6767: SDFS/SD make F() strings appear in RODATA, not flash
### Basic Infos

#### Platform

- Core Version:			[Core 2.6.0]

### Settings in IDE

- Flash Mode:			[dio]
- Flash Size: 		[4MB/1MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA|SERIAL]

### Problem Description

In my project, we have had log to SD present, but since it was disabled for a long time, I now made it compile again.
At first there is this strange thing when including SD.h, a lot of `__FlashStringHelper` related code would no longer compile.

For example I could not compile code like this anymore:
```c++
   static const __FlashStringHelper *gpMenu[8][3] = {
      F("&#8962;"),   F("Main"),          F("/"),              // 0
[...]
      F("&#128295;"), F("Tools"),         F("/tools"),         // 7
    };
```
Had to change it into separate functions with a switch statement returning F() macro defined strings.

After changing all of these, I ended up with this error:
```
c:/users/gijs/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: address 0x3fffe008 of .pio\build\debug_custom_ESP8266_4M1M\firmware.elf section `.bss' is not within region `dram0_0_seg'
c:/users/gijs/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: address 0x3fffe008 of .pio\build\debug_custom_ESP8266_4M1M\firmware.elf section `.bss' is not within region `dram0_0_seg'
*** [.pio\build\debug_custom_ESP8266_4M1M\firmware.elf] Error 1
```

Only when throwing out almost all code from my sketch, I was able to compile it wilt SD support enabled.
But I doubt it is useful, since you don't have any memory left.

Without SD support:
```
DATA:    [=====     ]  52.0% (used 42572 bytes from 81920 bytes)
PROGRAM: [=======   ]  69.0% (used 616020 bytes from 892912 bytes)
Creating BIN file ".pio\build\minimal_core_260_ESP8266_1M_OTA\firmware.bin" using ".pio\build\minimal_core_260_ESP8266_1M_OTA\firmware.elf"
```

With SD support:
```
DATA:    [========  ]  84.7% (used 69356 bytes from 81920 bytes)
PROGRAM: [=======   ]  70.9% (used 632760 bytes from 892912 bytes)
Creating BIN file ".pio\build\minimal_core_260_ESP8266_1M_OTA\firmware.bin" using ".pio\build\minimal_core_260_ESP8266_1M_OTA\firmware.elf"
```

That's an increase of over 26 kByte RAM usage.

I'm not entirely sure the issue with the `__FlashStringHelper` related code  has anything to do with this, but maybe it does somehow prevent strings to be stored in flash?

earlephilhower:
Fixes #6767 .  Remove the `undef F` from SysCall.h as it is not needed
nor used in the SD or SDFS libraries.  This puts F() strings back in
flash when using the SD/SDFS libs.


TD-er:
That leaves the question, why it was undef'ed in the first place?

earlephilhower:
He has several strings in the examples and core for status prints which use printf() which, several versions ago, did not support PROGMEM automatically w/o any change needed.

PR #6788, opened 2019-11-16
Make upload.py compatible with existing FS upload

earlephilhower:
PR #6765 introduced an incompatibility with the existing Java uploaders
for SPIFFS and LittleFS,  breaking them because of the upload.py
parameter format change to support single stage erase/upload.

Add in patch to silently eat the single --end and to parse the write
address and filename properly as generated by calls from the plugins,
while retaining compatibility with the current IDE changes.

### DIFF
--- a/tools/upload.py
+++ b/tools/upload.py
@@ -22,6 +22,7 @@ except:
 
 cmdline = []
 write_option = ''
+write_addr = '0x0'
 erase_addr = ''
 erase_len = ''
 
@@ -45,6 +46,16 @@ while len(sys.argv):
         erase_len = sys.argv.pop(0)
         thisarg = ''
 
+    # Backwards compatibility with fs upload tools, eat --end
+    if thisarg == '--end':
+        thisarg = ''
+
+    # Backwards compatibility with fs upload tools, parse write_flash for later use
+    if thisarg == 'write_flash':
+        write_addr = sys.argv.pop(0)
+        binary = sys.argv.pop(0)
+        thisarg = ''
+
     if os.path.isfile(thisarg):
         binary = thisarg
         thisarg = ''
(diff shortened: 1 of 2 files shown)


earlephilhower:
Fixes #6777 

Either this or #6781 should be merged ASAP.  #6781 reinstates the old path of pass-thru when `--end` is detected at the EOL, while this one tries to fit the old way into the new format and avoid separate parsing paths.  Either should solve it, but it's a style choice that I'm relatively indifferent about...

earlephilhower:
@d-a-v I'm sitting on the SC19 expo floor waiting for some networking to be done, so I can't actually test this.  It's what was discussed on the channel.  If you could give it a try before approving/requesting changed it'd be much appreciated.

tools/upload.py
```diff
@@ -39,28 +39,16 @@
...
         write_addr = sys.argv.pop(0)
         binary = sys.argv.pop(0)
-        thisarg = ''
-
-    if os.path.isfile(thisarg):
-        binary = thisarg
-        thisarg = ''
-
-    if len(thisarg):
+    elif len(thisarg):
```
gnu-emu: Sorry I'm still learning, but should this elif no stay an if? How do we pass the options to cmdline?

gnu-emu: I mean -chip esp8266 --port...etc.

earlephilhower: Those parameters won't match any of the other cases, so it will fall through to here.

gnu-emu: Is this elif not related to :
 if thisarg == "erase_flash":?

gnu-emu: Ah sorry. Now I got it.

gnu-emu:
I cloned https://github.com/earlephilhower/Arduino.git, installed it using-git-version doc on Arduino 1.8.10 (with portable directory on Ubuntu Bionic) and for me https://github.com/esp8266/arduino-esp8266fs-plugin/releases/download/0.4.0/ESP8266FS-0.4.0.zip is working. Sketch upload successful and working on the esp.

earlephilhower:
@gnu-emu thanks very much for the report and your other PRs on this!  It's one thing to report a bug, but a much harder thing to try and fix it...

Anyway, I think at some point we'll drop upload.py and call the esptool.py directly from the plugins and from the IDE (3.0 probably).  The suggestions you've got on the other repos will probably make sense then.

Issue #4183, opened 2018-01-17
Progress of HTTP-Update


Disane87:
Referecing the example for the HTTP-Update: https://github.com/esp8266/Arduino/libraries/ESP8266httpUpdate/examples/httpUpdate/httpUpdate.ino

Is there any possibility to get the current status of the download and flash progress?
I want to show a progressbar with the current percentage. 

Any chance to achive this?

CarlosJimenez:
I tried this (present in the latest version of Update.cpp) and it works for me!!:

```cpp
void progressFunc(unsigned int progress,unsigned int total) {
  Serial.printf("Progress: %u of %u\r", progress, total);
};
Update.onProgress(progressFunc);
```

earlephilhower:
Closing as fixed via #6796

PR #6796, opened 2019-11-17
Add start, end, progress, and error callbacks for ESP8266HTTPUpdate

earlephilhower:
Replaces abandoned #1817

Add an optional std::function callback (so it supports lambdas and
normal functions) with an optional data parameter (for passing in a
`this` or other state info).

Update example with its use.

From @baruch's original pull request:
The callback is called when the upgrade actually starts rather than just
the initial query so that the user can know that it will not take longer
and can also prepare for the upgrade by shutting down other works.


libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.h
```diff
@@ -64,6 +64,8 @@ enum HTTPUpdateResult {
 
 typedef HTTPUpdateResult t_httpUpdate_return; // backward compatibility
 
+typedef std::function<void(void *data)> HTTPUpdateCB;
```
devyte: prefer using = syntax

earlephilhower:
I'm going to merge this with an update for #2694, too, which has add'l CBs.

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.h
earlephilhower: Will do

earlephilhower:
Now implements the 4 CBs in@karlp's #2694.

earlephilhower:
Fixes #4183

PR #6797, opened 2019-11-17
Add const char* content to ESP8266WebServer::send()

earlephilhower:
Supercedes #3492

Allow sending raw binary data or strings directly without conversion to
a String to reduce memory overhead when possible.

From original @timw1971 PR #3492
Added public functions to allow content to be uploaded using const
char*. For some cases, this can remove the need for content to be copied
into a String, and thus can be considerably more space-efficient.


libraries/ESP8266WebServer/examples/HelloServer/HelloServer.ino
```diff
@@ -67,6 +67,17 @@ void setup(void) {
     server.send(200, "text/plain", "this works as well");
   });
 
+  server.on("/gif", []() {
+    uint8_t gif[] = {
```
devyte: This should be a static at least. Or be moved out to global space.

earlephilhower: Done.

PR #6798, opened 2019-11-17
[SSDP] add `schema(Print &) const`

earlephilhower:
Supercedes #2806

Make SSDP::schema(WiFiClient&) use a by-ref (reduce stack use)

Add a SSDP::schema(Print&)

From @Palatis' original PR:
useful when using AsyncWebServer.

### DIFF
--- a/libraries/ESP8266SSDP/ESP8266SSDP.cpp
+++ b/libraries/ESP8266SSDP/ESP8266SSDP.cpp
@@ -73,7 +73,7 @@ static const char _ssdp_packet_template[] PROGMEM =
   "SERVER: Arduino/1.0 UPNP/1.1 %s/%s\r\n" // _modelName, _modelNumber
   "USN: %s\r\n" // _uuid
   "%s: %s\r\n"  // "NT" or "ST", _deviceType
-  "LOCATION: http://%u.%u.%u.%u:%u/%s\r\n" // WiFi.localIP(), _port, _schemaURL
+  "LOCATION: http://%s:%u/%s\r\n" // WiFi.localIP(), _port, _schemaURL
   "\r\n";
 
 static const char _ssdp_schema_template[] PROGMEM =
@@ -88,7 +88,7 @@ static const char _ssdp_schema_template[] PROGMEM =
   "<major>1</major>"
   "<minor>0</minor>"
   "</specVersion>"
-  "<URLBase>http://%u.%u.%u.%u:%u/</URLBase>" // WiFi.localIP(), _port
+  "<URLBase>http://%s:%u/</URLBase>" // WiFi.localIP(), _port
   "<device>"
   "<deviceType>%s</deviceType>"
   "<friendlyName>%s</friendlyName>"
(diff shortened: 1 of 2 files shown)


libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -295,6 +295,22 @@ void SSDPClass::schema(WiFiClient client) {
                );
 }
 
+void SSDPClass::schema(Print &print) const {
```
devyte: This looks very similar to the WiFiClient overload. Why have both? Can they be unified?

earlephilhower:
@d-a-v, for the SSDP patch, I'm just the messenger.
I can't think of a good use case for a Printable vs. sending it over the wire, honestly.

However, maybe it makes sense to use just export `Print&` interface and drop the WiFiClient& one?  WiFiClient inherits a Print& interface from Stream...

earlephilhower:
Err, @devyte  I meant to ping!

devyte:
> maybe it makes sense to use just export Print& interface and drop the WiFiClient& one

Exactly why I asked. I suggest picking which of the two bodies looks best and keeping that one with Print arg.
Something that bothers me: the IP printout in both cases, one with the octets the other with IP2STR. I think the IPAddress::toString() method should be used instead, but that means updating the schema string, I think.

libraries/ESP8266SSDP/ESP8266SSDP.cpp
earlephilhower: Done, only ::schema(Print&) survives.

earlephilhower:
Both comments addressed.  SSDP will now use ip.toString() instead of octet extraction, and will export only the schema(Print&) which should work for any existing code as WiFiClient has a Print interface..

earlephilhower:
Well, I guess C++ can't convert silently from a WiFiClient to Print for us.  Multiple inheritance C++ classes might have such limitations (not like Java where there's a first-class interface type).

Any ideas?

````
/home/travis/build/esp8266/Arduino/libraries/ESP8266SSDP/examples/SSDP/SSDP.ino:29:32: error: no matching function for call to 'SSDPClass::schema(esp8266webserver::ESP8266WebServerTemplate<WiFiServer>::ClientType)'

       SSDP.schema(HTTP.client());

                                ^

/home/travis/build/esp8266/Arduino/libraries/ESP8266SSDP/examples/SSDP/SSDP.ino:29:32: note: candidate is:

In file included from /home/travis/build/esp8266/Arduino/libraries/ESP8266SSDP/examples/SSDP/SSDP.ino:3:0:

/home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266SSDP/ESP8266SSDP.h:65:10: note: void SSDPClass::schema(Print&) const

     void schema(Print &print) const;

          ^

/home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266SSDP/ESP8266SSDP.h:65:10: note:   no known conversion for argument 1 from 'esp8266webserver::ESP8266WebServerTemplate<WiFiServer>::ClientType {aka WiFiClient}' to 'Print&'
````

devyte:
Might be running into the 1-step implicit casting rule. I'll take a look at some point.

PR #6802, opened 2019-11-18
host build: optimistic_yield() wrongly delays() instead of avoiding excessive yields

dok-net:
On host, can work just like plain `yield()` there - do nothing at all.


dok-net:
PR is OK, CI once more has network issues?!

PR #6804, opened 2019-11-18
Fix optimistic_yield to not yield on each call after x µs

dok-net:
Before, once an invocation of `loop()` runs for x µs, each call to `optimistic_yield()` does really `yield()`.
Now, only yield every x µs between `optimistic_yield()` or `yield()` calls.


dok-net:
@d-a-v I've used the ESP.get… wrappers instead... do you approve of the possible wrapping for large interval numbers? Adding corrective code probably ruins any advantage from not using micros()?

dok-net:
Dear @earlephilhower, could you please check again what's up? Thank you!

cores/esp8266/Schedule.cpp
```diff
@@ -117,8 +117,6 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
 
 void run_scheduled_functions()
 {
-    esp8266::polledTimeout::periodicFastMs yieldNow(100); // yield every 100ms
```
earlephilhower: Can this be separated out into a separate PR?  Seems like changes in Schedule.cpp are not directly related to the optimistic_yield fix.

cores/esp8266/Arduino.h
```diff
@@ -204,7 +204,19 @@ void setup(void);
 void loop(void);
 
 void yield(void);
-void optimistic_yield(uint32_t interval_us);
+
+#ifndef F_CPU
```
earlephilhower: Is F_CPU ever not defined?  clockCyclesPerMicrosecond() is always defined, and it just uses that constant, no?

devyte: @earlephilhower that's a bit of jumping the gun for an idea I have to allow a 3rd option for cpu speed: runtime settable. I'm not sure yet if it's feasible

dok-net: The diff presentation here is quite unreadable, unless one remembers exactly what the surrounding code was. I'll have to check later. BUT: I'd strongly like to keep this check for F_CPU in any case, it doesn't hurt and IMHO makes the change and intention so much clearer.

cores/esp8266/Schedule.cpp
dok-net: This Github review presentation is useless, I'll have to browse the context to refresh my memory :-) So: I'll comment later.

dok-net: You're right, this was giving myself an unnecessary merge conflict. I've removed this.

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -57,8 +56,8 @@ cont_t* g_pcont __attribute__((section(".noinit")));
 /* Event queue used by the main (arduino) task */
 static os_event_t s_loop_queue[LOOP_QUEUE_SIZE];
 
-/* Used to implement optimistic_yield */
-static uint32_t s_micros_at_task_start;
+/* Used to implement __optimistic_yield */
+static uint32_t s_cycles_since_yield_start;
```
earlephilhower: Suggest a better name for this would be `s_cycles_at_yield_start`.  It is holding an absolute cycle count(at), not a delta(since).

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -125,14 +125,19 @@ extern "C" void __yield() {
 extern "C" void yield(void) __attribute__ ((weak, alias("__yield")));
 
 extern "C" void optimistic_yield(uint32_t interval_us) {
-    if (can_yield() &&
-        (system_get_time() - s_micros_at_task_start) > interval_us)
+    const uint32_t intvl_cycles = interval_us *
+#if defined(F_CPU)
+        clockCyclesPerMicrosecond();
+#else
+        getCpuFreqMHz();
```
earlephilhower: This function does not exist.  There is, however, an `ESP.getCpuFreqMHz`

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -57,8 +56,8 @@ cont_t* g_pcont __attribute__((section(".noinit")));
 /* Event queue used by the main (arduino) task */
 static os_event_t s_loop_queue[LOOP_QUEUE_SIZE];
 
-/* Used to implement optimistic_yield */
-static uint32_t s_micros_at_task_start;
+/* Used to implement __optimistic_yield */
```
earlephilhower: There is no `__optimistic_yield`, only `optimistic_yield`.

PR #6807, opened 2019-11-19
[BREAKING] Update FS plugin info for python3 compat

earlephilhower:
The FS uploader plugins need to be updated to use python3 and not
python, or they will fail on Windows (or Linux boxes without an
installed python2 interpreter).

Update the documents to point to the new versions.


earlephilhower:
The SPIFFS uploader is still maintained by @me-no-dev , so need to track 
https://github.com/esp8266/arduino-esp8266fs-plugin/pull/65
before merging.

earlephilhower:
Fixes #6829

Bubble007:
Hello alltogether,

I have a problem to install the Arduino plugin for uploading files to ESP8266 file system.

I get the error message: “Python was not found but can be installed from the Microsoft Store: https://go.microsoft.com/fwlink?linkID=2082640SPIFFS Upload failed!”

I have Win 10 and der Arduino Ide 1.8.12.

I installed also Python 3.

I have installed the ESP8266FS-0.5.0.zip from earlephilhower in the comment on 25 Nov 2019. (https://github.com/esp8266/Arduino/issues/6829#issuecomment-558295745)

All doesn’t help. Can you help?

PR #6813, opened 2019-11-19
Add flash mapping defines at compile time

### LINKED ISSUE #6806: [feature request] It would be nice to have the Arduino IDE add a #define for the configured flash size
In cases where specific code depends on the existence of FS configured flash, it would be nice to have a #define available to prevent compilation/upload if the Flash config is misconfigured, especially in an OTA situation where this could prevent the device from booting.

For example if the code depends on saving device settings to SPIFFS and therefore booting is impossible with no FS configured, it would be great to be able to code a safety measure like:
#ifndef FS1MB
   #error The Flash Config must be set to 1M FS
#endif 

This would prevent an OTA upload where the device couldn't possibly boot afterwards and save a lot of hassle.

d-a-v:
This let user code to take actions (like checking mapping) at compile time
defines example: `-DEEPROMSTART=0x3FB000 -DFSSTART=0x200000 -DFSEND=0x3FA000 -DFSBLOCKSIZE=8192`
Addresses are relative to SPI start address wich is `0x40200000`

Fixes #6806 @sprior


sprior:
Looks good, thanks for implementing.

d-a-v:
closed and included by #6690

PR #6816, opened 2019-11-20
Implement invert for HardwareSerial

### LINKED ISSUE #4896: Serial doesn't provide inverse operation as provided by Espressif
In section 3.5 of the espressif interface documentation is written there is functionality to inverse the signal lines of the uart.

I need to inverse the rxd line which is supported by the esp8266.

3.5. Reverse Operation
All the input and output signals can be configured to the reverse direction internally.
#define UART_DTR_INV (BIT(24))
#define UART_RTS_INV (BIT(23))
#define UART_TXD_INV (BIT(22))
#define UART_DSR_INV (BIT(21))
#define UART_CTS_INV (BIT(20))
#define UART_RXD_INV (BIT(19))
By setting the allocation of registers, the input and output of the corresponding signal lines can be
reversed.
Interface function: void UART_SetLineInverse(uint8 uart_no, UART_LineLevelInverse
inverse_mask);

http://bienonline.magix.net/public/esp8266-faq/8E-ESP8266__Interface_UART__EN_v0.2.pdf

I searched in the esp8266 arduino sdk without any luck. For now I have ordered some 7404 ic's from aliexpress but it would be nice if this functionality can be enabled in the sdk.

dok-net:
Fixes #4896

### DIFF
--- a/cores/esp8266/HardwareSerial.h
+++ b/cores/esp8266/HardwareSerial.h
@@ -84,7 +84,7 @@ public:
         begin(baud, config, mode, 1);
     }
 
-    void begin(unsigned long baud, SerialConfig config, SerialMode mode, uint8_t tx_pin);
+    void begin(unsigned long baud, SerialConfig config, SerialMode mode, uint8_t tx_pin, bool invert=false);
 
     void end();
 
--- a/cores/esp8266/HardwareSerial.h
+++ b/cores/esp8266/HardwareSerial.h
@@ -73,18 +73,23 @@ public:
 
     void begin(unsigned long baud)
     {
-        begin(baud, SERIAL_8N1, SERIAL_FULL, 1);
+        begin(baud, SERIAL_8N1, SERIAL_FULL, 1, false);
     }
     void begin(unsigned long baud, SerialConfig config)
     {
-        begin(baud, config, SERIAL_FULL, 1);
+        begin(baud, config, SERIAL_FULL, 1, false);
     }
     void begin(unsigned long baud, SerialConfig config, SerialMode mode)
     {
-        begin(baud, config, mode, 1);
+        begin(baud, config, mode, 1, false);
     }
 
-    void begin(unsigned long baud, SerialConfig config, SerialMode mode, uint8_t tx_pin, bool invert=false);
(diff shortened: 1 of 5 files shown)


cores/esp8266/HardwareSerial.h
```diff
@@ -84,7 +84,7 @@ class HardwareSerial: public Stream
         begin(baud, config, mode, 1);
     }
 
-    void begin(unsigned long baud, SerialConfig config, SerialMode mode, uint8_t tx_pin);
+    void begin(unsigned long baud, SerialConfig config, SerialMode mode, uint8_t tx_pin, bool invert=false);
```
devyte: This is a change in signature, which can be potentially breaking. Instead, prefer an overload with the new signature and forward the call with the added arg as default value.

jandrassy: it is not a virtual method, so there is no problem with a new parameter with default value

dok-net: Too late now :-)

Jeroen88:
Isn't it a bit counter intuitive to use uart_set_baudrate() to set inversion? Wouldn't it be better to have an additional low level function _uart_set_inversion(bool invert) that reads USC0 and clears or sets the bits for inversion?

dok-net:
@Jeroen88 what are you talking about, it's uart_init()

cores/esp8266/HardwareSerial.h
devyte: @jandrassy if you want to discuss, look me up on gitter

Jeroen88:
@dok-net You are right, my mistake... :)

dok-net:
The CI failure is (again) nothing to do with the PR, but network / setup related?

earlephilhower:
Looks like Sphinx (the doc generator) is barfing again on installation.

earlephilhower:
#6817 will make CI work again, then you can update and the CI should then pass.

morganchristiansson:
This is not documented @ https://arduino-esp8266.readthedocs.io/en/latest/reference.html#serial ? Had to read PR to figure out usage. Found no other leads searching on Google.


```
Serial.begin(115200, SERIAL_8N1, SERIAL_FULL, 1, true);
```

PR #6830, opened 2019-11-24
Fix pointer arithmetic

devyte:
Per @AdrianEddy
Actually advance position to write at while looping


earlephilhower:
Damn, sorry @d-a-v.  I could swear I did a test that passed with this using the WiFi Hotspot mode of Win10 before submitting, but looks like I flubbed it.

earlephilhower:
Err, Sorry, @devyte and @AdrianEddy, that is!

PR #6837, opened 2019-11-25
Handle HEAD requests for static files correctly

mathertel:
Some browser issue a HEAD before a GET requests when a &lt;object&gt; tag is used.
This change handles HEAD requests correctly for static files by returning the full header but no body according to https://tools.ietf.org/html/rfc7231#section-4.3.2.


libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -125,7 +125,7 @@ class StaticRequestHandler : public RequestHandler<ServerType> {
         if (_cache_header.length() != 0)
             server.sendHeader("Cache-Control", _cache_header);
 
-        server.streamFile(f, contentType);
+        server.streamFile(f, contentType, requestMethod);
```
d-a-v: Why not checking method before opening the file ?
(ie: move the File..open lines after cache header lines above,
 and surround file+streamfile block with `if method==HEAD`, no need to overload `::streamFile`)

Beside, in your version, `_streamFileCore` is called and sets content length which I'm not sure should be set in the `HEAD` case.

d-a-v: Forget my above comment,
- content-length must be set to the same value with HEAD as compared to GET. 
- consequently file must be opened to know its size

ref:
```
The Content-Length entity-header field indicates the size of the entity-body,
in decimal number of OCTETs, sent to the recipient or, in the case of the
HEAD method, the size of the entity-body that would have been sent had
the request been a GET.
```

mathertel: As you can see I tried to make minimal changes only.
The HEAD is exactly the same as GET including headers and Content-Length. - there is just no body.
I can move the check on level up but the I suggest to rename _streamFileCore to _sendFileHeaders.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -151,10 +151,15 @@ class ESP8266WebServerTemplate
 
   static String urlDecode(const String& text);
 
+  // send file based header and stream body when not "HTTP_HEAD" 
   template<typename T>
-  size_t streamFile(T &file, const String& contentType) {
+  size_t streamFile(T &file, const String& contentType, HTTPMethod requestMethod = HTTP_GET) {
```
devyte: Signature change. Prefer an overload and forwarder with the default value.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -151,10 +151,15 @@ class ESP8266WebServerTemplate
...
     _streamFileCore(file.size(), file.name(), contentType);
-    return _currentClient.write(file);
+    if (requestMethod != HTTP_HEAD) {
```
devyte: Prefer direct over inverse logic

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -151,10 +151,15 @@ class ESP8266WebServerTemplate
...
+      return _currentClient.write(file);
+    } else {
+      return 0;
```
devyte: Remove else and leave return statement as last line in the body

mathertel: done.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
mathertel: done.

PR #6854, opened 2019-11-28
Update ESP8266WifiGeneric.cpp to allow compile time selection of DNS …

altelch:
…resolution precedence IPv4/IPv6

eg. platformio uses prebuilt liblwip6-1460-feat.a which is built with DNS resolution order IPV4 before IPv6. This causes problems in IPv6 only environments if there is a DNS record with A and AAAA. lwip then tries to connect to the resolved v4 even if there is no global IPv4 address. Didn't find a way to get pio rebuild the complete library. That's why I patched ESP8266WifiGeneric.cpp to use dns_gethostbyname_addrtype instead of dns_gethostbyname (which is only a wrapper for gethostbyname_addrtype and the define to prefer v4). With this change it's possible to select resolution beaviour during compile time.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
@@ -644,6 +644,7 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
     return (err == ERR_OK) ? 1 : 0;
 }
 
+#if LWIP_VERSION_MAJOR > 1
 int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms, uint8_t resolveType)
 {
     ip_addr_t addr;
@@ -656,11 +657,7 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
     }
 
     DEBUG_WIFI_GENERIC("[hostByName] request IP for: %s\n", aHostname);
-#if LWIP_VERSION_MAJOR > 1
     err_t err = dns_gethostbyname_addrtype(aHostname, &addr, &wifi_dns_found_callback, &aResult, resolveType);
-#else
-    err_t err = dns_gethostbyname(aHostname, &addr, &wifi_dns_found_callback, &aResult);
-#endif
     if(err == ERR_OK) {
         aResult = IPAddress(&addr);
     } else if(err == ERR_INPROGRESS) {
@@ -682,6 +679,7 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
 
     return (err == ERR_OK) ? 1 : 0;
 }
+#endif
 
 /**
  * DNS callback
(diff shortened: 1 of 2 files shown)


d-a-v:
Because your PR might be a breaking change
(example: dual ipv4 ipv6 dns, with ipv4 only links)
(yes that's what I have at work: ipv6 is disabled on access points but not on wired network)

We have:
```cpp
int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms)
```
What about changing it to:
```cpp
int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms, resolveType_t resolveType)
```
and adding for backward compatibility:
```cpp
int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms)
{
    return hostByName(aHostname,aResult,timeout_ms,RESOLVE_IPV4);
}
```

altelch:
If you look at lwip source in dns.h you see the default behaviour. dns_gethostbyname is only a wrapper for dns_gethostbyname_addrtype and there we have LWIP_DNS_ADDRTYPE_DEFAULT as last argument. LWIP_DNS_ADDRTYPE_DEFAULT. if LWIP_IPV4 && LWIP_IPV6 the Default is LWIP_DNS_ADDRTYPE_IPV4_IPV6, if IPv4 only we have v4 only resolving and if IPv6 only we have IPv6 only resolving. So there is no change to default behaviour after this fix. You need to change the default to get another behaviour than now.

altelch:
But I have to admit, making this an runtime option would be more flexible than to have this only defined at compile time. So I could decide based on aquired addresses how to to my dns queries.

d-a-v:
I may not have understood.

[lwIP's code is](http://git.savannah.nongnu.org/cgit/lwip.git/tree/src/core/dns.c?h=STABLE-2_1_2_RELEASE#n1517):
```cpp 
err_t
dns_gethostbyname(const char *hostname, ip_addr_t *addr, dns_found_callback found,
                  void *callback_arg)
{
  return dns_gethostbyname_addrtype(hostname, addr, found, callback_arg, LWIP_DNS_ADDRTYPE_DEFAULT);
}
```

and your PR makes the same call:
```cpp
err_t err = dns_gethostbyname_addrtype(aHostname, &addr, &wifi_dns_found_callback, &aResult, LWIP_DNS_ADDRTYPE_DEFAULT);
```
So it looks like a no-op.

Do you check the number of answers given by the call ?
In case there are two, do they arrive always in the same order ?

I tend to think that if there is a need to force the type of an answer, the api update proposal above would help.

altelch:
The function doesn't return multiple results but works short circuit. If v4 is resolved it doesn't try v6. Only if no v4 is found a second try is made to get an AAAA record. I prefer your suggestion to make an api update. Will look into this tomorrow (maybe this evening).

altelch:
Hi, don't understand why Mock Trivial Test fails. Looks like the Test runs into a wrong branche LWIP_VERSION_MAJOR > 1 but has wrong lwip version. My real world tests with enabled lwip2 work as expected. New API call with resolvType also works with one drawback: lwip caches DNS results and if you do a v4 query and after that a v6 query to the same name you still get the v4 entry as the first DNS call is cached without considering the query type.

altelch:
Did https://arduino-esp8266.readthedocs.io/en/latest/installing.html#using-git-version, rebuilt FSBrowser.ino with my code changes and the build cleanly runs. Can't find a reason why the Mock Trivial Test fails during linking the binary.

altelch:
Ok, will close this PR and make a new one. Too many ugly commits caused by testing.

Issue #2162, opened 2016-06-17
I2C hang with very slow I2C slave (pic12f1840@500khz)


miky2k:
----------------------------- Remove above -----------------------------
#### Hardware

Hardware:           ESP-07
Core Version:       2.2.0
### Description

On very slow device i2c hang SDA low and give strange results  01111001 instead 00111100 .
### Settings in IDE

Module:             Generic ESP8266 Module
Flash Size:         512K/64K
CPU Frequency:  80Mhz
Flash Mode:     qio
Flash Frequency:    40Mhz?
Upload Using:       SERIAL?
Reset Method:       ck
### Sketch

Wire.begin(12,13); //I2C pin
Wire.setClockStretchLimit(2000);

In my software(l is not important) : 
    Wire.beginTransmission(Addr);
    Wire.write(0b00111100);
    Serial.println("I2C write---"+String(l));
    Serial.println(Wire.endTransmission(1));

Serial Output:

I2C write---4
3

![i2c](https://cloud.githubusercontent.com/assets/6616232/16148978/b0b028f4-348d-11e6-80ab-75bbe38f1bd6.jpg)
##

me-no-dev:
any particular reason why your slave does not release SDA?

far5893:
I disconnect physically only sda from device but bus remain low .

devyte:
@igrr this has milestone 2.4.0. However, I don't fully understand the issue, let alone what needs to be fixed in Wire.
Is the issue still relevant?

far5893:
If you have different clock stretch need on devices  is relevant!

enjoyneering:
check this - https://github.com/enjoyneering/ESP8266-I2C-Driver

d-a-v:
For reference ([link](https://github.com/esp8266/Arduino/issues/2330#issuecomment-391303255))
I will make a PR for that

devyte:
CC @Tech-TX

PR #6860, opened 2019-11-30
added clock stretch yield, [issue 2162] fixed twi::status

Tech-TX:
Finally figured out what I'd done wrong with the earlier commit.  No -a in a local commit!

issue #2162

Added devyte's PolledTimeout in a nested loop replacing the original WAIT_CLOCK_STRETCH().  The outer loop exits if SCL goes high or the loop times out.  The inner loop does a yield() every 5ms while we're waiting.
Additionally, two small changes to twi::status that add a clock stretch test before we exit if we find SCL stuck low, and the trailing **if (!write_start())** at the end is removed as it leaves SDA low, killing the bus.

### DIFF
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -22,6 +22,7 @@
 #include "twi.h"
 #include "pins_arduino.h"
 #include "wiring_private.h"
+#include "PolledTimeout.h"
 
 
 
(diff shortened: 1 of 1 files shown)


earlephilhower:
Failure looked to be a transient Chocolatey failure (Windows package manager).  Restarted, it's going OK now...

dok-net:
I propose amending this to, pending #6804 (for corrected timing):
```c++
        esp8266::polledTimeout::oneShotFastUs timeout(twi_clockStretchLimit);
        while(!timeout && !SCL_READ())  // loop is stretch duration up to stretch limit
        { 
            optimistic_yield(5000);
        }
```

dok-net:
@devyte Code smell, priority inversion?

PR #6861, opened 2019-11-30
fix device tests

d-a-v:
web server post test is (temporarily) removed due to general pyton2=>3 failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ Arduino core for ESP8266 WiFi chip
 
 # Quick links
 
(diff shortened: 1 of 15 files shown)


earlephilhower:
Thanks, @d-a-v.  Did you intend to remove the ResetMethod from the FQBN?

d-a-v:
@earlephilhower the reset method was wrong since recent changes in boards.txt.
I had changed to rely on default value but it is true that an explicit value is better.
Fixed.

PR #6865, opened 2019-12-01
Make DNS resolution order selectable during runtime or compile time

altelch:
Modified ESP8266WifiGeneric to allow selection of DNS resolution order. Can be controlled by dns.h defines or in an extended api call Wifi.hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms, uint8_t resolveType) in dual stack mode (LWIP_IPV4 and LWIP_IPV6 is set).


altelch:
Sorry for the commit-chaos. Hat no dev-environment at hand. Now IPv6.ino has an example on how to use the extended hostByName().

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -617,7 +617,47 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
...
+    }
+
+    return (err == ERR_OK) ? 1 : 0;
+}
+
+#if LWIP_IPV4 && LWIP_IPV6
+int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms, uint8_t resolveType)
+{
+    ip_addr_t addr;
+    aResult = static_cast<uint32_t>(0);
```
devyte: Instead of assigning 0, consider assigning INADDR_NONE.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -617,7 +617,47 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
...
+        DEBUG_WIFI_GENERIC("[hostByName] Host: %s lookup error: %d!\n", aHostname, (int)err);
+    } else {
+        DEBUG_WIFI_GENERIC("[hostByName] Host: %s IP: %s\n", aHostname, aResult.toString().c_str());
+    }
+
+    return (err == ERR_OK) ? 1 : 0;
+}
+
+#if LWIP_IPV4 && LWIP_IPV6
+int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms, uint8_t resolveType)
```
devyte: The arg resolveType is not type safe not is it being checked for wrong value. Instead, prefer an scoped enum or an enum class with values 0-3.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -617,7 +617,47 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
...
+    aResult = static_cast<uint32_t>(0);
+
+    if(aResult.fromString(aHostname)) {
+        // Host name is a IP address use it!
+        DEBUG_WIFI_GENERIC("[hostByName] Host: %s is a IP!\n", aHostname);
+        return 1;
+    }
+
+    DEBUG_WIFI_GENERIC("[hostByName] request IP for: %s\n", aHostname);
+    err_t err = dns_gethostbyname_addrtype(aHostname, &addr, &wifi_dns_found_callback, &aResult,resolveType & 3); // limit to defined types 0 -3
```
devyte: As explained above, the operation on resolveType will silently ignore errors. Instead, prefer receiving as a scoped enum or enum class, then error check, then cast to type here before passing to dns_gethostbyname_addrtype().

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
altelch: This is only a copy of the original code. I didn't make this assign but will change both occurences.

altelch:
Ok, changed to using an enum class for DNS resolution type. Hope this is what you intended.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -42,6 +42,14 @@ typedef std::shared_ptr<WiFiEventHandlerOpaque> WiFiEventHandler;
 
 typedef void (*WiFiEventCb)(WiFiEvent_t);
 
+enum class DNSResolveType: uint8_t
+{
+    DNS_AddrType_IPv4 = 0,	// LWIP_DNS_ADDRTYPE_IPV4 = 0
+    DNS_AddrType_IPv6,		// LWIP_DNS_ADDRTYPE_IPV6 = 1
+    DNS_AddrType_IPv4_IPv6,	// LWIP_DNS_ADDRTYPE_IPV4_IPV6 = 2
+    DNS_AddrType_IPv6_IPv4	// LWIP_DNS_ADDRTYPE_IPV6_IPV4 = 3
```
d-a-v: What do you think of 
```cpp
{
    DNS_AddrType_IPv4 = LWIP_DNS_ADDRTYPE_IPV4,
    DNS_AddrType_IPv6 = LWIP_DNS_ADDRTYPE_IPV6,
    DNS_AddrType_IPv4_IPv6 = LWIP_DNS_ADDRTYPE_IPV4_IPV6,
    DNS_AddrType_IPv6_IPv4 = LWIP_DNS_ADDRTYPE_IPV6_IPV4,
}
```

altelch: Hi, this was exactly my first try but it caused problems with the CI built as some of the tests failed because of LWIP_DNS_ADDRTYPE_* isn't defined during compile time. I did put the typedef into an IF clause to have it only built if Dualstack is enabled but that caused a even more strange built problem with test 1. This was the only working solution I could find without breakting the tests.

PR #6868, opened 2019-12-02
optimistic_yield is correct use inside libraries etc. code

dok-net:
optimistic_yield is there to prevent excess yielding in tight loops, just like the busy wait.

### DIFF
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -126,10 +126,10 @@ private:
     {
         esp8266::polledTimeout::oneShotFastUs timeout(twi_clockStretchLimit);
         esp8266::polledTimeout::periodicFastUs yieldTimeout(5000);
-        while(!timeout && !SCL_READ(twi_scl))  // outer loop is stretch duration up to stretch limit
+        while(!timeout && !SCL_READ(twi_scl))  // busy wait for stretch duration up to stretch limit
         { 
-            if (yieldTimeout)   // inner loop yields every 5ms
-                yield();
+            if (yieldTimeout)   // device response timeout exceeded
+                optimistic_yield(10000); // actually yield once every 10ms since last yield
         }
     }
 


dok-net:
@Tech-TX Please check comments' wording and reasoning. Thx.

cores/esp8266/core_esp8266_si2c.cpp
```diff
@@ -126,10 +126,10 @@ class Twi
     {
         esp8266::polledTimeout::oneShotFastUs timeout(twi_clockStretchLimit);
         esp8266::polledTimeout::periodicFastUs yieldTimeout(5000);
-        while(!timeout && !SCL_READ(twi_scl))  // outer loop is stretch duration up to stretch limit
+        while(!timeout && !SCL_READ(twi_scl))  // busy wait for stretch duration up to stretch limit
         { 
-            if (yieldTimeout)   // inner loop yields every 5ms
-                yield();
+            if (yieldTimeout)   // device response timeout exceeded
+                optimistic_yield(10000); // actually yield once every 10ms since last yield
```
earlephilhower: I think this misses the point of the `yieldTimeout`.  The original code makes a 5ms timer and then yields() when it fires (i.e. every 5ms).  Now you've got a timer that checks every 5ms but only yields every 10ms.

So, the original code is not yield()ing lots, anyway.  I don't see a reason to change it from 5 to 10ms, but if there was one the proper way would be to set yieldTimeout(10000) or just drop yieldTimeout completely and always optimizeic_yield(10ms) which would be essentially the same thing.

dok-net: Removing either of the timeouts has a negative impact in my reckoning.
Remove yieldTimeout, it may yield too early while the hardware is still expected to reply and things were intented to complete rapidly;
Increase PolledTimeout to 10ms and it busy-locks longer than the initial code intended to wait for the immediate hardware reply;
or leave the original code and it yields too often at 5ms interval.
I know it looks confusing, but think of it as a one-shot PolledTimeout to keep the first 5ms yield-free for performance, then thereafter let the global optimistic_yield() rate decide when to actually yield out of CONT.

earlephilhower: I was suggesting a simplification using your corrected `optimistic_yield` which now simply checks the ESP cycle counter and returns immediately if it's not time.  So the routine becomes
````
inline void WAIT_CLOCK_STRETCH()
{
    esp8266::polledTimeout::oneShotFastUs timeout(twi_clockStretchLimit); 
    while(!timeout && !SCL_READ(twi_scl))   {
        optimistic_yield(5000);  // Every 5ms do a yield(), OTW busy-wait
    }
}
````
The optimistic_yield already implements the same logic as the periodic timeout...

dok-net: @Tech-TX, could you please comment on this? It's my understanding that busy locking for up to 5ms without any `yield()` of unknown duration is important for the HW protocol to work. Once over 5ms delays may occur, because it's not time critical any longer. With the code in master it works like that, my patch to substitute `optimistic_yield(10000)` for `yield()` keeps the loop from calling `yield()` every further iteration, to instead just every 10000µs, letting the loop poll `SCL_READ` unless the 150ms timeout is exceeded.
@earlephilhower Your change has a a subtle difference to that, in that `optimistic_yield(5000)` yield early, depending on how long before `WAIT_CLOCK_STRETCH()` the most recent `yield()` occurred, I fear this breaks the timings.

earlephilhower: This is a master waiting on a possible clock stretch from the slave (i.e. the slave is pulling SCL low after we've written it to high using an open-drain output).  As the master you don't need to start driving the next clock edge until you want to, so there is no issue with yielding every 5ms, as the original code does more directly.

dok-net: @earlephilhower you almost had me, but this location reminded me of something. Yield is a weak binding, IS intended from Arduino AVR core down to be shimmed by multi-tasking libs (let's not get into a discussion about that please w.r.t. FreeRTOS now and here, please), and, but, definitely in ESP8266, yield executes the Schedule coroutines. So, no, yield can be far from a no-op.

dok-net: OTOH, here, yes, per your explanation it doesn't hurt. Given this is the only change in this PR, time has come to withdraw it.

dok-net:
Withdrawing due to review.

PR #6869, opened 2019-12-02
Minor yield() cleanup: optimistic_yield, proper polledTimeout use

dok-net:
Suggest by @devyte, this PR attempts to point out and refactor those places in cores and libraries that appear to be calling yield() more often than needed, which would hurt performance.
Please review carefully.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -431,6 +431,7 @@ int8_t ESP8266WiFiSTAClass::waitForConnectResult(unsigned long timeoutLength) {
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
+    // if probing doesn't trip, this yields
     using oneShotYieldMs = esp8266::polledTimeout::timeoutTemplate<false, esp8266::polledTimeout::YieldPolicy::YieldOrSkip>;
     oneShotYieldMs timeout(timeoutLength); // number of milliseconds to wait before returning timeout error
     while(!timeout) {
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -431,10 +431,9 @@ int8_t ESP8266WiFiSTAClass::waitForConnectResult(unsigned long timeoutLength) {
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    using esp8266::polledTimeout::oneShot;
-    oneShot timeout(timeoutLength); // number of milliseconds to wait before returning timeout error
+    using oneShotYieldMs = esp8266::polledTimeout::timeoutTemplate<false, esp8266::polledTimeout::YieldPolicy::YieldOrSkip>;
+    oneShotYieldMs timeout(timeoutLength); // number of milliseconds to wait before returning timeout error
     while(!timeout) {
-        yield();
         if(status() != WL_DISCONNECTED) {
             return status();
         }
--- a/libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
+++ b/libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
@@ -150,7 +150,7 @@ IPAddress ESP8266WiFiMesh::getStaticIP()
 void ESP8266WiFiMesh::disableStaticIP()
 {
   WiFi.config(0u,0u,0u);
-  optimistic_yield(10000);
+  yield();
   staticIPActivated = false;
 }
 
@@ -185,9 +185,9 @@ void ESP8266WiFiMesh::deactivateAP()
 void ESP8266WiFiMesh::restartAP()
 {
   deactivateAP();
-  optimistic_yield(10000);
+  yield();
   activateAP();
-  optimistic_yield(10000);
+  yield();
 }
 
 ESP8266WiFiMesh * ESP8266WiFiMesh::getAPController()
@@ -331,9 +331,9 @@ bool ESP8266WiFiMesh::latestTransmissionSuccessful()
 void ESP8266WiFiMesh::fullStop(WiFiClient &currClient)
 {
   currClient.stop();
-  optimistic_yield(10000);
+  yield();
   WiFi.disconnect();
-  optimistic_yield(10000);
+  yield();
 }
 
 /**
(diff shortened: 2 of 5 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -455,10 +455,9 @@ int8_t ESP8266WiFiSTAClass::waitForConnectResult(unsigned long timeoutLength) {
     if((wifi_get_opmode() & 1) == 0) {
         return WL_DISCONNECTED;
     }
-    using esp8266::polledTimeout::oneShot;
-    oneShot timeout(timeoutLength); // number of milliseconds to wait before returning timeout error
+    using oneShotYieldMs = esp8266::polledTimeout::timeoutTemplate<false, esp8266::polledTimeout::YieldPolicy::YieldOrSkip>;
```
earlephilhower: I know @devyte loves his templates, but this is one confusing template change.  Looking at the implementation it seems correct, but man...does it feel like its obfuscating things.

devyte: I actually didn't include that premade type for exactly this reason. However, it's a valid use (creating a new polledTimeout variation), and @dok-net isn't tjw first to realize that this particular variation is useful, if a bit... sophisticated. 
It was part of the design choice: if you want such variations, you can use them in exactly this way.

dok-net: I don't share your concern about the 10000µs, because it's open to optimization case-by-case. Changing it globally on the contrary looks dangerous to me.
As a compromise, given that 10000 does seem commonplace, could you agree to making that a default argument value for `optimistic_yield`? It's just gut feeling, but 10ms seems just right for a simple decimal duration.

dok-net: I don't share your concern about the 10000µs, this way it's open to optimization or fixing case-by-case. Introducing a common global on the contrary looks dangerous to me if it gets changed. Unfortunately, default argument isn't available due to extern "C" linkage.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -169,9 +169,9 @@ void ESP8266WiFiMesh::deactivateAP()
 void ESP8266WiFiMesh::restartAP()
 {
   deactivateAP();
-  yield();
+  optimistic_yield(10000);
```
earlephilhower: I'm concerned about the change here.  They're obviously worried that deactivateAP() can be very long and that activateAP() can also be long.

If deactivateAP() takes 9ms, the 1st optimistic_yield() won't fire.  Then, if activateAP() takes 9ms, you will have 18ms between yields() here, whereas before you'd be only 9ms between them.
Before, they'd get 2 yields no matter what, but here, if activateAP() is long, you could

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -315,9 +315,9 @@ bool ESP8266WiFiMesh::latestTransmissionSuccessful()
 void ESP8266WiFiMesh::fullStop(WiFiClient &currClient)
 {
   currClient.stop();
-  yield();
+  optimistic_yield(10000);
```
earlephilhower: Same comment as in ::restartAP()

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
```diff
@@ -359,7 +359,7 @@ transmission_status_t ESP8266WiFiMesh::exchangeInfo(WiFiClient &currClient)
   verboseModePrint("Transmitting");  // Not storing strings in flash (via F()) to avoid performance impacts when using the string.
     
   currClient.print(getMessage() + '\r');
-  yield();
+  optimistic_yield(10000);
```
earlephilhower: A yield here ensures the message actually is sent over the wifi, no?  So making this an optimistic_yield doesn't give the OS time to call the LWIP stack in the SYS context and actually get the work done, it seems.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
dok-net: @earlephilhower Of course you're making up the numbers to prove your case, but your on to something, and it's indeed a good catch that NOT yielding here make keep the firmware from running to actually carry the op to completion. I don't know, also optimistic_yield is best for loops, and I too don't think this will be called in a tight loop. So I'm reverting these two changes.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
dok-net: yes, better safe than sorry.

libraries/ESP8266WiFiMesh/src/ESP8266WiFiMesh.cpp
dok-net: rather err on the safe side, reverting.

dok-net:
A reply to the performance remark you made, @earlephilhower.
This sketch indicates otherwise, I think:
```
void setup()
{
	Serial.begin(74880);
	delay(500);
	Serial.println("opty");
}

void loop()
{
	uint32_t ts = ESP.getCycleCount();
	for (int i = 0; i < 1000000; ++i)
	{
		yield();
	}
	uint32_t ts2 = ESP.getCycleCount();
	Serial.print("1E6 iterations over yield, ms = ");
	Serial.println((ts2 - ts) / ESP.getCpuFreqMHz() / 1000);

	ts = ESP.getCycleCount();
	for (int i = 0; i < 1000000; ++i)
	{
		optimistic_yield(10000);
	}
	ts2 = ESP.getCycleCount();
	Serial.print("1E6 iterations over optimistic_yield(10000), ms = ");
	Serial.println((ts2 - ts) / ESP.getCpuFreqMHz() / 1000);
}
```
Result:
```
1E6 iterations over yield, ms = 6483
1E6 iterations over optimistic_yield(10000), ms = 400
1E6 iterations over yield, ms = 6483
1E6 iterations over optimistic_yield(10000), ms = 400
```

earlephilhower:
Thanks for the actual performance numbers!

earlephilhower (approved):
LGTM, seems like a very minor change with not much impact except in, say, hexdump().

devyte (approved):

PR #6871, opened 2019-12-02
LEAmDNS2 (Host Version)

LaborEtArs:
Completely refactored: An inner class (host) implements the mDNS responder functions for each netif that is added to the (outer) controller class. The API is incompatible to the 'old' LEAmDNS. It is planned to create a API-compatible subclass (later).


d-a-v:
Many thanks for all of your work @LaborEtArs !

PR #6881, opened 2019-12-05
Fix sending headers in #send_P(int, PGM_P, PGM_P, size_t)

jennytoo:
The method #send(int, char*, char*[, size_t])) is a virtual method which
calculates the size of the content then calls #send_P(int, PGM_P, PGM_P,
size_t). This particular implementation of #send_P differs from the other
implementations of #send and #send_P in that it uses #sendContent for
headers and always calls #sendContent_P for contents even when the
contents is not specified.

The method #sendContent is intended for body and prepends the chunksize
in chunk mode but this breaks the HTTP protocol which does not expect a
chunksize prior to the headers.

Fix is simply to do the same thing as all the other methods - call
_currentClient.write and only call #sendContent_P if there is content to
send.


jennytoo:
I don't know if there was a reason this method differed from the other implementations of #send and #send_P, but it resulted in a regression in where `send(200, "text/html", "")` was intended to terminate headers and start content, but was now resulting in a header block with a chunk size prior to the headers, followed by a chunksize of 0 indicating end of body. Rewriting the call to `send(200, "text/html", String(""))` was sufficient to work around it because it forced usage of a different implementation of #send

d-a-v:
What is the previous working version you were using ?

jennytoo:
I'll be honest, I don't know which was the previous working version. It was the one installed as part of Platform IO a couple months ago, and in debugging an issue in a newer installation I saw the ESP8266WebServer had been rewritten to use templates, and saw that the way it was being used resulted in calling the one #send / #send_P method which acted differently from the rest.

919c7535631d60e085f87d2b4c72d05a88651eed looks like the last good version with 36f903443bc1c545289adf36e3f56def5434d219 being the version which introduced the change.

The problem was not a change in #send_P, but the addition of a virtual #send method which matched the calling signature exactly (int, char*, char*) and then called #send_P. The actual code in this #send_P method has always worked this way it looks like - but it should never have worked in practice so either I'm misunderstanding it or no one ever relied on using this method while sending chunked data.

PR #6887, opened 2019-12-07
lwip2 updates: no more git sub-sub-module deps, faster checksum, backlog limitation and other fixes

### LINKED ISSUE #5380: Feature request: limit WiFiServer backlog size
Please limit the _unclaimed chain as listen backlog (http://man7.org/linux/man-pages/man2/listen.2.html)

https://github.com/esp8266/Arduino/blob/cd05bae0e8b69eea94adbf153eca4e7c9b96a901/libraries/ESP8266WiFi/src/WiFiServer.cpp#L164-L171

### LINKED ISSUE #6832: Unable to clone full platform (LWIP2 git offline)
The repository https://git.savannah.nongnu.org/git/lwip.git is offline.
This means the platform cannot be fetched.

Is it possible to define some archive somewhere as backup?
Now it is this external repository, but it may happen to some external repository some other time, so maybe it is good to be able to define some fallback location somewhere?

Since I'm not really sure where this could be done, I have also added a similar issue here: https://github.com/platformio/platform-espressif8266/issues/184

### LINKED ISSUE #6998: Switch to lwip checksum #3 / Discussion of general throughput improvements
### Basic Infos

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[8242d7227198dca2db39cf86d0b34f0e07e634cd]

### Settings in IDE

- Module:  			[Nodemcu]
- Flash Mode:			[unknown]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Higher Bandwidth]
- Reset Method:		[unknown]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]

### Problem Description

Benchmarking lwip_standard_chksum's implementations for a 1450 byte packet with CPU at 160MHz show:

- 46us - implementation 1
- 32us - implementation 2
- 19us - implementation 3

```cpp

#include "lwip/opt.h"

#include "lwip/inet_chksum.h"
#include "lwip/def.h"
#include "lwip/ip_addr.h"

#include <string.h>

/**
 * lwip checksum
 *
 * @param dataptr points to start of data to be summed at any boundary
 * @param len length of data to be summed
 * @return host order (!) lwip checksum (non-inverted Internet sum)
 *
 * @note accumulator size limits summable length to 64k
 * @note host endianess is irrelevant (p3 RFC1071)
 */
u16_t
lwip_standard_chksum1(const void *dataptr, int len)
{
  u32_t acc;
  u16_t src;
  const u8_t *octetptr;

  acc = 0;
  /* dataptr may be at odd or even addresses */
  octetptr = (const u8_t *)dataptr;
  while (len > 1) {
    /* declare first octet as most significant
       thus assume network order, ignoring host order */
    src = (*octetptr) << 8;
    octetptr++;
    /* declare second octet as least significant */
    src |= (*octetptr);
    octetptr++;
    acc += src;
    len -= 2;
  }
  if (len > 0) {
    /* accumulate remaining octet */
    src = (*octetptr) << 8;
    acc += src;
  }
  /* add deferred carry bits */
  acc = (acc >> 16) + (acc & 0x0000ffffUL);
  if ((acc & 0xffff0000UL) != 0) {
    acc = (acc >> 16) + (acc & 0x0000ffffUL);
  }
  /* This maybe a little confusing: reorder sum using lwip_htons()
     instead of lwip_ntohs() since it has a little less call overhead.
     The caller must invert bits for Internet sum ! */
  return lwip_htons((u16_t)acc);
}

/*
 * Curt McDowell
 * Broadcom Corp.
 * csm@broadcom.com
 *
 * IP checksum two bytes at a time with support for
 * unaligned buffer.
 * Works for len up to and including 0x20000.
 * by Curt McDowell, Broadcom Corp. 12/08/2005
 *
 * @param dataptr points to start of data to be summed at any boundary
 * @param len length of data to be summed
 * @return host order (!) lwip checksum (non-inverted Internet sum)
 */
u16_t
lwip_standard_chksum2(const void *dataptr, int len)
{
  const u8_t *pb = (const u8_t *)dataptr;
  const u16_t *ps;
  u16_t t = 0;
  u32_t sum = 0;
  int odd = ((mem_ptr_t)pb & 1);

  /* Get aligned to u16_t */
  if (odd && len > 0) {
    ((u8_t *)&t)[1] = *pb++;
    len--;
  }

  /* Add the bulk of the data */
  ps = (const u16_t *)(const void *)pb;
  while (len > 1) {
    sum += *ps++;
    len -= 2;
  }

  /* Consume left-over byte, if any */
  if (len > 0) {
    ((u8_t *)&t)[0] = *(const u8_t *)ps;
  }

  /* Add end bytes */
  sum += t;

  /* Fold 32-bit sum to 16 bits
     calling this twice is probably faster than if statements... */
  sum = FOLD_U32T(sum);
  sum = FOLD_U32T(sum);

  /* Swap if alignment was odd */
  if (odd) {
    sum = SWAP_BYTES_IN_WORD(sum);
  }

  return (u16_t)sum;
}

/**
 * An optimized checksum routine. Basically, it uses loop-unrolling on
 * the checksum loop, treating the head and tail bytes specially, whereas
 * the inner loop acts on 8 bytes at a time.
 *
 * @arg start of buffer to be checksummed. May be an odd byte address.
 * @len number of bytes in the buffer to be checksummed.
 * @return host order (!) lwip checksum (non-inverted Internet sum)
 *
 * by Curt McDowell, Broadcom Corp. December 8th, 2005
 */
u16_t
lwip_standard_chksum3(const void *dataptr, int len)
{
  const u8_t *pb = (const u8_t *)dataptr;
  const u16_t *ps;
  u16_t t = 0;
  const u32_t *pl;
  u32_t sum = 0, tmp;
  /* starts at odd byte address? */
  int odd = ((mem_ptr_t)pb & 1);

  if (odd && len > 0) {
    ((u8_t *)&t)[1] = *pb++;
    len--;
  }

  ps = (const u16_t *)(const void *)pb;

  if (((mem_ptr_t)ps & 3) && len > 1) {
    sum += *ps++;
    len -= 2;
  }

  pl = (const u32_t *)(const void *)ps;

  while (len > 7)  {
    tmp = sum + *pl++;          /* ping */
    if (tmp < sum) {
      tmp++;                    /* add back carry */
    }

    sum = tmp + *pl++;          /* pong */
    if (sum < tmp) {
      sum++;                    /* add back carry */
    }

    len -= 8;
  }

  /* make room in upper bits */
  sum = FOLD_U32T(sum);

  ps = (const u16_t *)pl;

  /* 16-bit aligned word remaining? */
  while (len > 1) {
    sum += *ps++;
    len -= 2;
  }

  /* dangling tail byte remaining? */
  if (len > 0) {                /* include odd byte */
    ((u8_t *)&t)[0] = *(const u8_t *)ps;
  }

  sum += t;                     /* add end bytes */

  /* Fold 32-bit sum to 16 bits
     calling this twice is probably faster than if statements... */
  sum = FOLD_U32T(sum);
  sum = FOLD_U32T(sum);

  if (odd) {
    sum = SWAP_BYTES_IN_WORD(sum);
  }

  return (u16_t)sum;
}

static unsigned char packet[1450];

void setup() {
  Serial.begin(115200);
  for( unsigned i = 0; i < sizeof packet; ++i )
      packet[i] = rand(); 
}

void loop() {
u32_t time0 = micros();
u16_t res1 = lwip_standard_chksum1(packet, sizeof packet);
u32_t time1 = micros();
u16_t res2 = lwip_standard_chksum2(packet, sizeof packet);
u32_t time2 = micros();
u16_t res3 = lwip_standard_chksum3(packet, sizeof packet);
u32_t time3 = micros();
Serial.print( time1-time0 );Serial.print(",");Serial.print(time2-time1);Serial.print(",");Serial.println(time3-time2);
}

```

### Debug Messages

```
46,33,19 #160MHz CPU
92,64,37 #80 MHz CPU
```

d-a-v:
* upstream lwIP is not sub-sub-moduled anymore (a light `git clone` with proper ref is used instead)
* Allows IPv4 and IPv6 DNS and SNTP server configured via DHCP to co-exist (patch against upstream)

fixes #6832

*edit - other lwip2 submodule updates*:

- Switches to lwIP checksum # 3 (closes #6998 - thanks @rsaxvc for https://github.com/d-a-v/esp82xx-nonos-linklayer/pull/42)
- Fixes `weak` usage (thanks @mcspr for https://github.com/d-a-v/esp82xx-nonos-linklayer/issues/43)
- Allows lwip2 compilation with absolute prefix path (thanks @mcspr for https://github.com/d-a-v/esp82xx-nonos-linklayer/pull/45)
- Enables tcp listen with backlog limitation (closes #5380 - thanks @douniwan5788 for suggesting)
- May be related with https://github.com/esp8266/Arduino/issues/6964

( ***updated 20-02-17*** ) About backlog limitation, `WiFiServer.h`:
```
// lwIP-v2 backlog facility allows to keep memory safe by limiting the
// maximum number of incoming *pending clients*.  Default number of possibly
// simultaneously pending clients is defined in WiFiServer.cpp
// (MAX_PENDING_CLIENTS_PER_PORT=5).  User can overide it at runtime from
// sketch:
//      WiFiServer::begin(port, max-simultaneous-pending-clients);
//
// An "incoming pending" client is a new incoming TCP connection trying to
// reach the TCP server.  It is "pending" until lwIP acknowledges it and
// "accepted / no more pending" when user calls WiFiServer::available().
//
// Before the backlog feature or with lwIP-v1.4, there was no pending
// connections: They were immediately accepted and filling RAM.
//
// Several pending clients can appear during the time when one client is
// served by a long not-async service like ESP8266WebServer.  During that
// time WiFiServer::available() cannot be called.
//
// Note: This *does not limit* the number of *simultaneously accepted
//       clients*.  Such limit management is left to the user.
//
// Thus, when the maximum number of pending connections is reached, new
// connections are delayed.
// By "delayed", it is meant that WiFiServer(lwIP) will not answer to the
// SYN packet until there is room for a new one: The TCP server on that port
// will be mute.  The TCP client will regularly try to connect until success
// or a timeout occurs (72s on windows).
//
// When user calls WiFiServer::available(), the tcp server stops muting and
// answers to newcomers (until the "backlog" pending list is full again).
```


libraries/ESP8266WiFi/src/WiFiServer.cpp
```diff
@@ -77,13 +82,21 @@ void WiFiServer::begin(uint16_t port) {
         return;
     }
 
+#if LWIP_VERSION_MAJOR == 1
+    (void)backlog;
     tcp_pcb* listen_pcb = tcp_listen(pcb);
+#else
+    if (backlog == 0)
```
devyte: instead of mapping 0=>5, how about malling 0=>disabled backlog?

d-a-v: Good question ! (disabling it by default is like setting to max `0xff`)
The goal of the feature is to protect the esp from OOM at worst or heap fragmentation at least when connection requests rate is greater than or around service time rate. My answer would be another question: what would we like by default for the average arduino sketch ?

devyte: I think you had it right before: the default should likely be the define, 5. My question here is about disabling it. 
Option 1: 0 means disabled. You could have a special case to go through a different code path and not do backlog at all. 
Option 2: 0 means max possible. If the arg is 0, replace the value with the max possible backlog number 0xff.
Option 3: 0 as arg is not allowed. A define or constexpr is provided so that the user can set it to th3 max value possible (0xff).

I prefer option 1, but will settle for option 2. I would prefer to avoid option 3.

TD-er: I have a sub-question.  This is about desired behavior when multiple clients access the unit, but what if a connection is not closed by a client?
So this brings to (my) mind the question whether a single counter here is enough?
This may lead to another and bigger change, so my expectation (desire?) is that it will NOT be part of this change, but it may give some ideas for this counter.

Is the WiFi Server, or LWIP2 aware of the continuation of receiving packets by the client?
If so, then there may be a difference in desired behavior when all data was sent but the client simply didn't close the connection, compared to being in the middle of a transfer.
There is also the other way around, when the client deliberately is not waiting for a reply/ack, is the connection then still considered to be an existing one?

So maybe we should accept a (single?) new one, even though the set amount is in use.

d-a-v: In the current state of this PR, the possible number of connections is limited to the max (5 by default).
We can change this behaviour that way:
- we can have an unlimited number of *accepted* connections:
  that would answer to you concern I think,
- ***and*** keep number of *pending* connections limited to 5:
  - if user doesn't care at all and treat one connection at a time (timeouts are meant to close dying clients), then ram usage is protected
  - if user wants to be able to handle many (>>5) connections simultaneously, it is only needed to **stop** checking `WiFiServer::available()` on user constraints.
    From RAM pov, the worst situation would be the already accepted connections, 5 connections in ram waiting to be acknowledged with `WiFiServer.available()` and the other newer requests waiting&retrying outside, not using a single byte of ram.

I'll make some tests and commit. Thanks for your sub-question @TD-er . Until now I was wondering why lwIP's `tcp_backlog_accepted()` was not included in `tcp_close()`. Their documentation on this is very poor. Now with your question, I think I understand why: We can call it before `tcp_close()`, once a connection is accepted.

TD-er: > we can have an unlimited number of accepted connections:
> that would answer to you concern I think,

Well, how many resources do "accepted connections" take?
And how do clients react on "accepted-but-never-answered" connections?
Will they actively close them after some timeout? If not, then the ESP may try to serve to a non existing connection

d-a-v: The issue is about the number of silently accepted connection.

Without backlog feature, there is no limitation and ESP can OOM because of many external requests connecting to a server port. Once connected (this is automatic, internal to lwIP, with a callback to store them in a waiting list) and even when sketch does not `WiFiServer::available()` them, ram is allocated (=> oom) while clients patiently wait for an answer. No timeout in that case (except when clients want to).

With an example: if one starts many simultaneous `telnet <esp-ip> <port>`, a `tcp_pcb` is allocated in esp for every single request. In that case OOM can happen even when `WiFiServer::available()` is called but requests keep coming faster than served and closed.

> Well, how many resources do "accepted connections" take?

some tens of bytes (`sizeof(tcp_pcb)`)

> And how do clients react on "accepted-but-never-answered" connections?
> Will they actively close them after some timeout?

Application defined. `telnet` waits forever.

> If not, then the ESP may try to serve to a non existing connection

There are timeouts on application level (our webserver is 5secs), and also at tcp levels. The latter can be very long (hours) before a connection is closed (that's why there is the tcp-keepalive option, to get a shorter/configurable internal tcp timeout).
On the other hand, if the client closes the connection, ram is released on esp side.

There is also another software way to deal with too many incoming connections which has not been addressed so far because it is not clean:
When the internal lwIP callback is called but the already accepted list is too large, WiFiServer could close the connection. That's bad behaviour. Backlog limitation handles this differently: it does not answer immediately until resources allow to answer properly.

d-a-v: @devyte 
As discussed, 0 for backlog is now a forbidden value which will not open the server port.

PR #6888, opened 2019-12-07
sntp callback: use a recurrent schedule function

d-a-v:
A scheduled function used as callback had to wait for next `loop()` but user may loop on `yield()`.
Using a recurrent scheduled function executed once instead of a simple scheduled function is a fix for this use case.

fixes https://gitter.im/esp8266/Arduino?at=5deb83131659720ca8145e26


hasenradball:
When using `schedule_function(_settimeofday_cb);` and a `delay(100)` for instancs of `yield()` to wait for the cb there is no change!
Do we have understood the root cause completely?

hasenradball:
Here is the test sketch:
```
#include <Arduino.h>
#include <time.h>
#include <coredecls.h>
#include <ESP8266WiFi.h>

const char * _ssid {"ssid"};
const char * _password {"password"};

// ### --- Class Unix_Time --- ###
class Unix_Time {
  private:
    time_t _ti{0};
    static bool _initOK;
    bool _initOK2{false};

  public:
    Unix_Time(void);
    ~Unix_Time(void);

    static void set_initOK(void);
    static bool get_initOK(void);
    void set_initOK2(void);
    bool get_initOK2(void);
    bool Init(void);
    time_t getTime(void);
};
bool Unix_Time::_initOK{false};

Unix_Time::Unix_Time() {
  //  settimeofday_cb([&]() {
  //    Unix_Time::set_initOK();
  //  });

  // for future usage if settimeofday_cb() can handle functional pointers
  settimeofday_cb(std::bind(&Unix_Time::set_initOK2, this));
  configTime(0, 0, "ptbtime1.ptb.de", "ptbtime2.ptb.de", "ptbtime3.ptb.de");
}

Unix_Time::~Unix_Time(void) {}

void Unix_Time::set_initOK(void) {
  _initOK = true;
}

void Unix_Time::set_initOK2(void) {
  _initOK2 = true;
}

bool Unix_Time::get_initOK2(void) {
  return _initOK2;
}

bool Unix_Time::get_initOK(void) {
  return _initOK;
}

time_t Unix_Time::getTime(void) {
  //if (!_initOK) return;
  _ti = time(nullptr);
  return _ti;
}
// ### --- Class Unix_Time --- ###

Unix_Time utc;

void setup() {
  Serial.begin(115200);
  while (!Serial) yield();

  Serial.printf(" WiFi mode set to WIFI_STA: %s\n", WiFi.mode(WIFI_STA) ? "Success!" : "Failed!");
  WiFi.begin(_ssid, _password);

  while ((WiFi.status() != WL_CONNECTED)) {
    delay(100);
    Serial.print(" ");
    Serial.print(WiFi.status());
  }
  if (WiFi.status() == WL_CONNECTED) {
    Serial.println("");
    Serial.print(" Verbunden mit: ");
    Serial.println(WiFi.SSID());
    Serial.print(" IP address: ");
    Serial.print("\n");
  }
  Serial.println(utc.getTime());
  Serial.println(">>> set Clock");
  while (!utc.get_initOK2()) {
    Serial.println("...no cb received");
    Serial.println(utc.getTime());
    delay(1000);
  }
  Serial.print(utc.getTime());
}

void loop() {

}
```

d-a-v:
You may have not applied the PR (which is not merged yet).
Your MCVE works correctly with it.
```
00:02:36.229 ->  IP address: 10.0.1.154
00:02:36.262 -> 
00:02:36.262 -> 28803
00:02:36.262 -> >>> set Clock
00:02:36.262 -> ...no cb received
00:02:36.262 -> 28803
00:02:37.258 -> 1575759756
(((then silent)))
```

hasenradball:
Hi I have tested it with the PR also, then it works, right. But with your description above I understood the rootcause was the yield(). But when using delay(500) for instance without the PR, there is no change.
So I did not understand the effect of of the PR.
Where can I look up the definition of the schedule_function?

hasenradball:
Is a cb as static function (2.5.2) differently invoked as the functional solution as now in 2.6.2?

d-a-v:
In 2.5.2, the function was called from within an ISR-like context that prevented to use most of the API.
It has been changed to a functionnal that runs like a normal/user context. But the call is deferred to the end of next `loop()`.
The current PR allows the call to be made at next `yield()`, `delay()` or `loop()` call.

hasenradball:
Ah now I think I catched the issue. If I had called it in the loop() funktion it would have worked, but when I am calling it in the setup() it can not work by definition, right?

PR #6889, opened 2019-12-07
WiFi: ARP gratuitous API for wifi station mode

### LINKED ISSUE #5998: WiFi dropouts unless using WiFi.setSleepMode(WIFI_NONE_SLEEP)
I've noticed lately that multiple devices I have have started dropping off WiFi and reconnecting (causing MQTT dropouts etc) unless I specify `WiFi.setSleepMode(WIFI_NONE_SLEEP)` when setting up the WiFi adapter.

ESP.getFullVersion() output:
```
SDK:3.0.0-dev(c0f7b44)/Core:2.5.0=20500000/lwIP:STABLE-2_1_2_RELEASE/glue:1.1/BearSSL:6778687
```
WiFi setup code in setup():
```
mConnectHandler = WiFi.onStationModeConnected(onConnected);
mDisConnectHandler = WiFi.onStationModeDisconnected(onDisconnect);
mGotIpHandler = WiFi.onStationModeGotIP(onGotIP);

WiFi.disconnect() ;
WiFi.persistent(false);
WiFi.mode(WIFI_STA);            // Client mode
WiFi.setOutputPower(17);        // 10dBm == 10mW, 14dBm = 25mW, 17dBm = 50mW, 20dBm = 100mW
WiFi.begin(ssid, password);     // Start WiFi.
```
And the functions:
```
void onConnected(const WiFiEventStationModeConnected& event){
update_status += millis();
update_status += ": Connected to AP.";
update_status += "\n";
}
void onDisconnect(const WiFiEventStationModeDisconnected& event){
update_status += millis();
update_status += ": Station disconnected";
update_status += "\n";
client.disconnect();
}
void onGotIP(const WiFiEventStationModeGotIP& event){
update_status += millis();
update_status += ": Station connected, IP: ";
update_status += WiFi.localIP().toString();
update_status += "\n";
timer.setTimeout(1000, checkForUpdate);
}
```
update_status is a global string that is included on the web page interface.

If I comment out the setSleepMode line, I get regular wifi disconnections.

### LINKED ISSUE #6886: ESP8266 Doesn't respond to ARP requests
My ESP8266 regularly stops responding to ARP requests. I raised this in #6873 but I was asked to troubleshoot with the Tasmota Devs first [(Github Issue)](https://github.com/arendst/Tasmota/issues/7098) and provide a MCVE example. Tasmota ruled out an issue with their code, but did identify that this may be an interoperability issue between the ESP8266 and Mikrotik access points.

### Basic Infos

#### Platform

- Hardware:				ESP8266
- Core Version:			2.6.2
- Operating System:		MacOS

### Problem Description

I have discovered if If `wifi_set_sleep_type` is light or modem ARP requests will consistently fail. When set to none I reliably get responses to ARP requests. This appears to affect ESP8266 devices connected to Mikrotik Access Points. 

ARP responses when sleep type set to none:  (1 lost is acceptable IMO)
```
sudo nping --arp-type ARP 10.0.130.109 -c 20
...
SENT (19.0551s) ARP who has 10.0.130.109? Tell 10.0.130.102
RCVD (19.0663s) ARP reply 10.0.130.109 is at 5C:CF:7F:88:E3:2F
...
Raw packets sent: 20 (840B) | Rcvd: 19 (532B) | Lost: 1 (5.00%)
```

ARP responses when sleep type set to light: 
```
sudo nping --arp-type ARP 10.0.130.109 -c 20
...
SENT (1.0130s) ARP who has 10.0.130.109? Tell 10.0.130.102
RCVD (1.0276s) ARP reply 10.0.130.109 is at 5C:CF:7F:88:E3:2F
...
Raw packets sent: 20 (840B) | Rcvd: 5 (140B) | Lost: 15 (75.00%)
```

ARP responses when sleep type set to modem: 
```
sudo nping --arp-type ARP 10.0.130.109 -c 20
...
SENT (1.0130s) ARP who has 10.0.130.109? Tell 10.0.130.102
RCVD (1.0276s) ARP reply 10.0.130.109 is at 5C:CF:7F:88:E3:2F
...
Raw packets sent: 20 (840B) | Rcvd: 3 (84B) | Lost: 17 (85.00%)
```

```
#include <ESP8266WiFi.h>

#ifndef STASSID
#define STASSID [YOUR SSID]
#define STAPSK  [YOUR PASSWORD]
#endif

const char* ssid     = STASSID;

void setup() {
  Serial.begin(115200);

  Serial.println();
  Serial.println();
  Serial.print("Connecting to ");
  Serial.println(ssid);

  wifi_set_sleep_type(NONE_SLEEP_T);
  //wifi_set_sleep_type(LIGHT_SLEEP_T);
  //wifi_set_sleep_type(MODEM_SLEEP_T);

  WiFi.mode(WIFI_STA);
  WiFi.begin(ssid, password);

    delay(500);
    Serial.print(".");
  }

  Serial.println("");
  Serial.println("WiFi connected");
  Serial.println("IP address: ");
}

void loop() {
  Serial.println("I'm alive");
  delay(30000);
}
```

d-a-v:
Based on multiple reports and examples

fixes #6886 
fixes #5998 (together with #6484, hopefully)


TD-er:
Just looking at the code and I think this is probably not optimal to use.

```c++
bool ESP8266WiFiSTAClass::stationKeepAliveEnabled ()
{
    return _keepStationAliveUs != 0;
}

void ESP8266WiFiSTAClass::stationKeepAliveStop ()
{
    _keepStationAliveUs = -1;
    // will be set to 0 at recurrent call
}

bool ESP8266WiFiSTAClass::stationKeepAliveSetupMs (int ms)
{
    if (_keepStationAliveUs != 0 || ms <= 0)
        return false;
[...]
```

So on other words, you can only change the value by calling stop and then wait for the next scheduled interval before you can even set a new value.
Worst case scenario is to wait for as long as the previous set interval.
It would make sense to set the interval to anything > 0 and if you want to stop it, call the stop function.
Now you just have to poll as long as you receive false to set a new interval.

Also I do miss to send (or restart the scheduler) when a wifi connection to the AP is established.
That's the most important one to send these ARP packets. If it was set to for example 10 seconds and you make a connection right after the scheduler tried sending these ARP packets, you may be unable to make a connection to another host, for about 10 seconds after successful wifi connection was established.

d-a-v:
@TD-er please review the API change

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -704,7 +711,52 @@ int32_t ESP8266WiFiSTAClass::RSSI(void) {
...
+        // mark disabled (0) if user cancelled it (<0)
+        _keepStationAliveUs = 0;
+}
 
+void ESP8266WiFiSTAClass::stationKeepAliveSetIntervalMs (int ms)
+{
+    bool wasEnabled = (_keepStationAliveUs > 0);
+    int us = ms * 1000;
+    if (wasEnabled)
+        _keepStationAliveUs = ms <= 0? /*disable*/-1: /*update*/us;
```
TD-er: Maybe you should also check for negative values of us (or ms > (max int  / 1000) )

d-a-v:
@TD-er the API has changed, based on your use case. Setting the interval will:
- cancel everything that was planned
- immediately send a packet
- schedule the next ones based on the new interval

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -704,7 +711,61 @@ int32_t ESP8266WiFiSTAClass::RSSI(void) {
...
+               (interface->flags & NETIF_FLAG_LINK_UP)
+            && (interface->flags & NETIF_FLAG_UP)
+#if LWIP_VERSION_MAJOR == 1
+            && interface == eagle_lwip_getif(STATION_IF) /* lwip1 does not set if->num properly */
+            && (!ip_addr_isany(&interface->ip_addr))
+#else
+            && interface->num == STATION_IF
+            && (!ip4_addr_isany_val(*netif_ip4_addr(interface)))
+#endif
+        )
```
TD-er: Just curious.
A lot of code examples (e.g. in AddrList.h) iterate over the `addrList` to see if one got a valid IP-address.
The code here does iterate over the interfaces.

What is the better approach?
I got the feeling one of my functions using the examples in AddrList.h does sometimes yield a false positive result when looping over the addrList, so that's why I am asking.

d-a-v: `AddrList` is meant to iterate over IP addresses (multiple per lwIP interfaces with IPv6), implicitly over all interfaces. But there's always only one IPv4 address per interface.

Here we only need to iterate over the (and once per) interface/s to announce the mac address  which is unique per lwIP interface. Checking IPv4 is sufficient to know whether the interface needs ARP announce.

When IPv6 is not enabled, looping with `AddrList` is like walking through `netif_list`.

marrold:
Novice question, is there a simple way I can test this branch in platformio?

d-a-v:
If you are using git for this repository in your platformIO setup, then yes.
You first need to checkout this PR ([help with this](https://gist.github.com/d-a-v/801cdfcf56d813dca2ff7b23f620c9d3)) then rebuild using PIO.

Jason2866:
Sorry for replying sooo late. We had no equipment until now to reproduce the issue (of not beeing reachable via http after a while) in our wifi setups. Since a few days i have a OpenWRT AP which has this issue...
@s-hadinger  implemented your changes as backport in a Tasmota test version https://github.com/s-hadinger/Tasmota/tree/arp_dev
Using this version (which sends every minute Gratiuitous ARP) device keeps beeing reachable.
The issue is solved (at least for OpenWRT)

It would be great if you merge this PR that all users can benefit from this option to solve issues
of devices not reachable because ARP seems to get lost if no Gratiuitous ARP is done.

s-hadinger:
I confirm this feature is useful with some wifi routers. I ported this feature in Tasmota, and it would be great for it to be integrated natively in Arduino Core. Some other projects might benefit from it.

d-a-v:
Per internal review comment:
- use a Ticker, but with adding a new ticker method using recurrent scheduled function once so:
  when tick is fired, function is executed once at next `yield()` (instead of at next `loop()`)
  This is needed to keep regularity the most we can in cont space.
- move to a new `experimental::ESP8266WiFiGratuitous` class

d-a-v:
@ivankravets @platformio I need help with [this issue](https://travis-ci.org/github/esp8266/Arduino/jobs/672718343#L271):

Following change does not help (because in that case arduino does not compile `Ticker.cpp`):
```
-#include <Ticker.h>
+#include "../../Ticker/src/Ticker.h" // instead of <Ticker.h> (platformio can't find it otherwise)
```

This `Ticker` is ours and does not appear in https://platformio.org/lib/search?query=header:Ticker.h .
This is a case where a local library uses another local library.
This is [not a problem](https://travis-ci.org/github/esp8266/Arduino/builds/672718342) with arduino.

d-a-v:
Moving `#include <Ticker.h>` from `.cpp` to `.h` didn't help

valeros:
Hi @d-a-v ! Looks like these changes are going to affect a lot of PlatformIO users since `Library Dependency Finder` in default mode doesn't look for dependencies in source files. Probably an optimal solution would be to add `depends` field to the `library.properties` file for `ESP8266WiFi` library, e.g.:
```
name=ESP8266WiFi
version=1.0
author=Ivan Grokhotkov
maintainer=Ivan Grokhtkov <ivan@esp8266.com>
sentence=Enables network connection (local and Internet) using the ESP8266 built-in WiFi.
paragraph=With this library you can instantiate Servers, Clients and send/receive UDP packets through WiFi. The shield can connect either to open or encrypted networks (WEP, WPA). The IP address can be assigned statically or through a DHCP. The library can also manage DNS.
category=Communication
url=
architectures=esp8266
dot_a_linkage=true
depends=Ticker
```

s-hadinger:
What about just leaving the `stationKeepAliveNow ()` function and leaving the timer on the client side? In Tasmota that's easy to add a timer for that, and it would avoid bringing yet another library to projects that don't need it.

d-a-v:
@valeros good idea but as @s-hadinger says, it may include Ticker for nothing if gratuitous is not used (unless it is moved in a separate directory ?).
Well, I chose to short-circuit the library and use the core api instead.

PR #6897, opened 2019-12-10
netifWrapper hostname() does not return hostname.

TD-er:
Not even sure why this was accepted by the compiler in the first place.

### DIFF
--- a/cores/esp8266/AddrList.h
+++ b/cores/esp8266/AddrList.h
@@ -124,7 +124,7 @@ struct netifWrapper
 
     // common to all addresses of this interface
     String ifname () const          { return String(_netif->name[0]) + _netif->name[1]; }
-    const char* ifhostname () const { return _netif->hostname?: emptyString.c_str(); }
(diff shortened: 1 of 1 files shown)


mcspr:
Having noticed this too some time ago... that's a proper syntax:
https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html
https://en.wikipedia.org/wiki/Elvis_operator

cores/esp8266/AddrList.h
```diff
@@ -124,7 +124,7 @@ struct netifWrapper
 
     // common to all addresses of this interface
     String ifname () const          { return String(_netif->name[0]) + _netif->name[1]; }
-    const char* ifhostname () const { return _netif->hostname?: emptyString.c_str(); }
```
d-a-v: [It is valid and returns the tested value if not '0'](https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html#Conditionals)

TD-er:
OK, I didn't know the Elvis operator. 
That's a great name for it :)

So this is a non-issue apparently.

PR #6894, opened 2019-12-09
put inline helpers in IRAM for slave ISRs [issue 6875]

Tech-TX:
https://github.com/esp8266/Arduino/issues/6875 describes exceptions when I2C hits a slave during OTA download.  Issue is caused by the 'inline helpers' not being in IRAM, and all 6 are used in the slave section ISRs.

PR #6898, opened 2019-12-11
change to make inline helpers truly inline [issue 6875]

Tech-TX:
heavily modified to insure that the 'inline helpers' are truly inline (they are, verified in the dump)

### DIFF
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -329,7 +329,7 @@ bool UpdaterClass::_writeBuffer(){
   bool modifyFlashMode = false;
   FlashMode_t flashMode = FM_QIO;
   FlashMode_t bufferFlashMode = FM_QIO;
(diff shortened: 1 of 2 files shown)


Tech-TX:
Did you want me to kill the other pull request https://github.com/esp8266/Arduino/pull/6894
as this one corrects the same issue in a different way?

devyte:
@Tech-TX yes please, let's keep this one.

Tech-TX:
Latest commit added devyte's comment about not inlining reply(), removed the inline attribute from 2 public functions in the slave side, and moved twi_scl_valley() out of the middle of the slave code and into the master area, as it's a master-only function.

Tech-TX:
As only one person has seen this error so far, would you mind waiting until Sunday evening or later to merge it?  I'd like to run every test against it that I can think of, since the changes touched nearly every lowest-level function.  I've eyeballed all of the changes 3 ways, but I'm the eternal pessimist.

PR #6902, opened 2019-12-12
Schedule.h|cpp refactored to class templates for performance and reuse

dok-net:
The interrupt/signal safe linked list in `Schedule.h|cpp` can very well be used as a universal queue or event delegate (c# terminology ;-) ) class.
This PR also supersedes `cores/esp8266/CallbackList.h`.

The new `Delegate.h` class template provides **compelling** performance improvements in cases where `std::function` is repeatedly constructed during execution, which is quite slow. While a simple seek&replace drop-in for `std::function` in APIs, call sites can in turn be refactored; basically, the binding via lambda capture or `std::bind` is substituted for by using the C-function pointer style `Delegate` constructor's `obj` argument.

### Using C function pointer or `std::function`
One assignment and one call via `bool(*f)(Foo*, int)` : 108 cycles; call only: 31 cycles.
One assignment and one call via `std::function<bool(int)>` : 1360 cycles; call only: 34 cycles.

### Using `Delegate<bool(int), Foo*>`
One assignment from bound C++-style Callable and one call: 1435 cycles; call only: 46 cycles.
One assignment from C-function pointer plus obj, and one call: 151 cycles; call only: 30 cycles.


hreintke:
@dok-net :
How should I use MultiDelegate as replacement for CallBackList ?

dok-net:
@d-a-v, PR #6888 would save approximately 1090 CPU cycles overall in
`cores/esp8266/sntp-lwip2.cpp`'s
```diff
diff --git a/cores/esp8266/sntp-lwip2.cpp b/cores/esp8266/sntp-lwip2.cpp
--- a/cores/esp8266/sntp-lwip2.cpp
+++ b/cores/esp8266/sntp-lwip2.cpp
@@ -463,7 +463,7 @@ int settimeofday(const struct timeval* tv, const struct timezone* tz)
         sntp_set_system_time(tv->tv_sec);
 
         if (_settimeofday_cb)
-            schedule_recurrent_function_us([](){ _settimeofday_cb(); return false; }, 0);
+            schedule_recurrent_function_us({ [](void*) { _settimeofday_cb(); return false; }, nullptr }, 0);
     }
     return 0;
 }
```

dok-net:
@hreintke I've tried for a few minutes to wrap `Multidelegate` as the `experimental::CBListImplementation::CallBackList`, but I think there's no need to mimic the interface exactly - right?
To get started on what you functionally need, I believe the two specializations used in `Schedule.h|cpp` provide that right match - execute exactly once, or forever until some condition applies, upon which the callback signals via it's `false` return value. So instead of resetting the `std::shared_ptr` returned from `add`, give the actual `Delegate `instances access to a corresponding predicate and have them return that.

```
struct SomeClass {
    void doSomething(int) const {};
    bool active = true; // for event handling sample
};
```
```
SomeClass obj;
```
```
// the function type void(int) passes a int value to all callbacks:
using CallbackOnceFuncT = Delegate<void(int), SomeClass>;
using CallOnceQueue = MultiDelegate< CallbackOnceFuncT, true >;
CallOnceQueue q;
q.add([obj](int i) { obj.doSomething(i); }); // construction very expensive.
// fast add, 712 fewer CPU cycles:
q.add({ [](SomeClass obj, int i) { obj.doSomething(i); }, obj });
```

```
// the function type bool(int) passes a int value to all event handlers.
// return false from the event handler to get it removed from the event multiplexer:
using EventHandlerFuncT = Delegate<bool(int), SomeClass>;
using EventHandlers = MultiDelegate< EventHandlerFuncT >;
EventHandlers ev;
ev.add([obj](int i) { obj.doSomething(i); return obj.active; }); // construction very expensive.
// fast add, 670 fewer CPU cycles:
ev.add({ [](SomeClass obj, int i) { obj.doSomething(i); return obj.active; }, obj });
```

hreintke:
@dok-net : 
First impression, did not get into all details of your remark.

Main functionality of CallBacklist is that it prevents executing the callback when the Object that registered the callback is already destructed.
Because of that, the deletion from the list cannot be dependent on returning false.
When the returned  `CallBackHandler add(cbFunctionT af, bool ad = true)` goes out of scope, the callback is deleted before executing it.

Do not (yet) see that possibility in (Multi)Delegate.
Maybe the `std::list<CallBackHandler> callBackEventList;` can be replaced my MultiDelegate but need to get more in the details to be sure.

PR #6930, opened 2019-12-21
Check eboot.elf is newer than any files in the dir

earlephilhower:
Eboot is prebuilt manually, and any changes to the bootloader source
code require a manual recompile and git commit to take effect.

Ensure that eboot.elf commit was after (or at same time) as every other
file in the bootloader directory as part of the CI build process.


earlephilhower:
This initial commit should fail because I have made a change to eboot.c (remove unneeded extern that doesn't actually exist).  I'll then commit a new eboot.elf and it should pass.

earlephilhower:
Failing job:
https://travis-ci.org/esp8266/Arduino/jobs/627986977?utm_medium=notification&utm_source=github_status

Issue #2524, opened 2016-09-16
Run i2c on < 100 kHz


vlast3k:
#### Hardware

Hardware:       ESP-12F
Core Version:       github-current
### Description

I wanted to try running I2C on a lower speed, e.g. 20khz, as i have a sensor that claims it needs <25khz. (though it works pretty well on 100 khz as well). But then i noticed that the minimum speed is limited to 100 khz as per

```
void twi_setClock(unsigned int freq){
#if F_CPU == FCPU80
  if(freq <= 100000) twi_dcount = 19;//about 100KHz
  else if(freq <= 200000) twi_dcount = 8;//about 200KHz
  else if(freq <= 300000) twi_dcount = 3;//about 300KHz
  else if(freq <= 400000) twi_dcount = 1;//about 400KHz
  else twi_dcount = 1;//about 400KHz
```

is there any reason for this? 
Else i could check the twi_dcount for ~25khz and compile with it.

devyte:
@vlast3k I still see the same code in latest git. The reason seems to be "standard speed", per [this](https://www.i2c-bus.org/speed/). In other words, if your device specifies a lower speed, it's apparently not compliant.
Still, that is supposed to be a theoretical max. The timing is supposed to be dicated by the clock, not the standard, and the clock is sent by the master. In noisy envs, I don't see harm in working at a lower speed.
Did you ever find the value for 25KHz?

teliot:
I have to lower the I2C clock for SMBus to work

https://www.totalphase.com/support/articles/200349186-Differences-between-I2C-and-SMBus
https://en.wikipedia.org/wiki/System_Management_Bus
http://smbus.org/specs/

```void twi_setClock(unsigned int freq){
#if F_CPU == FCPU80
  if(freq <= 50000) twi_dcount = 52;      //about 40KHz
  else if(freq <= 100000) twi_dcount = 19;//about 100KHz
  else if(freq <= 200000) twi_dcount = 8; //about 200KHz
  else if(freq <= 300000) twi_dcount = 3; //about 300KHz
  else if(freq <= 400000) twi_dcount = 1; //about 400KHz
  else twi_dcount = 1;//about 400KHz
#else```

Tech-TX:
[...]

PR #6934, opened 2019-12-22
Issue2524 allow clocks below 100KHz

Tech-TX:
Corrects the fixed I2C bus clock selections and allows slower bus clocks down to ridiculously low bus speeds if you need that.  Originally prompted by Earle's comment in https://github.com/esp8266/Arduino/issues/2524 about computing the bus clock timing.

Additionally, the dead Twi::stop was removed.  The compiler was removing it in slave compiles anyways, as it wasn't called by any of the functions except for one oversight in the slave state machine.  Slaves can't issue STOP commands.  Although the Twi::stop was listed as an exposed function, it wasn't linked from Wire.h, so it was optimized out.

### DIFF
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -160,6 +160,9 @@ static Twi twi;
 
 void Twi::setClock(unsigned int freq)
 {
+	if (freq < 250)  // minimum freq 250Hz to prevent overflow @ 232Hz
+		freq = 250;
+	
 	preferred_si2c_clock = freq;
 
 #if F_CPU == FCPU80
@@ -225,11 +228,6 @@ void Twi::enableSlave()
     {
 		asm("nop"); // minimum element to keep GCC from optimizing this function out.
     }
-
-//    {
-//        esp8266::polledTimeout::oneShotFastNs timeout(v);
-//        while(!timeout) { // 
-//    }
 }
 
 bool Twi::write_start(void)
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -265,7 +265,7 @@ bool Twi::write_bit(bool bit)
     {
         SDA_LOW(twi_sda);
     }
-    busywait(twi_dcount + 1);
+    busywait(twi_dcount);
     SCL_HIGH(twi_scl);
     WAIT_CLOCK_STRETCH();
     busywait(twi_dcount);
@@ -276,7 +276,7 @@ bool Twi::read_bit(void)
 {
     SCL_LOW(twi_scl);
     SDA_HIGH(twi_sda);
-    busywait(twi_dcount + 2);
+    busywait(twi_dcount);
     SCL_HIGH(twi_scl);
     WAIT_CLOCK_STRETCH();
     bool bit = SDA_READ(twi_sda);
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -225,6 +225,11 @@ void Twi::enableSlave()
     {
 		asm("nop"); // minimum element to keep GCC from optimizing this function out.
     }
+
+//    {
+//        esp8266::polledTimeout::oneShotFastNs timeout(v);
+//        while(!timeout) { // 
+//    }
 }
 
 bool Twi::write_start(void)
@@ -265,7 +270,7 @@ bool Twi::write_bit(bool bit)
     {
         SDA_LOW(twi_sda);
     }
-    busywait(twi_dcount);
+    busywait(twi_dcount + 1);
     SCL_HIGH(twi_scl);
     WAIT_CLOCK_STRETCH();
     busywait(twi_dcount);
@@ -276,7 +281,7 @@ bool Twi::read_bit(void)
 {
     SCL_LOW(twi_scl);
     SDA_HIGH(twi_sda);
-    busywait(twi_dcount);
+    busywait(twi_dcount + 2);
     SCL_HIGH(twi_scl);
     WAIT_CLOCK_STRETCH();
     bool bit = SDA_READ(twi_sda);
--- a/cores/esp8266/core_esp8266_si2c.cpp
+++ b/cores/esp8266/core_esp8266_si2c.cpp
@@ -63,7 +63,7 @@ class Twi
 {
 private:
     unsigned int preferred_si2c_clock = 100000;
-    unsigned char twi_dcount = 18;
+    uint32_t twi_dcount = 18;
     unsigned char twi_sda = 0;
     unsigned char twi_scl = 0;
     unsigned char twi_addr = 0;
@@ -112,7 +112,7 @@ private:
     bool _slaveEnabled = false;
 
     // Internal use functions
-    void ICACHE_RAM_ATTR busywait(unsigned char v);
+    void ICACHE_RAM_ATTR busywait(unsigned int v);
     bool write_start(void);
     bool write_stop(void);
     bool write_bit(bool bit);
@@ -148,7 +148,6 @@ public:
     void attachSlaveRxEvent(void (*function)(uint8_t*, size_t));
     void attachSlaveTxEvent(void (*function)(void));
     void ICACHE_RAM_ATTR reply(uint8_t ack);
-    void ICACHE_RAM_ATTR stop(void);
     void ICACHE_RAM_ATTR releaseBus(void);
     void enableSlave();
 };
@@ -161,65 +160,22 @@ static Twi twi;
 
 void Twi::setClock(unsigned int freq)
 {
-    preferred_si2c_clock = freq;
+	preferred_si2c_clock = freq;
+
 #if F_CPU == FCPU80
-    if (freq <= 50000)
-    {
-        twi_dcount = 38;  //about 50KHz
-    }
-    else if (freq <= 100000)
-    {
-        twi_dcount = 19;  //about 100KHz
-    }
-    else if (freq <= 200000)
-    {
-        twi_dcount = 8;  //about 200KHz
-    }
-    else if (freq <= 300000)
-    {
-        twi_dcount = 3;  //about 300KHz
-    }
-    else if (freq <= 400000)
-    {
-        twi_dcount = 1;  //about 400KHz
-    }
-    else
-    {
-        twi_dcount = 1;  //about 400KHz
-    }
+
+    if (freq > 400000)
+	    freq = 400000;
+    twi_dcount = (500000000 / freq);  // half-cycle period in ns
+    twi_dcount = (1000*(twi_dcount - 1120)) / 62500;  // (half cycle - overhead) / busywait loop time 
+	
 #else
-    if (freq <= 50000)
-    {
-        twi_dcount = 64;  //about 50KHz
-    }
-    else if (freq <= 100000)
-    {
-        twi_dcount = 32;  //about 100KHz
-    }
-    else if (freq <= 200000)
-    {
-        twi_dcount = 14;  //about 200KHz
-    }
-    else if (freq <= 300000)
-    {
-        twi_dcount = 8;  //about 300KHz
-    }
-    else if (freq <= 400000)
-    {
-        twi_dcount = 5;  //about 400KHz
-    }
-    else if (freq <= 500000)
-    {
-        twi_dcount = 3;  //about 500KHz
-    }
-    else if (freq <= 600000)
-    {
-        twi_dcount = 2;  //about 600KHz
-    }
-    else
-    {
-        twi_dcount = 1;  //about 700KHz
-    }
+
+	if (freq > 800000)
+	    freq = 800000;
+    twi_dcount = (500000000 / freq);  // half-cycle period in ns
+    twi_dcount = (1000*(twi_dcount - 560)) / 31250;  // (half cycle - overhead) / busywait loop time
+
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


Tech-TX:
BTW, here's a variation on the Master_Writer example so you can knob the bus clock from the serial terminal.  Enter your desired speed followed by enter, and it changes the Wire.setClock(freq) immediately.  It also displays the half-cycle period and how many loops busywait() will execute.

```cpp
#include <Wire.h>
#include <ESP8266WiFi.h>
#include <PolledTimeout.h>
#define SDA_PIN 4
#define SCL_PIN 5
const int16_t I2C_MASTER = 0x42;
const int16_t I2C_SLAVE = 0x08;
String i2c_clock = "";   // stores the command request

void setup() {
  Serial.begin(74880);           // start serial for output
  Serial.println();
  Serial.println("Starting...");
  Wire.begin(SDA_PIN, SCL_PIN, I2C_MASTER); // join i2c bus (address optional for master)
  Wire.setClock(100000L);
  WiFi.persistent(false);
  WiFi.mode(WIFI_OFF);
  delay(100);
  WiFi.forceSleepBegin();
  yield();

}

byte x = 0;

void loop() {
  using periodic = esp8266::polledTimeout::periodicMs;
  static periodic nextPing(100);

  if (nextPing) {
    Wire.beginTransmission(I2C_SLAVE); // transmit to device #8
    Wire.write("x is ");        // sends five bytes
    Wire.write(x);              // sends one byte
    Wire.endTransmission();    // stop transmitting

//    Serial.println (x);

    x++;
  }
    if (Serial.available()) {
    i2c_clock = Serial.readStringUntil('\n');  // enter the I2C bus speed in the serial monitor
    Serial.print("bus clock = ");
    Serial.println(i2c_clock);
    int freq = (i2c_clock.toInt());
    int period_nS = (500000000/freq);  // half-cycle period in ns
    Serial.print("period = ");
    Serial.print(period_nS);
    Serial.println("ns");
#if F_CPU == FCPU80
    int twi_dcount = (1000*(period_nS - 1120)) / 62500;  // (half cycle period - overhead) / busywait loop time 
#else
    int twi_dcount = (1000*(period_nS - 560)) / 31250;  // (half cycle period - overhead) / busywait loop time 
#endif
    Serial.print("busywait = ");
    Serial.print(twi_dcount);
    Serial.println(" loops");
    Serial.println();
    Wire.setClock(freq);
    }
}
```

cores/esp8266/core_esp8266_si2c.cpp
```diff
@@ -262,18 +218,18 @@ void Twi::enableSlave()
...
 {
     unsigned int i;
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
-    unsigned int reg;
-    for (i = 0; i < v; i++)
+    for (i = 0; i < v; i++)  // loop time is 5 machine cycles: 31.25ns @ 160MHz, 62.5ns @ 80MHz
     {
-        reg = GPI;
+		asm("nop"); // minimum element to keep GCC from optimizing this function out.
```
devyte: This will drift when there are interrupts happening in the middle of the loop. The effective wait doesn't take into account the time spent in the ISRs.
That is no different than what is happening in tje current code, but it is a potential point of failure. 
We discussed the polledTimeout approach, which is commented out below. I don't remember the conclusion about it anymore, was it just a bit slower, i. e. lower max clock speed?

Tech-TX: Quite a bit slower; the minimum granularity of oneShotFastNs was around 200ns, which limits the top speed to 379KHz @80 and 658KHz @160.  Most people are using either 100KHz or 400KHz, very few faster or slower than that.  If you look at one of the earlier commits, I'd tried it and forgot and left parts of it commented out in the commit.  If an interrupt hits, the actual clock period will be longer with a fixed busywait, but it's a static bus (purely edge-driven) so the slave will ride it out.  Best guess is interrupts are only 2-5% of the time even with busy WiFi traffic.  I clocked the WiFi services at anywhere from low microseconds to 2.5 milliseconds range.  Low microseconds won't effect the 5us period of 100KHz much, but the longer ones will.  

Hmmm.  Side issue for later: those occasional 2.5ms WiFi services could cause the I2C slave to timeout and abort, as it has a 10 millisecond timeout.  I don't know if ets_timer_arm_new() is higher priority than the WiFi events.

Here's the difference due to the granularity for the most common clock speeds:

requested bus clock = 50000 (currently needed for the ESP slave)
actual frequency using busywait    = 50012
actual frequency using polledTimer = 50200

requested bus clock = 100000
actual frequency using busywait    = 100050
actual frequency using polledTimer = 100806

requested bus clock = 400000
actual frequency using busywait    = 400801
actual frequency using polledTimer = 431034

requested bus clock = 800000
actual frequency using busywait    = 803212
actual frequency using PolledTimer = 657894

Below 100KHz the larger granularity doesn't have as much of an effect, but I've only seen a few people running slow.  Personally, I'd rather have the accuracy above 100KHz and the 800KHz top speed, but it's your call.

Tech-TX: One other thing I just thought of: I suspect the compiler is putting the busywait() inline, so the shorter routine I did accounts for the 1460 bytes of IRAM saved over the current implementation.  I'll have to go back and retest with the polledTimer to see what that does to the IRAM.  It's a bigger function, so the compiler SHOULD leave it in IRAM and do long calls to it.

Tech-TX: The polledTimer uses 40 bytes more IRAM and 16 bytes less Flash,

with busywait()

IROM   : 239308          - code in flash
IRAM   : 27596   / 32768 - code in IRAM
DATA   : 1264  )         - initialized variables
RODATA : 888   ) / 81920 - constants
BSS    : 25464 )         - zeroed variables

with polledTimer()

IROM   : 239292          - code in flash
IRAM   : 27636   / 32768 - code in IRAM
DATA   : 1264  )         - initialized variables
RODATA : 888   ) / 81920 - constants
BSS    : 25464 )         - zeroed variables

Tech-TX: I was wrong on the minimum delay from the polledTimer a little: the minimum is 214ns at 160, and 506ns at 80.  That means the maximum achievable speed is 328KHz @ 80 or 648KHz @ 160, quite short of the 400/800 I got with the modified busywait().

Here's the measured differences for desired vs. actual bus frequency (the numbers a couple of posts back were calculated):
at 80MHz CPU,
50KHz = 52.4
100KHz = 108
400KHz = 328

at 160MHz CPU,
50KHz = 50.74
100KHz = 103
400KHz = 429
800KHz = 648

Tech-TX: Here's the code difference for the polledTimer:
```cpp
void Twi::setClock(unsigned int freq)
{
    if (freq < 250)  // minimum freq 250Hz
        freq = 250;
    
    preferred_si2c_clock = freq;

#if F_CPU == FCPU80

    if (freq > 300000)
        freq = 300000;
    twi_dcount = (500000000 / freq);  // half-cycle period in ns
    twi_dcount = (twi_dcount - (1120 + 506));  // (half cycle - (overhead + polledTimer minimum) 
    
#else

    if (freq > 630000)
        freq = 630000;
    twi_dcount = (500000000 / freq);  // half-cycle period in ns
    twi_dcount = (twi_dcount - (560 + 214));  // (half cycle - (overhead + polledTimer minimum)

#endif
}


 void ICACHE_RAM_ATTR Twi::busywait(unsigned int v)
{
	esp8266::polledTimeout::oneShotFastNs timeout(v);
	while(!timeout) {}
}
```

devyte: I think polledTimeout probably needs further optimization and/or specific simplified policies to achieve a similar level of performance as your nop loop. Example: skip some checks. 
That requires further work in polledTimeout itseld, which is outside the scope of this fix. Let's handle that as a separate issue.
Your proposal here is a strict improvement over the current code, despite having the same drift flaw. I'm ok with that.

devyte:
> Might want to limit the lower end even further

I think a 1KHz lower limit is ok, I somehow can't picture I2C at such a low clock speed lol.

cores/esp8266/core_esp8266_si2c.cpp
```diff
@@ -161,65 +160,25 @@ static Twi twi;
 
 void Twi::setClock(unsigned int freq)
 {
+    if (freq < 250)  // minimum freq 250Hz to prevent overflow @ 232Hz
```
devyte: As discussed, this probably should be limited to 1KHz.

cores/esp8266/core_esp8266_si2c.cpp
```diff
@@ -518,35 +472,24 @@ void ICACHE_RAM_ATTR Twi::reply(uint8_t ack)
...
     {
         //TWCR = _BV(TWEN) | _BV(TWIE) | _BV(TWINT);
-        SCL_HIGH(twi.twi_scl);		// _BV(TWINT)
-        twi_ack = 0;	// ~_BV(TWEA)
+        SCL_HIGH(twi.twi_scl);      // _BV(TWINT)
+        twi_ack = 0;    // ~_BV(TWEA)
     }
 }
 
-void ICACHE_RAM_ATTR Twi::stop(void)
```
devyte: Why is this stop() being removed?
I know we discussed it, but I can't remember the reason at the moment. Let's leave a comment in this review to document the reason.

cores/esp8266/core_esp8266_si2c.cpp
```diff
@@ -112,7 +112,7 @@ class Twi
     bool _slaveEnabled = false;
 
     // Internal use functions
-    void ICACHE_RAM_ATTR busywait(unsigned char v);
+    void ICACHE_RAM_ATTR busywait(unsigned int v);
```
devyte: Comment of intent:
Strictly speaking, this is a signature change, and as such it could potentially be a breaking change.
However, anyone using the (very new and supposedly internal) Twi class directly deserves to get yelled at, so let's keep it as proposed.

Tech-TX: busywait()  is a private function,  so it shouldn't change the signature

cores/esp8266/core_esp8266_si2c.cpp
Tech-TX: I'm good with 1Khz, tested to 250Hz just in case.  Below that at some point the slave will timeout

cores/esp8266/core_esp8266_si2c.cpp
Tech-TX: My first comment at top: it's a dead function that's optimized out,  verified in the dump.  It was a left-over from the AVR master code,  and has no function in an I2C slave.

Tech-TX: here's the relevant part of twi.h in the cores directory:
```cpp
void twi_init(unsigned char sda, unsigned char scl);
void twi_setAddress(uint8_t);
void twi_stop(void);
void twi_setClock(unsigned int freq);
void twi_setClockStretchLimit(uint32_t limit);
uint8_t twi_writeTo(unsigned char address, unsigned char * buf, unsigned int len, unsigned char sendStop);
uint8_t twi_readFrom(unsigned char address, unsigned char * buf, unsigned int len, unsigned char sendStop);
uint8_t twi_status();

uint8_t twi_transmit(const uint8_t*, uint8_t);

void twi_attachSlaveRxEvent(void (*)(uint8_t*, size_t));
void twi_attachSlaveTxEvent(void (*)(void));
void twi_reply(uint8_t);
//void twi_stop(void);
void twi_releaseBus(void);
```
It looks like twi_stop was commented out at the bottom and then added back in again above that, unknown why as it's not a valid function for a slave.

Here's Wire.h from the library directory in it's entirety:
```cpp
/*
    TwoWire.h - TWI/I2C library for Arduino & Wiring
    Copyright (c) 2006 Nicholas Zambetti.  All right reserved.
    This library is free software; you can redistribute it and/or
    modify it under the terms of the GNU Lesser General Public
    License as published by the Free Software Foundation; either
    version 2.1 of the License, or (at your option) any later version.
    This library is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
    Lesser General Public License for more details.
    You should have received a copy of the GNU Lesser General Public
    License along with this library; if not, write to the Free Software
    Modified 2012 by Todd Krein (todd@krein.org) to implement repeated starts
    Modified December 2014 by Ivan Grokhotkov (ivan@esp8266.com) - esp8266 support
    Modified April 2015 by Hrsto Gochkov (ficeto@ficeto.com) - alternative esp8266 support
*/

#ifndef TwoWire_h
#define TwoWire_h

#include <inttypes.h>
#include "Stream.h"



#define BUFFER_LENGTH 128

class TwoWire : public Stream
{
private:
    static uint8_t rxBuffer[];
    static uint8_t rxBufferIndex;
    static uint8_t rxBufferLength;

    static uint8_t txAddress;
    static uint8_t txBuffer[];
    static uint8_t txBufferIndex;
    static uint8_t txBufferLength;

    static uint8_t transmitting;
    static void (*user_onRequest)(void);
    static void (*user_onReceive)(size_t);
    static void onRequestService(void);
    static void onReceiveService(uint8_t*, size_t);
public:
    TwoWire();
    void begin(int sda, int scl);
    void begin(int sda, int scl, uint8_t address);
    void pins(int sda, int scl) __attribute__((deprecated)); // use begin(sda, scl) in new code
    void begin();
    void begin(uint8_t);
    void begin(int);
    void setClock(uint32_t);
    void setClockStretchLimit(uint32_t);
    void beginTransmission(uint8_t);
    void beginTransmission(int);
    uint8_t endTransmission(void);
    uint8_t endTransmission(uint8_t);
    size_t requestFrom(uint8_t address, size_t size, bool sendStop);
    uint8_t status();

    uint8_t requestFrom(uint8_t, uint8_t);
    uint8_t requestFrom(uint8_t, uint8_t, uint8_t);
    uint8_t requestFrom(int, int);
    uint8_t requestFrom(int, int, int);

    virtual size_t write(uint8_t);
    virtual size_t write(const uint8_t *, size_t);
    virtual int available(void);
    virtual int read(void);
    virtual int peek(void);
    virtual void flush(void);
    void onReceive(void (*)(int));      // arduino api
    void onReceive(void (*)(size_t));   // legacy esp8266 backward compatibility
    void onRequest(void (*)(void));

    using Print::write;
};

#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_TWOWIRE)
extern TwoWire Wire;
#endif

#endif
```
I've compiled the master with all functions called, and slave with both send and receive functions used, and the Twi::stop didn't appear in the dumps for either.  It's listed as an exposed function in the core .cpp file and also listed in the core twi.h, but never in Wire.h.  
I've verified that it doesn't change any of the functional test results for master or slave.

cores/esp8266/core_esp8266_si2c.cpp
Tech-TX: Here's the commit with the change:
https://github.com/esp8266/Arduino/pull/6934/commits/9cf109546fc0d198c5d91005745dde80d2d8b611
Travis running with it now.

PR #6936, opened 2019-12-22
Code size optimisations for ESP.getFullVersion()

dirkmueller:
This saves about ~ 60 bytes of flash usage (50% reduction of
the total function size)

### DIFF
--- a/cores/esp8266/Esp-version.cpp
+++ b/cores/esp8266/Esp-version.cpp
@@ -28,29 +28,41 @@
 #define STRHELPER(x) #x
 #define STR(x) STRHELPER(x) // stringifier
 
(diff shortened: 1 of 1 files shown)


earlephilhower:
@dirkmueller, neat work, but it might end up with more fragmentation due to the multiple `+=`s.

A `.reserve()` of some sufficiently long number (probably empirically determined) would ensure no memory allocation for the string concat'ing.

Another idea, since you seem to be on a PMEM optimization kick :), most of this string concatenation can be replaced with a single constant string, so some multi-line `#define` may be able to make many of the String `+=`s unnecessary (i.e. no need to += when you're just adding two compile-time strings.

dirkmueller:
@earlephilhower its not worse than temporary strings (for the concatenation using +) created before, but you're right, I forgot to add a reserve() statement. fixed now.

PR #6937, opened 2019-12-22
Reduce temporary string creation in HTTPClient::sendHeader

dirkmueller:
This improves both performance due to fewer memory allocations/copies
as well as reduces code size by ~ 25% (150 bytes)

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -1227,8 +1227,9 @@ bool HTTPClient::sendHeader(const char * type)
     }
 
     String header;
-    // 168: Arbitrarily chosen to have enough buffer space for avoiding internal reallocations
-    header.reserve(_headers.length() + _base64Authorization.length() + _host.length() + 168);
+    // 128: Arbitrarily chosen to have enough buffer space for avoiding internal reallocations
+    header.reserve(_headers.length() + _uri.length() +
+            _base64Authorization.length() + _host.length() + _userAgent.length() + 128);
     header += type;
     header += ' ';
     if (_uri.length()) {
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -1076,12 +1076,14 @@ String HTTPClient::errorToString(int error)
 void HTTPClient::addHeader(const String& name, const String& value, bool first, bool replace)
 {
     // not allow set of Header handled by code
-    if(!name.equalsIgnoreCase(F("Connection")) &&
-       !name.equalsIgnoreCase(F("User-Agent")) &&
-       !name.equalsIgnoreCase(F("Host")) &&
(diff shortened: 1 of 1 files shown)


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1225,41 +1226,49 @@ bool HTTPClient::sendHeader(const char * type)
         return false;
     }
 
-    String header = String(type) + ' ' + (_uri.length() ? _uri : F("/")) + F(" HTTP/1.");
+    String header;
+    // 168: Arbitrarily chosen to have enough buffer space for avoiding internal reallocations
+    header.reserve(_headers.length() + _base64Authorization.length() + _host.length() + 168);
```
devyte: Choosing some estimated upper bound like this is fine, but how about doing a finer estimate here? You have at hand at least the following:
- _uri.length(), which in theory could be up to 2048 chars max, although an appropriate length is considered to be 75
- _host.length() which could be up to 255
- userAgent.length() I have no idea what the max length is, but a quick search shows upwards of 100

Of course, you could then reduce your 168 to something smaller.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1336,9 +1348,12 @@ int HTTPClient::handleHeaderResponse()
                         break; // We found a match, stop looking
                     }
                 }
+                continue;
```
devyte: Is this continue correct?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1336,9 +1348,12 @@ int HTTPClient::handleHeaderResponse()
...
             }
 
-            if(headerLine == "") {
+            headerLine.trim(); // remove \r
```
devyte: What is the reason to move this trim so far below?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
dirkmueller: There is a trade-off to make: correctness of the calculation vs code size overhead for the correct calculation. you're correct that we can add _uri.length and userAgent.length() that can be added, but lets not overdo it because it really doesn't matter whether we'll have one realloc() or zero realloc(). the important improvement here is that we no longer have on average 4-5 realloc()s.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
dirkmueller: yes, because the only thing after this line is to check whether the headerline is the "empty line" aka whether we're at the end of the section of headers. if we have determined that there is a ':' in the current line we already know it is no longer empty and can stop caring about the empty line case.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
dirkmueller: so that it doesn't run when it isn't needed - for the majority of cases we only need to trim the value, which is done earlier.

PR #6938, opened 2019-12-22
SNTP: expose sntp_set_daylight(int daylight) & add sntp_get_daylight(…

LeisureLadi:
It would be helpful for control of day light saving settings to expose the sntp_set_daylight() function to the application layer via time.cpp and have an additional function sntp_get_daylight() function to read back the current setting.
I made the corresponding changes to sntp-lwip2.cpp, time.cpp and Arduino.h.
The corresponding wrapper functions in time.cpp are getDST() and setDST().


LeisureLadi:
I thought about avoiding the wrapper, but then, as you correctly mentioned, decided to maintain the structure. However, both way would be ok for me. Please let me know, if I shall change the request.

d-a-v (approved):

d-a-v:
@LeisureLadi would you fix the merge conflicts ?

d-a-v:
Closing per unresolvable conflicts due to changed API.
daylight is now set with `configTime(timezone_in_seconds, daylight_saving_time_in_seconds)` overload, which is artificially added to timezone.
In any case, the nearly-POSIX way `configTime(timezone_string)` is preferred.
`localtime()` and `gmtime()` (see manpages) allow to guess what would DST be.

PR #6950, opened 2019-12-26
PROGMEM footprint cleanup for responseCodeToString

dirkmueller:
Doing returns with String/FlashStringHelper conversion for every
case means that the code for doing that is duplicated, which
is wasteful. doing it only once saves about 250 bytes (50%) of
code size.

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
@@ -801,9 +801,6 @@ String ESP8266WebServerTemplate<ServerType>::responseCodeToString(const int code
     case 417:
         r = F("Expectation Failed");
         break;
-    case 418:
-        r = F("I'm a teapot");
-        break;
     case 500:
         r = F("Internal Server Error");
         break;
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
@@ -695,49 +695,136 @@ void ESP8266WebServerTemplate<ServerType>::_finalizeResponse() {
 }
 
 template <typename ServerType>
-const String ESP8266WebServerTemplate<ServerType>::responseCodeToString(const int code) {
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -695,49 +695,136 @@ void ESP8266WebServerTemplate<ServerType>::_finalizeResponse() {
...
+    case 415:
+        r = F("Unsupported Media Type");
+        break;
+    case 416:
+        r = F("Range not satisfiable");
+        break;
+    case 417:
+        r = F("Expectation Failed");
+        break;
+    case 418:
```
earlephilhower: Can we drop the silly 418, for another 16 bytes of flash saved?

devyte: This is RFC2324, and supported by all major browsers.
But yes :)

dirkmueller: done

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -695,49 +695,133 @@ void ESP8266WebServerTemplate<ServerType>::_finalizeResponse() {
...
-    case 500: return F("Internal Server Error");
-    case 501: return F("Not Implemented");
-    case 502: return F("Bad Gateway");
-    case 503: return F("Service Unavailable");
-    case 504: return F("Gateway Time-out");
-    case 505: return F("HTTP Version not supported");
-    default:  return F("");
-  }
+String ESP8266WebServerTemplate<ServerType>::responseCodeToString(const int code) {
+    const __FlashStringHelper *r;
```
devyte: Please add a comment explaining why you're doing this. I can easily picture somebody "simplifying" the switch back in the future.

dirkmueller: good point, done!

PR #6954, opened 2019-12-28
Fix Updater potential overflow, add host tests

### LINKED ISSUE #4674: ESP8266 Updater class does not honour size parameter in begin(size)
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).

#### Platform

- Hardware:					All
- Core Version:			Latest auto install
- Operating System:		Ubuntu

### Settings in IDE

- Flash Mode:			qio
- Flash Size: 		4MB
- Reset Method:		ck
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz|
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

There is an existing bug in the code for the Updater class that is still present in the latest GIT source available online. The error is in the write() method and more particularly the check for running out of space with:

```cpp
if(len > remaining()){
```

I have been developing my own code to update OTA via a web server ESP8266WebServer that then calls Update.begin(), Update.write(), Update.end() and if I upload MORE bytes than I specify in begin(), write() does NOT trigger an error and continues writing to flash, eventually overwriting the SPIFFS directory.

This may be an issue with the _bufferSize = FLASH_SECTOR_SIZE = 4096 and the buffer sent to write being 2048 which are exact multiples of each other or it may be generically incorrect

The bug can be verified by walking through an example where _size = 5000, and manually considering what happens on each call to write() with a buffer of 2048

Call 1 -> 2048 bytes get stored in buffer, method returns 2048
Call 2 -> 2048 bytes get stored, buffer is full but NOT written yet, method returns 2048
Call 3 -> 2048 is now too many, but the test succeeds. Now the 4096 byte buffer gets flashed, the next 2048 bytes get stored in the buffer, and remaining() will forever more return a negative number (size_t being unsigned this means a huge number) and an unlimited amount of bytes can be written

I believe the test should be changed to check if written (progress() + bytes in buffer + len is greater than size

```cpp
if (progress() + _bufferLen + len > _size) {
```

### Why hasn't this been caught

It appears that most examples just use the:

```cpp
    uint32_t maxSketchSpace = (ESP.getFreeSketchSpace() - 0x1000) & 0xFFFFF000;
```
to pass to begin() which means that the file will typically finish before the problem is seen

earlephilhower:
Fixes #4674

The Updater class could, when exactly 4K bytes were in the buffer but
not yet written to flash, allow overwriting data written to it beyond
the passed-in size parameter.

Fix per @jason-but's suggestion, and add a host test (plus minor changes
to Updater code to support host testing).


cores/esp8266/Updater.cpp
```diff
@@ -104,7 +104,9 @@ bool UpdaterClass::begin(size_t size, int command, int ledPin, uint8_t ledOn) {
   _reset();
   clearError(); //  _error = 0
 
+#ifndef HOST_MOCK
   wifi_set_sleep_type(NONE_SLEEP_T);
+#endif
```
mcspr: Can this be defined in the user_interface.cpp mock to avoid ifndef?

cores/esp8266/Updater.cpp
```diff
@@ -115,7 +117,11 @@ bool UpdaterClass::begin(size_t size, int command, int ledPin, uint8_t ledOn) {
 
   if (command == U_FLASH) {
     //address of the end of the space available for sketch and update
+#ifdef HOST_MOCK
+    uintptr_t updateEndAddress = (uintptr_t) -1;
+#else
     uintptr_t updateEndAddress = (uintptr_t)&_FS_start - 0x40200000;
+#endif
```
mcspr: Can this ifndef be avoided if something provides _FS_start symbol?

cores/esp8266/Updater.cpp
earlephilhower: It actually is defined there already, but unfortunately `user_interface.cpp` is not for CI testing but for building host apps and including that ends up eventually requiring the whole host app sources (with its own main, etc.).

The host app and host tests have differing build setups and assumptions (and goals, obviously).  It's probably possible to refactor them completely to avoid the duplication here, but I'm not so sure this is the right time for it (3.0.0-ish feels right).

cores/esp8266/Updater.cpp
earlephilhower: Awesome idea.  Let me pull in your suggestion in the PR that fixes the uninitted write() buffer valgrind error.

PR #6280, opened 2019-07-10
WiFi Mesh Update 2.2

aerlon:
Main changes:
- Add new ESP-NOW mesh backend.
- Add HelloEspnow.ino example to demonstrate the current ESP-NOW mesh backend features.
- Deprecate the ESP8266WiFiMesh class in favour of the new ESP-NOW and TCP/IP backends.
- Update the TCP/IP mesh backend to use the new lwIP version preprocessor flag and remove obsolete preprocessor flags.

The code is still a bit of a work in progress, but it will give you an idea of what the ESP-NOW backend can do. The ESP-NOW mesh backend API is not guaranteed to be stable yet. Below is what remains before the PR is complete:

- [x] Change HMAC to use BearSSL.
- [x] Include the new ESP-NOW broadcast functionality. (#6257)
- [x] Make backends use separate ConnectionQueues so they won't interfere with each other.
- [x] Minor refactoring.
- [x] Move most Strings to Flash. Benchmark performance first.
- [x] Add mesh network implementation.
- [x] Update library documentation and keywords.

PR #6961, opened 2019-12-29
Add CryptoInterface library

aerlon:
- Add CryptoInterface library

- Add TypeConversion core files.

A library providing an easy-to-use frontend for the cryptographic functionality of BearSSL. Created as an offshoot from the code of the new mesh library PR, based on discussions [here](https://github.com/esp8266/Arduino/pull/6280#discussion_r342266973).


cores/esp8266/TypeConversion.cpp
```diff
@@ -0,0 +1,82 @@
...
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+   THE SOFTWARE.
+*/
+
+#include <assert.h>
+#include "TypeConversion.h"
+
+namespace esp8266 {
+  namespace TypeConversion {
+    const char base36Chars[36] = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z'};
```
earlephilhower: These two declarations add 110+ bytes to RODATA (i.e. shrink heap).  Is there a better way (either calc'ing them or PROGMEM) to handle them?

cores/esp8266/TypeConversion.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+      for (uint32_t i = 0; i < arrayLength; ++i) {
+        uint8Array[i] = (base36CharValues[hexString.charAt(i * 2) - '0'] << 4) + base36CharValues[hexString.charAt(i * 2 + 1) - '0'];
+      }
+
+      return uint8Array;
+    }
+
+    uint8_t *uint64ToUint8Array(const uint64_t value, uint8_t *resultArray) {
+      resultArray[7] = value;
```
earlephilhower: I appreciate the clean C code, but on the 8266 couldn't this just reduce to `memcpy(resultArray, &value, 8)` or a simple pointer-cast and assign?

cores/esp8266/TypeConversion.cpp
```diff
@@ -0,0 +1,82 @@
...
+      resultArray[3] = value >> 32;
+      resultArray[2] = value >> 40;
+      resultArray[1] = value >> 48;
+      resultArray[0] = value >> 56;
+
+      return resultArray;
+    }
+
+    uint64_t uint8ArrayToUint64(const uint8_t *inputArray) {
+      uint64_t result = (uint64_t)inputArray[0] << 56 | (uint64_t)inputArray[1] << 48 | (uint64_t)inputArray[2] << 40 | (uint64_t)inputArray[3] << 32
```
earlephilhower: Same as above.

cores/esp8266/TypeConversion.h
```diff
@@ -0,0 +1,78 @@
...
+*/
+
+#ifndef __ESP8266_TYPECONVERSION_H__
+#define __ESP8266_TYPECONVERSION_H__
+
+#include <Arduino.h>
+
+namespace esp8266 {
+  namespace TypeConversion {
+    extern const char base36Chars[36];
```
earlephilhower: Maybe these could be static and local to the TC.cpp file?  Not seeing the use in general code.

libraries/CryptoInterface/src/CryptoInterface.cpp
```diff
@@ -0,0 +1,566 @@
...
+     * The sample passed all tests."
+     * 
+     * Since ESP32 is the sequal to ESP8266 it is unlikely that the ESP8266 is able to generate random numbers more quickly than 5 MHz when run at a 80 MHz frequency.
+     * A maximum random number frequency of 0.5 MHz is used here to leave some margin for possibly inferior components in the ESP8266.
+     * It should be noted that the ESP8266 has no Bluetooth functionality, so turning the WiFi off is likely to cause RANDOM_REG32 to use pseudo-random numbers.
+     * 
+     * It is possible that yield() must be called on the ESP8266 to properly feed the hardware random number generator new bits, since there is only one processor core available. 
+     * However, no feeding requirements are mentioned in the ESP32 documentation, and using yield() could possibly cause extended delays during nonce generation.
+     * Thus only delayMicroseconds() is used below.
+     */
```
earlephilhower: That's something I did not know about the HW.  Assumed it was driven off of the usual Zener.  Do you think the BSSL port needs to include this?

libraries/CryptoInterface/src/CryptoInterface.cpp
```diff
@@ -0,0 +1,566 @@
...
+    return createBearsslHmacCT(&br_md5_vtable, MD5_NATURAL_LENGTH, message, hashKey, hashKeyLength, hmacLength);
+  }
+
+
+  // #################### SHA-1 ####################
+
+  // resultArray must have size SHA1_NATURAL_LENGTH or greater
+  void *sha1Hash(const void *data, const size_t dataLength, void *resultArray)
+  {
+    if(warningsEnabled())
```
earlephilhower: This should probably have a `deprecated` applied to all the sha1 functions in this case.  Get compiler warnings no matter what.

The warningsEnabled() really needs to be a compile-time constexpr.  Suggest a DEBUG-type macro instead.  Don't want these strings in code for prod ships...

libraries/CryptoInterface/src/CryptoInterface.cpp
```diff
@@ -0,0 +1,566 @@
...
+  bool warningsEnabled() { return _warningsEnabled; }
+
+  void setNonceGenerator(nonceGeneratorType nonceGenerator) { _nonceGenerator = nonceGenerator; }
+  nonceGeneratorType getNonceGenerator() { return _nonceGenerator; }
+
+
+  // #################### MD5 ####################
+    
+  // resultArray must have size MD5_NATURAL_LENGTH or greater
+  void *md5Hash(const void *data, const size_t dataLength, void *resultArray)
```
earlephilhower: We have MD5 stuff in the core, too, which should be migrated to use this (but with same API, probably, so turned into just a wrapper).

libraries/CryptoInterface/src/CryptoInterface.cpp
```diff
@@ -0,0 +1,566 @@
...
+    getNonceGenerator()(nonce, 12);
+    
+    chacha20Poly1305Kernel(1, data, dataLength, key, keySalt, keySaltLength, nonce, resultingTag, aad, aadLength);            
+  }
+  
+  bool chacha20Poly1305Decrypt(void *data, const size_t dataLength, const void *key, const void *keySalt, const size_t keySaltLength,
+                               const void *encryptionNonce, const void *encryptionTag, const void *aad, const size_t aadLength)
+  {
+    const uint8_t *oldTag = (const uint8_t *)encryptionTag;
+    uint8_t newTag[16] {0};
```
earlephilhower: Never saw that syntax.  Neat.

libraries/CryptoInterface/src/CryptoInterface.cpp
aerlon: Include yields or delays you mean? Good question. I think I have too little data to answer conclusively. I simply couldn't find any source on the topic other than the ESP32 docs, which may still be applicable since the two MCUs are closely related. However, it is possible that the random number generation happens in a different way for the ESP8266 compared to the ESP32, so I wouldn't completely rule out a Zener diode being used.

So until we have more data I guess the answer depends on the cost of making the changes. If they are too high, perhaps it is worthwhile seeking clarification from Espressif first.

cores/esp8266/TypeConversion.cpp
```diff
@@ -0,0 +1,82 @@
...
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+   THE SOFTWARE.
+*/
+
+#include <assert.h>
+#include "TypeConversion.h"
+
+namespace esp8266 {
+  namespace TypeConversion {
```
dirkmueller: the preferred indentation style is no indentations for namespaces and otherwise 4 spaces. could you please reformat it (you can do astyle with the tests/astyle_core.conf for example)

cores/esp8266/TypeConversion.cpp
aerlon: Putting the arrays in PROGMEM adds a bit to the conversion time (up to 15 %), but that's probably an OK trade-off.

cores/esp8266/TypeConversion.cpp
aerlon: A straight memory copy is not possible if we want to keep compatibility with how MAC addresses are handled in the Arduino Core at the moment (e.g. the conversion `MAC to uint64_t to String` would give a backwards result compared to `MAC to String`). This can be resolved by reversing the result of the memory copy, e.g. via `std::reverse`, but based on testing this would increase conversion time compared to the current approach by at least 75 % with the default compiler optimization flag (-Os).

cores/esp8266/TypeConversion.cpp
aerlon: Same reply as above, but with a conversion time increase of at least 92 %.

cores/esp8266/TypeConversion.h
aerlon: The base36 arrays are used in a few other type conversion functions which based on conversations with @devyte may end up in separate classes. I thought it could be a good idea to put the arrays in TypeConversion for use in all such conversion functions, but I'm open to other ideas.

libraries/CryptoInterface/src/CryptoInterface.cpp
aerlon: Excellent idea. I added the deprecated attribute to the MD5 and SHA1 hash functions, and moved the old warnings to the comments.

HMAC seems to be safe even though the underlying hash function is broken in terms of collision resistance [[1]](https://crypto.stackexchange.com/questions/9336/is-hmac-md5-considered-secure-for-authenticating-encrypted-data)[[2]](https://tools.ietf.org/html/rfc6151)[[3]](https://stackoverflow.com/questions/2889473/when-is-it-safe-to-use-a-broken-hash-function). So the warning is only applicable to the hash functions for now, although it is naturally prudent to choose a less broken underlying hash function if possible, just in case.

aerlon:
Thank you for the feedback. I've corrected most issues and depending on what you think of the remaining ones this PR may be merge ready.

@dirkmueller Ah, so there is a separate style for core files and example files. Good to know.

@earlephilhower Thanks! I think there's actually more comments than code in this library, but then again it is supposed to be a frontend.

@devyte I should probably finish rewriting the mesh library before starting to rewrite this, so I've put the entire CryptoInterface in namespace experimental for now.

libraries/CryptoInterface/src/CryptoInterface.h
```diff
@@ -0,0 +1,799 @@
...
+   * 1024 by default.
+   */
+  size_t ctMaxDataLength();
+
+  /**
+   * Turn on or off warning Serial prints from the CryptoInterface functions.
+   * 
+   * @param warningsEnabled If true, warnings will be printed to Serial.
+   */
+  void setWarningsEnabled(bool warningsEnabled);
```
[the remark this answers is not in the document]
aerlon: All runtime warnings removed.

cores/esp8266/TypeConversion.h
devyte: What is usually done in such cases is to move such arrays into another cpp, and export them in a .h which then gets included wherever.
I think we can clean that up later, this is all in experimental so we can make changes without having to explain or jump through hoops.

libraries/CryptoInterface/src/CryptoInterface.cpp
devyte: I think this is still valid. The core would likely make use of this interface for ease of use and for cleaning up current code. Also, crcXX should eventually be integrated here.

devyte:
@aerlon I think this needs some restructuring, and probably moving out of a library and into the core, but given that it's in namespace experimental we can merge it as-is and repolish later.

cores/esp8266/TypeConversion.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    for (uint32_t i = 0; i < arrayLength; ++i)
+    {
+        uint8Array[i] = (pgm_read_byte(base36CharValues + hexString.charAt(i * 2) - '0') << 4) + pgm_read_byte(base36CharValues + hexString.charAt(i * 2 + 1) - '0');
+    }
+
+    return uint8Array;
+}
+
+uint8_t *uint64ToUint8Array(const uint64_t value, uint8_t *resultArray)
```
earlephilhower, line 69: This code is unneeded.  A simple cast suffices on this machine due to byte ordering:

````
void setup() {
Serial.begin(115200);
Serial.printf("\n");
union { uint64_t b64; uint8_t b8[8]; } x;
x.b64 = 0x1122334455667788LL;
for (auto i=0; i<8; i++) Serial.printf("%02x ", x.b8[i]);
Serial.printf("\n");
}
void loop {}
````
gives
````
88 77 66 55 44 33 22 11 
````

File: cores/esp8266/TypeConversion.cpp [hunk 12]
```diff
@@ -0,0 +1,91 @@
...
+    resultArray[2] = value >> 40;
+    resultArray[1] = value >> 48;
+    resultArray[0] = value >> 56;
+
+    return resultArray;
+}
+
+uint64_t uint8ArrayToUint64(const uint8_t *inputArray)
+{
+    uint64_t result = (uint64_t)inputArray[0] << 56 | (uint64_t)inputArray[1] << 48 | (uint64_t)inputArray[2] << 40 | (uint64_t)inputArray[3] << 32
```
earlephilhower, line 85: Same as above

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
```diff
@@ -0,0 +1,102 @@
...
+*/
+constexpr char masterKey[] PROGMEM = "w86vn@rpfA O+S"; // Use 8 random characters or more
+
+void setup() {
+  // Prevents the flash memory from being worn out, see: https://github.com/esp8266/Arduino/issues/1054 .
+  // This will however delay node WiFi start-up by about 700 ms. The delay is 900 ms if we otherwise would have stored the WiFi network we want to connect to.
+  WiFi.persistent(false);
+
+  Serial.begin(115200);
+  delay(50); // Wait for Serial.
```
earlephilhower, line 31: Not needed on this chip.  I think it's a leftover from the original AVR.

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
```diff
@@ -0,0 +1,102 @@
...
+
+  //yield(); // Use this if you don't want to wait for Serial.
+
+  // The WiFi.disconnect() ensures that the WiFi is working correctly. If this is not done before receiving WiFi connections,
+  // those WiFi connections will take a long time to make or sometimes will not work at all.
```
earlephilhower, line 36: @d-a-v, I'm not sure I agree with the comment.  Can you please chime in?

libraries/CryptoInterface/src/CryptoInterface.h
```diff
@@ -0,0 +1,806 @@
...
+{
+  namespace CryptoInterface 
+  {
+    /**
+     * Regarding constant-time (CT) HMAC:
+     * 
+     * Basically, constant-time algorithms makes it harder for attackers to learn things about your system based on the execution time of code. 
+     * Good intro here: https://www.bearssl.org/constanttime.html
+     * 
+     * It should be noted that every HMAC is already partially constant-time. Quoting the link above:
```
earlephilhower, line 41: +1 here and elsewhere for the detailed comments!

libraries/CryptoInterface/src/CryptoInterface.h
```diff
@@ -0,0 +1,806 @@
...
+     * Create a MD5 hash of the data. The result will be MD5_NATURAL_LENGTH bytes long and stored in resultArray.
+     * Uses the BearSSL cryptographic library.
+     * 
+     * @param data The data array from which to create the hash.
+     * @param dataLength The length of the data array in bytes.
+     * @param resultArray The array wherein to store the resulting hash. MUST be be able to contain MD5_NATURAL_LENGTH bytes or more.
+     * 
+     * @return A pointer to resultArray.
+     */
+    void *md5Hash(const void *data, const size_t dataLength, void *resultArray) __attribute__((deprecated));
```
earlephilhower, line 128: It seems to me that the string of hashes all have the same API (since they're just wrappers for BSSL functions and, well, hashes have a well defined interface).  Is this not a good spot for a *little* abstraction with a `HashClass` parent and a set of descendent objects (MD5, SHA1, etc.)?  You'd then call with `Crypto::MD5::hash(x....)` and get sizes for arrays with `uint8_t dest[MD5:hashSize];` and be able to pass them around abstractly (say, on user choice or handshake options) as `HashClass` or something?

libraries/CryptoInterface/src/CryptoInterface.h
```diff
@@ -0,0 +1,806 @@
...
+     * 
+     * Must be called at least once before hkdfProduce() can be used.
+     * 
+     * @param keyMaterial An array containing the key material to use when deriving subkeys. Typically this would be the master key.
+     * @param keyMaterialLength The length of keyMaterial in bytes.
+     * @param salt An array containing the salt to use when ingesting key material. Salt is non-secret and can be empty.
+     *             Its role is normally to bind the input to a conventional identifier that qualify it within the used protocol or application.
+     * @param saltLength The length of the salt array, in bytes.
+     */
+    void hkdfInit(const void *keyMaterial, const size_t keyMaterialLength, const void *salt = nullptr, const size_t saltLength = 0);
```
earlephilhower, line 695: There's obviously some hidden state (since there are separate init and produce calls.  I'd also think this is an object-ification opportunity.

libraries/CryptoInterface/src/CryptoInterface.h
```diff
@@ -0,0 +1,806 @@
...
+     * @param keySaltLength The length of keySalt in bytes.
+     * @param resultingNonce The array that will store the nonce generated during encryption. Must be able to contain at least 12 bytes. The nonce is not secret and must be passed to the decryption function.
+     * @param resultingTag The array that will store the message authentication tag generated during encryption. Must be able to contain at least 16 bytes. The tag is not secret and must be passed to the decryption function.
+     * @param aad Additional authenticated data. This data will be covered by the Poly1305 MAC, but not encrypted. 
+     *            You can include the unencrypted parts of your message as AAD to ensure that the encrypted content cannot 
+     *            be re-sent with replaced unencrypted data by an attacker.
+     *            Defaults to nullptr.
+     * @param aadLength The length of the aad array in bytes. Defaults to 0.
+     */
+    void chacha20Poly1305Encrypt(void *data, const size_t dataLength, const void *key, const void *keySalt, const size_t keySaltLength, void *resultingNonce, void *resultingTag, const void *aad = nullptr, const size_t aadLength = 0);
```
earlephilhower, line 775: Same as above.  ChaCha is fast on the 8266, but AES is also widely used and abstracting one more layer into an encryption major class seems more logical to me.

libraries/CryptoInterface/src/CryptoInterface.cpp
```diff
@@ -0,0 +1,567 @@
...
+{
+  namespace
+  {
+    size_t _ctMinDataLength = 0;
+    size_t _ctMaxDataLength = 1024;
+    
+    br_hkdf_context _storedHkdfContext;
+    bool _hkdfContextStored = false;
+
+    uint8_t *defaultNonceGenerator(uint8_t *nonceArray, const size_t nonceLength)
```
earlephilhower, line 44: Great attention to detail on the lack of entropy.  I think a proper "get random number" call (maybe renaming it or something) would also be handy, returning just safe u32?  I can see it being handy everywhere `rand()` or `RANDOM_REG` is (improperly) used.  This fcn would also use it to grab the RANDOM_REG, of course, so that the timing would be preserved...

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
earlephilhower: To be specific, the `delay` isn't needed.  `Serial.begin` still is.

libraries/CryptoInterface/src/CryptoInterface.h
devyte: This is what I meant with my comment about something like struct md5 {} struct sha1 {} etc. It means restructuring.
A note here: if a parent class is to be implemented, I suggest doing that only as guidance for a common interface, and using non-virtualness in the parent. In other words, aim for templated usage and not polymorphism,

cores/esp8266/TypeConversion.cpp
aerlon: The suggestion is neat, but note that the result is reversed compared to what the original function provides. (i.e. with the original function you get `11 22 33 44 55 66 77 88` as output) I wrote this code a while back now, but if I remember correctly it used this byte order to be compatible with the existing functions in the core that return a MAC array (`WiFi.softAPmacAddress()` etc.). An easy way to get exactly the same result as before would be to use `std::reverse( std::begin(x.b8), std::end(x.b8) );`, but that is slightly slower than just using the current solution (6 % or so).

cores/esp8266/TypeConversion.cpp
aerlon: As above.

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
aerlon: I've previously noticed that initial prints are sometimes dropped without the delay. This has not always been the case though, so maybe it was a bug that has been fixed now?

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
aerlon: Seems like I added this in a commit 2018-09-14, and back then I know it was required. I can't see any problem now when removing it in the example code, so maybe this was a bug that has been fixed. With the caveat that my original test case may have been different, it would seem as though `WiFi.disconnect()` can be removed here.

libraries/CryptoInterface/src/CryptoInterface.h
aerlon: Thanks!

libraries/CryptoInterface/src/CryptoInterface.cpp
aerlon: If I understand this correct, you would like a version of the `defaultNonceGenerator` function that returns an uint32_t instead of an uint8_t array? I'm thinking it may be possible to break out the `defaultNonceGenerator` functionality into a separate base function that looks exactly the same as  `defaultNonceGenerator` does now. This base function can then be called from two wrapper functions, one being `defaultNonceGenerator` and the other being the new function you describe that returns an uint32_t. Was this what you had in mind?

libraries/CryptoInterface/src/CryptoInterface.h
aerlon: There is some room for abstraction here, indeed. @devyte Do you have some existing code base as an example for the templated usage you would like? It would help me avoid misunderstandings.

libraries/CryptoInterface/src/CryptoInterface.cpp
earlephilhower: I'd go further.  The standard random generator(Arduino `random`) at least.

Also, I'd suggest because this is useful *everywhere* that "nonce" is probably not the right naming.  `uint32_t ESP.random()` might be the right spot to expose it.  Want to make it obvious that this is the right way to get real random #s (as opposed to newlib pseudorand which cannot be changed).

libraries/CryptoInterface/src/CryptoInterface.h
earlephilhower: There's already a HashClass used in the BearSSLHelpers.h.  I know @devyte hates virtual methods, but IMO a nice virtual base and overridden methods is good enough and lets us have 1 copy of the code in flash that can handle any kind of hash (vs. templates where you regen a completely new class...which can't be used by others without also making them templated...it's a virus!)

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
earlephilhower: It could have been fixed before I started working on the core, I don't really know.  But for several years now it's been fine to send/receive immediately after a `Serial.begin()`.  All other examples that I know of have removed the delay...

cores/esp8266/TypeConversion.cpp
earlephilhower: I missed the endianness, mybad!

I would suggest in that case that this needs to also include the endinanness in the function name (BE/LE) as appropriate, please.

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
earlephilhower: Yes, please.  If we do have a problem we should fix it in the core and not workaround it in this example.

libraries/CryptoInterface/src/CryptoInterface.h
devyte: I was thinking just free standing structs, e. g. struct md5, struct sha1, struct sha256, etc, each with methods hash(), and whatever else is needed.
I somehow don't think they can all have the exact same interface, which would make polymorphism very messy, because the base class would need to have a superset of all cases. Let's try free standing objects first, and if needed we can implement a base class later and have all inherit from that.
About the BearSSL HashClass, it can be reworked as a thin polymorphic wrapper around these if needed.

libraries/CryptoInterface/examples/HelloCrypto/HelloCrypto.ino
aerlon: Hmm... I do remember having problems with the first few `Serial.println` not showing up as late as the autumn of 2019, but removing the `delay` now seems to cause no issue. Perhaps it only happens in some special cases, or perhaps the problem has been fixed. I'll remove the `delay` for now since it does not seem to have an effect.

cores/esp8266/TypeConversion.cpp
aerlon: BE added to function names.

libraries/CryptoInterface/src/CryptoInterface.h
aerlon: Code classified.

libraries/CryptoInterface/src/CryptoInterface.cpp
aerlon: New ESP.random methods have been added to the core based on the crypto code, and these new random methods are now used by the defaultNonceGenerator.

libraries/CryptoInterface/src/CryptoInterface.h
aerlon: There is now an object with encrypt and decrypt methods. Since free standing objects [were requested ](https://github.com/esp8266/Arduino/pull/6961#discussion_r415436174) no further abstraction has been done.

libraries/esp8266/examples/HelloCrypto/HelloCrypto.ino
```diff
@@ -0,0 +1,96 @@
...
+  WiFi.persistent(false);
+
+  Serial.begin(115200);
+
+  Serial.println();
+  Serial.println();
+}
+
+void loop() {
+  // This serves only to demonstrate the library use. See the header file for a full list of functions.
```
devyte, line 39: Up here can do something like:
```cpp
using namespace experimental::crypto;
```

Then further down you can use directly, e.g.:
```cpp
...
getNonceGenerator()(...);
...
SHA256::hash(...);
...
SHA256::hmac(...);
```
etc. This eases use and reduces the long fully qualified names. It also allows easy remapping of namespaces, because you just change the namespace import at the top instead of having to change each fully qualified name through the code.

aerlon:
@devyte Thanks. The issues you mentioned are fixed in PR #7252.

PR #6972, opened 2020-01-01
Corrected board definitions to include ESP8266

ttytyper:
According to the sonoff_*.json board files from PlatformIO, the build flags are supposed to be ARDUINO_ESP8266_SONOFF_BASIC etc. However, pins_arduino.h expects them to be ARDUINO_SONOFF_BASIC etc, without ESP8266. This causes sketches that make use of i2c with Wire.h to fail, as the necessary pins do not get defined.

I suspect that pins_arduino.h is incorrect and propose this correction. My apologies if this is not the correct way to fix the issue - it's difficult to wrap my head around how the different frameworks, IDEs, build tools etc interact :)

Reference: https://github.com/platformio/platform-espressif8266/tree/develop/boards


mcspr:
It is a bit more complicated than that, since Arduino IDE uses a different build config through platform.txt and boards.txt ini-like files:
https://github.com/esp8266/Arduino/blob/fa5040d5dafaa82334a2c5c345b2cb9449bacc7b/platform.txt#L104 (note the `-DARDUINO_{build.board}`, which means that the IDE will use the variable provided by the menu entry / variable depending on the selected board type)

Resulting boards.txt is generated by this script:
[/tools/boards.txt.py#L866](https://github.com/esp8266/Arduino/blob/fa5040d5dafaa82334a2c5c345b2cb9449bacc7b/tools/boards.txt.py#L866) (part with sonoff boards)
So you'd need to change it too and regenerate boards.txt so that IDE uses correct definitions

Having `_ESP8266_` seems like a correct way to name boards, but not sure what are implications for backwards compatibility are.

mhightower83:
@ttytyper  If you like, I can supply you with the `boards.txt.py` changes for your PR or I can create a PR and include your changes?

d-a-v:
@mcspr you are right, in this PR `board.txt.py` needs this change:
```
            '.menu.BoardModel.sonoffSV.build.board': 'SONOFF_SV',
```
to
```
            '.menu.BoardModel.sonoffSV.build.board': 'ARDUINO_SONOFF_SV',
```
(and similarly with other itead/sonoff models)
@mhightower83 you are the initial author for the itead board update, would you agree with this changes ?
@ttytyper would you be able to update your PR with these changes ? Or leave @mhightower83 do it as proposed ?

mhightower83:
@d-a-v actually, it looks like the recipe in platform.txt adds the `ARDUINO_` part, we need to add the `ESP8266_` to `SONOFF_SV` to complete the correction, etc.
```
'.menu.BoardModel.sonoffSV.build.board': 'ESP8266_SONOFF_SV',
```

```
## Compile c++ files
recipe.cpp.o.pattern=... -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" ...
```

d-a-v:
True, `ESP8266_`  not `ARDUINO_`

ttytyper:
It looks like you have a much better idea of what needs to be done than I do :) Thanks for looking into it.

@mhightower83 Feel free to make the necessary changes. This is all a bit beyond what I'm comfortable messing around with.

earlephilhower:
Superseded by #7024

PR #6988, opened 2020-01-06
SPI: Add async write functions

rsaxvc:
Add new SPI write functions suffixed with underscore that are allowed to return before SPI data is fully written.

Internally, this allows the CPU to run up to one SPI FIFO ahead of the SPI transmitter, for writes only.

A benchmark: Updating a string of 30 APA102 LEDs on 160MHz ESP8266:
    235us - FastGPIO bitbang
      97us - SPI.write32() used for pixel updates
      79us - SPI.write32_() used for pixel updates

This is particularly useful for complex pixel calculations. Approximately 1.5 microseconds of calculations per pixel can now be pipelined in parallel with the previous pixel transmission.

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -147,6 +147,7 @@ void SPIClass::beginTransaction(SPISettings settings) {
 }
 
 void SPIClass::endTransaction() {
+    while(SPI1CMD & SPIBUSY) {}
 }
 
 void SPIClass::setDataMode(uint8_t dataMode) {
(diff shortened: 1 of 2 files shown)


rsaxvc:
Sounds fine to me, I'll put up a new patch when I have time.

Kouzeru:
Any update of this?

rsaxvc:
Sorry, I switched to ESP32 for other reasons, which was fast enough without this mechanism, and I forgot about it. You're welcome to take over or rebase and reupload this patch if it's helpful.

PR #6993, opened 2020-01-06
configTime(tzsec,dstsec,): fix UTC/local management

### LINKED ISSUE #6678: sntp_set_timezone_in_seconds doesn't seem to have any effect
### Basic Infos

#### Platform

- Hardware:				ESP-12E (LoLin  NodeMCU V3)
- Core Version:			ba50bd57b81cf4106fe44ab831d5a061facd4be9
- Development Env:		Arduino IDE 1.8.10
- Operating System:		Ubuntu LTS 18.04.2

### Settings in IDE

- Module:  			NodeMCU 1.0 (ESP-12E Module)
- Flash Size: 		4MB
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

No matter what value I assign to the `timezone_sec` arg in
https://github.com/esp8266/Arduino/blob/36c369bb8fc8c76945203018e97db70387da6df4/cores/esp8266/time.cpp#L101
the output of `time_t now = time(nullptr); Serial.println(time(&now)); Serial.println(ctime(&now));` will be UTC+8.

So if i set `timezone_sec=0` i would expect the output of GMT/UTC but it won't be.

If i set
https://github.com/esp8266/Arduino/blob/36c369bb8fc8c76945203018e97db70387da6df4/cores/esp8266/sntp-lwip2.cpp#L103
to 0 instead, then you will get the time in GMT/UTC.

It seems that the method
https://github.com/esp8266/Arduino/blob/36c369bb8fc8c76945203018e97db70387da6df4/cores/esp8266/sntp-lwip2.cpp#L437
hasn't any effect to the global variable `time_zone`

Also have a look at #6675 

```cpp

#include <ESP8266WiFi.h>
#include <time.h>

const char* ssid       = "...";
const char* password   = "...";

const int timezone_sec = 0;
const int daylightOffset_sec = 0;
const char* ntpServer = "pool.ntp.org";

void setup()
{
  Serial.begin(115200);
  
  Serial.printf("Connecting to %s ", ssid);
  WiFi.begin(ssid, password);
      delay(500);
      Serial.print(".");
  }
  Serial.println(" CONNECTED");
  
  Serial.println("Call configTime...");
  configTime(timezone_sec, daylightOffset_sec, ntpServer);

  Serial.println("Now it's:");
  time_t now = time(nullptr);
  Serial.println(time(&now));
  Serial.println(ctime(&now));

  WiFi.disconnect(true);
  WiFi.mode(WIFI_OFF);
}

void loop()
{
}

```

### Debug Messages

Run at Mon Oct 28 08:03:51 2019 (GMT/UTC)

```
Connecting to xxx ....... CONNECTED
Call configTime...
Now it's:
1572275031
Mon Oct 28 15:03:51 2019
```

### LINKED ISSUE #6921: configTime with TZ offset causes incorrect UTC time
### Basic Infos

#### Platform

- Hardware:					ESP-12
- Core Version:			d40dbb458420939bc83e3d9a2cb81a07c4d2d6b8
- Development Env:		Arduino IDE 1.8.10

### Settings in IDE

- Module:  			Wemos D1 mini
- Flash Size: 		4MB
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   921600

### Problem Description

When using the `void configTime(int timezone, int daylightOffset_sec, const char* server1,
    const char* server2 = nullptr, const char* server3 = nullptr)` with non-zero timezone offset to synchronize time, device doesn't differentiate between local and UTC time, where UTC time is wrongly set to local time. 

This works OK on ESP32 (1.0.4).
 
I've checked #6678, but it seems it is just about initialization issue.

The workaround is using `void configTime(const char* tz, const char* server1,
    const char* server2 = nullptr, const char* server3 = nullptr)`.

```cpp
#if defined(ESP32)
# include <WiFiMulti.h>
  WiFiMulti wifiMulti;
#elif defined(ESP8266)
# include <ESP8266WiFiMulti.h>
  ESP8266WiFiMulti wifiMulti;
#endif

void setup() {
  Serial.begin(115200);

  Serial.println();
  
  // Setup wifi
  WiFi.mode(WIFI_STA);
  wifiMulti.addAP("SSID","PASSWORD");

  Serial.print("Connecting to wifi");
  while (wifiMulti.run() != WL_CONNECTED) {
    Serial.print(".");
    delay(100);
  }
  Serial.println();

  // TZ offset +1h
  configTime(3600, 0 , "pool.ntp.org", "time.nis.gov");
  //configTime("CET-1CEST,M3.5.0,M10.5.0/3" , "pool.ntp.org", "time.nis.gov");
  // Wait till time is synced
  Serial.print("Syncing time");
  int i = 0;
  while (time(nullptr) < 1000000000ul && i<100) {
    Serial.print(".");
    delay(100);
    i++;
  }
  Serial.println();

  time_t tnow = time(nullptr);
  struct tm *timeinfo;
  char buffer [80];


  timeinfo = localtime (&tnow);
  strftime (buffer,80,"Local time: %H:%M.",timeinfo);
  Serial.println(buffer);
  
  timeinfo = gmtime (&tnow);
  strftime (buffer,80,"UTC time: %H:%M.",timeinfo);
  Serial.println(buffer); 
}

void loop() {

}

```

### Debug Messages

```
Connecting to wifi......................
Syncing time.
Local time: 11:17.
UTC time: 11:17.
```

d-a-v:
This PR also removes dead code since probably newlib updates.
The NTP-TZ-DST example is updated.

fixes #6921
fixes #6678


earlephilhower:
Any PR that drops 300+ lines of legacy, unmaintained code w/o dropping functionality is awesome!

cores/esp8266/sntp-lwip2.cpp
```diff
@@ -99,368 +99,26 @@ int settimeofday(const struct timeval* tv, const struct timezone* tz)
...
-        sntp_set_daylight(0);
-    }
+    (void)tz; // obsolete (cf. man settimeofday)
     if (tv) /*after*/
     {
         // reset time subsystem
         tune_timeshift64(tv->tv_sec * 1000000ULL + tv->tv_usec);
 
-        sntp_set_system_time(tv->tv_sec);
+		sntp_real_timestamp = tv->tv_sec;
```
earlephilhower: This seems to rely on a OS timer to handle second ticks.  Is this not going to drift like mad?  If so, does it matter (i.e. are you just looking for a monotonically increasing int somewhere)?

d-a-v: It is like that since the beginning (patched into lwip1 by espressif).
I was thinking this has the same accuracy than anything else handled by the OS or cpu even the cycle counter because everything is driven by the local crystal. Do you think we can have a better accuracy with another local source ?
SNTP readjusts it every hour anyway (by default).

earlephilhower: > Timer APIs can be found in/ESP8266_NONOS_SDK/include/osapi.h. Please note that os_timer APIs listed below are software timers executed in tasks, thus timer callbacks may not be precisely executed at the right time; it depends on priority. If you need a precise timer, please use a hardware timer which can be executed in hardware interrupt.

From the API reference at https://www.espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf

So it's not IRQ driven, but from the description I can't say if it's based on a HW timer (ie. you'll get jitter but not miss anything).  Since you say it's reset every 3600 seconds, I guess it's not a problem.

d-a-v: I think `"may not be precisely executed at the right time"` would indeed mean that it may introduce jitter but the frequency average would stay still (or have the same drift as with HW timers).

cores/esp8266/sntp-lwip2.cpp
```diff
@@ -65,7 +65,7 @@ extern "C" {
 
 static const char stod14[] PROGMEM = "settimeofday() can't set time!\n";
 bool sntp_set_timezone(sint8 timezone);
-bool sntp_set_timezone_in_seconds(sint32 timezone)
+bool sntp_set_timezone_in_seconds(uint32_t timezone)
```
earlephilhower: I'm no expert o/n timezones, but I'm pretty sure UTC-8 exists.  Wouldn't this change make it impossible to set negative UTC timezones

cores/esp8266/sntp-lwip2.cpp
```diff
@@ -99,368 +99,26 @@ int settimeofday(const struct timeval* tv, const struct timezone* tz)
...
 int settimeofday(const struct timeval* tv, const struct timezone* tz)
 {
-    if (tz) /*before*/
-    {
-        sntp_set_timezone_in_seconds(tz->tz_minuteswest * 60);
-        // apparently tz->tz_dsttime is a bitfield and should not be further used (cf man)
-        sntp_set_daylight(0);
-    }
+    (void)tz; // obsolete (cf. man settimeofday)
     if (tv) /*after*/
```
earlephilhower: Since tz must be NULL, and tv must != NULL, I think this needs to be
````
if (tz || !tv) {
  return EFAULT; // or EINVAL as you wish
}
````
And the if (tv) comparison can be removed

File: cores/esp8266/sntp-lwip2.cpp [hunk 4]
```diff
@@ -99,368 +99,26 @@ int settimeofday(const struct timeval* tv, const struct timezone* tz)
...
     {
         // reset time subsystem
         tune_timeshift64(tv->tv_sec * 1000000ULL + tv->tv_usec);
 
-        sntp_set_system_time(tv->tv_sec);
+		sntp_real_timestamp = tv->tv_sec;
+		os_timer_disarm(&sntp_timer);
+		os_timer_setfn(&sntp_timer, (os_timer_func_t *)sntp_time_inc, NULL);
+		os_timer_arm(&sntp_timer, 1000, 1);
```
earlephilhower: Minor tab/space change here, looks like.

cores/esp8266/time.cpp
```diff
@@ -71,14 +73,27 @@ int clock_gettime(clockid_t unused, struct timespec *tp)
...
 }
 
+#if LWIP_VERSION_MAJOR == 1
+// hack for espressif time management included in patched lwIP-1.4
+#define sntp_real_timestamp sntp_get_current_timestamp()
+#endif
+
+#if LWIP_VERSION_MAJOR == 2
+// backport api
+bool sntp_set_timezone_in_seconds (uint32_t timezone_sec)
```
earlephilhower: Same comment as above re: signed/unsigned timezone_sec param

cores/esp8266/time.cpp
```diff
@@ -88,7 +103,7 @@ int _gettimeofday_r(struct _reent* unused, struct timeval *tp, void *tzp)
     if (tp)
     {
         if (!timeshift64_is_set)
-            tune_timeshift64(sntp_get_current_timestamp() * 1000000ULL);
+            tune_timeshift64(sntp_real_timestamp * 1000000ULL);
```
earlephilhower: Tab/space?

libraries/esp8266/examples/NTP-TZ-DST/NTP-TZ-DST.ino
```diff
@@ -212,8 +213,18 @@ void setup() {
...
   // (see below)
+
+  // ----> Here is the ONLY ONE LINE needed in your sketch
+
   configTime(MYTZ, "pool.ntp.org");
 
+  //       Here is the ONLY ONE LINE needed in your sketch <----
+  // pick a value from TZ.h (search for this file in your filesystem) for MYTZ
+
+  // former configTime is still valid, here is the call for 7 to the west
```
earlephilhower: If this is no longer preferred, maybe drop completely from demo?  Don't need to show *all* ways, just the *recommended* ones IMHO.

d-a-v: The example shows what should be used, but also tells the former API is still good for users wanting to upgrade but also seeking for compatibility with other APIs (esp32).

Thanks for the review, all other points are addressed I guess.

arendst:
For some reason after this was merged my sntp code stopped working. 

Undefined references to the once documented SNTP API calls by espressif:
- sntp_get_real_time
- sntp_get_current_timestamp
- sntp_set_timezone

earlephilhower:
@arendst can you open a new issue on this?  This is closed-merged, so anything here won't show up on our lists w/o explicitly looking for it.

arendst:
Done.

PR #7000, opened 2020-01-08
Add fileCreation/getCreation create-time accessors

### LINKED ISSUE #6992: SD file system, request to add file.CreationDate ().
In version 2.5.0 was a directory of utilities included.
A few years ago I found an example on the internet that made it possible to make the creation date of an SD file visible.
In version 2.6.3, Arduino SdFat Library by William Greiman is still used.  (2009)

**..\2.6.3\librarie\ESP8266\SdFat\src\Fatlib\FatFile.cpp**

This file contains code to process the SD file creation date / time.
Is it possible to add this to the options of the SDFS library?

```cpp
// set up variables using the SD utility library functions:
Sd2Card card;
SdVolume volume;
SdFile root;
File sdFile;

SdFile::dateTimeCallback(dateTime);                  // set callback for SD file date


//=============================================================================
// this function is called by SD if it wants to know the time
void dateTime(uint16_t* date, uint16_t* time)
{
  // return date using FAT_DATE macro to format fields
  *date = FAT_DATE(year(), month(), day());
  // return time using FAT_TIME macro to format fields
  *time = FAT_TIME(hour(), minute(), second());
}
//=============================================================================

void handleSDfilelist() {
  const char* br = "<br>\n";
  message = "<!DOCTYPE html> <html> <head> <title>SD files</title></head> <body>";
  message += "<table style='font-size:14px'><tr><td><b>SD files</b></td></tr>";

root.openRoot(volume);
  dir_t p;
  root.rewind();
  delay(1);  //yield();   //201909 probleem WDT restart oplossen (?)
  while (root.readDir(p) > 0) {
    if (p.name[0] == DIR_NAME_FREE) break;
    // skip deleted entry and entries for . and  ..
    if (p.name[0] == DIR_NAME_DELETED || p.name[0] == '.') continue;
    // only list subdirectories and files
    if (!DIR_IS_FILE_OR_SUBDIR(&p) ) continue;
    message += "<tr><td><a href=\"../";
    for (uint8_t i = 0; i < 11; i++) {
      if (p.name[i] == ' ') continue;
      if (i == 8) {
        message += ".";
      }
      message +=  (char)p.name[i];
    }
    message += "\">";

    // print file name with possible blank fill
    for (uint8_t i = 0; i < 11; i++) {
      if (p.name[i] == ' ') continue;
      if (i == 8) {
        message += ".";
      }
      message += (char)p.name[i] ;
    }
    message += "</a></td><td>";
    message += p.fileSize;
    message += F("</td>");
    SDdatum(p.creationDate, p.creationTime);
    message += ioBuffer;
    message += "   ";
    SDdatum(p.lastWriteDate, p.lastWriteTime);
    message += ioBuffer;
  }
  message += "</tr> \n";

  message += "</table></body></html> \n";
  server.send(200, "text/html", message);
}

```

earlephilhower:
For SDFS and LittleFS, enable a creation time accessor for files and Dir
iterators, similar to the existing fileTime/getLastWrite calls.

Remove spurious Dir::getLastWrite method (the proper and only documented
way is really Dir::fileTime).

Update json to point to new mklittlefs which copies the creation date of
files to the image.

Fixes #6992


cores/esp8266/FSImpl.h
```diff
@@ -51,6 +51,8 @@ class FileImpl {
     // as the FS is allowed to return either the time of the last write() operation or the
     // time present in the filesystem metadata (often the last time the file was closed)
     virtual time_t getLastWrite() { return 0; } // Default is to not support timestamps
+    // Same for creation time.
+    virtual time_t getCreation() { return 0; } // Default is to not support timestamps
```
devyte: The function name seems a bit confusing to me. Consider getCreationTime() to be explicit.

libraries/LittleFS/src/LittleFS.h
```diff
@@ -592,6 +598,26 @@ class LittleFSDirImpl : public DirImpl
...
     }
 
+    uint32_t _getAttr4(char attr) {
+        if (!_valid) {
+            return 0;
+        }
+        int nameLen = 3; // Slashes, terminator
+        nameLen += _dirPath.get() ? strlen(_dirPath.get()) : 0;
+        nameLen += strlen(_dirent.name);
+        char *tmpName = (char*)malloc(nameLen);
```
devyte: While here, I suggest replacing this with new[] into a unique_ptr and dropping the free below

cores/esp8266/FSImpl.h
devyte: Maybe other getters of time values should also get a Time posfix to be clear.

earlephilhower: > ... getters of time values should also get a Time posfix to be clear.

This name is actually an ESP32-ism, part of their API I tried to copy when consolidating the Filesystems.  I suppose we could deprecate this one and direct to a `getLastWriteTime`, but then we'll differ.  I like the Time postfix myself...

libraries/LittleFS/src/LittleFS.h
earlephilhower: Practically the MAXPATH should be <128 bytes.  I should really just make this a stack var with a dynamic size.  C++ is cool that way. :)

earlephilhower: These are now stack temps in the latest push.

cores/esp8266/FSImpl.h
earlephilhower: Updated in latest pushed.

PR #7003, opened 2020-01-08
Updating _port to really used port and made _port accessible

LaborEtArs:
When calling WiFiServer with a 'port = 0', a free port is auto-selected, but this port isn't updated to the '_port' member. In addition, the used port is made accessible.

PR #7004, opened 2020-01-08
Added input_netif to the stored data

LaborEtArs:
- Added a 'setMulticastInterface' method for netifs.
- The input_netif is also only available, while 'inside' of the recv-callback. Added this to the temp-saved data.
- Found problems with omitted UDP messages, therefore disabled the the omitting of messages.
- Made '_pcb' accessible

### DIFF
--- a/libraries/ESP8266WiFi/src/include/UdpContext.h
+++ b/libraries/ESP8266WiFi/src/include/UdpContext.h
@@ -172,8 +172,11 @@ public:
      */
     void setMulticastInterface(netif* p_pNetIf)
     {
-        uint8_t mcast_ifindex = (p_pNetIf ? netif_get_index(p_pNetIf) : 0);
-        udp_set_multicast_netif_index(_pcb, mcast_ifindex);
+#if LWIP_VERSION_MAJOR == 1
+        udp_set_multicast_netif_addr(_pcb, (p_pNetIf ? ip_2_ip4(p_pNetIf->ip_addr) : ip_addr_any));
+#else
+        udp_set_multicast_netif_index(_pcb, (p_pNetIf ? netif_get_index(p_pNetIf) : 0));
+#endif
     }
 
     /*
--- a/libraries/ESP8266WiFi/src/include/UdpContext.h
+++ b/libraries/ESP8266WiFi/src/include/UdpContext.h
@@ -478,7 +478,7 @@ private:
 
 #if LWIP_VERSION_MAJOR == 1
     #define TEMPDSTADDR (&current_iphdr_dest)
-    #define TEMPINPUTNETIF (ip_current_input_netif())
+    #define TEMPINPUTNETIF (current_netif)
 #else
     #define TEMPDSTADDR (ip_current_dest_addr())
     #define TEMPINPUTNETIF (ip_current_input_netif())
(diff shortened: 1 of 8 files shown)


libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -441,24 +462,13 @@ class UdpContext
...
-            int count = 0;
-            for (p = _rx_buf; p && ++count < rxBufMaxDepth*2; p = p->next);
-            if (p)
-            {
-                // pbuf chain too deep, dropping
-                pbuf_free(pb);
-                DEBUGV(":udr\r\n");
-                return;
-            }
-        }
```
d-a-v: did you try to increase `rxBufMaxDepth` instead of removing that block ?

LaborEtArs: No, might also help, but what would be the right value?
What is the reason to omit UDP messages? Memory usage?

LaborEtArs: The 'submodules' problem should exist here too, right?

LaborEtArs: Redid the same process as with the WiFiServer change

d-a-v: Yes, memory usage

LaborEtArs:
Reinserted the code that omits UDP messages

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -167,6 +167,26 @@ class UdpContext
 
 #endif // !LWIP_IPV6
 
+    /*
+     * Add a netif (by its index) as the multicast interface
+     */
+    void setMulticastInterface(netif* p_pNetIf)
+    {
+#if LWIP_VERSION_MAJOR == 1
+        udp_set_multicast_netif_addr(_pcb, (p_pNetIf ? ip_2_ip4(p_pNetIf->ip_addr) : ip_addr_any));
```
d-a-v: `ip_2_ip4()` is useless with lwIP-1.4
https://github.com/esp8266/Arduino/blob/00440cd84a4ed9467de6305cb768875e09c12303/cores/esp8266/IPAddress.h#L33

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -265,7 +290,6 @@ class UdpContext
             // ref'ing it to prevent release from the below pbuf_free(deleteme)
             pbuf_ref(_rx_buf);
         }
-        // remove the already-consumed head of the chain
```
d-a-v: Perhaps it has to be replaced by
"remove parts of the pbuf chain that are not anymore "ref"ed

d-a-v:
There are unwanted submodule updates in this PR
- libraries/LittleFS/lib/littlefs 
- libraries/SoftwareSerial
- tools/sdk/ssl/bearssl 

(see https://github.com/esp8266/Arduino/pull/7003#issuecomment-572148633)

LaborEtArs:
Can't remove these submodules. Every time I do checkout a new branch, these (and other) submodule are added automagically by git. Closing the PR now.

Issue #7008, opened 2020-01-10
Enhancement request: support for Server Sent Events aka EventSource


ewaldc:
This is a request for support of HTML5 Server Sent Events aka [Event Source](https://html.spec.whatwg.org/multipage/server-sent-events.html).
It allows e.g. to implement an event bus over HTML.

The required changes to the ESP8266WebServer library are absolutely minimal: zero impact on existing code and increase in code size/memory footprint of just few bytes (basically sizeof bool).

Essentially, the variable _bool keepCurrentClient_ which is currently hardcoded to _false_, becomes a protected class variable in the ESP8266WebServer class that can be modified via a new function called _keepCurrentClient(bool)_.  

To enable Server Sent Events/event subscription, it requires two web server instances, one being changed to keep the client socket/session (an event listener) open until disconnected. 
```	
ESP8266WebServer _server(port);
ESP8266WebServer _SSEserver(port + 1);
_server.begin();
_SSEserver.keepCurrentClient(true);
_SSEserver.begin();
```
Everything else can implemented in user code (event subscription, event bus keepalive messages, event payload etc.).
I have this currently working on a clean fork from master branch with just this one change.
Happy to post a PR (it's ready to post), if this is a reasonable request.
I can also contribute a working example, I suppose in the ESP8266WebServer examples section.

Thanks once again for this brilliant library and for taking the time to consider this request,
Ewald

d-a-v:
Thanks !
Please post your PR and its example

ewaldc:
@d-a-v,
Example submitted.  It turns out that there is no need for the _keepCurrentClient_ enhancement.
I found another way to make it work: 
1. set the contentLength to _CONTENT_LENGTH_UNKNOWN_ which allows to submit a virtually endless payload (needed for an event bus)
2. add a KeepAlive timer, using the Ticker library, which keeps the socket open by sending e.g. every minute a keep-alive message in case there are no (normal) state update events.  The use of a KeepAlive timer is not a violation of the with SSE - EventSource standard, it just requires the client to handle/ignore the KeepAlive events.

Hence, no need for the enhancement request. Sorry for that, but I hope the example is still useful.  I tried for days to make it work using WebSockets and AsyncWebServer but both solutions are much more complex, consume much more memory and could not be made stable.

[...]

PR #7009, opened 2020-01-10
Support for Server Side Events #7008

ewaldc:
Enhancement request: support for Server Sent Events aka EventSource #7008
Code for review, examples to follow

### DIFF
--- a/package/package_esp8266com_index.template.json
+++ b/package/package_esp8266com_index.template.json
@@ -320,6 +320,13 @@
                      "size": "36871"
                   },
                   {
+                     "host": "i686-pc-linux-gnu",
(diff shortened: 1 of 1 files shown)


earlephilhower:
@ewaldc, looks like you didn't `git add` your changes to `ESP8266WebServer.h` so the new variable you're using is not defined.  Can you add it and re-push?

ewaldc:
@earlephilhower, 
I have come to realize that it is not needed to change the API to support Server Sent Event as I was coding the example.  Hence you may ignore/reject this PR.
I have submitted an SSE example using a separate PR for the ESPWebServer library though.
I may have messed up the example a bit as I forgot to take out my SSID/password (no big harm though, it's a test network with no internet access)

PR #7010, opened 2020-01-11
Hardware WDT Stack Dump Tool

mhightower83:
A Hardware WDT Reset stack dump tool by Michael Hightower

This Sketch demonstrates the use of a tool that prints a stack dump after a Hardware WDT reset. After a Hardware Watchdog Timer reset, the module `hwdt_app_entry.cpp` writes a stack dump to the serial interface.

The goal was to have a file (`hwdt_app_entry.cpp`) that could be dropped into a sketch directory, then open the sketch project, build, upload, and debug.

When the ESP8266 restarts because of a Hardware WDT reset, the serial port speed defaults to 115200 bps. For support of other speeds, review the comments on option `HWDT_UART_SPEED` in `wdt_app_entry.cpp`.

See `hwdt_app_entry.cpp` for more options and details.


d-a-v:
I haven't tried yet but I read the comments.
Did you try `Serial.flush()` instead of `delay()` ?
also `eboot.elf` has grown in size I believe, that should be reflected in the comments

mhightower83:
@d-a-v, Sorry for the confusion, In order to capture and print the stack, I am running as an `app_entry_redefinable` replacement, before the SDK and Core has started. (I really miss the Core APIs.) Fortunately, a UART speed change is not needed, if you are running at the default speed of 115200 and your printing the stack dump for a Hardware WDT reset. It is when I tried to print at a different time or speed that creates the issue. It works for now. I want to go back and try again polling the FIFO for empty then wait 1 character time for the character to leave the shift register, before changing UART speed. That should work; however, I have found that in the same sentence the words "should" and "work" just don't get along very often. :smile: 

eboot.elf's stack usage, with just a flash boot, (w/o running the uncompress OTA logic) appears to be down around 720 bytes. At this time I have a conservative 1024 reserved for ROM and eboot. I think that can come down. Just waiting to see if anything grows.

d-a-v:
Serial speed is generally 74880 at boot time (115200*26/40).
Does your board have a 40MHz crystal ? Should it be mentioned in the comments (115200 vs 74880 according to crystal) ?
By Serial.flush() I indeed intended to wait for a character (10/baud secs).

mhightower83:
@d-a-v, ahh, I tend to over detail things. I think I went too far the other way. I have a 26MHz crystal. I understand that the boot ROM has initially set the UART clock divisor to support 115200 bps with the assumption that it has a 40MHz crystal. For a 26MHz crystal, the resulting error gives us a real rate of 74880 bps and prints jibberish at first until the CPU clock frequency is recalibrated. While CH_PD and EST_RST bring the CPU back to this state of underclocking with a 26MHz crystal, the Hardware WDT reset does not appear to do the same. The UART appears to continues to be clocked at a rate to support a device connected at 115200 bps. And, this allows us to see the famous and cryptic WDT message.

I'll come up with a clearer description so people that are aware of the crystal issue are not distracted by my too brief description.

On the delay for FIFO to clear, as I was thinking this through again, I remembered another factor in just using the delay was to save on IRAM. It only runs at startup, so I opted for keeping code smaller vs faster in this area.

earlephilhower:
@mhightower83 can correct if I'm wrong, but mt $0.02:

> Does it make sense to integrate with the stack dump code that is already implemented?

IMO, no.  This is run at machine startup.  The main stack dump code is run right before a reset and has a very different machine state to work with.

> Or does this replace the current stack dump code?

No, this complements it.  Existing dump code happens with a fully booted and running machine, executed as part of graceful system reset.  This lives at startup right after the ROM calls the main app.

mhightower83:
@earlephilhower thanks, that is a good summary.

Because of the added burden in the way of resources stack, DRAM, and flash space, I always thought of this as a debug option, that would not always be built.

Also, because some of the stack space is zeroed out by the Boot ROM during reset, the default for this module is to run with the NO4KEXTRA option. This keeps the system stack out of the zero-out region. If you know, crashes are not on the system stack this can be turned off to get more DRAM back. In short, there are things that I am doing that you would not always want to be done.

In its current form, it is very self-contained. At the minimum, you take one file drop it into a failing build. Add one line to the user's sketch at the beginning of setup(). Then build and load. Hmm, I think I just repeated the introduction to this PR.

devyte:
@earlephilhower thanks, so no to 2 and 3, which leaves 1: integration with the core. 

@mhightower83 said:
>... you take one file drop it into a failing build. Add one line to the user's sketch at the beginning of setup()

So most of the implementation is in its own file, which I take to be hwdt_app_entry. h/cpp. How about that file gets moved into our core, and we have the user just add the one line in setup(), which I take to be enable_debug_hwdt_at_link_time()? That would pull in the file for compiling, which would trigger the link time magic. The usage would then be consistent with the disable extra 4k() call that @d-a-v implemented. Also, it may allow unification/refactoring of some of the file content with core code, like the g_cont stuff.
What do you think?

mhightower83:
My take on changes that would be needed.

I gave moving `hwdt_app_entry. h/cpp` to core a try.
I currently use `preinit()` in `core_esp8266_main.cpp` so would need to add `hwdt_init()` or something like that to use instead.

I am thinking these current defines may need to change for a more global usage context:
```cpp
// These I think are more likely to be used.
#define DEBUG_HWDT_NO4KEXTRA
#define HWDT_UART_SPEED (115200)

/// These are less likely
#define HWDT_INFO
#define HWDT_PRINT_GREETING
#define ROM_STACK_SIZE (1024)
#define ROM_STACK_DUMP
```
Here is a link to a description of the options: https://github.com/esp8266/Arduino/blob/7288abf1b90d2a081fe1b1e7ead14ae6179c5d0e/libraries/esp8266/examples/HwdtStackDump/hwdt_app_entry.cpp#L123-L235

Defines will have to be specified through `platform.local.txt`.
```txt
compiler.cpp.extra_flags=-DDEBUG_HWDT_NO4KEXTRA -DHWDT_INFO -DROM_STACK_DUMP -DHWDT_UART_SPEED=74880 
```
A strange bug appeared I discovered I duplicated the name HWDR_INFO. I had it as a #define and a structure name. C compiler was fine when it was local to the module. It didn't work so well when defined globally.

Let me know if you want me to do a PR with the move. For now, I am going to do an update for some missed comments, defined, and the duplicate structure/define name.

libraries/esp8266/examples/HwdtStackDump/hwdt_app_entry.cpp
```diff
@@ -334,7 +319,7 @@ extern uint32_t stack_thunk_get_stack_bot() __attribute__((weak, alias("__zero_r
 #if !defined(HWDT_STACK_DUMP_H) || defined(HWDT_VERIFY_HWDT_INFO)
 #define HWDT_STACK_DUMP_H
 
-typedef struct HWDT_INFO {
+typedef struct HWDT_INFO_S {
```
mcspr, line 322: (minor style nitpick)
- since this is an example, to we still need to have `typedef struct`? yes, c++ still supports it, but we can just do `struct name { ... };`
- should the struct name be `HwdtInfo` to avoid confusion with UPPERCASE macros?

mhightower83: @mcspr thanks for the suggestion. 

I agree it is best to avoid confusion with UPPERCASE for macros. My intent was to offer an example with a self-contained debug module(hwdt_app_entry.cpp) that would in effect be supplementing the core for debugging purposes. The example files use camel-case and to set it apart, the debug module uses snake-case, similar to most of the core files I see.  So for the structure, I have matched the style to that I see in the core files.

The intent was that `hwdt_app_entry.cpp` would be more a portable module, not an example.
With the exception of some user-configurable `#defines` at the top of `hwdt_app_entry.cpp`, there should be little need to change the code. A least that was my goal.

mcspr: Thanks for the clarification!

I just noticed the
> A strange bug appeared I discovered I duplicated the name HWDR_INFO. I had it as a #define and a structure name. 

Which is cleared up by the commit after this one. I rest my case :)

d-a-v:
> Defines will have to be specified through platform.local.txt.

This deserves new entries in the [debug menu options](https://github.com/esp8266/Arduino/blob/c3c61a5f751fcc75440e571b5aa89254dce6e4f7/tools/boards.txt.py#L1237).
You are welcome to find a better way of managing the python `options` variable.

mhightower83:
Oops, looks like I have broken the +4K Extra path with my attempt to resolve the toolchain 10.1 change. The NO4KEXTRA still works which because of Boot ROM issues is the better option to use (more recoverable stack space). Several things going on at once will try to fix soon. 

Curious, has anyone tried to used this PR to debug a problem? Did it work for you?

cores/esp8266/hwdt_app_entry.cpp
```diff
@@ -0,0 +1,1051 @@
...
+ *
+ * Using this tool's alternate `app_entry_redefinable()`, we can gain control
+ * before the SDK is started. And dump the contents of the "sys" and "cont"
+ * stacks.
+ *
+ * By making some adjustments to start of the stack pointer, at the entry to
+ * `app_entry_redefinable()`, and also to the stack pointer passed to the SDK,
+ * we can preserve the stack during an HWDT event.
+ *
+ * To use, just place this file in the sketch directory before opening the
```
devyte, line 39: Are these instructions still valid, given that the functionality is now integrated into the core?

mhightower83: Updated.

mhightower83:
I missed some details in meshing with the original extra4k heap feature. I have more integration details to work out.

mhightower83:
Findings from exploring the Extra 4K heap option crash further:

For the possible SYS stack memory range 0x3FFFE000 up to 0x40000000, there is hole at 0x3FFFEA80 up to 0x3FFFEB30 
This region turns to zeros after a WiFi connection is made. It looks like it is a buffer used by `aes_decrypt_init`. It also appears the SDK has a wrapper around the function.

Looking at one of my projects I see the SYS stack high watermark at 0x3FFFE720. This places the hole within the active stack range and yet it does not crash. I assume when the WiFi connection is being done, the actively used stack space is low at that moment. However, when HWDT (with ROM Stack set to 1024) is used with Extra 4K Heap I do crash. My assumption is that the "actively used stack space" is overlapping the hole. In contrast to when it works with reserved ROM Stack set to 752.

mhightower83:
Some HWDT crashes are improved by https://github.com/esp8266/Arduino/pull/7704#issue-520013473. This PR moves them to crashes postmortem can report with much-improved context.

devyte:
@mhightower83 there are conflicts in this one. Could you please resolve them?

devyte (approved):

earlephilhower (approved):
LGTM.  Our debug menu is enormous already, adding a few more entries is no problem.  The replacement startup looks good (if massively over-engineered and over-configurable, but it's mostly hidden if you don't care about it), and only comes into effect in specific debug modes.

Thx again, @mhightower83 !

mhightower83:
@earlephilhower 
> The replacement startup looks good (if massively over-engineered and over-configurable

Yea, I either under-plan or over-plan ahead. Also, with the discovery of how to turn on Flash execution, this PR transitioned from running mostly in IRAM to running mostly from Flash. Assuming we have no need for the IRAM build path, I should go back and trim that out. I don't think features in that build path have kept up with the Flash side. Assuming yes, to avoid introducing new bugs, it might be best to do in a new PR.

On the over-plan side, I do have a generalization in the startup, that I think will allow for a possible user-supplied function (`hwdt_pre_sdk_init()`) that could be used to do post-crash printing of stuff that could aid in analyzing a sketch's state at the time of the HWDT. Also (might need more changes) I was thinking that it would be nice to run a heap integrity check after a previous crash and maybe a full poison check as well. This PR is too big, these are things I have been thinking about. It is something I can look at if you see a need or value.

Anytime you think I have gone too far, feel free to encourage me to come back :smile:

PR #7013, opened 2020-01-12
Fix for windows

kuzfix:
signing.py:
Adding ", shell=True" to line 25 seams to correct an issue with signing a binary on windows. Someone would need to test whether that change brakes it on linux or other OSs.

### DIFF
--- a/tools/signing.py
+++ b/tools/signing.py
@@ -22,7 +22,7 @@ def sign_and_write(data, priv_key, out_file):
     """Save the signed firmware to out_file (file path)."""
 
     signcmd = [ 'openssl', 'dgst', '-sha256', '-sign', priv_key ]
-    proc = subprocess.Popen(signcmd, stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.PIPE)
+    proc = subprocess.Popen(signcmd, stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
     signout, signerr = proc.communicate(input=data)
     if proc.returncode:
         sys.stderr.write("OpenSSL returned an error signing the binary: " + str(proc.returncode) + "\nSTDERR: " + str(signerr))


kuzfix:
Sorry I couldn't reply sooner. "openssl" is an exe. Given the security considerations, it might be better to leave it as it is for now and just add a comment into documentation as a sidenote to windows users, that this change might help. 
Maybe someone with more knowleadge of python than me can figure out what can be done to really fix the issue.

earlephilhower:
I've just installed OpenSSL on my Win10 laptop from the first entry pointed to by the OpenSSL.org pages:  https://slproweb.com/download/Win64OpenSSL_Light-1_1_1d.exe  

I added the installed dir to the environment PATH via System Properties->Environment Variables dialog.

I then built the httpUpdateSigned and it signed the binary just fine:
````
Enabling binary signing
Signed binary: C:\Users\earle\AppData\Local\Temp\arduino_build_510072/httpUpdateSigned.ino.bin.signed
Legacy signed binary: C:\Users\earle\AppData\Local\Temp\arduino_build_510072/httpUpdateSigned.ino.bin.legacy_sig
Executable segment sizes:

IROM   : 311912          - code in flash         (default or ICACHE_FLASH_ATTR) 
IRAM   : 29728   / 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...) 
DATA   : 1272  )         - initialized variables (global, static) in RAM/HEAP 
RODATA : 1420  ) / 81920 - constants             (global, static) in RAM/HEAP 
BSS    : 25728 )         - zeroed variables      (global, static) in RAM/HEAP 

Sketch uses 344332 bytes (32%) of program storage space. Maximum is 1044464 bytes.
Global variables use 28420 bytes (34%) of dynamic memory, leaving 53500 bytes for local variables. Maximum is 81920 bytes.
````

Given that proof point, and the issue with using a shell to run apps vs. a pure exec, I'm closing this for now.

kuzfix:
Thank you for testing this out.

PR #7023, opened 2020-01-19
Unacceptable defaults that destroy servos

dok-net:
Compatibility to AVR Arduino is not premium in this case, as the very much out of spec defaults for the servo PWM timings cause my servos to RUN HOT and PERMANENTLY FAIL.
The defaults can be overridden, but they must be within public specification for unsuspecting initial use!

Still under investigation is the recently patched maximum servo count limit, 12 seems, from current experience with running fan controls and servos, too high - besides, where can one connect 12 servos to an ESP8266 default GPIOs? AFAIK the servo lib does not apply to I2C servo extension boards.

EDIT: Now also contains a fix for servo jerk on detach. Before, detach would cancel the PWM to servo at a random time, which could be interpreted by the servo as an adjustment command to a random angle.


earlephilhower:
You should consider doing this against the main Arduino core if these (imported) defaults are destructive.

https://github.com/arduino-libraries/Servo

libraries/Servo/src/Servo.cpp
```diff
@@ -68,11 +68,11 @@ uint8_t Servo::attach(int pin, uint16_t minUs, uint16_t maxUs)
     _attached = true;
   }
 
-  // keep the min and max within 200-3000 us, these are extreme
+  // keep the min and max within 500-2500 us, these are extreme
```
Makuna: I disagree with this change.  There are servos out there that require the wider range, this is meant to keep outrageous values from being used, not to maintain "safe" values for your servos.

libraries/Servo/src/Servo.h
```diff
@@ -46,10 +46,10 @@
 
 // the following are in us (microseconds)
 //
-#define MIN_PULSE_WIDTH       544     // the shortest pulse sent to a servo  
-#define MAX_PULSE_WIDTH      2400     // the longest pulse sent to a servo 
-#define DEFAULT_PULSE_WIDTH  1500     // default pulse width when servo is attached
-#define REFRESH_INTERVAL    20000     // minumim time to refresh servos in microseconds 
+#define MIN_PULSE_WIDTH       800     // the shortest duty cycle sent to a servo
```
Makuna: what are the values from the standard Arduino servo library?

earlephilhower: The ones in master are the same as in Arduino's Servo library.  This PS restricts the values to a smaller range.

````
#define MIN_PULSE_WIDTH       544     // the shortest pulse sent to a servo  
#define MAX_PULSE_WIDTH      2400     // the longest pulse sent to a servo 
#define DEFAULT_PULSE_WIDTH  1500     // default pulse width when servo is attached
#define REFRESH_INTERVAL    20000     // minumim time to refresh servos in microseconds 
````

File: libraries/Servo/src/Servo.cpp [hunk 3]
```diff
@@ -84,21 +84,20 @@ void Servo::detach()
...
+    _valueUs = DEFAULT_NEUTRAL_PULSE_WIDTH;
     digitalWrite(_pin, LOW);
   }
 }
 
 void Servo::write(int value)
 {
-  // treat values less than 544 as angles in degrees (valid values in microseconds are handled as microseconds)
+  // treat values less than _minUs as angles in degrees (values equal or larger are handled as microseconds)
   if (value < _minUs) {
```
Makuna: what does the AVR Arduino do in this case?  Does it use the_minUs and _maxUs or does it use some other hardcoded values.
What I am getting at here, is the core issue your concerned about is that the angle set and get are calculated act differently (dangerously as you put it) than the standard servo library?

dok-net: @makuna have you looked at the ostensible default from Arduino and how nothing has been done there in 3.5 years over an active issue https://github.com/arduino-libraries/Servo/issues/3 ?
The "safety limits" in the ESP8266 impl are a local invention, btw, I don't see the AVR impl enforcing any such restriction, but the code is obfuscated, probably to conserve RAM (int8_t instead of int) at the expense of CPU time.
Now, what is that default behavior...

libraries/Servo/src/Servo.h
dok-net: @earlephilhower No, no restrictions now in the updated PR, but safer defaults, it is no one's best interest to destroy quality manufacturer (Tower Pro) servos for people that just unsuspectingly try the example for the first time.

libraries/Servo/src/Servo.cpp
Makuna: @dok-net Let me be a little more clear.  Provide more details as to why deviating from the AVR/SAM/etc libraries is a good thing.  What research have you done?  Links to data to back up your statements.  
I wrote my own for AVR that uses a separate timer for each channel so NO ISRs happen; and as you state, I used pulses in the range of 1ms - 2ms with a 1.5ms center as defaults; but mine doesn't limit the user, it just gives them something to play with and lets them keep it in range.(https://github.com/Makuna/PwmServo/blob/master/PwmServo.h)

dok-net: At to what happens, here's the story again, this can happen to anyone with any servo, the results may be more or less severe, but the gist of it is, the defaults should be inside safe limits, finding out what the exact 0° and 180° values are for any given serve is a matter of calibration.

I inspected the sweep example, adopted the initialization code into my own sketch, started it, set a few values for the outputs, among them the servo - my device makes a little noise that can mask the servo's - and found the servo pressing against it's internal mechanical endstop, running hot while doing so, and before I knew it, it stopped working for good.
Mind you, this happens well below requesting 180° with the lib's defaults.

dok-net:
https://asset.conrad.com/media10/add/160267/c1/-/en/000404753DS01/datenblatt-404753-modelcraft-standard-servo-bms-410c-analog-servo-getriebe-material-kunststoff-stecksystem-jr.pdf
https://asset.conrad.com/media10/add/160267/c1/-/en/000275487DS01/datenblatt-275487-bluebird-midi-servo-bms-390-dmh-digital-servo-getriebe-material-metall-stecksystem-jr.pdf

dok-net:
This PR would fix the issue underlying #7084.

Tech-TX:
I'm behind dok-net  on this issue.  The trim pots on my radios would allow up to ~20% deviation from standard timing,  maximum.  The default shouldn't exceed that without good reason as servos use max current when slammed to either stop.  If one or two brands need wider timing,  the users are allowed to recompile with whatever they need,  but the default should always be the 'normal'  range and not significantly wider.  The Arduino library is flawed,  so compatibility with it is still possible with a recompile.

libraries/Servo/src/Servo.cpp
```diff
@@ -74,29 +79,31 @@ uint8_t Servo::attach(int pin, uint16_t minUs, uint16_t maxUs)
...
     // assumed to be 0-180 degrees servo
     value = constrain(value, 0, 180);
-    // writeMicroseconds will contrain the calculated value for us
-    // for any user defined min and max, but we must use default min max
     value = improved_map(value, 0, 180, _minUs, _maxUs);
+  } else if (value > _maxUs) {
+    value = _maxUs;
```
earlephilhower: Good bugfix!, however, this should really be done in `writeMicroseconds` because it's still possible to send in an illegal variable via that API  Move that check into the lower-lever routine (probably ensuring it's >=min and <=max) and you're good.

libraries/Servo/src/Servo.cpp
earlephilhower: I checked the Arduino issue on this and yes, it's been years without any response or motion, and in those years only had 2 thumbs up (one from dok-net, one from another user).  It seems unlikely to move, but frankly that's normal for Arduino stuff.

I'm not really able to find a meaningful datasheet for consumer-type servos.  Some websites report 500-2500 for certain models, while others are reporting 900-2200 for the same exact models.

That said, the only thing changing are the *defaults*, from a wider range to a (safer) narrower one.  `Attach`ing with specific mac/min values is still supported, and given the variability in the servos out there (aka clones) for real apps you're going to need an encoder or a manual tweaking anyway.  

@devyte, @d-a-v, any comments?

libraries/Servo/src/Servo.cpp
dok-net: @earlephilhower This API is a bit silly, now I have to check for _minUs twice, but if we are about protecting servos, let's do it right. I'm also setting the min and max DEFAULTs to 1000 and 2000 per the spec now, damnit :-)

earlephilhower: > ...have to check for minus 2x...

Well, the 1st one in this fcn. is just a bad API choice they made.  "I have an int, so why not silently use it as a degree or as a us value" seems silly vs adding a `writeDegrees` and making everything explicit.  In the lowest level `writeMicroseconds`, value===us always, though, so that's the only spot where "val<minUs" is an error.

Tech-TX:
I haven't checked this latest commit, but with current master and a new-old-stock Futaba FP-S148 servo it buzzes at 180 degrees and heats up; I immediately cut the power rather than burn up the motor.  I just received the Futaba from eBay today for $8, free shipping. ;-)  This was THE standard servo for many years in planes.

I'll re-check the PR Real Soon Now.

Tech-TX:
With the new (old-school) defaults, all 3 of my servos move ~90 degrees end-to-end (0-180 control values), which is pretty common for a flight servo (+/- 45 degrees travel).  Since you're changing an old default, I'd suggest an exposed set of variables so the users can change the desired min/max defaults in their program _without_ modifying the library **servo.h** file, something like:
#define MIN_PULSE_WIDTH 550
#define MAX_PULSE_WIDTH 2330
That's the specified limits for my Futaba S148 servo as an example, with the variable names copied from the Arduino library.

If the user doesn't override the defaults, then it reverts to the new 1000us/2000us limits.

Other than that, no issues found.

dok-net:
@Tech-TX Stan, what are you talking about :-) You must have an outdated version of this PR's code branch.

libraries/Servo/src/Servo.cpp
```diff
@@ -74,35 +79,36 @@ uint8_t Servo::attach(int pin, uint16_t minUs, uint16_t maxUs)
   _maxUs = max((uint16_t)250, min((uint16_t)3000, maxUs));
   _minUs = max((uint16_t)200, min(_maxUs, minUs));
```
dok-net, line 80: @Tech-TX These are the hard-coded limits.

libraries/Servo/src/Servo.h
```diff
@@ -63,8 +63,15 @@ class Servo
...
-    uint8_t attach(int pin, uint16_t min, uint16_t max); // as above but also sets min and max values for writes. 
+    // attach the given pin to the next free channel, sets pinMode, returns channel number or 0 if failure.
+    uint8_t attach(int pin);
+    // attach the given pin to the next free channel, sets pinMode, min, and max values for writes.
+    // returns channel number or 0 if failure.
+    uint8_t attach(int pin, uint16_t min, uint16_t max);
+    // attach the given pin to the next free channel, sets pinMode, min, max values for writes,
+    // and sets the initial value, the same as write.
+    // returns channel number or 0 if failure.
+    uint8_t attach(int pin, uint16_t min, uint16_t max, int value);
```
dok-net, line 74: @Tech-TX Have you tried either of these two overloads of `attach()`?

Tech-TX: I pulled it this morning, Dirk, but I missed the attach() parameters.  :wink:  

I *really* need to add a bunch of *.rst or *.md help files or additions for the stuff in the libraries, especially where things differ from the original Arduino libraries.  That would help minimize confusion for our Gentle Users (and me!).  The current info at ReadTheDocs is woefully incomplete for the servo library.  Here's the full quote:

>**Servo**
This library exposes the ability to control RC (hobby) servo motors. It will support up to 24 servos on any available output pin. By default the first 12 servos will use Timer0 and currently this will not interfere with any other support. Servo counts above 12 will use Timer1 and features that use it will be affected. While many RC servo motors will accept the 3.3V IO data pin from a ESP8266, most will not be able to run off 3.3v and will require another power source that matches their specifications. Make sure to connect the grounds between the ESP8266 and the servo motor power supply.

I didn't know we had 24 I/O pins that could drive servos.  You learn something every day...  and now I need to hunt down the 13 other (missing) GPIO pins. :smile: 

Here's the info on the Arduino library:  https://www.arduino.cc/en/Reference/ServoAttach
so the 8266 library has an INT VALUE option for the initial starting degrees that the Arduino library doesn't.

Your PR is working OK with my test bed and servos.

FedericoBusero:
The detach function now does stopWaveform and delay to have a correct last pulse. This is only a workaround, not a solution: not only it is not recommended to have a delay of 20 ms in such a function, but it also doesn't solve the issue.
When you only have one servo, it works ok. But if you have concurrent analogWrite's going on, the problem still exists. So I guess, the problem should be fixed in the stopWaveform interrupt implementation.

rileyjshaw:
I just stumbled upon this issue. I opened a similar PR in the official Arduino library in April 2019, but sadly there's been no word from the project maintainers. Some of the details from the PR might be relevant / useful here: https://github.com/arduino-libraries/Servo/pull/24.

d-a-v:
Is it a silly idea to:
```
uint8_t Servo::attach(int pin) // leave "deprecated" but do not remove this function
    __attribute__((deprecated("default min/max are 1000/2000, use instead attach(pin, min, max [,init])")));
```
That will produce a useful message and encourage users to make more robust code.

devyte:
After some discussion, @d-a-v 's proposal has merit, and in conjunction with @dok-net proposal in this PR seems to cover all bases. Our choice of different defaults is meant for safety, to not destroy hardware, and the user should look at setting min/max values that fit his specific servo type based on that servo's specs. A warning when using defaults should cover that, and let the user know that specific min/max values should be used.
In this case, deprecation is meant only to warn about incomplete usage. It does NOT mean that the method will be removed in the future, given that it is a method that is currently valid in the Arduino reference.

@dok-net could you please update the PR to include:
- deprecation warning as proposed by @d-a-v 
- comment at the definition of the defaults that explains that we chose defaults different than the Arduino reference because <insert reasons here>, and see discussion <insert url to this discussion here>
- details about the defaults and deprecation in readthedocs [here](https://arduino-esp8266.readthedocs.io/en/latest/libraries.html?highlight=servo#servo)

devyte:
... or rather, [there](https://github.com/esp8266/Arduino/blob/master/doc/libraries.rst#servo).

dok-net:
@devyte Besides looking really ugly to me (which is just a personal opinion, nevermind), the deprecation breaks sample sketches. Additionally, the Arduino AVR Servo lib has pending changes to the same effect as this PR: https://github.com/arduino-libraries/Servo/pull/24, which was mentioned earlier in this thread by @rileyjshaw. I would like to think that the whole matter of being different from AVR is short-lived, we should focus instead on helping the AVR implementation getting updated.

d-a-v:
> the deprecation breaks sample sketches

Your use of words like "ugly" that we are supposed to omit despite the fact that you wrote it on purpose does not help the discussion. Yes, deprecated messages are supposed to be fixed in sketches, that's why examples must be updated.

**That is the purpose**

The vast majority of arduino users do not care at all about warning messages, including deprecated API messages.
Giving this deprecated information will however help those who care to use a more robust API.

Examples are here to show users how to use the API.  Such a proposal will not make anything not compatible anyway.

> I would like to think that the whole matter of being different from AVR is short-lived,

Maybe doing this here, and telling them there will help.
Last time I checked, the issue and the PR don't get frequently updated. With more triggers they will be more likely to come back to it.
Again, ***it would not make us not compatible with them***.

> we should focus instead on helping the AVR implementation getting updated.

.. but you made a PR here instead of highlighting and getting maintainers attention in the PR there.

d-a-v:
After discussion and because deprecation might be too intrusive, there's maybe a more acceptable way to emit a warning when the function is used the first time [when in debug mode](https://github.com/esp8266/Arduino/blob/5d2563eee98ae116dc3d4b6bc14225efb3c85098/cores/esp8266/Print.cpp#L37).
That can be added in a further PR.

Issue #6982, opened 2020-01-04
MDNS.end() exception 28


philbowles:
### Basic Infos

- [x ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x ] I have tested that the issue is present in current master branch (aka latest git).
- [x ] I have searched the issue tracker for a similar issue.
- [x ] If there is a stack dump, I have decoded it.
- [x ] I have filled out all fields below.

#### Platform

- Hardware:				any ESP8266
- Core Version:			2.6.3
- Development Env:		Arduino IDE 1.8.10
- Operating System:		Windows7

### Problem Description

Calling MDNS.end() when already closed causes exception 28. Yes, I know it's not a normal thing to do, but if your code gets "out of step" for some other reason, an exception is a heavy punishment! Can we change it to just silently do nothing, as most other APIs do? For example calling WiFi.disconnect() when already disconnected causes no ill effects 

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>
#include <ArduinoOTA.h>

void setup() {
  Serial.begin(115200);
  Serial.println("Booting");
  WiFi.mode(WIFI_STA);
  WiFi.begin("XXXXXXXX", "XXXXXXXX");
  WiFi.waitForConnectResult();

  ArduinoOTA.begin();
  Serial.println("Ready");
  Serial.print("IP address: ");
  Serial.println(WiFi.localIP());

  Serial.print("Close MDNS OK\n");
  MDNS.end();
  Serial.print("Close Again\n");
  MDNS.end();
  Serial.print("You will never see this message\n");

}

void loop() {
  ArduinoOTA.handle();
}
```

### Debug Messages

```
Booting
Ready
IP address: 192.168.1.106
Close MDNS OK
Close Again

Exception (28):
epc1=0x40207a2b epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffd30 end: 3fffffc0 offset: 01a0
3ffffed0:  3ffe86b7 0000204a 00000000 00000000  
3ffffee0:  4020c438 00000000 0000019d 401009a0  
[...]

PR #7025, opened 2020-01-20
Update LEAmDNS.cpp (fix issue #6982)

LeisureLadi:
Added check 'm_pUDPContext' before calling functions to reset WiFi event callbacks, stop probing ... to fix issue reported under #6982


d-a-v:
Running [this script](https://github.com/esp8266/Arduino/blob/master/tests/restyle.sh) will save you a lot of time.

PR #7027, opened 2020-01-21
Fix WMath's map() implementation for inverse/round-trip mapping

dok-net:
Inspired by Servo.cpp's `improved_map`, this superseded that code, and has better performance (identical to prior WMath implementation).
I expect this is better used to replace the core implementation, then drop "improved_map" from Servo.cpp, than keep it local and have a map() function that may be compatible but works horribly ;-)

### DIFF
--- a/cores/esp8266/WMath.cpp
+++ b/cores/esp8266/WMath.cpp
@@ -70,11 +70,11 @@ long secureRandom(long howsmall, long howbig) {
 }
 
 long map(long x, long in_min, long in_max, long out_min, long out_max) {
-    long divisor = (in_max - in_min);
-    if(divisor == 0){
-        return -1; //AVR returns -1, SAM returns 0
-    }
-    return (x - in_min) * (out_max - out_min) / divisor + out_min;
+    const long dividend = out_max - out_min;
+    const long divisor = in_max - in_min;
+    const long delta = x - in_min;
+
+    return (delta * dividend + (divisor / 2)) / divisor + out_min;
 }
 
 unsigned int makeWord(unsigned int w) {


dok-net:
Here's the sketch for testing correctness and timing:
```
uint32_t expired;
uint32_t iterations;

void setup()
{
	Serial.begin(115200);
	while (!Serial);
	Serial.println("check map");
	expired = 0;
	iterations = 0;
}

constexpr int imin = 0;
constexpr int imax = 180;
constexpr int omin = 100;
constexpr int omax = 200;

int i = 0;
int fails = 0;

void loop()
{
	++iterations;
	i = (i + 1) % (imax + 1);
	auto start = ESP.getCycleCount();
	auto o = map(i, imin, imax, omin, omax);
	expired += ESP.getCycleCount() - start;
	
	if (i != map(o, omin, omax, imin, imax))
	{
		++fails;
	}

	if (iterations > 1000000)
	{
		Serial.printf("Cycles per map(...) = %i, fails = %i\n", expired / iterations, fails);
		expired = 0;
		iterations = 0;
		fails = 0;
	}
}
```

dok-net:
https://github.com/arduino/ArduinoCore-avr/pull/309

cores/esp8266/WMath.cpp
```diff
@@ -70,11 +70,11 @@ long secureRandom(long howsmall, long howbig) {
...
-    long divisor = (in_max - in_min);
-    if(divisor == 0){
-        return -1; //AVR returns -1, SAM returns 0
-    }
-    return (x - in_min) * (out_max - out_min) / divisor + out_min;
+    const long dividend = out_max - out_min;
+    const long divisor = in_max - in_min;
+    const long delta = x - in_min;
+
+    return (delta * dividend + (divisor / 2)) / divisor + out_min;
```
devyte: This doesn't handle divisor == 0.

dok-net:
This today got merged into Arduino Core for ESP32, https://github.com/espressif/arduino-esp32/pull/3655

PR #7029, opened 2020-01-21
Code size optimisation of ESP.getResetReason()

dirkmueller:
doing if/else snakes for something that is a switch/case
is wasteful, as it repeatedly evaluates the same if() condition.
Also repeating strcpy_P is adding code bloat.

This simplification reduces size from 111 to 41 bytes.

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -485,7 +485,7 @@ String EspClass::getResetReason(void) {
         case REASON_DEEP_SLEEP_AWAKE: buff = F("Deep-Sleep Wake"); break;
         // // external system reset
         case REASON_EXT_SYS_RST:      buff = F("External System"); break;
-        default:                      buff = F("Unknown");
+        default:                      buff = F("Unknown"); break;
     }
     return String(buff);
 }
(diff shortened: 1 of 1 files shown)


cores/esp8266/Esp.cpp
```diff
@@ -468,23 +468,24 @@ bool EspClass::checkFlashCRC() {
...
+        case REASON_EXCEPTION_RST:    buff = F("Exception"); break;
+        // software watch dog reset, GPIO status won’t change
+        case REASON_SOFT_WDT_RST:     buff = F("Software Watchdog"); break;
+        // software restart ,system_restart , GPIO status won’t change
+        case REASON_SOFT_RESTART:     buff = F("Software/System restart"); break;
+        // wake up from deep-sleep
+        case REASON_DEEP_SLEEP_AWAKE: buff = F("Deep-Sleep Wake"); break;
+        // // external system reset
+        case REASON_EXT_SYS_RST:      buff = F("External System"); break;
+        default:                      buff = F("Unknown");
```
earlephilhower: Could we add a `break` here on the default case, please?  Just in case someone comes in and reworks things and we wind up with a case after default in the list.  More code hygiene than anything wrong with the code as-is.

dirkmueller: done. Not sure we would want to encourage default: not being the last case..

PR #7030, opened 2020-01-21
Reduce mem footprint of ESP.getResetInfo()

dirkmueller:
This function has excessively long datastrings that can
better be stored in flash, reducing runtime memory footprint.

Also detailed formatting only makes sense when there is an
exception or a watchdog. In other cases instead of printing
an unhelpful "flag: 0" we can just return the ResetReason, which
is much more readable.

Saves about 120 bytes of (data) memory.


fsommer1968:
Thanks for your effort reducing memory footpront 👍🏻

PR #7036, opened 2020-01-26
Fix/enable UDP packet reassembly

szekelyisz:
UdpContext didn't care about pbuf chaining when receiving datagrams, leading
to fragments delivered to the application as individual packets.


szekelyisz:
Thanks @d-a-v! Yeah I scratched my head a little because I found out in the meantime that reassembly is completely disabled in the LwIP 1.4 binary (although the feature is present in the source). Still I chose this approach for two reasons. First, having `#if`s around every reference of `pbuf->tot_len` or `pbuf->len` is ugly. Second, someone could choose to recompile LwIP 1.4 with reassembly support, and it would be impolite to leave them with the original bug intentionally.

I came across this bug when I wanted to send data for WS2812 LED strips encoded in OSC packets as blobs. The usual MTU of 1500 bytes would leave me with around maximum 470 LEDs per strip, so I was wondering if I could use IP-level fragmentation to squeeze more data into a single datagram and avoid application-level fragmentation/reassembly. After making sure that LwIP had this feature and worked correctly I realized the bug was in the Arduino code. Now I can receive data for thousands of LEDs.

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -238,6 +238,8 @@ class UdpContext
         }
 
         auto deleteme = _rx_buf;
+
+        while(_rx_buf->len != _rx_buf->tot_len) _rx_buf = _rx_buf->next;
```
devyte: while body on its own line please

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -287,22 +289,26 @@ class UdpContext
...
             return 0;
 
-        size_t max_size = _rx_buf->len - _rx_buf_offset;
+        size_t max_size = _rx_buf->tot_len - _rx_buf_offset;
         size = (size < max_size) ? size : max_size;
-        DEBUGV(":urd %d, %d, %d\r\n", size, _rx_buf->len, _rx_buf_offset);
+        DEBUGV(":urd %d, %d, %d\r\n", size, _rx_buf->tot_len, _rx_buf_offset);
+
+        void* buf = pbuf_get_contiguous(_rx_buf, dst, size, size, _rx_buf_offset);
+        if(!buf) return 0;
```
devyte: if body on its own line please

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -287,22 +289,26 @@ class UdpContext
...
+
+        if(buf != dst) memcpy(dst, buf, size);
```
devyte: memcpy on its own line please

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -522,6 +528,90 @@ class UdpContext
...
+    /*
+     * Code in this conditional block is copied/backported verbatim from
+     * LwIP 2.1.2 to provide pbuf_get_contiguous.
+     */
+
+    static const struct pbuf *
+    pbuf_skip_const(const struct pbuf *in, u16_t in_offset, u16_t *out_offset)
+    {
+      u16_t offset_left = in_offset;
+      const struct pbuf *q = in;
```
devyte: what is q? please either add a comment, or rename to something meaningful (e.g. it which is typical of iterator, or similar)

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -522,6 +528,90 @@ class UdpContext
         reinterpret_cast<UdpContext*>(arg)->_recv(upcb, p, srcaddr, srcport);
     }
 
+#if LWIP_VERSION_MAJOR == 1
+    /*
+     * Code in this conditional block is copied/backported verbatim from
+     * LwIP 2.1.2 to provide pbuf_get_contiguous.
+     */
+
+    static const struct pbuf *
```
devyte: It looks to me like this method doesn't have to be in the class definition. I suggest moving to the .cpp, maybe even move it out of the class altogether.
Of course, the other methods that make use of it should be moved as well.

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -522,6 +528,90 @@ class UdpContext
...
+        q = q->next;
+      }
+      if (out_offset != NULL) {
+        *out_offset = offset_left;
+      }
+      return q;
+    }
+
+    u16_t
+    pbuf_copy_partial(const struct pbuf *buf, void *dataptr, u16_t len, u16_t offset)
```
devyte: Does this have to be in the class definition? Can it be moved out into the cpp?

libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -522,6 +528,90 @@ class UdpContext
...
+          left = (u16_t)(left + buf_copy_len);
+          len = (u16_t)(len - buf_copy_len);
+          offset = 0;
+        }
+      }
+      return copied_total;
+    }
+
+    void *
+    pbuf_get_contiguous(const struct pbuf *p, void *buffer, size_t bufsize, u16_t len, u16_t offset)
```
devyte: Can this be moved to the cpp?

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: Fixed.

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: Fixed.

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: Fixed.

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: This code was copy-pasted from LwIP2 source, nevertheless I renamed it to`pbuf_it`.

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: There is no cpp. UdpContext is a header-only class. These 2 methods are used exclusively in this class, so I believe it's better to enclose it inside and not pollute the global namespace.

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: Again, same as above.

libraries/ESP8266WiFi/src/include/UdpContext.h
szekelyisz: Again, same as above.

libraries/ESP8266WiFi/src/include/UdpContext.h
devyte: I meant create one and move this in there. But there is additional cleanup to do here, so let's just leave this as is.

PR #7040, opened 2020-01-28
Extend Print class for 64bit integers.

dok-net:
@d-a-v We've talked about it yesterday. Even if PR'ing AVR Arduino is a good idea to do first|also, could you give this an eye, please?
Compiled size is not affected if not used with (u)int64_t, if print(int64_t) gets used, 288 bytes IROM are used additionally.

### DIFF
--- a/cores/esp8266/Print.cpp
+++ b/cores/esp8266/Print.cpp
@@ -181,12 +181,6 @@ size_t Print::print(double n, int digits) {
     return printFloat(n, digits);
 }
 
-size_t Print::println(const __FlashStringHelper *ifsh) {
-    size_t n = print(ifsh);
-    n += println();
-    return n;
-}
-
 size_t Print::print(const Printable& x) {
     return x.printTo(*this);
 }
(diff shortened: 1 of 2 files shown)


dok-net:
On AVR it's different from Xtensa, I've now optimized both PRs for size, thanks for asking the question.

cores/esp8266/Print.cpp
```diff
@@ -261,6 +283,28 @@ size_t Print::printNumber(unsigned long n, uint8_t base) {
...
+    char* str = &buf[sizeof(buf) - 1];
+
+    *str = '\0';
+
+    // prevent crash if called with base == 1
+    if (base < 2)
+        base = 10;
+
+    do {
+        unsigned long m = n;
```
devyte: Shouldn't this be unsigned long long?

cores/esp8266/Print.cpp
```diff
@@ -261,6 +283,28 @@ size_t Print::printNumber(unsigned long n, uint8_t base) {
...
         char c = m - base * n;
+
+        *--str = c < 10 ? c + '0' : c + 'A' - 10;
+    } while(n);
+
+    return write(str);
+}
+
+size_t Print::printNumber(unsigned long long n, uint8_t base) {
+    char buf[8 * sizeof(long long) + 1]; // Assumes 8-bit chars plus zero byte.
```
devyte: Prefer sizeof(n) here

cores/esp8266/Print.cpp
```diff
@@ -261,6 +283,28 @@ size_t Print::printNumber(unsigned long n, uint8_t base) {
...
         n /= base;
         char c = m - base * n;
+
+        *--str = c < 10 ? c + '0' : c + 'A' - 10;
+    } while(n);
+
+    return write(str);
+}
+
+size_t Print::printNumber(unsigned long long n, uint8_t base) {
```
devyte: I think it would be best to unify this in a single template method where the type of n is the template parameter. It will reduce mistakes. Something like:
```cpp
template <typename uintType>
size_t Print::printNumber(uintType n, uint8_t base) {
    char buf[8 * sizeof(n) + 1]; // Assumes 8-bit chars plus zero byte.
...
    do {
        uintType m = n;
        ...
    }
...
}
```

dok-net: Not really, this might require specializations anyway, maybe not on the ESP8266 right now, but if the compiler intrinsics change, I don't know about the CPU instruction set? uint32_t and uint64_t modulo are implemented differently on the other boards.

devyte: If other boards require specializations, they can do that, or they can reimplement this, or something else. We're on the ESP8266, so we focus on the ESP8266. 
The intrinsics aren't going to change. It's a 32bit platform, and our build system is defined. Also, templatization should cover that anyways even if it does happen.

devyte:
@dok-net just the one change and this is good from me.

dok-net:
Built binary size remains unchanged.

dok-net:
This today got merged in Arduino Core for ESP32 master, https://github.com/espressif/arduino-esp32/pull/3688

devyte (approved):

PR #7044, opened 2020-01-29
Pull GPIO initialization into its own 'weak' function.

### LINKED ISSUE #7041: GPIOs get re-init during software restarts
I suggest to remove the lines:

```
  for (int i = 0; i <= 5; ++i) {
    pinMode(i, INPUT);
  }
  // pins 6-11 are used for the SPI flash interface
  for (int i = 12; i <= 16; ++i) {
    pinMode(i, INPUT);
  }
```
in: 
https://github.com/esp8266/Arduino/blob/9e9515b49fe635767afb1ae5c8a2f61694ffd794/cores/esp8266/core_esp8266_wiring_digital.cpp#L238

or maybe check the reset reason and only init the pins if `resetInfo.reason == REASON_DEFAULT_RST` ?

opinions?

everslick:
By pulling GPIO init into its own weak function, it can be overridden by the user. This is important in cases when GPIOs should not toggle during reboot, exceptions or other crashes. Fixes #7041.


d-a-v:
@everslick 
Will you fix this error ?
```
core_esp8266_wiring_digital.cpp: In function 'void initPins()':
core_esp8266_wiring_digital.cpp:254:13: error: 'resetPins' was not declared in this scope
   resetPins();
             ^
```

everslick:
I think it is only the missing prototype `void resetPins();` in wiring_private.h. I submitted the PR via the github online editor, which only allows me to change ONE file per PR (I think).

Unfortunately I'm not at my workstation for at least 3 days. If it is not a big hassle to you I'd kindly ask you (@d-a-v) to add the line this time.

everslick:
Looks good! :-)

PR #7047, opened 2020-01-29
Fixes eboot.elf CS issue

mhightower83:
The CS issue is only seen when you try to use an `eboot.elf` that is built from scratch.
The eboot.elf checked in works fine. The `.text` section is significantly smaller than the one built from scratch.

Added dependencies for eboot.ld and Makefile to Makefile.
Updated eboot.ld to not fill with zeros through the CS field
on its way to the CRC.
Added size test to elf2bin.py

Edited: clarification


bootloaders/eboot/eboot.ld
```diff
@@ -154,12 +154,24 @@ SECTIONS
...
+/*
+The filling (padding) and values for _crc_size and _crc_val are handled by
+elf2bin.py. With this, we give values to the symbols without explicitly
+assigning space. This avoids the linkers back *fill* operation that causes
+trouble.
+
+The CRC info is stored in last 8 bytes.    */
+    _crc_size = _stext + 4096 - 8;
+    _crc_val = _stext + 4096 - 4;
+    ASSERT((4096 > (17 + (. - _stext))), "Error: No space for CS and CRC in bootloader sector.");
```
earlephilhower: You calculated `_free_space` above, why not use it at as the assertion (`_free_space >= 0`) here?

bootloaders/eboot/eboot.ld
```diff
@@ -154,12 +154,24 @@ SECTIONS
...
-    . = _stext + 4096 - 8;
-    _crc_size = .;
-    . = . + 4;
-    _crc_val = .;
+    _free_space = 4096 - 17 - (. - _stext);
+/*
+The boot loader checksum must be before the CRC, which is written by elf2bin.py.
+This leaves 16 bytes after the checksum for the CRC placed at the end of the
+4096-byte sector. */
+    _cs_here = (ALIGN((. + 1), 16) == ALIGN(16)) ? (ALIGN(16) - 1) : (. + 0x0F);
```
earlephilhower: Sorry, but I don't follow what exactly is here.  The ROM segment checksum is placed outside of the ELF address space.  In the flash it follows the first sector's data (always 4096 bytes).  What is _cs_here representing, and who is setting/using it?

bootloaders/eboot/eboot.ld
mhightower83: I thought I tried that and could not get it to work, but wasn't sure. I tried it again, it looks like the numbers are unsigned and always true. I created an over the limit case, with _free_space at `0xfffffffffffffffe` the assert passed. Is there something special I can do to make the compare signed?

bootloaders/eboot/eboot.ld
mhightower83: Sorry, my experience with the gnu linker is lacking. The goal of this ugly expression was to calculate a location for the CS that landed just outside the `.text` section at the end of a 16 bit aligned. Its purpose was for a human cross-check. I used `hexdump` to view the .bin file created by `elf2bin.py`. I verified that CS and CRC were in the right places.

> In the flash it follows the first sector's data (always 4096 bytes).

Sorry, I am not sure I am parsing that correctly. So I maybe miss answering your question.
I could not find any detailed documentation on what the Boot ROM expected. So I resorted to the empirical method. From that, the Boot ROM expects a checksum at the end of the 16-byte alignment immediately following the last segment's payload. Since the CRC is at the end of the 4096-byte sector, the CS needs to at least proceed it. Otherwise when the CRC is written the CS will be in error.

bootloaders/eboot/eboot.ld
earlephilhower: On second though, I don't think it can.  The `free_space` is a linker symbol, not a variable. And there are no such tings as negative addresses.  So, NP here.

bootloaders/eboot/eboot.ld
earlephilhower: Got it.  Didn't understand the reasoning before.  This seems fine and fixes something I overlooked.

PR #7048, opened 2020-01-31
Additions to UdpContext needed for LEAmDNS2

LaborEtArs:
- Added 'setMulticastInterface(netif*)'
- Added a getter for the used 'udp_pcb' -> 'udp_pcb* pcb()'
- The information about the originating netif is also only available while processing 'onRx'. Added the netif to the temp stored data and added a getter for this.


libraries/ESP8266WiFi/src/include/UdpContext.h
```diff
@@ -167,6 +167,26 @@ class UdpContext
 
 #endif // !LWIP_IPV6
 
+    /*
+     * Add a netif (by its index) as the multicast interface
+     */
+    void setMulticastInterface(netif* p_pNetIf)
+    {
+#if LWIP_VERSION_MAJOR == 1
+        udp_set_multicast_netif_addr(_pcb, (p_pNetIf ? ip_2_ip4(p_pNetIf->ip_addr) : ip_addr_any));
```
d-a-v: FYI `ip_2_ip4(x)` is like `(x)` [with lwIP-v1](https://github.com/esp8266/Arduino/blob/a2141803f11d91a5654c7a7065b552302cac527c/cores/esp8266/IPAddress.h#L33) (because IPv6 does not exist in it)

LaborEtArs: Updated the PR

PR #6979, opened 2020-01-03
Stream::send()

d-a-v:
*edit*:

fixes #6907 (datasource is replaced by stream) ([comment](https://github.com/esp8266/Arduino/issues/6907#issuecomment-633109307))

*edit*:

References to "String is Stream" are removed:
An helper class `S2Stream myStringVarNowIsAStream(myStringVar)` has been added to make an already existing `String` into a `Stream`, and `StreamString` derives from it.

*edit*:

- \+ [device test](https://github.com/d-a-v/Arduino/blob/streamto/tests/device/test_sw_StreamString/test_sw_StreamString.ino) combined with [an example](https://github.com/d-a-v/Arduino/blob/streamto/libraries/esp8266/examples/StreamString/StreamString.ino)

- \+ [documentation](https://github.com/d-a-v/Arduino/blob/streamto/doc/reference.rst#streams)

*edit*:

- `WiFiEcho` server sketch example, with provided python echo client:

| code | BW (Kibits/s) | min. free stack (Bytes) |
| ----- | ----- | ----- |
| [byte by byte](https://github.com/d-a-v/Arduino/blob/streamto/libraries/ESP8266WiFi/examples/WiFiEcho/WiFiEcho.ino#L98L103) | 539 | 2864 |
| [512 bytes stack buffer](https://github.com/d-a-v/Arduino/blob/streamto/libraries/ESP8266WiFi/examples/WiFiEcho/WiFiEcho.ino#L108L119) | 4295 | 2212 |
| [Stream::to()](https://github.com/d-a-v/Arduino/blob/streamto/libraries/ESP8266WiFi/examples/WiFiEcho/WiFiEcho.ino#L127) | 4980 | 2576  |

*(these results are environment-dependent and have been made under the same conditions)*

- [Optional virtual peekBuffer API to allow direct access to data](https://github.com/d-a-v/Arduino/blob/bf9ad754ea0ab57133ac1147fce4d5d6bd2c95f4/cores/esp8266/Stream.h#L115L139) (internal Stream API, currently implemented by HardwareSerial and WiFiClient/Secure)

- [API addition is summarized here](https://github.com/d-a-v/Arduino/blob/bf9ad754ea0ab57133ac1147fce4d5d6bd2c95f4/cores/esp8266/Stream.h#L141L199) (user facing API)

Additional notes:
[...]

PR #7051, opened 2020-01-31
Fixes large payload issues for sendRequest

### LINKED ISSUE #7049: ESP8266HTTPClient issue in library
Message bigger than 1460 bytes can't be send.
ESP8266HTTPClient.CPP

Problem is in:
<pre>
Function: int HTTPClient::sendRequest(const char * type, const uint8_t * payload, size_t size)

// send Payload if needed
        if (payload && size > 0) {
            size_t bytesWritten = 0;
            const uint8_t *p = payload;
            while (bytesWritten < size) {
                int written;
                int towrite = std::min((int)size, (int)HTTP_TCP_BUFFER_SIZE);
                written = _client->write(p + bytesWritten, towrite);
                if (written < 0) {
                     return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
                } else if (written == 0) {
                     return returnError(HTTPC_ERROR_CONNECTION_LOST);
                }
                bytesWritten += written;
                size -= written;
            }
        }
</pre>

Problematic is: 
`size -= written;` - because SIZE is subtracting every loop and then in WHILE is compared to overal bytes sent. Only first chunk 1460 bytes can be send.

Solution:
`size -= written;` - this should be erased.
`bytesWritten += written;` - this is enough for this function.

Proper code is:
<pre>
// send Payload if needed
        if (payload && size > 0) {
            size_t bytesWritten = 0;
            const uint8_t *p = payload;
            while (bytesWritten < size) {
                int written;
                int towrite = std::min((int)size, (int)HTTP_TCP_BUFFER_SIZE);
                written = _client->write(p + bytesWritten, towrite);
                if (written < 0) {
                     return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
                } else if (written == 0) {
                     return returnError(HTTPC_ERROR_CONNECTION_LOST);
                }
                bytesWritten += written;
                //size -= written;
            }
        }
</pre>

ramirocarra:
sendRequest has a major problem when sending a big payload, the comparator in the IF loop has its two operators changed at the same time, so the last part of payload is never sent. This is seen when you try to send a payload that needs more than one operation.

The modification has been tested for a 1500 bytes payload.

*edit from maintainer:* fixes #7049

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -681,7 +681,8 @@ int HTTPClient::sendRequest(const char * type, const uint8_t * payload, size_t s
         if (payload && size > 0) {
             size_t bytesWritten = 0;
             const uint8_t *p = payload;
-            while (bytesWritten < size) {
+            size_t originalSize = size;
+            while (bytesWritten < originalSize) {
                 int written;
                 int towrite = std::min((int)size, (int)HTTP_TCP_BUFFER_SIZE);
                 written = _client->write(p + bytesWritten, towrite);


ramirocarra:
commenting `size -= written` may work, but that will make the "amount to write" wrong for the last chunk, in line: 

`int towrite = std::min((int)size, (int)HTTP_TCP_BUFFER_SIZE);`

It may take some random info from payload pointer

d-a-v:
That is very true. But `HTTP_TCP_BUFFER_SIZE` is totally useless and we should have
`towrite = size - byteswritten` there. This is the kind of code that is coded and recoded many times through the core than can be simplified by [one transfer function for all](https://github.com/esp8266/Arduino/pull/6979/files#diff-15a79083ae1917043eb1d24feccca784L685).

(I'm still approving your PR, other maintainers will comment)

ramirocarra:
That´s right too. I haven´t ckeck the code in function `_client->write()` but i guess it will take only the amount of info the buffer size allows (keeping `HTTP_TCP_BUFFER_SIZE` is an unnecessary double verification?)

d-a-v:
> keeping HTTP_TCP_BUFFER_SIZE is an unnecessary double verification?

It's more of a History/Legacy leftover.

devyte:
This will likely get simplified with #6979 , but merging for the immediate need.

TD-er:
Is it possible this also may fix an issue I was having with running out of memory after a number of failed connection attempts to an MQTT broker?

d-a-v:
This is unrelated because pubsubclient does not use HTTPClient.

TD-er:
OK, thanks, that does limit my search to the real problem that may have been fixed now :)

PR #7060, opened 2020-02-04
PoC cache configuration control

mhightower83:
## MMU - Adjust the Ratio of ICACHE to IRAM
The Arduino IDE Tools menu has a new option, `MMU`.

Possible selections are:
 1.  `32KB cache + 32KB IRAM (balanced)`
 1.  `16KB cache + 48KB IRAM (IRAM)`
 2.  `16KB cache + 48KB IRAM and 2nd Heap (shared)` 
      * The 2nd heap size will vary with free IRAM
      * Enables Non-32-Bit Access for IRAM by default. Required for `umm_malloc` library to work.
      * The 2nd heap is supported by the standard `malloc` APIs.
      * Heap selection is handled through a `HeapSelect` class.
 1.  `16KB cache + 32KB IRAM + 16KB 2nd Heap (not shared)`
      * Not managed by `umm_malloc` library
      * Non-32-Bit Access for IRAM must be enabled separately

New build defines and possible values.These are the results of the menu options described above:
|`#define`|balanced|IRAM|IRAM and Heap<br>shared|IRAM and Heap<br>not shared|
|---|:--:|:---:|:---:|:---:|
|`MMU_IRAM_SIZE`|`0x8000`|`0xC000`|`0xC000`|`0x8000`|
|`MMU_ICACHE_SIZE`|`0x8000`|`0x4000`|`0x4000`|`0x4000`|
|`MMU_IRAM_HEAP`| -- | -- |defined, enables<br>`umm_malloc`| -- |
|`MMU_SEC_HEAP`| -- |`> _text_end` ** | `> _text_end` ** |`0x40108000`|
|`MMU_SEC_HEAP_SIZE`| -- | variable ** | variable ** |`0x4000`|

** These defines are to inline functions that calculate the values, based on unused code space.

IRAM, unlike DRAM, must be accessed as aligned full 32-bit words, no byte or short access.
I assume pgm_read macros would work; however, the store operation would remain an issue. ets_memcpy - appears to work well as long as byte count is rounded up to be evenly divided by 4.
## Non-32-Bit Access
Pulled in work from earlephilhower's PR https://github.com/esp8266/Arduino/pull/6978, updated/refactored to handle writes to iRAM and more. This allows word and byte access to iRAM through a load/store exception handler. This would best be used, for infrequently accessed data. Expect it to be very slow, each character access will require a complete save and restore of all 16+ registers.

The Arduino IDE Tools menu has a new option, `Non-32-Bit Access`.

Selections are:
* `Use pgm_read macros for IRAM/PROGMEM`
* `Byte/Word access to IRAM/PROGMEM (very slow)`

To get a sense of how memory access time is effected, see examples `MMU48K` and `irammem` in `ESP8266`.

#### For calls to `umm_malloc` with interrupts disabled.
* `malloc` will always allocate from `DRAM` when called with interrupts disabled.
* `realloc` will fail if not built with `USE_ISR_SAFE_EXC_WRAPPER` defined. 
  * The current build has `USE_ISR_SAFE_EXC_WRAPPER` defined in `mmu_iram.h`.
* `realloc` requests that require `malloc` to complete, will allocate from `DRAM`.

#### ISR/Exception Handler Issue
The non-32-bit exception handler is called by a "C" wrapper function in ROM. This ROM function enables interrupts before calling our registered handler. Defining `USE_ISR_SAFE_EXC_WRAPPER` in `mmu_iram.h` will install a replacement that does not enable interrupts (now default). The effects on Network performance are unknown.

To keep ISR execution time with interrupts disabled at a minimum, avoid the use of IRAM from ISRs. Especially the use of non-32-bit read/writes on IRAM.

#### How to Select Heap
The `MMU` selection `16KB cache + 48KB IRAM and 2nd Heap (shared)` allows you to use the standard heap API function calls (`malloc`, `calloc`, `free`, ... ). to allocate memory from DRAM or IRAM. The selection can be made by instantiating the class `HeapSelectIram` or `HeapSelectDram`.The usage is similar to that of the `InterruptLock` class. The default/initial heap source is DRAM. The class is in `umm_malloc/umm_malloc.h`
```cpp
...
    char *bufferDram;
    bufferDram = (char *)malloc(33);
    char *bufferIram;
    {
        HeapSelectIram ephemeral;
        bufferIram = (char *)malloc(33);
    }
...
    free(bufferIram);
    free(bufferDram);
...
```
Low level functions for selecting a heap. These are used by the above Classes:
* `umm_get_current_heap_id()`
* `umm_set_heap_by_id( ID value )`
* Possible ID values
   * `UMM_HEAP_DRAM`
   * `UMM_HEAP_IRAM`
   * ~`UMM_HEAP_EXTERNAL`~ (code present in umm_malloc only, not enabled)

Also, APIs added from earlephilhower's PR https://github.com/esp8266/Arduino/pull/6978 are:
* `ESP.setIramHeap()` Pushes current heap onto a stack and sets IRAM heap.
* `ESP.setDramHeap()` Pushes current heap onto a stack and sets DRAM heap.
* `ESP.resetHeap()` Restores previously pushed heap.

Updated to reflect current features in the PR.


earlephilhower:
Very cool, and probably of much more general use than my virtual memory setup (although being able to malloc an 8MB block was kind of neat...).  I heard the impact to performance was minimal in your tests which is even more amazing.

One thing I see you didn't pull in is the changes I did to make the memory accessible.  When using the non32b exception handler, UMM can manage the memory as just another heap, allowing users to malloc() and new() with impunity.  I think that would make it much more usable for general folks.

The interface could be the stack one I tried (worked well for me and let me hack libs to silently use internal/fast or external/slow), or @devyte was thinking more of a single flag (normal/iram/external) for simplicity.

Any thought from you on malloc interfaces?

Also, PlatformIO is not liking the .ld->.h conversions you've done.  I can't quite grok what's wrong and have never been able to get a working general PIO install running on my system, but you might want to take a look.

mhightower83:
@earlephilhower Thanks for the kind words.

> One thing I see you didn't pull in is the changes I did to make the memory accessible. When using the non32b exception handler, UMM can manage the memory as just another heap, allowing users to malloc() and new() with impunity. I think that would make it much more usable for general folks.

Yes, that was in a different PR that I meant to go back and look at. I was hoping that I could adapt UMM to do that when I added the non32b; however, I think I see an issue in the ROMs `_xtos_c_wrapper_handler` for processing exceptions. It does a `rsil 0` a little bit before calling the registered exception handler. 

I also had trouble with the Hello Bear example when I fixed it up with an iRAM stack directly. My memory has faded on the specifics, I don't remember the kind of crash. I suppose it could be interrupt related. --- Just looked through your PR --- Not sure why I had trouble and you didn't with the Bear stuff. I'll have to try that again. 

I am also thinking that UMM should revert back to internal DRAM, anytime it is called with interrupts disabled. That would result in all the iRAM allocations occurring in the foreground and maybe for exception processing disable IRQs at the start and leave it to the exit logic to restore PS.

> Any thought from you on malloc interfaces?

Not sure. While simplicity is usually best, this is not looking very simple. DRAM, IRAM, FLASH, and external SRAM each have different performance properties and concerns. For now, I'll look at merging in your umm_malloc changes.

>  PlatformIO

Yea I didn't know what to do about that. I have an idea now that might work better w/o all the renaming to `.h`, I'll push it in a couple of days.

earlephilhower:
> I also had trouble with the Hello Bear example when I fixed it up with an iRAM stack directly. My memory has faded on the specifics, I don't remember the kind of crash. I suppose it could be interrupt related. --- Just looked through your PR --- Not sure why I had trouble and you didn't with the Bear stuff. I'll have to try that again.

I was unable to get the SSL *stack* in external SRAM working, and I doubt that you'll be able to make it work in IRAM, either.  There is no problem other than it's too slow and accessed on non-word boundaries, and accessed too intensely.  I get a WDT w/external SRAM for the BSSL stack, and I would bet that's what you're seeing too.

I actually moved the 17KB SSL *buffer* (i.e. the per-connection info) to SRAM and that ran fine and any perf. difference was undetectable.  I also moved the String() allocator to it, too, and also had no issue in anything I tested.

It's probably a matter of adding more optimistic_yield()s in the library, but I didn't look into it.

mcspr:
Re eb9882e. I don't see any errors showing up in the CI... if that's the reason for confusion about PIO issues, here some locally rolled back to https://github.com/esp8266/Arduino/pull/7060/commits/91fc39130e6b39d3f4cf25c3956fda07a7d5c981. Paths are cleaned up for readability:
```
cores/esp8266/exc-c-wrapper-handler.S: Assembler messages:
cores/esp8266/exc-c-wrapper-handler.S:99: Error: literal pool location required for text-section-literals; specify with .literal_position
cores/esp8266/exc-c-wrapper-handler.S:123: Error: literal pool location required for text-section-literals; specify with .literal_position
cores/esp8266/exc-c-wrapper-handler.S:99: Error: literal pool location required for text-section-literals; specify with .literal_position
```

tools/platformio-build.py
```diff
@@ -111,7 +111,8 @@ def scons_patched_match_splitext(path, suffixes=None):
         join(FRAMEWORK_DIR, "tools", "sdk", "include"),
         join(FRAMEWORK_DIR, "tools", "sdk", "libc",
              "xtensa-lx106-elf", "include"),
-        join(FRAMEWORK_DIR, "cores", env.BoardConfig().get("build.core"))
+        join(FRAMEWORK_DIR, "cores", env.BoardConfig().get("build.core")),
+        join(FRAMEWORK_DIR, "tools", "xtensa-lx106-elf", "include")
```
mcspr: PIO error about the missing include comes from here. Proper location should be:
```
join(platform.get_package_dir("toolchain-xtensa"), "include")
```
Because the coreasm.h is here, not inside of `tools/` subdir like with Arduino IDE:
```
~/.platformio/packages/toolchain-xtensa/include/xtensa/coreasm.h
```

mhightower83:
@mcspr Thankyou! That was my problem.

tools/platformio-build.py
mhightower83: Success! Thanks again!

mhightower83:
For a module that has a 4M XMC flash chip, there appear to be some compatibility issues with this PoC when using 16K ICACHE and the new `experimental::initFlashQuirks()`. A hardware WDT kicks in during `user_init()` just before a call to set up the `cont` stack best guess it is in `experimental::initFlashQuirks()`. The crash is occurring before the sketch code has the opportunity to run; however, changing the sketch code appears to affect whether a crash occurs. I assume it has to do with code size or placement, very strange, and difficult to create an MCVE. This is a very confusing problem.

I will do a push that comments out `experimental::initFlashQuirks()` for now, since some new fixes appear to be in the works. https://github.com/esp8266/Arduino/issues/7267#issuecomment-629657724
```

wdt reset
load 0x4010f000, len 3456, room 16 
tail 0
chksum 0x84
csum 0x84
v1a9d9092
~ld

Hardware WDT reset

>>>stack>>>

ctx: sys
... long sys stack dump decoded and deleted
... cont stack dump also deleted it was all zeros

Decoding stack results
0x4023258b: experimental::_SPICommand(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t*, uint32_t, uint32_t, uint32_t) at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_spi_utils.cpp line 103
0x40232530: experimental::_SPICommand(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t*, uint32_t, uint32_t, uint32_t) at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_spi_utils.cpp line 80
0x40240000: _tzset_r at /home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/time/tzset_r.c line 158
0x40232530: experimental::_SPICommand(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t*, uint32_t, uint32_t, uint32_t) at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_spi_utils.cpp line 80
0x40232524: experimental::_SPICommand(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t*, uint32_t, uint32_t, uint32_t) at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_spi_utils.cpp line 75
0x40232691: experimental::SPI0Command(uint8_t, uint32_t*, uint32_t, uint32_t) at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_spi_utils.cpp line 182
0x4023245a: experimental::initFlashQuirks() at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_flash_quirks.cpp line 65
0x40230494: init() at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring.cpp line 215
0x4022ffcb: user_init() at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_main.cpp line 349
0x4023fe6b: _tzset_r at /home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/time/tzset_r.c line 76
0x40100498: schedule_function(std::function  const&) at /home/mhightow/Arduino/hardware/esp8266com/esp8266/cores/esp8266/Schedule.cpp line 82
0x4023fe86: _tzset_r at /home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/time/tzset_r.c line 82
```

ChocolateFrogsNuts:
My initial thoughts are that if it is in initFlashQuirks() something is happening to bump the PRECACHEd code out of the cache (an interrupt?).
As far as I remember, the critical code is <200 bytes and apart from one call to a small rom function (Wait_SPI_Idle) only does register manipulation, so if something is pushing that out of cache it must be doing something big in an interrupt... or it could be causing access to the SPI controller (eg running some non IRAM_ATTR code, though that should be impossible now).

> For a module that has a 4M XMC flash chip, there appear to be some compatibility issues with this PoC when using 16K ICACHE and the new `experimental::initFlashQuirks()`. A hardware WDT kicks in during `user_init()` just before a call to set up the `cont` stack best guess it is in `experimental::initFlashQuirks()`. The crash is occurring before the sketch code has the opportunity to run; however, changing the sketch code appears to affect whether a crash occurs. I assume it has to do with code size or placement, very strange, and difficult to create an MCVE. This is a very confusing problem.

tools/platformio-build.py
```diff
@@ -258,17 +259,27 @@ def scons_patched_match_splitext(path, suffixes=None):
     env.Append(CPPDEFINES=[current_vtables])
 assert current_vtables
 
+current_mmu_iram_size = None
+for d in flatten_cppdefines:
+    if str(d).startswith("MMU_IRAM_SIZE"):
+        current_mmu_iram_size = d
```
mcspr, line 265:
```suggestion
mmu_iter = iter(flatten_cppdefines)
while True:
    try:
        item = next(mmu_iter)
        if str(item).startswith("MMU_IRAM_SIZE"):
            val = next(mmu_iter)
            current_mmu_iram_size = "{}={}".format(item, val)
    except StopIteration:
        break
```
This is a weird one. flattened list value is actually right after the d match, so we never get the real flag into the build. I just noticed this when building, iram len was 1. Directly iterating with env["CPPDEFINES"] is weird, too, since we either get a 2-item tuple or a string. SCons leaves this as an exercise for the user :)
sry not noticing this earlier.

mcspr: Or, directly in defines:
```suggestion
for flag in env["CPPDEFINES"]:
    try:
        d, val = flag
        if str(d).startswith("MMU_IRAM_SIZE"):
            current_mmu_iram_size = "{}={}".format(d, val)
    except ValueError:
        continue
```

devyte:
@mhightower83 @ChocolateFrogsNuts what is the status with this? Was the xmc incompatibility resolved?

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -314,15 +431,90 @@ void umm_init( void ) {
...
+  for (size_t i = 0; i < UMM_NUM_HEAPS; i++) {
+    heap_context[i].heap = NULL;
+  }
+  memset(&heap_context[0], 0, sizeof(heap_context));
+  umm_init_common( UMM_HEAP_DRAM, (void *)UMM_MALLOC_CFG_HEAP_ADDR, UMM_MALLOC_CFG_HEAP_SIZE, true );
+  // umm_heap = (void *)&heap_context;
+}
+
+#ifdef UMM_HEAP_IRAM
+#include <core_esp8266_non32xfer.h>
```
devyte, line 474: This creates a circular dependency: the core depends on umm, and umm depends on the core. This is never a good thing.
Instead, I suggest one of the following:
1. a weak function that in here is implemented empty, and is implemented as calling install_non32xfer_exception_handler(). I'm not fond of adding additional weak symbols.
2. receive a function pointer as arg, to be called on init. I'm not sure this is possible.
3. use the umm config header to define a binding to an external init function to be called. The external function is implemented in the core.

cores/esp8266/umm_malloc/umm_malloc.cpp
```diff
@@ -547,7 +744,13 @@ void *umm_malloc( size_t size ) {
 
   UMM_CRITICAL_ENTRY(id_malloc);
 
-  ptr = umm_malloc_core( size );
+#if 1 // !defined(USE_ISR_SAFE_EXC_WRAPPER)
```
devyte, line 747: why #if 1? should this be cleaned up?

cores/esp8266/umm_malloc/umm_malloc.h
```diff
@@ -27,8 +35,86 @@ extern void  umm_free( void *ptr );
...
+  }
+
+  ~HeapSelectDram() {
+    umm_set_heap_by_id(_heap_id);
+  }
+
+protected:
+    size_t _heap_id;
+#endif
+};
```
devyte, line 117: These are objects, they likely should be put in a separate file, e.g.: umm_malloc_scoped.h or umm_malloc_select.h or similar.

cores/esp8266/xtruntime-frames.h
```diff
@@ -0,0 +1,162 @@
+/*  xtruntime-frames.h  -  exception stack frames for single-threaded run-time  */
```
devyte, line 1: Is this header used anywhere?

libraries/esp8266/examples/MMU48K/MMU48K.ino
```diff
@@ -0,0 +1,354 @@
...
+  dump_mem32((void *)(0x4010C000 - 16 * 4), 32);
+
+  probe_b = (char *)gobble;
+  probe_s = (short *)((uintptr_t)gobble);
+  unaligned_probe_s = (short *)((uintptr_t)gobble + 1);
+
+}
+
+
+extern uint32_t mmu_non32xfer_count;
```
devyte, line 182: Rather than do this, it's better practice to implement an accessor function that returns the value.

libraries/esp8266/examples/MMU48K/MMU48K.ino
```diff
@@ -0,0 +1,354 @@
...
+extern "C" void _text_end(void);
+
+extern "C" void umm_init_iram(void) {
+  uint32_t sec_heap = (uint32_t)_text_end + 32;
+  sec_heap &= ~7;
+  size_t   sec_heap_sz = 0xC000UL - (sec_heap - 0x40100000UL);
+
+  umm_init_iram_ex((void *)sec_heap, sec_heap_sz, true);
+}
+#endif
```
devyte, line 353: Cleanup?

bootloaders/eboot/eboot.c
```diff
@@ -68,7 +68,7 @@ int load_app_from_flash_raw(const uint32_t flash_addr)
             load = true;
         }
 
-        if (address >= 0x40100000 && address < 0x40108000) {
+        if (address >= 0x40100000 && address < 0x4010C000) {
```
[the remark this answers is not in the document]
mhightower83: Yes. Will do.

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -335,10 +344,16 @@ extern "C" void user_init(void) {
 
     initVariant();
 
-    experimental::initFlashQuirks(); // Chip specific flash init.
+    // experimental::initFlashQuirks(); // Chip specific flash init.
```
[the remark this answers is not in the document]
mhightower83: I am confused. I have tried this from time to time as merges have been done. Until now it has failed. It is now working. I'll remove the commenting for now and keep an eye on it.

The module I have with an XMC chip in it is quite a stable w/o the patch.

cores/esp8266/Esp.h
```diff
@@ -172,6 +172,17 @@ class EspClass {
 #else
         uint32_t getCycleCount();
 #endif
+
+        void enableVM();
+        void setDramHeap();
+        void setIramHeap();
+        void setExternalHeap();
+        void setInternalHeap() {setDramHeap();}; // depricated
```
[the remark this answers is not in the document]
mhightower83: 1) At this time it is not allowed. 
> Because of the extended processing time that it takes to process a malloc using the  non32-bit exception handler. I chose to force all new malloc's that occur from an IRQ over to DRAM. If they really want IRAM allocated, they have to do the allocation at INTLEVEL 0.

2) It is designed to work. 
> When I took the changes @earlephilhower had made to `umm_malloc` for multiple heaps, it morphed quite a bit more. There is now a heap context pointer. It can with little overhead, switch to an alternate heap context. Not in this PR, I have an experimental branch where WString does all reallocs with IRAM and when it runs out, a slightly modified umm_realloc will OOM internally and switch over and malloc from the DRAM heap and copy. On return the heap context is still IRAM; however, the allocation returned is in DRAM.

cores/esp8266/mmu_iram.cpp
```diff
@@ -0,0 +1,214 @@
...
+        if (!is_gpio_persistent()) {
+            /* Restore pin to TX after Power-on and EXT_RST */
+            __pinMode(pin, FUNCTION_0);
+        }
+        in_initPins = false;
+        return;
+    }
+
+    __pinMode( pin, mode );
+}
```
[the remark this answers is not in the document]
mhightower83: This was more for development debugging. In my mind using this feature had a lot of uncertainty about it. Since it is scarcely documented. I have held back from deleting a log of development code used to monitor what is going on. I think there is more of this scattered around that needs to be looked at. I suppose now may be the time to purge.

cores/esp8266/mmu_iram.h
```diff
@@ -0,0 +1,274 @@
...
+  uint32_t v_cfg;
+  int32_t state;      // -1 - not initialized, 0 - disabled, 1 - enabled
+  uint32_t enable_count;
+  uint32_t disable_count;
+  uint32_t map;
+  uint32_t p;
+  uint32_t v;
+}  mmu_cre_status_t;
+
+extern mmu_cre_status_t mmu_status;
```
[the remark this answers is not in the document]
mhightower83: Yes, keep private. This looks like more of the development monitor/debug code. I am not sure how much to keep / what to keep. I think this should be moved under a `#ifdef DEV_DEBUG_MMU_IRAM`

cores/esp8266/mmu_iram.h
```diff
@@ -0,0 +1,274 @@
...
+
+
+#ifdef DEBUG_MMU
+
+static inline bool is_iram(uint32_t addr) {
+// constexpr uint32_t _start = 0x40100000UL;
+#define IRAM_START 0x40100000UL
+#ifndef MMU_IRAM_SIZE
+#define MMU_IRAM_SIZE 0x8000UL
```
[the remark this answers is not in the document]
mhightower83: There was a time when the Cache configuration did not specify MMU_IRAM_SIZE for the balanced, legacy, option and this filled in the blank. The #ifdef is no longer needed. I can remove it.

libraries/esp8266/examples/MMU48K/MMU48K.ino
mhightower83: I'll remove this for now. 

Someday I'll create an example to illustrate how to reserve some of the IRAM from the heap in order to create an always available IRAM buffer that can be used across boots. Examples: a crash stack save buffer, a diagnostic message print buffer, etc.

This block shows that there is some external control over defining the IRAM heap; however, it needs comments and an example.

cores/esp8266/xtruntime-frames.h
mhightower83: It is used by cores/esp8266/exc-c-wrapper-handler.S.

Is there a better way to organize this? Should this and its friends be moved into a subdirectory?

libraries/esp8266/examples/MMU48K/MMU48K.ino
mhightower83: Unless you feel there is some value in this count, I'll remove it?

It was more to convince me my code was executing. Beyond the initial development use, I see no need to expose it.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: The need is to have the  non32xfer_exception_handler installed before creating an IRAM heap. 

There is already a call to install the non32xfer_exception_handler in `core_esp8266_main.cpp`. I can extend the #if around that call to include a test for `UMM_HEAP_IRAM`. 

This would mean no earlier initialization of IRAM heap before `user_init` runs. Which seems highly unlikely anyways. We already have the ability to handle an early init of the DRAM heap. Which is what the SDK needs at startup.

cores/esp8266/umm_malloc/umm_malloc.cpp
mhightower83: Yes. This is where I force the use of DRAM when called with interrupts disabled. IRAM was unsafe to use before replacing the exception C wrapper because it enabled interrupts. 

While it should now be possible to handle allocating IRAM from an ISR, I lean toward not allowing it, because of the increased run time with interrupts blocked. The timing gets even longer if UMM_POISON checking is running. Every byte check becomes an expensive exception.

If you agree with imposing this limitation, I'll remove the #if conditional and keep the test?



In contrast, if a realloc is done with an existing allocation in IRAM and you are calling from an ISR and we have `USE_ISR_SAFE_EXC_WRAPPER` it will process the request. My thinking is to just allow it if they already have it. This logic path takes less IRAM. I don't feel strongly about this.


Additionally, in regards to USE_ISR_SAFE_EXC_WRAPPER define. I am thinking it may be time to clean up/consolidate and always follow the USE_ISR_SAFE_EXC_WRAPPER path when NON32XFER_HANDLER is defined. I don't think we will ever want to not use the NON32XFER_HANDLER w/o  USE_ISR_SAFE_EXC_WRAPPER. The code evolved over time and I initially placed conditionals to make it possible to take a step back. This is one of those. At this time USE_ISR_SAFE_EXC_WRAPPER is required for the umm_malloc library to safely manage an IRAM heap.

If you agree I'll consulate this conditional build branch?

cores/esp8266/umm_malloc/umm_malloc.h
mhightower83: Would `umm_heap_select.h` be acceptable?

cores/esp8266/mmu_iram.h
```diff
@@ -0,0 +1,274 @@
...
+#define ASSERT_RANGE_TEST_WRITE(a) do {} while(false)
+#define ASSERT_RANGE_TEST_READ(a) do {} while(false)
+#endif
+
+/*
+ * Some inlines to allow faster random access to non32bit access of iRAM or
+ * iCACHE data elements. These remove the extra time and stack space that would
+ * have occured by relying on exception processing.
+ */
+static inline uint8_t get_uint8_iram(const void *p8) {
```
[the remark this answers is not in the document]
mhightower83: I'll change them to mmu_..., e.g. mmu_get_uint8().

Looks like the only thing the get_uint8 has to offer is range checking when `DEBUG_ESP_CORE` is present.

cores/esp8266/mmu_iram.h
```diff
@@ -0,0 +1,274 @@
...
+  uint32_t p;
+  uint32_t v;
+}  mmu_cre_status_t;
+
+extern mmu_cre_status_t mmu_status;
+
+
+#ifdef DEBUG_MMU
+
+static inline bool is_iram(uint32_t addr) {
```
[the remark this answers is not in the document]
mhightower83: Okay.

ChocolateFrogsNuts:
> @mhightower83 @ChocolateFrogsNuts what is the status with this? Was the xmc incompatibility resolved?

From memory I "resolved" it by disabling interrupts during the critical section - when the precached code/access to SPI0 happens.
The fact that this works potentially points to something else being an issue - as in something in an interrupt is causing an IRAM cache miss, or some other weird thing - but as the critical code executes for a minimal and finite time, and for now only a few times at startup so I didn't pursue it further.

cores/esp8266/core_esp8266_main.cpp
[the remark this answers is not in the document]
ChocolateFrogsNuts: > Does this still need to be commented out? CC @ChocolateFrogsNuts

Not now that the minor change of disabling interrupts during the critical part of the SPI0 access has been merged.
Mine are quite stable too... until you start using a flash based file system at higher clock speeds.

cores/esp8266/StackThunk.cpp
```diff
@@ -48,7 +50,14 @@ void stack_thunk_add_ref()
 {
   stack_thunk_refcnt++;
   if (stack_thunk_refcnt == 1) {
+    DBG_MMU_PRINTF("\nStackThunk malloc(%u)\n", _stackSize * sizeof(uint32_t));
+    // The stack must be in DRAM, or an Soft WDT will follow. Not sure why,
+    // maybe too much time is consumed with the non32-bit exception handler.
+    // Also, interrupt handling on an IRAM stack would be very slow.
+    // Strings on the stack would be very slow to access as well.
+    HeapSelectDram ephemeral;
```
devyte, line 58: This makes me think about interrupt safety in the ephemeral constructor... are we 100% sure the heap change is interrupt safe, even from itself?

doc/mmu.rst
```diff
@@ -0,0 +1,257 @@
...
+      ...
+
+``free`` will always return memory to the correct heap. There is no need
+for tracking and selecting before freeing.
+
+``realloc`` with a non-NULL pointer will always resize the allocation
+from the original heap it was allocated from. When the supplied pointer
+is NULL, then the current heap selection is used.
+
+Low level primatives for selecting a heap. These are used by the above
```
devyte, line 199: primitives*

libraries/esp8266/examples/MMU48K/MMU48K.ino
devyte: Sure, remove away. I was wondering about its use.

cores/esp8266/core_esp8266_main.cpp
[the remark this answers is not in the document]
devyte: In that case, this needs to be uncommented

cores/esp8266/StackThunk.cpp
mhightower83: I think it is safe as long as an ISR does not get interrupts enabled. 
There was a window with the non32-exception handler; however, I think I have that one closed.
ISRs should not try and change heaps. Even then all should be well if they put it back the way it was. On return, everything should be the same. The foreground thread will continue with the same information that was there before the interrupt.

Changing the heap involves changing a single variable. From the umm_mallac side, that variable is used to load a context pointer by index. From there it stays in registers or the current stack as the request is processed. With a single variable to reference for heap selection, I am thinking it is unlikely that umm_malloc can be called with things in an unusable transition state.

Additionally from the umm_malloc side, it can decide as in the ISR case, to ignore the current context and locally load the context for IRAM. No need to touch the current heap index variable.

Async logic without locks is always tricky. I keep asking the question: "Have I missed an angle?"

One concern I have about using IRAM instead of DRAM would be DMA. Do we have any DMA operation taking place with the ESP8266? I would expect the Network buffer transfers to be DMA. For TCP transfers I believe lwIP is configured to always copy sends to its buffers which would be in DRAM so that should not be a problem. I am less familiar with TCP receive and the UDP configuration options.

cores/esp8266/umm_malloc/umm_malloc.h
mhightower83: Moved class to umm_heap_select.h

doc/mmu.rst
mhightower83: done

libraries/esp8266/examples/MMU48K/MMU48K.ino
mhightower83: Done. Added example IramReserve.ino.

libraries/esp8266/examples/MMU48K/MMU48K.ino
mhightower83: Done.

mhightower83:
@earlephilhower I am having trouble with memmove_P getting its directions mixed up when IRAM is used.
```cpp
#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <umm_malloc/umm_heap_select.h>

//#define memmove_P ets_memmove

char str[] = "Move over";

void setup() {
  WiFi.persistent(false);
  WiFi.mode(WIFI_OFF);
  Serial.begin(115200);
  delay(10);
  Serial.printf_P(PSTR("\r\nBegin ...\r\n"));

  {
    HeapSelectIram ephemeral;
    size_t sz = strlen(str) + 1;
    char *newstr = (char *)malloc(sz + 8);
    memcpy(newstr, str, sz);
    Serial.println(newstr);
    memmove_P(&newstr[1], newstr, sz);
    newstr[0] = ' ';
    Serial.println(newstr);
    free(newstr);
  }
}

void loop() {}
```
Prints:
```
Begin ...
Move over
 MMMMMMMMMM
```
Works when not using IRAM or using ets_memmove.

earlephilhower:
> @earlephilhower I am having trouble with memmove_P getting its directions mixed up when IRAM is used.

The problem is that it assumed there would never be a dest that was in ROM/IRAM space...

https://github.com/earlephilhower/newlib-xtensa/blob/671feddab9539cef4daf72dfb246cae178cefd28/newlib/libc/sys/xtensa/string_pgmspace.c#L184-L190

So it's actually falling into a `memcpy` (because there was no chance of a PGM_P *write*, memmove is always a memcpy since they couldn't overlap).  Given that memcpy is working, I assume your patch has a misaligned write/read handler.

memmove's a little complicated and has a few corner cases, so I need to think about the right solution.  Maybe it's as simple as "if dest is in 4b-only space, default to memmove and pray, else if src is in 4b space then use memcpy_P, else memmove"...

mhightower83:
> The problem is that it assumed there would never be a dest that was in ROM/IRAM space...

Sounds like a reasonable assumption to me. When would anyone ever want to rewrite byte strings in instruction memory? :P

> ... patch has a misaligned write/read handler.

Apologies I forgot to mention that. 

I currently have a sketch running with an experimental version of WString that is configured to use IRAM first and DRAM on OOM. I am waiting to see what jumps out next.

earlephilhower:
Can you open a bug on https://github.com/earlephilhower/newlib-xtensa about the memmove_P?  It's a simple fix I don't want to lose.

earlephilhower:
@mhightower83 I have built a new libc.  Can you give it a try w/your patches?

[libc.zip](https://github.com/esp8266/Arduino/files/5046641/libc.zip)

Unzip over the existing `tools/sdk/libc/xtensa-lx106-elf/lib/libc.a`  and re-run your memmove_P test.

If it works, I can commit to the main newlib repo.  Logic is very simple

mhightower83:
@earlephilhower Thanks! It works with the memmove_P test and in the original failing use case.

cores/esp8266/umm_malloc/umm_malloc.cpp
devyte: I would prefer to not have special cases, if at all possible and if it makes sense, i.e. I'd like to avoid "if you're using THIS heap then you can do THIS, but if you're using THAT heap you can't". Other differences are ok, e.g.: "if you use THIS heap, performance is slower, because blahblah insert explanation here".
So let's consolidate this and avoid yet another build configuration parameter.
Your explanation above about the exception wrapper should be included somewhere visible as a comment. There seem to be a lot of exception handlers and wrappers throughout the core code, and I'm rather lost as to what replaces which wrapper around what call.
E.g.:
"This is a wrapper around the exception handler xyz in file blah.c. It is needed because blahblah DRAM blah IRAM etc."

cores/esp8266/mmu_iram.h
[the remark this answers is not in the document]
devyte: Does this still need to be cleaned up?

devyte (changes requested):
Minor things, then I'm good for an epic merge :)

cores/esp8266/StackThunk.cpp
d-a-v: Network (wifi) uses `malloc()` too. I believe it only happens at `yields()` and (end of) `loop()`.
One/I should conduct some tests under heavy network load and see what happens.
If a network allocation happens in IRAM without the unaligned handler, bad things may happen.

d-a-v (approved):
Default menu option changes nothing so it should be safe to merge.
Changes are numerous but quite clear if time is spent to read them.
In my humble opinion, this feature will give a second life to esp8266/arduino where memory is sometimes tight.
I tried it with esp8266Audio which requires a fair amount of cpu power for decoding, and all went well with the shared second heap allowing a bigger audio buffer and much more free DRAM.
Thank you for this huge work @mhightower83 !

cores/esp8266/StackThunk.cpp
mhightower83: Then, I think the changes that were made here should keep Network malloc()s in DRAM.
https://github.com/esp8266/Arduino/blob/a3bf35cff7abc0bd89090601b112efc218c07f33/cores/esp8266/core_esp8266_main.cpp#L207-L216

Any SYS callbacks that changes the Heap must restore before they return.

mhightower83:
@devyte 
> https://github.com/esp8266/Arduino/pull/7060#discussion_r499788793
```cpp
  uint32_t v;
}  mmu_cre_status_t;

extern mmu_cre_status_t mmu_status;
```
> Does this still need to be cleaned up?

That has been removed.

devyte (approved):

devyte:
Congratulations @mhightower83, this is an awesome feature!

mhightower83:
@devyte Thanks! It looks like that little POC for 16K/32K cache selection grew a little bit.

I want to thank @earlephilhower for his work on a virtual memory PR which I leverage to get started with the exception handler and 1st Heap selection API.

efitrillo:
> @devyte Thanks! It looks like that little POC for 16K/32K cache selection grew a little bit.
> 
> I want to thank @earlephilhower for his work on a virtual memory PR which I leverage to get started with the exception handler and 1st Heap selection API.

Thank you for this amazing option. It really comes in handy specially when using SSL, helping the ESP8266 instead of replacing with ESP32.

As a sidenote for newcomers when you use the tool, the extra HEAP won't show with the usual freeHeap function, but its there working and having extra heap.

mhightower83:
@efitrillo That is good to hear.

The well-established Heap APIs, like freeHeap, work with the current Heap selected.
Do something like this to see the free IRAM Heap:
```cpp
#include <umm_malloc/umm_heap_select.h>

#ifdef UMM_HEAP_IRAM
  {
    // Note, the current heap does not change if the IRAM Heap was not in the 
    // build option. In that case, ESP.getFreeHeap() will report free DRAM space.
    HeapSelectIram ephemeral;  
    Serial.printf("IRAM free: %6d\r\n", ESP.getFreeHeap());
  }
#else 
  Serial.printf("IRAM free: 0\r\n");
#endif 
```

PR #7068, opened 2020-02-09
Use sprintf to output floats in Print/dtostrf

### LINKED ISSUE #7043: [BUG]: Rounding Values when Using String Type Cast
### Basic Infos

#### Platform

- Hardware:			[ESP-12]
- Core Version:			[2.6.3]
- Development Env:		[Arduino IDE|]

### Settings in IDE

- Flash Size: 		[4MB/1MB]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- Upload Using:		[OTA|SERIAL]

### Problem Description

When using a type cast to cast an float to a String like:
`String(pressure, 0)` the result is not rounded correctly.
If I make a type cast like this:
`String(pressure, 1)`the result is rounded correctly.

```
#include <Arduino.h>

void setup() {
  Serial.begin(115200);
  while (!Serial) yield(); 
  int a = 1014;
  float b = 1014.45;
  float c = 1014.5;

  Serial.println();
  Serial.println(a);
  
  Serial.println();
  Serial.println(b, 1);
  Serial.println(String(b,1));

  Serial.println();
  Serial.println(c, 0);
  Serial.println(String(c ,0));
}

void loop() {
}
```

### Debug Messages

```
1014

1014.5
1014.5

1015
1014

```

earlephilhower:
Fixes #7043

Two slightly different custom routines were implemented by hand in
dtostrf (an AVR-lib non-ISO function) and Print.  This resulted in
inconsistent output of float/double vars when rounding was needed.

Replace them all with a call to sprintf(), removing the duplicated, not
quite correct code.

Print(String(float)) and Print(float) now generate the same output.

### DIFF
--- a/cores/esp8266/core_esp8266_noniso.cpp
+++ b/cores/esp8266/core_esp8266_noniso.cpp
@@ -22,6 +22,7 @@
 
  */
 
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <stdbool.h>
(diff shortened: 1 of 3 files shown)


earlephilhower:
The example code in the original bug (with an INF and NAN added for checks) now reports the same results no matter if it's Print(float) or Print(String(float)):
````
#include <Arduino.h>

void setup() {
  Serial.begin(115200);
  while (!Serial) yield(); 
  int a = 1014;
  float b = 1014.45;
  double c = 1014.5;

  Serial.println();
  Serial.println(a);
  
  Serial.println();
  Serial.println(b, 1);
  Serial.println(b, 2);
  Serial.println(String(b,1));

  Serial.println();
  Serial.println(c, 0);
  Serial.println(String(c ,0));

  float n = nanf("x");
  Serial.println(n);

  float u = 1/0.0;
  Serial.println(u);

}

void loop() {
}
````

````
1014

1014.5
1014.45
1014.5

1014
1014
nan
inf
````

File: cores/esp8266/core_esp8266_noniso.cpp [hunk 1]
```diff
@@ -40,75 +41,9 @@ char* ultoa(unsigned long value, char* result, int base) {
...
-            *out++ = '.';
-        }
-        number -= digit;
-        number *= 10.0;
-    }
-
-    // make sure the string is terminated
-    *out = 0;
+    char fmt[32];
+    sprintf(fmt, "%%%d.%df", width, prec);
```
mcspr: Might need to be labelled as breaking? See
https://github.com/arendst/Tasmota/blob/1598e6227a31be5f27081c5a660d22d21613c0ca/platformio.ini#L83
https://github.com/arendst/Tasmota/blob/development/pio/strip-floats.py
https://github.com/xoseperez/espurna/blob/dev/code/scripts/espurna_utils/float_support.py

It removes `-u _printf_float` from linker flags, so this might come as a surprise b/c code uses dtostrf routinely :)

earlephilhower: Ah, crap.  Thank you for reminding me about that!  I had a niggle about it, but it's been maybe 2 years since the whole -u print_float stuff.

I'll open a new issue.  The solution is not clear to me now.

PR #7080, opened 2020-02-13
Made ESP8266WebServer::client() return a reference (fixes #7075)

### LINKED ISSUE #7075: ESP8266WebServer::client() should return a reference
Hi,

Currently `ESP8266WebServer::client()` returns a `WiFiClient` by value, which forbids passing the result directly to a function taking a `Client&`.

For example, it causes a compilation error with [ArduinoJson](https://arduinojson.org):

```c++
serializeJson(doc, server.client());
```

Because [`serializeJson()`](https://arduinojson.org/v6/api/json/serializejson/)  takes a `Print&`, the line above produces the following error:

```
no instance of overloaded function "serializeJson" matches the argument list -- argument types are: (ArduinoJson6141_0000010::StaticJsonDocument<768U>, WiFiServer::ClientType)
```

To fix this problem, `ESP8266WebServer::client()` must return a reference (and not a copy), just as `HTTPClient::getStream()` does.

(see also [this question on StackOverflow](https://arduino.stackexchange.com/q/72433/3430))

Best regards,
Benoit

bblanchon:
*edit from maintainer:* fixes #7075


devyte:
I am in favor of this, in fact I completely agree to move away from return-by-value. However, there is worry that this could cause pain due to moving away from the Arduino way of return-by-value (object slicing or whatever), as has happened in the past.
Whatever the case, this is a breaking change due to a signature change, so at best it must be targeted for v3.

PR #7093, opened 2020-02-19
restore dtostrf when floats are disabled in printf/scanf + round fix

### LINKED ISSUE #7043: [BUG]: Rounding Values when Using String Type Cast

### LINKED ISSUE #7073: dtostrf/Print(float) inconsistent value/methods, downstream use
#7068 removed two slightly different (and both slightly wrong) implementations of `double` to `ASCII string` in `Print::print(float)` and `dtostrf`(a non-std. AVR C addition AFAICT) and replaced it with a call to newlib's `sprintf(%f)` which is known good.

Unfortunately, as @mcspr noted, there are folks who are building the core w/o float-printf support (`printf_float` in platform.txt), so the above PR will end up borking their output.

> Might need to be labelled as breaking? See
https://github.com/arendst/Tasmota/blob/1598e6227a31be5f27081c5a660d22d21613c0ca/platformio.ini#L83
https://github.com/arendst/Tasmota/blob/development/pio/strip-floats.py
https://github.com/xoseperez/espurna/blob/dev/code/scripts/espurna_utils/float_support.py

The main question now is what is the right thing to do in this case.  I suggest it's one of the following:
1) Revert to duplicated code and inconsistent output diffs between Print() and dtostrf() in corner cases.  (i.e. undo #7068).  Larger code for most people (3 different versions of float->ascii in ROM), but downstream keeps working.
2) Keep sprintf(%f) and disallow disabling printf_float, requiring changes in downstream projects (ROM will go up ~1K for them, but down ~1KB for most users since the 2 dup'd versions will be expunged)
3) Pick method from Print(float) and replicate it in dtostrf() so they match.  Then replace Print(float) with a call to dtostrf().  Smaller ROM for all, both match, but may not match the correct output in some cases.
4) Same as 3, but use the dtostrf() as golden and not Print()s version.

FWIW, looking at the AVC-LIBC implementation, it is closest in logic to number 2 above.  They go down to the binary FP level and extract exponents and mantissas (as the AVR ASM level) and work with those parts to actually generate strings.

Any opinions from downstream?

d-a-v:
replaces and closes #7087

(*edit*: tested)

fixes #7073 with solution 6
updates #7068 
fixes #7043 
(no need for `-DNOPRINTFLOAT`)


d-a-v:
I am too interested in another solution because I see only the use of the define to not have the real implementation when printf-float is available. That's because the symbol presence cannot be known at compile time.

@earlephilhower with a patch to newlib, we could condition linking of `_prinf_float` the very same way we do with stack in sys space when WPS is not used. No more `-u printf...` nor any define. A simple call to a function could disable printf/floats at link time.

cores/esp8266/core_esp8266_noniso.cpp
```diff
@@ -40,11 +41,95 @@ char* ultoa(unsigned long value, char* result, int base) {
...
+    // make sure the string is terminated
+    *out = 0;
     return s;
 }
 
+// extern weak declaration (= function pointer) to check whether _printf_float is defined
+extern int _printf_float () __attribute__((__weak__));
+
+char * dtostrf(double number, signed char width, unsigned char prec, char *s) {
+    if (!_printf_float) {
```
devyte: This function is used directly only by users/maintainers of builds that have float support disabled. Supporting sprintf in here doesn't make sense to me, it should be the __dtostrf body directly. That will give consistent results to the users regardless of whether float support is enabled or not.

devyte: Whatever the case, please prefer snprintf over sprintf.

earlephilhower: > Whatever the case, please prefer snprintf over sprintf

Normally, I'd agree, but in this case it is a false sense of added security.  This API does not have a way of passing in sizeof(destbuff), and anything you artificially create in the function itself is bogus.

The "width" param is the integer width, not including the "." and the prec after the decimal point.  The "prec" param is the number of fractional digits.  No destination size can be calculated from these numbers...

d-a-v:
diff from original dtostrf is 
```
index b39edbb2..c742904d 100644
--- a/cores/esp8266/core_esp8266_noniso.cpp
+++ b/cores/esp8266/core_esp8266_noniso.cpp
@@ -27,6 +27,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <math.h>
+#include <limits>
 #include "stdlib_noniso.h"
 
 extern "C" {
@@ -77,11 +78,15 @@ char * dtostrf(double number, signed char width, unsigned char prec, char *s) {
     // Figure out how big our number really is
     double tenpow = 1.0;
     int digitcount = 1;
-    while (number >= 10.0 * tenpow) {
-        tenpow *= 10.0;
+    double nextpow;
+    while (number >= (nextpow = (10.0 * tenpow))) {
+        tenpow = nextpow;
         digitcount++;
     }
 
+    // minimal compensation for possible lack of precision (#7087 addition)
+    number *= 1 + std::numeric_limits<decltype(number)>::epsilon();
+
     number /= tenpow;
     fillme -= digitcount;
```

PR #7096, opened 2020-02-21
TCP: fixes backlog limit handling, and allows ephemeral TCP connections

### LINKED ISSUE #2569: Server closes connection after client shuts down the sending side of its socket
#### Hardware

Hardware:           ESP-12F on NodeMCU clone (Geekcreit Doit)
Core Version:       2.3.0
### Description

A WiFiClient instance (or the underlying TCP/IP lib) obtained from WiFiServer::available() seems to close the TCP connection when it receives a FIN packet. According to the TCP spec, it should go to the CLOSE_WAIT state, which would still allow data to be sent in the opposite direction.

This happens if a client connects to a server running on the ESP, sends some data and shuts down the writing side of the socket before receiving a response. The client socket's reading side is still open in this case, and the server should be able to send a response to the client.

This issue also occurs with ESP8266WebServer, because it uses WiFiServer/Client.
### Settings in IDE

Module: NodeMCU 1.0 (ESP-12E Module)
Flash Size: 4MB (3M SPIFFS)
CPU Frequency: 80Mhz
Upload Using: SERIAL
### Sketch

``` cpp
#include "ESP8266WiFi.h"

#define WIFI_SSID "..."
#define WIFI_PASS "..."

WiFiServer server(1400);

void setup() {
  Serial.begin(115200);
  delay(10);

  WiFi.begin(WIFI_SSID, WIFI_PASS);
  Serial.print("Connecting to WiFi.");
    delay(1000);
    Serial.print(".");
  }
  Serial.println("OK");

  server.begin();
}

void loop() {
  WiFiClient client = server.available();
  if (client) {
    Serial.printf("status before delay: %d\n", client.status());
    delay(1000);
    client.write("Hello, World!\n");
    Serial.printf("status after delay: %d\n", client.status());
    client.stop();
    Serial.printf("status after stop: %d\n", client.status());
  }
}
```
#### Client that shuts down the sending side immediately

Shell command:

```
sleep 0 | nc 192.168.178.42 1400
```

(sending side is shut down immediately, because input pipe of netcat is closed after `sleep 0`)

Client output:

```
<nothing>
```

Serial output:

```
status before delay: 4
status after delay: 0
status after stop: 0
```

(4 is ESTABLISHED, 0 is CLOSED)

Wireshark capture:
screen shot 2016-10-02 at 15 03 01
#### Client that waits for the server to close the connection (works correctly)

Shell command:

```
sleep 2 | nc 192.168.178.42 1400
```

(this gives the server enough time to respond and close the connection)

Client output:

```
Hello, World!
```

Serial output:

```
status before delay: 4
status after delay: 4
status after stop: 0
```

(4 is ESTABLISHED, 0 is CLOSED)

Wireshark capture:
screen shot 2016-10-02 at 15 02 01

### LINKED ISSUE #7103: PR #6887 generates execptions in Project Tasmota
#### Platform
-  device                                 |ESP01 ]
- Operating System:		[Windows|Linux]

### Settings in IDE

- Module:  			[|Nodemcu|other]
- Flash Mode:			[DOUT]
- Flash Size: 		[1MB]
- Reset Method:		[|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]

### Problem Description

Tasmota does generate exceptions when using latest git.

Reverting https://github.com/esp8266/Arduino/pull/6887 does solve the issue.
Without this and using all other PR until to the latest Tasmota does not generate exceptions anymore.

d-a-v:
When a sketch does not check `server::available()` often enough, short connections are *now* received and stored in background (backlog limit number of connections at most) and can be retrieved and used by subsequent `client = server::available()`. Those clients will be `status() == CLOSED`, `connected() == false` but `available() > 0`.

@kirillandy Could you check this ? I get what's expected by #2569 with these changes.

@jandrassy This is still not what you are waiting for but it would be nice if you checked nothing has changed on the arduino side (`WiFiClient::operator bool()` has changed).

fixes #2569 

*edit*:

fixes #7103


kirillandy:
Hi d-a-v! Currently trying to check your changes out, but I'm not too certain about my actions.
So I installed the current ESP Arduino library version using Git (as the docs instruct). After a bit of research I then used "git merge https://github.com/d-a-v/Arduino.git FixBacklogHonorEphemeral" to apply your changes on what I have locally stored on my PC.

Is that all I need to do before I compile my code to test out on my ESPs?
(Also, if my actions were correct, was there an easier way to do what I did? :D)

kirillandy:
UPDATE:
So, if everything I did to get hold of your commit was correct (still need some friendly guidance on that, since this is my first time properly using Git and its commands :)), then everything works like a charm! I repeated my previous experiment by sending a few bytes when the server was not available-looping and the server was able to get the client and read the bytes 👍 
I also tried sending bytes in 3 different connections one after the other to check if they'd all be in the queue and the server managed to .available() all 3 and read every message.
Thanks a lot for your work, @d-a-v!

Jason2866:
Fix #7103

PR #7101, opened 2020-02-22
Revert "Fix optimistic_yield to not yield on each call after x µs"

Jason2866:
Reverts esp8266/Arduino#6804


Jason2866:
Errorly done

PR #7105, opened 2020-02-23
Add flash helper constructor to Uri

### LINKED ISSUE #7102: Use of Uri in webserver breaks current implementations
### Basic Infos

#### Platform

- Breaking commit:			https://github.com/esp8266/Arduino/commit/4eca62cb530297d98c00f463dba03113811a444e

### Problem Description

When calling the `WebServer.on(...)` function with a flash string, the compiler refuses to build.

Example error:
```
C:/GitHub/TD-er/ESPEasy/src/WebServer.ino:603:6: error: no matching function for call to 'esp8266webserver::ESP8266WebServerTemplate<WiFiServer>::on(const __FlashStringHelper*, HTTPMethod, WebServerInit()::__lambda4)'
```

Simple work-around is to replace all calls like this with a `String()` wrapper around the flash string.

I guess the best fix would be to add a constructor for Uri with `const __FlashStringHelper*` as parameter.

N.B. some of the derived classes use explicit in the constructor definition, but Uri not. Is this intentional?

devyte:
Fixes #7102

### DIFF
--- a/libraries/ESP8266WebServer/examples/PathArgServer/PathArgServer.ino
+++ b/libraries/ESP8266WebServer/examples/PathArgServer/PathArgServer.ino
@@ -37,7 +37,7 @@ void setup(void) {
     Serial.println("MDNS responder started");
   }
 
-  server.on("/", []() {
+  server.on(F("/"), []() {
     server.send(200, "text/plain", "hello from esp8266!");
   });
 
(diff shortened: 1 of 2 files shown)


devyte:
@earlephilhower I could do a different pr for that after this one is merged. I'm Chuck Norrising this from my phone via webedit while on the road.

PR #7109, opened 2020-02-24
simple TZ api: bypass sprintf/sscanf: + 7KB

### LINKED ISSUE #7107: Code size increase of 6k in esp-quick-toolchain
### Basic Infos

#### Platform

- Hardware:					[ESP-12|other]
- Core Version:			[latest git hash vs. commit 6be561617]
- Operating System:		[Windows|Ubuntu]

### Settings in IDE

- Module:  			[Nodemcu]
- Flash Mode:			[dout]
- Flash Size: 		[1MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA|SERIAL]

### Problem Description

Not a bug. But since commit #6be561617f645f6a2ae82b8211f6af8c43e834cf and latest stage
there is a code size increase about 6k in the esp-quick-toolchain.
Code size with commit #6be.. 
image

and with latest stage

image

Is there a big change in functions that explains that code size increase?
Thx!

d-a-v:
fixes #7107 
@Jason2866 do you confirm tasmota fits in its previous size ?
About that, are you using compressed OTA ?

### DIFF
--- a/cores/esp8266/time.cpp
+++ b/cores/esp8266/time.cpp
@@ -137,9 +137,8 @@ void configTime(int timezone_sec, int daylightOffset_sec, const char* server1, c
     // So DST is always integrated in TZ
     // The other API should be preferred
 
-#if 0
-
-    // using posix API (calls sprintf here, sscanf internally)
+    /*** portable version using posix API
+         (calls sprintf here, then sscanf internally)
 
     int tzs = daylightOffset_sec + timezone_sec;
     int tzh = tzs / 3600;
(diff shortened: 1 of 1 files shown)


cores/esp8266/time.cpp
```diff
@@ -133,21 +133,58 @@ int _gettimeofday_r(struct _reent* unused, struct timeval *tp, void *tzp)
 
 void configTime(int timezone_sec, int daylightOffset_sec, const char* server1, const char* server2, const char* server3)
 {
-    char tzstr [64];
-
     // There is no way to tell when DST starts or stop with this API
     // So DST is always integrated in TZ
     // The other API should be preferred
 
+#if 0
```
devyte: Instead of an #if #else I suggest a comment block along the lines of: "the following is a hack of the internal structures of libc because... (insert explanation)... It is equivalent to the following code: (insert code currently inside the #if 0).

Jason2866:
@d-a-v  Thx!, The PR eliminates the firmware size increase. :-)
Yes, we do use compressed OTA. Works like a charm!

d-a-v:
I was asking about OTA because I have been thinking that increased size on one side would be compensated by size decrease on the other (thanks to compressed OTA). But as @earlephilhower noted, there are still 512K flash chips around, these ones could use any flash size optimization.

PR #7110, opened 2020-02-25
Set a value for SSDP notify interval

PA4WD:
Change the SSDP notify message interval.
Defaults to the value of SSDP_INTERVAL.


libraries/ESP8266SSDP/ESP8266SSDP.h
```diff
@@ -104,6 +105,7 @@ class SSDPClass{
     SSDPTimer* _timer;
     uint16_t _port;
     uint8_t _ttl;
+    uint8_t _interval;
```
devyte: This can't be uint8_t. SSDP_INTERVAL is 1200 which is already above the max of 255 for a uint8_t.
Please change to uint32_t.

libraries/ESP8266SSDP/ESP8266SSDP.h
```diff
@@ -92,6 +92,7 @@ class SSDPClass{
     void setManufacturerURL(const char *url);
     void setHTTPPort(uint16_t port);
     void setTTL(uint8_t ttl);
+    void setInterval(uint8_t interval);
```
devyte: uint32_t

PR #7121, opened 2020-02-28
unblobbing commit

d-a-v:
fixes https://github.com/esp8266/Arduino/commit/2c435b1f427dfdc1567bd52ca6d966d5640d6bb9#commitcomment-37529995

Sorry for that @progtronix

Issue #7124, opened 2020-02-28
OTA broken after commit #7036


MarkyAD:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:					nodemcu
- Core Version:			e6decacd0a03a61f00068264b024e7
- Development Env:		Makefile
- Operating System:		Linux

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			qio
- Flash Size: 		        4M
- lwip Variant:                 v2 Lower Memoru
- Reset Method:		nodemcu
- Flash Frequency:	       40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		OTA
- Upload Speed:              115200

### Problem Description

@d-a-v @devyte After commit #7036 OTA no longer works, it just hangs. I traced it back to this specific commit.

d-a-v:
I tried the OTALeds (ArduinoOTA lib using UDP to negociate) example sketch and it's still working.
Can you give more details on your setup ?

MarkyAD:
I have several nodemcu and d1_mini_pro and they all had the same symptom, all refused authentication. I do all my own dns and dhcp services. Once I reverted my git repo back before that commit, OTA worked again on all of them. Host environment is x86_64 and arm (both 32 and 64).

MarkyAD:
Just tried again on a pi3, same thing but this time espota.py spills "No Answer". I was beginning to think it was a linux python3 issue before I tried reverting my repo. This time on the pi with a board currently flashed with a pre #7036 sketch, ota works the first time, then ota hangs after being flashed with a post #7036 sketch.

Build environment is plain Makefiles and/or Makefiles with arduino-builder (1.8.12).

d-a-v:
Ah I didn't try with auth.

PR #7129, opened 2020-03-03
The properly built lwip2 binaries for #7121

### LINKED ISSUE #7127: Regression in master merged with PR #7022: WDT resets
Details in #7022 and #7121.
PR #7022 works without lock-ups or reboots up to and including master commit 2c435b1f427dfdc1567bd52ca6d966d5640d6bb9. So commit c61b70de865531f62218541fa0f32715c153a449 introduces some change that breaks the timing and causes continual WDT resets (1 per minute or so) in the MVCE I've supplied in #7121.
Without PR #7022, I have not been able to reproduce this malfunction, but that shouldn't make it any less significant.

dok-net:
From clean git checkout, "make install".
Fixes #7127


d-a-v:
`tcp_priv.h` is correct but willingly not merged like you propose to avoid gcc warns. To solve it cleanly, a lwIP patch would be needed but has not been done yet.

I don't understand the rest of this PR, please elaborate what is wrong with current master.

dok-net:
tcp_priv.h reverted, thanks for clarifying.

dok-net:
These binaries are the result of `make install`, as per your and the README details, for lwIP2 in ESP8266 git master fresh clone. It can only guess that the binaries you merges the other day are broken.

As these are binaries, produced from a make process that's mostly opaque, I cannot comment further, except that in my tests, that fail on ESP8266 git master, this PR fixes the WDT resets completely.

d-a-v:
OK. Because the make process is not opaque at all from this side of internet, 
I will produce these binaries myself (or if another maintainer propose to do it) as I always did, from a computer and a clean repository, different than the one used to produce the binaries that you find offending so you can test them.

dok-net:
Great, could you elaborate for the sake of quick research how to match what is landed in ESP8266 Arduino to the sources used, i.e. GNU lwIP? The Makefiles do traverse up and down and do a cascade of includes, on both sides of the internet, I assume :-)
As of writing this, I haven't had a WDT reboot in hours, now, using the "contributor" binaries instead of the "collaborator" ones.

d-a-v:
@dok-net will this file work for you ?
https://github.com/d-a-v/Arduino/raw/90347e719d6c5082ad88d20e617cb79f3aabef1d/tools/sdk/lib/liblwip2-536-feat.a

dok-net:
This clean-built binary doesn't WDT reset every so often, please commit them all, or mine to your liking.
With regard to the source revision, you wouldn't give me the answer, so it appears to be as follows:
lwip2-src is cloned from `https://git.savannah.nongnu.org/git/lwip.git`, checkout is done for the tag given in `lwip2/builder/makefiles/Makefile.defs`, currently `UPSTREAM_VERSION = STABLE-2_1_2_RELEASE`.
Also, the contents of `tools/sdk/lwip2/include/lwip-git-hash.h`,
```
// generated by makefiles/make-lwip2-hash
#ifndef LWIP_HASH_H
#define LWIP_HASH_H
#define LWIP_HASH_STR "STABLE-2_1_2_RELEASE/glue:1.2-30-g92add50"
#endif // LWIP_HASH_H
```
detail the revisions of the lwip repository, obviously, and the `builder` submodule, a.k.a. `lwip2`, a.k.a. `https://github.com/d-a-v/esp82xx-nonos-linklayer.git`, a.k.a. `glue` , in `git describe --tag` syntax.

dok-net:
@d-a-v closed this in anticipation of your fix commit for the lwip binaries.
PR #7128 is (loosely) related to this, in any case, the nullptr checks I've added fix the broken OTA for me (issue #7124) and should be best practice anyway.

PR #7132, opened 2020-03-03
udp: fix again pbuf management

### LINKED ISSUE #7124: OTA broken after commit #7036

d-a-v:
This follows #7036 where udp pbuf chain handling is made hairy because of interleaved information fragment introduced with IPv6.

@szekelyisz Could you please test this change ?
fixes #7124

### DIFF
--- a/libraries/ESP8266WiFi/src/include/UdpContext.h
+++ b/libraries/ESP8266WiFi/src/include/UdpContext.h
@@ -202,6 +202,30 @@ public:
         _on_rx = handler;
     }
 
+#ifdef DEBUG_ESP_CORE
+    // this helper is ready to be used when debugging UDP
+    void printChain (const pbuf* pb, const char* msg) const
+    {
+        // printf the pb pbuf chain, bufferred and all at once
+        char buf[128];
+        int l = snprintf(buf, sizeof(buf), "UDP: %s: ", msg);
+        while (pb)
+        {
+            l += snprintf(&buf[l], sizeof(buf) -l, "%p(H=%d,%d<=%d)-",
+                pb, pb->flags == PBUF_HELPER_FLAG, pb->len, pb->tot_len);
+            pb = pb->next;
+        }
+        l += snprintf(&buf[l], sizeof(buf) - l, "(end)");
+        DEBUGV("%s\n", buf);
+    }
+#else
+    void printChain (const pbuf* pb, const char* msg) const
+    {
+        (void)pb;
+        (void)msg;
+    }
+#endif
+
     size_t getSize() const
     {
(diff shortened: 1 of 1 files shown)


MarkyAD:
Confirming this does indeed fix #7124

szekelyisz:
Reassembly on IPv4 still works, no problem. Something's up though.

I'm sending 4 UDP packets with 1024 bytes payload each in quick succession. The first one is delivered immediately, but the remaining 3 are buffered by LwIP. How those are delivered depends on the commit.

* Vanilla 2.6.2: all packets are correctly delivered to the application.
* #7036: packets in the buffer are dropped.
* This PR: three packets are delivered, the first one contains all three of them interleaved with some garbage data, the second one contains the last two, again interleaved with some garbage, and the last one contains the last packet. I'm suspecting the garbage is the address info pbuf.

I can take a look tomorrow. Not tested IPv6 yet.

d-a-v:
> This PR: three packets are delivered, the first one contains all three of them interleaved with some garbage data, the second one contains the last two, again interleaved with some garbage, and the last one contains the last packet. I'm suspecting the garbage is the address info pbuf.

I think (and while I'm writing it becomes clear) it is because of 
```
    size_t getSize() const
    {
        if (!_rx_buf)
            return 0;

        return _rx_buf->tot_len - _rx_buf_offset;
    }
```

The info pbuf (the garbage)  should not be accessible but it is because of `getSize()` (which should only report size before next info pbuf) and your fix (use of pbuf_contiguous()).
So an update of this PR is needed in order to cache this value for the current packet.

> I can take a look tomorrow. Not tested IPv6 yet.

This is not needed. The code needed for IPv6 is also used for IPv4 and is the info pbuf management.

d-a-v:
@szekelyisz Can you please retry with latest commit ?

szekelyisz:
> @szekelyisz Can you please retry with latest commit ?

Reassembly still works and smaller packets are also delivered correctly.

d-a-v:
Thanks for testing !

PR #7133, opened 2020-03-04
EspSoftwareSerial release 6.7.2 fixes warnings with gcc 9.2

dok-net:


dok-net:
@earlephilhower heads-up

PR #7134, opened 2020-03-04
Expose webserver's chunk api

d-a-v:
This API is already existing but not quite exposed nor used in examples.
It allows to dynamically build and send large webpages without needing to store everything in a single memory buffer (namely String with all bad implications).

FSBrowser example has been slightly modified. 
As a consequence it does not support HTTP/1.0 anymore.


d-a-v:
@mhightower83 This is going to be merged so #7182 can use it,
Last chance to comment about your :eyes: - CI is working right now

mhightower83:
@d-a-v This is great! The other day, I was looking for an API call to do this and didn't find it.

d-a-v:
I thank the original designer too :)

PR #7146, opened 2020-03-10
Delay(0) is an obfuscation of yield() or esp_schedule(); esp_yield()

dok-net:
Let's unroll the code:
yield() =>
```
if (can_yield()) {
  ets_post(LOOP_TASK_PRIORITY, 0, 0); // esp_schedule();
  esp_yield_within_cont();
}
else {
  panic();
}          
```

delay(0): =>
```
ets_post(LOOP_TASK_PRIORITY, 0, 0); // esp_schedule();
if (can_yield()) {
  esp_yield_within_cont();
}
```
So, in detail, there is a difference. Yield() panics in ISRs (and SYS?), delay(0) posts CONT from ISR and SYS just fine, but doesn't actually yield unless in CONT.
I don't think using delay(0), requiring an extra literal 0 argument passed, is more expressive in SYS or ISRs to express a restart for suspend-yielded CONT.

For compatibility expectations, in AVR Arduino core, delay(0) is basically a no-op, it does not yield.

On the ESP32:
```
void delay(uint32_t ms) {
  vTaskDelay(ms / portTICK_PERIOD_MS);
}
```
This obviously is illegal to call from ISRs.

The take-away, as far as I am concerned, is that for the legal uses, yield() succinctly expresses what delay(0) does, and is slightly more efficient and safer.

In ClientContext.h, the `delay(0)` is probably unintentional, the `esp_schedule()` direct call is used in a few places already, and is functionally identical in this place, while also more expressive and saves resources.

### DIFF
--- a/cores/esp8266/uart.cpp
+++ b/cores/esp8266/uart.cpp
@@ -528,7 +528,7 @@ uart_wait_tx_empty(uart_t* uart)
         return;
 
     while(uart_tx_fifo_available(uart->uart_nr) > 0)
-        delay(0);
+        yield();
 
 }
 
@@ -892,7 +892,7 @@ inline void
 uart_write_char_delay(const int uart_nr, char c)
 {
     while(uart_tx_fifo_full(uart_nr))
-        delay(0);
+        yield();
 
     USF(uart_nr) = c;
 
--- a/libraries/ESP8266WiFi/src/include/ClientContext.h
+++ b/libraries/ESP8266WiFi/src/include/ClientContext.h
@@ -342,7 +342,7 @@ public:
                 last_sent = millis();
             }
 
-            delay(0); // from sys or os context
+            yield(); // from sys or os context
 
             if ((state() != ESTABLISHED) || (sndbuf == TCP_SND_BUF)) {
                 break;
--- a/cores/esp8266/PolledTimeout.h
+++ b/cores/esp8266/PolledTimeout.h
@@ -44,7 +44,7 @@ struct DoNothing
 
 struct YieldOrSkip
 {
-  static void execute() {delay(0);}
+  static void execute() {yield();}
 };
 
 template <unsigned long delayMs>
--- a/cores/esp8266/core_esp8266_main.cpp
+++ b/cores/esp8266/core_esp8266_main.cpp
@@ -113,12 +113,9 @@ extern "C" IRAM_ATTR void esp_schedule() {
 }
 
 extern "C" void __yield() {
+    esp_schedule();
     if (can_yield()) {
-        esp_schedule();
         esp_yield_within_cont();
-    }
-    else {
-        panic();
     }
 }
 
--- a/cores/esp8266/core_esp8266_waveform.cpp
+++ b/cores/esp8266/core_esp8266_waveform.cpp
@@ -139,7 +139,7 @@ int startWaveform(uint8_t pin, uint32_t timeHighUS, uint32_t timeLowUS, uint32_t
       }
     }
     while (waveformToEnable) {
-      delay(0); // Wait for waveform to update
+      yield(); // Wait for waveform to update
     }
   }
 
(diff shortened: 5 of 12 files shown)


d-a-v:
`panic()` is called when `yield()` is called from SYS,
but not when `delay(0)` is called from SYS.

This is history, I don't know the exact reason for this. It may be improved.

`can_yield()`:
```
bool ICACHE_RAM_ATTR cont_can_yield(cont_t* cont) {
    return !ETS_INTR_WITHINISR() &&
           cont->pc_ret != 0 && cont->pc_yield == 0;
}
```

dok-net:
@d-a-v Yes, I implied that. Are you thus approving of what I call cleanup?

libraries/ESP8266WiFi/src/include/ClientContext.h
```diff
@@ -342,7 +342,7 @@ class ClientContext
                 last_sent = millis();
             }
 
-            delay(0); // from sys or os context
+            esp_schedule(); // from sys or os context
```
d-a-v: I would also add `esp_yield()` because 
- that's what `delay(0)` does
- it does not panic when in ISR
- this last change was following an issue that was annoying to track

dok-net: Absolutely, thanks - done.

cores/esp8266/uart.cpp
```diff
@@ -892,7 +892,7 @@ inline void
 uart_write_char_delay(const int uart_nr, char c)
 {
     while(uart_tx_fifo_full(uart_nr))
-        delay(0);
+        yield();
```
mcspr: Looking at `uart_write_char_delay`, it is used by the `uart{0,1}_write_char` and those specific write funcs are installed as `ets_install_putc1(...)` when debug mode is set. Does this mean that SYS calls this part, not CONT?

d-a-v: This is right, `esp_schedule()` should be also called here.
So `uart{n}_write_char()`  should be in IRAM right ?
And `uart_write_char_delay()` is inline, maybe the attribute `__attribute__((always_inline))` should be added ?

dok-net:
@devyte One step at a time, a change in core and the included libs doesn't mean 3rd parties have to change alongside, if what's there before still works afterward.
That said, and having reviewed everything more deeply - here I find PRs are just better suited to code reviewing and discussions based on the code than mere issues, please be forgiving if you feel otherwise - I come to this conclusion:

The panic() in yield() doesn't strike me as in anyone's particular best interest. But without the panic() branch, yield() can - and I suggest should - be restated as the rolled out version of `delay(0)`, doing in each context what's the correct thing to do:

```c++
extern "C" void __yield() {
    esp_schedule();
    if (can_yield()) {
        esp_yield_within_cont();
    }
}
```
This has the benefit of allowing us to return to `yield();` from `esp_schedule(); esp_yield();`.
I take it that you've voted against this PR in the previous form, so I'm modifying it to this formulation.
It also is in the spirit of the expectation, that something that looks like regular Arduino `yield()` will resume, whereas for indefinite suspension, `esp_yield()` exists on ESP8266.
@d-a-v please take notice.

dok-net:
Concluding this:
It has become obvious that in current master, the `esp_yield()` performs something that would be more aptly named `esp_suspend()`. Commonly, `yield()` is expected to return from the scheduler on the next round, but for `esp_yield()`, it only does so after being `esp_schedule()`d before or elsewhere.
This can and will be tackled in another timeline.

cores/esp8266/uart.cpp
mcspr: re IRAM and inline. It seems to be inlined anyway, no symbol is created for any functions down the chain besides delay(). Is it a concern for sdk code in any way?
Other q i think is answered below.

d-a-v: @mcspr 
`inline` is an indication (like `register` which is deprecated in latest gcc version I believe) and compiler is free to not honor this request, which is fair.
But in that case, the function needs to be in iram because it can be called from an ISR.
Furthermore we can't declare a function to be at the same time inline and in iram.
So the proposal above is to force the compiler to always use this function inline.

dok-net:
Resuscitated in #7148 with a unique wrapper for `esp_schedule(); esp_yield()` => `esp_resume()`.
Also added the ICACHE_RAM_ATTR and inlining fixed.

PR #7148, opened 2020-03-11
Delay(0) is an obfuscation of yield() or esp_schedule(); esp_yield()

dok-net:
@d-a-v liked #7146, I've reworked it and here it is back, now.


dok-net:
CI bombed again, this time for MacOS :-(

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -112,6 +112,11 @@ extern "C" IRAM_ATTR void esp_schedule() {
     ets_post(LOOP_TASK_PRIORITY, 0, 0);
 }
 
+extern "C" IRAM_ATTR void esp_resume() {
+    esp_schedule();
+    esp_yield();
+}
+
```
d-a-v: How about commenting the choice of this name (is it interrupting tasks calling (`delay()`) from cont stack) and also that it is equivalent to `delay(0)` ?

cores/esp8266/uart.cpp
```diff
@@ -48,6 +48,8 @@
 #include "user_interface.h"
 #include "uart_register.h"
 
+extern "C" void esp_resume();
```
d-a-v: How about using `coredecls.h` to declare functions from only one place ?

cores/esp8266/core_esp8266_main.cpp
d-a-v: If new function names are to be added, with restrictions on usage, like "use esp_break() if code is called from SYS", I suggest adding `_from_sys` in its name to make it clear.

d-a-v: `yield()` historically panics on purpose when it is called from SYS.
`esp_break()` is exactly `yield()` without `panic()`, callable from both cont and sys.
Considering that  the `yield()`'s `panic()` is avoided by `delay(0)` or its new flavour `esp_break()`, then we may just update `yield()` to not panic no ?

dok-net: I don't know what happened, but you must not have available the current sources somehow.
The actual comment now states, I think I adapted it after an earlier discussion:
```
// Replacement for delay(0). In CONT, same as yield(). Whereas yield() panics
// in SYS, esp_break() is safe to call and only schedules CONT. Use yield()
// whereever only called from CONT, use esp_break() if code is called from SYS
// or both CONT and SYS.
```
Meaning, `esp_break()` is intended for both SYS and CONT.
With regard to `yield()` panicking in SYS, whereas `delay(0)` does not and `esp_break()` of course does neither, I think this was discussed and it was stated that `yield()` shall intentionally continue panicking when called from anywhere else but CONT.

d-a-v: I forgot to disable my previous answer before the last one. Sorry that this confused you.

So you propose `esp_break()` for both sys and cont because `yield()` is for cont only, while they both do the same job, like `delay(0)` which currently is used to replace `yield()` everywhere where it is needed for sys and cont.
Then, you replace `delay(0)` by the new sys+cont `yield()` taste.
So we have now `yield()`, `delay(0)`, `esp_break()` ?
Why not simply `yield()` with a comment ?

If the earlier discussion you refer to is [that one](https://github.com/esp8266/Arduino/pull/7146#pullrequestreview-371951203), it also says:
> We could fix it for a major release

@devyte maybe you can elaborate on this:
> The expressibility issue has been discussed before (a long time ago shortly after I arrived). It was deemed valid

dok-net: @d-a-v Thanks, one can never look too many times at the code. I'm going to explain next that it's a bit different than you think, but that wasn't helped by me replacing `delay(0)` by `yield()` on time too often - instead of  by `esp_break()` (https://github.com/esp8266/Arduino/pull/7148/commits/5c39094e4831e6e937df47eae2ebe991b34c2130).

> Then, you replace delay(0) by the new sys+cont yield() taste.
So we have now yield(), delay(0), esp_break() ?
Why not simply yield() with a comment ?

`Delay(0)` is either pointless, in those places where the code runs only ever from CONT, so for final clarity and to let the runtime `panic` to express that contract, I'm replacing it by `yield()`. Again, only in those places.
Wherever code may run from SYS (and perhaps from CONT, too), the new `esp_break()` must be used to replace `delay(0)`. This is to make the intention clear not to `panic` in SYS, which I find obfuscated by a zero time `delay(0)` call.
As you can see, `yield()` (intentional panic in SYS) is not the same as `esp_break()`, and my reservations about `delay(0)` I've explained above.

d-a-v:
(to be noted: @d-a-v *did not approve* this very PR 1.5 year ago and had also created a similar PR that was simpler but literally killed and wiped out after a childish discussion about who copied who)

@dok-net #8317 stands outside from core changes, is necessary for ethernet and is not directed against your proposal.

dok-net:
@mcspr commented:
> blocking() callback order is different, we check timeout then attempt the call. but, isn't checking blocking() first more important than timeout?

In most cases, it is assumed that without suspending the running task, the `blocking` predicate cannot be updated - like SYS needs time to run first. I wonder if this is isn't universally true on the ESP8266 Arduino core running of NonOS SDK. 

> why std::function, and not template in the header? main use-case is to immediately discard the lambda / std-func obj, might as well save on the implicit type punning and std::function dtor call? plus, not need to introduce extra aliases

Valid point, I didn't see how much validation has gone into using an rvalue-ref in the function template variant. I will inspect how adopting that pattern here could turn out. Thanks for reminder.

> why opt out of local timer in the scheduled funcs and share yield() timer with everything else?

Isn't `esp8266::polledTimeout::oneShotFastMs` a busy wait, as such preventing any power saving? I believe using the timer is the correct way to fix #7969, and not inferior for the use-cases of #8317. Mind you, I merged #8317 into this PR and found no cases that weren't already covered here, instead, there were fewer. Strangely enough, I can't find any use of `yieldUntil` in the libraries/lwIP_Ethernet code itself, which we are told it is presented for.

dok-net:
@mcspr Thank you for your contribution. I may not have taken the time analyze further otherwise. Yes, your change shaves 160 bytes off the IROM usage - that's for the pristine WiFiScan.ino example sketch. I'm committing your change, with a small refactoring of the const references on uint32_t that I think were unintended.

mcspr (changes requested):
lgtm
besides the comments and the commit & title could be improved for squashing?

some historical reference
- #6212, here is a more complete replacement of delay(0) usage across the Core
- #8317 here, implement delay on a lower level
- #7969 since we deal with a bunch of weak funcs, here internal libraries reference the exact functionality they want
- #7146 discussion originated the UART changes, as noticed with the delay(0) vs. yield() usage

cores/esp8266/core_esp8266_waveform_pwm.cpp
```diff
@@ -392,7 +393,7 @@ int startWaveformClockCycles_weak(uint8_t pin, uint32_t timeHighCycles, uint32_t
     initTimer();
     forceTimerInterrupt();
     while (wvfState.waveformToEnable) {
-      delay(0); // Wait for waveform to update
+      yield(); // Wait for waveform to update
```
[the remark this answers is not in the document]
dok-net: Again, MT pattern.

cores/esp8266/coredecls.h
```diff
@@ -32,9 +38,44 @@ uint32_t crc32 (const void* data, size_t length, uint32_t crc = 0xffffffff);
...
+    do {
+        esp_suspend();
+    } while (blocked());
+}
+
+// Try to delay for timeout_ms relative to start_ms. Returns false if the delayed task
+// is asynchronously resumed before the timeout is reached.
+// Also returns false if intvl_ms have expired during the active call.
+// Otherwise return true to indicate the timeout_ms have completely expired.
+bool try_esp_delay(const uint32_t start_ms, const uint32_t timeout_ms, const uint32_t intvl_ms);
```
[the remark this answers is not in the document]
dok-net: Not sure about leading `__`, I'm come to expect weak bindings when I see that in the core, but you are right about the `esp_` prefix, I'm changing that - nice catch.

cores/esp8266/Schedule.cpp
```diff
@@ -161,13 +159,10 @@ void run_scheduled_functions()
...
-        {
-            // because scheduled functions might last too long for watchdog etc,
-            // this is yield() in cont stack:
-            esp_schedule();
-            cont_yield(g_pcont);
-        }
+        // scheduled functions might last too long for watchdog etc.
+        // yield() is allowed in scheduled functions, therefore
+        // recursion into run_scheduled_recurrent_functions() is permitted
+        optimistic_yield(100000);
```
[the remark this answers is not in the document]
dok-net: Is this an approval? I'm not sure what you are hinting at, sorry about that.

tests/host/common/Arduino.cpp
```diff
@@ -53,10 +54,33 @@ extern "C" void optimistic_yield (uint32_t interval_us)
...
+bool esp_try_delay(const uint32_t start_ms, const uint32_t timeout_ms, const uint32_t intvl_ms) {
+    decltype(millis()) expired;
+
+    if ((expired = millis() - start_ms) >= timeout_ms) {
+        return true;
+    }
+    const auto remaining = timeout_ms - expired;
+    esp_delay(remaining <= intvl_ms ? remaining : intvl_ms);
+    return false;
+}
```
mcspr, line 83: revisiting the tests, nothing actually calls it :/
`return true;` might suffice for now (...and if it's not working, host test Checks will show)

cores/esp8266/Schedule.cpp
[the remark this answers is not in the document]
dok-net: Beats me how PolledTimeout looping over schedule/suspend is more efficient than optimistic_yield. Remember the purpose of that construction in the particular place is not to spend some time, but to yield once so often. I consider optimistic_yield the right use and more efficient in that place.

tests/host/common/Arduino.cpp
dok-net: I agree with your remarks regarding esp_yield() in startWaveform; done.
I don't see how ignoring the specification in the comment is the right conclusion from any absence of a test :-)
Careful examination of the code in review hopefully is sufficient for this change.

mcspr: > I don't see how ignoring the specification in the comment is the right conclusion from any absence of a test :-)

Now there are two implementations though, one in core\_esp8266\_main.cpp and another esp\_try\_delay is here. The way tests work right now, there's no async waiting like here since the rest of the mock code makes blocking operations instead. So that is why my first though is to just remove the implementation until it is actually doing something useful.

Another option is to actually share the code, however that might happen (actually build Core's .cpp, extra include with implementation, macro resulting in the implementation code, etc.)

> // No mem barrier here, the call to a global function implies global state updated

Was the comment wrong?

dok-net: The host-test version of `esp_try_delay` is identical to the main one, now. Does no harm this way, despite you are right that at present, there cannot be any async cancellation.
Fixed the comment specifying `esp_try_delay`.
The memory fence in the PWM routine is safer for any code changes, like inlining, at some point in time, I don't think such micro-optimizations by hand are a great idea for stability.

mcspr (approved):

mcspr:
@dok-net can you merge master branch here or unlock the modifications by the maintainers on the left column?

dok-net:
@mcspr Done

Issue #7156, opened 2020-03-16
`HTTPClient` does not follow POST requests (feature request)


PsychoXIVI:
Feature request: `HTTPClient` stands for using `HTTP 1.1` so it should allow to follow POST requests.

More information:
https://softwareengineering.stackexchange.com/questions/99894/why-doesnt-http-have-post-redirect

[The RFC document](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) mentioned in code require client confirmation to follow request with method other than GET and HEAD, so I propose additional flag to force follow the redirection (or extension to existing one, by enum?).

_I will try to implement it tonight myself._

PsychoXIVI:
_Please add [fixed-by-pr] by #7157 ?_

PsychoXIVI:
Solved by #7157 , which had just merged :)

PR #7157, opened 2020-03-17
Better follow redirection for `HTTPClient`

PsychoXIVI:
Solves #7156 

First commit adds way to force follow redirections. [The RFC document](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) does not allow to redirect request with methods other than GET/HEAD without any user confirmation. To workaround that, and there is now `setFollowRedirects(followRedirects_t follow)` which allow library user to choose between following redirects as RFC allow without user confirmation (default); or force redirects - as user confirmed the redirection. There is still `setFollowRedirects(bool follow)` left, marked as deprecated.

Second commit makes the client follow most other implementations about `302` HTTP code (`Found` (1.1), previously `Moved temporarily` (1.0)). As mentioned already in the RFC:
```
Note: RFC 1945 and RFC 2068 specify that the client is not allowed
      to change the method on the redirected request.  However, **most
      existing user agent implementations treat 302 as if it were a 303
      response**, performing a GET on the Location field-value regardless
      of the original request method. The status codes 303 and 307 have
      been added for servers that wish to make unambiguously clear which
      kind of reaction is expected of the client.
```
This behavior is required in some cases, even in modern world. My today's example: Google Scripts Apps endpoint. I have `POST` handler setup to run some macro. It should response with JSON containing number of affected rows, but the server handle first POST request (execute macro) and response with redirect request with `302` and other location. Then my code needs to GET from the other location. By browsers (Chrome&Firefox) it works just nicely - as they understand 302 "correctly" (well, not in terms of HTTP 1.0). This commit brings this important behavior to the library.

Also, there is a small rewrite of `sendRequest` code just to make it more clear, without `do {} while();`. Used recursion - as it was also used in other parts. There is only one `int` in this whole function declared, so I think its fine.


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
```diff
@@ -130,6 +130,12 @@ typedef enum {
     HTTPC_TE_CHUNKED
 } transferEncoding_t;
 
+typedef enum {
```
devyte: This requires a Big Phat Comment explaining what each means (e. g. your pr explanation). At a glance, I can't tell the difference between follow and force. (I'm thinking of looking at this in the future).
Also, maybe better naming? Although I can't come up with anything better off the top of my head. If you can't either, just leave it as it is.

d-a-v:
@PsychoXIVI 
CI requires you also updates libraries so there's no warning:
```
libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp:264:45: error: 'void HTTPClient::setFollowRedirects(bool)' is deprecated (declared at /home/travis/arduino_ide/hardware/esp8266com/esp8266/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h:184) [-Werror=deprecated-declarations]
     http.setFollowRedirects(_followRedirects);
                                             ^
cc1plus: all warnings being treated as errors
```

PsychoXIVI:
About that warning, should I make compiler ignore this warning using something like:
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored  "-Wdeprecated-declarations"
// Here the deprecated function declaration
#pragma GCC diagnostic pop
```
Or what?

d-a-v:
Just fix this file, commit your changes and push them to your branch.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
PsychoXIVI: See `af6f69a` commit

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -652,87 +661,100 @@ int HTTPClient::sendRequest(const char * type, const String& payload)
...
-                // return the redirect instead of handling on failure of setURL()
-                redirect = false;
+                    _redirectCount += 1;
+                    DEBUG_HTTPCLIENT("[HTTP-Client][sendRequest] following redirect (the same method): '%s' redirCount: %d\n", _location.c_str(), _redirectCount);
+                    if (!setURL(_location)) {
+                        DEBUG_HTTPCLIENT("[HTTP-Client][sendRequest] failed setting URL for redirection\n");
+                        // no redirection
+                        break;
+                    }
+                    code = sendRequest(type, payload, size);
```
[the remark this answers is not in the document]
PsychoXIVI: See `15e69a5` commit. Rewritten to remove all recursion (including already existing).

PsychoXIVI:
I will also update `ESP8266httpUpdate` library and tests in `test_sw_http_client.ino`

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
PsychoXIVI: @devyte any further work is required? PTAL

devyte: Still not clear at a glance. Quoting the rfc is good as authoritative reference, but a minimal summary of the meanings should be here to not have to go look it up and dig through it.

devyte: e. g. : strict means get redirects to get and put to put, force means <insert meaning here> despite not being rfc compliant because <insert reason here>.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -700,42 +708,67 @@ int HTTPClient::sendRequest(const char * type, const uint8_t * payload, size_t s
...
+                        // redirect using the same request method and payload, diffrent URL
+                        redirect = true;
+                    }
+                    break;
+                }
+                // redirecting with method dropped to GET or HEAD
+                // note: it does not need `HTTPC_FORCE_FOLLOW_REDIRECTS` for any method
+                case HTTP_CODE_FOUND:
+                case HTTP_CODE_SEE_OTHER: {
+                    redirectCount += 1;
```
devyte: Prefer ++redirectCount here

PsychoXIVI: Is it really that bad? Is there some official style guide to follow? This is really picky one TBH

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
PsychoXIVI: See https://github.com/esp8266/Arduino/pull/7157/commits/9510d166e74aaa567804a38ac96857ce05fce30f . I don't think that can be more explanatory, without citing RFC or something. Also, most users (as default) will use RFC strict mode. I think Google will also do the job, if someone is looking for ["how to follow post redirection for httpclient in esp8266"](http://google.com/search?q=how%20to%20follow%20post%20redirection%20for%20httpclient%20in%20esp8266).

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
devyte: About the comment, it's better. In a nutshell, you summarized the problem when opening the issue, by explaining the distinction between the two cades, and that's what's needed here. 
About the increment, it was more a "fix it while you're here" kind of thing.

PR #7167, opened 2020-03-26
Attribute `__real_system_restart_local()` as `(noreturn)`

jjsuwa:
As well as `abort()`.


d-a-v:
What would be nice is to move repeated declarations into a single dedicated file
`core/esp8266/coredecls.h`

PR #7179, opened 2020-04-01
Tone limit 20KHz 

Tech-TX:
Add 2 comments since I found out where the 5KHz upper limit on Tone output frequency was, apparently added for compatibility with the Arduino library some time in the past.

If you want the full 20KHz audio frequency range, change the two lines in Tone.cpp at line 38 to:
```cpp
  high = std::max(high, (uint32_t)25);  // new 20KHz maximum Tone frequency
  low = std::max(low, (uint32_t)25);
```

### DIFF
--- a/cores/esp8266/Tone.cpp
+++ b/cores/esp8266/Tone.cpp
@@ -38,9 +38,7 @@ static void _startTone(uint8_t _pin, uint32_t high, uint32_t low, unsigned long 
   high = std::max(high, (uint32_t)100);  // 5KHz maximum tone for Arduino compatibility
   low = std::max(low, (uint32_t)100);    // (100us high + 100us low period = 5KHz)
 
-  duration *= 1000UL;
-  duration -= duration % (high + low);
-  if (startWaveform(_pin, high, low, duration)) {
+  if (startWaveform(_pin, high, low, (uint32_t) duration * 1000)) {
     _toneMap |= 1 << _pin;
   }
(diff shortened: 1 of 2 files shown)


cores/esp8266/Tone.cpp
```diff
@@ -35,8 +35,8 @@ static void _startTone(uint8_t _pin, uint32_t high, uint32_t low, unsigned long
 
   pinMode(_pin, OUTPUT);
 
-  high = std::max(high, (uint32_t)100);
-  low = std::max(low, (uint32_t)100);
+  high = std::max(high, (uint32_t)100);  // 5KHz maximum tone for Arduino compatibility
```
earlephilhower, line 38: Why not just set it to 25 here (i.e. 20Khz max)?  I think you're fine allowing a larger range than Arduino old-style core.

Tech-TX:
Done, and done!  I'll good with removing artificial restraints on operation. :smile:
Travis CI running now.

cores/esp8266/Tone.cpp
Tech-TX: I'm good with it as well.  It's in.

Tech-TX:
FYI, I've already tested this change yesterday much farther than 20KHz.  When I tried to take it to 65535 it blew up, but I sort of expected that based on your notes elsewhere about CPU utilization at 40KHz PWM.

PR #7180, opened 2020-04-01
Move floating point from libgcc out of IRAM

s-hadinger:
In Tasmota I realized the following floating point function are in precious IRAM:

```
                0x0000000040107184                __divsf3
                0x0000000040107240                __eqsf2
                0x0000000040107240                __nesf2
                0x0000000040107268                __gtsf2
                0x0000000040107288                __lesf2
                0x00000000401072d4                __gesf2
                0x00000000401072f4                __ltsf2
                0x0000000040107340                __unordsf2
                0x0000000040107364                __fixsfsi
                0x00000000401073a4                __nedf2
                0x00000000401073a4                __eqdf2
                0x00000000401073d4                __gtdf2
                0x00000000401073f8                __ledf2
                0x000000004010745c                __gedf2
                0x0000000040107480                __ltdf2
                0x00000000401074e4                __unorddf2
```

This PR adds a `#define FP_IN_IROM` to move floating point function out of IRAM to IROM.

- `*f2.o`,`*f3.o` : `float` and `double` functions
- `*fsi.o`: fp to int
- `*fdi.o`: fp to long
- `*ifs.o`: int to fp
- `*ids.o`: long to fp

In Tasmota, the only functions from `libgcc` remaining in IRAM are: `__modsi3` (integer modulus) and `__ashrdi3` (bit shifting).

### DIFF
--- a/tools/sdk/ld/eagle.app.v6.common.ld.h
+++ b/tools/sdk/ld/eagle.app.v6.common.ld.h
@@ -151,6 +151,14 @@ SECTIONS
 
     *libc.a:(.literal .text .literal.* .text.*)
     *libm.a:(.literal .text .literal.* .text.*)
+#ifdef FP_IN_IROM
+    *libgcc.a:*f2.o(.literal .text)
+    *libgcc.a:*f3.o(.literal .text)
+    *libgcc.a:*fsi.o(.literal .text)
+    *libgcc.a:*fdi.o(.literal .text)
+    *libgcc.a:*ifs.o(.literal .text)
+    *libgcc.a:*idf.o(.literal .text)
+#endif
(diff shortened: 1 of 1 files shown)


s-hadinger:
I totally agree. Actually this is already the case. By default the FP libs are still in IRAM as before, and you need to explicitly `#define FP_IN_IROM` to move them out of IRAM. No braking change.

The added lines are in the ` .irom0.text` section.

Gisi0:
How can I use it?
Use of #define FP_IN_IROM in the source code seems not working.
I have no changes in the size.

using: framework-arduinoespressif8266 483bfde

s-hadinger:
I just realized the #defines were not automatically used when generating ld scripts. I created a new PR to solve this: #7188

Issue #7002, opened 2020-01-08
isDir always false, so code has no effect


brownrb:
in FSBrowser.ino

Line 195 sets isDir false
    bool isDir = false;

then a test in line 197 tests the state of isDir which will always be false
 output += (isDir) ? "dir" : "file"

Nowhere is the value of isDir changed, so this code has an error.

Jeroen88:
Maybe bool isDir = dir.isDirectory() will work

francescolavra:
Indeed the line:
```
bool isDir = false;
```
should be changed to:
```
bool isDir = entry.isDirectory();
```
(The OP is talking about the ESP8266WebServer example sketch at https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino)

brownrb:
This fails for ESP8266 because entry.isDirectory() is invalid for ESP8266

francescolavra:
`entry.isDirectory()` returns always false if you use SPIFFS (because of a limitation in SPIFFS), but should work if you use LittleFS.

earlephilhower:
@brownrb would you like to do a PR with this small change, since you found it?

brownrb:
What is a PR

d-a-v:
Pull Request on github
Merge Request elsewhere
It's a patch proposal

earlephilhower:
@brownrb , could you generate a pull request with the fix?  You've found the bug and solution, and it would be a good way to get your name in the commit logs for posterity. :)

brownrb:
What is a pull request?

     On Monday, February 10, 2020, 06:19:56 AM GMT+13, Earle F. Philhower, III <notifications@github.com> wrote:  
 
 
@brownrb , could you generate a pull request with the fix? You've found the bug and solution, and it would be a good way to get your name in the commit logs for posterity. :)

earlephilhower:
A quick test with LittleFS and directories shows the fix to be more complicated than it looks.

Replacing with `isDir = entry.isDirectory()` does result in the specific results working (i.e. the JSON this pseudi-cgi generates will contain 'dir's), but the actual web page JavaScript (minified so it's not human readable) doesn't know what to do with an entry of type 'dir' so simply drops directories from the list.

So, the fix needs to also incorporate a new .js file that can handle subdirs.  It's a can of worms. :(

vdeconinck:
[...]

PR #7182, opened 2020-04-02
Revamp of the FSBrowser and SDWebServer examples

vdeconinck:
Following a discussion in issue https://github.com/esp8266/Arduino/issues/7002 , I proposed to unify the versions (FSBrowser and SDWebServer) and make it a "universal" FS browser. I also tried to modernize calls (e.g. using SDFS instead of SD).
One thing leading to another, I ended up improving usability and adding many features, such as a file size, free space bargraph and rename/move features.
So this is kind of "FSBrowser on steroïds". Hope you'll like it. 
Please don't hesitate to test and give feedback.

Here is the hi-level changelog :
### Fixes to work on LittleFS based on SDFS
- #define logic to select FS
- switched from SD to SDFS
- begin() does not support parameters > removed SS and added optional config
- LittleFS.open() second parametsr is mandatory > specified "r" where needed
- 'FILE_WRITE' was not declared in this scope > replaced by "w"

### UI/usability improvements
- Never format filesystem, just return "FS INIT ERROR" when FS cannot be mounted 
- Tree panel width is now proportional (20%) to see long names on big screens 
- Added icons for files, and indented them to the same level as folders
- Changed file/folder icon set to use lighter and more neutral one, and added specific "text" and "image" icons for formats recognized as such
- Items are now sorted (folders first, then plain files, each in alphabetic order)
- Added file size after each file name
- Added FS status information at the top right
- Made clear that an async operation is in progress by dimming screen and showing opertation status
- Filled filename box in header with the name of the last clicked file
- Selecting a file for upload defaults to putting it in the same folder as the last clicked file
- Removed limitation to 8.3 lowercase filenames
- Support Filenames without extension, Dirnames with extension
- Improved recursive refresh of parts of the tree (e.g. refresh folder upon file delete, show last folder upon creating nested file)
- Added Save/Discard/Help buttons to ACE editor, discard confirmation on leave, and refresh tree and status upon save
- Removed "Upload" from context menu (which didn't work anyway)
- Added "Rename/Move" feature to context menu

### Other
- Added a readme.md, as well as the PNG source for the icons and both full text and minified gz versions of the web page


d-a-v:
Thanks !
You should run the style script before next push
```
$ cd tests
$ ./restyle.sh
$ git add -p 
  # (add/ the re-formated changes)
```

earlephilhower:
Awesome, @vdeconinck !

In your commit you'll need to do astyle like @d-a-v mentioned to make sure the example has consistent indentation/etc.

Also, please `#define` a filesystem in the example near the top so your error goes away.  We build everything as-is in CI and need no errors/warnings.

````
/home/travis/build/esp8266/Arduino/libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino:58:4: error: #error Please select a filesystem first by uncommenting one of the "#define USE_xxx" lines at the beginning of the sketch.
   #error Please select a filesystem first by uncommenting one of the "#define USE_xxx" lines at the beginning of the sketch.
    ^
````

You could also do something like
````
//#define USE_SPIFFS
//#define USE_LITTLEFS
//#define USE_SDFS
#if !defined(USE_SPIFFS) && !defined(USE_LITTLEFS) && !defined(USE_SDFS)
  #define USE_LITTLEFS
#endif
````

**Comment by vdeconinck** [comment]
CI seemed to be stuck, so I just made a dummy edit following @earlephilhower 's advice.
Also, here's a picture with most of the UI changes to the FSBrowser :
New features

**Comment by d-a-v** [comment]
Github being currently broken, here's the link [to the latest CI](https://travis-ci.org/github/esp8266/Arduino/jobs/670545264?utm_medium=notification&utm_source=github_status) to help you fixing.
You can also use the `test/run_CI_locally.sh` script.

**Comment by vdeconinck** [comment]
Hi @d-a-v. 
Thanks for the link. I had a look at the output and fixed the only "warning as error" it found.
I also tried run_CI_locally.sh but, being on Windows, I'm using a Ubuntu VM and it does not include dev tools so It's not straightforward. Installed git but now I'm getting "Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set)." error and I think I might as well wait for Github to fix the CI integration suite.
Unless you think time is running against us of course.

**Comment by d-a-v** [comment]
> Unless you think time is running against us of course.

I have learned to be very patient :)
Hopefully GH is fixed very soon, we asked them about this issue with CI.

**Comment by vdeconinck** [comment]
Hi. The "Mock trivial test" keeps failing with errors like "‘getFileError’ was not declared in this scope" because it seems to enforce the order of function declarations (like in C) while the .ino sketch concept normally takes care of declaring functions even if they are defined further in the source file. 
Do I need to reorder the functions to please the test or is it too strict ?
KR, Vincent

**Comment by earlephilhower** [comment]
@vdeconinck yes, please do move the function definition (or prototype) before the use.  Arduino runs the `.ino` file through a hackish preprocessor to generate a header file which it silently includes.  The host mock tests don't do that, so you get the same kinds of C++ errors as if you were compiling in straight g++.

**Comment by d-a-v** [comment]
You finally got through this, congrats!
Before final review, would you consider checking #7134 which updates FSbrowser too.
If you are interested, you are free to integrate this webserver API minor update in your PR so your examples may be simplified, as reference examples for how to efficiently make a webserver.

**Comment by vdeconinck** [comment]
Sorry in advance if I'm asking stupid questions.
I get what #7134 is about, and adding the more efficient chunk API is great indeed. 
I'd be glad to update FSBrowser following what you did, but I guess #7134 should be merged first before I can pull the changes to ESP8266WebServer.h and update #7182 , right ?
In other words, I'm all for it but I don't know how to proceed... :-)

**Comment by d-a-v** [comment]
I was proposing you integrate the changes in your PR.
I was feeling bad to merge #7134 before yours since it will generate a conflict for you, given that #7134 is old and you have been working on yours very recently.
We can indeed merge it first, then you integrate it afterwards.

**Comment by vdeconinck** [comment]
Yes, I suggest you merge first and will resolve the conflict afterwards to keep things separate. 
You were there first in the end :-)

**Comment by devyte** [comment]
CI running for #7134, will be merged shortly.

**Comment by devyte** [comment]
#7134 is merged, please resolve conflicts.

**Comment by vdeconinck** [comment]
OK, had to rebase to get all the changes to my tree (don't know why), but here we are, hopefully :-)
Question : may I still make a few improvements ? 
I'd like to borrow the idea from [here](https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/SPIFFSEditor.cpp) to (optionally) embed index.htm.gz in the source code. That way, the example can be flashed to an ESP with valuable data on the internal FS without requiring to format it to upload the html page.
Now if you think this should be commited as-is first, no problem of course...

File: libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino [hunk 1]
```diff
@@ -41,197 +67,534 @@ FS* filesystem = &LittleFS;
...
+*/
 void handleFileCreate() {
-  if (server.args() == 0) {
-    return server.send(500, "text/plain", "BAD ARGS");
+  if (!fsOK) {
+    return returnFail("FS INIT ERROR");
   }
-  String path = server.arg(0);
-  DBG_OUTPUT_PORT.println("handleFileCreate: " + path);
+  if (server.args() == 0 || server.args() > 2) {
```
d-a-v, line 365: Checking `server.args()` will lead to confusion, at least until #6768 which restores the original and expected `::args()` definition is merged.
Indeed in a previous commit from this PR (see comments there), `args()` is increased sometimes.
Can you avoid using it and instead check for the existence (or not) of the expected arguments ?

d-a-v:
> Question : may I still make a few improvements ?

Yes, the idea of demonstrating how to include compressed files in source code is interesting but must remain optional as it requires more work and disallow dynamically changing these files (ie: reflashing is needed). 
On the other hand, it can be interesting to propose a *default* file when the real one is not present in the filesystem.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
vdeconinck: No problem. I rewrote the tests to check for presence of server arguments.

earlephilhower:
@vdeconinck another INO style error.  You can run `tests/restyle-examples-only.sh` to fix it before committing.

vdeconinck:
Oh Damn. I'll fix that of course.
Sorry :-/

vdeconinck:
OK, I think I'm done now :-). 
All comments are welcome.

vdeconinck:
As discussed a few weeks ago, FSBrowser can now work without web access. 

Hope I didn't break anything :-/ 
I promise it's the last feature I wanted to add, so this PR is final, apart from potential change requests of course...
Thanks for your patience.

vdeconinck:
Thanks for your kind words. 
I'm standing on the shoulders of giants, you know, so my name in the commit history will be enough... Plus, I promised not to do any more commits ;-P

vdeconinck:
(fixed upload on a full filesystem that got silently ignored + use <meter> tag for disk usage, which changes color when approaching filesystem full). 
Sorry, I know I promised, but it seemed worth it.

devyte:
It's ok if you want to continue tweaking and fixing bugs as you find them. Better now than later!

vdeconinck:
Thanks :-).
But I think I'm really done now, so it would be great if you could review it.

vdeconinck:
Hi Earle, thanks for reviewing and for the feedback.
The .gz version was first passed through a minifier (see `extras/reduce_index.sh`) before being compressed, so the code is different but should behave exactly the same.
The goal was to help people with limited FS space to use the minimal version by just deleting index.htm (index.htm.gz then takes over), as indicated in the "options" part of the readme.
But you're right, it makes things harder to understand, and the goal of examples is obviously not to make things unnecessarily complex.
I just changed so that the gzipped version only stays in the "extras" folder, and updated the readme instructions on how to use it.
(Note: I chose to include both the gzipped file and the .h version in the extras folder because they are not trivial to build on a Windows system)
Don't hesitate to tell me if you think it's not the best way.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+   The "Not Found" handler catches all URI not explicitely declared in code
+   First try to find and return the requested file from the filesystem,
+   and if it fails, return a 404 page with debug information
+*/
+void handleNotFound() {
+  if (!fsOK) {
+    return returnFail("FS INIT ERROR");
+  }
+  String uri = urlDecode(server.uri());
+  if (!handleFileRead(uri)) {
```
devyte, line 547: Prefer direct logic over inverted:
```cpp
if(handleFileRead(uri)
  return;

String message ...
...
return replyNotFound(message);
}
````

Again, please avoid/fix such sequences of concats. Call String::reserve() first with an appropriate value.
And please move the strings to flash.

File: libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino [hunk 3]
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+    // Dump debug data
+    String message = "Error: File not found\n\n";
+    message += "URI: ";
+    message += uri;
+    message += "\nMethod: ";
+    message += (server.method() == HTTP_GET) ? "GET" : "POST";
+    message += "\nArguments: ";
+    message += server.args();
+    message += "\n";
```
devyte, line 556: prefer single char '\n' over single char strings.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+    for (uint8_t i = 0; i < server.args(); i++) {
+      message += String(" NAME:") + server.argName(i) + "\n VALUE:" + server.arg(i) + "\n";
```
devyte, line 558: strings to flash

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+}
+
+/*
+   This specific handler returns the index.htm (or a gzipped version) from the /edit folder.
+   If the file is not present but the flag INCLUDE_FALLBACK_INDEX_HTM has been set, falls back to the version
+   embedded in the program code.
+   Otherwise, fails with a 404 page with debug information
+*/
+void handleGetEdit() {
+  if (!handleFileRead("/edit/index.htm")) {
```
devyte, line 573: Prefer direct logic over reversed to reduce cyclomatic complexity

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -257,54 +628,48 @@ void setup(void) {
...
 
+  // Filesystem status
+  server.on("/status", HTTP_GET, handleStatus);
 
-  //SERVER INIT
-  //list directory
+  // List directory
   server.on("/list", HTTP_GET, handleFileList);
-  //load editor
-  server.on("/edit", HTTP_GET, []() {
```
devyte, line 270: I seem to remember there was a reason for this special handling, but I can't for the life of me remember what it was.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -257,54 +628,48 @@ void setup(void) {
...
-  //get heap status, analog input value and all GPIO statuses in one json call
-  server.on("/all", HTTP_GET, []() {
-    String json('{');
-    json += "\"heap\":" + String(ESP.getFreeHeap());
-    json += ", \"analog\":" + String(analogRead(A0));
-    json += ", \"gpio\":" + String((uint32_t)(((GPI | GPO) & 0xFFFF) | ((GP16I & 0x01) << 16)));
-    json += "}";
-    server.send(200, "text/json", json);
-    json.clear();
-  });
```
devyte, line 302: Why remove this?

libraries/ESP8266WebServer/examples/FSBrowser/data/edit/index.htm
```diff
@@ -0,0 +1,1122 @@
...
+        border: 16px solid #eee;
+        border-right-color: transparent;
+        border-radius: 32px;
+        box-sizing: border-box;
+        overflow: hidden;
+        text-indent: -9999px;
+        width: 64px;
+        height: 64px;
+      }
+    </style>
```
devyte, line 252: Does it make sense to move this out into a .css?

libraries/ESP8266WebServer/examples/FSBrowser/data/edit/index.htm
```diff
@@ -0,0 +1,1122 @@
...
+        var vars = {};
+        var parts = window.location.href.replace(/[?&]+([^=&]+)=([^&]*)/gi, function(m,key,value) { vars[key] = value; });        
+        if (typeof ace != "undefined") {
+          var editor = createEditor("editor", vars.file, vars.lang, vars.theme);
+        }
+        tree = createTree("tree", editor);
+        createHeader("header", tree, editor);
+        refreshStatus();
+      };
+    </script>
```
devyte, line 1102: Does it make sense to move this whole thing into its own js?
It's considered best practice to try to maintain separation between html/js/css as far as is reasonable.

libraries/ESP8266WebServer/examples/FSBrowser/extras/index_htm.h
```diff
@@ -0,0 +1,525 @@
+unsigned char index_htm_gz[] = {
```
devyte, line 1: There should be a BigPhatComment somewhere explaining that modifications to index.htm should be accompanied by an update to this.
In fact, that comment should be in in the .ino, inside index.htm, and in here, at the minimum.

libraries/ESP8266WebServer/examples/FSBrowser/extras/index_htm.h
```diff
@@ -0,0 +1,525 @@
...
+  0xa2, 0x9e, 0x3a, 0x05, 0x31, 0xe5, 0xb2, 0xd7, 0x86, 0x8c, 0x34, 0x76,
+  0x8a, 0xc8, 0x1d, 0x06, 0xef, 0x9f, 0x77, 0x28, 0x62, 0xb7, 0xb6, 0xbd,
+  0x89, 0xbb, 0x63, 0xdd, 0xe2, 0x6d, 0xb7, 0x6e, 0x2a, 0xd3, 0x44, 0xb9,
+  0xd5, 0xfa, 0xbf, 0xcd, 0x58, 0x1d, 0x95, 0xb7, 0x13, 0x75, 0xf5, 0x64,
+  0x5a, 0x19, 0x67, 0x5b, 0xd4, 0x75, 0x92, 0xce, 0xb8, 0x3f, 0x17, 0x95,
+  0xe9, 0x65, 0xb0, 0x91, 0x92, 0x44, 0x84, 0xad, 0xa4, 0x71, 0xf9, 0xb3,
+  0x20, 0x88, 0x0e, 0x8e, 0x67, 0x82, 0x76, 0xfa, 0x27, 0xc4, 0xbd, 0xa3,
+  0x7f, 0xb9, 0xa1, 0x68, 0x0e, 0x38, 0x4b, 0x00, 0x00
+};
+unsigned int index_htm_gz_len = 6261;
```
devyte, line 525: What is the purpose of this? Manually setting such size values is error prone. Prefer use of sizeof().

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
+# FSBrowser readme
+
+## What is this sketch about ?
+
+This example is a FileSystem Browser for the ESP8266 using http requests and a html/javascript frontend, 
+working for both SPIFFS, LittleFS and SDFS.
```
devyte, line 6: "work for all of" (both implies count of 2, here it's 3)

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+This unified version is based on the previous examples named FSWebServer, FSBrowser and SDWebServer, Copyright (c) 2015 Hristo Gochkov. All rights reserved.
+
+## How to use ?
```
devyte, line 9: How to use **it**

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+1. Uncomment one of the `#define USE_xxx` directives in the sketch
+2. Add the credentials of your WiFi network (search for `STASSID`)
+3. Compile and upload the sketch to your ESP8266 device
+4. For normal use, copy the contents of the `data` folder to the filesystem. To do so:
+- for SDFS, copy that contents (not the data folder itself, just its contents) to the root of a FAT/FAT32-formated SD card connected to the SPI port of the ESP8266
```
devyte, line 14: ..., copy the files inside the data folder... to the root of...

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+- for SPIFFS or LittleFS, please follow the instructions at https://arduino-esp8266.readthedocs.io/en/latest/filesystem.html#uploading-files-to-file-system
+5. Once the data and sketch have been uploaded, access the editor by pointing your browser to http://fsbrowser.local/edit
```
devyte, line 16: /edit or /edit.htm?

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Options
+If you need to free some space on the ESP filesystem, you can delete all the sample files at the root but also replace the `index.htm` file in the `data/edit` subfolder by the `index.htm.gz` file from the `extras` folder. That compressed version is not suited for learning or debugging, but will bring the total FS usage under 7KB.
```
devyte, line 19: at the expense of a higher binary size, i.e.: the fs size reduction is not for free, as explained below.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+If you want to use the browser on a an existing filesystem or don't want to perform step 4 above, you have two possibilities :
+- either upload the `index.htm` file to the filesystem by opening a command shell in the `data` folder and running the following cURL command:
+`curl -F file=@edit/index.htm;filename=/edit/index.htm fsbrowser.local/edit`
+- or embed a version of the html page in the source code itself by uncommenting the following line in the sketch and rebuilding:
+`#define INCLUDE_FALLBACK_INDEX_HTM`
+That embedded version is functionally equivalent and will be returned if no `/edit/index.htm` or `/edit/index.htm.gz` file can be found on the filesystem, at the cost of a larger compiled file.
+
+To regenerate the files required for the previous options or understand how they are build, see the `reduce_index.sh` script located in the `extras` subfolder.
```
devyte, line 27: ...how they are built, ...

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Dependency
+The html page uses the [ace.js](https://ace.c9.io/) text editor which is loaded from a CDN. Consequently, internet access from your web browser is required for the FSBrowser editing feature to work as-is.
```
devyte, line 30: Which version, at the time of this writing?

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+If your browser has no web access (e.g. if you are connected to the ESP8266 as an access-point), you can copy the `ace.js` file (and dependencies) to the `edit` subfolder of the ESP filesystem.
```
devyte, line 31: How do I figure out the dependencies for ace? It's not a matter of going in there and poking around.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+If `ace.js` cannot be found on the ESP filesystem either, the page will default to a plain text viewer, with a warning message.
+
+## Notes
+- See https://arduino-esp8266.readthedocs.io/en/latest/filesystem.html for more information on FileSystems supported by the ESP8266.
+- For SDFS, if your card's CS pin is not connected the default pin (4), enable the `fileSystemConfig.setCSPin(chipSelectPin);` line, specifying the GPIO the CS pin is connected to
```
devyte, line 36: ...not connected **to** the default...

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
```diff
@@ -0,0 +1,121 @@
...
+- In subdir : MkFile '/dir/2.txt' / List / Edit / Download / Delete / Upload '/dir/2.png' / View image / Mkdir '/dir/sub'
+- Delete root folder '/dir'
+- Create nested file '/a/b.txt' and delete file 'b.txt', then delete '/a'
+#### Long filenames
+- At root : MkFile '/My text file 1.txt' / List / Edit / Download / Delete / Upload '/My image file 1.png' / View image / Delete image / Mkdir '/My Directory'
+- In subdir : MkFile '/My Directory/My text file 2.txt' / List / Edit / Download / Delete / Upload '/My Directory/My image file 2.png' / View image / Mkdir '/My Directory/My Subdirectory'
+- Delete root folder '/My Directory'
+- Create nested file '/My folder/My test file.txt' and delete file 'My test file.txt' 
+
+## Credits
```
devyte, line 116: Maybe add something like:
Original version of FSBrowser written by me-no-dev, contributions over time by various contributors

devyte:
I still need to test this a bit.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: Hmm, I admit I'm a bit puzzled too, and can't find a definitive answer. 
Seems ace.js dynamically loads additional files according to supported file formats, preferred keymaps, themes, etc. The full list is at https://github.com/ajaxorg/ace-builds/tree/master/src but I believe the editor works with none loaded and just falls back to default settings if files cannot be loaded. 
I'd suggest including this in the readme :
`copy the `ace.js` file to the `edit` subfolder of the ESP filesystem, as well as optional themes, plugins etc. according to your needs (see https://github.com/ajaxorg/ace-builds for a full list).`
What do you think ?

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+   ---------------+--------------------------------------------------------------
+   Delete file    | parent of deleted file, or remaining ancestor
+   Delete folder  | parent of deleted folder, or remaining ancestor
+*/
+void handleFileDelete() {
+  if (!fsOK) {
+    return returnFail("FS INIT ERROR");
+  }
+
+  String path = server.arg(0);
```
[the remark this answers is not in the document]
vdeconinck: Testing for argument presence makes sense of course.
Regarding isEmpty(), I always hesitate to use "sub-documented methods" ;-).
E.g., isEmpty() and clear() are not listed on https://www.arduino.cc/reference/en/language/variables/data-types/stringobject/ , but they seem to work fine so I'm switching to isEmpty.
Being mostly a Java programmer, I admit I often find that Arduino classes documentation is lacking, but probably it's just me ;-) .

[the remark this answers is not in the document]
vdeconinck: Err... Just remembered this review comment by @d-a-v against the use of args():
https://github.com/esp8266/Arduino/pull/7182#pullrequestreview-389887075
So in the end I'm not sure...

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -41,150 +77,180 @@ FS* filesystem = &LittleFS;
...
+  }
+  if (filename.endsWith(".zip")) {
     return "application/x-zip";
-  } else if (filename.endsWith(".gz")) {
+  }
+  if (filename.endsWith(".gz")) {
     return "application/x-gzip";
   }
   return "text/plain";
 }
```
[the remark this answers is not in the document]
vdeconinck: The octet-stream (download) case was moved up from getContentType() to the caller:
```
  if (server.hasArg("download")) {
    contentType = "application/octet-stream";
  } else {
    contentType = getContentType(path);
  }
```
because I found it weird for getContentType() to return a value based on the `server` global variable, and ignoring the parameter. 
Btw it now makes it compatible with the implementation you suggest from StaticRequestHandler :-)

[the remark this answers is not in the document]
vdeconinck: Anyway, using that StaticRequestHandler implementation makes perfect sense, but I admit I don't know how to call it... Tried this:
```
contentType = StaticRequestHandler::getContentType(path);
```
but I'm getting a `'StaticRequestHandler' has not been declared` error.
Do I need to include something specific ?
I guess my C++ limitations are showing up... :-(

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
vdeconinck: I thought it was a leftover from another example, because as far as I could see, it was unused by the html code, and had nothing to do with filesystem browsing. 
Now if you think it's of any use here, I can add it back of course...

libraries/ESP8266WebServer/examples/FSBrowser/data/edit/index.htm
vdeconinck: Hmm. That's a good question.
For the sake of cleanliness, I'd move the css, js and the icons to separate files, but for ease of use, just dropping a single, self-contained, html file on the filesystem sounds much simpler.
It also makes compression easier, and makes the "embed html inside the program" relatively simple.
So all in all, I would leave it as is...

libraries/ESP8266WebServer/examples/FSBrowser/data/edit/index.htm
vdeconinck: See comment for the CSS.
I think this sketch is half an example of how to use the ESP8266WebServer lib (plus many other things, such as FileSystem calls), and half a real utility that could become a separate project on its own.
So it is a balance between education (which would tend to make things more explicit, reduce fancy cases, etc) and real-world usage.
I admit what motivated me to dig down inside the code is that I include that sketch in all my ESP projects as a way to debug and manipulate my filesystem contents.

libraries/ESP8266WebServer/examples/FSBrowser/extras/index_htm.h
vdeconinck: This file is the untouched output of `xxd -i`.
I agree that is seems redundant with a sizeof() but I guess it's zero risk here as it's auto-generated.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: Thanks. 
English is not my mother tongue, so all such tips are greatly appreciated.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: Well, I meant both files and folders have to be copied, so I was looking for a generic term and found 'contents'. 
Although technically a folder is a file, I guess most people tend to think otherwise.
If you think "files" is best, just tell me.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: It's `edit`.
`edit` is the folder, and the default file is `index.htm`.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -41,150 +77,180 @@ FS* filesystem = &LittleFS;
...
-    filename.clear();
-  } else if (upload.status == UPLOAD_FILE_WRITE) {
-    //DBG_OUTPUT_PORT.print("handleFileUpload Data: "); DBG_OUTPUT_PORT.println(upload.currentSize);
-    if (fsUploadFile) {
-      fsUploadFile.write(upload.buf, upload.currentSize);
-    }
-  } else if (upload.status == UPLOAD_FILE_END) {
-    if (fsUploadFile) {
-      fsUploadFile.close();
+String urlDecode(String str) {
```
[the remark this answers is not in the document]
vdeconinck: Same here: just replacing the urlDecode() call by ESP8266WebServerTemplate::urlDecode() causes a `'ESP8266WebServerTemplate' has not been declared` error.
I must be missing something. Just a hint please ?

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+  }
+
+  String path = server.arg(0);
+  if (path == "") {
+    return returnFail("BAD ARGS");
+  }
+
+  DBG_OUTPUT_PORT.println(String("handleFileDelete: ") + path);
+  if (path == "/" || !fileSystem->exists(path)) {
+    return returnFail("BAD PATH");
```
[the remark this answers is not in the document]
vdeconinck: Right.
Trying to remove "/" should return an error 500, but a non-existing file should return a 404 for sure.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+   Move file      | parent of source file, or remaining ancestor
+   Rename folder  | parent of source folder
+   Move folder    | parent of source folder, or remaining ancestor
+*/
+void handleFileCreate() {
+  if (!fsOK) {
+    return returnFail("FS INIT ERROR");
+  }
+
+  String path = server.arg("path");
```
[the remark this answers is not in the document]
vdeconinck: Can you advise regarding args() and @d-a-v 's comment https://github.com/esp8266/Arduino/pull/7182#pullrequestreview-389887075 ?

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -226,29 +302,324 @@ void handleFileList() {
...
+  }
+}
+
+
+/*
+   Delete the file or folder designed by the given path.
+   If it's a file, delete it.
+   If it's a folder, delete all nested contents first then the folder itself
+*/
+void deleteRecursive(String path) {
```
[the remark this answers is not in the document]
vdeconinck: Been thinking about this for a few hours, but I'm not sure at all how to do it :-(.
I guess one way or another, I'd have to push the "yet-to-process" items on a custom-made stack, so I wonder if it's worth it... 
Now regarding the passing by reference, the only solution I can see would be to replace the "path" String by a linked list of pointers to each level.
But again, I fear that such an improvement, altough required for a high quality project, would make code more complex and less legible for an example meant to demonstrate usage of the web server library...

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
vdeconinck: By the way, I also removed the data/graphs.js.gz file. Probably they were meant to work together, but no HTML code referred to that javascript file anymore.
Maybe they are valuable and could be used in a separate example, but we'd have to dig through the project history to find how they were meant to be used...

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
```diff
@@ -41,150 +77,180 @@ FS* filesystem = &LittleFS;
...
+    return returnFail("FS INIT ERROR");
+  }
+
   if (!server.hasArg("dir")) {
-    server.send(500, "text/plain", "BAD ARGS");
-    return;
+    return returnFail("BAD ARGS");
+  }
+
+  String path = urlDecode(server.arg("dir"));
```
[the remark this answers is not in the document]
vdeconinck: Hmm. Unfortunately I have a very volatile memory...
All I can say is that I introduced server argument decoding a few weeks ago as part of this commit: https://github.com/esp8266/Arduino/pull/7182/commits/d709ae53049262c8dcf19eba9e5a40a98effca76#diff-d926380a25a7af1975d069c86e3d27d4 , and it was required to fix a bug with filenames containing blanks.
Now I admit after removing the URLdecode() I cannot reproduce the exact bug anymore :-(. 
It would be great to get rid of this code, but if I spent time finding how to URLdecode, I guess there was a good reason.
I'll have to dig further...

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
vdeconinck: Err, that special handler is still there, I just moved it up so that the GET method appears before PUT, DELETE and POST. Sounded more logical to me...

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: The goal is to refactor, not just replace the calls.
An example (off the top of my head) of what you could do is:
- pull the StaticRequestHandler<>::getContentType() function body out into a free standing function inside mime.h|cpp in the mime namespace.
- Replace the StaticRequestHandler::getContentType() body with a call to mime::getContentType() (i.e.: it becomes just a forwarder) to keep compatibility with current code
- #include "mime.h" in the sketch and use mime::getContetType() directly in the sketch if needed
- Add additional mime types to the mimetable if any are missing

This would unify code but still allow reuse by other users who need to getContentType() for whatever reason.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: Does calling ESP8266WebServer::urlDecode(str) not work?

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: This does need corroboration. My thinking is that you get the same return for arg(0) when there was no arg provided as when the first arg has no value.
So e.g. this:
http://myserver.org/index.html
would return the same arg(0) as this:
http://myserver.org/index.html?blah
Which is a Bad Example™, so you need to add the additional condition of how many args are present.
Let's see what @d-a-v says.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: Alright, leave it as is for now, but because this is an example please add a comment above saying something along the lines of "This function uses recursion, don't do this at home or you'll blow up the stack"

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
devyte: I meant the lambda wrapper. NVM, if something breaks we can handle it in a later PR.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
devyte: They **are** supposed to work together, I have a variation of this working in my own webapp, where I read a bunch of values and show a bunch of graphs in a webpage using graphs.js.gz. Most likely the html page front end for /all got lost at some point in the near past.
In this PR I suggest:
- keeping this code, maybe rename /all to something like /data
- keeping graphs.js.gz, maybe see if there's some updated version of it

Then, if you have the time/inclination, do one of the following in a subsequent PR:
- implement the web front-end to show the graphs
or
- propose an alternative way of graphing data that is as self-contained as this one. This one works, but is rather limited.

libraries/ESP8266WebServer/examples/FSBrowser/extras/index_htm.h
devyte: Then that should be explained somewhere in a BigPhatComment™, something like "don't touch that file because it's autogenerated with xxd -i"

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
devyte: ah yes, nvm then

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
devyte: Is /edit then both a service endpoint and a folder? How does that work?

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
devyte: Agreed. However, the most widely used cases for this editor will be html, js, css and json. What do you think of figuring out the minimum files for those specific formats and suggesting those explicitly together with the above comment? I'm trying to think like the newcomer looking at the example for the first time. Looking at that full list is rather overwhelming.
So the comment would be something like:
```
Copy the ace.js file to the edit subfolder of the ESP filesystem, as well as optional themes, plugins etc. according to your needs (see https://github.com/ajaxorg/ace-builds for a full list).
Example:
mode-css.js
mode-html.js
blah.js
...
```

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
d-a-v: https://github.com/esp8266/Arduino/pull/6768#issuecomment-620246696

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
vdeconinck: Oh, I see.  OK, I didn't intend to jump out of the scope of this example.
I'll do my best, but what about making that refactoring a separate PR once this one is out ?
(By the way, comparing the code, it seems "jpeg" is missing from the StaticRequestHandler version)

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
vdeconinck: Oh I get it now. Well I thought the lambda was used because the code was so simple it didn't justify the overhead of a named function.
As the code became a bit more complex with the option to return the embedded binary version, I found it cleaner to create a dedicated handler function like the other calls.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
vdeconinck: Hmm, I'm still tempted to think an example should do one thing and do it well, and this part clearly has nothing to do with filesystem browsing. 
But I agree that if there is valuable code that shows how to display data from the ESP in a web page (I guess), it shouldn't just get lost.
What about creating a dedicating example with that code ? 
I volunteer to do it once this PR is wrapped up...

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: Well, from the user's point of view, you call `/edit` URL from your browser, and you get the `index.htm` file from the `/edit` folder.
Now admittedly `/edit` has specific handlers for the different HTTP methods, and when called in GET, the dedicated handler optionally returns the embedded gzipped version if the `index.htm` file is not found, but functionally, the user must still call `http://fsbrowser.local/edit` to open the FS Browser page.

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: Indeed, I admit as a newcomer to ace.js, I find the list overwhelming ;-).
OK, I'll try to figure out a "reasonable" list and will update the readme accordingly. For example, the theme is fixed in index.htm IIRC, so it is easy to determine which theme file is needed.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: Alright, agreed. 
BTW, you can look me up in our [gitter channel](https://gitter.im/esp8266/Arduino), I could provide better guidance there, it would be much faster.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: It's quite possible. If you can't reproduce the bug, then I suggest removing this.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: Looking at this with fresh eyes, I think your intent is that path must have a value. Not providing the arg and providing the arg without a value fall under the same case: it's a bad path, so let's keep this as it is.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: Given the following test with isEmpty(), leave it as it is for now.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
devyte: In this case I remember that the first time I tried the example I had a problem with SPIFFS self/destructing, so no files were present on boot no matter what I did. That meant that I would pretty much get a 404 everywhere. The only way I could know that the server was actually ok was thanks to this endpoint.
However, agreed, let's split into a specialized example for graphs. It's something that deserves an update anyways. In particular, I think the graph library should be replaced with something standard rather than an in-house lib.

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
devyte: In which case you no longer need the local urlDecode()?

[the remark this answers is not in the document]
vdeconinck: Yep, can't seem to be able to reproduce the bug.
Although I'm pretty sure I introduced it for a reason, the latest source works without it... And I went back to a 2.6.0 install and it still worked without it, so I don't know. Maybe I was chasing the wrong target when I introduced it. 
So indeed, no more call => no more urlDecode() implementation => no more hexDigitToInt(). I like that :-)

libraries/ESP8266WebServer/examples/FSBrowser/FSBrowser.ino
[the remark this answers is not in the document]
vdeconinck: Not needed anymore ;-).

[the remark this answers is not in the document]
vdeconinck: Yes it does ! 
Cleaning up the local version. Thanks :-)

libraries/ESP8266WebServer/examples/FSBrowser/readme.md
vdeconinck: Done. 
I just opened the "Network" developer tool and played with the editori to collect what was actually loaded from the CDN and compile a "typical" list :-).

vdeconinck:
Hi devbyte, First, many thanks for the extensive review and advice. Much appreciated.

Rereading some comments, I realize some moves I made may seem to come out of the blue, but this version is really a merge between the original FSBrowser and SDWebServer. For example, the fact that the page is `/edit/index.htm` and not `/edit.htm[.gz]` comes from SDWebServer. Same for the recursive deletion routine.
I guess it means I really didn't invent anything, I'm mostly the cook who mixed the right ingredients :-).

With the latest requested changes (and removal of local urlDecode()), I think the version is quite OK.

With respect to the "is the server alive ?" use of `/all`, I just compiled a FSBrowser for LittleFS while the filesystem was formated as SPIFFS (note that no formating ever happens anymore in such a case) and all calls return a clean "FS INIT ERROR" message. OTOH, a call to ´/status´ returns the status of the filesystem, which in that case was `{"type":"LittleFS", "isOk":"**false**","unsupportedFiles":""}`, so  I think it's quite usable that way.

Please don't hesitate to test in real life and report of course.

If everything is OK, I now have two PRs on my TODO list:
- Make the lib version of getContentType() callable from user sketches, and adapt FSBrowser accordingly
- Create a basic example showing a live graph of ESP data.

Regarding that last point, do you still have a html page at hand that calls the graphs.js file ? I would like to know how it's meant to work before refactoring.

devyte:
About the two PRs, let's discuss in gitter, just @ me in the channel.

PR #7188, opened 2020-04-03
Ability to define -DFP_IN_IROM when generating ld script

s-hadinger:
Follow up of #7180, there was no way to enable the `-DFP_IN_IROM` flag when generating the ld script.

### DIFF
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -247,9 +247,12 @@ else:
 #
 
 current_vtables = None
+fp_in_irom = ""
 for d in flatten_cppdefines:
     if str(d).startswith("VTABLES_IN_"):
         current_vtables = d
+    if str(d) == "FP_IN_IROM":
+        fp_in_irom = "-DFP_IN_IROM"
 if not current_vtables:
     current_vtables = "VTABLES_IN_FLASH"
     env.Append(CPPDEFINES=[current_vtables])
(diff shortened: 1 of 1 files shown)


tools/platformio-build.py
```diff
@@ -260,7 +263,7 @@ def scons_patched_match_splitext(path, suffixes=None):
     join("$BUILD_DIR", "ld", "local.eagle.app.v6.common.ld"),
     join(FRAMEWORK_DIR, "tools", "sdk", "ld", "eagle.app.v6.common.ld.h"),
     env.VerboseAction(
-        "$CC -CC -E -P -D%s $SOURCE -o $TARGET" % current_vtables,
+        "$CC -CC -E -P -D%s %s $SOURCE -o $TARGET" % (current_vtables, fp_in_irom),
```
mhightower83, line 266: I don't speak pythonese or use platformio; however, shouldn't there be a -D in front of the new '%s'.
Déjà Vu, I changed the same line in a PoC PR https://github.com/esp8266/Arduino/blob/91fc39130e6b39d3f4cf25c3956fda07a7d5c981/tools/platformio-build.py#L274

s-hadinger: No, because the second parameter can be absent. This would result in `-D` being inserted and potentially causing an error.
The `-D` is already included above here: `fp_in_irom = "-DFP_IN_IROM"`

Issue #3812, opened 2017-11-09
Unable to use http.setAuthorization for ESPhttpUpdate


pjunni:
I did not find a way to use http.setAuthorization for ESPhttpUpdate, but I solved the issue by adding the line http.setAuthorization("username","password"); on line 181 in https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp. Is there a more "correct" way to do this or could there be an overloaded handleUpdate function that takes these "username" and "password" and pass them into http.setAuthorization("username","password");?

devyte:
@pjunni if I understand the request correctly, what you want to do is update from a webserver that requires user/pass. Is that correct?

pjunni:
That's correct.

pjunni:
@devyte Would you have any suggestions on this matter?

jatayu86:
I face the similar issue,could someone please let us know how to add http authorization headers for ESPHttpUpdate?

ngashok:
Facing same issue. How to do OTA from https server having authorisation with username/password.

Swiftnesses:
Was this ever solved, I have the same issue?

drhideg:
If you are still interested, would you give it a try this PR? #7190

earlephilhower:
Fixed via the mentioned PR (#7190).  Closing.

PR #7190, opened 2020-04-04
Basic authentication with ESP8266httpUpdate

drhideg:
Add ability to use basic access authentication with ESP8266httpUpdate


drhideg:
Possible soultion for issue #3812 where unable to use http.setAuthorization for ESPhttpUpdate.

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp
```diff
@@ -282,6 +304,16 @@ HTTPUpdateResult ESP8266HTTPUpdate::handleUpdate(HTTPClient& http, const String&
         http.addHeader(F("x-ESP8266-version"), currentVersion);
     }
 
+    if (_user && _user[0] != 0x00 && _password && _password[0] != 0x00)
+    {
+        http.setAuthorization(_user.c_str(), _password.c_str());
+    }
+
+    if (_auth && _auth[0] != 0x00)
```
devyte, line 312: Prefwr _auth. isEmpty()

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp
```diff
@@ -282,6 +304,16 @@ HTTPUpdateResult ESP8266HTTPUpdate::handleUpdate(HTTPClient& http, const String&
         http.addHeader(F("x-ESP8266-version"), currentVersion);
     }
 
+    if (_user && _user[0] != 0x00 && _password && _password[0] != 0x00)
```
devyte, line 307: prefer _user. isEmpty () and _password. isEmpty ()

devyte:
Also, is authorization the right term here? our webserver seems to use authentication for user/pass.

libraries/ESP8266httpUpdate/src/ESP8266httpUpdate.cpp
drhideg: Actually it was a copy-paste solution to meet my needs, from the [ESP8266HTTPClient](https://github.com/esp8266/Arduino/tree/83166f948bedff85543b1cb5532b905746744df7/libraries/ESP8266HTTPClient/src) library.

PR #7192, opened 2020-04-04
Preserve phase relationship when waveform updated

earlephilhower:
Preserve the phase relationship as close as possibly by taking an entire
period of the waveform and copying it over to the machine state, instead
of doing it on rising and falling edges.

### DIFF
--- a/cores/esp8266/core_esp8266_waveform.cpp
+++ b/cores/esp8266/core_esp8266_waveform.cpp
@@ -114,13 +114,17 @@ void setTimer1Callback(uint32_t (*fn)()) {
 // waveform smoothly on next low->high transition.  For immediate change, stopWaveform()
 // first, then it will immediately begin.
 int startWaveform(uint8_t pin, uint32_t timeHighUS, uint32_t timeLowUS, uint32_t runTimeUS) {
+  return startWaveformCycles(pin, microsecondsToClockCycles(timeHighUS), microsecondsToClockCycles(timeLowUS), microsecondsToClockCycles(runTimeUS));
+}
+
+int startWaveformCycles(uint8_t pin, uint32_t timeHighCycles, uint32_t timeLowCycles, uint32_t runTimeCycles) {
   if ((pin > 16) || isFlashInterfacePin(pin)) {
     return false;
   }
   Waveform *wave = &waveform[pin];
 
-  wave->expiryCycle = runTimeUS ? GetCycleCount() + microsecondsToClockCycles(runTimeUS) : 0;
-  if (runTimeUS && !wave->expiryCycle) {
+  wave->expiryCycle = runTimeCycles ? GetCycleCount() + runTimeCycles : 0;
+  if (runTimeCycles && !wave->expiryCycle) {
     wave->expiryCycle = 1; // expiryCycle==0 means no timeout, so avoid setting it
   }
 
@@ -130,14 +134,14 @@ int startWaveform(uint8_t pin, uint32_t timeHighUS, uint32_t timeLowUS, uint32_t
     // Need to manually do RSIL here because this is "C" code and can't use the InterruptLock class
     uint32_t _state = xt_rsil(15);
 
-    wave->gotoTimeLowCycles = microsecondsToClockCycles(timeLowUS);
-    wave->gotoTimeHighCycles = microsecondsToClockCycles(timeHighUS);
+    wave->gotoTimeLowCycles = timeLowCycles;
+    wave->gotoTimeHighCycles = timeHighCycles;
 
     // Restore interrupt state
     xt_wsr_ps(_state);
   } else { //  if (!(waveformEnabled & mask))
-    wave->timeHighCycles = microsecondsToClockCycles(timeHighUS);
-    wave->timeLowCycles = microsecondsToClockCycles(timeLowUS);
+    wave->timeHighCycles = timeHighCycles;
+    wave->timeLowCycles = timeLowCycles;
     wave->gotoTimeHighCycles = wave->timeHighCycles;
     wave->gotoTimeLowCycles = wave->timeLowCycles;
     // Actually set the pin high or low in the IRQ service to guarantee times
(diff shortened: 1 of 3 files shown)


cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -112,20 +114,30 @@ void setTimer1Callback(uint32_t (*fn)()) {
...
+  if (runTimeCycles && !wave->expiryCycle) {
     wave->expiryCycle = 1; // expiryCycle==0 means no timeout, so avoid setting it
   }
 
   uint32_t mask = 1<<pin;
-  if (!(waveformEnabled & mask)) {
+  if (waveformEnabled & mask) {
+    esp8266::InterruptLock il; // Make the variable sets atomic
+    wave->gotoTimeLowCycles = timeLowCycles;
+    wave->gotoTimeHighCycles = timeHighCycles;
```
mcspr, line 135: If waveformEnabled is set from timer1 cb, should the lock also cover condition check itself?
(i.e. are we safe while doing & operation?)

devyte: That's a good question. Usually in a case like this it's pass/fail (i. e. bool result) so it doesn't matter, because you can't corrupt a bool. Worst case is you miss the condition in this pass, which doesn't matter if the timing is so tight, and it'll be caught correctly in the next pass. 
However, on our esp I believe the operation is done in sw, so I guess the answer depends on whether the sw operation is guaranteed to be stable or not when faced with the value changing in the middle of it.

earlephilhower: @mcspr did find a race condition.  You could do the `if (a&mask)`, take an interrupt stopping the specific pin, and then return to set the next cycle's numbers...which will be ignored because the timeout hit.

It's a corner-case, though, and only when a Tone with a period is specified but is overwritten before it ends.  I should also update the remaining run length here, too.

mcspr: Ah, sorry. But does lock work at all? Looking at initTimer(), timer1 is set as NMI (== meaning we can't stop / hide from it?)

re previous question, since I have a very basic understanding of it. In case of normal isr, [xtensa ISA doc that i found (ref 3.5.4.1, 3.5.4.2 about fetch, 4.4.5.4 about checkInterrupts)](https://0x04.net/~mwk/doc/xtensa.pdf) describes interrupt trigger as beginning of instruction cycle. Does it mean that literal asm lines from `objdump -S` / `g++ -S` location can be used to point to possible 'stop' locations?

dok-net: I was going to complain about the same fundamental issue here, @mcspr, there's multiple mention of NMI in the waveform code, and given that
```
for (int o = 0; o < 8; ++o) {
	tone(D5, o % 2 ? 440 * 4 : 440 / 2);
	yield();
	{
		esp8266::InterruptLock il;
		delayMicroseconds(500000);
	}
}
noTone(D5);
```
produces sounds just fine, I think this PR is broken. Besides, I don't get the point of duplicating my (and @Tech-TX's) extensive work in PR #7022 :-(

devyte: @dok-net the point of this pr is to provide an immediate fix on top of the current code, which passes @Tech-TX 's checks. This PR shows an almost perfect graph at 40 KHz, with only the extreme high and low having plateaus. @Tech-TX could you please add your linearity graph capture to the PR's main thread?
This is meant to get an interim fix done in time for 2.7.0. It is not meant to replace #7022.

dok-net:
Master, 40kHz PWM:
MASTER_40_20200406_190832
PR7022, 40kHz PWM:
PR7022_40_20200406_190238
PR7022 reverted to analogWrite() µs resolution instead of CPU cycles, 40kHz PWM:
PR7022-1_40_20200406_191256
PR7022, 40kHz PWM, "color temperature", red is most frequent, blue least frequent measurement:
PR7022_40_20200406_194215
PR7022, 20kHz PWM, "color temperature", red is most frequent, blue least frequent measurement:
PR7022_20_20200406_194823
PR7192, 40kHz PWM:
PR7192_40_20200406_192343
PR7192, 40kHz PWM, "color temperature", red is most frequent, blue least frequent measurement:
PR7192_40_20200406_193411
PR7192, 20kHz PWM, "color temperature", red is most frequent, blue least frequent measurement:
PR7192_20_20200406_195951

I find PR7022 has a more narrow red line versus PR7192, which should indicate that overall the error distribution is better for PR7022.

cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -256,23 +281,26 @@ static ICACHE_RAM_ATTR void timer1Interrupt() {
...
+          state.waveformState ^= mask;
+          if (state.waveformState & mask) {
             if (i == 16) {
               GP16O |= 1; // GPIO16 write slow as it's RMW
             } else {
               SetGPIO(mask);
             }
-            wave->nextServiceCycle = now + wave->nextTimeHighCycles;
-            nextEventCycles = min_u32(nextEventCycles, wave->nextTimeHighCycles);
+            wave->nextServiceCycle = now + wave->timeHighCycles;
```
devyte, line 291: I think this is causing a drift in the waveform, because it doesn't consider the time elapsed between the timer firing and the measurement of now (assuming a single waveform).
Instead, consider:
´´´cpp
wave->nextServiceCycle += wave->timeHighCycles;
´´´
which comes from #7057, as pointed out by @dok-net.

cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -256,23 +281,26 @@ static ICACHE_RAM_ATTR void timer1Interrupt() {
...
+            nextEventCycles = min_u32(nextEventCycles, wave->timeHighCycles);
           } else {
             if (i == 16) {
               GP16O &= ~1; // GPIO16 write slow as it's RMW
             } else {
               ClearGPIO(mask);
             }
-            wave->nextServiceCycle = now + wave->nextTimeLowCycles;
-            nextEventCycles = min_u32(nextEventCycles, wave->nextTimeLowCycles);
+            wave->nextServiceCycle = now + wave->timeLowCycles;
```
devyte, line 299: Same as previosu comment, consider:
´´´cpp
wave->nextServiceCycle += wave->timeHighCycles;
´´´

devyte:
To be clear: this PR is not expected to fix all recent issues with PWM. This PR is only meant to fix the most immediate and urgent issues while #7022 gets sorted out and optimized to match or improve on current performance and behavior. When #7022 is ready, and once everyone is in agreement about it, it will supersede the current implementation and then it will be merged.

earlephilhower:
Closing in favor of #7231 which directly addresses the issue of PWM alignment.

Issue #7195, opened 2020-04-06
Possible memory leak in MD5Builder


lrodorigo:
Hi all,
I noticed a possible memory leak in MD5 Builder `addStream` member function.

https://github.com/esp8266/Arduino/blob/eea9999dc5eaf464a432f77d5b65269f9baf198d/cores/esp8266/MD5Builder.cpp#L58

As you can see, if the `readBytes` fails, the function returns without releasing the allocated buffer.

earlephilhower:
@lrodorigo Looks like a potential leak to me!  Want to do a PR for it and add the `free` before the `return false`?

PR #7197, opened 2020-04-07
Fix for MD5 leak bug, issue #7195

lrodorigo:
Added the `free` before the return.
I was wondering if a more "C++ish" resource handling could be fine for you (e.g. buffer as unique_ptr).

### DIFF
--- a/cores/esp8266/MD5Builder.cpp
+++ b/cores/esp8266/MD5Builder.cpp
@@ -55,6 +55,7 @@ bool MD5Builder::addStream(Stream & stream, const size_t maxLen){
         // read data and check if we got something
         int numBytesRead = stream.readBytes(buf, readBytes);
         if(numBytesRead< 1) {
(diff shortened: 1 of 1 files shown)


devyte:
@lrodorigo I wouldn't mind a unique_ptr solution at all, at the very least it would be one small step towards exception friendly code.

devyte:
@lrodorigo ping? will you be reworking this with a smart ptr?

lrodorigo:
Sorry, I read in hurry your message and I misunderstood your message.

I can try to rework for a smart_ptr, no-except code. I can start to work on it after easter holidays.

d-a-v:
> I can try to rework for a smart_ptr, no-except code. I can start to work on it after easter holidays.

That will be nice !
We are going to merge this fix now because it solves a bug,
then when you're ready you can post another pull request using `unique_ptr<>`

PR #7204, opened 2020-04-09
Fix espota completion success/fail check

### LINKED ISSUE #7162: OTA Unsigned binary after first one is signed, espota.py reports [INFO]: Result: OK but update fails on esp8266
I have signed a binary and uploaded via OTA - so next OTA updates should be signed.
Then tried to see if the esp8266 would accept an unsigned binary - it's expected that it should not accept.
So I uploaded via OTA an unsigned binary, and espota.py reported  **[INFO]: Result: OK**
But it actually failed on the esp8266 side with **Error[4]: End Failed**
First: why is espota reporting OK if it failed?
And second, shouldn't the error on the esp side be UPDATE_ERROR_SIGN ?

- Hardware:				ESP-12e
- Core Version:			latest git hash
- Development Env:		Arduino IDE && Platformio
- Operating System:		Windows

I have used ArduinoOTA sketch for this tests.

earlephilhower:
The OTA script was not reporting the actual reported upload status from
the ESP8266, and instead always printed "Result: OK" no matter what
happened.

Now check for ERROR or OK in final message (and ensure the message is
not accidentally merged with the final byte count) and report properly.

Fixes #7162

### DIFF
--- a/tools/espota.py
+++ b/tools/espota.py
@@ -132,7 +132,7 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
         sys.stderr.write('FAIL\n')
         logging.error('%s', data)
         sock2.close()
-        sys.exit(1);
+        sys.exit(1)
         return 1
       sys.stderr.write('OK\n')
     else:
@@ -172,7 +172,7 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
         connection.sendall(chunk)
         if connection.recv(32).decode().find('O') >= 0:
           # connection will receive only digits or 'OK'
-          received_ok = True;
+          received_ok = True
       except:
         sys.stderr.write('\n')
         logging.error('Error Uploading')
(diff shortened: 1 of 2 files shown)


tools/espota.py
```diff
@@ -188,19 +188,24 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
...
       f.close()
       sock.close()
-      if (data != "OK"):
-        sys.stderr.write('\n')
-        logging.error('%s', data)
-        return 1;
-      return 0
+      if received_ok:
+        return 0
+      else:
```
devyte, line 207: I'm not aware of py best practices, but from C/C++ experiencie it's best to remove the else and leave just the return as is done in the current code.

tools/espota.py
```diff
@@ -188,19 +188,24 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
...
-      while not received_ok:
-        if connection.recv(32).decode().find('O') >= 0:
-          # connection will receive only digits or 'OK'
-          received_ok = True;
-      logging.info('Result: OK')
+      received_ok = False
+      received_error = False
+      while not (received_ok or received_error):
+        reply = connection.recv(64).decode()
+        if reply.find('E') >= 0:
```
devyte, line 195: How about a comment or two about what this if else means?

tools/espota.py
earlephilhower: I can't find anything in particular in PEP 8 (the Python style guide) about this.  Not a big deal, will change in a PR.

lorol:
Hi, on Arduino IDE 1.8.12 Win64 after this commit the **network port** of a sketch w/ usual OTA is there after an upload but later (on next start of the IDE) it disappears. Please verify.

devyte:
Please open a new issue and follow the issue template instructions. Please be specific about how to reproduce.

lorol:
Sorry, the problem is not here. It is a side effect on my setup after this change:
https://github.com/esp8266/Arduino/issues/7213

I have 2 x active network interfaces on a Win10 PC. 
Not sure if this needs to be corrected but it is annoying. Yes I will open an issue if needed. Thank you!

PR #7207, opened 2020-04-11
Fix TCP send to IPv6 link local addresses

d-a-v:
credit: Simon Arlott @nomis - similar to #6541 for TCP


nomis:
I don't think the API should be changed to remove `const`.

d-a-v:
That's right. Would you like to make a fixing PR ?

nomis:
No, it's your API change, you can fix it...

Right now I'm getting enough extra notifications from GitHub every time someone merges this commit into their fork.

PR #7208, opened 2020-04-11
MD5 Builder: Rework for unqiue_ptr instead of malloc for internal dynamic buffers.

lrodorigo:
In order to guarantee exception-safety I used `unique_ptr` to `uint8_t[]`, with `std::nothrow` decorator.
I cannot mark the functions as no-except because I don't know if the called functions (especially the Stream implementations) are no-except.
 
I also fixed the const-correctness of const member functions, such as toString() and getters. 

I executed the host tests, I still need to run tests on the ESP8266, but this is the first time that I contribute with you, so I have to study how to setup and run your tests suite on my device.

### DIFF
--- a/cores/esp8266/MD5Builder.cpp
+++ b/cores/esp8266/MD5Builder.cpp
@@ -1,10 +1,11 @@
 #include <Arduino.h>
 #include <MD5Builder.h>
+#include <memory>
 
-uint8_t hex_char_to_byte(uint8_t c){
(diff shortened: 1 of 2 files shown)


earlephilhower:
Thanks, @lrodorigo !

lrodorigo:
@earlephilhower I'm so happy to cooperate with you, in this important project. If you think that I'm appropriate, please let me know if I can cooperate with you in a more stable way and what I can do now.

Regards

devyte:
@lrodorigo if you would like to help move the core forward, please msg me in our gitter channel gitter.im/esp8266/.

PR #7216, opened 2020-04-15
mDNS: protect MDNSResponder::queryService against misuse

### LINKED ISSUE #7215: Exception (28) on MDNSResponder::queryService
MDNSResponder::queryService crashes esp8266 with Exception (28) in case it is called before begin has been called

d-a-v:
fixes #7215

### DIFF
--- a/libraries/ESP8266mDNS/src/LEAmDNS.cpp
+++ b/libraries/ESP8266mDNS/src/LEAmDNS.cpp
@@ -767,6 +767,10 @@ uint32_t MDNSResponder::queryService(const char* p_pcService,
                                      const char* p_pcProtocol,
                                      const uint16_t p_u16Timeout /*= MDNS_QUERYSERVICES_WAIT_TIME*/)
 {
+    if (0 == m_pUDPContext)
+        // safeguard against misuse
+        return 0;
+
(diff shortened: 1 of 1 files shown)


libraries/ESP8266mDNS/src/LEAmDNS.cpp
```diff
@@ -767,6 +767,10 @@ uint32_t MDNSResponder::queryService(const char* p_pcService,
                                      const char* p_pcProtocol,
                                      const uint16_t p_u16Timeout /*= MDNS_QUERYSERVICES_WAIT_TIME*/)
 {
+    if (0 == m_pUDPContext)
```
devyte, line 770: Isn't this a ptr? check against nullptr?

d-a-v: Yes. For coherency I sticked to this file's style.
https://github.com/esp8266/Arduino/blob/e1789ddf0c73ff00cdca104e420ead5bc8bc016f/libraries/ESP8266mDNS/src/LEAmDNS.cpp#L101

I was afraid to modify too much for when LEAmDNS2 is coming to avoid merge complication.

PR #7218, opened 2020-04-17
Pickup some loose ends from PR #7211

dok-net:
Future-proofing #7122, newly introduced public `startWaveformCycles()` is more aptly named `startWaveformClockCycles()` (like in rest of core API for this type of use).
Fix/clarify comments.

Fix redundancies in Tone, end Tone waveform on exact period limit for proper sound.
Fix redundancies in wiring_pwm.
Extend Servo to map in-use pins, Tone already has this.

### DIFF
--- a/cores/esp8266/Tone.cpp
+++ b/cores/esp8266/Tone.cpp
@@ -22,14 +22,14 @@
 */
 
 #include "Arduino.h"
-#include "user_interface.h"
 #include "core_esp8266_waveform.h"
+#include "user_interface.h"
 
 // Which pins have a tone running on them?
 static uint32_t _toneMap = 0;
 
 
-static void _startTone(uint8_t _pin, uint32_t high, uint32_t low, unsigned long duration) {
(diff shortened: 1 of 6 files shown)


cores/esp8266/Tone.cpp
```diff
@@ -49,7 +52,7 @@ void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) {
   if (frequency == 0) {
     noTone(_pin);
   } else {
-    uint32_t period = (1000000L * system_get_cpu_freq()) / frequency;
+    uint32_t period = microsecondsToClockCycles(1000000UL) / frequency;
```
devyte, line 55: I'm not sure whether this change is ok. The microsecondsToClockCycles() define uses the F_CPU macro, while system_get_cpu_freq() is the SDK api call. 
I _think_ it is ok because these functions aren't supposed to be called from an ISR. Also, if built for 160MHz, execution will be done at that speed something like 95% of the time.

PR #7219, opened 2020-04-17
Add isRunning getter to check is ::begin already called

### LINKED ISSUE #7214: No way to check, MDNSResponder::begin has been called or not
There's currently no way to check, is MDNSResponder::begin already called. There shall be a method that returns true if begin has been called and false if not (or close/end has been called). There shall be a bool member field default to false, set to true if begin succeeded and back to false when close/end succedded, and some method like bool isBegan(void) to check it

BbIKTOP:
Fixes #7214

### DIFF
--- a/libraries/ESP8266mDNS/src/LEAmDNS.h
+++ b/libraries/ESP8266mDNS/src/LEAmDNS.h
@@ -196,7 +196,7 @@ public:
 
     bool isRunning(void)
     {
-        return(m_pUDPContext!=0);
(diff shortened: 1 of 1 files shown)


earlephilhower:
@BbIKTOP you have an astyle error.  You can fix them by running `tests/restyle-examples-only.sh` or manually by just adding the missing space in this case.

BbIKTOP:
Yes, it's a new computer, just found that astyle was not installed. Fixed.

PR #7222, opened 2020-04-17
CI: fix style checking

d-a-v:
- (some) core files are astyled
- ci was only checking `libraries/`

=> now everything is checked


earlephilhower:
Missed it by *that* much.  😆

PR #7228, opened 2020-04-19
Create new method connectAsync

ruggi99:
Created a new method connectAsync to connect asynchronously to a remote server.
Function connectAsync takes one more parameter to understand if connection is still in progress or has failed. If connection is successful, waiting flag is set to 0. If connection isn't successful, waiting flag is set to 1 if connection has not failed and therefore there's a possibility that in future will be successful.
Added an example showing how connectAsync works
Based on dead PR #5595


ruggi99:
Ping

libraries/ESP8266WiFi/examples/ConnectAsync/ConnectAsync.ino
```diff
@@ -0,0 +1,71 @@
+/*
+    This sketch tries to connect asincronously to a remote server.
```
earlephilhower, line 2: "asynchronously"

libraries/ESP8266WiFi/examples/ConnectAsync/ConnectAsync.ino
```diff
@@ -0,0 +1,71 @@
...
+  Serial.println(WiFi.localIP());
+}
+
+void loop() {
+  // Use WiFiClient class to create TCP connections
+  if (client.connected()) {
+    //do some stuff
+  } else {
+    // Try to connect
+    if ((millis() - lastTry) > 5000) {
```
earlephilhower, line 55: This should be a `PolledTimeout`

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -718,6 +718,48 @@ void wifi_dns_found_callback(const char *name, const ip_addr_t *ipaddr, void *ca
     esp_schedule(); // break delay in hostByName
 }
 
+void wifi_dns_found_callback_async(const char* name, const ip_addr_t* ipaddr, void* callback_arg);
```
earlephilhower, line 721: `static` as it should not exist out of this file

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -718,6 +718,48 @@ void wifi_dns_found_callback(const char *name, const ip_addr_t *ipaddr, void *ca
...
+        if (waiting != NULL) {
+            *waiting = 1;
+        }
+    } else { // ERR_TIMEOUT
+        DEBUG_WIFI_GENERIC("[hostByName] DNS search failed. Host %s\n", aHostname);
+    }
+    return 0;
+}
+
+void wifi_dns_found_callback_async(const char* name, const ip_addr_t* ipaddr, void* callback_arg) {
```
earlephilhower, line 757: `static` as mentioned above.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -121,6 +121,7 @@ class ESP8266WiFiGenericClass {
     public:
         int hostByName(const char* aHostname, IPAddress& aResult);
         int hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms);
+        int hostByNameAsync(const char* aHostname, IPAddress& aResult, uint8_t* waiting = NULL);
```
earlephilhower, line 124: `waiting` should be a `bool` since it only has `0` or `1` states.

libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -137,6 +137,22 @@ int WiFiClient::connect(const String& host, uint16_t port)
     return connect(host.c_str(), port);
 }
 
+int WiFiClient::connectAsync(const char* host, uint16_t port, uint8_t* waiting) // Non blocking
```
earlephilhower, line 140: Should be a `bool` and return `false` or `true`

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -718,6 +718,48 @@ void wifi_dns_found_callback(const char *name, const ip_addr_t *ipaddr, void *ca
...
+ * @param aResult       IPAddress structure to store the returned IP address
+ * @param waiting       waiting flag if DNS search is still in progress
+ * @return 1 if aHostname was successfully converted to an IP address,
+ *         else 0
+ */
+
+int ESP8266WiFiGenericClass::hostByNameAsync(const char* aHostname, IPAddress& aResult, uint8_t* waiting) {
+    ip_addr_t addr;
+
+    if (waiting != NULL) {
```
earlephilhower, line 735: `NULL`->`nullptr` everywhere, please.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -718,6 +718,48 @@ void wifi_dns_found_callback(const char *name, const ip_addr_t *ipaddr, void *ca
...
 }
 
+void wifi_dns_found_callback_async(const char* name, const ip_addr_t* ipaddr, void* callback_arg);
+
+/**
+ * Resolves the given hostname to an IP address asynchronously.
+ * @param aHostname     Name to be resolved
+ * @param aResult       IPAddress structure to store the returned IP address
+ * @param waiting       waiting flag if DNS search is still in progress
+ * @return 1 if aHostname was successfully converted to an IP address,
```
earlephilhower, line 728: `if a*space* hostname...` and the alignment seems kind of funky, too.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -235,6 +235,24 @@ int WiFiClientSecure::connect(const String& host, uint16_t port) {
   return connect(host.c_str(), port);
 }
 
+int WiFiClientSecure::connectAsync(const char* name, uint16_t port, uint8_t* waiting) {
+  if (waiting != NULL) *waiting = 0;
+  if (connected()) return 1;
+
+  IPAddress remote_addr;
+
+  if (!WiFi.hostByNameAsync(name, remote_addr, waiting)) {
```
earlephilhower, line 244: I don't think this fcn can work.  If `hostByNameAsync` is IN_PROGRESS, then it returns 0.  And so this routine will return failure if the async part actually takes effect.

libraries/ESP8266WiFi/examples/ConnectAsync/ConnectAsync.ino
ruggi99: Done in dd24fb4539ccc02105a420f215da8c8af46ddda9

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
ruggi99: Done in df3075cd47a18f96de540cf2dc5a994fbcbf5a84

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
ruggi99: Done in df3075cd47a18f96de540cf2dc5a994fbcbf5a84

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
ruggi99: Done in 37b35d78c2acc701b3755640674ed068c193188a

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
ruggi99: Done in 0093cca821e180a92ba5e6e474faa3f5c5daef7e

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
ruggi99: Done in 16d52717a6603ee422cb61c58741f518de233d19. Should work

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
ruggi99: It's the same description as ESP8266WiFiGenericClass::hostByName.
I just copy-pasted and added a line for waiting parameter.

libraries/ESP8266WiFi/src/WiFiClient.cpp
ruggi99: The function has the same formal parameters and return type as WiFiClient::connect.
I just added the waiting parameter. Since they are similar I've not changed it for consistency.

libraries/ESP8266WiFi/examples/ConnectAsync/ConnectAsync.ino
ruggi99: Done in 06bac8e44de87c6ce548d4d79eb1a1249f333f11. Don't know if oneShot type is better than periodic in PolledTimeout here.

ruggi99:
@devyte I agree with you. The function is not completely async, but only in part.

> Finally, the same behavior as proposed can be achieved by just implementing an async or pollable dns lookup method, and then those users that need it can make use of it.

Would you mind explaining me this paradigm better? Are you saying that I can leave hostByNameAsync function and delete connectAsync, as it's not completely asynchronous as you may think?

ruggi99:
While testing another way to get async mode I found something wrong that sounds me like a bug.
If I call hostByName with a very low timeout (even 0), it always returns 1. The wrong line is this in my opinion:
https://github.com/esp8266/Arduino/blob/5b3d290de8df597a98f1c13fc216754040ab1a7b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp#L616
It should be INADDR_ANY.
Because when this line is reached:
https://github.com/esp8266/Arduino/blob/5b3d290de8df597a98f1c13fc216754040ab1a7b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp#L638
isSet() returns true when aResult is INADDR_NONE and false when INADDR_ANY.
If the callback doesn't fire, hostByName returns 1 instead of 0.

ruggi99:
Another way to get async mode, is to use a timeout of 0 (or very low anyway).
Works perfectly with hostByName (the callback is ignored) but doesn't work with connect function.
There are problems either in WiFiClient and ClientContext class.
- Each time `WiFiClient::connect` is called a new ClientContext (and a new tcp buffer) is created, deleting the existing one.
- Whenever `ClientContext::connect` fails (`tcp_connect` returns false), the close function is called, deleting the tcp buffer.
- Most of users do call disconnect function before trying a new connection, which calls `WiFiClient::stop`, which deletes `_client`

ruggi99:
https://github.com/esp8266/Arduino/blob/5b3d290de8df597a98f1c13fc216754040ab1a7b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L223
I think should be:
`WiFi.hostByName(name, remote_addr, _timeout)`

ruggi99:
@earlephilhower what do you think of the above messages and corrections I made?

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -235,6 +235,28 @@ int WiFiClientSecure::connect(const String& host, uint16_t port) {
   return connect(host.c_str(), port);
 }
 
+int WiFiClientSecure::connectAsync(const char* name, uint16_t port, bool* waiting) {
+  if (waiting != nullptr) *waiting = false;
+  if (connected()) return 1;
```
earlephilhower, line 240: A `connect()` on a previously connected client should drop the current connection and open a new one.  This won't, it will return success.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -235,6 +235,28 @@ int WiFiClientSecure::connect(const String& host, uint16_t port) {
   return connect(host.c_str(), port);
 }
 
+int WiFiClientSecure::connectAsync(const char* name, uint16_t port, bool* waiting) {
+  if (waiting != nullptr) *waiting = false;
```
earlephilhower, line 239: Singe-line if

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -121,6 +121,7 @@ class ESP8266WiFiGenericClass {
     public:
         int hostByName(const char* aHostname, IPAddress& aResult);
         int hostByName(const char* aHostname, IPAddress& aResult, uint32_t timeout_ms);
+        int hostByNameAsync(const char* aHostname, IPAddress& aResult, bool* waiting = nullptr);
```
earlephilhower, line 124: Just in general, if `waiting` is `nullptr`, then how would an app actually know that it's done?  I don't think waiting makes sense being optional here.

libraries/ESP8266WiFi/examples/ConnectAsync/ConnectAsync.ino
```diff
@@ -0,0 +1,75 @@
...
+#endif
+
+const char* ssid     = STASSID;
+const char* password = STAPSK;
+
+const char* host = "djxmmx.net";
+const uint16_t port = 17;
+
+bool isDNSSearchInProgress = false;
+unsigned long lastTry = 0;
```
earlephilhower, line 20: Unused variable

libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -137,6 +137,22 @@ int WiFiClient::connect(const String& host, uint16_t port)
     return connect(host.c_str(), port);
 }
 
+int WiFiClient::connectAsync(const char* host, uint16_t port, bool* waiting) // Non blocking
+{
+    if (waiting != nullptr)
+        *waiting = false;
+    if (connected())
```
earlephilhower, line 144: If a Client is connected(), another `connect` call will disconnect it and open a new one.  This doesn't.

libraries/ESP8266WiFi/src/WiFiClient.h
```diff
@@ -55,7 +55,11 @@ class WiFiClient : public Client, public SList<WiFiClient> {
   uint8_t status();
   virtual int connect(IPAddress ip, uint16_t port) override;
   virtual int connect(const char *host, uint16_t port) override;
+  virtual int connectAsync(const char* host, uint16_t port, bool* waiting = nullptr);
```
earlephilhower, line 58: Same comments about `waiting`.  I don't think `nullptr` makes any sense because the app would never know if it was a failure or just that it needs to wait for the answer.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -235,6 +235,28 @@ int WiFiClientSecure::connect(const String& host, uint16_t port) {
...
+  if (connected()) return 1;
+
+  IPAddress remote_addr;
+
+  if (!WiFi.hostByNameAsync(name, remote_addr, waiting)) {
+    if (waiting != nullptr and *waiting == true) {
```
earlephilhower, line 245: For consistency, `and` -> `&&`, because it's that way everywhere else in the core.

ruggi99:
Hi @earlephilhower, thanks for your suggestions.
I've changed my mind since your first message because I agree with you and I think this is not the best approach to making an async connection. Please review the 4 comments I left 2 days ago. Maybe I should open a question to discuss it?

ruggi99:
Created a new PR on my fork with a new approach to achieve the same result

ruggi99/Arduino#10

devyte:
This is not a breaking change, removintg 3.0.0 milestone.

Issue #7202, opened 2020-04-09
mklittlefs support for 32 bit Linux is missing in version 2.6.0 and later


andrews999:
### Basic Infos

- [ y] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [y ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [y ] I have tested that the issue is present in current master branch (aka latest git).
- [y ] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [y ] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12|ESP-01|ESP-07|ESP8285 device|other]
- Core Version:			[2.6.3]
- Development Env:		[Arduino IDE]
- Operating System:		[Linux 32 bit]

### Settings in IDE

- Module:  			[Generic ESP8266 Module|Wemos D1 mini r2|Nodemcu|other]
- Flash Mode:			[qio|dio|other]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v1.4|v2 Lower Memory|Higher Bandwidth]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz|160MHz]
- Upload Using:		[OTA|SERIAL]
- Upload Speed:   [115200|other] (serial upload only)

### Problem Description

When mklittlefs was introduced in version 2.6.0 the lines to support 32 bit Linux were missing from package_esp8266com_index.json. They can be manually added in a local file as the files exist in https://github.com/earlephilhower/esp-quick-toolchain/releases/download/2.5.0-4/i686-linux-gnu.mklittlefs-7f77f2b.1563313032.tar.gz.
Can the build be changed to include support for 32 bit Linux?

d-a-v:
This has been fixed since 2.6.3, you need to wait for release 2.7.0 or use the [beta (0.0.1)](https://d-a-v.github.io) version.

andrews999:
Thanks for your help.

PR #7231, opened 2020-04-19
Re-implement PWM generator logic

earlephilhower:
Add special-purpose PWM logic to preserve alignment of PWM signals for
things like RGB LEDs.

Keep a sorted list of GPIO changes in memory.  At time 0 of the PWM
cycle, set all pins to high.  As time progresses bring down the
additional pins as their duty cycle runs out.  This way all PWM signals
are time aligned by construction.

This also reduces the number of PWM interrupts by up to 50%.  Before,
both the rising and falling edge of a PWM pin required an interrupt (and
could shift arround accordingly).  Now, a single IRQ sets all PWM rising
edges (so 1 no matter how many PWM pins) and individual interrupts
generate the falling edges.

The code favors duty cycle accuracy over PWM period accuracy (since PWM
is simulating an analog voltage it's the %age of time high that's the
critical factor in most apps, not the refresh rate).  Measurements give
it about 35% less total error over full range at 20khz than master.

@me-no-dev used something very similar in the original PWM generator.


dok-net:
Here is an unsolicited test, for comparing against PR #7022, https://github.com/esp8266/Arduino/pull/7022#issuecomment-616221739:
PR7231_20200419_231458

earlephilhower:
Thanks!  Can you give test parameters for the above graph?

earlephilhower:
I changed the test definition and reran and I can *kind of* repro that hump, but it's not looking like your output.  Smells like its related to the handoff between busy-wait and IRQ handled updates.

image

Test sketch:
https://gist.github.com/earlephilhower/26f0f5695e2a3f401a4173c00c9926ed

Test Master (yes, it's in PowerShell for now)
https://gist.github.com/earlephilhower/8ab3a22e61aa584a860162922fbb572f
[newpwm25k2tone.zip](https://github.com/esp8266/Arduino/files/4500422/newpwm25k2tone.zip)

s-hadinger:
As stated, I confirm there is no more visible flickering with leds in Tasmota. 

Any chance this PR is integrated in v2.7?

earlephilhower:
With multiple (different) PWMs there's the non-linear bumps you see on the graph in https://github.com/esp8266/Arduino/pull/7231#issuecomment-616235682 (kind of a worst-case test where we have the PWM periods going in opposite directions).  I wrote it literally during a bout of insomnia yesterday AM and haven't optimized it at all.

There's also #7022 which @dok-net has worked hard on and which he seems to get better performance for his app from.  It does show better linearity than this one, on the little logic analyzer I'm using, looked pretty stable, too, but I've not tried 3 channels or other stuff yet.

Net result, it's very complicated.  You can always merge this PR into your own master before building if needed, while we try and figure out the best solution...

earlephilhower:
I actually had 6 PWMs running in the prior graph, not 2 (bug just fixed).

After the latest push, here is what I get with 2 opposing 25khz PWMs going in opposing directions:

image

Much nicer.  Error is <1% for *everywhere* except the hump at 160-230.

dok-net:
@earlephilhower I found that aggregating the GPIO changes and writing to the two set and clear registers just once each, had overall worse performance, due to the added code for collecting state, I guess. While it does give perfect phase sync, the performance impact was unacceptable. I've reverted this now in PR #7202.
About your graph, it's unfortunate that it hides the jitter completely...
Here's 40kHz, 80MHz, single GPIO:
PR7231_20200421_004458
And this is 25kHz, 160MHz (which @devyte says not to use for proving anything), two GPIO PWM in opposing increments:
PR7231_2_20200421_005316
I'm posting the same test's results for PR#7022 here:

dok-net:
@earlephilhower We additionally need performance figures - I take I you were speaking from experience, not hard numbers. Can you please look into how you'll get well defined 0V and Vmax bands? I derive from the images that those are maintained only for exactly 0 and 1023, if at all.

earlephilhower:
Thanks for the screenshots!

Jitter's actually recorded and in the CSV/XLS.  I just didn't graph it to keep things simple.  With the jitter, you can't see the average value which it sawtooth in #7022, too, so I don't think there's a good single answer.

Here's what he HW recorded.  I took the std.dev of the high period (i.e. duty cycle jitter) and divided by the average period reported over the ~100 samples it examined.
image

I can measure anything we need if the analyzer can record it and I can teach PS or Python how to look for it, no sweat!

For 0 and 100%, I'm not sure I understand you.  Aren't those special-cased as digitalWrite(0/1)?  That's what's in master now and in this PR (although, as you noticed, there was a bug and that analogWrite(100%)->digitalWrite(1) mapping was busted...it's fixed with last push)

earlephilhower:
Oh, and it's very cool that your performance shows the same dip at about the same time (the bump on the 1 pin has a corresponding dip on the other pin, so I guess we're just looking at different pins of the pair)!  Independent verification is always best.

dok-net:
@earlephilhower I don't know what commit 160435d was exactly, but since I've rebased and squashed a serious bug fix and a commit (GPOS and GPOC use) that was so bad in performance that wanted to get rid of that completely, it just may be that you are still testing that badly badly broken revision!!! Please rerun and probably take down your graphs with that revision... I'm sorry.

earlephilhower:
@dok-net "106-435d" is "160mhz, commit 435d."  It's the last force-push.  Do you have unpushed commits, maybe?

dok-net:
@earlephilhower 0 and 1023, as digitalWrite(…) etc. in master, is one of the things that got started on #7022, because that totally loses phase. Let me think, try switching off servos perfectly with your code such that they don't ever sweep into nirvana on detach. My hope is still to use waveform for communications signal generation, like software serial. For that, exactly level on exit from a wave defined by precise runtime is a must. You see, the features I've built in are what I'm fighting for here, your PR wouldn't cut it for those (imaged) uses (just yet). My worry is that this competion, where you keep merging small portions, are going to drive me nuts rebasing or merging my code - probably I'll have to keep forcing my stuff over everything, which ends up badly,

dok-net:
@earlephilhower OMG, 160 "-" 435d - I'm sorry, your graphics is slightly blurry and small, I thought it was 160435d :-) :-) :-) So, 435d5d8fd750c6c5631244ee0592b42fd24421bb is just right.

earlephilhower:
Ah, I see.  For master and this PR, then `analogWrite(max/min)` is a no-op.

So I think I get:
* Master has *no* phase guarantee between anything.
* This PR *will* maintain phase even if you turn off a PWM pin *if and only if* you have another PWM pin active
* #7022 will maintain phase *always*, even if no PWM is running(?)

dok-net:
#7022 maintains phase if the waveform generation running, but with either zero duty or 100% duty. Also, the runtime parameter exits on the exact level at that time, I am sure anymore, but master at least always switches back to LOW, which is not what I expect.

dok-net:
And, you haven't picked that up yet, while I have removed ns-precise phase alignment due to the overall performance issue with that, one can specify phase delay to another previously started wave, so you could generate three waves at 120° phase shift to each other, for instance :-)

earlephilhower:
The hump is caused by having a PWM falling edge just before the cycle restart, while having the other channel requesting a 1->0 transition just outside the busy-loop window of 10us.  So it gets an IRQ for channel B 0->1, then waits 2..8us for the next PWM full cycle 0->1, and ends up returning from interrupt and not scheduling another IRQ for 10us...hence the horizontal leg of the bump...
image

earlephilhower:
Latest average PWM results:
image

earlephilhower:
Single PWM at 40khz, 160mhz.  **This is different than the other graphs show here which are 2-pwm at 25khz moving in opposite directions**

image

earlephilhower:
2 opposing pwm @ 25khz pwm and  80MHz CPU.  

image

s-hadinger:
I tried latest commit, still good results. I will try later on a TECKIN bulb as it was reported #7054 created flicker on this bulb.

s-hadinger:
@earlephilhower For now I'm switching to #7231 instead of #7054 in Tasmota. I haven't seen any downside in this PR, and waiting for #7022 to be complete and solve the drop of brightness at wifi connection.

s-hadinger:
@earlephilhower For now I'm switching to #7231 instead of #7054 in Tasmota. I haven't seen any downside in this PR, and waiting for #7022 to be complete and solve the drop of brightness at wifi connection.

s-hadinger:
@earlephilhower For now I'm switching to #7231 instead of #7054 in Tasmota. I haven't seen any downside in this PR, and waiting for #7022 to be complete and solve the drop of brightness at wifi connection.

s-hadinger:
@earlephilhower For now I'm switching to #7231 instead of #7054 in Tasmota. I haven't seen any downside in this PR, and waiting for #7022 to be complete and solve the drop of brightness at wifi connection.

Tech-TX:
> @earlephilhower For now I'm switching to #7231 instead of #7054 in Tasmota. I haven't seen any downside in this PR, and waiting for #7022 to be complete and solve the drop of brightness at wifi connection.

I think that might happen with _any_ version of software-driven PWM (brightness change with WiFi) as WiFi has higher priority than PWM, and it can hog the interrupts for quite a long time.  I've seen my servos go completely crazy during an OTA upload.  I could easily be wrong here, but my read is that PWM can't interrupt the WiFi once WiFi has grabbed control of the interrupt; WiFi has to release the lock before PWM can resume.

Tech-TX:
Here's my version of Earle's two 25KHz plots with opposing channels, much the same as his:
7231 two channels opposing @ 25KHz

There's a bit of noise in my test bed that I haven't been able to walk down, but other than that we're pretty close to each other's results.

earlephilhower:
The automated test suite can now report on two signals, so here's the latest push's results on the usual 25khz,80MHz,2 opposing PWMs:
image

earlephilhower:
CPU slowdown for same test.  1.0=normal no PWM, 2.0 = computations take 2x as many seconds to finish, etc.

image

dok-net:
In order to mention it here, too: setting the BUILTIN_LED GPIO to 1022, it's quite unsteadying, flashing, with this PR right now. Using the same analogWrite API to set 40kHz mode, and updating two other channels continually in opposite direction.

earlephilhower:
Sorry, I'm not able to reproduce that.  40khz and 3 PWMs at 80MHz is gonna be dicey no matter what, but I did just run the following test and didn't notice any flashed on the LED (d1 mini):
````
sut = analyzer.ESP8266(port = 'COM6')
la = analyzer.LogicAnalyzer()
sut.reset()
idle = sut.perfTest()
freqs = [ 40000 ]
for f in freqs:
    sut.analogWriteFreq(f)
    for i in range(1, 1023, 1): # also tried step=10 and it was OK
        sut.analogWrite(1, i);
        sut.analogWrite(2, i-1);
        sut.analogWrite(4, 1022)
        csv = la.capture(pins = 'D1,D2', samples = 1200, rate = '24mhz')
        cycleCnt, avgPeriod, avgDuty, periods, highs, duties, stdP, stdD = la.decodePWM(csv, 1)
        running = sut.perfTest()
        print(running/idle)
````

**Comment by dok-net** [comment]
So now that's Python OMG - running where exactly? Don't reset the LED for every step, btw.

**Comment by dok-net** [comment]
Just beginning to wonder if integrating audio and/or servos into the tests alters the landscape. Basically anything that doesn't have the same period like running everything off analogWrite() has. Admittedly, I haven't checked if this PR can do different periods on different GPIOs.

**Comment by earlephilhower** [comment]
Here it is locking on to a PWM in an idle system at 80mhz:
image

Add a new (low frequency in this case) tone and the IRQ loop is a little longer because of the add'l pin it's checking, and you can see the lock re-acquire:
image

It's unrelated to PWM, but still kinda neat IMHO.

**Comment by earlephilhower** [comment]
> @earlephilhower please move on from the c preprocessor to c++ constexpr, specifically, my support for 80/160MHz in [6cb1699](https://github.com/earlephilhower/Arduino/commit/6cb16997d8cedacc0dc5b1acda9a3e868b7b4395), it's in master, I am using it too.

You're lucky I don't go back to K&R C.  😆

I'll be sure to go over it next commit!

**Comment by earlephilhower** [comment]
The individual optimization just pushed might apply to #7022 and master, too.  They've reduced IRAM usage by 336 bytes.

It's still larger than master by ~450 bytes, and ~90 bytes larger than #7240 

* ~50 bytes are due to the fixing of waveforms when WiFi connects and CPU speed changes under our feet
* ~50 bytes are due to the full-period copy over of waveforms (i.e. smoothly transition to next waveform on next full wvf period)
* ~ 250 bytes are due to PWM generation in the ISR and support code
* ~ 100 bytes are due to the frequency feedback loop on waveforms

Need to do full sweep, but test B and G look OK on it.

**Comment by earlephilhower** [comment]
Last couple reduce IRAM size by another ~50 bytes, so we're at just under 400b larger than master now.

**Comment by dok-net** [comment]
@earlephilhower If you're sure about that GPIO16 finding, adding global macros like the rest of them would be great idea.

File: cores/esp8266/core_esp8266_waveform.cpp [hunk 1]
```diff
@@ -259,24 +527,58 @@ static ICACHE_RAM_ATTR void timer1Interrupt() {
...
-          waveformState ^= mask;
-          if (waveformState & mask) {
+          uint32_t nextEdgeCycles;
+          uint32_t desired = 0;
+          uint32_t *timeToUpdate;
+          wvfState.waveformState ^= mask;
+          if (wvfState.waveformState & mask) {
             if (i == 16) {
-              GP16O |= 1; // GPIO16 write slow as it's RMW
+              GP16O = 1; // GPIO16 write slow as it's RMW
```
dok-net, line 536: With revision control, please never leave commented code or outdated comments in commits.

cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -259,24 +527,58 @@ static ICACHE_RAM_ATTR void timer1Interrupt() {
...
+            if (wave->lastEdge) {
+              desired = wave->desiredLowCycles;
+              timeToUpdate = &wave->timeLowCycles;
+            }
+#endif
+            nextEdgeCycles = wave->timeHighCycles;
           } else {
             if (i == 16) {
-              GP16O &= ~1; // GPIO16 write slow as it's RMW
+              GP16O = 0; // GPIO16 write slow as it's RMW
```
dok-net, line 549: No RMW now, remove the confusing comment, too.

s-hadinger:
As I tested LED with #7022, I also re-tested this PR. It's still super stable: no flickering in steady state, and absolutely no visible change during Wifi connection. I don't know how you did this magic but it's impressive.

I was running at 80MHz with a PWM frequency of 223Hz.

cores/esp8266/core_esp8266_waveform.cpp
dok-net: @earlephilhower The GPIO 16 macros in `esp8266_peri.h` are:
```
//GPIO 16 Control Registers
#define GP16O  ESP8266_REG(0x768)
#define GP16E  ESP8266_REG(0x774)
#define GP16I  ESP8266_REG(0x78C)

//GPIO 16 PIN Control Register
#define GP16C  ESP8266_REG(0x790)
#define GPC16  GP16C

//GPIO 16 PIN Function Register
#define GP16F  ESP8266_REG(0x7A0)
#define GPF16  GP16F

//GPIO 16 PIN Function Bits
#define GP16FFS0 0 //Function Select bit 0
#define GP16FFS1 1 //Function Select bit 1
#define GP16FPD  3 //Pulldown
#define GP16FSPD 5 //Sleep Pulldown
#define GP16FFS2 6 //Function Select bit 2
#define GP16FFS(f) (((f) & 0x03) | (((f) & 0x04) << 4))
```
Would you feel confident enough to apply the change to `core_esp8266_wiring_digital.cpp` `__digitalWrite()` as well? Otherwise, I'm not buying it just yet myself in #7022.

earlephilhower: `digitalWrite` change would be for another PR (but digitalWrite's not really timing critical like this code that's firing 80,000 per second).

I'm pretty confident that only 1 bit is implemented at that address.  Have you ever read anything but 0 for the upper bits in GP16O that makes you concerned?

earlephilhower:
> As I tested LED with #7022, I also re-tested this PR. It's still super stable: no flickering in steady state, and absolutely no visible change during Wifi connection. I don't know how you did this magic but it's impressive.

@s-hadinger can you try the latest push?  There should be no functional changes, but IRAM usage for PWM has gone down by ~60% vs. the version you were using.  If you do find something has broken with the code reduction, do please let me know!

s-hadinger:
Sure, will try tomorrow.

s-hadinger:
I tested latest commit `#e44171f`:

- ESP at 160MHz, PWM Freq at 223Hz: no visible artifact, clear difference between 1/1023 and 2/1023 (full 10 bits range).
- ESP at 160MHz, PWM Freq at 880Hz: no visible artifact, clear difference between 1/1023 and 2/1023 (full 10 bits range).

- ESP at 80MHz, PWM Freq at 223Hz: almost invisible blinks during Wifi connection (they are subtle, I may have missed them earlier), clear difference between 1/1023 and 2/1023 (full 10 bits range).
- ESP at 160MHz, PWM Freq at 880Hz: no visible artifact, clear difference between 1/1023 and 2/1023 (full 10 bits range).

It's all good to me, the visible artifact at Wifi connection is hard to see and way better than current implementation.

However, when I change the PMWFrequency, the PWMs stop completely. With previous versions, the change would only be taken after restart.

earlephilhower:
> However, when I change the PMWFrequency, the PWMs stop completely. With previous versions, the change would only be taken after restart.

Yes, I didn't think that was something that people used and removed the code that redid PWM values when things were changed.  Is that something you're doing in your code?  Or was this just a directed test?  It's easy to put it back in, but it does cost ~100 bytes of IROM for the logic.

s-hadinger:
Thanks, 80 bytes of IROM is reasonable indeed.

s-hadinger:
There is something wrong with https://github.com/esp8266/Arduino/pull/7231/commits/7fe9a2d7d8f3ecc1e82781d923d977edb87b55df

Now I can't turn off the LEDs any more.  When I set analogWrite to 0, the GPIO is set to HIGH instead of LOW.

earlephilhower:
> Now I can't turn off the LEDs any more. When I set analogWrite to 0, the GPIO is set to HIGH instead of LOW.

I swapped the LOW/HIGH values in the PR when removing a temp variable.  Sorry!

Fixed with latest push and verified on the LA.

s-hadinger:
Thanks, this looks good to me. We have reverted the default PWM frequency to 977Hz and even running at 80MHz, there are no blinks/flickers, and resolution is full 10 bits.

I'm updating Tasmota with this latest commit.

earlephilhower:
Updated to new PWMRANGE, but not yet tested with new core/GCC

BigMike71:
Hello, with the latest versions of tasmota (lite) 8.4.x I have strong visible flickering again with my LED lamps at low brightness!
are the changes in the current Tasmota already included?

s-hadinger:
@BigMike71 Tasmota currently uses code from #7022.

Jason2866:
@BigMike71  Tasmota is back to #7231  we had sometimes execptions when using PWMi with #7022  Using #7231 there is no flicker and no execption.

PR #7232, opened 2020-04-20
Turn off IRQ on stopWaveform after Tone timeout

### LINKED ISSUE #7230: stopWaveform() doesn't deinit the timer if waveform had limited runtime
```
[...]
startWaveform(GPIO, 100, 100, 200);
delay(200);
stopWaveform();
[…]
```
Due to code in master branch, core_esp8266_waveform.cpp, `stopWaveform(uint8_t pin)`:
```
  uint32_t mask = 1<<pin;
  if (!(waveformEnabled & mask)) {
    return false; // It's not running, nothing to do here
  }
```
the timer keeps running even when there isn't any other waveform or the callback attached.

Fixed in #7022

earlephilhower:
Fix #7230 (until 3.0 when this whole code will probably change).

Before, a timed out pin would clear its bit in `waveformEnabled`.  The
logic in `stopWaveform`.  If only one tone was running and timed out,
`stopWaveform` would check the `enabled` bitmask and see nothing active
and immediately return w/o cancelling the IRQ.  So, the IRQ would be
called every 1/100th of a second and return immediately when no work to
be done was detected.

Remove the check and always send in a `waveformToDisable` bit.  It's
save to disable an already disabled pin, so no logical consequences will
occur, and the final IRQ disable will be executed when appropriate.

### DIFF
--- a/cores/esp8266/core_esp8266_waveform.cpp
+++ b/cores/esp8266/core_esp8266_waveform.cpp
@@ -182,7 +182,7 @@ int ICACHE_RAM_ATTR stopWaveform(uint8_t pin) {
     timer1_write(microsecondsToClockCycles(10));
   }
   while (waveformToDisable) {
-    /* no-op */ // Can't delay() since stopWaveform may be called from an IRQ
+    /* no-op */ // Can't delay() because stopWaveform may be called from an IRQ
   }
   if (!waveformEnabled && !timer1CB) {
     deinitTimer();
--- a/cores/esp8266/core_esp8266_waveform.cpp
+++ b/cores/esp8266/core_esp8266_waveform.cpp
@@ -176,12 +176,6 @@ int ICACHE_RAM_ATTR stopWaveform(uint8_t pin) {
   if (!timerRunning) {
     return false;
   }
-  // If user sends in a pin >16 but <32, this will always point to a 0 bit
-  // If they send >=32, then the shift will result in 0 and it will also return false
-  uint32_t mask = 1<<pin;
-  if (!(waveformEnabled & mask)) {
-    return false; // It's not running, nothing to do here
-  }
   waveformToDisable |= mask;
   // Ensure timely service....
   if (T1L > microsecondsToClockCycles(10)) {
--- a/cores/esp8266/core_esp8266_waveform.cpp
+++ b/cores/esp8266/core_esp8266_waveform.cpp
@@ -176,13 +176,14 @@ int ICACHE_RAM_ATTR stopWaveform(uint8_t pin) {
   if (!timerRunning) {
     return false;
   }
+  uint32_t mask = 1<<pin;
   waveformToDisable |= mask;
   // Ensure timely service....
   if (T1L > microsecondsToClockCycles(10)) {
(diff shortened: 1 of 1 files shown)


dok-net:
@earlephilhower Off the top of my head, your fix isn't right, if you'd just cherry-pick mine from PR#7022 it will be correct?

earlephilhower:
@dok-net I'm looking at your commit https://github.com/esp8266/Arduino/pull/7022/commits/bb3ee99ecdf8280ece2ba6c57b56bc3af6fa8c5a .  I think we're both fine here.

It's doing the same thing, but this small PR sends in the `waveformToDisable` even if it was already off.  This is safe to do, as that flag is just removed from the waveform (a no-op in this case).  It does waste up to 10us, but stopping waveforms isn't a timing critical operation. 

In your PR it's also possible to have a race condition where the bit is cleared in `wvfEnabled` just *after* the main code does its check and you'd still send in the `disable` request.  It's no biggie and safe on both versions...

cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -176,19 +176,13 @@ int ICACHE_RAM_ATTR stopWaveform(uint8_t pin) {
   if (!timerRunning) {
     return false;
   }
-  // If user sends in a pin >16 but <32, this will always point to a 0 bit
-  // If they send >=32, then the shift will result in 0 and it will also return false
-  uint32_t mask = 1<<pin;
-  if (!(waveformEnabled & mask)) {
-    return false; // It's not running, nothing to do here
-  }
   waveformToDisable |= mask;
```
devyte, line 179: You deleted mask, which you then use here

earlephilhower: Ah, poop.  That's what I get for PR'ing on 3 hrs sleep.  🥱

cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -176,12 +176,6 @@ int ICACHE_RAM_ATTR stopWaveform(uint8_t pin) {
   if (!timerRunning) {
     return false;
   }
-  // If user sends in a pin >16 but <32, this will always point to a 0 bit
-  // If they send >=32, then the shift will result in 0 and it will also return false
-  uint32_t mask = 1<<pin;
```
s-hadinger, line 181: I'm getting a compilation error
```
/Users/xxx/.platformio/packages/framework-arduinoespressif8266/cores/esp8266/core_esp8266_waveform.cpp:179:24: error: 'mask' was not declared in this scope
   waveformToDisable |= mask;
```

earlephilhower: Yeah, please pull latest.  Silly mistake no my part.  Current PR works.

earlephilhower: @s-hadinger , I meant to check out #7231 in my other mail. Sorry, please use that one instead for your test.

This is unrelated to PWM stuff.

earlephilhower:
Closing

PR #7235, opened 2020-04-20
WIP - Add device tests for PWM and Tone test/dev

earlephilhower:
This is a WIP and includes a PowerShell script (yes, I hear the groans
already), but it will be moved to Python soon to allow cross-platform
compatibility.

A new directory under tests is added (because the tests/device are
self-contained and doesn't easily support the needs of this test).

A USB logic analyzer is required to be hooked up to the 8266, and a test
slave script should be uploaded.

The main test script, run on the host, will reset the chip, then run a
series of tests varying some parameter or other and collect individual
waveforms at each test point.

These raw waveforms are the post-processed to determine exact, per-cycle
phase, period, etc. parameters at that test point.

This will be repeated for every test point desired (the checked in
version now starts 2 25MHz analogWrites, one moviing from 0...1000 and
one moving from 1000...0) and calculates the performance.

Eventually this will be part of a makefile with a series of test scripts
running on a more general-purpose host test driver.

### DIFF
--- a/tests/pwm/pwmtestersut.ino
+++ b/tests/pwm/pwmtestersut.ino
@@ -1,0 +1,42 @@
+#include <ESP8266WiFi.h>
+
+const int pin[] = { D0, D1, D2, D3, D4, D5, D6, D7, D8 };
+
+void setup() {
+  WiFi.disconnect(true);
+  WiFi.forceSleepBegin();
+  Serial.begin(115200);
+  Serial.printf("\nPWM tester\n");
+  Serial.printf("z        = Reset ESP8266\n");
+  Serial.printf("d # 0/1  = pinMode(D#, OUTPUT); digitalWrite(D#, h/l)\n");
+  Serial.printf("t # freq = pinMode(D#, OUTPUT); tone(D#, freq)\n");
+  Serial.printf("f freq   = analogWriteFreq(freq)\n");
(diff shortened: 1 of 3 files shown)


tests/pwm/pwmtestersut.ino
```diff
@@ -0,0 +1,42 @@
+#include <ESP8266WiFi.h>
+
+const int pin[] = { D0, D1, D2, D3, D4, D5, D6, D7, D8 };
+
+void setup() {
+  WiFi.disconnect(true);
+  WiFi.forceSleepBegin();
+  Serial.begin(115200);
+  Serial.printf("\nPWM tester\n");
+  Serial.printf("z        = Reset ESP8266\n");
```
devyte, line 10: Instead of putting this here, I suggest an h command that prints it out.

tests/pwm/pwmtestersut.ino
```diff
@@ -0,0 +1,42 @@
+#include <ESP8266WiFi.h>
+
+const int pin[] = { D0, D1, D2, D3, D4, D5, D6, D7, D8 };
```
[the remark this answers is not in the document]
earlephilhower: Done

tests/pwm/pwmtestersut.ino
earlephilhower: Done

PR #7236, opened 2020-04-21
Waveform stopped by runtime limit in iSR doesn't deinit the timer

dok-net:
But `stopWaveform()` refuses to do anything about it if the waveform was previously stopped by expired runtime from inside ISR, either. Fix for #7230 from original issue reporter.

For some reason `stopWaveform()` is in IRAM cache and optimized for performance over size.
Also, `stopWaveform()` would be explicitly called (if indirectly) from user code, which for waveform with a runtime argument can be reasonably expected to be well adjusted in time to avoid race conditions between ISR-based cutoff vs. `stopWaveform()`.

Both facts talk back to https://github.com/esp8266/Arduino/pull/7232#issuecomment-616660681, which curiously mentions that 10µs isn't a big thing - for a function that bloats under gcc -O2 and is in IRAM cache :-)


cores/esp8266/core_esp8266_waveform.cpp
```diff
@@ -178,17 +178,15 @@ int ICACHE_RAM_ATTR stopWaveform(uint8_t pin) {
...
-  }
-  waveformToDisable |= mask;
-  // Ensure timely service....
-  if (T1L > microsecondsToClockCycles(10)) {
-    timer1_write(microsecondsToClockCycles(10));
-  }
-  while (waveformToDisable) {
-    /* no-op */ // Can't delay() since stopWaveform may be called from an IRQ
+  if (waveformsEnabled & (1UL << pin)) {
+    waveformToDisable = 1UL << pin;
```
dok-net, line 182: Everyone can read 1UL << pin, absolutely no need to introduce an object mask for this idiom. Use constness if you really feel the need to.

earlephilhower: Removing a named variable like this, used immediately after definition in the code, is not going to affect GCC's generated code unless it's god a really bad hangover. It's for readability by the next maintainer, only.

PR #7237, opened 2020-04-22
Add DNS forwarder to DNSServer 

mhightower83:
**The key functions added are:**
<br>

`bool enableForwarder(const String &domainName=emptyString, const IPAddress &dns=uint32_t)0)`

If specified, `enableForwarder` will update the `domainName` that is used to match DNS request to this AP's IP Address. A non-matching request will be forwarded to the DNS server specified by `dns`. 

Returns `true` on success.

Returns `false`, 
 * when forwarding `dns` is not set, or 
 * unable to allocate resources for managing the DNS forward function.

<br>

`void disableForwarder(const String &domainName=emptyString, bool freeResources=false)`

`disableForwarder` will stop forwarding DNS requests. If specified, updates the `domainName` that is matched for returning this AP's IP Address.
Optionally, resources used for the DNS forward function can be freed.


d-a-v:
Is it expensive in RAM when it is enabled ?
Is it always the same price, even when STA is not enabled ?
I guess it isn't useful when NAPT is not enabled.

Those questions because I wonder what is against enabling it by default, and its impact on sketch that are not using STA.

mhightower83:
The forwarder queue size is a power of 2. It is currently at 8 items, total size of 96 bytes.

I don't think an STA configured, enters directly into the DNSForwarder resource requirements.

As you indicated it isn't very useful outside of using NAPT. From time to time I see complaints about the slowly increasing of resources with each release, so I tried to make it resource flexible. Also as it is written, it is easy to increase or decrease the queue size. It is a simple constant in the .h. However, for me, 8 items seemed to work well with one host browsing. With a smaller value, I saw more dropped requests.

mhightower83:
> I don't think an STA configured, enters directly into the DNSForwarder resource requirements.

Sorry, I wasn't thinking that all the way through. W/O STA there is no need for the forwarder.

libraries/DNSServer/examples/NAPTCaptivePortal/WifiHttp.h
```diff
@@ -0,0 +1,330 @@
+#ifndef WIFIHTTP_H_
```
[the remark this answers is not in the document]
mcspr: ...and also, why not a single .ino entrypoint, with .cpp actually implementing things?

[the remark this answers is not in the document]
mhightower83: > ...and also, why not a single .ino entrypoint, with .cpp actually implementing things?

I don't think I follow this point.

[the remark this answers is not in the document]
mcspr: > > ...and also, why not a single .ino entrypoint, with .cpp actually implementing things?
> 
> I don't think I follow this point.

Only keeping NAPTCaptivePortal.ino for setup() & loop(), for the sole purpose of avoiding ide concatenation of .ino files. Just a consideration, as I noted I may be out-of-loop and this is actually easier to manage within IDE

libraries/DNSServer/src/DNSServer.cpp
```diff
@@ -1,33 +1,139 @@
...
+{
+  disableForwarder(domainName, false); // Just happens to have the same logic needed here.
+
+  if (dns.isSet()) {
+    _dns = dns;
+  }
+
+  if (_dns.isSet()) {
+    if (!_que) {
+      _que = std::unique_ptr<struct DNSS_REQUESTER[]> (new (std::nothrow) struct DNSS_REQUESTER[kDNSSQueSize]);
```
mcspr, line 106: like typedef, struct is redundant as c++ will deduce what it is

libraries/DNSServer/src/DNSServer.h
```diff
@@ -45,31 +53,76 @@ struct DNSHeader
...
     // stops the DNS server
     void stop();
 
   private:
     WiFiUDP _udp;
-    uint16_t _port;
     String _domainName;
-    unsigned char _resolvedIP[4];
+    IPAddress _dns;
+    std::unique_ptr<struct DNSS_REQUESTER[]> _que;
```
mcspr, line 113: ^ redundant struct prefix

libraries/DNSServer/src/DNSServer.h
```diff
@@ -45,31 +53,76 @@ struct DNSHeader
...
+#ifdef DEBUG_DNSSERVER
+    // There are 2 possiblities for OverFlow:
+    //   1) we have more than kDNSSQueSize request already outstanding.
+    //   2) we have request that never received a reply.
+    uint32_t _que_ov;
+    uint32_t _que_drop;
+#endif
     DNSReplyCode _errorReplyCode;
+    bool _forwarder;
+    unsigned char _resolvedIP[4];
```
mcspr, line 124: btw shouldn't this be IPAddress?

mcspr (approved):
Thanks!
Just tried with a couple of devices, all seems to work like intended

libraries/DNSServer/examples/NAPTCaptivePortal/NAPTCaptivePortal.ino
```diff
@@ -0,0 +1,391 @@
...
+unsigned int status = WL_IDLE_STATUS;
+
+void setup() {
+  WiFi.persistent(false);  // w/o this a flash write occurs at every boot
+  WiFi.mode(WIFI_OFF);     // Prevent use of SDK stored credentials
+  CONSOLE.begin(115200);
+  CONSOLE_PRINTLN("\r\n\r\nNAPT with Configuration Portal ...");
+
+  staModeConnectedHandler = WiFi.onStationModeConnected(
+    [](const WiFiEventStationModeConnected data) {
```
mcspr, line 179:
```suggestion
    [](const WiFiEventStationModeConnected& data) {
```
(what I meant before, see the signature. still a copy)
```
### instance-method `onStationModeConnected`
──────────────────────────────────────────────────────────────────────────────────────
Parameters:
 - `std::function<void (const WiFiEventStationModeConnected &)>`

Subscribe to specific event and get event information as an argument to the callback
 ──────────────────────────────────────────────────────────────────────────────────────
 // In ESP8266WiFiGenericClass
public: WiFiEventHandler onStationModeConnected(
    std::function<void(const WiFiEventStationModeConnected &)>)
```

libraries/DNSServer/examples/NAPTCaptivePortal/NAPTCaptivePortal.ino
```diff
@@ -0,0 +1,391 @@
...
+      // Keep a copy of the BSSID for the AP that WLAN connects to.
+      // This is used in the WLAN report on WiFi Details page.
+      memcpy(bssid, data.bssid, sizeof(bssid));
+    });
+
+  staModeDisconnectedHandler = WiFi.onStationModeDisconnected(
+    [](const WiFiEventStationModeDisconnected) {
```
mcspr, line 186:
```suggestion
    [](const WiFiEventStationModeDisconnected&) {
```

PR #7254, opened 2020-04-29
Make getContentType available for 3rd party usage

vdeconinck:


vdeconinck:
@devyte I guess it's OK now. Let's see what the CI says...

Issue #7262, opened 2020-05-02
mDNS sudden death


jjsuwa:
**MCVE:**
Simple WiFiSTA + mDNS, and LED blink as coalmine canary :)
- Hardware: [Generic ESP8266 (ESP-WROOM-02 breakout PCB w/DTR reset circuitry)]
- Core Version: commit 4e3a4b6d210054242e5d7a13bcd3467819518cfa (latest @ the moment)
- **[Edited 3]** CPU Frequency: 160 MHz
- **[Edited 3]** lwIP Variant: v2 Lower Memory (no features)
- **[Edited 2]** Espressif FW: nonos-sdk 2.2.1+119 (191122)
```cpp
#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>

#define AP_SSID       "YOUR-SSID"
#define AP_PASSWORD   "YOUR-PASSWORD"
#define MDNS_HOSTNAME "WiFiSta"

#define LED_PIN 4
#define DIVIDER 10000

void setup() {
  pinMode(LED_PIN, OUTPUT);
  WiFi.persistent(false);
  WiFi.mode(WIFI_OFF);
  Serial.begin(115200);
  delay(250);

  WiFi.mode(WIFI_STA);
  Serial.print(F("\n\n\n"
                 "WiFi(STA): SSID=\"" AP_SSID "\".\n"
                 "WiFi(STA): connecting"));
  WiFi.begin(F(AP_SSID), F(AP_PASSWORD));
  for (; ; delay(500)) {
    if (WiFi.status() == WL_CONNECTED) {
      Serial.printf_P(PSTR(", done.\n"
                           "WiFi(STA): IP address=%s/%s.\n"
                           "mDNS: hostname=\"" MDNS_HOSTNAME ".local\".\n"), WiFi.localIP().toString().c_str(), WiFi.subnetMask().toString().c_str());
      MDNS.begin(F(MDNS_HOSTNAME));
      break;
    }
    Serial.print('.');
  }
}

void loop() {
  MDNS.update();

  // visible canary
  static unsigned int counter = DIVIDER;
  if (--counter == 0) {
    counter = DIVIDER;
    digitalWrite(LED_PIN, digitalRead(LED_PIN) == 0);
  }
}
```
**Symptom:**
- After WiFi connecting done, 1st mDNS response will almost always be fine.
- But later attempts will not often be responded without any signs, especially at some interval (a few minites~).
[...]

PR #7266, opened 2020-05-03
Revert "Changing listen to listen the current iface only instead of 0 (#7217)"

jjsuwa:
* #7262 hotfix


devyte:
Merging this as a temporary workaround.

PR #7276, opened 2020-05-06
Bring SPIFFS up to latest version

earlephilhower:
Manually merged the PATCH files from upstream up until Jan 26, 2020.

Looks to only fix a single bug related to reading past the end of a file
in certain cases.  Other new features (secure erase) are included to
make it easier to merge any future releases, but are not enabled in the
core.

Merges included:
````
20fc6df0ab2bee391961b5f823a327887abc96b4
da1cf494796d68bb5c02ba70cf78a904db08a170
13935985cdde6d6b4ec77f2685264242ee55e7ac
ec68ba8208d7550860e4e78299d58a529b88bf85
f59d958e0b2f169b549e5cdc293a70bc6873cd45
f7d3e9f2b207958dfc2c01cf3fc42f98a4b9b239
554b59c147d96223ccce18374989f191f7222b45
07e013056dd976d12ae36db5c54a275e1497f6c8
d0d44c4908efff7a443ed9b1174d9173fb784a72
3cb24dbfd32a396dca46a7ca56a8e75bdcb97078
8172b40813ff36aa0df4c88a6e9c76e5300d7e32
````

### CODE REVIEW

**Comment by earlephilhower** [comment]
@TD-er, you seem to be a big SPIFFS user.  Can you give this a try?  I really don't think it should change any behavior, but would love independent verification beyond our host tests.

File: cores/esp8266/spiffs/spiffs_config.h [hunk 1]
```diff
@@ -326,6 +326,17 @@ typedef uint8_t u8_t;
...
 
+// By default SPIFFS in some cases relies on the property of NOR flash that bits
+// cannot be set from 0 to 1 by writing and that controllers will ignore such
+// bit changes. This results in fewer reads as SPIFFS can in some cases perform
+// blind writes, with all bits set to 1 and only those it needs reset set to 0.
+// Most of the chips and controllers allow this behavior, so the default is to
+// use this technique. If your controller is one of the rare ones that don't,
+// turn this option on and SPIFFS will perform a read-modify-write instead.
+#ifndef SPIFFS_NO_BLIND_WRITES
+#define SPIFFS_NO_BLIND_WRITES                0
```
TD-er, line 337: Hmm looks like we have also our PUYA patch in a single define here, or am I missing something?

devyte: > or am I missing something?

This applies to spiffs directly, it's meant for other platforms. We have it disabled, because our patch is implemented at a lower level: at flash read/write. If we were to just use this one, we wouldn't be able to use eeprom or ota, etc.

TD-er: Ah yes EEPROM is the one I forgot.
I am not entirely sure we also need it for OTA, unless there are blocks partially updated during the OTA. Otherwise the data to be written is also going to be written on PUYA chips.

devyte:
@TD-er ping

TD-er:
Sorry, it slipped my mind to make a test build for it.
Just by looking at the code I don't see any obvious issues why it shouldn't work, but I will try to make a test build.  (should be less problematic these days with the `platform_packages` introduced in PlatformIO)

TD-er:
I've made a test build with it and have not seen any issues so far.

It was only tested by me, so still no heavy testing, but the normal things in ESPEasy run just fine with it.

PR #7284, opened 2020-05-09
Fix pathValid() filename length checking

abderraouf-adjal:
Fix pathValid() filename length checking.  
Ref: https://github.com/ARMmbed/littlefs/blob/4c9146ea539f72749d6cc3ea076372a81b12cb11/lfs.c#L2428


devyte:
CI fails because a test fails with this change. The test must be checked and updated as well.

abderraouf-adjal:
Related to this 2 lines probably: [/tests/host/fs/test_fs.inc#L115](https://github.com/esp8266/Arduino/blob/56bbb99e6f12b5425d45246907b7496d61b4bd9d/tests/host/fs/test_fs.inc#L115)

earlephilhower:
Also need to adjust the too-long filename to make it really too long:  https://github.com/esp8266/Arduino/blob/56bbb99e6f12b5425d45246907b7496d61b4bd9d/tests/host/fs/test_fs.cpp#L70

abderraouf-adjal:
LittleFS passed, but SPIFFS test failed.
FS test need separated functions...

```
fs/test_fs.inc:123: FAILED:
  REQUIRE( SPIFFS.open(longName_32, "w") )
with expansion:
  false
```

abderraouf-adjal:
Closed: LittleFS routines strip leading slashes before doing anything.

PR #7296, opened 2020-05-11
ESP8266WebServer: Add variadic template version of collectHeaders()

jjsuwa:
* More user-friendly, less RODATA usage.

eg. `webServer.collectHeaders(F("Content-Type"), F("Origin"));`

In this example, less about 20 bytes than the traditional way.


earlephilhower:
@jjsuwa, it looks like this changes the API in a way that breaks existing examples.  Can you check the Travis CI logs and see?

jjsuwa:
Hmm, Travis CI notification stalled?
Build #10881 [https://travis-ci.org/github/esp8266/Arduino/builds/685840835](url) seems done.

d-a-v:
CI has restarted, sorry for the interruption

earlephilhower:
@devyte, this is some deep template voodoo.  Can you give it a look-see?

jjsuwa:
Should `libraries/ESP8266HTTPClient` be done like this?

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -644,14 +643,27 @@ template <typename ServerType>
...
-     delete[]_currentHeaders;
+    delete[] _currentHeaders;
   _currentHeaders = new RequestArgument[_headerKeysCount];
   _currentHeaders[0].key = FPSTR(AUTHORIZATION_HEADER);
   for (int i = 1; i < _headerKeysCount; i++){
     _currentHeaders[i].key = headerKeys[i-1];
   }
 }
 
+template <typename ServerType>
```
devyte, line 654: It's been a while since I last saw this use of variadic args. Pretty nice!
There are several things at play here, but most should be ok for 99% of use cases, so it's ok. 
However, I don't like the code duplication. This method body is almost the same as the method above it, and I was already confused for a moment due to the different +1.
How about implementing this method in terms of the above, variadic args and all? Something along the lines of declaring an array of headers with args, and then call the above method?

jjsuwa: > However, I don't like the code duplication.

Who likes code duplication? :)

> How about implementing this method in terms of the above, variadic args and all? Something along the lines of declaring an array of headers with args, and then call the above method?

What you wanna say I guess is:
```
// [ ESP8266WebServer.h ]
protected:
  void _collectHeaders(String headerKeys[], const size_t headerKeysCount);
```
```
// [ ESP8266WebServer-impl.h ]
void ESP8266WebServerTemplate<ServerType>::_collectHeaders(String headerKeys[], const size_t headerKeysCount) {
  _headerKeysCount = headerKeysCount + 1;
  if (_currentHeaders)
    delete[] _currentHeaders;
  _currentHeaders = new RequestArgument[_headerKeysCount];
  _currentHeaders[0].key = FPSTR(AUTHORIZATION_HEADER);
  for (int i = 1; i < _headerKeysCount; i++){
    _currentHeaders[i].key = headerKeys[i-1];
  }
}

void ESP8266WebServerTemplate<ServerType>::collectHeaders(const char* headerKeys[], const size_t headerKeysCount) {
  String headers[headerKeysCount];
  for (int i = 0; i < headerKeysCount; i++) {
    headers[i] = String(headerKeys[i]);
  }
  _collectHeaders(headers, headerKeysCount);
}

template <typename... Args>
void ESP8266WebServerTemplate<ServerType>::collectHeaders(const Args&... args) {
  String headers[] = {
    args...
  };
  _collectHeaders(headers, sizeof...(args));
}
```
and inspection code:
```
#include <ESP8266WebServer.h>
static ESP8266WebServer webServer(80);
void setup() {
  webServer.collectHeaders(F("Content-Type"), F("Origin"));
}
void loop() {
  webServer.handleClient();
}
```
Finally, IROM grows 251072 bytes to 251152... inefficient, vain efforts.

devyte (approved):

PR #7299, opened 2020-05-12
Graph example

vdeconinck:
Hello. As [promised](https://github.com/esp8266/Arduino/pull/7182#issuecomment-620929417) during refactoring of FSBrowser in #7182 , I created a new stand-alone Graph application, demonstrating the periodic HTTP call to the ESP to get GPIO, analog and heap values and display them as charts.
The result looks like this:
image
@devyte: Many many thanks for the help. I'll be glad to get your feedback on the code.


earlephilhower:
Very cool looking.  One thing I'm noting on FF latest is that the lower 2 graphs (analogRead/freeHeap) have negative values on the Y axis, and should really go from 0....n

image

libraries/ESP8266WebServer/examples/Graph/data/index.htm
```diff
@@ -0,0 +1,527 @@
...
+        xAxes: [{
+          id:"x-axis",
+          ticks: {
+            minRotation:0,
+            maxRotation:0
+          }
+        }],
+        yAxes: [{
+          ticks: {
+            beginAtZero:true
```
earlephilhower, line 145: Maybe add 'min:0'  to this one since the analogInput can't be < 0?

libraries/ESP8266WebServer/examples/Graph/data/index.htm
```diff
@@ -0,0 +1,527 @@
...
```
earlephilhower, line 218: Same 'min:0' comment here as above.

libraries/ESP8266WebServer/examples/Graph/Graph.ino
```diff
@@ -0,0 +1,318 @@
...
+#define STAPSK  "your-password"
+#endif
+
+// Indicate which digital I/Os should be displayed on the chart.
+// From GPIO16 to GPIO0, a '1' means the corresponding GPIO will be shown
+// Note: SD GPIOs are hidden by default:
+#define DEFAULT_GPIO_MASK 0b11111000000111111
+
+unsigned int gpioMask = DEFAULT_GPIO_MASK;
+
```
earlephilhower, line 70: A better way of getting the "safe" GPIO mask programatically using the `isFlashInterface` API
````
mask = 0x1ffff;
for (auto pin=0; pin <= 16; pin++) if (isFlashInterfacePin(pin)) mask &= ~(1<<pin);
````

File: libraries/ESP8266WebServer/examples/Graph/Graph.ino [hunk 4]
```diff
@@ -0,0 +1,318 @@
...
+        rgbMode = 0;
+      }
+
+      modeChangeRequested = false;
+    }
+
+    // act according to mode
+    switch (rgbMode) {
+      case 0: // off
+        gpioMask = 0b10100000000111111; // GPIOs 12-13-15 are hidden
```
earlephilhower, line 288: Same comment about gpio mask as above

libraries/ESP8266WebServer/examples/Graph/data/index.htm
vdeconinck: Although I cannot reproduce the issue of negative scale upon first load (FFx 76.0.1 (64 bits) Win10), I can get negative numbers using the "unzoom" feature (e.g. using mouse wheel).
Unfortunately, min:0 does not seem to prevent negative values from being shown :-(.
I guess the zoom plugin overrides the values set for the Y axis.
A single click on the graph should reset the scale, though...
I proposed the zoom plugin following a discussion with @devyte to allow both a "full range" view as well as a focus on small changes, because each has use cases, but it's not perfect :-/.

libraries/ESP8266WebServer/examples/Graph/data/index.htm
vdeconinck: see above

libraries/ESP8266WebServer/examples/Graph/Graph.ino
vdeconinck: Cool, didn't know that. Will change accordingly.

libraries/ESP8266WebServer/examples/Graph/Graph.ino
vdeconinck: see above

libraries/ESP8266WebServer/examples/Graph/Graph.ino
vdeconinck: Done. Thanks for the tip.

libraries/ESP8266WebServer/examples/Graph/Graph.ino
```diff
@@ -0,0 +1,334 @@
...
+  DBG_OUTPUT_PORT.begin(115200);
+  DBG_OUTPUT_PORT.setDebugOutput(true);
+  DBG_OUTPUT_PORT.print('\n');
+
+  ////////////////////////////////
+  // FILESYSTEM INIT
+
+  fileSystemConfig.setAutoFormat(false);
+  fileSystem->setConfig(fileSystemConfig);
+  boolean fsOK = fileSystem->begin();
```
devyte, line 185: Please use bool

libraries/ESP8266WebServer/examples/Graph/Graph.ino
```diff
@@ -0,0 +1,334 @@
...
+      mask &= ~(1 << pin);
+    }
+  }
+  return mask;
+}
+
+int rgbMode = 1; // 0=off - 1=auto - 2=manual
+int rgbValue = 0;
+long lastChangeTime = 0;
+boolean modeChangeRequested = false;
```
devyte, line 269: bool

libraries/ESP8266WebServer/examples/Graph/Graph.ino
```diff
@@ -0,0 +1,334 @@
...
+  MDNS.update();
+
+  if (digitalRead(4) == 0) {
+    // button pressed
+    modeChangeRequested = true;
+  }
+
+  // see if one second has passed since last change
+  long now = millis();
+  if (now - lastChangeTime > 1000) {
```
devyte, line 282: Although tbis ks technically correct, it reinvents the wheel. Please use a global esp8266::polledTimeout::periodicMs, see core examples for use.

devyte: Also, prefer early return to reduce cyclomatic complexity of what follows, i. e. 
```cpp
esp8266::polledTimeout::periodicMs timeToChange(1000);
...

if(!timeToChange)
    return;

if(modeChangeRequested) {
... 
```

vdeconinck: Didn't know about periodicMs. Nice :-)

PR #7311, opened 2020-05-18
Flash size reduction for mime-type

s-hadinger:
The current mime-type array is based on `Entry` structure with fixed size string that waste a lot of Flash space (~600 bytes):
```
struct Entry
{
  const char endsWith[16]; 
  const char mimeType[32];
};
```

I replaced with 2 arrays for suffix and mime-type, with PROGMEM strings.

I also added `#define MIMETYPE_MINIMAL` to reduce the footprint to mime-types that are strictly necessary: `html`, `txt`, `gz`

Overall this saves a little over 1kB for Tasmota.


d-a-v:
Nice but it's an API breaking change.

s-hadinger:
Oh, if `mimeTable[]` is part of the API, I can put it back in a single array which would be API-compatible. Let me do the change.

s-hadinger:
Closed in favor of #7312

PR #7312, opened 2020-05-18
Flash size reduction for mime-type

s-hadinger:
Superseded #7311 

The current mime-type array is based on `Entry` structure with fixed size string that waste a lot of Flash space (~400 bytes):
```
struct Entry
{
  const char endsWith[16]; 
  const char mimeType[32];
};
```

I replaced with standard PROGMEM strings.

I also added `#define MIMETYPE_MINIMAL` to reduce the footprint to mime-types that are strictly necessary: `html`, `txt`, `gz`

Overall this saves 1kB for Tasmota.

### DIFF
--- a/libraries/ESP8266WebServer/src/detail/mimetable.cpp
+++ b/libraries/ESP8266WebServer/src/detail/mimetable.cpp
@@ -7,8 +7,9 @@ namespace mime
 
 static const char kHtmlSuffix[] PROGMEM = ".html";
 static const char kHtmSuffix[] PROGMEM = ".htm";
-static const char kCssSuffix[] PROGMEM = ".css";
 static const char kTxtSuffix[] PROGMEM = ".txt";
+#ifndef MIMETYPE_MINIMAL
+static const char kCssSuffix[] PROGMEM = ".css";
 static const char kJsSuffix[] PROGMEM = ".js";
 static const char kJsonSuffix[] PROGMEM = ".json";
 static const char kPngSuffix[] PROGMEM = ".png";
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WebServer/src/detail/mimetable.cpp
```diff
@@ -5,48 +5,93 @@
 namespace mime
 {
 
-// Table of extension->MIME strings stored in PROGMEM, needs to be global due to GCC section typing rules
-const Entry mimeTable[maxType] PROGMEM = 
+static const char kHtmlSuffix[] PROGMEM = ".html";
+static const char kHtmSuffix[] PROGMEM = ".htm";
+static const char kCssSuffix[] PROGMEM = ".css";
```
devyte, line 10: Missing MIMETYPE_MINIMAL conditional

libraries/ESP8266WebServer/src/detail/mimetable.cpp
```diff
@@ -5,48 +5,93 @@
...
+static const char kSfntSuffix[] PROGMEM = ".sfnt";
+static const char kXmlSuffix[] PROGMEM = ".xml";
+static const char kPdfSuffix[] PROGMEM = ".pdf";
+static const char kZipSuffix[] PROGMEM = ".zip";
+static const char kGzSuffix[] PROGMEM = ".gz";
+static const char kAppcacheSuffix[] PROGMEM = ".appcache";
+static const char kDefaultSuffix[] PROGMEM = "";
+
+static const char kHtml[] PROGMEM = "text/html";
+static const char kCss[] PROGMEM = "text/css";
```
devyte, line 34: Missing MIMETYPE_MINIMAL conditional

libraries/ESP8266WebServer/src/detail/mimetable.cpp
s-hadinger: The conditional is not strictly needed, the linker will remove the strings that are not referenced. I can add an explicit conditional.

s-hadinger: Conditional added.

earlephilhower:
One of the VMs died during the CI process.  Just restarted, once it's good I'll hit merge.

PR #7313, opened 2020-05-18
fixed crash excpetion 28 at 0x33

st0ff3r:
netif is NULL if interface is down


st0ff3r:
it happens randomly when I call wifi_set_opmode_current(NULL_MODE) and later 	wifi_set_opmode_current(STATIONAP_MODE); to restart network interface

earlephilhower:
LWIP 1.x has been removed, closing.

Issue #7307, opened 2020-05-18
[Core 2.7.0/2.7.1] ESP.restart() or ESP.reset() cause bootloop


rev1204:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:					ESP-12S
- Core Version:			2.7.0 and 2.7.1
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			dio
- Flash Size: 		4MB
- lwip Variant:   v2 Lower Memory
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	80Mhz
- CPU Frequency:		160MHz
- Upload Using:		SERIAL
- Upload Speed:   921600

### Problem Description

Core 2.7.0 and 2.7.1 ESP.restart() and ESP.reset() causes ESP to bootloop (ESP keep restarting, never start the program again). Unfortunately ESP didn't print any exception. In the debug message below, the program ran first time after Uploading. After first restart, ESP not printing serial and keep restarting forever at 1 sec interval. This is not happen in Core 2.6.3.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
void setup() {
	delay(500);
	Serial.begin(115200);
	Serial.println("\nSTART");
	delay(5000);
	ESP.restart();
}

void loop() {
}
```

### Debug Messages

```
START

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 3456, room 16 
tail 0
chksum 0x84
csum 0x84
va5432625
~ld

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 3456, room 16 
tail 0
chksum 0x84
csum 0x84
va5432625
~ld
```

d-a-v:
Unfortunately your sketch is working as expected here (using a d1 mini running latest git master).

Are the IOs connected to something or left unconnected ?
Did you try with the full flash erase option ?
[...]

Issue #7267, opened 2020-05-04
OTA update fails with XMC Flash chip


Jason2866:
OTA update fails (used gz Tasmota firmware) when a XMC Flash is on the ESP board. 
Module Model: ESP-12F Vendor: DOITING
Flash Chip Id  0x164020

To get it working (again) a serial flash is needed

```
 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 3656, room 16

tail 8

chksum 0x0c

csum 0x0c

v9c56ed1f

```

### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:			[ESP-12F]
- Core Version:			[2.7.0]
- Development Env:		[Platformio]
- Operating System:		[Windows|Ubuntu]

### Settings in IDE

- Module:  			[Generic ESP8266]
- Flash Mode:			[DOUT]
- Flash Size: 		        [4MB]
- lwip Variant:                   [LWIP2_HIGHER_BANDWIDTH_LOW_FLASH]
- Reset Method:		[nodemcu]
- Flash Frequency:	        [40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA]
- Upload Speed:               [115200] (serial upload only)

### Problem Description

Flashing via serial does work and device works as expected.
Trying a OTA update results in this

### Debug Messages

```
 ets Jan  8 2013,rst cause:2, boot mode:(3,7)

load 0x00000000, len 201326592, room 16


```

Jason2866:
Did some test #6725 causes the issue. Using a eboot.elf version before this commit solves the problem

d-a-v:
Thanks @Jason2866 for the bissect !
@ChocolateFrogsNuts Does it ring a bell ?

crashcoq:
Same problem for me on OTA only

Start OK
`
21:38:58.882 ->  ets Jan  8 2013,rst cause:2, boot mode:(3,6)
21:38:58.882 -> 
[...]

PR #7317, opened 2020-05-19
Xmc flash 2

ChocolateFrogsNuts:
Remove the unnecessary  XMC chip support from eboot and verify the flash contents after copying new firmware.

Testing shows there should be no issue with the XMC chips in eboot because the flash access speed is always 20MHz, however verifying the content written will help identify any issues if they do occur.
The additional code is minimal, and as it only reads, the verify is very quick.


Tech-TX:
That works with both failing cases for me (OTA upload and ESP.restart() at 160MHz),
https://github.com/esp8266/Arduino/issues/7307
https://github.com/esp8266/Arduino/issues/7267

I've already merged https://github.com/esp8266/Arduino/pull/7277 into my local fork, so I'm not sure if #7267 is needed in addition to this PR.

devyte:
@Tech-TX #7277 was a rollback of the first implementation. This should be the fixed final solution.

devyte:
The verify is supposed to cover the case when the flash write seems to succeed, i. e. no error, but in truth it didn't take at hw level due to whatever reason.

ChocolateFrogsNuts:
@earlephilhower the check is relatively cheap (in terms of code size and execution time) and if we do have problems with OTA updates to XMC chips will make it easy to identify if it is a problem with writing to the chip.
Theoretically we could store ACTION_COPY_FAILED in the RTC and have that cause printing the error and a while(true){}
That way if an OTA update copy failed, the chip would halt, then you could connect to the USB and press reset to get the reason dumped to monitor. If the eboot command is moved to a non-volatile location, it would even work over power cycles (although you would need to clear that the eboot command as part of a USB update).

PR #7339, opened 2020-05-29
comments in variant files to centralize useful informations in one place

d-a-v:
It is not really a webpage, but it helps prototyping without having many opened window.

### DIFF
--- a/variants/d1_mini/pins_arduino.h
+++ b/variants/d1_mini/pins_arduino.h
@@ -32,19 +32,25 @@
 static const uint8_t SDA = PIN_WIRE_SDA;
 static const uint8_t SCL = PIN_WIRE_SCL;
 
+#ifndef LED_BUILTIN
 #define LED_BUILTIN 2
+#endif
 
-static const uint8_t D0   = 16;
-static const uint8_t D1   = 5;
-static const uint8_t D2   = 4;
-static const uint8_t D3   = 0;
-static const uint8_t D4   = 2;
-static const uint8_t D5   = 14;
-static const uint8_t D6   = 12;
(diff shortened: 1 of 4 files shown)


variants/d1_mini/pins_arduino.h
```diff
@@ -32,19 +32,25 @@
...
-static const uint8_t D6   = 12;
-static const uint8_t D7   = 13;
-static const uint8_t D8   = 15;
-static const uint8_t RX   = 3;
-static const uint8_t TX   = 1;
+#endif
+
+
+//                D1-mini  GPIO
+static const uint8_t D0   = 16; // (no pwm, no intr, internal-pull-down-poosible-no-pullup)
```
earlephilhower, line 41: We do actually support PWM on GPIO16.  Also, typo in "poosible"=>"possible"

earlephilhower: Also, if you're listing functions, then I would suggest adding the I2S :)

image

d-a-v: The purpose of this proposal is to have only one place without having to refer elsewhere (ie. with the board specific pinout naming).
It's true it's a lot of work. That's why I started with the most common ones.
I can populate the other files and add urls somewhere to point them so they also can be turned into an eternal link/webpage like "[d1 mini full info](https://github.com/d-a-v/Arduino/blob/in-source-gpio-doc/variants/d1_mini/pins_arduino.h#L40)".

Didn't know about GPIO16 :+1: (so now we know that this link will be more up-to-date :grin:).
I2S will be added for further review iteration.

PR #7351, opened 2020-06-05
elf2bin.py objcopy cleanup

mcspr:
Per https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp
> Unlike TemporaryFile(), the user of mkstemp() is responsible for deleting the temporary file when done with it.

`read_segment()`'s `dumpfile` was never deleted, producing 5 'segment' files each build.

Instead, this patch creates a temporary directory which is later used for objcopy output. `TemporaryDirectory()` context manager will handle the deletion.
Plus, some minor refactoring to reduce line length and add some pythonisms (hopefully, not too much)

Noticed this issue accidentally by inspecting a specific Windows PC install using WSL circa January. `/tmp` was ~2.5GB and there was a *lot* of files.

### DIFF
--- a/tools/elf2bin.py
+++ b/tools/elf2bin.py
@@ -55,10 +55,10 @@ def get_segment_size_addr(elf, segment, path):
     raise Exception('Unable to find size and start point in file "' + elf + '" for "' + segment + '"')
 
 def read_segment(elf, tmpdir, segment, path):
-    dumpfile = os.path.join(tmpdir, segment + '.read_segment')
-    p = subprocess.check_call([path + "/xtensa-lx106-elf-objcopy", '-O', 'binary', '--only-section=' + segment, elf, dumpfile], stdout=subprocess.PIPE)
-    with open(dumpfile, "rb") as binfile:
-        raw = binfile.read()
+    tmpfile = os.path.join(tmpdir, segment + '.read_segment')
+    p = subprocess.check_call([path + "/xtensa-lx106-elf-objcopy", '-O', 'binary', '--only-section=' + segment, elf, tmpfile], stdout=subprocess.PIPE)
+    with open(tmpfile, "rb") as f:
+        raw = f.read()
 
     return raw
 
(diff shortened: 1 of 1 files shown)


tools/elf2bin.py
```diff
@@ -152,12 +151,24 @@ def main():
...
+        )
 
-    out = open(args.out, "wb")
-    write_bin(out, args.eboot, ['.text'], 4096, args.flash_mode, args.flash_size, args.flash_freq, args.path)
-    write_bin(out, args.app, ['.irom0.text', '.text', '.text1', '.data', '.rodata'], 0, args.flash_mode, args.flash_size, args.flash_freq, args.path)
-    out.close()
+        wrapper(
+            elf=args.app,
+            segments=[".irom0.text", ".text", ".text1", ".data", ".rodata"],
+            to_addr=0,
```
earlephilhower, line 170: Extra comma here

tools/elf2bin.py
```diff
@@ -55,28 +55,29 @@ def get_segment_size_addr(elf, segment, path):
...
-    p = subprocess.check_call([path + "/xtensa-lx106-elf-objcopy", '-O', 'binary', '--only-section=' + segment, elf, dumpfile], stdout=subprocess.PIPE)
-    binfile = open(dumpfile, "rb")
-    raw = binfile.read()
-    binfile.close()
+    fd, tmpfile = tempfile.mkstemp()
+    os.close(fd)
+    p = subprocess.check_call([path + "/xtensa-lx106-elf-objcopy", '-O', 'binary', '--only-section=' + segment, elf, tmpfile], stdout=subprocess.PIPE)
+    with open(tmpfile, "rb") as f:
+        raw = f.read()
+    os.remove(tmpfile)
```
mcspr, line 63: re TemporaryDirectory, just removing the file right here also works.

earlephilhower: Nice and simple!

PR #7354, opened 2020-06-07
EspSoftwareSerial service release 6.8.2

dok-net:


d-a-v:
Is it stable enough, or is it an urgent / bugfix commit that needs to be marked for 2.7.2 ?
Or should we wait for after 2.7.2 ?

dok-net:
@d-a-v It's a correction for defaults handling in the `begin()` function, fixing RX buffer size deduction. And it's already published via Arduino library manager due to my no-risk-involved assessment. I would like to ask you to include it ASAP, ahead of 2.7.2, to avoid the usual confusion among our users. Thanks for the quick response!

PR #7362, opened 2020-06-08
libraries/SPI: abs -> std::abs and cast fixes

mcspr:
Not to complicate comment chain in #6294 further

SPI library code erroneously assumed that <s>abs() is a C function, so include stdlib.h is required.
  what happens instead is Arduino.h shadows `abs()` with it's own macro</s>
`freq - calFreq` or `freq - bestFreq` may produce negative results and require `abs()`
In the referenced PR, when adding `using std::abs;` it confuses GCC by introducing multiple functions named abs(), neither of which have uint32_t argument. As it turns out, `uint32_t() - int32_t()` promotes to `uint32_t`, where old libc abs() silently converted it back to `int`

Fixing casts for `freq` to explicitly make it a signed int when comparing with generated values, both of which are signed.

*edit: noting that abs() isn't a macro and we actually want signedness*

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -203,6 +203,7 @@ void SPIClass::setFrequency(uint32_t freq) {
     static uint32_t lastSetRegister = 0;
 
     if(freq >= ESP8266_CLOCK) {
+        // magic number to set spi sysclock bit (see below.)
         setClockDivider(0x80000000);
         return;
     }
@@ -215,7 +216,7 @@ void SPIClass::setFrequency(uint32_t freq) {
     const spiClk_t minFreqReg = { 0x7FFFF020 };
     uint32_t minFreq = ClkRegToFreq((spiClk_t*) &minFreqReg);
     if(freq < minFreq) {
-        // use minimum possible clock
+        // use minimum possible clock regardless
         setClockDivider(minFreqReg.regValue);
         lastSetRegister = SPI1CLK;
         lastSetFrequency = freq;
@@ -227,8 +228,14 @@ void SPIClass::setFrequency(uint32_t freq) {
     spiClk_t bestReg = { 0 };
     int32_t bestFreq = 0;
 
-    // find the best match
-    while(calN <= 0x3F) { // 0x3F max for N
+    // aka 0x3F, aka 63, max for regN:6
+    const uint8_t regNMax = (1 << 6) - 1;
+
+    // aka 0x1fff, aka 8191, max for regPre:13
+    const int32_t regPreMax = (1 << 13) - 1;
+
+    // find the best match for the next 63 iterations
+    while(calN <= regNMax) {
 
         spiClk_t reg = { 0 };
         int32_t calFreq;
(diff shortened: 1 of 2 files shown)


libraries/SPI/SPI.cpp
```diff
@@ -260,7 +260,7 @@ void SPIClass::setFrequency(uint32_t freq) {
                 break;
             } else if(calFreq < (int32_t) freq) {
                 // never go over the requested frequency
-                if(abs(freq - calFreq) < abs(freq - bestFreq)) {
+                if((freq - calFreq) < (freq - bestFreq)) {
```
devyte, line 263: I'm not convinced this is correct.
Strictly speaking, the operation is happening on two unsigned values, bjt that is only due to "Usual Arithmetic Conversions", i. e. freq is unsigned, calFreq is signed, so calFreq is implicitly converted to unsigned for the subtraction, which then of course makes the abs() call redundant. 
However, looking at the code above, there are several casts to int32. Either:
1. the subtract operation makes sense to happen on signed args, in which case abs() *should be kept* and freq should be cast to int32 before passing it in
or
2. having signed types doesn't make sense, in which case the types should all be changed to unsigned, and the abs() call should be removed as proposed 

Given that the statement is a comparison between deltas, I'm inclined to think 1 is correct, but my opinion is without fully understanding the surrounding code.

devyte: If 1 ks found to be correct, then std::abs should be used instead of abs

mcspr: Having just checked that calFreq is less than freq, abs() is still redundant? We can only calculate ≥0 values there. bestFreq is calFreq of the previous loop

I am not really following what "we calculate in int so we miss the decimals, testing is the easyest and fastest way" means at the start of the while() loop.

devyte: That would be true for the 1st delta, but there's still the 2nd delta: bestFreq is signed, freq is unsigned.

earlephilhower: The code in question is trying to find if the absolute value of a potential frequency is closer to the requested one.  Answering the question, "Is this tested frequency closer to the requested one than my current best guess?"

For that, you'd want `abs(signed-diff(thisFreq,desired)) < abs(signed-diff(bestFreq,desired))`.

While it is true that `signed - unsigned == *un*signed`, that *un*signed is interpreted as a *signed* by `abs()` so "it just works."

So, as far as I can see, this patch is not correct.  The abs needs to be left in.  For best correctness, the `uint32_t` subtrahend should be cast to a `int32_t` to be completely legit, but functionally there's no issue right now.

mcspr: > hat unsigned is interpreted as a signed by abs() so "it just works."

True, abs() is actually from libc. aghr, sorry. I just mentioned the `#include <cmath>` earlier in the PR, which does silently remove the Arduino.h's abs definition in the same header. Now it does make sense where from the two `abs aXX aXX` assembly lines come from (and why it did not break anything after all this time. funnily enough, arduino-esp32 SPI HAL code shares this exact code snippet).

Not to go on a loop... We already defensively protect value from overflow + it never goes over magic 0x4c4b400 ESP8266_CLOCK, which is already less-than both INT_MAX and UINT_MAX. I already mentioned the change from calFreq to bestFreq goes through the same less-than check with (int32_t)freq, so both sutractions are always either 0 or more than 0. 

Nonetheless, subtraction is indeed signed, so wide array of casts should be replaced with a singled named cast somewhere above.

*But, ff we are protecting from future code changes... I'd rather not predict the future too much.*

Issue #5627, opened 2019-01-17
Why is default softap ip 192.168.244.1 ?


tablatronix:
It seems a bunch of people seem to think this is the default ip for softap mode, when it is not.

192.168.4.1 is the default, but it does look like this IS the default ip by lib when ip config fails.

Any ideas why, and why not use the same ip as SDK

https://github.com/esp8266/Arduino/blob/dd6333ee8b26dd9cc82dea252ba150e85c0ebb11/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp#L173

devyte:
No idea. The referenced code was introduced by @Links2004 and dates back to Jan 2016.

If we change it, it would have to be in 3.0.0.

tablatronix:
I need to PR this..

PR #7363, opened 2020-06-08
Fix the bad address of the the AP if the settings are wrong

dumarjo:
Hi,
This should fix the problem seen from time to time on our device. The default IP should be 192.168.4.1 and not 192.168.244.1

This should fix this issue: https://github.com/esp8266/Arduino/issues/5627


devyte:
This is a long standing bug. We don't support user reliance on a bug. Moving to 2.7.2.

devyte:
@dumarjo thanls for this. Also, shame one, because I know I've seen that line in the past, and wondered about it.

PR #7367, opened 2020-06-11
User requested fix: be less restrictive in the use of GPIOs.

dok-net:
Release 6.8.3 of EspSoftwareSerial.


dok-net:
@d-a-v I know you are going to ask this:
> Is it stable enough, or is it an urgent / bugfix commit that needs to be marked for 2.7.2 ?
Or should we wait for after 2.7.2 ?

And I'll reply: I am sorry, a user explained a compelling use case that requires this change, just this morning. Otherwise, there's no activity in EspSoftwareSerial, so I'm PRing ASAP case by case.

PR #7369, opened 2020-06-11
Add a "-- CUT HERE --" marker to exception output

earlephilhower:
Users aren't reporting/seeing the actual PC of failures because they
are only copying the >>stack<< into the decoder and not the Exception
line.  Add an ALL CAPS "CUT HERE FOR EXCEPTION DECODER" line to the
output to make it blindingly obvious to include the full crashdump
both in bug reports as well as exception decodes.

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -92,6 +92,18 @@ static void ets_printf_P(const char *str, ...) {
     }
 }
 
+static void cut_here() {
+    ets_putc('\n');
+    for (auto i = 0; i < 15; i++ ) {
+        ets_putc('-');
+    }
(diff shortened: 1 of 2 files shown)


earlephilhower:
An example of the modified output:
````

Exception (29):
epc1=0x4020104f epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffe00 end: 3fffffc0 offset: 0190
3fffff90:  3fffdad0 00000000 3ffee304 40201049  
3fffffa0:  feefeffe feefeffe 3ffee32c 40201970  
3fffffb0:  feefeffe feefeffe 3ffe84e0 40100b85  
<<<stack<<<

````

The decoder works w/ or w/o the `---` lines (it ignores what it can't understand.).

PR #7375, opened 2020-06-13
WiFiUDP::endPacketMayRetry()

d-a-v:
This method will allow to later resend an UDP packet when an issue occurs, like when sending too fast.

The added example is demonstrated below.
`q` is sent, then replied back 20 times. every `!` means a failed send and an attempt to resend.
Serial console:
```
Received packet of size 2 from 10.0.1.7:60116
    (to 10.0.1.163:8888, free heap = 41776 B)
Contents:
q

0 
1 
2 
3 
4 
5 
6 
7 
8 ! ! ! ! ! 
9 ! 
10 ! 
11 ! ! ! ! ! ! ! ! ! ! ! ! 
12 ! 
13 ! 
14 ! 
15 ! 
16 ! 
17 ! 
18 ! 
19 ! ! ! 
```

`nc / netcat` side:
```
q   <= me sending
Reply 'q' #0 heap=41776  <= replies
Reply 'q' #1 heap=41616
Reply 'q' #2 heap=41456
Reply 'q' #3 heap=41296
Reply 'q' #4 heap=41136
Reply 'q' #5 heap=40976
Reply 'q' #6 heap=40816
Reply 'q' #7 heap=40656
Reply 'q' #8 heap=40496
Reply 'q' #9 heap=40264
Reply 'q' #10 heap=40264
Reply 'q' #11 heap=40264
Reply 'q' #12 heap=40264
Reply 'q' #13 heap=40264
Reply 'q' #14 heap=40264
Reply 'q' #15 heap=40264
Reply 'q' #16 heap=40264
Reply 'q' #17 heap=40264
Reply 'q' #18 heap=40264
Reply 'q' #19 heap=40264
```


d-a-v:
Reasons for this PR.

Arduino UDP API `::endPacket()` always discards data to be sent even when there is an error while effectively trying to send it (for some reason like sending too fast / output queue still full - which happens with UDP because there is no integrated flow control like in TCP).

Sometimes the building process to obtain data to be sent is complex, like it is in our mDNS implementation: resending a discarded data is difficult to handle.

Our internal API in `UdpContext` was following Arduino API (by always discarding). This PR updated it to make discarding optional (there was `::send()`, there are now `::trySend()` and `::cancelBuffer()`. Now `::send()` uses them.

Included in the PR the counterpart for Arduno API's `WiFiUDP` and an example showing how to use.

libraries/ESP8266WiFi/src/WiFiUdp.cpp
```diff
@@ -176,6 +176,24 @@ int WiFiUDP::endPacket()
     return (_ctx->send()) ? 1 : 0;
 }
 
+bool WiFiUDP::endPacketMayRetry()
```
devyte, line 179: As I tried to explain in our internal discussion, there is an interaction between this new method and scheduled functions, if both try to use the same Udp object. The problem happens because this new method blocks while waiting for the pbuf to progress enough to make room, and during that time a scheduled function could get triggered and try to beginPacket().
As you pointed out, this is a reentrancy case, and the current Udp implementation doesn't support reentrancy. However, the case can be hit easily by a user, as I try to show in the gist [there](https://gist.github.com/devyte/2c7c5404294f93c20c841179e88291cd). The example is trivial, and it is not clear why it doesn't work as expected.
To mitigate the problem, I think the beginPacket() method should detect whether there is data pending in the UdpContext tx buffer, and if so error out. That can mean either return false (up to the user to check the return value of beginPacket()), or fail an assert (programming error, so crater on purpose with some debug message).
I'm not sure if the detection can be done within beginPaket() itself. but if not, I think it would be valid to detect inside UdpContext::connect(), something along the lines of:
```cpp
if(tx_buffer_has_data())
  return false;
```
at the top.

d-a-v:
Even offering a `WiFiUDP::beginPacket()`'s return value would not help since an average arduino user won't use it. Those who are aware of reentrancy issues know they can't use the same udp instance in the main loop and a scheduled function. They would use a polledTimeout-like test for the ping service. A user who would open an issue with your gist would be told about the reentrancy issue which is inherent to this particular UDP API (begin, later write, later send).

UdpContext needed changes will be included in the incoming new mDNS implementation, *without* changing the exposed WiFiUDP API. Advanced users will be able to directly use UdpContext (like it is currently done and soon improved in the mDNS library).

Closing this PR because UDP-sending API is considered as being too dangerous.

Issue #5530, opened 2018-12-20
Calling round() inside print() result differs between v2.4.2 and v2.5.0-beta2


Paraphraser:
### Basic Infos

- [Y] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [Y] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [N] I have tested that the issue is present in current master branch (aka latest git). I simply upgraded the IDE to 2.5.2-beta2 and then compared it with 2.4.2 behaviour and observed an unexpected and unwelcome difference.
- [Y] I have searched the issue tracker for a similar issue.
- [N/A] If there is a stack dump, I have decoded it.
- [Y] I have filled out all fields below.

#### Platform

- Hardware:				[ESP8286]
- Core Version:			[2.5.0-beta2 & 2.4.2]
- Development Env:		[Arduino IDE]
- Operating System:		[MacOS]

### Settings in IDE

- Module:  			[Wemos D1 mini r2]
- Flash Mode:			[dio - as far as I am aware]
- Flash Size: 		       [4MB/1MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[whatever the IDE does at the end of a download]
- Flash Frequency:	[no menu item for this - can't answer]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600)

### Problem Description

Under version 2.4.2, round() appears to return an integer quantity. Under 2.5.0-beta2, round() appears to return a real quantity. I am inferring this from a change in print() behaviour.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <Arduino.h>

void setup() {

    Serial.begin(115200);
    while (!Serial);
    delay(100);
    Serial.println();
    Serial.println();

}

void loop() {

    // wait a while
    delay(1511);

    // precalculate upTime to the nearest second
    unsigned long upTime = round(millis()/1000.0);

    // display upTime with an inline calculation
    Serial.print("Up time: inline = ");
    Serial.print(round(millis()/1000.0));
    Serial.print(", pre-calculated = ");
    Serial.println(upTime);
    
}

```

### Debug Messages

Output from compiling under 2.4.2 libraries. Note ABSENCE of decimal places for "inline =" values. This is the expected behaviour.

```
[...]

PR #7377, opened 2020-06-14
Allow test framework to use cores/esp8266/Arduino.h directly

### LINKED ISSUE #7359: Host tests: different Arduino.h contents
### Basic Infos

#### Platform

- Hardware:					Host tests
- Core Version:			https://github.com/esp8266/Arduino/commit/e70092ca3e7a974f268521d85833ac58187f4649
- Development Env:		esp8266 Arduino Mock

### Problem Description

As noticed in the #6294 , [host tests Arduino.h](https://github.com/esp8266/Arduino/blob/master/tests/host/common/Arduino.h) should be in sync with the one used to build the real firmware.

### <s>[MCVE](https://stackoverflow.com/help/mcve)</s>

This does not cause any immediate issues, just introduces different build settings for host. Some (but not all) examples are:
- boolean typedef
- implicit `#include <cmath>`
- abs redefinition

This may cause more problems in the future, perhaps both headers should use the common source and adjust platform-specific settings.

mcspr:
resolve #7359 , @d-a-v

- Removes 'tests/host/common/Arduino.h' in favor of 'cores/esp8266/Arduino.h'
- Clean-up math imports, comment some reasoning behind define / undef
- Add a small example that does not build with the current master

Some minor stuff
- Esp.h change is cosmetic, we might as well use class body instead of this weird forced inline definition
- I also noticed malloc debug include at the end
```cpp
#ifdef DEBUG_ESP_OOM
// reinclude *alloc redefinition because of <cstdlib> undefining them
// this is mandatory for allowing OOM *alloc definitions in .ino files
#include "umm_malloc/umm_malloc_cfg.h"
```
This does not cause any issues since mock does not set DEBUG_ESP_OOM, but it is unclear why it is outside of guards and whether it should be some place else ('core_esp8266_... .h'?)


cores/esp8266/Arduino.h
```diff
@@ -233,25 +240,22 @@ const int TIM_DIV265 __attribute__((deprecated, weak)) = TIM_DIV256;
...
 #include <cmath>
-#include <pgmspace.h>
-
-#include "WCharacter.h"
-#include "WString.h"
-
-#include "HardwareSerial.h"
-#include "Esp.h"
-#include "Updater.h"
-#include "debug.h"
```
mcspr, line 248: 'debug.h' seems to be internal header?

cores/esp8266/Esp.h
```diff
@@ -168,21 +169,14 @@ class EspClass {
         uint32_t random() const;
 
 #ifndef CORE_MOCK
-        inline uint32_t getCycleCount() __attribute__((always_inline));
+        uint32_t getCycleCount() __attribute__((always_inline)) {
+            return esp_get_cycle_count();
+        }
```
mcspr, line 174: I wonder if attribute is required?

cores/esp8266/libc_replacements.cpp
```diff
@@ -127,6 +127,7 @@ void _exit(int status) {
     abort();
 }
 
+int atexit(void (*func)()) __attribute__((weak));
```
mcspr, line 130: I think this was mentioned before, weak definition will apply to every symbol otherwise. Placing it here we specify that libc_replacements.o version is weak

cores/esp8266/Arduino.h
```diff
@@ -233,25 +240,22 @@ const int TIM_DIV265 __attribute__((deprecated, weak)) = TIM_DIV256;
...
 
 using std::min;
 using std::max;
 using std::isinf;
 using std::isnan;
 
+// these are important, as we may end up using C versions otherwise
+using std::abs;
+using std::round;
```
mcspr, line 257: Meaning, right now anyone using both of these uses C stdlib versions, undef + define never did anything because of the cstdlib include in algorithm

cores/esp8266/Arduino.h
```diff
@@ -127,21 +128,26 @@ void timer0_isr_init(void);
...
 #ifdef abs
 #undef abs
+#define abs(x) ((x)>0?(x):-(x))
 #endif
 
-#define abs(x) ((x)>0?(x):-(x))
+#ifdef round
+#undef round
+#define round(x) ((x)>=0?(long)((x)+0.5):(long)((x)-0.5))
+#endif
```
mcspr, line 143: :oops: this is actually wrong as there is no `define round`, missed when editing.
meaning, right now this pr only uses std::round and master uses macro

libraries/esp8266/examples/math/math.ino
```diff
@@ -0,0 +1,30 @@
+/*
+  Small math example, checking whether we can support various Arduino math operations.
```
devyte, line 2: I strongly advise rewriting this as a device side test instead of an example. You can then add explicit checks for specific functionality.

cores/esp8266/Esp.h
devyte: It is needed. I think it was to force inline when optimizations are turned off. Whatever the case was, if it is not inlined, and this is called e. g. from an ISR, you'll get a crash.

cores/esp8266/libc_replacements.cpp
devyte: I don't understand this comment. Please explain in detail.

dok-net:
@devyte This is another - from my POV, excuse the language - arbitrary PR that at the time of writing this gets fast-tracked to 3.0.0 past some of my work, thus both ignoring that and making merging it hard/impossible (finite resources for merging).

devyte:
Fixes #5530 as well, if the abs and round changes are ok (I haven't checked in detail)

libraries/esp8266/examples/math/math.ino
mcspr: Ah. Meaning, the right place is `tests/device/test_sw_...`?
Only issue is (as far as I can understand the makefile MOCK=0 / MOCK=1), it gets picked up only by the host gcc and it will not be compiled by the xtensa toolchain too? The general issue was different code, so it needs to be able to be built by both.

cores/esp8266/libc_replacements.cpp
mcspr: `__attribute__((weak));` when in header is applied to every file including Arduino.h and implementing the function, so every atexit becomes weak and we depend in which order they are linked.

libraries/esp8266/examples/math/math.ino
d-a-v: Device tests can always be compiled to run on device.
Some of them can also be compiled by both host and device compilers.
Current convention is to call them `test_sw_<name>`.

mcspr: Sure. Tried it out both with g++ 9.3.1 on host and running PR tree on device.
b/c gcc 10 pr is merged, it is less relevant to the testing, but it is still a useful patch

d-a-v: No, CI currently does not run device tests.
But we could add device tests that can also be run in the test framework.

d-a-v:
from OP:
> I also noticed malloc debug include at the end

This is history. Files and dependencies have changed from that time.
Maybe it can be moved in a better place like you suggest.

cores/esp8266/Arduino.h
```diff
@@ -218,28 +216,31 @@ void optimistic_yield(uint32_t interval_us);
 } // extern "C"
 #endif
 
+// undefine stdlib's definitions when encountered, provide abs that supports floating point for C code
+#ifndef __cplusplus
+#undef abs
+#define abs(x) ((x)>0?(x):-(x))
```
d-a-v, line 222: For sanity I would
```
#define abs(x) ({ __typeof__(x) _x = (x); _x > 0 ? _x : -_x; })
#define round(x) ({ __typeof__(x) _x = (x); _x >= 0 ? (long)(_x + 0.5) : (long)(_x - 0.5); })
```

dok-net:
In `libraries/ESP8266WiFi/src/BearSSLHelpers.cpp`, `libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp`, `libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp`, `tests/device/test_sw_http_client/test_sw_http_client.ino` there are

`#if !CORE_MOCK`

instead of the (canonical)

`#ifndef CORE_MOCK`

Could this be fixed/streamlined in the course of this PR as well?

libraries/esp8266/examples/math/math.ino
mcspr: As noted above, we don't really even need to run it, I am not really keen on it beeing a test at all. fwiw, I have tried a small example having *only* `static_assert(...condition..., "");`, because everything is already known at compile time.

Can you please point to the correct place, then? Only requirement is that something tries to build a .cpp with both xtensa and host's compiler.

mcspr:
> While testing tests/device/test_sw_arduino_math_overrides, I had to add #include <cerrno> in tests/host/common/{litlefs,spiffs}_mock.cpp. Maybe because my host gcc version is 10.

will update, seeing this too (was jumping between gcc-9 and gcc-10 machines...)
```
common/littlefs_mock.cpp:34:1: note: ‘errno’ is defined in header ‘<cerrno>’; did you forget to ‘#include <cerrno>’?
   33 | #include <unistd.h>
  +++ |+#include <cerrno>
   34 | #include "flash_hal_mock.h"
...
common/spiffs_mock.cpp:29:1: note: ‘errno’ is defined in header ‘<cerrno>’; did you forget to ‘#include <cerrno>’?
   28 | #include <unistd.h>
  +++ |+#include <cerrno>
   29 |
...
```

> In libraries/ESP8266WiFi/src/BearSSLHelpers.cpp, libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp, libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp, tests/device/test_sw_http_client/test_sw_http_client.ino there are
>
> #if !CORE_MOCK
>
> instead of the (canonical)
>
>#ifndef CORE_MOCK
>
>Could this be fixed/streamlined in the course of this PR as well?

:shrug:
this would also require changing mock.h to a plain `#define CORE_MOCK` so that does not happen in the future
besides, that that is only cosmetic and might as well be added after?

bearssl(helpers|client) needs to change
littlefs comment is definitely wrong
http_client test does not need ifndef'ed block at all because axtls is removed

libraries/esp8266/examples/math/math.ino
d-a-v: It's in `test/device`. There is for example `test_time` which may not run with emulation, and `test_sw_http_client` which runs on both (it was first called `test_http_client` then later renamed after proving to also work on host).

cores/esp8266/Arduino.h
devyte: Why are these defined at all? I think eboot.c and umm is the only C code in our core, and Arduino.h isn't included in either AFAIK.

mcspr: Not the Core, these are for users' .c code

cores/esp8266/umm_malloc/umm_malloc_cfg.h
```diff
@@ -17,9 +17,10 @@
 #include <stdint.h>
 #include <stddef.h>
 #include <stdbool.h>
-#include <debug.h>
+
 #include <pgmspace.h>
-#include <esp8266_undocumented.h>
+#include "../debug.h"
+#include "../esp8266_undocumented.h"
```
mcspr, line 23: It is sort of hard to remember why I cared about debug.h 2 month ago...

Having a PIO project with debug.h file in the tree, including Arduino.h caused it to include local debug.h instead of the one in the core/esp8266. I suppose a correct solution would be to either:
- use a sub-folder for debug.h e.g. include \<esp8266/debug.h\>)
- prefix it like the other headers, esp8266_debug.h

I removed other references as those aren't really necessary. This one, however, very quickly breaks when OOM debug is enabled

d-a-v: I wouldn't mind if you rename it.

mcspr: I'll prepare a separate PR for that, since it is not necessary here.

d-a-v:
> Minimum for -std=c17 is 8.x.0, but my understanding is we would need to call it via gcc-{8,9}, same as g++-{7,8,9}

It seems gcc-8 is bundled with ubuntu-latest that we use in CI.

Maybe that could be useful
```
diff --git a/tests/host/Makefile b/tests/host/Makefile
index 1c534356..b1785d83 100644
--- a/tests/host/Makefile
+++ b/tests/host/Makefile
@@ -11,6 +11,11 @@ DEFSYM_FS ?= -Wl,--defsym,_FS_start=0x40300000 -Wl,--defsym,_FS_end=0x411FA000 -
 
 MAKEFILE = $(word 1, $(MAKEFILE_LIST))
 
+ifeq ($(shell g++-8 -v; echo $$?),0)
+CC  = gcc-8
+CXX = g++-8
+endif
+
```

d-a-v:
@mcspr do you intend to fix these conflicts ?

mcspr:
Done.
What's holding this, anyway?

devyte:
Me, I need to re-review.
Don't forget to please mark resolved comments as, well, resolved :) 
I expect to take a look later today.

devyte (approved):
Apologies for taking so long!

earlephilhower:
@mcspr we now have a merge conflict in Arduino that I can't fix via the Web GUI.  Can you please do a manual merge and re-push so we can pull this into master?

mcspr:
Sure, will do (as sooon as I have more than 5min at the PC...)
a3281fe2f3436a3bce9604033344ae483bf5d0ed seems to be the only change, but I am not yet sure what that change is supposed to be doing

d-a-v:
Merge error is indeed:
> CONFLICT (modify/delete): tests/host/common/Arduino.h deleted in HEAD and modified in master. Version master of tests/host/common/Arduino.h left in tree.
```
@@ -270,3 +270,10 @@ extern "C" void configTime(long timezone, int daylightOffset_sec,
 #include "pins_arduino.h"
 
 #endif /* Arduino_h */
+
+#if __cplusplus
+#include <WString.h>
+#include <map>
+#define MOCKARGS 1
+extern std::map<String,String> mockArgs;
+#endif
```

After inspection, `MOCKARGS` is unused anywhere, and `mockArgs` is only used in `ArduinoMain.cpp`. These are potential helpers, unused in the core, that I added to allow mocking with parameters without recompilation in user (mocking) sketches.
I think you can ignore/remove them in your PR. They shall be shared in another emulation-specific header file.

mcspr:
> These are potential helpers, unused in the core, that I added to allow mocking with parameters without recompilation in user (mocking) sketches.

Thanks for clarifying. But, any reason why not just use {get,set}env instead?

d-a-v:
>  But, any reason why not just use {get,set}env instead?

Not a single tiny one :blush:
I will remove this key/value facility asap

PR #7388, opened 2020-06-17
add ESP8266WebServerTemplate<ServerType>::enableCORS(bool _ec) like E…

Mjxkill:
add ESP8266WebServerTemplate<ServerType>::enableCORS(bool _ec) like ESP32 arduino plateform

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
@@ -44,8 +44,7 @@ static const char Content_Length[] PROGMEM = "Content-Length";
 
 template <typename ServerType>
 ESP8266WebServerTemplate<ServerType>::ESP8266WebServerTemplate(IPAddress addr, int port)
-: _corsEnabled(false)
-, _server(addr, port)
+: _server(addr, port)
 , _currentMethod(HTTP_ANY)
 , _currentVersion(0)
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -45,7 +44,8 @@ static const char Content_Length[] PROGMEM = "Content-Length";
 
 template <typename ServerType>
 ESP8266WebServerTemplate<ServerType>::ESP8266WebServerTemplate(IPAddress addr, int port)
-: _server(addr, port)
+: _corsEnabled(false)
```
d-a-v, line 47: `_corsEnabled` needs to be last in the initialization list because you put it last in members (CI is not happy).

Also, for clarity, can you replace `_ec` by `enable` ?

Mjxkill: yes you're right ! i will correct this... it was my first pull :-) how to correct ? will I post a new pull request ?

earlephilhower: Just commit and push to the same branch and it will automatically pick it up here.

Mjxkill: ok Done

PR #7401, opened 2020-06-23
httpclient: fix error meaning

d-a-v:
Changed HTTPClient error name and log from "Refused" to "Failed" (both macro and string).
"refused" can lead to false assumptions.
"refused" macro is kept for backward compatibility. This is not a breaking change.

*update*:

A TCP-"connection refused" *is an answer* from peer telling that there is nothing listening on this tcp port (when a firewall is in the way, the error becomes a timeout error).
But this httpclient error can be returned even when the question cannot be asked because it can be caused by a local error (wifi not connected for example).

For the correct distinction between a local error and a real "connection refused", WiFiClient API has to be updated with an additional function returning the last real error (returned by lwIP (*refused*, *unreachable*, ...) or by local hardware (*abort*, ...)).


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
```diff
@@ -48,7 +48,8 @@
 #define HTTPCLIENT_DEFAULT_TCP_TIMEOUT (5000)
 
 /// HTTP client errors
-#define HTTPC_ERROR_CONNECTION_REFUSED  (-1)
+#define HTTPC_ERROR_CONNECTION_REFUSED  (-1) // backward compatibility
```
earlephilhower, line 51: The third answer ("You could make sure that these macros would expand to something that would include a expression that would trigger the attribute((deprecated)) warning...." ) here
https://stackoverflow.com/questions/2681259/how-to-deprecate-a-macro-in-gcc
shows a little convoluted way of making it work.  Did not try myself, though.

Also, might be possible to make a single `typedef __attribute__ (( deprecated )) enum {HTTPC_ERROR_CONNECTION_REFUESD=-1};` ?

d-a-v: what about `constexpr int HTTPC_ERROR_CONNECTION_REFUSED = -1 __attribute__((deprecated));`
and push milestone to v3 (because of `deprecated`) ?

earlephilhower: Ah, was thinking C and not C++, but this is a class so you're not using it in C anyway.  D'oh!

Why not deprecate now?  If we can deprecate SPIFFS in 2.7.1, doing a single, poorly named constant in 2.7.2 seems relatively OK.

PR #7420, opened 2020-07-01
Ability to set connect timeout in WifiMulti->run

MikaelBertze:
WifiMulti->run has a built in timeout when connecting set to 5s. In some cases this is not enough time to set up the connection.

I noticed that my ESP-boards quite often failed to connect to my wifi if my repeater mode AP was selected as the strongest SSID. Don't know the reason for the slow connection setup, but my other devices connect without issue.

For Esp32, the 5s timeout was added some time ago based on this project (https://github.com/espressif/arduino-esp32/pull/1220/files). They implemented it a bit different where the timeout could be altered and with a default set to 5s.

I made a PR that changes the API to the same as for Esp32.


MikaelBertze:
> Thanks !
> For the sake of clarity I would rename the variable `connectTimeoutMs`

Done!

PR #7435, opened 2020-07-07
WIP - Remove AXTLS from core

earlephilhower:
axTLS has been deprecated for over a year now, so finally remove it from
the code.

Rename BearSSL* examples to plain WiFiSecure*


earlephilhower:
This is turning into a hydra.  Lots of interdependencies sneaking up.  Closing for now while I think of a less intrusive way of doing this...

PR #7437, opened 2020-07-08
Remove axTLS from code and documentation

earlephilhower:


jeroenst:
Due to this change I can't use version 3.x.x with ESPAsyncTCP and I have to keep using 2.x.x.

PR #7459, opened 2020-07-15
webserver hook: allow to handle external http protocol

d-a-v:
`webserver.addHook()` addition, allowing to handle external http protocols **on the same port**, at least like WebDAV and WebSocket.

Tested with ESPWebDAV [branch v3](https://github.com/d-a-v/ESPWebDAV/tree/v3).

A hook function can use URI and http method to check if it would handle the http request.
There are 4 possible return values:
- CLIENT_REQUEST_CAN_CONTINUE
  Hook function is not interested
- CLIENT_REQUEST_IS_HANDLED
  Hook function took and handled the current http request, webserver must skip it but can handle the next request
- CLIENT_MUST_STOP
  Issue with http protocol, or general malfunction, or connection is done. Connection should be closed.
- CLIENT_IS_GIVEN
  The tcp connection has been given to someone else, it must not be closed neither anymore handled
  (this is typically for websockets)

Example of a hook function: [call on webserver](https://github.com/d-a-v/ESPWebDAV/blob/7fa7d8c6638c39d5ee77cfaff297370f7b22b64c/examples/Hooked/Hooked.ino#L102), [function itself](https://github.com/d-a-v/ESPWebDAV/blob/7fa7d8c6638c39d5ee77cfaff297370f7b22b64c/WebDav4WebServer.cpp#L4).


libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -35,6 +35,12 @@
 #define DEBUG_OUTPUT Serial
 #endif
 
+#ifdef DEBUG_ESP_HTTP_SERVER
```
earlephilhower, line 38: Could you convert the other `DEBUG_OUTPUT.printf()`s to your new macro as a bit of cleanup?

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -192,11 +196,28 @@ class ESP8266WebServerTemplate
 
   static String responseCodeToString(const int code);
 
+  void addHook (Hook_f hook)
+  {
```
earlephilhower, line 200: Might be a bit too anal, but can we keep the cuddle consistent in the file?  Everything else has opening bracket on the fcn. line.

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -80,6 +81,9 @@ class ESP8266WebServerTemplate
   using ClientType = typename ServerType::ClientType;
   using RequestHandlerType = RequestHandler<ServerType>;
   using WebServerType = ESP8266WebServerTemplate<ServerType>;
+  using ContentType_f = std::function<String(const String&)>;
+  enum ClientFuture_e { CLIENT_REQUEST_CAN_CONTINUE, CLIENT_REQUEST_IS_HANDLED, CLIENT_MUST_STOP, CLIENT_IS_GIVEN };
+  using Hook_f = std::function<ClientFuture_e(const String& method, const String& url, WiFiClient* client, ContentType_f contentType)>;
```
earlephilhower, line 86: I think we're still pulling all std::functions in to IRAM in the linker which might make this a problem.  Could you check on your own build where the webDAV hook ends up?

libraries/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -80,6 +81,9 @@ class ESP8266WebServerTemplate
   using ClientType = typename ServerType::ClientType;
   using RequestHandlerType = RequestHandler<ServerType>;
   using WebServerType = ESP8266WebServerTemplate<ServerType>;
+  using ContentType_f = std::function<String(const String&)>;
```
earlephilhower, line 84: Another anal thing, we're not using Hungarian notation on other types/usings elsewhere in the code.

d-a-v:
@earlephilhower I addressed your review, I still need to check whether lambdas/functionals are in flash (not iram) thanks to your snippet
```
    uint32_t addr;
    __asm__ ("movi %0, ." : "=r" (addr) : : );
    // 401x:iram 402x:flash
    Serial.printf("%08x\n", addr);
```

libraries/ESP8266WebServer/src/ESP8266WebServer.h
d-a-v: Hooks are in FLASH (not IRAM). It is verified with your snippet and updated HelloServer example.

d-a-v:
Example is provided in [`Helloserver.cpp`](https://github.com/d-a-v/Arduino/blob/webhook/libraries/ESP8266WebServer/examples/HelloServer/HelloServer.ino#L89L155).

libraries/ESP8266WebServer/examples/HelloServer/HelloServer.ino
```diff
@@ -86,6 +86,74 @@ void setup(void) {
...
   server.onNotFound(handleNotFound);
 
+  /////////////////////////////////////////////////////////
+  // Hook examples
+
+  server.addHook([](const String & method, const String & url, WiFiClient * client, ESP8266WebServer::ContentTypeFunction contentType) {
+    (void)method;
+    (void)url;
+    (void)client;
+    (void)contentType;
```
mcspr, line 96: Minor style nitpick - by omitting argument name in the signature, there would be no need for these 4 lines.
There is also [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) / `[[gnu::unused]]` for local declaration, since now we use c++17. But that one is mostly for stack variables

Issue #7458, opened 2020-07-15
OTA of large files results in device hangs due to bootloader compare error


arendst:
After a recent core change Tasmota fails to load binaries over 560k due to new bootloader functionality.

After the file is loaded this is reported:

```
load 0x4010f000, len 3664, room 16 
tail 0
chksum 0xee
csum 0xee
v39c79d9b
@cp:0
cmp:
 ets Jan  8 2013,rst cause:3, boot mode:(3,7)

ets_main.c 
```

It seems the bootloader fails to compare the copied binary with the uploaded binary (for obvious reasons as part of the uploaded file are being overwritten by the bootloader copy routine).

A recent commit 51daecc236fbff0d1ed30a9c63a3af6e8b5c7392 introduces this cmp functionality not present in previous bootloader versions and leads to the device hangs.

Pls remove the compare functionality as it breaks OTA functionality

EDIT: Using core 2.7.1 the OTA uploads work fine (as it has no compare functionality)

earlephilhower:
> It seems the bootloader fails to compare the copied binary with the uploaded binary (for obvious reasons as part of the uploaded file are being overwritten by the bootloader copy routine).

Can you please explain that a little more?  At first glance, the code looks like it on a COPY_RAW (i.e. OTA) it writes then verifies once.  After that, the verify is never executed again (it's inside the `if (cmd==COPY_RAW)` block).

earlephilhower:
Oh poop, I think I see what you're saying.

You have a firmware of 100K and a 1MB flash chip.  You OTA a 200K app placed at 100K-300K and copied down to 0-200K...overwriting the 1st 100K of the original OTA space.  Is that the general gist of the issue?

arendst:
From what I see here:
```
    if (cmd.action == ACTION_COPY_RAW) {
        ets_putc('c'); ets_putc('p'); ets_putc(':');

        ets_wdt_disable();
        res = copy_raw(cmd.args[0], cmd.args[1], cmd.args[2], false);
        ets_wdt_enable();

        ets_putc('0'+res); ets_putc('\n');

        // Verify the copy
        ets_putc('c'); ets_putc('m'); ets_putc('p'); ets_putc(':');
        if (res == 0) {
            ets_wdt_disable();
            res = copy_raw(cmd.args[0], cmd.args[1], cmd.args[2], true);
            ets_wdt_enable();
            }
```
[...]

PR #7468, opened 2020-07-18
Comment out verify step in eboot.c

devyte:
Meant for #7458 , but still requires a recompiled eboot.elf.


devyte:
Testing by @Tech-TX shows ok.
ToDo, but in a different PR and target: implement some alternate form of verification, e. g. with hash calc.

PR #7486, opened 2020-07-25
CVE-2020-12638 workaround for WPA downgrade attack

earlephilhower:
When connected to an encrypted (WEP/WPA) router, a rogue packet can
cause the ESP8266 WiFi stack to drop to an unecrypted rogue network of
the same SSID.

Handle this by dropping the WiFi connection immediately and reconnecting
to the stored WPA/WEP network requested by the application, whenever the
AUTHMODE changes to OPEN from a secured mode.

https://lbsfilm.at/blog/wpa2-authenticationmode-downgrade-in-espressif-microprocessors
for more details.


s00500:
I could confirm that this successfully mitigates an attack as a workaround, hopefully the arduino core can be upgraded to the latest NONOS master and this workaround can be removed in the future

Thanks for the fast response!

earlephilhower:
Thanks @s00500 !

PR #7491, opened 2020-07-29
Fixed PUYA flash write buffer alignment

drzony:
This merge fixes flash writing on PUYA flashes when write size is not a multiple of 4.
This bug prevents updating via Arduino OTA when using gzipped image.

### DIFF
--- a/cores/esp8266/Esp.cpp
+++ b/cores/esp8266/Esp.cpp
@@ -697,7 +697,7 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
         } else {
             bytesLeft = 0;
         }
-        size_t bytesAligned = (bytesNow & 3) ? ((bytesNow & ~3) + 4) : bytesNow;
+        size_t bytesAligned = (bytesNow + 3) & ~3;
         rc = spi_flash_read(pos, flash_write_puya_buf, bytesAligned);
         if (rc != SPI_FLASH_RESULT_OK) {
             return rc;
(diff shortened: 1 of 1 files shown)


cores/esp8266/Esp.cpp
```diff
@@ -697,15 +697,16 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
         } else {
             bytesLeft = 0;
         }
-        rc = spi_flash_read(pos, flash_write_puya_buf, bytesNow);
+        size_t bytesAligned = (bytesNow & 3) ? ((bytesNow & ~3) + 4) : bytesNow;
```
earlephilhower, line 700: I think it's more commonly done as
````
         size_t bytesAligned = (bytesNow + 3)  & ~3; 
````

drzony: Yup, got a little brain freeze there 😄

yumkam:
FTR, this over-reads `data` by up to 3 bytes. Assuming `data` is 4-bytes aligned and does not point to some kind of "special" memory (mmio/etc), this should be practically harmless. But formally this is still undefined behavior and compiler is free to kill kittens and start WWIII.

drzony:
Thanks for the observation @yumkam , must have missed this somehow. Another pull request is on the way.

drzony:
As for the overreading I have not seen a flash that is not 4-byte aligned. All the functions for spi_flash ops have `uint32 *` as parameters.

PR #7496, opened 2020-08-01
Add synthetic IntegerDivideByZero exception

earlephilhower:
The ROM routine __divsi3 is called by code whenever a division is needed,
because there is no divide unit on the ESP8266 core.  When the divide
routine in ROM hits a div-by-zero case, it jumpt to an ILL(egal instruction)
at a fixed address which causes a HW exception 0 (IllegalInsnException).

In the postmortem dump, when an ILL exception is detected at this address
in ROM, convert it to a DivByZeroException for printout (6).

Divde by zero errors now print as follows:
````
Exception (6):
epc1=0x4000dce5 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>
...
<<<stack<<<
````

And will decode as follows:
````
Exception 6: IntegerDivideByZero: QUOS, QUOU, REMS, or REMU divisor operand is zero
PC: 0x4000dce5
EXCVADDR: 0x00000000

Decoding stack results
...
````


earlephilhower:
Unfortunately, the mod (%) operator intrinsic function is not in ROM, so the same trick can't be used as it will move around in flash depending on the code.

PR #7500, opened 2020-08-02
Clean up minor warnings from LGTM.com

earlephilhower:
LGTM (Semmie) is a tool, bought by GitHub last year, that conducts basic
linting tasks on code and HTML.

Clean up the warnings identified in the latest report:
https://lgtm.com/projects/g/esp8266/Arduino/?mode=list

No functionality should change, however this may fix some issues with
the python utilities not exiting properly on a Ctrl-C from the command
line.

### DIFF
--- a/cores/esp8266/Crypto.cpp
+++ b/cores/esp8266/Crypto.cpp
@@ -100,6 +100,7 @@ void *createBearsslHmac(const br_hash_class *hashType, const void *data, const s
 
 String createBearsslHmac(const br_hash_class *hashType, const uint8_t hashTypeNaturalLength, const String &message, const void *hashKey, const size_t hashKeyLength, const size_t hmacLength)
 {
+    (void) hashTypeNaturalLength;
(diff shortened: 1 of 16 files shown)


mcspr:
Re:
> No functionality should change, however this may fix some issues with
the python utilities not exiting properly on a Ctrl-C from the command
line.

ref. https://lgtm.com/rules/6780080/, KeyboardInterrupt *wont* trigger the `except Exception:`, but it did before:
```python
try:
    while True:
        pass
except KeyboardInterrupt:
    print("fu")
except Exception:
    print("foo")

print("bar")

# $ python3 script.py
# ^Cfu
# bar
# $
```
espota.py and elf2bin.py seems to be the only ones with code iside there
idk about espota right now though, but elf2bin change seems legit

earlephilhower:
> KeyboardInterrupt wont trigger the except Exception:, but it did before:

That's the idea.  Before Ctrl-C would get caught in the `except: pass`es, now it should stop the app completely as an unhandled exception.

PR #7501, opened 2020-08-03
Add valgrind mocked test to CI

earlephilhower:
Run valgrind on host mock example runs to catch more bugs in CI.  These
tests would have caught the problem in #7464 before users did.

Add a list of some randomly picked examples to run, and add an option to
run the loop exactly once in the host mock routine, so the test will
actually exit under valgrind.

### DIFF
--- a/tests/ci/host_test.sh
+++ b/tests/ci/host_test.sh
@@ -6,5 +6,18 @@ set -ev
 
 cd $TRAVIS_BUILD_DIR/tests/host
 
-make CI
+
+make -j2 ssl
+for i in ../../libraries/ESP8266WiFi/examples/WiFiClient/WiFiClient \
+	../../libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation \
+	../../libraries/ESP8266WebServer/examples/HelloServer/HelloServer \
+	../../libraries/SD/examples/Files/Files \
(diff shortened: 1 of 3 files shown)


tests/ci/host_test.sh
```diff
@@ -6,5 +6,18 @@ set -ev
...
-make CI
+
+make -j2 ssl
+for i in ../../libraries/ESP8266WiFi/examples/WiFiClient/WiFiClient \
+	../../libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation \
+	../../libraries/ESP8266WebServer/examples/HelloServer/HelloServer \
+	../../libraries/SD/examples/Files/Files \
+	../../libraries/LittleFS/examples/LittleFS_Timestamp/LittleFS_Timestamp \
+	../../libraries/LittleFS/examples/SpeedTest/SpeedTest ; do
+	make -j2 D=1 $i
```
d-a-v, line 17: In a 64-bits gcc-multilib environment, 32 bits binaries are generated by default but valgrind does not work with them. `FORCE32=0` disables that and 64 bits binaries are generated.
In a non-gcc-multilib 64-bits environment, everything is 64 bits.
Just in case, I would `make -j2 D=1 FORCE32=0 $i`

PR #7505, opened 2020-08-05
Add SerialEvent() callback to loop processing

### LINKED ISSUE #752: A small nothing left aside. serialEvent()
I, you guys seeem to forgot the serialEvent() trigger, that is required to be fully compatible with the basics example of the Arduino IDE.

I guess we could just program a trigger from within the program, you while you guys are at it, well here is the constructive remark ;)

Thanx :)
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26489482-a-small-nothing-left-aside-serialevent?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

earlephilhower:
Match the AVR SerialEvent implicit callback.  Callback is executed
in normal user mode, not IRQ, so standard processing can be uses.

Fixes #752 after 5 years. :)

### DIFF
--- a/cores/esp8266/HardwareSerial.cpp
+++ b/cores/esp8266/HardwareSerial.cpp
@@ -32,6 +32,14 @@
 #include "HardwareSerial.h"
 #include "Esp.h"
 
+
+// SerialEvent functions are weak, so when the user doesn't define them,
+// the linker just sets their address to 0 (which is checked below).
+// The Serialx_available is just a wrapper around Serialx.available(),
+// but we can refer to it weakly so we don't pull in the entire
+// HardwareSerial instance if the user doesn't also refer to it.
+void serialEvent() __attribute__((weak));
+
 HardwareSerial::HardwareSerial(int uart_nr)
     : _uart_nr(uart_nr), _rx_size(256)
 {}
@@ -162,6 +170,12 @@ size_t HardwareSerial::readBytes(char* buffer, size_t size)
 
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL)
 HardwareSerial Serial(UART0);
+
+// Executed at end of loop() processing when > 0 bytes available in the Serial port
+void serialEventRun(void)
+{
+  if (serialEvent && Serial.available()) serialEvent();
+}
 #endif
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL1)
 HardwareSerial Serial1(UART1);
--- a/cores/esp8266/HardwareSerial.cpp
+++ b/cores/esp8266/HardwareSerial.cpp
@@ -174,7 +174,9 @@ HardwareSerial Serial(UART0);
 // Executed at end of loop() processing when > 0 bytes available in the Serial port
 void serialEventRun(void)
 {
-  if (serialEvent && Serial.available()) serialEvent();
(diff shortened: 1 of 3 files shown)


cores/esp8266/HardwareSerial.cpp
```diff
@@ -162,6 +170,12 @@ size_t HardwareSerial::readBytes(char* buffer, size_t size)
 
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL)
 HardwareSerial Serial(UART0);
+
+// Executed at end of loop() processing when > 0 bytes available in the Serial port
+void serialEventRun(void)
+{
+  if (serialEvent && Serial.available()) serialEvent();
```
devyte, line 177: Single line if

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -195,6 +195,7 @@ static void loop_wrapper() {
     }
     loop();
     loop_end();
+    if (serialEventRun) serialEventRun();
```
devyte, line 198: Single line if

cores/esp8266/HardwareSerial.cpp
devyte: I'm curious here: if serialEventRun is null, does this whole thing get optimized away?

earlephilhower: No, because `serialEventRun` is a linker-set variable and the GCC compile stage doesn't know about it.  It's one of the minor things LTO would help with (but LTO breaks everything else we're using w/weak functions...)

earlephilhower: I just disassembled a test w/no user `serialEvent`:

The `if (serialEventRun())` in `main.cpp` *is* still present:
````
40201120:       000000                  ill
...
40201144:       201110                  or      a1, a1, a1
40201147:       fffbc5                  call0   40201104 <__loop_end>
    if (serialEventRun) {
4020114a:       fff521                  l32r    a2, 40201120 <__loop_end+0x1c>
4020114d:       428c                    beqz.n  a2, 40201155 <_ZL12loop_wrapperv+0x31>
        serialEventRun();
4020114f:       fff401                  l32r    a0, 40201120 <__loop_end+0x1c>
40201152:       0000c0                  callx0  a0
    esp_schedule();
````

However, the `serialEventRun` function isn't present at all, so you just load that `00000000` and jump around the call each main loop_wrapper cycle.

**Comment by wolfbert** [comment]
Hi, just stumbled over this change and noticed that the code in [line 170 ff.](https://github.com/esp8266/Arduino/blob/d3eddeb5014d04aa9e61b97474ef98d0214eab1a/cores/esp8266/HardwareSerial.cpp#L170) ended up within the `#define` block. Seems strange.

```
#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL)
HardwareSerial Serial(UART0);

// Executed at end of loop() processing when > 0 bytes available in the Serial port
void serialEventRun(void)
{
  if (serialEvent && Serial.available()) {
    serialEvent();
  }
}
#endif
```

PR #7510, opened 2020-08-06
sys/pgmspace.h: Refactor inline asm

jjsuwa:
* Refactor asm tabs for `.S` cosmetics

* Avoid fixed regname (`a14`) allocation

* Remove unnecessary explicit cast and word masking


earlephilhower:
Howdy, @jjsuwa!  Looks good if a bit pedantic (dropping casts).

Let me do a couple checks on real HW just to verify the `a14` change.

In the meantime, this file comes from upstream in the newlib toolchain.  Can you put in the same PR there?  https://github.com/earlephilhower/newlib-xtensa

OTW, this change will be lost as soon as a new toolchain build is done.

PR #7514, opened 2020-08-07
New flash writing method with offset/memory/size alignment handling

drzony:
Fixes issues raised in https://github.com/esp8266/Arduino/pull/7491#issuecomment-670452122


yumkam:
I meant different issue: over-read of `data` (via `*ptr`), not over-read of flash.
(and I suspect `spi_flash_read`/`spi_flash_write` don't support `(size % 4) != 0`. Something like
[this](https://github.com/esp8266/Arduino/commit/b6bfe9907f5eae3f29817fe9358611270f4338b0)

drzony:
flash_write_puya_buf size is checked against bytesNow, which is enough, since it's aligned to flash sector size:
```cpp
  // Good alternative for buffer size is: SPI_FLASH_SEC_SIZE (= 4k)
  // Always use a multiple of flash page size (256 bytes)
  #define PUYA_BUFFER_SIZE 256
```

drzony:
@earlephilhower My commit does not change the over-read, it only changes the over-write, which in my opinion is a valid point. I see that you requested changes to my request, what should I change?

earlephilhower:
Thanks for the explanation.  I was going thru the original issue and thought the purpose was to not read past end-of-data.

This section of code has been very much trouble for us, so honestly what I'd need is a failing and passing valid test case on real XMC hardware with a couple folks' confirmations.  XMC just has some strange behavior which has caused something like 3 dot-releases so far, and stuff that looks like it should be fine often isn't when it meets that abomination of a flash chip.

drzony:
I have PUYA chip in my Sonoff S20, so I'm verifying each time before creating a pull request.
A minimal test case for me is just a gzipped OTA upgrade (via Arduino OTA). Before my original patch from #7491 when gzip did not have 4 byte alignment the update failed, since the last `(size % 4)` bytes were not written at all. Unfortunately my first patch wrote additional `(4 - (size % 4))` bytes to the flash, that is what this one fixes. I'm still modifying `(size + 3) & ~3` in temporary buffer (which is the fix for gzipped OTA), but only reading/writing `size`.

cores/esp8266/Esp.cpp
```diff
@@ -698,15 +698,15 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
             bytesLeft = 0;
         }
         size_t bytesAligned = (bytesNow + 3) & ~3;
-        rc = spi_flash_read(pos, flash_write_puya_buf, bytesAligned);
+        rc = spi_flash_read(pos, flash_write_puya_buf, bytesNow);
```
earlephilhower, line 701: The API docs seem to specify that the size needs to be quad-aligned, too ( https://www.espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf )

image

Sorry for the image paste.  The PDF wouldn't text-copy properly for me.

cores/esp8266/Esp.cpp
```diff
@@ -698,15 +698,15 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
         if (rc != SPI_FLASH_RESULT_OK) {
             return rc;
         }
         for (size_t i = 0; i < bytesAligned / 4; ++i) {
             flash_write_puya_buf[i] &= *ptr;
             ++ptr;
         }
-        rc = spi_flash_write(pos, flash_write_puya_buf, bytesAligned);
+        rc = spi_flash_write(pos, flash_write_puya_buf, bytesNow);
```
earlephilhower, line 709: Same comment as above, size must be x4 for writes, too.

earlephilhower:
@devyte, if you're still relaxing, can you eyeball this?  Am I missing something?

drzony:
@devyte 
AND is not out of bounds since the `flash_write_puya_buf` is 4 byte aligned and there is a check for `bytesNow > PUYA_BUFFER_SIZE`
@devyte @earlephilhower 
If the flash reads must be 4-byte aligned, then it should be fixed in `EspClass::flashWrite` and `EspClass::flashRead`, this means adding something similar to what is done currently for PUYA, for the generic case i.e.:
```cpp
if (!aligned) {
  readLast4BytesFromFlash();
  andReadWithData();
  writeLast4BytesToFlash();
}
```

yumkam:
See [my suggested change linked above](https://github.com/esp8266/Arduino/commit/b6bfe9907f5eae3f29817fe9358611270f4338b0), it does exactly this (and also does not over-read `data`).

drzony:
@yumkam It does not handle the generic case

drzony:
My additional explanation about flash writes:
Extra write will not reduce life of flash:
1. SPI flashes can usually write in 1-byte increments, on SPI it's 4 probably due to speed (i.e. DIO/QIO instructions)
2. Flash life is usually measured in erase cycles
3. On SPI we can write to all bytes on page no matter the order, but only once before erase is needed
4. The additional write will not erase flash, so it will not wear it more
5. Example of such approach can be seen in Linux UBIFS driver, which keeps accounting of written "sub-pages" and fills the whole page before erasing it
6. The only problem here is when someone tries to write on unaligned offsets i.e. write 3 bytes at offset 0, then write 1 byte at offset 2. If `spi_flash_write` somehow aligns the write, which seems to be the case, then the second write will fail as the bytes 0-3 were already written.

TD-er:
@drzony 
I don't agree with the statement about the writes.
The original PUYA chips for which the patch was made, is doing something in the background.
So what we're seeing is the data is exactly written to the flash as it is sent to it.
For flash memory, this is not what you would expect, as it should only change a 1 into a 0 and not the other way around.
But that's exactly what the PUYA chips do. If there is 0xAA stored in the flash and you write 0x55 to it, you would expect the cell to contain 0x00, but instead it holds 0x55.

So unless they have come up with some revolutionary new type of memory, they must perform an erase somewhere.
It could be the chip itself does buffer a page and does perform the erase/write cycle in the background, or it is using some page mapping.
But there must be extra erase/write cycles on those memory chips to get the result we're seeing.

drzony:
@TD-er 
I can see the problem, but I don't have an idea on how to test whether read(4), write(4), read(4), write(4) is any different from read(8), write(8), because this is what seems to be the worry here.

TD-er:
More like this:
write 3 bytes, start at 4N + 0  => read(4), write(4)
write 3 bytes, start at 4N + 3   => read(4), write(4), read(4), write(4)   (or read(8), write(8))

drzony:
Ok, so I have confirmed that spi_flash_write self-aligns, this code:
```cpp
uint32_t flash_offset = 0xa0000;
uint32_t wr_buf1[2]   = {0x55555555, 0x55555555};
uint32_t wr_buf2[2]   = {0xAAAAAAAA, 0xAAAAAAAA};
uint32_t rd_buf1[2]   = {0x00000000, 0x00000000};

spi_flash_erase_sector(flash_offset / 0x1000);
spi_flash_write(flash_offset, wr_buf1, 3);
spi_flash_read(flash_offset, rd_buf1, 4);
DEBUG("Read from flash + 0 %08x\n", rd_buf1[0]);
flash_read(flash_offset + 4, rd_buf1, 4);
DEBUG("Read from flash + 4 %08x\n", rd_buf1[0]);
spi_flash_write(flash_offset + 4, wr_buf1, 4);
spi_flash_read(flash_offset + 8, rd_buf1, 4);
DEBUG("Read from flash + 8 %08x\n", rd_buf1[0]);
spi_flash_write(flash_offset + 8, wr_buf2, 4);
spi_flash_read(flash_offset + 4, rd_buf1, 4);
DEBUG("Read from flash + 4 %08x\n", rd_buf1[0]);
spi_flash_read(flash_offset + 8, rd_buf1, 4);
DEBUG("Read from flash + 8 %08x\n", rd_buf1[0]);
```
Produces (on esp12e with "normal" flash and on S20 with PUYA flash):
```txt
Read from flash + 0 55555555
Read from flash + 4 ffffffff
Read from flash + 8 ffffffff
Read from flash + 4 55555555
Read from flash + 8 aaaaaaaa
```
So always 4 bytes are written

drzony:
Next test:
```cpp
uint32_t flash_offset = 0xa0000;
uint32_t wr_buf1[2]   = {0x55555555, 0x55555555};
uint32_t wr_buf2[2]   = {0xAAAAAAAA, 0xAAAAAAAA};
uint32_t rd_buf1[2]   = {0x00000000, 0x00000000};

spi_flash_erase_sector(flash_offset / 0x1000);
spi_flash_write(flash_offset, wr_buf1, 4);
spi_flash_write(flash_offset, wr_buf2, 4);
spi_flash_read(flash_offset, rd_buf1, 4);
DEBUG("Read from flash + 0 %08x\n", rd_buf1[0]);
```
Produces:
On "normal" flash:
```txt
Read from flash + 0 00000000
```
On PUYA flash:
```txt
Read from flash + 0 aaaa2aaa
```
So completely messed up.

TD-er:
Hmm that `2` is strange in the PUYA output.
If it were only `a` HEX values, then I could understand it, but this `2` I can't explain.

drzony:
And the last test for today:
```cpp
uint32_t flash_offset = 0xa0000;
uint32_t wr_buf1[2]   = {0xDEADBEEF, 0xDEADBEEF};
uint32_t rd_buf1[2] = {0x00000000, 0x00000000};
uint32_t rd_buf2[2] = {0x00000000, 0x00000000};
spi_flash_erase_sector(flash_offset / 0x1000);
spi_flash_read(flash_offset, rd_buf1, 4);
spi_flash_write(flash_offset, wr_buf1, 4);
spi_flash_read(flash_offset + 4, rd_buf1 + 1, 4);
spi_flash_write(flash_offset + 4, wr_buf1 + 1, 4);
spi_flash_read(flash_offset, rd_buf2, 8);
DEBUG("Read from flash + 0 %08x %08x %08x %08x\n", rd_buf1[0], rd_buf1[1], rd_buf2[0], rd_buf2[1]);
memset(rd_buf1, 0, 8);
memset(rd_buf2, 0, 8);
spi_flash_erase_sector(flash_offset / 0x1000);
spi_flash_read(flash_offset, rd_buf1, 8);
spi_flash_write(flash_offset, wr_buf1, 8);
spi_flash_read(flash_offset, rd_buf2, 8);
DEBUG("Read from flash + 0 %08x %08x %08x %08x\n", rd_buf1[0], rd_buf1[1], rd_buf2[0], rd_buf2[1]);
```
Produces
```txt
Read from flash + 0 ffffffff ffffffff deadbeef deadbeef
Read from flash + 0 ffffffff ffffffff deadbeef deadbeef
```
On both flashes, so if there is some erasing happening, then it's not affecting other data on flash.
As for the magic `2` in read, I also have no idea what is happening other than some weird buffering.

drzony:
@TD-er One thing keeps bugging me, what is the reason for the read/write combo on PUYA flashes?
As I've shown above if we keep to the standard erase/write-once cycle everything works as expected.
It seems that the problem was (and still is) with unaligned writes, which are self-aligned by `spi_flash_write`, which may cause double write on some bytes (which is always bad since we can only set 0s)

drzony:
I've done more tests and it seems that whole `spi_flash_write_puya` is not needed, does anybody have a device that was used when the original patch was created and can confirm my findings?

My thoughts:
1. Writing multiple times to a single address on the flash, although sometimes possible, is not a good practice. One reason is that we can only change 1s to 0s, which would mean that we would have to verify if the data can be written (read flash, check if the only changes are 1s->0s, write or return error). The second reason is that not all flashes react well to this. Puya is one example, but I've also encountered some NAND flashes which had similar issues (even if they contained all 1s, the second write without erase would fail)
2. `spi_flash_write` always writes 4 bytes to the flash, so it triggers the previous point on unaligned write. I we do `spi_flash_write(4n + 0, data1, 3)` and `spi_flash_write(4n + 3, data2, 1)` my PR will work on "nice" flashes, but it will fail on "weird" flashes
3. The additional write in my PR will not cause any additional flash wear, since there is no erase.
4. As an additional "speedup" I can remove the additional `spi_flash_read` and initialize `tempData = 0xFFFFFFFF`, this will work on "nice" flashes as intended (I still think that writing multiple times is not a good way), it will fail on "weird" flashes regardless of whether we do `read` or set `FF`.
5. As for weird behavior of PUYA flashes, I think what we see can be explained by bitflips in hardware. The physical flash in those chips may need less power to flip to 1 (or is not keeping 0s that well), so that is why we see the change in value. This means that I do not agree with @TD-er that there must be some erase on the way, it can be explained by just poor persistence.

drzony:
Just for completeness, a test with both aligned and unaligned `offset`
```cpp
uint32_t flash_offset = 0xa0003; // Testing with 0xa0000 gives the same result
uint32_t wr_buf[4]    = {0xDEADBEEF, 0x55555555, 0xAAAAAAAA, 0x88888888};
uint32_t rd_buf[4]    = {0x00000000, 0x00000000, 0x00000000, 0x00000000};
ESP.flashEraseSector(flash_offset / 0x1000);
ESP.flashWrite(flash_offset, wr_buf, 13);
ESP.flashRead(flash_offset, rd_buf, 16);
DEBUG("Flash read %08x %08x %08x %08x\n", rd_buf[0], rd_buf[1], rd_buf[2], rd_buf[3]);
ESP.flashWrite(flash_offset + 16, wr_buf, 14);
ESP.flashRead(flash_offset + 16, rd_buf, 16);
DEBUG("Flash read %08x %08x %08x %08x\n", rd_buf[0], rd_buf[1], rd_buf[2], rd_buf[3]);
ESP.flashWrite(flash_offset + 32, wr_buf, 15);
ESP.flashRead(flash_offset + 32, rd_buf, 16);
DEBUG("Flash read %08x %08x %08x %08x\n", rd_buf[0], rd_buf[1], rd_buf[2], rd_buf[3]);
ESP.flashWrite(flash_offset + 48, wr_buf, 16);
ESP.flashRead(flash_offset + 48, rd_buf, 16);
DEBUG("Flash read %08x %08x %08x %08x\n", rd_buf[0], rd_buf[1], rd_buf[2], rd_buf[3]);
```
Output:
```txt
Flash read deadbeef 55555555 aaaaaaaa ffffff88
Flash read deadbeef 55555555 aaaaaaaa ffff8888
Flash read deadbeef 55555555 aaaaaaaa ff888888
Flash read deadbeef 55555555 aaaaaaaa 88888888
```
The same output is on all 4 cases (PUYA-aligned, PUYA-unaligned, Other-aligned, Other-unaligned)

d-a-v:
> a device that was used when the original patch was created and can confirm my findings?

Do you suspect there are several versions of puya chips ?

drzony:
I suspect that they are the same, but I still do not understand why the read/AND/write patch was created and based on what findings. To my understanding the only problem with the code from the beginning (I mean before adding PUYA_SUPPORT at all) was the unaligned size passed to spi_flash_write, which was self-aligned by the function, writing excess bytes and over-reading memory.
My update problem can be solved in 2.7.3 just by disabling PUYA_SUPPORT.

TD-er:
The original patch was written to make writing to PUYA chips the same as traditional flash.
Some libraries, like SPIFFS filesystem, use the traditional behavior of flash to keep records of what blocks are used, free and ready to be erased.
So you need to have the behavior where you can only flip a bit to 0.
On PUYA chips the last data written is the value that's read the next time and not the NAND value of the last + previous data.

If you don't need this specific behavior, then it doesn't matter if you use a PUYA chip or any other if you only erase a block + only a single write to an address in that block.
But for using SPIFFS you really need this behavior.
I don't know if LittleFS uses similar optimizations.

drzony:
@TD-er Yes, I got to that after reading the SPIFFS issue, however I do not think that `spi_flash_write_puya` is properly implemented. It's relying on an undocumented feature that permits writes without erase, I think the proper solution would be to do erase before write. If the erase is done in the background as @igrr somehow found out, it would be nice to know if he got some feedback from the manufacturer on those issues.
Optimizations in SPIFFS are made on the assumption that "it usually works that way", which is not a good assumption. I think it would be better to make changes there (it's deprecated however, so I don't see the point) and make sure that LittleFS does not do the same.

drzony:
I've looked through the code of LittleFS and its documentation and I don't see any SPI tricks in there. It seems that it's always erase then write cycle.

d-a-v:
Would you have a chance to try LittleFS on your puya chip without the puya patch ?
If that works, maybe you can update the code to call the patched spi write functions only from spiffs, and use the regular one from everywhere else ?
That also would help reducing flash footprint when using LittleFS (spiffs is deprecated).

TD-er:
SPIFFS is one of the libraries I know uses this.
I have no idea if there are others, like perhaps the EPROM library, or maybe users that store their own counters.

If you try to do an erase every time something is marked as being used or marked free, then you're going to write a lot more frequently then needed.
Also it takes a lot more time as an erase does take a while and often can only be done on a block containing a number of pages. (32 pages of 256 bytes as far as I remember)
So you need to either every time use a new page for your administration, or read always the data before writing again to be sure you always keep the existing bits.
Worst case scenario, this administration process takes a bit longer.
Writing data (file content) should be done only on empty blocks, to make sure it is stored correctly.

About contact with the manufacturer... I don't expect a lot from it if you would try, as the PUYA brand is a bit shady. (at least that's how I see it).
They suddenly appeared with products on the market for apparently a price which was lower than all competitors as you saw them being used on the cheapest of modules.
Also several people noticed their datasheets did resemble those of other brands, but the behavior of the chips was completely different from all other flash chips on the market... which is something you would expect to be mentioned in such elaborate datasheets.

drzony:
I'll try to do some test withing the next few days. For me the idea, that the core tries to emulate some other flash, is a bit ridiculous. I have a few years of experience with flash based consumer devices and writing to the same offset without erase was always a no-go (no matter what flash was used).
EEPROM library stores data in memory and writes a full block on commit.
Writing to flash on a bit-by-bit basis as is the case with SPIFFS is asking for trouble.
The only thing, I'm still wondering about, is where the idea about the erases in the background came from? Is it somehow confirmed by someone with proper tools?
For me the issue can be explained by poor persistence and bit flipping on second write and would be inline with "shady" cost-cutting company, rather than having some kind of internal buffer, copying the data, erasing the flash and then copying it back, which seems like a very expensive piece of hardware, considering the cost of those flashes.

drzony:
@TD-er 
Relying on bit access is a bad design, You don't need to write the bitmap (taking FS as an example) each time, you do accounting i.e. write to next byte (or 4-bytes in our case) and then erase when you reach end of the page. Reusing bytes from a page is always bad, since there are a lot of flashes that have problems with this. More expensive flashes can keep their own accounting.
As I've written before I have encountered a mid tier NAND flash that, on second write without erase, just didn't write the data, while reporting success.
This considers NAND flashes, but [Linux UBI docs](http://www.linux-mtd.infradead.org/doc/ubi.html#L_flasher_algo) mention that some flashes don't like second writes without erase, the same is probably true for cheap NOR/SPI flashes.

TD-er:
> Relying on bit access is a bad design, You don't need to write the bitmap (taking FS as an example) each time, you do accounting i.e. write to next byte (or 4-bytes in our case) and then erase when you reach end of the page. Reusing bytes from a page is always bad, since there are a lot of flashes that have problems with this. More expensive flashes can keep their own accounting.

These two lines contradict eachother.

There is a bitmap representing what pages are free and one holding the pages used, but no longer allocated for storage.
If you simply write the new content of this bitmap as it is intended to be after writing (which on itself is not that big of a deal), then you still need to take into account the fact you cannot change a 0 into a 1 on most flash chips (so far only PUYA chips show other behavior) without an erase.
So the meaning of a 0 and a 1 in these tables may appear a bit strange at first sight.
A 1 means it is the initial state after an erase and the 0 means it is altered.
This means you need to rely on bit access of the storage medium, or else you have to read an etire block (8 kB), alter it in memory, erase the block and write it back. Or you copy the block to a new block and while copying you alter it to the new state, just to save the very expensive 8k allocation.
As a result you perform an erase on every single file update, which is going to wear out the flash very quickly as an erase (or at least every change from a 0 to a 1) counts as one write cycle.
All updates on a bitmap until it must be erased are 1 write cycle in total.

Performing an erase, just to make it "feel right" is going to destroy the sectors used for the file table very quickly.
So it may feel as a bad design choice, or whatever programmer's itch it may cause, but it is clearly done with a very very good reason.
If you want to read the data, alter it in memory and write it back to be sure it is written as intended... well that was exactly what the PUYA patch did.

Now back to the "bitflip" of the PUYA chips.
It is very well possible that's what may be happening here. The datasheet does mention some page buffer size, but like I said I doubt that datasheet is reflecting the true inner workings of these chips, so an unintended bitflip may also be very plausible.
If it is an unintended bitflip, then I am rather worried what happens to neighboring bits if you write data to some address.

But as you may already have guessed, I'm absolutely not a fan of PUYA chips. I just consider them a plague we just have to learn to live with as lots and lots of manufacturers (still) use them.

drzony:
I've simplified the approach, but those lines do not contradict each other. I was writing about reusing not using all of them once. Generally idea is to write new versions of bitmap in the next block of the page. For example we have 64 pages of flash, so we need 64 bits of data, each page is 256 bytes, so we can write 4 versions of bitmap before erase. UBI/UBIFS does this, LittleFS does this, other flash FSes do it. If you want to go deeper, the design document for LittleFS is great, UBI documents are also noteworthy.
That is why I think it's the responsibility of the filesystem to treat the flash in a robust way, this is raw flash, not an SSD with a 25$-50$ controller inside.
I know that the erase on every write is a bad thing, don't get me wrong, but what is done in the spi_flash_write_puya is fixing the design flaw of the filesystem in the flash "driver", by using undocumented "magic" property that may or may not work depending on the manufacturing run of the chip.
SPIFFS should not use bit writes to flash, if it does, then it should have a whitelist of flashes that are thoroughly tested as supporting this feature (which is not guaranteed, since even the "not shady" manufacturers do not put it in their spec, so new batch of chips may behave differently)

TD-er:
Well then you run into another problem.
You can't erase a 256 byte page.
You can only erase  a block of consecutive pages (32 if memory serves me well, but with these temperatures, I also may experience bit flips or shifts)
So your approach would then use at least 8 k to keep these bitmaps, regardless the number of blocks you need to administer.
And probably an extensive check at boot to see what block you were left the last time.

drzony:
I'm just giving an example, not providing a fool-proof design for a filesystem. LittleFS design docs provide a great insight on how to store metadata [here is the file](https://github.com/ARMmbed/littlefs/blob/a049f1318eecbe502549f9d74a41951985fb956f/DESIGN.md). Yes sometimes a check on mount is a viable tradeoff, if not, then there are some other methods to avoid it (with other tradeoffs) and so on.
I'm just saying that relying on "magic", undocumented properties of the medium when designing a filesystem is something that will break the whole thing at some point in the future, and that future is now.

TD-er:
> I'm just saying that relying on "magic", undocumented properties of the medium when designing a filesystem is something that will break the whole thing at some point in the future, and that future is now.

Well that was also the cause for the "PUYA patch" in the first place.

drzony:
Yes and I strongly think that this should be fixed on the SPIFFS side and the "PUYA patch" should be removed as it "cheats" the flash by doing a bitbanging combo on the medium to pretend that it is a "normal" flash that supports an undocumented feature.
If someone else uses the flash in this manner, then he should also read a lot about how flashes work and what might go wrong, before doing such a thing (or use EEPROM library that does the thinking for him).
I do not have a ready-to-go solution on how to fix it on SPIFFS side, but the easiest way would be to move the logic of detecting flash type and choosing a write method based on it. This is an almost "no effort" solution. Right now we are poisoning a generic flashWrite function with logic to circumvent a bad library. As my example shows, the workaround for SPIFFS comletely broke gzipped OTA on my device (which does the right thing by erasing before write).

TD-er:
Well the problem is, you don't know if SPIFFS is the only one affected.
So by moving the patch from one place to another, you may cause a lot of frustration on those users that are completely unaware why their system suddenly breaks.

It would be nice if it could be fixed in all those spots where they made optimizations based on how almost all flash chips operate, but the sad truth is that you probably will not be able to do so.

For me, I really don't know why PUYA support is needed for an OTA (gzipped or not). Is it to allow a little more usable space which may already erase a sector that's being read?

drzony:
PUYA support is not required for gzipped OTA, but spi_flash_write_puya has a bug, that prevents it from writing last `size % 4` bytes to flash (see #7491 ).
So although it's not needed, it's on by default and poisons any flash write. One thing is that it consumes memory, second thing is that it makes additional read, third thing is that it does a lot of unnecessary ANDs on memory.
A workaround for a bad FS design is causing breakage and performance hit for things that do not require it (and they do not require it, because they are written according to spec).

TD-er:
Well the original PUYA patch did look at the vendor ID of the flash chip. So it did not use extra resources other than a single look at the vendor ID at boot.
Only the size of the binary was a bit larger, but that was avoidable using a define.

drzony:
Have you read the #7491 ?
1. I have PUYA flash, `spi_flash_write_puya` is used for all flash writes.
2. Due to a bug in it's logic it breaks gzipped OTA on PUYA flashes
3. OTA on PUYA flashes does not require use of `spi_flash_write_puya` because it behaves properly (erase then write), but Esp::flashWrite does not know whether this write is a proper write or "weird optimized write", so it just calls `spi_flash_write_puya` which has a performance hit and had a bug before my fix.
4. This means that on PUYA flashes all writes including those that work without "PUYA patch" take a performance hit for no reason other than "we want a workaround for SPIFFS and bad flash access".

TD-er:
> 4\. This means that on PUYA flashes all writes including those that work without "PUYA patch" take a performance hit for no reason other than "we want a workaround for SPIFFS and bad flash access".

You can also turn this argument around.
By using PUYA flash chips, you should not worry about the performance hit.
Those are the cheapest and dirtiest chips around, operating completely different from all other brands.
If there is a one-fix-fixes-all, then sure there is a penalty in performance, but I truly cannot be sure it will work properly in all other situations (other then SPIFFS).

I do agree fixing it in SPIFFS does sound great, but on the other hand it is a bit strange to remove a perfectly fine working optimization for all other nice behaving flash chips just to be sure those that don't act like proper flash chips will also work.
Also a file system library should be hardware agnostic (as you also pointed out), so enabling/disabling optimization for specific brands of chips sounds a bit strange.
Given the nature of SPIFFS (SPI Flash File System), it is not that strange it does use some very specific known features of SPI flash chips.
Sure you could allocate a single bit in the file system structure to detect what kind of chip you're dealing with.
Just write a 0 to it and later try setting it to 1 without calling an erase and you know what kind of chip you're dealing with.
But the same could be used in this library.
N.B. this can be done at the first initialization of the chip/file system, so you only need to test it once.

To be honest, if you're using PUYA chips, you may be glad it has support after all, as it is complete and utter rubbish.
So if you have to pay a penalty for it by having some performance loss, so be it.   (at least something is "paid" then...)
Adding extra code to detect when to use it and when not is not worth the effort as you will always loose a lot of debugging time finding out where you missed a spot where the patch may be needed after all.

I'm glad you fixed the alignment issue, but I do think trying to "optimize" PUYA support is going to bite you later on.
For example by increasing the write speed you may suddenly remove the extra delays that now magically prevent data corruption.
Like I said, it is absolute garbage.

drzony:
My argument is that using multiple writes on the same address is out of spec for all flashes (each byte can be written only once).
PUYA flashes behave correctly and according to spec. I'm not trying to optimize PUYA writes, I'm trying to write according to spec. I'm also trying to convince you that writing in an out of spec way on other flashes, may cause problems sometimes in the future.
If a library really wants to write in a out of spec manner, then it can do it using Esp::flashWrite.
What is done now is adding additional hidden logic to Esp::flashWrite that does a read.
Maybe PUYA flashes are crap, maybe you just don't like them, but the fact is that I do not have to "pay" anything on them unless I want to do out of spec stuff.
If we really want to support this out of spec method, then let's add Esp::flashWriteNoErase or something like that. Libraries/users can use it knowing that it's a "works if you're lucky" function. And we will not be breaking properly behaved libraries.

TD-er:
> If we really want to support this out of spec method, then let's add Esp::flashWriteNoErase or something like that. Libraries/users can use it knowing that it's a "works if you're lucky" function. And we will not be breaking properly behaved libraries.

That's how it is now, with the `PUYA_SUPPORT` build flag.
If you make sure it is set to 0, you will not have to endure the write penalty if you don't use SPIFFS.
By the way, last time I checked, there was also some mention in the SPIFFS library about this kind of behavior, so maybe we can already toggle another define to let SPIFFS handle the PUYA specific quirks.
When the SPIFFS library was "updated" (read: cherry picked) recently, it had some remarks in the notes about being able to handle chips which does sound like PUYA chip behavior.

drzony:
Now it's all or nothing, either I use `spi_flash_write_puya` everywhere, or nowhere.
What I'm suggesting is letting library/user decide which method should be used. In spec libraries can use traditional `Esp::flashWrite`, out of spec libraries that want that "optimization" can opt in to use `Esp::flashWriteNoErase`.

[Here](https://github.com/pellepl/spiffs/issues/172) is another issue with SPIFFS "optimization" where it relied on 0->1 transitions being ignored.
It turned out to be causing an error on a high end flash, which had integrated protection against such behavior.
That is why I'm very reluctant to accept such optimizations anywhere.

drzony:
Mentioning #4061 just for reference if anyone would like to read up on the "PUYA patch".

arendst:
Oh my, if I need to erase for every write my Quick Power Cycle detection gos haywire....

```
void UpdateQuickPowerCycle(bool update)
{
  if (Settings.flag3.fast_power_cycle_disable) { return; }  // SetOption65 - Disable fast power cycle detection for device reset

  uint32_t pc_register;
  uint32_t pc_location = SETTINGS_LOCATION - CFG_ROTATES;

  ESP.flashRead(pc_location * SPI_FLASH_SEC_SIZE, (uint32*)&pc_register, sizeof(pc_register));
  if (update && ((pc_register & 0xFFFFFF80) == 0xFFA55A80)) {
    uint32_t counter = ((pc_register & 0x7F) << 1) & 0x7F;
    if (0 == counter) {  // 7 power cycles in a row
      SettingsErase(3);  // Quickly reset all settings including QuickPowerCycle flag
      EspRestart();      // And restart
    } else {
      pc_register = 0xFFA55AB0 | counter;
      ESP.flashWrite(pc_location * SPI_FLASH_SEC_SIZE, (uint32*)&pc_register, sizeof(pc_register));
      AddLog_P2(LOG_LEVEL_DEBUG, PSTR("QPC: Flag %02X"), counter);
    }
  }
  else if (pc_register != 0xFFA55AFF) {
    pc_register = 0xFFA55AFF;
    // Assume flash is default all ones and setting a bit to zero does not need an erase
    if (ESP.flashEraseSector(pc_location)) {
      ESP.flashWrite(pc_location * SPI_FLASH_SEC_SIZE, (uint32*)&pc_register, sizeof(pc_register));
    }
    AddLog_P2(LOG_LEVEL_DEBUG, PSTR("QPC: Reset"));
  }
}
```

TD-er:
@arendst Just curious, what area do you use for this counter? Some area where you normally keep the OTA image?

drzony:
@arendst 
You do not need to erase with every write, just erase with every write to the same location. For counter you can use consecutive bytes, you are using the whole sector anyway so instead of single 4-byte value use 7*4 bytes for counter.
The problem I'm raising here is that if you do this the way you are doing it now, it may stop working at any time. If the flash chip changes in new modules, if a new batch of chips do not allow for bit writes etc.
That is why I'm proposing the new Esp::flashWriteNoErase, you would be able to use it in your code, if you are sure that you really want it and have tested this with all devices that you are going to use.
From my perspective the "PUYA patch" is unstable, because it relies on a weird design/manufacturing flaw in the chip itself

arendst:
Yep. I keep config data in 8 rolling pages just below the EEPROM area. The page before that I use a uint32_t to keep track of power cycles using the bit change method.

If OTA is requested I move the current config to the eeprom area, execute OTA and reinit the QPC uint32_t after restart.

arendst:
Thx for the info I changed my QPC code to the below which now even seems to be running much better; the original didn't seem to work on some devices for now obvious reasons:

> just erase with every write to the same location

```
  const uint32_t QPC_COUNT = 7;  // Number of Power Cycles before Settings erase
  const uint32_t QPC_SIGNATURE = 0xFFA55AFF;
  const uint32_t qpc_sector = SETTINGS_LOCATION - CFG_ROTATES;
  const uint32_t qpc_location = qpc_sector * SPI_FLASH_SEC_SIZE;

  uint32_t qpc_buffer[QPC_COUNT +1];
  ESP.flashRead(qpc_location, (uint32*)&qpc_buffer, sizeof(qpc_buffer));
  if (update && (QPC_SIGNATURE == qpc_buffer[0])) {
    uint32_t counter = 1;
    while ((0 == qpc_buffer[counter]) && (counter <= QPC_COUNT)) { counter++; }
    if (QPC_COUNT == counter) {  // 7 power cycles in a row
      SettingsErase(3);          // Quickly reset all settings including QuickPowerCycle flag
      EspRestart();              // And restart
    } else {
      qpc_buffer[0] = 0;
      ESP.flashWrite(qpc_location + (counter * 4), (uint32*)&qpc_buffer, 4);
      AddLog_P2(LOG_LEVEL_INFO, PSTR("QPC: Count %d"), counter);
    }
  }
  else if ((qpc_buffer[0] != QPC_SIGNATURE) || (0 == qpc_buffer[1])) {
    qpc_buffer[0] = QPC_SIGNATURE;
    // Assume flash is default all ones and setting a bit to zero does not need an erase
    if (ESP.flashEraseSector(qpc_sector)) {
      ESP.flashWrite(qpc_location, (uint32*)&qpc_buffer, 4);
      AddLog_P2(LOG_LEVEL_INFO, PSTR("QPC: Reset"));
    }
  }
```

drzony:
@arendst Glad I could help, if in the future you would need more space, you can squeeze more info in those 4 bytes. For example:
Byte.bit | 0.0 | 0.1-0.3 | 0.4-0.7 | 1 | 2 | 3 |
-|-|-|-|-|-|-|
Data | 0 - Used, 1 - free | Counter | Some config | Other config | Next config | Last config |

Then you can only look at the last 4 bytes with bit 0 = 0, which will be the valid 4-byte "data". Even in the case of counter you can then use whole sector before erase, which will preserve the flash. Instead of erasing the whole sector you can just write 0 at byte 0 bits 1-3. Here is some pseudo code:
```cpp
uint32_t data, valid_data;
uint32_t offset = 0;
read(offset, &data, 4);
if (data & 0x80000000) {
  // No valid data, initialize
} else {
  while (!(data & 0x80000000) && offset < sector_size) {
    valid_data = data;
    offset += 4;
    read(offset, &data, 4);
  }
  // We have valid data now, do something with it
  uint32_t new_data = do_something(valid_data);
  new_data &= ~0x80000000; // Set used bit to 0
  if (offset == sector_size) {
    erase();
    offset = 0;
  }
  write(offset, &new_data, 4);
}
```

Above is just an example (sector size checking must be done better), but should show my point.

devyte:
I am completely lost wrt this PR. @drzony , please look me up in [gitter](https://gitter.im/esp8266/Arduino) for a direct discussion.

cores/esp8266/Esp.cpp
```diff
@@ -715,20 +720,35 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
+        }
     }
     return rc == SPI_FLASH_RESULT_OK;
 }
 
 bool EspClass::flashRead(uint32_t offset, uint32_t *data, size_t size) {
-    auto rc = spi_flash_read(offset, (uint32_t*) data, size);
+    size_t sizeAligned = size & ~3;
+    auto rc = spi_flash_read(offset, data, sizeAligned);
+    if (sizeAligned < size) {
```
drzony, line 747: Should have rc check

cores/esp8266/Esp.cpp
```diff
@@ -715,20 +720,35 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
-        rc = spi_flash_write_puya(offset, data, size);
+        rc = spi_flash_write_puya(offset, data, sizeAligned);
     }
     else
 #endif
     {
-        rc = spi_flash_write(offset, data, size);
+        rc = spi_flash_write(offset, data, sizeAligned);
+    }
+    if (sizeAligned < size) {
```
drzony, line 733: rc check

drzony:
The code from latest commit is not tested yet. I've pushed it to let @devyte have a preview of the changes

cores/esp8266/flash_hal.cpp
```diff
@@ -28,141 +28,27 @@ extern "C" {
...
- relevant bytes into result buffer.
-
-alignment:       012301230123012301230123
-bytes requested: -------***********------
-read directly:   --------xxxxxxxx--------
-read pre:        ----aaaa----------------
-read post:       ----------------bbbb----
-alignedBegin:            ^
-alignedEnd:                      ^
-*/
```
devyte, line 43: These comments shouldn't be dropped, but should instead be moved to where the logic is now implemented

cores/esp8266/flash_hal.cpp
```diff
@@ -28,141 +28,27 @@ extern "C" {
...
 
-/*
- Like spi_flash_read, spi_flash_write has a requirement for flash address to be
- aligned. However it also requires RAM address to be aligned as it reads data
- in 32-bit words. Flash address (mis-)alignment is handled much the same way
- as for reads, but for RAM alignment we have to copy data into a temporary
- buffer. The size of this buffer is a tradeoff between number of writes required
- and amount of stack required. This is chosen to be 512 bytes here, but might
- be adjusted in the future if there are good reasons to do so.
-*/
```
devyte, line 98: Same here, these comments shouldn't be lost, but rather should be moved to the new code and get updated

cores/esp8266/Esp.h
```diff
@@ -150,8 +150,10 @@ class EspClass {
         bool checkFlashCRC();
 
         bool flashEraseSector(uint32_t sector);
-        bool flashWrite(uint32_t offset, uint32_t *data, size_t size);
+        bool flashWrite(uint32_t offset, const uint32_t *data, size_t size);
+        bool flashWrite(uint32_t offset, const uint8_t *data, size_t size);
         bool flashRead(uint32_t offset, uint32_t *data, size_t size);
+        bool flashRead(uint32_t offset, uint8_t *data, size_t size);
```
devyte, line 156: These should have comments explaining the differences, e.g.: 
```cpp
//This call requires that offset be aligned to 4-byte boundary and size be multiple-of-4, or the call fails.
bool flashRead(uint32_t offset, uint32_t *data, size_t size);
//This allows misaligned offset and size that is not multiple-of-4, but performance slows down to handle all the corner cases.
bool flashRead(uint32_t offset, uint8_t *data, size_t size);
```
Same for Writes.

cores/esp8266/Esp.cpp
```diff
@@ -692,39 +698,141 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
-#endif
+#endif // PUYA_SUPPORT
     {
-        rc = spi_flash_write(offset, data, size);
+        rc = spi_flash_write(offset, const_cast<uint32_t *>(data), size);
     }
     return rc == SPI_FLASH_RESULT_OK;
 }
 
+static const int UNALIGNED_WRITE_BUFFER_SIZE = 512;
```
devyte, line 733: Instead of a static const int, please use a constexpr size_t.

cores/esp8266/Esp.cpp
```diff
@@ -692,39 +698,141 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
+
+    return true;
+}
+
+bool EspClass::flashRead(uint32_t offset, uint8_t *data, size_t size) {
+    size_t sizeAligned = size & ~3;
+    size_t currentOffset = 0;
+
+    if ((uintptr_t)data % 4 != 0) {
+        uint32_t alignedData[UNALIGNED_WRITE_BUFFER_SIZE / sizeof(uint32_t)] __attribute__((aligned(4)));
```
devyte, line 798: This uses up 512 bytes on the stack, which is just 4KB. I'm not sure whether that's ok, but we're trying to not have large stack usage.

cores/esp8266/Esp.cpp
```diff
@@ -664,10 +664,16 @@ bool EspClass::flashEraseSector(uint32_t sector) {
...
 
 #if PUYA_SUPPORT
+#if PUYA_BUFFER_SIZE % 256 != 0
+#error PUYA_BUFFER_SIZE is not 256 byte aligned
+#endif
 static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, size_t size) {
     if (data == nullptr) {
       return SPI_FLASH_RESULT_ERR;
     }
+    if (size % 4 != 0) {
```
devyte, line 674: Shouldn't 4-byte alignment also be enforced on offset here, given that the corner cases are handled above this function?

cores/esp8266/Esp.cpp
```diff
@@ -692,39 +698,141 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
 }
 
+static const int UNALIGNED_WRITE_BUFFER_SIZE = 512;
+
+bool EspClass::flashWrite(uint32_t offset, const uint8_t *data, size_t size) {
+    size_t sizeAligned = size & ~3;
+    size_t currentOffset = 0;
+    if ((uintptr_t)data % 4 != 0) {
+        // Memory is unaligned, so we need to copy it to an aligned buffer
+        uint32_t alignedData[UNALIGNED_WRITE_BUFFER_SIZE / sizeof(uint32_t)] __attribute__((aligned(4)));
```
devyte, line 740: This uses up 512 bytes of the 4KB stack.

cores/esp8266/Updater.cpp
```diff
@@ -236,7 +236,7 @@ bool UpdaterClass::end(bool evenIfRemaining){
     DEBUG_UPDATER.printf_P(PSTR("[Updater] Adjusted binsize: %d\n"), binSize);
 #endif
       // Calculate the MD5 and hash using proper size
-    uint8_t buff[128];
+    uint8_t buff[128] __attribute__((aligned(4)));
```
devyte, line 239: This is 128 bytes on our stack of 4KB. There is a Esp.flashRead() that follows, which uses up an additional 512 bytes of stack.

devyte (changes requested):

cores/esp8266/Updater.cpp
drzony: `Esp.flashRead()` will not use 512 bytes since `buff` is aligned, that's why I added `__attribute__((aligned(4)))`

cores/esp8266/flash_hal.cpp
drzony: This logic is dropped, since it was not true, address does not need to be aligned (apart from page boundary case that we talked)

cores/esp8266/flash_hal.cpp
drzony: Same as in read

cores/esp8266/Esp.cpp
drzony: This is an internal function, it's only used by flashWrite, but I can add the restrictions

cores/esp8266/Esp.cpp
drzony: From the comment in `flash_hal.c`:
>The size of this buffer is a tradeoff between number of writes required
 and amount of stack required. This is chosen to be 512 bytes here, but might
 be adjusted in the future if there are good reasons to do so.
I think I can lower it to 256 (page size), plus it only happens when unaligned memory is passed

drzony:
Test code:
```
void testFlash(uint32_t start_offset, uint8_t data_offset, size_t amount)
{
    static uint32_t *write_buffer = (uint32_t *)malloc(4096);
    static uint32_t *read_buffer = (uint32_t *)malloc(4096);

    for (uint32_t i = 0; i < 1024; i++)
    {
        write_buffer[i] = (i + 100) * 33;
        read_buffer[i] = 0xAAAAAAAA;
    }
    Serial.println("---------------------------------------------------");
    ESP.flashEraseSector(start_offset / 0x1000);
    Serial.printf("Testing %d bytes @ %08x + %d\n", amount, start_offset, data_offset);
    unsigned long start = micros();

    if (!ESP.flashWrite(start_offset, (uint8_t *)write_buffer + data_offset, amount))
    {
        Serial.printf("Write fail\n");
    }
    if (!ESP.flashRead(start_offset, (uint8_t *)read_buffer + data_offset, amount))
    {
        Serial.printf("Read fail\n");
    }
    if (memcmp((uint8_t *)write_buffer + data_offset, (uint8_t *)read_buffer + data_offset, amount) != 0)
    {
        Serial.printf("Compare fail\n");
        dumpBuffer((uint8_t *)write_buffer, 8);
        dumpBuffer((uint8_t *)read_buffer, 8);
    }
    Serial.printf("Write took %lu us\n", DL::Timing::diff(start, micros()));
}
testFlash(0xa0000, 0, 100);
testFlash(0xa0000, 0, 101);
testFlash(0xa0000, 0, 512);
testFlash(0xa0000, 0, 515);

testFlash(0xa0000, 1, 100);
testFlash(0xa0000, 2, 101);
testFlash(0xa0000, 3, 512);
testFlash(0xa0000, 1, 515);

testFlash(0xa0001, 0, 100);
testFlash(0xa0001, 0, 128);

testFlash(0xa0001, 1, 100);
testFlash(0xa0001, 2, 128);
testFlash(0xa0001, 3, 512);

testFlash(0xa0003, 1, 100);
testFlash(0xa0003, 2, 128);

testFlash(0xa0000 + 255, 0, 1);
testFlash(0xa0000 + 255, 1, 1);

testFlash(0xa0000 + 253, 0, 3);
testFlash(0xa0000 + 253, 1, 3);

testFlash(0xa0000 + 255, 0, 2);
testFlash(0xa0000 + 255, 1, 2);
```

Results:

```
---------------------------------------------------
Testing 100 bytes @ 000a0000 + 0
Write took 502 us
---------------------------------------------------
Testing 101 bytes @ 000a0000 + 0
Write took 665 us
---------------------------------------------------
Testing 512 bytes @ 000a0000 + 0
Write took 1607 us
---------------------------------------------------
Testing 515 bytes @ 000a0000 + 0
Write took 1774 us
---------------------------------------------------
Testing 100 bytes @ 000a0000 + 1
Write took 550 us
---------------------------------------------------
Testing 101 bytes @ 000a0000 + 2
Write took 709 us
---------------------------------------------------
Testing 512 bytes @ 000a0000 + 3
Write took 1830 us
---------------------------------------------------
Testing 515 bytes @ 000a0000 + 1
Write took 1994 us
---------------------------------------------------
Testing 100 bytes @ 000a0001 + 0
Write took 501 us
---------------------------------------------------
Testing 128 bytes @ 000a0001 + 0
Write took 554 us
---------------------------------------------------
Testing 100 bytes @ 000a0001 + 1
Write took 549 us
---------------------------------------------------
Testing 128 bytes @ 000a0001 + 2
Write took 606 us
---------------------------------------------------
Testing 512 bytes @ 000a0001 + 3
Write took 2176 us
---------------------------------------------------
Testing 100 bytes @ 000a0003 + 1
Write took 550 us
---------------------------------------------------
Testing 128 bytes @ 000a0003 + 2
Write took 606 us
---------------------------------------------------
Testing 1 bytes @ 000a00ff + 0
Write fail
Write took 491 us
---------------------------------------------------
Testing 1 bytes @ 000a00ff + 1
Write fail
Write took 486 us
---------------------------------------------------
Testing 3 bytes @ 000a00fd + 0
Write fail
Write took 495 us
---------------------------------------------------
Testing 3 bytes @ 000a00fd + 1
Write fail
Write took 489 us
---------------------------------------------------
Testing 2 bytes @ 000a00ff + 0
Write took 427 us
---------------------------------------------------
Testing 2 bytes @ 000a00ff + 1
Write took 422 us
```

earlephilhower (changes requested):
Nice work, thanks!

Because this touches a lot of things in the core and blob, can we please get a device test for it with all the new corner cases enumerated and verified?  Something that goes thru all possible aligned/unaligned source/dest/lens?  That would give great confidence that we're not breaking some corner case only hit by a FS or something very rarely.

drzony:
> 
> 
> Nice work, thanks!
> 
> Because this touches a lot of things in the core and blob, can we please get a device test for it with all the new corner cases enumerated and verified? Something that goes thru all possible aligned/unaligned source/dest/lens? That would give great confidence that we're not breaking some corner case only hit by a FS or something very rarely.

Code from my comment above covers that, I'll put that into a test

drzony:
@devyte @earlephilhower Can any of you run those tests on the device. I'm using Platformio for development on Windows, so setting up a testing environment will take me some time.

earlephilhower:
@drzony I'll give this PR a run thru my system and report back.  Thanks for adding tests!

earlephilhower:
@drzony, a minor patch is required to get it to build and run:
````
diff --git a/tests/device/test_spi_flash/test_spi_flash.ino b/tests/device/test_spi_flash/test_spi_flash.ino
index a0cc1c583..33746d10b 100644
--- a/tests/device/test_spi_flash/test_spi_flash.ino
+++ b/tests/device/test_spi_flash/test_spi_flash.ino
@@ -1,6 +1,8 @@
 #include <BSTest.h>
 #include <ESP8266WiFi.h>
 
+BS_ENV_DECLARE();
+
 void preinit() {
     // (no C++ in function)
     // disable wifi
@@ -59,7 +61,7 @@ bool testFlash(uint32_t start_offset, uint8_t data_offset, size_t amount)
         dumpBuffer((uint8_t *)read_buffer, 8);
         return false;
     }
-    Serial.printf("Write took %lu us\n", DL::Timing::diff(start, micros()));
+    Serial.printf("Write took %lu us\n", micros() - start);
     return true;
 }
 
````


Once done, I'm seeing a failure on the "Last bytes of page" test:
````
No SPIFFS to upload
Uploading binary
Running tests
running test "|+o|+m|+s|0p|"
test "|+o|+m|+s|0p|" passed
running test "|+o|+m|+s|1p|"
test "|+o|+m|+s|1p|" passed
running test "|+o|+m|-s|0p|"
test "|+o|+m|-s|0p|" passed
running test "|+o|+m|-s|2p|"
test "|+o|+m|-s|2p|" passed
running test "|+o|-m|+s|0|"
test "|+o|-m|+s|0|" passed
running test "|+o|-m|+s|1p|"
test "|+o|-m|+s|1p|" passed
running test "|+o|-m|-s|0p|"
test "|+o|-m|-s|0p|" passed
running test "|+o|-m|-s|2p|"
test "|+o|-m|-s|2p|" passed
running test "|-o|+m|+s|0p|"
test "|-o|+m|+s|0p|" passed
running test "|-o|+m|+s|1p|"
test "|-o|+m|+s|1p|" passed
running test "|-o|+m|-s|0p|"
test "|-o|+m|-s|0p|" passed
running test "|-o|+m|-s|1p|"
test "|-o|+m|-s|1p|" passed
running test "|-o|-m|+s|0p|"
test "|-o|-m|+s|0p|" passed
running test "|-o|-m|+s|1p|"
test "|-o|-m|+s|1p|" passed
running test "|-o|-m|-s|0p|"
test "|-o|-m|-s|0p|" passed
running test "|-o|-m|-s|1p|"
test "|-o|-m|-s|1p|" passed
running test "Last bytes of page"
test "Last bytes of page" failed
running test "Unaligned page cross only"
test "Unaligned page cross only" passed
````

earlephilhower:
Verbose output:
````
running test "Last bytes of page"
test output was:
17
>>>>>bs_test_start file="test_spi_flash.ino" line=153 name="Last bytes of page" desc="[spi_flash]"
---------------------------------------------------
Testing 1 bytes @ 000a00ff + 0
Write fail
>>>>>bs_test_check_failure line=155
---------------------------------------------------
Testing 1 bytes @ 000a00ff + 1
Write fail
>>>>>bs_test_check_failure line=156
---------------------------------------------------
Testing 2 bytes @ 000a00fe + 0
Write fail
>>>>>bs_test_check_failure line=157
---------------------------------------------------
Testing 2 bytes @ 000a00fe + 1
Write fail
>>>>>bs_test_check_failure line=158
---------------------------------------------------
Testing 3 bytes @ 000a00fd + 0
Write fail
>>>>>bs_test_check_failure line=159
---------------------------------------------------
Testing 3 bytes @ 000a00fd + 1
Write fail
>>>>>bs_test_check_failure line=160
>>>>>bs_test_end line=0 result=0 checks=6 failed_checks=6

test "Last bytes of page" failed
````

drzony:
@earlephilhower Thanks for the assist :) Everything should be fixed now

drzony:
Still testing on PUYA flash, do not merge yet

drzony:
I've managed to run those tests on PUYA device and they passed, so the PR is ready to merge.

earlephilhower:
Latest version passes in device tests for me, too:
````
--------------------------------
Running test 'test_spi_flash/test_spi_flash.ino' of 26 tests
Compiling test_spi_flash.ino
Executable segment sizes:
IROM   : 250596          - code in flash         (default or ICACHE_FLASH_ATTR) 
IRAM   : 27072   / 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...) 
DATA   : 1328  )         - initialized variables (global, static) in RAM/HEAP 
RODATA : 1880  ) / 81920 - constants             (global, static) in RAM/HEAP 
BSS    : 25264 )         - zeroed variables      (global, static) in RAM/HEAP 
Sketch uses 280876 bytes (26%) of program storage space. Maximum is 1044464 bytes.
Global variables use 28472 bytes (34%) of dynamic memory, leaving 53448 bytes for local variables. Maximum is 81920 bytes.
No SPIFFS to upload
Uploading binary
Running tests
running test "|+o|+m|+s|0p|"
test "|+o|+m|+s|0p|" passed
running test "|+o|+m|+s|1p|"
test "|+o|+m|+s|1p|" passed
running test "|+o|+m|-s|0p|"
test "|+o|+m|-s|0p|" passed
running test "|+o|+m|-s|2p|"
test "|+o|+m|-s|2p|" passed
running test "|+o|-m|+s|0|"
test "|+o|-m|+s|0|" passed
running test "|+o|-m|+s|1p|"
test "|+o|-m|+s|1p|" passed
running test "|+o|-m|-s|0p|"
test "|+o|-m|-s|0p|" passed
running test "|+o|-m|-s|2p|"
test "|+o|-m|-s|2p|" passed
running test "|-o|+m|+s|0p|"
test "|-o|+m|+s|0p|" passed
running test "|-o|+m|+s|1p|"
test "|-o|+m|+s|1p|" passed
running test "|-o|+m|-s|0p|"
test "|-o|+m|-s|0p|" passed
running test "|-o|+m|-s|1p|"
test "|-o|+m|-s|1p|" passed
running test "|-o|-m|+s|0p|"
test "|-o|-m|+s|0p|" passed
running test "|-o|-m|+s|1p|"
test "|-o|-m|+s|1p|" passed
running test "|-o|-m|-s|0p|"
test "|-o|-m|-s|0p|" passed
running test "|-o|-m|-s|1p|"
test "|-o|-m|-s|1p|" passed
running test "Last bytes of page"
test "Last bytes of page" passed
running test "Unaligned page cross only"
test "Unaligned page cross only" passed
````

earlephilhower (approved):

devyte (approved):

PR #7527, opened 2020-08-15
Netump Initial commit

hreintke:
This is a continuation from #6518

To ease tracking I first commit the same code, updates come with next commits.


hreintke:
Reply to comments @d-a-v @devyte 

On default constructor/deconstructor being generated by compiler. 
Need to include when there are also other (de)constructors, even when empty.

>PacketType destructor 
No need for this, if the destructor is the default (empty body in the cpp)

Done
>PacketType constructor 

Needed -> Not done

>ntoh16 
I meant: rather than reimplement here, why not call the lwip functions directly? or wrap the lwip calls in these methods.

These are netdump specific. wrapping lwip calls only complicates the code.
Not done

>NetdumpIP 
What I mean is that there doesn't seem to be an equivalent for IPv6, and also: is this constructor valid for the case of a sketch with only IPv6? If the answer to that is "no", then should the constructor have different args? have an overload with different args? something else?

NetdumpIP is created because lwip callback ans as such netdump always will have both IPv4 ans IPv6. Regardless of the lwip seting.
No change

>Netdump IP (de)constructor.

Constructor needed -> not done
Deconstructor -> done

>tcpBufferSize* (uppercase S)

Done

>Does it also not apply to WiFiClientSecure?

Do not understand your comment.

>Given that what comes next is an infinite loop, I suggest: "WiFiFailed, stopping sketch". I've been confused by these things myself.

Done

>The switch was changed, but the arg is still received as an int.

Done

earlephilhower:
This is too valuable to wait for minor tweaks.  Merging and will open an issue pointing to some of the comments herein.

PR #7530, opened 2020-08-16
Update PIO CI to be less chatty, fix CI fail

earlephilhower:


tests/platformio.sh
```diff
@@ -3,19 +3,21 @@
 cache_dir=$(mktemp -d)
 
 source "$TRAVIS_BUILD_DIR"/tests/common.sh
-set -ex
 
 function install_platformio()
 {
     pip3 install --user -U https://github.com/platformio/platformio/archive/develop.zip
     platformio platform install "https://github.com/platformio/platform-espressif8266.git"
```
mcspr, line 10: This *may* be fixed soon, as this is pretty recent git change in PIO repo :/ Which is pulled by using develop.zip, making each CI run different.
I wonder if it is better to target 4.3.4 release instead of develop branch for the time being

earlephilhower: Thanks for the idea!  PIO itself is dead with the latest branch from the devel repo anyway, not related to the JSON munging I removed.  Let me adjust the installer...

PR #7536, opened 2020-08-17
new w/ OOM now aborts by defaults, or throw an exception

d-a-v:
Changes:
- remove legacy `new` behavior, c++ `operator new` is supposed to ***never*** return `nullptr`
    - when exceptions are enabled: a catchable exception is raised
    - when exceptions are disabled (defaults): abort is called: caller address is showed for debuggers
- unused `arduino_new` has exactly the same behavior as `new (std::nothrow)`, removing the first.
- all source code with `new` which are checking the returned value are now calling `new (std::nothrow)` instead
  (the others will abort on oom, then reboot)
- `malloc` is untouched: it returns `nullptr` on oom like `new (std::nothrow)`


TD-er:
Can we also have a build option to deliberately keep the old situation?
Or else my code will cause crashes where it is checking the returned pointer as soon as this gets merged.
And right now I cannot change the code to call `new(std::nothrow)` as it is not yet merged.
I may need to make a macro for it or else my code will become quite messy while I still want to support old and new core libs.

d-a-v:
`new (std::nothrow)` is currently supposed to work. Did you try it ?

devyte:
@TD-er the whole point of this PR is to clean up the current situation and better conform to the C++ language standard. That means removing the legacy behavior.
Although this is expected to be merged soon, you have ample time to migrate, assuming you really need it. 
@d-a-v 's question is valid: new(std::nothrow) is supposed to work now. If it doesn't, We Need To Know.

TD-er:
I have not tried it yet.
I just assumed it wouldn't because this PR implemented it...

I will try to see if it compiles using the current core code and will start making a macro for it.

TD-er:
OK, next time I should try it first before panic.
It does compile just fine, and the changes do appear quite easy to do in my code (not that many `new` calls)

devyte:
Thanks for this, finally this is cleaned up!

libraries/ESP8266WiFi/src/CertStoreBearSSL.cpp
```diff
@@ -50,9 +50,13 @@ CertStore::CertInfo CertStore::_preprocessCert(uint32_t length, uint32_t offset,
...
   // Process it using SHA256, same as the hashed_dn
-  br_x509_decoder_context *ctx = new br_x509_decoder_context;
-  br_sha256_context *sha256 = new br_sha256_context;
+  br_x509_decoder_context *ctx = new (std::nothrow) br_x509_decoder_context;
+  br_sha256_context *sha256 = new (std::nothrow) br_sha256_context;
   if (!ctx || !sha256) {
+    if (ctx)
+      delete ctx;
+    if (sha256)
+      delete sha256;
```
mcspr, line 59: is `if (...)` redundant here? afaik, delete will be no-op with nullptr

d-a-v: We don't know which one is nullptr

mcspr: meaning, we can delete both. one nullptr, another created (or both nullptr)
i.e. https://godbolt.org/z/63PKa3

PR #7543, opened 2020-08-23
fixed bug in parsing POST file uploads

### LINKED ISSUE #7542: Bug in parsing file upload through http form
#### Platform
- Hardware:			ESP8285 / ESP8266
- Core Version:			2.7.4 (Package installed from Arduino IDE board manager)
- Operating System:		Windows

### Settings in IDE

- Flash Mode:			dout
- Flash Size: 	                1MB
- lwip Variant:                   v2 Lower Memory
- Reset Method:		nodemcu
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:               3000000

### Problem Description

The scenario is uploading a file through an HTML form. This uses a POST request to send multipart form data, although the files I'm using are small enough that they fit within one request buffer (i.e. <2048 bytes). The small program below accepts the file from the form and prints out the buffer contents. This usually works. For my application I am trying to upload new root certificates in X.509 format. This causes a problem because the file contains "-----END CERTIFICATE-----" at the end of the file. In multipart form data, the separator for fields is a line called "boundary," which starts with (at least) two dashes. For example, I'm using FireFox and each form field is separated by some unique string like "-----------------------------26263539015755059383526350366" Here is a  [link](https://github.com/esp8266/Arduino/blob/1ff927d04caa28e1294e94ec10edf758bec6edf1/libraries/ESP8266WebServer/src/Parsing-impl.h#L456) to the relevant section of the library. There is a bug in the logic checking if it is actually the boundary line. In my example I included an extra field below the file entry. When I print the file buffer, it includes the boundary string and the data of the next form entry. If I don't include the extra form entry, the upload hangs while the server waits for the end of the entry. 

What is going wrong in the algorithm: You notice the current line starts with two dashes and decide to check if it is the boundary line. You read in enough characters to check if it is the boundary string. However, the actual line is shorter than the boundary string, so you end up reading in the current line plus half of the next one (the actual boundary string). Then you see it does not match the boundary string, so you go back to reading the file. Again you notice two dashes, and read in the length of the boundary string. Since you were already halfway through the line containing the boundary, the stuff you read in doesn't fully match the boundary. So you continue on reading into the next entry until you hit the second entry's boundary, which is handled correctly.

How to fix it: Instead of reading the length of the boundary string, make sure to only read to the end of the current line.

### Example Sketch
Here is the file I used for testing: [DST_Root_CA_X3.pem](https://www.identrust.com/dst-root-ca-x3)
Don't forget to fill in your own WiFi credentials
```cpp
#include <ESP8266WiFi.h>
#include <WiFiClient.h>
#include <ESP8266WebServer.h>

ESP8266WebServer server(80);

const String htmlDoc = "<html>\
  <head>\
    <meta charset=\"UTF-8\">\
  </head>\
  <body>\
    <form method=\"post\" enctype=\"multipart/form-data\" action=\"/index.html\">\
      <label for=\"caCert\">CA root certificate:</label><br><input type=\"file\" id=\"caCert\" name=\"caCert\"><br>\
      <label for=\"ssid\">SSID:</label><input type=\"text\" id=\"ssid\" name=\"ssid\"><br>\
      <input type=\"submit\" value=\"Submit\">\
    </form>\
  </body>\
</html>";


void setup(void) {
  Serial.begin(115200);
  WiFi.begin("your-ssid", "your-password");
    delay(500);
  Serial.print("\n\rIP address: ");

  server.onFileUpload([]() { 
    if(server.upload().status == UPLOAD_FILE_WRITE)
      Serial.println(String((const char*)server.upload().buf)); 
  });
  
  server.on("/index.html", [](){server.send(200, "text/html", htmlDoc);});
  server.begin();
}

void loop(void) {
  server.handleClient();
}
```

### Debug Messages

```
New client
request: POST /index.html HTTP/1.1
method: POST url: /index.html search: 
headerName: Host
headerValue: 192.168.9.114
headerName: User-Agent
headerValue: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0
headerName: Accept
headerValue: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
headerName: Accept-Language
headerValue: en-US,en;q=0.5
headerName: Accept-Encoding
headerValue: gzip, deflate
headerName: Content-Type
headerValue: multipart/form-data; boundary=---------------------------32093566031658169028293882063
headerName: Content-Length
headerValue: 1573
headerName: Origin
headerValue: http://192.168.9.114
headerName: Connection
headerValue: keep-alive
headerName: Referer
headerValue: http://192.168.9.114/index.html
headerValue: 1
args: 
args count: 0
args: 
args count: 0
Parse Form: Boundary: ---------------------------32093566031658169028293882063 Length: 1573
PostArg FileName: DST_Root_CA_X3.pem
PostArg Name: caCert
PostArg Type: application/octet-stream
Start File: DST_Root_CA_X3.pem Type: application/octet-stream
-----BEGIN CERTIFICATE-----
MIIDSjCCAjKgAwIBAgIQRK+wgNajJ7qJMDmGLvhAazANBgkqhkiG9w0BAQUFADA/
MSQwIgYDVQQKExtEaWdpdGFsIFNpZ25hdHVyZSBUcnVzdCBDby4xFzAVBgNVBAMT
DkRTVCBSb290IENBIFgzMB4XDTAwMDkzMDIxMTIxOVoXDTIxMDkzMDE0MDExNVow
PzEkMCIGA1UEChMbRGlnaXRhbCBTaWduYXR1cmUgVHJ1c3QgQ28uMRcwFQYDVQQD
Ew5EU1QgUm9vdCBDQSBYMzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEB
AN+v6ZdQCINXtMxiZfaQguzH0yxrMMpb7NnDfcdAwRgUi+DoM3ZJKuM/IUmTrE4O
rz5Iy2Xu/NMhD2XSKtkyj4zl93ewEnu1lcCJo6m67XMuegwGMoOifooUMM0RoOEq
OLl5CjH9UL2AZd+3UWODyOKIYepLYYHsUmu5ouJLGiifSKOeDNoJjj4XLh7dIN9b
xiqKqy69cK3FCxolkHRyxXtqqzTWMIn/5WgTe1QLyNau7Fqckh49ZLOMxt+/yUFw
7BZy1SbsOFU5Q9D8/RhcQPGX69Wam40dutolucbY38EVAjqr2m7xPi71XAicPNaD
aeQQmxkqtilX4+U9m5/wAl0CAwEAAaNCMEAwDwYDVR0TAQH/BAUwAwEB/zAOBgNV
HQ8BAf8EBAMCAQYwHQYDVR0OBBYEFMSnsaR7LHH62+FLkHX/xBVghYkQMA0GCSqG
SIb3DQEBBQUAA4IBAQCjGiybFwBcqR7uKGY3Or+Dxz9LwwmglSBd49lZRNI+DT69
ikugdB/OEIKcdBodfpga3csTS7MgROSR6cz8faXbauX+5v3gTt23ADq1cEmv8uXr
AvHRAosZy5Q6XkjEGB5YGV8eAlrwDPGxrancWYaLbumR9YbK+rlmM6pZW87ipxZz
R8srzJmwN0jP41ZL9c8PDHIyh8bwRLtTcm1D9SZImlJnt1ir/md2cXjbDaJWFBM5
JDGFoqgCWjBH4d1QB7wCCZAA62RjYJsWvIjJEubSfZGL+T0yjWW06XyxV3bqxbYo
Ob8VZRzI9neWagqNdwvYkQsEjgfbKbYK7p2CNTUQ
-----END CERTIFICATE-----

-----------------------------32093566031658169028293882063
Content-Disposition: form-data; name="ssid"


End File: DST_Root_CA_X3.pem Type: application/octet-stream Size: 1329
Done Parsing POST
Request: /index.html
Arguments: 
final list of key/value pairs:
pm open,type:2 0
```

mjforanVT:
This PR fixes #7542. I rewrote the section of the Web Server library which handles files being uploaded via POST requests.


libraries/ESP8266WebServer/src/Parsing-impl.h
```diff
@@ -440,74 +441,61 @@ bool ESP8266WebServerTemplate<ServerType>::_parseForm(ClientType& client, const
...
+            _currentUpload->status = UPLOAD_FILE_END;
+            if (_currentHandler && _currentHandler->canUpload(_currentUri))
+                _currentHandler->upload(*this, _currentUri, *_currentUpload);
+            DBGWS("End File: %s Type: %s Size: %d\n",
+                _currentUpload->filename.c_str(),
+                _currentUpload->type.c_str(),
+                (int)_currentUpload->totalSize);
+            line = client.readStringUntil('\r');
+            client.readStringUntil('\n');
+            if (line == "--") {     // extra two dashes mean we reached the end of all form fields
```
TD-er, line 494: This now seems to be the only way to get out of this function.
Is it guaranteed there is this `--` line?
What if it got lost either due to lost connection, or buffers not long enough to receive all data?

libraries/ESP8266WebServer/src/Parsing-impl.h
```diff
@@ -440,74 +441,61 @@ bool ESP8266WebServerTemplate<ServerType>::_parseForm(ClientType& client, const
...
+            _currentUpload->totalSize += _currentUpload->currentSize;
+            _currentUpload->status = UPLOAD_FILE_END;
+            if (_currentHandler && _currentHandler->canUpload(_currentUri))
+                _currentHandler->upload(*this, _currentUri, *_currentUpload);
+            DBGWS("End File: %s Type: %s Size: %d\n",
+                _currentUpload->filename.c_str(),
+                _currentUpload->type.c_str(),
+                (int)_currentUpload->totalSize);
+            line = client.readStringUntil('\r');
+            client.readStringUntil('\n');
```
TD-er, line 493: Is it guaranteed in the specs there will be a `\r\n` terminating the line?  (thus making `\r` as line termination impossible)

mjforanVT: I couldn't find a specification that mandated line endings but the existing code used '\r\n' so that's what I did. I tested on Windows, Android, and Linux and there were no problems with the line endings.

libraries/ESP8266WebServer/src/Parsing-impl.h
mjforanVT: Yes, I believe '--' is the standard for ending multi-part form data so that line will always be there. There is a `if (!client.connected()) return _parseFormUploadAborted();` check before reading data from the client, so that's how the function will exit if the client disconnects. I just committed a few more of these lines that I missed the first time. I'm not sure what you mean about the buffers; large uploads will be processed in chunks of 2048 bytes. I have tested with files up to 100MB. They take a long time to upload but they don't cause any errors.

TD-er: Chunks of 2k will work OK I guess.
I've seen sometimes POST requests did not always have all data as it was somehow truncated or already re-used for something else.
But I guess for a file upload it may be dealt with differently.
It is just that I would also expect some timeout per transfer chunk, or else it may be stalled forever or WDT reboot if the connection is lost during transfer.

mjforanVT: I haven't looked in-depth at the rest of the code; I just used what was in place from the former implementation of form parsing. That being said, I don't think my code should cause any problems with timing out or missing data. We're only reading one byte at a time from the client, and each time we're checking if the client is still connected. If it disconnects then we abort the transfer. If the client data is being clobbered before we read it, that sounds like a bug somewhere else. This PR is just to fix the parsing bug. I haven't noticed any problems during my testing, but if you think there is a fundamental problem in how we handle POST data then we should work on that as a separate issue.

libraries/ESP8266WebServer/src/Parsing-impl.h
d-a-v: [HTTP/1.0 line break is CRLF](https://tools.ietf.org/html/rfc1945#section-2.2)

d-a-v (approved):
LGTM, nothing jumps at me after carefully reading.

libraries/ESP8266WebServer/src/Parsing-impl.h
```diff
@@ -442,74 +443,64 @@ bool ESP8266WebServerTemplate<ServerType>::_parseForm(ClientType& client, const
...
-                _uploadWriteByte((uint8_t)('-'));
-                _uploadWriteByte((uint8_t)('-'));
-                uint32_t i = 0;
-                while(i < boundary.length()){
-                  _uploadWriteByte(endBuf[i++]);
+                // current line does not contain boundary, upload all bytes in boundary buffer
+                for (int j = 0; j < i; j++)
+                    _uploadWriteByte(boundBuf[j]);
+                // the initial pass (filling up the boundary buffer) did not reach the end of the line. Upload the rest of the line now
+                if (i >= 2 + bLen) {
```
earlephilhower, line 473: I can only ever be <= 2+blen per the above for() loop, so this is really checking `if (i==2+blen)`.  It's not a big deal, but made me do a double-take as I thought there was an overflow condition...

earlephilhower (approved):
Strangely enough I was unable to reproduce your failure with `master` using the same PEM as you did, but I do see the issue you found in the existing code.  Things have to align poorly in the HTTP stream to hit it, I guess.

Thx for the PR!

libraries/ESP8266WebServer/src/Parsing-impl.h
mjforanVT: I have a habit of using <= or >= instead of == to avoid bugs where a variable gets incremented one too many times and skips the exit case for an infinite loop (for example). Maybe it is a bad habit if it confuses people reading my code!

earlephilhower: No, you're fine.  It's safer (protects against spurious bitflips, too) and logically correct!

PR #7544, opened 2020-08-23
Exception menu: by default new aborts on OOM

d-a-v:
Previously: new was returning nullptr after calling constructor on null object

replaces and closes #7536


d-a-v:
Closing in favor of #7536 which is now simplified, merged with this PR and consider the above review

PR #7546, opened 2020-08-23
Add CI test for eboot build

earlephilhower:
Check that building the eboot.c block generates the same binary as
the verison checked into the repo.  Catches the case where a library
or eboot.c file is changed, but an updated eboot.elf isn't included
in a PR.

Can't do simple binary diff of the ELFs because paths and compile
times will change, so dump the two sections we care about.


earlephilhower:
Verified CI fail when eboot.c changed w/o corresponding eboot.elf update.

PR #7547, opened 2020-08-24
Don't overwrite boot sector unless OTA changes it

earlephilhower:
There is a window where the eboot sector is erased and
unwritten/partially written.  If there's a power cycle at this time, the
chip will brick due to eboot being corrupted.

Avoid this by checking if the new eboot 4K sector is identical to the
one already in flash, and if so don't rewrite it.


davisonja:
Are we still looking at merging in the power-cycle-protection? #6538

devyte:
Yes, we're merging for v3 these days. We'll get there.

davisonja:
It's a two pronged query (though I hadn't realised it had an assigned
milestone).

Partly as this was one of the changes. that had been discussed to go into
#6538, and partly because I'm keen to get that PR merged, but it's going to
involve a bunch of changes from where it stands as the changes the PR
included when submitted now conflict with a small collection of more recent
changes (such as this PR would likely make).

Which probably boils down to looking for clarity on the order of things to
avoid having too many conflicts to resolve before #6538 can actually be
merged cleanly at all.

earlephilhower:
@davisonja, sorry, I actually thought #6538 was abandoned.  If you want to integrate this boot-block check there, I'm happy to axe this PR.  It should be orthogonal to your changes, though.

Can you spend a few mins to make #6538 mergeable?  It also has some issues w/the branch pulling in old versions of multiple libraries, so it's in kind of a weird state beyond just the code merge issues.

devyte:
@davisonja @earlephilhower #6538 needs to be fixed anyways, so I'll merge this now in the interest of moving forward.
About #6538, once it gets updated, the one thing we need is testing to make sure nothing was broken. Also, I'm not clear about forward/backwards compatibility, but let's discuss all that over there after the PR is updated.

PR #7551, opened 2020-08-26
Improve tests in Actions

ruggi99:
Fix tests for BearSSL in Actions.
Added exit() function to stop `Host tests` in case of errors in BearSSL_Validation.ino. Should affect only test.
Updated Server certificates of `api.github.com` in the sketch to pass the test. The certificate has been updated about 2 months ago, which means that tests should have always failed since then, but that didn't happened.
Added the new fingerprint and left there the old one to add a new test that should fail.


ruggi99:
There's an error in `fetchCertAuthority()` where the time should not be set initially, but it does.
Does it have anything to do with emulation?

d-a-v:
I restarted CI, it is now fine

ruggi99:
Converted to draft because I need help understading why time is already set before `setClock()` call.
In the code it has been ignored for the moment.
Also running test locally, time is already set.
I Could also add a test that should fail for Public Key.

d-a-v:
Time is automatically set when esp is dhcp client and when dhcp server provides an ntp server.
Check the ntp-tz-dst example to know how to disable these features.

earlephilhower:
As part of CI, this test isn't really testing that BearSSL *works* on the host, it's checking that there are no memory leaks, uninitialized variables, or random pointers in it.  The error code we're checking is from valgrind, not the INO.  So, while I appreciate the effort I think it's kind of wasted here.

Now, if BearSSL_Validatiaon.ino is not working *on the chip itself*, we do have an error and need to fix it.

earlephilhower:
@ruggi99 can you explain the reasoning for the PR?  This example is not a test, and it's not used as such in the host environment.  It's just a vehicle to verify that BearSSL wrapper code doesn't have mem leaks/out-of-bounds accesses.

ruggi99:
I opened this by mistake, because I wanted to open on my fork and play with Actions.
This is not a fix (title should be changed), rather an improvement of the test, so that it checks for any memory leaks, etc and also for the correct work of the SSL. This also applies to future changes to the library or SDK. It doesn't add workload.
It's extendible to all examples, to check that libraries are working correctly in the future, not only from a memory point of view

ruggi99:
Also every example has a sort of memory leak caused by `Serial` that it's not destructed at the end of the program.
Hovewer I fixed certificates in the SSL example that wasn't working.

libraries/ESP8266WiFi/examples/BearSSL_Validation/BearSSL_Validation.ino
```diff
@@ -196,15 +234,21 @@ can also be used.  ESP8266 time needs to be valid for checks to pass as
...
+  if (fetchURL(&client, host, port, path) != true) // Error here, time already set
+    ret = false;
+#endif
+  Serial.println("\nTry again after setting NTP time (should pass):\n");
   setClock();
-  fetchURL(&client, host, port, path);
+  if (fetchURL(&client, host, port, path) != true)
+    ret = false;
+  return ret;
 }
```
bchampp, line 252: why not just return fetchURL?

bchampp (changes requested):

PR #7553, opened 2020-08-27
WString: Optimize a bit

jjsuwa:
* move bodies of dtor, `init()` and `charAt()` to .h (implicitly inlined)

* unify the description of initialization into one: `init()` (literally), and called from each ctors, `invalidate()` and `move()`.

* invert the SSO state logic in order to make init state zeroed (as a result, inlined `init()` insns: 4 to 3)

IROM size @ FSBrowser: 304836 -> 304596 (saved 240 from master)

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -55,14 +55,6 @@ String::String(StringSumHelper &&rval) noexcept {
     move(rval);
 }
 
-String::String(char c) {
-    init();
-    char buf[2];
-    buf[0] = c;
-    buf[1] = 0;
-    *this = buf;
-}
-
 String::String(unsigned char value, unsigned char base) {
     init();
     char buf[1 + 8 * sizeof(unsigned char)];
@@ -91,7 +83,7 @@ String::String(unsigned int value, unsigned char base) {
 String::String(long value, unsigned char base) {
     init();
     char buf[2 + 8 * sizeof(long)];
-    if (base==10) {
+    if (base == 10) {
         sprintf(buf, "%ld", value);
     } else {
         ltoa(value, buf, base);
@@ -118,31 +110,21 @@ String::String(double value, unsigned char decimalPlaces) {
     *this = dtostrf(value, (decimalPlaces + 2), decimalPlaces, buf);
 }
 
-String::~String() {
-    invalidate();
-}
-
 /*********************************************/
 /*  Memory Management                        */
 /*********************************************/
 
-inline void String::init(void) {
-    setSSO(true);
-    setLen(0);
-    wbuffer()[0] = 0;
-}
-
 void String::invalidate(void) {
-    if(!isSSO() && wbuffer())
+    if (!isSSO() && wbuffer())
         free(wbuffer());
     init();
 }
 
 unsigned char String::reserve(unsigned int size) {
-    if(buffer() && capacity() >= size)
+    if (buffer() && capacity() >= size)
         return 1;
-    if(changeBuffer(size)) {
-        if(len() == 0)
+    if (changeBuffer(size)) {
+        if (len() == 0)
             wbuffer()[0] = 0;
         return 1;
     }
@@ -157,35 +139,32 @@ unsigned char String::changeBuffer(unsigned int maxStrLen) {
             uint16_t oldLen = len();
             setSSO(true);
             setLen(oldLen);
-            return 1;
         } else { // if bufptr && !isSSO()
             // Using bufptr, need to shrink into sso.buff
-            char temp[sizeof(sso.buff)];
-            memcpy(temp, buffer(), maxStrLen);
-            free(wbuffer());
+            const char *temp = buffer();
             uint16_t oldLen = len();
             setSSO(true);
             setLen(oldLen);
             memcpy(wbuffer(), temp, maxStrLen);
-            return 1;
+            free((void *)temp);
         }
+        return 1;
     }
     // Fallthrough to normal allocator
     size_t newSize = (maxStrLen + 16) & (~0xf);
     // Make sure we can fit newsize in the buffer
     if (newSize > CAPACITY_MAX) {
-        return false;
+        return 0;
     }
     uint16_t oldLen = len();
-    char *newbuffer = (char *) realloc(isSSO() ? nullptr : wbuffer(), newSize);
+    char *newbuffer = (char *)realloc(isSSO() ? nullptr : wbuffer(), newSize);
     if (newbuffer) {
         size_t oldSize = capacity() + 1; // include NULL.
         if (isSSO()) {
             // Copy the SSO buffer into allocated space
             memmove_P(newbuffer, sso.buff, sizeof(sso.buff));
         }
-        if (newSize > oldSize)
-        {
+        if (newSize > oldSize) {
             memset(newbuffer + oldSize, 0, newSize - oldSize);
         }
         setSSO(false);
@@ -201,7 +180,7 @@ unsigned char String::changeBuffer(unsigned int maxStrLen) {
 /*  Copy and Move                            */
 /*********************************************/
 
-String & String::copy(const char *cstr, unsigned int length) {
+String &String::copy(const char *cstr, unsigned int length) {
     if (!reserve(length)) {
         invalidate();
         return *this;
@@ -211,7 +190,7 @@ String & String::copy(const char *cstr, unsigned int length) {
     return *this;
 }
 
-String & String::copy(const __FlashStringHelper *pstr, unsigned int length) {
+String &String::copy(const __FlashStringHelper *pstr, unsigned int length) {
     if (!reserve(length)) {
         invalidate();
         return *this;
@@ -227,44 +206,35 @@ void String::move(String &rhs) noexcept {
     rhs.init();
 }
 
-String & String::operator =(const String &rhs) {
+String &String::operator =(const String &rhs) {
     if (this == &rhs)
         return *this;
-
     if (rhs.buffer())
         copy(rhs.buffer(), rhs.len());
     else
         invalidate();
-
     return *this;
 }
 
-String & String::operator =(String &&rval) noexcept {
+String &String::operator =(String &&rval) noexcept {
     if (this != &rval)
         move(rval);
     return *this;
 }
 
-String & String::operator =(StringSumHelper &&rval) noexcept {
-    if (this != &rval)
-        move(rval);
-    return *this;
-}
-
-String & String::operator =(const char *cstr) {
+String &String::operator =(const char *cstr) {
     if (cstr)
         copy(cstr, strlen(cstr));
     else
         invalidate();
-
     return *this;
 }
 
-String & String::operator = (const __FlashStringHelper *pstr)
-{
-    if (pstr) copy(pstr, strlen_P((PGM_P)pstr));
-    else invalidate();
-
+String &String::operator =(const __FlashStringHelper *pstr) {
+    if (pstr)
+        copy(pstr, strlen_P((PGM_P)pstr));
+    else
+        invalidate();
     return *this;
 }
 
@@ -285,7 +255,7 @@ unsigned char String::concat(const String &s) {
             return 0;
         memmove_P(wbuffer() + len(), buffer(), len());
         setLen(newlen);
-        wbuffer()[len()] = 0;
+        wbuffer()[newlen] = 0;
         return 1;
     } else {
         return concat(s.buffer(), s.len());
@@ -313,22 +283,17 @@ unsigned char String::concat(const char *cstr) {
 }
 
 unsigned char String::concat(char c) {
-    char buf[2];
-    buf[0] = c;
-    buf[1] = 0;
-    return concat(buf, 1);
+    return concat(&c, 1);
 }
 
 unsigned char String::concat(unsigned char num) {
     char buf[1 + 3 * sizeof(unsigned char)];
-    sprintf(buf, "%d", num);
-    return concat(buf, strlen(buf));
+    return concat(buf, sprintf(buf, "%d", num));
 }
 
 unsigned char String::concat(int num) {
     char buf[2 + 3 * sizeof(int)];
-    sprintf(buf, "%d", num);
-    return concat(buf, strlen(buf));
+    return concat(buf, sprintf(buf, "%d", num));
 }
 
 unsigned char String::concat(unsigned int num) {
@@ -339,8 +304,7 @@ unsigned char String::concat(unsigned int num) {
 
 unsigned char String::concat(long num) {
     char buf[2 + 3 * sizeof(long)];
-    sprintf(buf, "%ld", num);
-    return concat(buf, strlen(buf));
+    return concat(buf, sprintf(buf, "%ld", num));
 }
 
 unsigned char String::concat(unsigned long num) {
@@ -351,22 +315,25 @@ unsigned char String::concat(unsigned long num) {
 
 unsigned char String::concat(float num) {
     char buf[20];
-    char* string = dtostrf(num, 4, 2, buf);
+    char *string = dtostrf(num, 4, 2, buf);
     return concat(string, strlen(string));
 }
 
 unsigned char String::concat(double num) {
     char buf[20];
-    char* string = dtostrf(num, 4, 2, buf);
+    char *string = dtostrf(num, 4, 2, buf);
     return concat(string, strlen(string));
 }
 
-unsigned char String::concat(const __FlashStringHelper * str) {
-    if (!str) return 0;
+unsigned char String::concat(const __FlashStringHelper *str) {
+    if (!str)
+        return 0;
     int length = strlen_P((PGM_P)str);
-    if (length == 0) return 1;
+    if (length == 0)
+        return 1;
     unsigned int newlen = len() + length;
-    if (!reserve(newlen)) return 0;
+    if (!reserve(newlen))
+        return 0;
     memcpy_P(wbuffer() + len(), (PGM_P)str, length + 1);
     setLen(newlen);
     return 1;
@@ -376,79 +343,78 @@ unsigned char String::concat(const __FlashStringHelper * str) {
 /*  Concatenate                              */
 /*********************************************/
 
-StringSumHelper & operator +(const StringSumHelper &lhs, const String &rhs) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, const String &rhs) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(rhs.buffer(), rhs.len()))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, const char *cstr) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, const char *cstr) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!cstr || !a.concat(cstr, strlen(cstr)))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, char c) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, char c) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(c))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, unsigned char num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, unsigned char num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, int num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, int num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, unsigned int num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, unsigned int num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, long num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, long num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, unsigned long num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, unsigned long num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, float num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, float num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator +(const StringSumHelper &lhs, double num) {
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, double num) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(num))
         a.invalidate();
     return a;
 }
 
-StringSumHelper & operator + (const StringSumHelper &lhs, const __FlashStringHelper *rhs)
-{
-    StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+StringSumHelper &operator +(const StringSumHelper &lhs, const __FlashStringHelper *rhs) {
+    StringSumHelper &a = const_cast<StringSumHelper &>(lhs);
     if (!a.concat(rhs))
         a.invalidate();
     return a;
@@ -459,11 +425,11 @@ StringSumHelper & operator + (const StringSumHelper &lhs, const __FlashStringHel
 /*********************************************/
 
 int String::compareTo(const String &s) const {
-    if(!buffer() || !s.buffer()) {
-        if(s.buffer() && s.len() > 0)
-            return 0 - *(unsigned char *) s.buffer();
-        if(buffer() && len() > 0)
-            return *(unsigned char *) buffer();
+    if (!buffer() || !s.buffer()) {
+        if (s.buffer() && s.len() > 0)
+            return 0 - *(unsigned char *)s.buffer();
+        if (buffer() && len() > 0)
+            return *(unsigned char *)buffer();
         return 0;
     }
     return strcmp(buffer(), s.buffer());
@@ -521,7 +487,7 @@ unsigned char String::equalsConstantTime(const String &s2) const {
     //at this point lengths are the same
     if (len() == 0)
         return 1;
-    //at this point lenghts are the same and non-zero
+    //at this point lengths are the same and non-zero
     const char *p1 = buffer();
     const char *p2 = s2.buffer();
     unsigned int equalchars = 0;
@@ -541,19 +507,19 @@ unsigned char String::equalsConstantTime(const String &s2) const {
 }
 
 unsigned char String::startsWith(const String &s2) const {
-    if(len() < s2.len())
+    if (len() < s2.len())
         return 0;
     return startsWith(s2, 0);
 }
 
 unsigned char String::startsWith(const String &s2, unsigned int offset) const {
-    if(offset > (unsigned)(len() - s2.len()) || !buffer() || !s2.buffer())
+    if (offset > (unsigned)(len() - s2.len()) || !buffer() || !s2.buffer())
         return 0;
     return strncmp(&buffer()[offset], s2.buffer(), s2.len()) == 0;
 }
 
 unsigned char String::endsWith(const String &s2) const {
-    if(len() < s2.len() || !buffer() || !s2.buffer())
+    if (len() < s2.len() || !buffer() || !s2.buffer())
         return 0;
     return strcmp(&buffer()[len() - s2.len()], s2.buffer()) == 0;
 }
@@ -562,16 +528,12 @@ unsigned char String::endsWith(const String &s2) const {
 /*  Character Access                         */
 /*********************************************/
 
-char String::charAt(unsigned int loc) const {
-    return operator[](loc);
-}
-
 void String::setCharAt(unsigned int loc, char c) {
     if (loc < len())
         wbuffer()[loc] = c;
 }
 
-char & String::operator[](unsigned int index) {
+char &String::operator[](unsigned int index) {
     static char dummy_writable_char;
     if (index >= len() || !buffer()) {
         dummy_writable_char = 0;
@@ -596,7 +558,7 @@ void String::getBytes(unsigned char *buf, unsigned int bufsize, unsigned int ind
     unsigned int n = bufsize - 1;
     if (n > len() - index)
         n = len() - index;
-    strncpy((char *) buf, buffer() + index, n);
+    strncpy((char *)buf, buffer() + index, n);
     buf[n] = 0;
 }
 
@@ -604,29 +566,13 @@ void String::getBytes(unsigned char *buf, unsigned int bufsize, unsigned int ind
 /*  Search                                   */
 /*********************************************/
 
-int String::indexOf(char c) const {
-    return indexOf(c, 0);
-}
-
 int String::indexOf(char ch, unsigned int fromIndex) const {
     if (fromIndex >= len())
         return -1;
-    const char* temp = strchr(buffer() + fromIndex, ch);
+    const char *temp = strchr(buffer() + fromIndex, ch);
     if (temp == NULL)
         return -1;
     return temp - buffer();
-}
-
-int String::indexOf(const __FlashStringHelper *s2) const {
-    return indexOf(s2, 0);
-}
-
-int String::indexOf(const __FlashStringHelper *s2, unsigned int fromIndex) const {
-    return indexOf((const char*) s2, fromIndex);
-}
-
-int String::indexOf(const char *s2) const {
-    return indexOf(s2, 0);
 }
 
 int String::indexOf(const char *s2, unsigned int fromIndex) const {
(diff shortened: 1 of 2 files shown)


cores/esp8266/WString.cpp
```diff
@@ -203,24 +204,20 @@ unsigned char String::changeBuffer(unsigned int maxStrLen) {
...
     setLen(length);
     memmove_P(wbuffer(), cstr, length + 1);
-    return *this;
 }
 
-String & String::copy(const __FlashStringHelper *pstr, unsigned int length) {
+void String::copy(const __FlashStringHelper *pstr, unsigned int length) {
     if (!reserve(length)) {
         invalidate();
-        return *this;
```
earlephilhower, line 219: This change breaks the copy by letting a `return` fall through in the PR.

cores/esp8266/WString.cpp
```diff
@@ -203,24 +204,20 @@ unsigned char String::changeBuffer(unsigned int maxStrLen) {
 // /*  Copy and Move                            */
 // /*********************************************/
 
-String & String::copy(const char *cstr, unsigned int length) {
+void String::copy(const char *cstr, unsigned int length) {
     if (!reserve(length)) {
         invalidate();
-        return *this;
```
earlephilhower, line 209: Breaks the code with a fallthru instead of a return.

cores/esp8266/WString.h
```diff
@@ -314,8 +313,8 @@ class String {
         unsigned char changeBuffer(unsigned int maxStrLen);
 
         // copy and move
-        String & copy(const char *cstr, unsigned int length);
-        String & copy(const __FlashStringHelper *pstr, unsigned int length);
+        void copy(const char *cstr, unsigned int length);
```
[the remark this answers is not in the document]
jjsuwa: >  The return value comes from the this pointer which is already in a register, so I don't think this saves anything code/runtime-wise.

really saves 1 insn, eg. `mov.n  a2, a12` (copy `this` to retval-reg)

cores/esp8266/WString.cpp
```diff
@@ -58,70 +58,71 @@ String::String(StringSumHelper &&rval) {
 #endif
 
 String::String(char c) {
+    uint32_t buf = (uint32_t)c;
```
yumkam: This assumes little-endian. Yeap, sure, there are no such thing as big-endian esp8266; but then someone will copy-paste this code, and kitten will die. It worth at least leaving comment on this dependency (and intent: alignment).
More alternatives - `const char buf __attribute__((__aligned__(4))) = c;` or `union {const char c; uint32_t align; } buf = {c};`, YMMV.

earlephilhower:
Can you please give some hard #s on the size difference for a few examples using this vs. master, and the runtime differences, too?

You could do a build and then `objdump -T` to get the function sizes to do it at a mcro-level, or you could build FSBrowser.ino which makes extensive use of Strings and check total bin size.

jjsuwa:
`FSBrowser.ino` build results:

- commit e53aa87160ad26d45e06a5f5fa3474b67f7f4642 (master in Aug 26, 2020):
  FSBrowser.ino.bin : 338304 bytes
    - IROM   : 305020
    - IRAM   : 26340
    - DATA   : 1272
    - RODATA : 1512
    - BSS    : 25576
  `objdump -t` (excerpt) :
```
00000008 g     F .text._ZN6StringC2Ec	00000031 _ZN6StringC2Ec
00000008 g     F .text._ZN6StringC2Ec	00000031 _ZN6StringC1Ec
0000000c g     F .text._ZN6StringC2Ehh	0000003d _ZN6StringC2Ehh
0000000c g     F .text._ZN6StringC2Ehh	0000003d _ZN6StringC1Ehh
00000014 g     F .text._ZN6StringC2Eih	0000004f _ZN6StringC2Eih
00000014 g     F .text._ZN6StringC2Eih	0000004f _ZN6StringC1Eih
0000000c g     F .text._ZN6StringC2Ejh	0000003b _ZN6StringC2Ejh
0000000c g     F .text._ZN6StringC2Ejh	0000003b _ZN6StringC1Ejh
00000014 g     F .text._ZN6StringC2Elh	0000004f _ZN6StringC2Elh
00000014 g     F .text._ZN6StringC2Elh	0000004f _ZN6StringC1Elh
0000000c g     F .text._ZN6StringC2Emh	0000003b _ZN6StringC2Emh
0000000c g     F .text._ZN6StringC2Emh	0000003b _ZN6StringC1Emh
00000010 g     F .text._ZN6StringC2Efh	0000005b _ZN6StringC2Efh
00000010 g     F .text._ZN6StringC2Efh	0000005b _ZN6StringC1Efh
0000000c g     F .text._ZN6StringC2Edh	0000004e _ZN6StringC2Edh
0000000c g     F .text._ZN6StringC2Edh	0000004e _ZN6StringC1Edh
00000004 g     F .text._ZN6StringD2Ev	00000013 _ZN6StringD1Ev
00000004 g     F .text._ZN6StringD2Ev	00000013 _ZN6StringD2Ev
00000010 g     F .text._ZN6String4copyEPKcj	0000005b _ZN6String4copyEPKcj
00000010 g     F .text._ZN6String4copyEPK19__FlashStringHelperj	0000005b _ZN6String4copyEPK19__FlashStringHelperj
```

- commit f4352b1d72180401f0669054f860d150cd025712 (this PR, latest):
  FSBrowser.ino.bin : 338320 bytes
    - IROM   : 305036
    - IRAM   : 26340
    - DATA   : 1272
    - RODATA : 1512
    - BSS    : 25576
  `objdump -t` (excerpt) :
```
00000008 g     F .text._ZN6StringC2Ec	0000002a _ZN6StringC2Ec
00000008 g     F .text._ZN6StringC2Ec	0000002a _ZN6StringC1Ec
00000010 g     F .text._ZN6StringC2Ehh	0000003d _ZN6StringC2Ehh
00000010 g     F .text._ZN6StringC2Ehh	0000003d _ZN6StringC1Ehh
00000018 g     F .text._ZN6StringC2Eih	00000051 _ZN6StringC2Eih
00000018 g     F .text._ZN6StringC2Eih	00000051 _ZN6StringC1Eih
00000010 g     F .text._ZN6StringC2Ejh	0000003b _ZN6StringC2Ejh
00000010 g     F .text._ZN6StringC2Ejh	0000003b _ZN6StringC1Ejh
00000018 g     F .text._ZN6StringC2Elh	00000051 _ZN6StringC2Elh
00000018 g     F .text._ZN6StringC2Elh	00000051 _ZN6StringC1Elh
00000010 g     F .text._ZN6StringC2Emh	0000003b _ZN6StringC2Emh
00000010 g     F .text._ZN6StringC2Emh	0000003b _ZN6StringC1Emh
00000014 g     F .text._ZN6StringC2Efh	00000055 _ZN6StringC2Efh
00000014 g     F .text._ZN6StringC2Efh	00000055 _ZN6StringC1Efh
00000010 g     F .text._ZN6StringC2Edh	00000045 _ZN6StringC2Edh
00000010 g     F .text._ZN6StringC2Edh	00000045 _ZN6StringC1Edh
00000004 g     F .text._ZN6StringD2Ev	0000001c _ZN6StringD1Ev
00000004 g     F .text._ZN6StringD2Ev	0000001c _ZN6StringD2Ev
00000010 g     F .text._ZN6String4copyEPKcj	00000059 _ZN6String4copyEPKcj
00000010 g     F .text._ZN6String4copyEPK19__FlashStringHelperj	00000059 _ZN6String4copyEPK19__FlashStringHelperj
```

devyte:
That shows that your optimization attempts are going the wrong way: 16 bytes increase in IROM.

jjsuwa:
- commit cb6dca9df0e677f336f842660de1a61eed705d75 (this PR: latest):
  FSBrowser.ino.bin : 338288 bytes
    - IROM   : 305004
    - IRAM   : 26340
    - DATA   : 1272
    - RODATA : 1512
    - BSS    : 25576
  `objdump -t` (excerpt) :
```
00000008 g     F .text._ZN6StringC2Ec	00000029 _ZN6StringC2Ec
00000008 g     F .text._ZN6StringC2Ec	00000029 _ZN6StringC1Ec
0000000c g     F .text._ZN6StringC2Ehh	00000033 _ZN6StringC2Ehh
0000000c g     F .text._ZN6StringC2Ehh	00000033 _ZN6StringC1Ehh
00000014 g     F .text._ZN6StringC2Eih	00000047 _ZN6StringC2Eih
00000014 g     F .text._ZN6StringC2Eih	00000047 _ZN6StringC1Eih
0000000c g     F .text._ZN6StringC2Ejh	00000031 _ZN6StringC2Ejh
0000000c g     F .text._ZN6StringC2Ejh	00000031 _ZN6StringC1Ejh
00000014 g     F .text._ZN6StringC2Elh	00000047 _ZN6StringC2Elh
00000014 g     F .text._ZN6StringC2Elh	00000047 _ZN6StringC1Elh
0000000c g     F .text._ZN6StringC2Emh	00000031 _ZN6StringC2Emh
0000000c g     F .text._ZN6StringC2Emh	00000031 _ZN6StringC1Emh
00000010 g     F .text._ZN6StringC2Efh	0000004b _ZN6StringC2Efh
00000010 g     F .text._ZN6StringC2Efh	0000004b _ZN6StringC1Efh
0000000c g     F .text._ZN6StringC2Edh	0000003b _ZN6StringC2Edh
0000000c g     F .text._ZN6StringC2Edh	0000003b _ZN6StringC1Edh
00000004 g     F .text._ZN6StringD2Ev	0000001c _ZN6StringD2Ev
00000004 g     F .text._ZN6StringD2Ev	0000001c _ZN6StringD1Ev
00000010 g     F .text._ZN6String4copyEPKcj	00000059 _ZN6String4copyEPKcj
00000010 g     F .text._ZN6String4copyEPK19__FlashStringHelperj	00000059 _ZN6String4copyEPK19__FlashStringHelperj
```

cores/esp8266/WString.h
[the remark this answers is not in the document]
devyte: The copy() methods are protected, so only derived classes would have access to them anyways. That's too limited a use case imho, so I suspect it's ok to lose this.

cores/esp8266/WString.h
```diff
@@ -279,7 +278,7 @@ class String {
 
     protected:
         // Contains the string info when we're not in SSO mode
-        struct _ptr { 
+        struct _ptr {
             char *   buff;
```
TD-er, line 282: Just curious.
Why are those members not initialized to null_ptr and 0?

I guess if the allocation of a buffer fails, it could be anything in there and when the destructor is called it will try to destruct something that's not to be destructed -> crash.
As we have seen while using the std::list emplace function, this could happen.
My work around for now is to emplace an empty object and then fill it while being in the list already.

earlephilhower: @TD-er, I'm not following you here.

This is 1/2 of the union{} and neither this struct nor the _sso one have any C++-type new/delete operations associated with them.  Iff the string was not in SSO mode, this buff will be freed in the ~String().  The union is initted in the String() constructor.

TD-er: @earlephilhower Every now and then I'm still looking for the explanation why the String class caused crashes when being destructed like I showed you before.
And the only thing I can think of is that there must be some possible code path where sso is set to false but the buffer pointer is not set due to OOM allocation while the String object is constructed.
So far I have not found any possible code path that could lead to this, but apparently there must be one as it otherwise could not cause these crashes while being destructed on members of a larger struct created using the emplace function of a list.
But even though I cannot find the code path to it, it would make sense for the buffer pointer to always being set to `nullptr` for every occasion where sso is set.

Setting it at the construction of the `_prt` struct is one thing and I agree with you that it is no guarantee it will be correct as it shares the memory with the `_sso` struct.

So maybe this function could be extended to something like this:
```c++
inline void setSSO(bool set) { 
    if (sso.isSSO && !set) {
        ptr.buff = nullptr;
    }
    sso.isSSO = set;
}
```

But on the other hand the only places where `setSSO(false)` is called, also the `setBuffer` is set.
Of those very few calls, only the call to `changeBuffer` can be slightly suspicious.
In which a call to `realloc` is made which should return either `nullptr` or a valid pointer.


Maybe a bit long way of me to show I don't like the open end where I simply cannot explain the crashes I had and have been working around them since. But I am always afraid I will one day forget to do so or merge a PR which doesn't take it into account and thus run into those hard to reproduce issues again.
So still looking at each and every change made to the String class to hopefully one day see something that may explain those issues.

earlephilhower: This SSO optimization was moved into the ESP32 repo as well, and they don't seem to be reporting issues related to OOM cases.  Not to say there's no bugs, but if there are we don't seem to be hitting them in other folks' code.

`free()` is an interesting beast because UMM doesn't keep anything like a list of allocated blocks.  They're just a few bits in memory right before the returned `malloc()` value.  So if there has been memory overwrite of even a byte past the end of the previous pointer, really weird things can happen.  But that's something you'd need `valgrind` or similar to detect at runtime (and that's not gonna run on the 8266!).

If you think there should be a reset in String, drop a PR and we can see about it.  Actually, if you think that's the issue it might be worthwhile to change your own version and run your code to see if it fixes the crashes.  That would be a good smoking gun test case (although it would be best to find and repro a failing case in the host unit tests, of course).

Good luck!

TD-er: Just a thought.

The issues I saw were with (multiple) String objects as members of a larger struct.
What would happen if the OOM occurred when constructing this larger struct?
What would the state be of those String objects? Has the constructor run?
If the constructor has not yet run, the init() function is not called, but the members are there.
So would it be useful to at least intialize the `isSSO` member of the `_sso` struct?
Like this:
```c++
        struct _sso {
            char     buff[SSOSIZE] = { 0 };
            unsigned char len   : 7 = 0; // Ensure only one byte is allocated by GCC for the bitfields
            unsigned char isSSO : 1 = 1;
        } __attribute__((packed)); // Ensure that GCC doesn't expand the flag byte to a 32-bit word for alignment issues
```

This to make sure the desctructor is not doing funny things as it needs `isSSO` to be 0 to call `free` on the buffer.
```c++
String::~String() {
    if(!isSSO() && wbuffer())
        free(wbuffer());
}
```

jjsuwa: @TD-er:
> Like this:

+1
this also eliminates explicit `init()` from constructors

cores/esp8266/WString.h
```diff
@@ -284,17 +284,17 @@ class String {
...
             char     buff[SSOSIZE];
             unsigned char len   : 7; // Ensure only one byte is allocated by GCC for the bitfields
             unsigned char isSSO : 1;
         } __attribute__((packed)); // Ensure that GCC doesn't expand the flag byte to a 32-bit word for alignment issues
-        enum { CAPACITY_MAX = 65535 }; // If typeof(cap) changed from uint16_t, be sure to update this enum to the max value storable in the type
         union {
             struct _ptr ptr;
             struct _sso sso;
         };
+        static constexpr auto CAPACITY_MAX = (1U << sizeof(ptr.cap) * 8) - 1;
```
TD-er, line 300: This is not used in the .h file, so why not move it to the .cpp?

TD-er:
Is the `init()` function still needed?
Its declaration is already protected and it is only used now in the invalidate function.

cores/esp8266/WString.h
jjsuwa: > (it is protected, so only those inheriting from this class)

just only, but still possible.
if so, we should warn: [API change] (of course `.isNotSSO` too)

i'd like to know why these internal implementation members are protected but not private.

TD-er: Well the `init()` function was really needed for proper functioning of the class, so if someone wants to make a derived class, this function must be protected (or public).
I can imagine it is not needed for using the String class, so then protected makes perfect sense.
Now init() is no longer needed in the current implementation, but it may indeed break API design if the function is removed. The same for the `CAPACITY_MAX` value.

cores/esp8266/WString.cpp
```diff
@@ -159,7 +159,8 @@ unsigned char String::changeBuffer(unsigned int maxStrLen) {
     // Fallthrough to normal allocator
     size_t newSize = (maxStrLen + 16) & (~0xf);
     // Make sure we can fit newsize in the buffer
-    if (newSize > CAPACITY_MAX) {
+    size_t maxCapacity = (1U << sizeof(ptr.cap) * 8) - 1;
```
devyte, line 162: Please be careful when refactoring, the original definition was static constexpr auto. At least make this constexpr.

jjsuwa: thanks for the advice, @devyte 

i have periodically checked the disassembled binaries using `objdump -d`.
in this case, the same as before.

cores/esp8266/WString.h
```diff
@@ -275,41 +274,42 @@ class String {
...
         };
         // This allows strings up up to 11 (10 + \0 termination) without any extra space.
-        enum { SSOSIZE = sizeof(struct _ptr) + 4 - 1 }; // Characters to allocate space for SSO, must be 12 or more
+        static constexpr auto SSOSIZE = sizeof(struct _ptr) + 4 - 1; // Characters to allocate space for SSO, must be 12 or more
         struct _sso {
             char     buff[SSOSIZE];
-            unsigned char len   : 7; // Ensure only one byte is allocated by GCC for the bitfields
-            unsigned char isSSO : 1;
+            unsigned char len      : 7; // Ensure only one byte is allocated by GCC for the bitfields
+            unsigned char isNotSSO : 1;
```
devyte, line 291: What is this? Why the change to reverse logic? I see in addition that this is used as if(! isSSO()) in code, so this is a reverse of a reversed name that reverses a bit.
If there is a good reason for reversing, then please change the name to something that reflects the new meaning without the Not. Having to look at things like e. g. !isSSO() and then bool isSSO() {return !isNotSSO;} is just painful.
And I won't even mention the setSSO() body ....

cores/esp8266/WString.h
```diff
@@ -275,41 +274,42 @@ class String {
...
         // Contains the string info when we're not in SSO mode
-        struct _ptr { 
+        struct _ptr {
             char *   buff;
             uint16_t cap;
             uint16_t len;
         };
         // This allows strings up up to 11 (10 + \0 termination) without any extra space.
-        enum { SSOSIZE = sizeof(struct _ptr) + 4 - 1 }; // Characters to allocate space for SSO, must be 12 or more
+        static constexpr auto SSOSIZE = sizeof(struct _ptr) + 4 - 1; // Characters to allocate space for SSO, must be 12 or more
```
devyte, line 287: The reason for the enum was that we were certain that way didn't create a constant with the value with old gcc compiler, even when using the value in code. We checked with sizeof and looked at the bin dump.
Are you positive that there is *no* case where this new definition creates a constant that increases the mem footprint?

cores/esp8266/WString.h
```diff
@@ -275,41 +274,42 @@ class String {
...
         struct _sso {
             char     buff[SSOSIZE];
-            unsigned char len   : 7; // Ensure only one byte is allocated by GCC for the bitfields
-            unsigned char isSSO : 1;
+            unsigned char len      : 7; // Ensure only one byte is allocated by GCC for the bitfields
+            unsigned char isNotSSO : 1;
+            _sso() : len(0), isNotSSO(0) {
```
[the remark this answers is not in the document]
jjsuwa: direct member init for bitfield requires c++20, will bring some of CI tests failure (possibly host emulation)

cores/esp8266/WString.cpp
```diff
@@ -30,50 +30,40 @@
...
 }
 #endif
 
 String::String(char c) {
-    init();
-    char buf[2];
-    buf[0] = c;
-    buf[1] = 0;
-    *this = buf;
+    uint16_t buf = __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__ ? (uint16_t)c << 8 : (uint16_t)c;
```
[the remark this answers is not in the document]
jjsuwa: ok, i understand that (but, sorry, hard to satisfy...)

cores/esp8266/WString.h
jjsuwa: as you indicated, this change has no performance effect (either better or worse).
according to the principle of minimum action, leaving it alone is probably the right way, i understood.

cores/esp8266/WString.cpp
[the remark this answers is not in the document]
devyte: Why hard? To make sure there is no misunderstanding:
```cpp
uint16_t buf = __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ ? 
(uint16_t)c : (uint16_t)c << 8;
```

cores/esp8266/WString.h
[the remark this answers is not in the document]
devyte: That's not what I meant. You should be able to do direct member init below in the union in some form or other.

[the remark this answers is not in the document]
TD-er: For bitfield members that's not possible, as that's one of the new features of C++20.
So for this special case of the bitfields you can only initialize them in the constructor of the `_sso` struct.

cores/esp8266/WString.h
jjsuwa: hmm, `unsigned char disabled : 1;` or something, might be unambiguous.

TD-er: If you need a better term, perhaps "heapBuffer" ? (`isSSO` was about not using a heap allocated buffer)

devyte: Before moving forward with any change, please explain why it needs to change at all.

cores/esp8266/WString.h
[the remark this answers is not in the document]
devyte: As a quick test:
```cpp
  union
  {
    _sso sso = {{0},{0}};
    _ptr ptr;
  };
```

works fine with C++11.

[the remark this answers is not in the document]
jjsuwa: thx @devyte:

that's surely fine, in terms of the logic: wholly zeroed, due to funccall of `memset(this, 0, sizeof(_sso))`.
zeroing `_sso.buff[0]` and `_sso.`bitfields would be sufficient (and of course most efficient).

both class member initializers (`char member = 0;`) and braces initialization (`struct _foo foo = { 0, ... };`) cause whole initialization
(assumed that each of unspecified members has default value).

member initializer (`: bar(0), ...`) seems to be only way to realize sparse(holed) initialization.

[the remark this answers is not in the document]
devyte: The sso struct is 12 bytes in size, yet you seem keen on doing microoptimizations here. I don't consider this worthwhile, the only case where this would be significant is in a tight loop where >90% of the time is spent in initing a String object, and in such a case I would likely yell at the author. Personally, I think there are other places where spending time and effort would bring far better returns. 
If you decide to pursue this, please:
- find bin size for initialization with {{0},{0}}
- compare against bin size with your member initializer
- compare runtime performance of both cases
- report back here

BTW, we're already on gcc 10.x. Although we have C++17 enabled, in theory we have access to C++20 experimental features.

cores/esp8266/WString.cpp
```diff
@@ -30,50 +30,40 @@
...
 }
 #endif
 
 String::String(char c) {
-    init();
-    char buf[2];
-    buf[0] = c;
-    buf[1] = 0;
-    *this = buf;
+    uint16_t buf = __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ ? (uint16_t)c : (uint16_t)c << 8;
```
devyte, line 56: This looks duplicated, please refactor.

cores/esp8266/WString.cpp
```diff
@@ -342,10 +328,8 @@ unsigned char String::concat(const char *cstr) {
 }
 
 unsigned char String::concat(char c) {
-    char buf[2];
-    buf[0] = c;
-    buf[1] = 0;
-    return concat(buf, 1);
+    uint16_t buf = __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ ? (uint16_t)c : (uint16_t)c << 8;
```
devyte, line 331: Duplicate, please refactor

cores/esp8266/WString.h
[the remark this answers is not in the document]
TD-er: > BTW, we're already on gcc 10.x. Although we have C++17 enabled, in theory we have access to C++20 experimental features.

It was already mentioned before, this class is also ported to ESP32.
Are they also using gcc 10.x?
To be honest, if the braces initialization also works, then I would say to stick with what's working and not forcing to only be able to compile with the newest C++ standard.

What I'm more worried about is how the initialization is done when in a union.
In other words, what will be guaranteed to be initialized with the isSSO set to 1 (or the inverted one set to 0)?
We can talk about semantics and how the code may be the most aestetic pleasing to the reader.
But in the end it is important this one flag is always default set to mean `isSSO` is active.
If that should be done via automatic zero'ed initialization (thus inverting the flag) or the suggested:
```c++
  union
  {
    _sso sso = {{0},{0}};
    _ptr ptr;
  };
```
That's all fine by me.
Most important is that this `isSSO` is in its 'enabled' state (which isn't with this suggested code by the way) as default member value.

earlephilhower:
Guys, I'm seeing this PR morphing into two things.
1. An explicit init of the SSO/PTR union to catch potential OOM invalid frees (ie. "bug fix")
2. Some tweaks to the constructors to save +/- a few bytes but with no logical changes

Number 1 seems like a fine change and should be a 2-line PR, while number 2 is big and introduces issues of taste ans style.

Maybe the best way forward is to drop any functional changes from this and open a new one with only the "init sso/ptr union" change?  That init PR should be easily approvable and get us runtime in end users to see if it has any effect.

devyte:
I agree, let's split the "bug fix" off into its own pr to discuss.

cores/esp8266/WString.cpp
dirkmueller: I think also that most compilers perform this optimisation automatically. there might not be an observable behavior difference between old and new code.

cores/esp8266/WString.h
dirkmueller: I think its easy: the default constructors are setting "SSO" to true. by making the default "false" or "0", the initialization can be similified because almost all compilers and all architectures have some kind of loophole optimisation to find the best possible code pattern for initializing a variable or a set of variables to zero. 

I think instead of using "SSO" which is an acronym that is hard to understand in the first place, I would suggest to call it "HeapString" or "isHeapString" or "isLargeString" . "isnotSSO" which is "is not small string optimisation" is really awkward english.

cores/esp8266/WString.h
```diff
@@ -288,19 +292,24 @@ class String {
...
-            unsigned char isSSO : 1;
+            unsigned char len    : 7; // Ensure only one byte is allocated by GCC for the bitfields
+            unsigned char isHeap : 1;
         } __attribute__((packed)); // Ensure that GCC doesn't expand the flag byte to a 32-bit word for alignment issues
         enum { CAPACITY_MAX = 65535 }; // If typeof(cap) changed from uint16_t, be sure to update this enum to the max value storable in the type
         union {
+            struct {
+                char _offset0 = 0; // wbuffer()[0] = 0;
+                char _offset1[SSOSIZE - 1];
+                char _offset11 = 0; // setSSO(true); setLen(0);
```
TD-er, line 303: I get why you want to do this, but I think the naming is a bit strange here.
Also, why not just use a constructor in one (not both) of the existing structs `_ptr` or `_sso` ?

I think `_sso` is the best choice here for having the constructor to zero all values.
And since it has to be done in the body of the constructor anyway, you can also make a function in that struct and call it both from its constructor and from the `String::init()` function.

As done here, using an unnamed struct with `_offsetNN` names is confusing.

cores/esp8266/WString.h
jjsuwa: See #7590

cores/esp8266/WString.h
```diff
@@ -68,7 +71,9 @@ class String {
         explicit String(unsigned long, unsigned char base = 10);
         explicit String(float, unsigned char decimalPlaces = 2);
         explicit String(double, unsigned char decimalPlaces = 2);
-        ~String(void);
+        inline ~String() {
```
TD-er, line 74: Just curious, but what extra does `inline` add to a function implemented in a .h file?
I always thought implementing a function in a .h file always implies `inline`.

jjsuwa: as you said, function bodies inside a class definition (typically in .h) are implicitly marked as `inline`.

like extra arithmetic parentheses, may be removed later.

mcspr (approved):
@jjsuwa also note of the #7611 / d3b3a432fee29484d387c9e3b8392399c6792470

Adding this test to **tests/host/core/test_string.cpp** fails on master, but not with this patch:
```cpp
...

TEST_CASE("String::move", "[core][String]")
{
    const char buffer[] = "this string goes over the sso limit";

    String target;
    String source(buffer);

    target = std::move(source);
    REQUIRE(source.c_str() != nullptr);
    REQUIRE(target == buffer);
}

...
```
Perhaps you could add this test (or my commit) here?

re ALL: all existing change requests seem resolved, please re-review? I do believe initialization issue discussed above was resolved through the use of `init()`. Naming might not, but idk if it is still relevant.

earlephilhower:
What is the new code size difference with these changes, @jjsuwa ?

The changing polarity of the `isSSO` bit into `isHeap` does not affect any actual code, it's just an unnecessary change IMO which makes things obtuse since the code is still using the name `SSO` in method names.  Saving 3-4 bytes of *flash* would not be worth that obfuscation IMO (at the level of build noise).  If the change saves 100 bytes or more, then it's meaningful.

cores/esp8266/WString.cpp
```diff
@@ -225,32 +215,9 @@ String & String::copy(const __FlashStringHelper *pstr, unsigned int length) {
...
-        setBuffer(rhs.wbuffer());
-    }
-    setCapacity(rhs.capacity());
-    setLen(rhs.len());
-    rhs.setSSO(false);
-    rhs.setCapacity(0);
-    rhs.setLen(0);
-    rhs.setBuffer(nullptr);
+    invalidate();
+    sso = rhs.sso; // memcpy(this, &rhs, sizeof(*this));
```
earlephilhower, line 219: I think this comment would be best removed.

cores/esp8266/WString.cpp
```diff
@@ -225,32 +215,9 @@ String & String::copy(const __FlashStringHelper *pstr, unsigned int length) {
 
 #ifdef __GXX_EXPERIMENTAL_CXX0X__
 void String::move(String &rhs) noexcept {
```
earlephilhower, line 217: Very nice cleanup of code here!

earlephilhower (changes requested):
Can you also please add in @mcspr 's add'l String test which passes w/the fixed/minimized `::move()`, too?

cores/esp8266/WString.h
```diff
@@ -287,19 +291,19 @@ class String {
...
         // Accessor functions
-        inline bool isSSO() const { return sso.isSSO; }
+        inline bool isSSO() const { return !sso.isHeap; }
         inline unsigned int len() const { return isSSO() ? sso.len : ptr.len; }
         inline unsigned int capacity() const { return isSSO() ? (unsigned int)SSOSIZE - 1 : ptr.cap; } // Size of max string not including terminal NUL
-        inline void setSSO(bool set) { sso.isSSO = set; }
+        inline void setSSO(bool set) { sso.isHeap = !set; }
         inline void setLen(int len) { if (isSSO()) sso.len = len; else ptr.len = len; }
         inline void setCapacity(int cap) { if (!isSSO()) ptr.cap = cap; }
 	inline void setBuffer(char *buff) { if (!isSSO()) ptr.buff = buff; }
```
earlephilhower, line 309: I know this was already there, but could you fix the tab/space issue here?

d-a-v (approved):
Changes do not reflect the long discussion in this PR.
However approving the proposed changes as they are today.

Jason2866:
Nice one. Saves 544 bytes when used with Tasmota. Without 609844 bytes. With 609300 bytes.

dirkmueller (approved):
looks good to me. Oh please, please pretty please.

earlephilhower (approved):
-edit- latest push has add'l changes that need to be reviewed before re-approval.

While I'm not that big of a fan of the `isSSO`->`isHeap` conversion, what's left here does seem legit, minimal, functional, and did have a measurable impact on other large projects, so approving.  Thx!

cores/esp8266/WString.h
```diff
@@ -287,34 +288,38 @@ class String {
...
+        unsigned int capacity() const { return isSSO() ? (unsigned int)SSOSIZE - 1 : ptr.cap; } // Size of max string not including terminal NUL
+        void setSSO(bool set) { sso.isHeap = !set; }
+        void setLen(int len) { if (isSSO()) sso.len = len; else ptr.len = len; }
+        void setCapacity(int cap) { if (!isSSO()) ptr.cap = cap; }
+        void setBuffer(char *buff) { if (!isSSO()) ptr.buff = buff; }
         // Buffer accessor functions
-        inline const char *buffer() const { return (const char *)(isSSO() ? sso.buff : ptr.buff); }
-        inline char *wbuffer() const { return isSSO() ? const_cast<char *>(sso.buff) : ptr.buff; } // Writable version of buffer
+        const char *buffer() const { return isSSO() ? sso.buff : ptr.buff; }
+        char *wbuffer() const { return const_cast<char *>(buffer()); } // Writable version of buffer
```
d-a-v, line 309: Instead of `const_cast<>` I would do the more readable other way around:
```
const char* buffer() const { return wbuffer(); }
char* wbuffer() const { return isSSO() ? sso.buff : ptr.buff; } // Writable version of buffer
```
(I wonder why the compiler don't complain against the const in `char* wbuffer() const`)

TD-er: > (I wonder why the compiler don't complain against the const in `char* wbuffer() const`)

You don't change the pointer, only where the pointer points to?
Otherwise you would need something like `const char* const wbuffer() const`, which is not open for debate on ugliness IMHO ;)

cores/esp8266/WString.cpp
```diff
@@ -157,18 +147,16 @@ unsigned char String::changeBuffer(unsigned int maxStrLen) {
             uint16_t oldLen = len();
             setSSO(true);
             setLen(oldLen);
-            return 1;
         } else { // if bufptr && !isSSO()
             // Using bufptr, need to shrink into sso.buff
-            char temp[sizeof(sso.buff)];
-            memcpy(temp, buffer(), maxStrLen);
-            free(wbuffer());
+            const char *temp = buffer();
```
earlephilhower, line 152: Good optimization!

cores/esp8266/WString.cpp
```diff
@@ -637,13 +614,14 @@ int String::lastIndexOf(char theChar) const {
 int String::lastIndexOf(char ch, unsigned int fromIndex) const {
     if (fromIndex >= len())
         return -1;
-    char tempchar = buffer()[fromIndex + 1];
-    wbuffer()[fromIndex + 1] = '\0';
-    char* temp = strrchr(wbuffer(), ch);
-    wbuffer()[fromIndex + 1] = tempchar;
+    char *wbuffer = this->wbuffer();
```
earlephilhower, line 617: The local variable `wbuffer` is the same as the class method `wbuffer()` so I'd really ask to change this name to something else.

I think the logic is still correct, it's just a readability and sanity thing.

cores/esp8266/WString.cpp
```diff
@@ -677,10 +655,11 @@ String String::substring(unsigned int left, unsigned int right) const {
         return out;
     if (right > len())
         right = len();
-    char temp = buffer()[right];  // save the replaced character
-    wbuffer()[right] = '\0';
-    out = wbuffer() + left;  // pointer arithmetic
-    wbuffer()[right] = temp;  //restore character
+    char *wbuffer = this->wbuffer();
```
earlephilhower, line 658: Same comment as above.

Also, for these two changes have you seen any code size difference?  wbuffer() is a pretty simple call and I think it has a good chance of being inlined to a simple reference to the string array...

earlephilhower (changes requested):

Jason2866:
@earlephilhower  the last change does indeed save some bytes more.
With this change we have a 72 bytes smaller Tasmota build

d-a-v:
In order for this PR to finally reach merging state, and thanks for this, can you please, as suggested by @earlephilhower rename `char *wbuffer = this->wbuffer();` to avoid name collision ?
I'd suggest `char *writableBuffer`. But it could be `writableb`, or `wb`, `writb`, `iamdifferentfromwbuffer` or even just `wbuf`.
Well... I guess you got it @jjsuwa :)

earlephilhower (approved):
I see the local variable wbuffer was renamed, thanks!  We should put this in and then see if we have any issues at end users, then in a month or so submit the same change to the ESP32 repo.

cores/esp8266/WString.h
```diff
@@ -323,10 +323,16 @@ class String {
         char *wbuffer() const { return isSSO() ? const_cast<char *>(sso.buff) : ptr.buff; } // Writable version of buffer
 
     protected:
-        void init(void) {
-            sso.buff[0] = 0;
-            sso.len     = 0;
-            sso.isHeap  = 0;
+        void init(void) __attribute__((always_inline)) {
+            sso.buff[0]  = 0; // In the Xtensa ISA, in fact, 32-bit store insn ("S32I.N") is one-byte shorter than 8-bit one ("S8I").
```
TD-er, line 327: About the comment; One byte or one bit shorter?

devyte:
@jjsuwa What was the total gain from this optimization effort? Flash size, IRAM, heap, speed etc. Could you please redo your checks from end of August and report here for completeness?

jjsuwa-sys3175:
sorry, now @jjsuwa is unable to login due to my mistake (DNS MX expiration)...

@devyte :
> Could you please redo your checks from end of August and report here for completeness?

end of August... such a long ago :)

`FSBrowser.ino` build results:

- commit faf59f5190d77eb5190580f87477f451c81f31e6, "Update to GCC 10.2 (#7607)" (Sep 24, 2020):
FSBrowser.ino.bin : 338512 bytes
  - IROM : 305252
  - IRAM : 26324
  - DATA : 1272
  - RODATA : 1512
  - BSS : 25552

- commit faf59f5190d77eb5190580f87477f451c81f31e6 + this PR (replacing `core/esp8266/WString.*`):
FSBrowser.ino.bin : 337744 bytes
  - IROM : 304484
  - IRAM : 26324
  - DATA : 1272
  - RODATA : 1512
  - BSS : 25552

for speed... i guess, at least it shouldn't become slow by this PR because of code shrinkage, funccall overhead elimination and so on.

PR #7557, opened 2020-08-28
Fixup #7554: Inline, fewer LOC, remove redundant definition in cpp.

dok-net:


d-a-v:
Title should be
> Fixup #7554: ...

dok-net:
@d-a-v totally right, sorry, switched lines on copy&paste.

cores/esp8266/Esp.h
```diff
@@ -123,12 +123,12 @@ class EspClass {
 
 #if defined(F_CPU) || defined(CORE_MOCK)
         constexpr uint8_t getCpuFreqMHz() const
+#else
+        uint8_t getCpuFreqMHz() const
+#endif
```
devyte, line 128: Minor nitpick: please put the constexpr specifier on its own line here. It took me several seconds to understand that this is alternates signatures.
E.g.:
```
#if defined(F_CPU) || defined(CORE_MOCK)
        constexpr 
        uint8_t getCpuFreqMHz() const
#else
        uint8_t getCpuFreqMHz() const
#endif
```
or maybe this:
```cpp
#if defined(F_CPU) || defined(CORE_MOCK)
        constexpr 
#endif
        uint8_t getCpuFreqMHz() const
```

dok-net: You are right.

dirkmueller:
there are out of band inlines in Esp.h at the end. I don't really care which way, but it would be great to stay consistent..

d-a-v (approved):

PR #7558, opened 2020-08-28
Use direct member initialization instead of ctr initialisation

dirkmueller:
This removes a bit of code repetition.


libraries/ESP8266SSDP/ESP8266SSDP.cpp
```diff
@@ -47,12 +47,10 @@ extern "C" {
 #include "include/UdpContext.h"
 //#define DEBUG_SSDP  Serial
 
-#define SSDP_INTERVAL     1200
 #define SSDP_PORT         1900
 #define SSDP_METHOD_SIZE  10
 #define SSDP_URI_SIZE     2
 #define SSDP_BUFFER_SIZE  64
-#define SSDP_MULTICAST_TTL 2
```
d-a-v, line 55: I would leave `SSDP_INTERVAL` and `SSDP_MULTICAST_TTL` defined and used in member initializers for clarity (maybe also adding the unity: `SSDP_INTERVAL_SECONDS`)

dirkmueller: that'd defeat the purpose of the PR to only have initialization once in one place. I can rename the _interval member variable to _interval_s to keep the unit.

this define is used only in one place, adding this #define does not add make things easier (you need to look it up anyway when you come across reading the constructor). 

if thats a hard objection the only thing I'd do is then to remove ESP82666SSDP* completely from this cleanup, leaving it as it is now.

earlephilhower:
Other than trading one style for another, @dirkmueller, it doesn't look like this has any actual effect on the generated code, no?  I'm not really convinced one way is better than the other (and the whole codebase mixes and matches between the two ways in my experience)...

dirkmueller:
@earlephilhower yes, generated code will remain identical. We could decide for one style (e.g the one proposed in this PR) and I can make the code base consistent. There are advantages (reordering class members doesn't require code changes in the constructor code, multiple constructors don't have to repeate initializations)

TD-er:
Isn't there a difference here?
Don't know what to search for to check in the standard, but I thought it also allowed to initialize the members directly on the stack, before the constructor was called.
If members are initialized in the constructor, then it depends on whether the constructor is actually called (and also which constructor).
This can be an issue if you have specified a destructor which depends on variables to be initialized correctly (like pointers).


But like I said, have to check in the standard to make sure that's what is happening.

libraries/ESP8266SSDP/ESP8266SSDP.cpp
devyte: What @d-a-v means is expose the defines, i.e.: move them to the header, maybe turn them into constexpr, and use the relevant ones in the direct member assignment instead of using literal numbers that don't convey meaning.

devyte: BTW, if it is true that the new MDNS responder also handles SSDP, then this SSDP class will likely be pulled out.

dirkmueller:
> Isn't there a difference here?
> Don't know what to search for to check in the standard, but I thought it also allowed to initialize the members directly on the stack, before the constructor was called.

No, its identical, see this FAQ: https://isocpp.org/wiki/faq/cpp11-language-classes#member-init

you're right that initalizer lists (e.g. foo::foo() : member(0) {} ) allow avoiding default constructor invocation of the member (which optimizes runtime) as well as allow the compiler to be clever in initializing objects. however one or the other is just syntactical sugar that is making error-prone repetition of code obsolete (DRY - don't repeat yourself).

devyte (approved):

d-a-v (approved):
Thanks !

PR #7559, opened 2020-08-28
Avoid float-double-conversion

dirkmueller:
Converting floats to doubles is very expensive on esp8266, so prefer
calculations or comparisons as float. This saves 10% (20 bytes) of the
String::parseFloat() code size and probably quite a bit of runtime
overhead.

### DIFF
--- a/cores/esp8266/Stream.cpp
+++ b/cores/esp8266/Stream.cpp
@@ -173,7 +173,7 @@ float Stream::parseFloat(char skipChar) {
     boolean isFraction = false;
     long value = 0;
     int c;
-    float fraction = 1.0;
+    float fraction = 1.0f;
 
     c = peekNextDigit();
     // ignore non numeric leading characters
@@ -190,7 +190,7 @@ float Stream::parseFloat(char skipChar) {
         else if(c >= '0' && c <= '9') {      // is c a digit?
             value = value * 10 + c - '0';
             if(isFraction)
-                fraction *= 0.1;
+                fraction *= 0.1f;
         }
         read();  // consume the character we got with peek
         c = timedPeek();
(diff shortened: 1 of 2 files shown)


d-a-v:
`float fraction = 1.0;` has been missed

dirkmueller:
@d-a-v thats just initialization, which seems to be handled fine by the compiler. anyway, fixed, thanks!

earlephilhower:
Can you please build an example that uses these calls w/ and w/o this change so we can measure the code differences.

dirkmueller:
so there are examples that use the code, but they're not run in the ci checks. no idea why :-(

Here's what I did to reproduce: 

add a (nonsensical) usage:
```diff
--- a/libraries/esp8266/examples/TestEspApi/TestEspApi.ino
+++ b/libraries/esp8266/examples/TestEspApi/TestEspApi.ino
@@ -242,6 +242,7 @@ void setup() {
   // messages are also readable.
 
   Serial.begin(74880);
+  Serial.parseFloat();
 
   // Try pushing frequency to 160MHz.
   system_update_cpu_freq(SYS_CPU_160MHZ);
```

then build it:


```bash
$ pio ci ./libraries/esp8266/examples/TestEspApi --board nodemcuv2
...
RAM:   [===       ]  33.9% (used 27752 bytes from 81920 bytes)
Flash: [===       ]  25.5% (used 265836 bytes from 1044464 bytes)
.pio/build/nodemcuv2/firmware.elf  :
section                                        size         addr
.data                                          1248   1073643520
.noinit                                           4   1073644768
.text                                           232   1074790400
.irom0.text                                  236132   1075843088
.text1                                        26656   1074790632
.rodata                                        1568   1073644784
.bss                                          24936   1073646352
.comment                                       5065            0
```

then I applied the patch to ~/.platformio/packages/framework-arduinoespressif8266/cores/esp8266/Stream.cpp
and built again:

```bash
$ pio ci ./libraries/esp8266/examples/TestEspApi --board nodemcuv2
...
RAM:   [===       ]  33.9% (used 27752 bytes from 81920 bytes)
Flash: [===       ]  25.4% (used 265804 bytes from 1044464 bytes)
.pio/build/nodemcuv2/firmware.elf  :
section                                        size         addr
.data                                          1248   1073643520
.noinit                                           4   1073644768
.text                                           232   1074790400
.irom0.text                                  236100   1075843088
.text1                                        26656   1074790632
.rodata                                        1568   1073644784
.bss                                          24936   1073646352

```

PR #7573, opened 2020-09-02
Update certs-from-mozilla.py

gnorbsl:
Looks like Mozilla changed the csv, certs are now at row 32


earlephilhower (approved):

earlephilhower:
At some point the script should get smart enough to find the "PEM Info" column and use that instead of an absolute column number, but for now this gets everyone going.  Thx!

gnorbsl:
@earlephilhower I added a pull request that should address the changing rows

Issue #7300, opened 2020-05-12
Problem with certs-from-mozilla.py


gjt211:
#### Platform

- Hardware:			ESP8266
- Core Version:			2.7.1
- Development Env:		[Arduino IDE 1.8.12]
- Operating System:		[Windows 7]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[DIO]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description
I am trying to test the example BearSSL_CertStore1.ini and the python script fails when trying to run it on my Windows7 system.
I need to run the python script to generate the .AR file but it keeps failing.

I have installed the latest python 3.8.2.
The following is the error from the executing `py certs-from-mozilla.py` and note that the `data` directory gets created but there are no files in there.

It says file not found, but doesn't say what file (maybe it is the ca_000.der file but I am not sure). Is there something else that I need to install?
I have tried running cmd.exe as administrator and it makes no difference.
Can anyone suggest how to proceed please?

```
C:\py>py certs-from-mozilla.py
AC Camerfirma, S.A.:AC Camerfirma SA CIF A82743287:http://www.chambersign.org -> data/ca_000.der
Traceback (most recent call last):
  File "certs-from-mozilla.py", line 55, in <module>
    ssl = Popen(['openssl','x509','-inform','PEM','-outform','DER','-out', certName], shell = False, stdin = PIPE)
  File "C:\Users\username\AppData\Local\Programs\Python\Python38-32\lib\subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\username\AppData\Local\Programs\Python\Python38-32\lib\subprocess.py", line 1307, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified

C:\py>
```

JiriBilek:
Do you have openssl.exe on your path? It won't work without it.

earlephilhower:
As @JiriBilek said, you need to have OpenSSL in the PATH variable.

gjt211:
[...]

PR #7578, opened 2020-09-04
Update certs-from-mozilla.py

gnorbsl:
Check if ar exists, if not tell the user what to get to prevent issue #7300 also dynamically get certs instead of hardcoded row item https://github.com/esp8266/Arduino/pull/7573#issuecomment-686192353

### DIFF
--- a/libraries/ESP8266WiFi/examples/BearSSL_CertStore/certs-from-mozilla.py
+++ b/libraries/ESP8266WiFi/examples/BearSSL_CertStore/certs-from-mozilla.py
@@ -11,6 +11,8 @@ from __future__ import print_function
 import csv
 import os
 import sys
+from shutil import which
+
 from subprocess import Popen, PIPE, call
 try:
     from urllib.request import urlopen
@@ -20,7 +22,9 @@ try:
     from StringIO import StringIO
 except Exception:
     from io import StringIO
-
+#check if ar is available
+if which('ar') is None and not os.path.isfile('./ar') and not os.path.isfile('./ar.exe'):
+    raise Exception("You need the program 'ar' from xtensa-lx106-elf: https://docs.espressif.com/projects/esp8266-rtos-sdk/en/latest/get-started/index.html#setup-toolchain")
 # Mozilla's URL for the CSV file with included PEM certs
 mozurl = "https://ccadb-public.secure.force.com/mozilla/IncludedCACertificateReportPEMCSV"
 
@@ -35,7 +39,9 @@ csvFile = StringIO(csvData)
 csvReader = csv.reader(csvFile)
 for row in csvReader:
     names.append(row[0]+":"+row[1]+":"+row[2])
-    pems.append(row[32])
+    for item in row:
+        if item.startswith("'-----BEGIN CERTIFICATE-----"):
+            pems.append(item)
 del names[0] # Remove headers
 del pems[0] # Remove headers
 
--- a/libraries/ESP8266WiFi/examples/BearSSL_CertStore/certs-from-mozilla.py
+++ b/libraries/ESP8266WiFi/examples/BearSSL_CertStore/certs-from-mozilla.py
@@ -24,8 +24,9 @@ except Exception:
     from io import StringIO
 #check if ar is available
 if which('ar') is None and not os.path.isfile('./ar') and not os.path.isfile('./ar.exe'):
-    raise Exception("You need the program 'ar' from xtensa-lx106-elf: found here: (esp8266-arduino-core-root-path)/tools/xtensa-lx106-elf/bin/xtensa-lx106-elf-ar")
-# Mozilla's URL for the CSV file with included PEM certs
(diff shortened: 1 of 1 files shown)


d-a-v:
Thanks !
Our `ar` does not come from espressif but rather from the script `tools/get.py` which reads its sources from [`package/package_esp8266com_index.template.json`](https://github.com/esp8266/Arduino/blob/master/package/package_esp8266com_index.template.json).

gnorbsl:
@d-a-v So assuming someone is a new user that just wants to make SSL requests with this. How does he find ar or get.py?

Is there any documentation that I can link to that shows the user how to get ar?

d-a-v:
> Is there any documentation that I can link to that shows the user how to get ar?

There are three way to install:
- arduino ide: toolchain is automatically downloaded to `tools/xtensa-lx106-elf/`
- git: documentation asks user to run `cd tools; python get.py`. It will download the right toolchain to the same place
- platformio: we don't document it (pio has it own documentation)

> I assume ar is /Java/hardware/tools/avr/bin/avr-ar?

You are running Arduino IDE on macOS. That one is not the good one.
You should have: `Java/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/xtensa-lx106-elf/bin/ar` and `Java/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/bin/xtensa-lx106-elf-ar`.
That is the automatic installation mentioned above.

gnorbsl:
@d-a-v Thank you,

can I just link your comment in the python script so users that stumble over this issue can see what to do?

d-a-v:
Not this comment's URL. That would be sufficient :
```
(esp8266-arduino-core-root-path)/tools/xtensa-lx106-elf/bin/xtensa-lx106-elf-ar
```

d-a-v:
You do have it.
Is it not `/Application/Arduino/whatever/Java/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/bin/xtensa-lx106-elf-ar` ?

d-a-v:
@earlephilhower Do we really need xtensa's ar ? Any version of `ar` seems to be sufficient.

Anyway, we know for sure that we always have this one `(esp8266-arduino-core)/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/xtensa-lx106-elf/bin/ar` because it is shipped with the necessary toolchain.
Other versions may work but are not be always available on every setup (AVR's `ar` not with PIO, linux's `ar` now with bare windows)

gnorbsl:
I don't have Arduino in Application since I never moved it there.  All I have in the Arduino package contents is a folder named arduino and tools, both have an avr folder inside.

"we" excludes myself then I guess. If others have the path then thats fine. I just got ar from somewhere else and ran the script inside the folder with ar and that fixed it.

JiriBilek:
I suggest you to test the existence of openssl as well.

d-a-v:
I think we should update the path with this one because it is always present in a working setup (the current one is `.../xtensa-lx106-elf-ar`, but the script uses `ar`) :
```
(esp8266-arduino-core)/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/xtensa-lx106-elf/bin/ar
```

@JiriBilek that's right

```
if which('ar') is None and not os.path.isfile('./ar') and not os.path.isfile('./ar.exe'):
    raise Exception("You need the program 'ar' from xtensa-lx106-elf found here: (esp8266-arduino-core)/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/xtensa-lx106-elf/bin/ar")
if which('openssl') is None and not os.path.isfile('./openssl') and not os.path.isfile('./openssl.exe'):
    raise Exception("You need to have openssl in PATH, installable from https://www.openssl.org/")
```

gnorbsl:
changed path and added openssl check @d-a-v

libraries/ESP8266WiFi/examples/BearSSL_CertStore/certs-from-mozilla.py
```diff
@@ -20,8 +22,11 @@
     from StringIO import StringIO
 except Exception:
     from io import StringIO
-
-# Mozilla's URL for the CSV file with included PEM certs
+#check if ar is available
+if which('ar') is None and not os.path.isfile('./ar') and not os.path.isfile('./ar.exe'):
+    raise Exception("You need the program 'ar' from xtensa-lx106-elf found here: (esp8266-arduino-core)/hardware/esp8266com/esp8266/tools/xtensa-lx106-elf/xtensa-lx106-elf/bin/ar")
+if which('openssl') is None and not os.path.isfile('./openssl') and not os.path.isfile('./openssl.exe'):
+    raise Exception("You need to have openssl in PATH, installable from https://www.openssl.org/")# Mozilla's URL for the CSV file with included PEM certs
```
d-a-v, line 29: Maybe a line break/feed for the comment ?

gnorbsl: fixed?

d-a-v (approved):

d-a-v:
Thanks for making this script more generic !

PR #7588, opened 2020-09-09
test: use -std=gnu17 instead of -std=c17

mcspr:
amend #7552, same build options everywhere

https://github.com/esp8266/Arduino/blob/a460cb793574dd4b055ef4c409ebc57d6394888f/platform.txt#L58
https://github.com/esp8266/Arduino/blob/a460cb793574dd4b055ef4c409ebc57d6394888f/tools/platformio-build.py#L57


earlephilhower (changes requested):
@mcspr this is a good catch, but I think you found a problem in *platform.txt* and not the host tests.

Since we're using `gnu++17` for C++, it seems only right to use `gnu17` for C, not `c17` in the real app build.

Would you like to swap the change around (i.e. modify platform.txt to match existing host makefile)?

mcspr:
My assumption was possible portability of the code b/c I do remember some of the  `__asm__` vs `asm` comments here:
https://github.com/esp8266/Arduino/pull/6294#issuecomment-635676547
:shrug:

Will revise / re-test

earlephilhower:
@mcspr ping on this.  Could you swap the platform.txt of Makefile to `gnu17` instead of the other way around?

earlephilhower (approved):
Thx!

PR #7590, opened 2020-09-10
WString: make to be safer init state, in order to tolerate (potential) stray dtor calls

jjsuwa:
derived from #7553

### DIFF
--- a/cores/esp8266/WString.h
+++ b/cores/esp8266/WString.h
@@ -288,8 +288,8 @@ class String {
         enum { SSOSIZE = sizeof(struct _ptr) + 4 - 1 }; // Characters to allocate space for SSO, must be 12 or more
         struct _sso {
             char     buff[SSOSIZE];
-            unsigned char len   : 7; // Ensure only one byte is allocated by GCC for the bitfields
-            unsigned char isSSO : 1;
+            unsigned char len    : 7; // Ensure only one byte is allocated by GCC for the bitfields
+            unsigned char isHeap : 1;
         } __attribute__((packed)); // Ensure that GCC doesn't expand the flag byte to a 32-bit word for alignment issues
         enum { CAPACITY_MAX = 65535 }; // If typeof(cap) changed from uint16_t, be sure to update this enum to the max value storable in the type
         union {
@@ -297,10 +297,10 @@ class String {
             struct _sso sso;
         };
         // Accessor functions
-        inline bool isSSO() const { return sso.isSSO; }
+        inline bool isSSO() const { return !sso.isHeap; }
         inline unsigned int len() const { return isSSO() ? sso.len : ptr.len; }
         inline unsigned int capacity() const { return isSSO() ? (unsigned int)SSOSIZE - 1 : ptr.cap; } // Size of max string not including terminal NUL
-        inline void setSSO(bool set) { sso.isSSO = set; }
+        inline void setSSO(bool set) { sso.isHeap = !set; }
         inline void setLen(int len) { if (isSSO()) sso.len = len; else ptr.len = len; }
         inline void setCapacity(int cap) { if (!isSSO()) ptr.cap = cap; }
 	inline void setBuffer(char *buff) { if (!isSSO()) ptr.buff = buff; }


earlephilhower (changes requested):
I don't see how this affects the c'tor/d'tor one way or the other.  It just seems to change the name and inverting the logic for SSO flags.

If that flag itself is not set somewhere before being used in a d'tor, you have to assume it's random.  So, still the 50% chance of trying to `free` something that wasn't allocated.

TD-er:
> potential... (negligible in fact)

Steps to reproduce it...

- Create a struct with several String objects in it.
- Make sure you're running out of memory ;)
- Create such a struct directly by adding it to a std::list using `emplace_back`, preferrably with some constructor values of the struct which will cause a memory allocation fail in one of the first members making sure the latter String members are not properly constructed.
- Try to erase the item from the list. => crash
- Edit your "negligible" statement here ;)

earlephilhower:
@TD-er can you give an MCVE for that, one that fails for you?

Seems straightforward enough (and you can do a `void *unused = malloc(47000);` at the beginning to eat up most RAM and then keep adding strings until there is a failure.  But, again, I think the issue is as likely to be in the `std::list` (which wants to throw an exception on OOM but can't in normal operating mode on the 8266) as in `String`.

jjsuwa:
> Make sure you're running out of memory ;)

post-OOM behavior... virtually impossible to keep heap integrity, should dump so to debug serial and `panic()` if possible.

TD-er:
> > Make sure you're running out of memory ;)
> 
> post-OOM behavior... virtually impossible to keep heap integrity, should dump so to debug serial and `panic()` if possible.

Well I do always check large object allocations whether they return nullptr.
As my sketch is quite often running into almost out of memory situations, it is nice to have those be handled a bit more gracefully than a crash.
But this also means whatever member exists in a class must be able to be destructed regardless whether the class was fully constructed.

TD-er:
> @TD-er can you give an MCVE for that, one that fails for you?
> 
> Seems straightforward enough (and you can do a `void *unused = malloc(47000);` at the beginning to eat up most RAM and then keep adding strings until there is a failure. But, again, I think the issue is as likely to be in the `std::list` (which wants to throw an exception on OOM but can't in normal operating mode on the 8266) as in `String`.

I will try to find some time for it this weekend.

jjsuwa:
> As my sketch is quite often running into almost out of memory situations, it is nice to have those be handled a bit more gracefully than a crash.

post-OOM condition is typical corner-case and thus `malloc()` without nullptr check are so ubiquitous despite of bad manners.
but in your work, is not...

the main interest: is this PR sufficient to resolve your problem?

earlephilhower:
On further inspection, I'm *really* not seeing anything this changes and I'm not seeing how String() could free an invalid pointer and cause a crash like @TD-er is thinking.

My reasoning:

All `String` constructors call `String::init()`:
https://github.com/esp8266/Arduino/blob/a460cb793574dd4b055ef4c409ebc57d6394888f/cores/esp8266/WString.cpp#L131-L135

If the constructor is called with a `nullptr` `this`, that will immediately crash as it is writing values to `NULL`.

So, the only way it doesn't crash is with a valid `this` pointer.  It sets the `isSSO` flag and makes a 0-len string in the structure.

Now on destruction (which cannot ever happen if the c'tor `this` was `nullptr`) we do:
https://github.com/esp8266/Arduino/blob/a460cb793574dd4b055ef4c409ebc57d6394888f/cores/esp8266/WString.cpp#L123-L125
https://github.com/esp8266/Arduino/blob/a460cb793574dd4b055ef4c409ebc57d6394888f/cores/esp8266/WString.cpp#L137-L141

So, we check the `isSSO` flag and potentially free a heap pointer and then do the same as the c'tor (i.e. set some flags which, in this case, are ignored).

So, this PR doesn't change anything or affect the behavior in any way.  And, as far as I can see, the existing behavior is safe and sane.

If memory gets corrupted elsewhere, or the String `this` is corrupted by other code, well yeah it could crash but that's not `String`'s fault or problem and there's nothing we can do about it.

jjsuwa:
thanks for the discussion, this PR is melt back into #7553 and closed.

PR #7603, opened 2020-09-17
Update path of gdb

amfasis:
Used this nice tutorial several times, but each time had to use find to find gdb since it's not in hardware but in tools and also in a version-numbered folder


d-a-v (approved):
Thanks!
`gdb` path depends on installation type, and version.
Your link is good with current release 2.7.4 with an Arduino IDE installation type.

PR #7604, opened 2020-09-17
EspSoftwareSerial 6.9.0: data/strings in flash alignment handling fix

dok-net:
Also: yielding during write() improvement.


dok-net:
@d-a-v speed of light, thanks :-)

PR #7605, opened 2020-09-17
WiFiSever - Arduino API available() and accept()

jandrassy:
According to documentation, the `server.available()` should return first client with data available. 
https://www.arduino.cc/en/Reference/WiFiServerAvailable (esp8266 doc points there). 

ESP8266WiFi library now does what a new EthernetServer.accept() method does: https://www.arduino.cc/en/Reference/EthernetServerAccept. It was added to Ethernet library by Paul Stoffregen in 2018 for Ethernet 2.0. I added it to UIPEthernet in February and I have it in my WiFiEspAT library too.

This PR shows a draft of a possible implementation of `available()`.  (This quick implementation could let some client wait to long.)

My opinion is that it is strange to have a function like this available(). My explanation how they defined it like this, is that the networking libraries by Arduino are always remote APIs for a firmware and there it doesn't look so strange.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiServer.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiServer.cpp
@@ -115,7 +115,7 @@ bool WiFiServer::hasClient() {
     return false;
 }
 
-WiFiClient WiFiServer::available(byte* status) {
+WiFiClient WiFiServer::accept(byte* status) {
     (void) status;
     if (_unclaimed) {
         WiFiClient result(_unclaimed);
@@ -133,6 +133,24 @@ WiFiClient WiFiServer::available(byte* status) {
 
     optimistic_yield(1000);
     return WiFiClient();
+}
+
+WiFiClient WiFiServer::available() {
+  for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
+    WiFiClient& client = connectedClients[i];
+    if (client && client.status() == CLOSED) {
+      client = WiFiClient();
+    }
+    if (!client) {
+      client = accept();
+    }
+  }
+  for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
+    WiFiClient& client = connectedClients[i];
+    if (client.available())
+      return client;
+  }
+  return WiFiClient();
 }
 
 uint8_t WiFiServer::status()  {
--- a/libraries/ESP8266WiFi/src/WiFiServer.h
+++ b/libraries/ESP8266WiFi/src/WiFiServer.h
@@ -65,6 +65,10 @@ extern "C" {
 class ClientContext;
 class WiFiClient;
 
+#ifndef MAX_MONITORED_CLIENTS
+#define MAX_MONITORED_CLIENTS 5
+#endif
+
 class WiFiServer : public Server {
   // Secure server needs access to all the private entries here
 protected:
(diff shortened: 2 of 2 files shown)


d-a-v:
It seems to be a fair request, in the name of compatibility with Arduino API.

However, it would break many sketches and libraries because of the significant `WiFiServer::available()` semantic change.

There will be no change for simple TCP server answering to a simple request then closing.
There will be an undefined behavior / a total mess when several clients can be managed in parallel: such code would believe to have several independent connections, while they would be in fact all the same one. I would call this a collapsing change.

We can't really fully deprecate `WiFiServer::available()`, can we ? 

What could be done:

- in all case, `::accept()` must be `::available()` old implementation 
- add an optional user `#define` like `ESP8266_ARDUINO_WIFISERVER_OLD_API` to keep up with current = old API
- without this define, make `WiFiServer::available()` unavailable
- another `#define` is needed, like `ESP8266_ARDUINO_WIFISERVER_LEGAL_API` would enable official Arduino implementation for enabling `::available()` like the one proposed in this PR.

***pros:***
- current sketches/libraries would not compile (= *will not silently break*) after these changes, unless by using `ESP8266_ARDUINO_WIFISERVER_OLD_API` which is harmless with older versions of this core.
- Arduino compatibility is possible but requires another define.

***cons:***
- Arduino compatibility is not totally transparent and requires a define: using `ESP8266_ARDUINO_WIFISERVER_LEGAL_API` in a sketch / a library also requires from the user to check that this define is supported
```
#ifndef ESP8266_ARDUINO_WIFISERVER_LEGAL_API_SUPPORTED
#error must use core 3.0.0+
#else 
#define ESP8266_ARDUINO_WIFISERVER_LEGAL_API
#endif
```

This proposal raises several potential issues depending on how these defines are really defined by the user/sketch/library.
Before moving forward, we must all reach an agreement about the above.
Any other proposal is welcome. The condition is that an existing sketch ***must not silently break***.

jandrassy:
In the years esp8266 Arduino exists, missed someone the true Arduino behaviour of available()? I think not. 
If it would be my call, it would add accept() as alias to available() and document available() as it is now.

jandrassy:
now I realized that the Arduino implementation of `available()` is tied with the write-to-all-clients functionality of `Server` base class as `Print` implementation. this is not implemented in ESP8266WiFi library.

`WiFiServer telnetLogging;` where every `telnetLogging.print` goes to all connected clients (or is ignored if there is no connected client) could be passed to any function or class which would take `Print& log`.

maybe a WiFiArduinoServer could inherit from WiFiServer and add the Arduino style available() function and the write-to-all-clients functionality. (same for SSL server)

d-a-v:
> maybe a WiFiArduinoServer could inherit from WiFiServer and add the Arduino style available() function and the write-to-all-clients functionality. (same for SSL server)

That would be a very good first step.

jandrassy:
> 
> 
> > maybe a WiFiArduinoServer could inherit from WiFiServer and add the Arduino style available() function and the write-to-all-clients functionality. (same for SSL server)
> 
> That would be a very good first step.

I rework this PR. Should I use WiFiClient or ClientContext to store the clients? (ClientContext with ref and unref)

d-a-v:
I think you have noticed that `WiFiClient` is just a wrapper around `ClientContext`.
If you don't intend to indeed use but only store the client, just use the simple version.

jandrassy:
this will be the example called PagerServer,  a simple server that echoes any incoming messages to all connected telnet clients.
```
void loop() {
  WiFiClient client = server.available(); // returns first client which has data to read
  if (client) { // client is true only if it is connected and has data to read
    String s = client.readStringUntil('\n'); // read the message incoming from one of the clients
    Serial.println(s); // print the message to Serial Monitor
    client.print("echo: "); // this is only for sending client
    server.println(s); // send the message to all connected clients
    server.flush(); // flush the buffers
  }
}
```
I think it is Arduino way elegant.

https://github.com/jandrassy/WiFiEspAT/blob/master/examples/Basic/PagerServer/PagerServer.ino

jandrassy:
next attempt https://github.com/esp8266/Arduino/pull/7612

PR #7608, opened 2020-09-22
FIX MIME-Type for txt-suffix

The-MEO:
In addition to #7312 the MIME type for txt files is to be corrected.

Currently the MimeType for txt files is the file extension ".txt". 
In the commit the variable with text/plain is used again.


d-a-v (approved):

earlephilhower:
Good catch, that was obviously a typo in the code!

PR #7609, opened 2020-09-22
FIX StaticRequestHandler crashes when cache_header is null

The-MEO:
In the StaticRequestHandler method, exception 28 (LoadProhibitedCause) occurs when debugging with the DEBUG_ESP_CORE option and if the default parameter null is used for cache_header.

In this commit, the pointer is checked for null before output.


d-a-v (approved):
Thanks!

PR #7611, opened 2020-09-25
WString: c_str() returns null pointer after move

mcspr:
target = std::move(source) does not reset source's buffer pointer back to the sso
#7553 fixes this behaviour

Added as a separate test, perhaps it could go into the existing SSO

*edit:* failing, as intended:
https://github.com/esp8266/Arduino/pull/7611/checks?check_run_id=1166504473#step:4:2364
```
-------------------------------------------------------------------------------
core/test_string.cpp:22
...............................................................................

core/test_string.cpp:30: FAILED:
  REQUIRE( origin.c_str() != nullptr )
with expansion:
  {null string} != nullptr
```

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -45,7 +45,6 @@ String::String(const __FlashStringHelper *pstr) {
     *this = pstr; // see operator =
 }
 
-#ifdef __GXX_EXPERIMENTAL_CXX0X__
 String::String(String &&rval) noexcept {
     init();
     move(rval);
@@ -55,7 +54,6 @@ String::String(StringSumHelper &&rval) noexcept {
     init();
     move(rval);
 }
-#endif
 
 String::String(char c) {
     init();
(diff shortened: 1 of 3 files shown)


earlephilhower (changes requested):
Can you add a patch that fixes this to the PR?  I guess it was only discovered when the noexcept was added, allowing GCC to actually use the move method?

everslick:
Is this also affecting the WString implementation on ESP32? I think it got thee SSO patches at one point.

mcspr:
> Can you add a patch that fixes this to the PR?

Done as https://github.com/esp8266/Arduino/pull/7611/commits/7ce820aeba4b7405324cab5c7d0ae48dec94b0ee
Although, I also *could* rebase with `--author=...` rewrite of the git history and co-authred-by: $myself instead, since I repeat the existing patch almost 1-to-1

> I guess it was only discovered when the noexcept was added, allowing GCC to actually use the move method?

Exactly. Which I noticed when running tests for the library assigning the existing object throughout the loop and moving it elsewhere periodically

earlephilhower (approved):
Good catch, thanks!

PR #7612, opened 2020-09-26
ArduinoWiFiServer with send-to-all-clients functionality

jandrassy:
and with available() working according to Arduino documentation. plus PagerServer example.

should the ESP8266WiFi library have this?


earlephilhower:
I think this is pretty cool, @jandrassy!  I would be careful about how `available()` on SSL connections works with this.  IIRC, when I did the SSL server/client `available` and `connected` were the 2 big stumbling blocks.

jandrassy:
> I think this is pretty cool, @jandrassy! I would be careful about how `available()` on SSL connections works with this. IIRC, when I did the SSL server/client `available` and `connected` were the 2 big stumbling blocks.

Thank you @earlephilhower . I am not sure if it is useful, but at least it shows the difference between Servers in Arduino libraries and in ESP libraries.

I attempted to test the SSL server, but telnet-ssl could not connected. I will test it again if this PR should be merged.

jandrassy:
@earlephilhower, multiple copies of  BearSSL::WiFiClientSecure for one connection are not in sync. so ArduinoWiFiServer doesn't work because the copy it holds still returns available() based on _recvapp_len, even if the copy returned before already read the data from the connection. if the data is read again esp8266 crashes.

axTLS works (the copies share the context object)

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
```diff
@@ -0,0 +1,127 @@
...
+
+  using Print::write;
+
+  virtual void flush() override {
+    flush(0);
+  }
+
+  virtual void flush(unsigned int maxWaitMs) {
+    for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
+      if (connectedClients[i]) {
```
d-a-v, line 93: `::connected()` is enough() (this test is `::connected() or ::available()`)

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  virtual size_t write(uint8_t b) override {
+    return write(&b, 1);
+  }
+
+  virtual size_t write(const uint8_t *buf, size_t size) override {
+    size_t ret = 0;
+    for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
+      if (connectedClients[i]) {
```
d-a-v, line 78: `::connected()` is enough() (this test is `::connected() or ::available()`)

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
```diff
@@ -0,0 +1,127 @@
...
+#define arduinowifiserver_h
+
+#include <ESP8266WiFi.h>
+
+#ifndef MAX_MONITORED_CLIENTS
+#define MAX_MONITORED_CLIENTS 5
+#endif
+
+template <class TServer, class TClient>
+class ArduinoComptibleWiFiServerTemplate : public Server {
```
d-a-v, line 31: Would it be wise to derive from `TServer` instead of `Server` ?

d-a-v:
> multiple copies of BearSSL::WiFiClientSecure for one connection are not in sync.

From what I understand, there's only one copy for each client, given by `::accept()`. Am I wrong ?

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
jandrassy: it will be much better. it removes the redirecting methods and getServer

jandrassy:
> > multiple copies of BearSSL::WiFiClientSecure for one connection are not in sync.
> 
> From what I understand, there's only one copy for each client, given by `::accept()`. Am I wrong ?

the concept of Arduino Client is that multiple copies of WiFiClient/EthernetClient (created as return values or parameters passed by copy) can interact with the same TCP connection. WiFiClient should be only a pointer to internal representation. In ESP8266WiFi library the WiFiClient does it with ClientContext. the axTLS does it very very nice with the std shared pointer to SSLContext.

(I think this strange concept of Arduino API classes has origin in simulating Java objects behaviour for some sketch source compatibility with Processing framework. so no `->` or `&` only `.`)

In the ArduinoWiFiServer template the array of 'monitored' WiFiClient objects has instance pointing to the connection and server.available() returns its copy always when again data are available for it. So the WiFiClient object in the array 
must have the same information about count of data available as the copy used to read the data.

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
```diff
@@ -0,0 +1,117 @@
...
+    return TClient(); // no client with data found
+  }
+
+  virtual size_t write(uint8_t b) override {
+    return write(&b, 1);
+  }
+
+  virtual size_t write(const uint8_t *buf, size_t size) override {
+    size_t ret = 0;
+    for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
```
d-a-v, line 71: Should `availableForWrite()` be tested for all writable clients and be compared to `size` and use the min value instead of `size` ?

d-a-v:
> the concept of Arduino Client is that multiple copies of WiFiClient/EthernetClient (created as return values or parameters passed by copy) can interact with the same TCP connection.

Right, I think we agree, I was meaning: "From what I understand, there's only one ***TCP instance*** for each client, given by ::accept().".
Do we really agree ? If we do I don't see the reason why several SSL clients should not work.

However maybe you should call on every occasion and even if it's useless `::connected()` *and* `::available()` on *every* client occurence to make the TLS state machine move forward. All of them, but of course only the `index` one is returned to the user.

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
jandrassy: yes. to avoid blocking. sorry I am not used to reliable availableForWrite. last time a checked it was missing  in Print here. I usually write code for many platforms. I am little confused with this nice limited space :-). but we can't return how many bytes were written to individual clients. so it is shot and forget

jandrassy:
> > the concept of Arduino Client is that multiple copies of WiFiClient/EthernetClient (created as return values or parameters passed by copy) can interact with the same TCP connection.
> 
> Right, I think we agree, I was meaning: "From what I understand, there's only one _**TCP instance**_ for each client, given by ::accept().".
> Do we really agree ? If we do I don't see the reason why several SSL clients should not work.
> 
> However maybe you should call on every occasion and even if it's useless `::connected()` _and_ `::available()` on _every_ client occurence to make the TLS state machine move forward. All of them, but of course only the `index` one is returned to the user.

for BearSSL the problem are the two copies for the same connection. one in the array and second in sketch filled at return from server.available(). the sketch reads data over its copy. the buffer pointer _recvapp_len gets to 0. but in the instance in the array still has the original value of _recvapp_len and returns the client again as if it has data available. 

available() calls bool() for every monitored client and bool() calls connected.

d-a-v:
> for BearSSL the problem are the two copies of WiFiClientSecure for the same connection

That is true. I wasn't aware. `WiFiClientSecure`'s context must move away to an external context for your class to work with it in the arduino way.
Like with `WiFiClient` which is stateless and pointing to a `ClientContext` which holds the TCP connection.

jandrassy:
> > for BearSSL the problem are the two copies of WiFiClientSecure for the same connection
> 
> That is true. I wasn't aware. `WiFiClientSecure`'s context must move away to an external context for your class to work with it in the arduino way.
> Like with `WiFiClient` which is stateless and pointing to a `ClientContext` which holds the TCP connection.

d-a-v:
Can you try with #7680 ?

jandrassy:
I didn't have much time. First attempt shows that
`connectedClients[i].status() == ESTABLISHED` is never true

and second client gets
```
juraj@nuc ~ $ telnet-ssl -z ssl 192.168.1.117 2323
Trying 192.168.1.117...
SSL: reason(0)
SSL_Connect: Input/output error
```

jandrassy:
@d-a-v 

`connectedClients[i].status() == ESTABLISHED` is never true

and second client gets
```
juraj@nuc ~ $ telnet-ssl -z ssl 192.168.1.117 2323
Trying 192.168.1.117...
SSL: reason(0)
SSL_Connect: Input/output error
```
for the second client "BSSL:_connectSSLServerRSA: OOM error" is logged
WiFiServerSecureAxTLS could handle 4 clients

d-a-v:
Thanks to @earlephilhower 's suggestion in gitter `server.setBufferSizes()` (512 was not enough, but 1024 is OK),
and his ssl server example from which I copied keys and `server.setECCert()`,
and also thanks to a fix on `ssl::status()` not properly routed to ssl context in #7680 that I'm about to push,
I could run your PR with 3 SSL clients and it's working well (showing two below):

```
$ telnet-ssl -z ssl 10.0.1.144 2323
Trying 10.0.1.144...
SSL: Server has a self-signed certificate
SSL: unknown Issuer: /C=AU/ST=Some-State/O=Internet Widgits Pty Ltd
Connected to 10.0.1.144.
Escape character is '^]'.
toto
echo: toto
titi
echo: titi
```

```
$ telnet-ssl -z ssl 10.0.1.144 2323
Trying 10.0.1.144...
SSL: Server has a self-signed certificate
SSL: unknown Issuer: /C=AU/ST=Some-State/O=Internet Widgits Pty Ltd
Connected to 10.0.1.144.
Escape character is '^]'.
toto
titi

```
patched example:
```
  // Attach the server private cert/key combo
  BearSSL::X509List *serverCertList = new BearSSL::X509List(server_cert);
  BearSSL::PrivateKey *serverPrivKey = new BearSSL::PrivateKey(server_private_key);
  server.setECCert(serverCertList, BR_KEYTYPE_KEYX|BR_KEYTYPE_SIGN, serverPrivKey);
  server.setBufferSizes(1024,1024);
  server.begin();
```

libraries/ESP8266WiFi/src/ArduinoWiFiServer.h
d-a-v: > sorry I am not used to reliable availableForWrite

I guess it is reliable in this core :)

> last time a checked it was missing in Print here

[It is there now](https://github.com/esp8266/Arduino/pull/7658), in master, as you know.

> EDIT: I think the user would prefer a little blocking over incomplete data

That's indeed the purpose of using it

> but we can't return how many bytes were written to individual clients. so it is shot and forget

We can get max availableForWrite over all opened client, then write this max size and return it.
The writer would have to deal with the returned value.

jandrassy: ok, I code it that way

jandrassy: done, but I had the PR7680 merged in the ArduinoWiFiServer branch, now I am not sure of the result. I think I can restore the branch to clean state without the PR7680 if necessary

d-a-v (approved):
Approving but to merge after #7680 (needed for this PR when SSL is used)
(or after #7680 is removed from this PR)

jandrassy:
> Approving but to merge after #7680 (needed for this PR when SSL is used)
> (or after #7680 is removed from this PR)

#7680 is removed

earlephilhower (approved):

PR #7627, opened 2020-10-03
Prevent rewriting Updater_Signing.h if content unchanged

mongozmaki:
Updater_Signing.h is rewritten at each compilation forcing the recompilation of some ESP libs.
This fix compares the content if the file already exists and updates it only if changed.

Very similar to "core_version.h" fix in #6414


earlephilhower (approved):
LGTM, thanks.

PR #7628, opened 2020-10-03
Add missing sntp_init/sntp_stop

mongozmaki:
Since #7109, `sntp_stop` and `sntp_init` are not called anymore when using
`configTime(int timezone_sec, int daylightOffset_sec, const char* server1, const char* server2, const char* server3)`

This results in NTP sync not working. (`settimeofday_cb` callback is never called).


d-a-v (approved):
Thanks !

Issue #7423, opened 2020-07-02
Change EEPROM.end() return type from void to bool


mrv96:
According to docs, EEPROM.end() in https://github.com/esp8266/Arduino/blob/master/libraries/EEPROM/EEPROM.cpp is a commit() followed by a free of used RAM.

Since commit() return a bool, also end() should return a bool to check if the commit failed or not.

d-a-v:
Will you propose a pull-request for that suggestion ?

PR #7630, opened 2020-10-03
BREAKING:  Change return EEPROM.end() to bool

Erriez:
As requested in https://github.com/esp8266/Arduino/issues/7423.


earlephilhower (approved):
LGTM.  This will diverge slightly from the base Arduino EEPROM class, now, but be forward compatible.

PR #7633, opened 2020-10-04
Update OTA HTTP Server Header Information

sidwarkd:
The headers sent when an OTA update is requested of an HTTP server have changed in the code. This change is to update the documentation accordingly. PHP sample code was changed but not tested.


earlephilhower (approved):
LGTM.  I did not run, but did double-check the strings in ESP8266httpUpdate.cpp and these seem to match.  Good catch!

PR #7636, opened 2020-10-06
MDNS MultiInterface

hreintke:


devyte (approved):
My one comment here is: I see very few comments in the changed code. If at all possible, please add. comments explaining what the code lines do.
Approved.

PR #7637, opened 2020-10-06
settimeofday_cb: distinguish from user or sntp

### LINKED ISSUE #7616: Setting a callback for SNTP time sync event [feature request]
### Basic Infos

#### Platform

- Hardware:				[ESP-12]
- Core Version:			[2.7.4]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Nodemcu]
- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]

### Problem Description

Function `settimeofday_cb()` sets a callback executed every time when `settimeofday()` is invoked. SNTP synchronization event change system time using `settimeofday()` - which is ok, but if I run `settimeofday()` manually also callback function is invoked. There is no mechanism to show what was the source of callback - network synchronization or manual set.

### Expected Behavior

I can see at least three possible solutions:
1. Add a parameter to callback function declaration - maybe some pointer to `bool` type variable which will be set to true if SNTP was a source of sync.
2. Add another `settimeofday`-like function which do the same, but do not fire any callback.
3. Expose ESP SDK function `sntp_set_time_sync_notification_cb()` from `lwip/apps/sntp/sntp.c` which sets a callback ONLY for SNTP synchronization.

Personally I prefer the 3rd one (only because I tested it in "plain" not "Arduino-based" ESP projects and works like a charm).

d-a-v:
This PR overrides `settimeofday_cb()` with a new callback signature:
- currently `void myCB ()` is and remains accepted
- `void myBetterCB (bool from_sntp)` is added: `from_sntp` is true when time is set by SNTP, otherwise it is false.

Consequently:
- lwip2 is updated with a trivial change
- [NTP_TZ_DST](https://github.com/esp8266/Arduino/blob/a0648ee62f60a24e67bc79176c0a8a24307ae440/libraries/esp8266/examples/NTP-TZ-DST/NTP-TZ-DST.ino) example is also updated to show this new callback

Fixes #7616 @JakubRakus


JakubRakus:
Tested, works like a charm.

earlephilhower (approved):
LGTM

PR #7639, opened 2020-10-07
ESP8266WiFilibrary -ESP8266 core 2.7.2

Kunal0cr7:
To hook up the ESP module to Wi-Fi (like hooking up a mobile phone to a hot spot), you need only a couple of lines of code:

#include <ESP8266WiFi.h>

void setup()
{
  Serial.begin(115200);
  Serial.println();

  WiFi.begin("network-name", "pass-to-network");

  Serial.print("Connecting");
  {
    delay(500);
    Serial.print(".");
  }
  Serial.println();

  Serial.print("Connected, IP address: ");
}

void loop() {}
In the line WiFi.begin("network-name", "pass-to-network") replace network-name and pass-to-network with the name and password of the Wi-Fi network you would like to connect to. Then, upload this sketch to ESP module and open the serial monitor. You should see something like:

Connection log on Arduino IDE's Serial Monitor
How does it work? In the first line of the sketch, #include <ESP8266WiFi.h> we are including the ESP8266WiFi library. This library provides ESP8266 specific Wi-Fi routines that we are calling to connect to the network.

The actual connection to Wi-Fi is initialized by calling:

WiFi.begin("network-name", "pass-to-network");
The connection process can take couple of seconds and we are checking for whether this has completed in the following loop:

{
  delay(500);
  Serial.print(".");
}
The while() loop will keep looping as long as WiFi.status() is other than WL_CONNECTED. The loop will exit only if the status changes to WL_CONNECTED.

The last line will then print out the IP address assigned to the ESP module by DHCP:

If you don’t see the last line but just more and more dots ........., then likely name or password to the Wi-Fi network is entered incorrectly in the sketch. Verify the name and password by connecting from scratch to this Wi-Fi network with a PC or a mobile phone.

Note: if connection is established, and then lost for some reason, ESP will automatically reconnect to the last used access point once it is again back on-line. This will be done automatically by Wi-Fi library, without any user intervention.

That’s all you need to connect ESP8266 to Wi-Fi. In the following chapters we will explain what cool things can be done by the ESP once it’s connected.

Who is Who
Devices that connect to Wi-Fi networks are called stations (STA). Connection to Wi-Fi is provided by an access point (AP), that acts as a hub for one or more stations. The access point on the other end is connected to a wired network. An access point is usually integrated with a router to provide access from a Wi-Fi network to the internet. Each access point is recognized by a SSID (Service Set IDentifier), that essentially is the name of network you select when connecting a device (station) to the Wi-Fi.

ESP8266 modules can operate as a station, so we can connect it to the Wi-Fi network. It can also operate as a soft access point (soft-AP), to establish its own Wi-Fi network. When the ESP8266 module is operating as a soft access point, we can connect other stations to the ESP module. ESP8266 is also able to operate as both a station and a soft access point mode. This provides the possibility of building e.g. mesh networks.

ESP8266 operating in the Station + Soft Access Point mode
The ESP8266WiFi library provides a wide collection of C++ methods (functions) and properties to configure and operate an ESP8266 module in station and / or soft access point mode. They are described in the following chapters.

Class Description
The ESP8266WiFi library is broken up into several classes. In most of cases, when writing the code, the user is not concerned with this classification. We are using it to break up description of this library into more manageable pieces.

Index of classes of ESP8266WiFi library
Chapters below describe all function calls (methods and properties in C++ terms) listed in particular classes of ESP8266WiFi. The description is illustrated with application examples and code snippets to show how to use functions in practice. This information is broken up into the following documents.

Station
Station (STA) mode is used to get the ESP module connected to a Wi-Fi network established by an access point.

ESP8266 operating in the Station mode
Station class has several features to facilitate the management of a Wi-Fi connection. In case the connection is lost, the ESP8266 will automatically reconnect to the last used access point, once it is available again. The same happens on module reboot. This is possible since ESP saves the credentials to the last used access point in flash (non-volatile) memory. Using the saved data ESP will also reconnect if sketch has been changed but code does not alter the Wi-Fi mode or credentials.

Station Class documentation

Check out separate section with examples.

Soft Access Point
An access point (AP) is a device that provides access to a Wi-Fi network to other devices (stations) and connects them to a wired network. The ESP8266 can provide similar functionality, except it does not have interface to a wired network. Such mode of operation is called soft access point (soft-AP). The maximum number of stations that can simultaneously be connected to the soft-AP can be set from 0 to 8, but defaults to 4.

ESP8266 operating in the Soft Access Point mode
The soft-AP mode is often used and an intermediate step before connecting ESP to a Wi-Fi in a station mode. This is when SSID and password to such network is not known upfront. ESP first boots in soft-AP mode, so we can connect to it using a laptop or a mobile phone. Then we are able to provide credentials to the target network. Then, the ESP is switched to the station mode and can connect to the target Wi-Fi.

Another handy application of soft-AP mode is to set up mesh networks. The ESP can operate in both soft-AP and Station mode so it can act as a node of a mesh network.

Soft Access Point Class documentation

Check out the separate section with examples.

Scan
To connect a mobile phone to a hot spot, you typically open Wi-Fi settings app, list available networks and pick the hot spot you need. Then enter a password (or not) and you are in. You can do the same with the ESP. Functionality of scanning for, and listing of available networks in range is implemented by the Scan Class.

Scan Class documentation

Check out the separate section with examples.

Client
The Client class creates clients that can access services provided by servers in order to send, receive and process data.

ESP8266 operating as the Client
Check out the separate section with list of functions

BearSSL Client Secure and Server Secure
BearSSL::WiFiClientSecure and BearSSL::WiFiServerSecure are extensions of the standard Client and Server classes where connection and data exchange with servers and clients using secure protocol. It supports TLS 1.2 using a wide variety of modern ciphers, hashes, and key types.

ESP8266 operating as the Client Secure
Secure clients and servers require significant amounts of additional memory and processing to enable their cryptographic algorithms. In general, only a single secure client or server connection at a time can be processed given the little RAM present on the ESP8266, but there are methods of reducing this RAM requirement detailed in the relevant sections.

BearSSL::WiFiClientSecure contains more information on using and configuring TLS connections.

BearSSL::WiFiServerSecure discusses the TLS server mode available. Please read and understand the BearSSL::WiFiClientSecure first as the server uses most of the same concepts.

Check out the separate section with examples .

Server
The Server Class creates servers that provide functionality to other programs or devices, called clients.

ESP8266 operating as the Server
Clients connect to sever to send and receive data and access provided functionality.

Check out separate section with examples / list of functions.

UDP
The UDP Class enables the User Datagram Protocol (UDP) messages to be sent and received. The UDP uses a simple “fire and forget” transmission model with no guarantee of delivery, ordering, or duplicate protection. UDP provides checksums for data integrity, and port numbers for addressing different functions at the source and destination of the datagram.

Check out separate section with examples / list of functions.

Generic
There are several functions offered by ESP8266’s SDK and not present in Arduino WiFi library. If such function does not fit into one of classes discussed above, it will likely be in Generic Class. Among them is handler to manage Wi-Fi events like connection, disconnection or obtaining an IP, Wi-Fi mode changes, functions to manage module sleep mode, hostname to an IP address resolution, etc.

Check out separate section with examples / list of functions.

Diagnostics
There are several techniques available to diagnose and troubleshoot issues with getting connected to Wi-Fi and keeping connection alive.

Check Return Codes
Almost each function described in chapters above returns some diagnostic information.

Such diagnostic may be provided as a simple boolean type true or false to indicate operation result. You may check this result as described in examples, for instance:

Serial.printf("Wi-Fi mode set to WIFI_STA %s\n", WiFi.mode(WIFI_STA) ? "" : "Failed!");
Some functions provide more than just a binary status information. A good example is WiFi.status().

Serial.printf("Connection status: %d\n", WiFi.status());
This function returns following codes to describe what is going on with Wi-Fi connection:

0 : WL_IDLE_STATUS when Wi-Fi is in process of changing between statuses

1 : WL_NO_SSID_AVAILin case configured SSID cannot be reached

3 : WL_CONNECTED after successful connection is established

4 : WL_CONNECT_FAILED if password is incorrect

6 : WL_DISCONNECTED if module is not configured in station mode

It is a good practice to display and check information returned by functions. Application development and troubleshooting will be easier with that.

Use printDiag
There is a specific function available to print out key Wi-Fi diagnostic information:

WiFi.printDiag(Serial);
A sample output of this function looks as follows:

Mode: STA+AP
PHY mode: N
Channel: 11
AP id: 0
Status: 5
Auto connect: 1
SSID (10): sensor-net
Passphrase (12): 123!$#0&*esP
BSSID set: 0
Use this function to provide snapshot of Wi-Fi status in these parts of application code, that you suspect may be failing.

Enable Wi-Fi Diagnostic
By default the diagnostic output from Wi-Fi libraries is disabled when you call Serial.begin. To enable debug output again, call Serial.setDebugOutput(true). To redirect debug output to Serial1 instead, call Serial1.setDebugOutput(true). For additional details regarding diagnostics using serial ports please refer to the documentation.

Below is an example of output for sample sketch discussed in Quick Start above with Serial.setDebugOutput(true):

Connectingscandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 1
cnt

connected with sensor-net, channel 6
dhcp client start...
chg_B1:-40
...ip:192.168.1.10,mask:255.255.255.0,gw:192.168.1.9
.
Connected, IP address: 192.168.1.10
The same sketch without Serial.setDebugOutput(true) will print out only the following:

Connecting....
Connected, IP address: 192.168.1.10
Enable Debugging in IDE
Arduino IDE provides convenient method to enable debugging for specific libraries.

What’s Inside?
If you like to analyze in detail what is inside of the ESP8266WiFi library, go directly to the ESP8266WiFi folder of esp8266 / Arduino repository on the GitHub.

To make the analysis easier, rather than looking into individual header or source files, use one of free tools to automatically generate documentation. The class index in chapter Class Description above has been prepared in no time using great Doxygen, that is the de facto standard tool for generating documentation from annotated C++ sources.

Example of documentation prepared by Doxygen
The tool crawls through all header and source files collecting information from formatted comment blocks. If developer of particular class annotated the code, you will see it like in examples below.

Example of documentation for station begin method by Doxygen
Example of documentation for station hostname property by Doxygen
If code is not annotated, you will still see the function prototype including types of arguments, and can use provided links to jump straight to the source code to check it out on your own. Doxygen provides really excellent navigation between members of library.

Example of documentation for UDP begin method (not annotated in code)by Doxygen
Several classes of ESP8266WiFi are not annotated. When preparing this document, Doxygen has been tremendous help to quickly navigate through almost 30 files that make this library.


ElectricBeat:
@earlephilhower 

Hello Sir 

What is the advantage of arduino esp8266 core "2.7.4" vs arduino esp8266 core "2.4.2"

JAndrassy:
> @earlephilhower
> 
> Hello Sir
> 
> What is the advantage of arduino esp8266 core "2.7.4" vs arduino esp8266 core "2.4.2"

https://github.com/esp8266/Arduino/releases?page=2

PR #7652, opened 2020-10-14
BREAKING: Add Wrong Password wifi status case

### LINKED ISSUE #4152: View specific wrong password error upon connecting to network
Suggestion for future versions: make a new status() output to specifically reflect wrong password.

https://github.com/esp8266/Arduino/blob/ca250687339659f336157509202f57c2612c8cf6/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L493

Of course, it is possible to call wifi_station_get_connect_status() from user_interface.h, but it would be cleaner if this was possible with ESP8622WiFi.h.

devyte:
Fixes #4152


devyte:
This PR is considered breaking, because with the prior code a sketch could just be testing for FAILED, while now a test for either FAILED or WRONG_PASSWORD should be done.

d-a-v (approved):

earlephilhower (approved):

PR #7658, opened 2020-10-17
BREAKING: Add Print::availableForWrite method

### LINKED ISSUE #7650: Sd File::availableForWrite()
Hi, I have recently discovered `int availableForWrite()` to check the amount of available bytes without waiting for new "sector" (hence avoiding blocking writings). [here](https://github.com/arduino-libraries/SD/blob/85dbcca432d1b658d0d848f5f7ba0a9e811afc04/examples/NonBlockingWrite/NonBlockingWrite.ino)) the example that demonstrate it.

 Is there any plan to add to esp8266 core?

earlephilhower:
Adds an availableForWrite() method to the Print class, matching current
ArduinoCore-API commit 398e70f188e2b861c10d9ffe5e2bfcb6a4a4f489 .

Hook availableForWrite into the SDFS filesystem (other FSes don't have
this capability built-in).

Fixes #7650


cores/esp8266/FSImpl.h
```diff
@@ -35,6 +35,7 @@ class FileImpl {
     virtual bool seek(uint32_t pos, SeekMode mode) = 0;
     virtual size_t position() const = 0;
     virtual size_t size() const = 0;
+    virtual int availableForWrite() { return 0; }
```
d-a-v, line 38: There seem to be no afw in LittleFS.
If there is not, or if AFW is unsupported, what do you think of returning -1 or File::UNSUPPORTED or whatever < 0 ?

earlephilhower: The way I understand it, `availableForWrite()` means "how many bytes can be written w/o a potential blocking call."  And `0` is what the Arduino core is returning in the "I don't know case" (which is safe and agrees with the definition).

So, since this is for Arduino core compatibility, I'd say stick with 0.  If someone cares, they can dig into LittleFS's guts and figure out how to peek the temp sector usage and return it. :)

d-a-v: Let's stick with Arduino API then. 0 = may block or (can't write <= unsupported)

d-a-v (approved):

PR #7662, opened 2020-10-18
New WiFi password manager library

Erriez:
New WiFi password manager library to encode/decode WiFi SSID/password with XTEA algorithm.

@devyte @earlephilhower,
I'm looking for someone to discuss/review technical details of this full functional proof of concept.

---

**The problem**

By default, the ESP8266 saves WiFi passwords in plain text to on-chip flash and can be easily retrieved. This is a known design flaw in the ESP8266 which cannot be resolved by software. A hacker can simply dump the credentials by overwriting the sketch when it has physical access to the ESP8266 device, or remotely using OTA. The retrieved SSID and password can then be misused to login to this network.

To improve security, WiFi credentials must never be saved in plain text to on-chip flash and requires at least `WiFi.persistent(false);` at the beginning of each sketch. The disadvantage is that fast-connect at startup is disabled which takes more time to connect.

**Proof of concept**

This library makes it more difficult for hackers to retrieve the WiFi SSID and password by encoding the credentials with a fast and simple XTEA algorithm. Decoding SSID and password on a ESP8266 at 80MHZ takes around 0.5ms.

The application designer is responsible to check if WiFi credentials are saved into on-chip flash with `isWiFiCredentialSaved()` and securely erase it with `eraseWiFiPasswords()`. This function uses SDK functions `system_restore()` followed by `system_reset()` to generate a CPU reset.

**Usage**

The library consists of 3 sketches:

Sketch | Function
---------  | ------------
01-WiFiPassManSetup.ino | Generate secret key / encode WiFi SSID/password
02-WiFiPassManConnect.ino | Decode WiFi SSID/password and connect to AP
03-WiFiPassManErase.ino | Dump plain text password and erase it

The application designer generates a unique 128-bit secret key and encodes the WiFi SSID (optional) and password (required) by calling `encodeWiFiCredentials()`. Maximum length SSID and password is 63 characters + NULL. By default the output is printed to the `Serial` and the flash tables must be included in each sketch for decoding.

Writing a new sketch to the ESP8266 overwrites the secret key and encoded WiFi SSID/password. By design, the `LittleFS` is not used, otherwise another sketch can be used to read and print the secret key.

Decoding must be performed once before calling `WiFi.begin(SSID, PASS)`.

Encoding/decoding uses the hardware AP MAC address by default to bind to only one specific ESP8266 device. In this case the encoded WiFi SSID/password flash table cannot be decoded on another ESP8266 with a different MAC address and requires a re-encoding.

**DISCLAIMER**

The XTEA algorithm contains known and documented attacks and therefore the wording encoding is used instead of real encryption.

Macro XTEA_NUM_ROUNDS can be increased to 128 or higher to increase complexity. This also increases the decoding time.

Note: A hacker with physical access to the device can dump the entire flash to a binary file and analyze the location of the secret key with encoded WiFi SSID/password table. For this architecture, it is technically not possible to protect the flash for readout. It requires a lot of research for the hacker to decode the WiFi credentials, hardware/software and crypto knowledge.

100% security does not exist. It is up to crypto experts to determine the safeness of the used algorithm and library.
Encoding WiFi SSID and password is always better than saving it in plain text which can be read by everyone.

### DIFF
--- a/libraries/ESP8266WiFiPassMan/examples/01-WiFiPassManSetup/01-WiFiPassManSetup.ino
+++ b/libraries/ESP8266WiFiPassMan/examples/01-WiFiPassManSetup/01-WiFiPassManSetup.ino
@@ -1,0 +1,46 @@
+/*
+    This sketch generates encoded WiFi SSID/password.
+*/
+
+#include <ESP8266WiFiPassMan.h>
+
+// WiFi credentials to be encoded
+#define WIFI_SSID   "my-ssid"
+#define WIFI_PASS   "my-secret-password-123456789"
+
+// Use predefined secret key for encoding. Keep 0 to generate new secret key.
+const uint32_t secretKey[4] PROGMEM = {
+  0x00000000, 0x00000000, 0x00000000, 0x00000000
+};
+
+// true:  WiFi credentials can be decoded on this ESP8266 only (recommended).
+// false: WiFi credentials can be decoded on any ESP8266 (less secure).
+bool encodeAPMAC = true;
+
+
+void setup() {
+  // Initialize Serial
+  delay(500);
+  Serial.begin(115200);
+  Serial.println(F("\nESP8266 WiFi password manager setup\n"));
+
+  // Create WiFi password manager object.
+  ESP8266WiFiPassMan wpm(encodeAPMAC);
+
+  // Optional: Run selftest encoding/decoding string up to 63 characters
+  if (!wpm.selftest("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!")) {
+    return;
+  }
+  if (!wpm.selftest(WIFI_SSID)) {
+    return;
+  }
+  if (!wpm.selftest(WIFI_PASS)) {
+    return;
+  }
+
+  // Encode and print WiFi credentials
+  wpm.encodeWiFiCredentials(WIFI_SSID, WIFI_PASS, secretKey);
+}
+
+void loop() {
+}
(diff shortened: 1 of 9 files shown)


JiriBilek:
Nice idea, I really dislike plaintext passwords in MCU's flash.

Have you considered to use a BearSSL cipher algorithm instead of XTEA? My point of view is that when I am concerned about plaintext passwords, I also communicate using TLS and so the standard ciphers from BearSSL are already linked in my application. So we could use e.g. AES cipher instead.

Erriez:
@JiriBilek Thanks for your reply. Yes, I started with ChaCha20 as listed in `Crypto.h` to encode/decode the password and got it up and running, but decided to switch to XTEA. The reason is that XTEA requires less flash space and is much faster.

In the mean time I figured out that `esptool.py` contains a `flash_read` command. It simply means that an attacker can dump the entire flash to a binary file and open it in a disassembler. In this case, all encryption is useless when the arguments of `WiFi.begin()` can be retrieved. This library prevents reading WiFi SSID/password for 'normal' users.

Rule of thumb: Yes, SSID/password encoding is possible on ESP8266, but must use `WiFi.persistent(false)`. Simply retrieving the SSID and password by loading a new sketch is no longer possible with this library. However, with sufficient technical knowledge and time, it is possible to retrieve the password from flash. The ESP8266 and ESP32 chips are cheap, but also weak in security.

earlephilhower:
@Erriez, interesting PR!  However, I think the best spot for this would be a separate library and not part of the core, similar to the WiFi connection managers and other utilities.  It also allows you to update freely w/o us getting in the way or requiring a new release of the core before people see your changes.  We can add links in the README and you can submit to the Arduino library manager to make it visible to users who care about obfuscating SSID passwords.

That said, I think you have placed a titanium lock on a cardboard box. A lot of the same issues came up while doing the SSL port with client certs (where you need a private key stored on the 8266), for example.  :(

As an aside, maybe a generalization of this into an all-purpose obfuscator would make sense?  So a user could use it for their SSID, their MQTT cert private keys, private web API keys, etc.

Erriez:
@earlephilhower, I understand what you mean and I fully agree with you. A separate library is much easier to maintain for me and it can be used to obfuscate other data as well. I was not fully aware of the weak spots in the ESP8266 when I started this as a proof of concept. Now I learned a lot more about the chip.

I'll delete the source branch and move it to my own Erriez libraries later. We can close this MR.

PR #7666, opened 2020-10-21
Fix eboot crash with exception 28. ets_printf() should take pointer argument.

dirkenstein:
The latest update to eboot was trying to pass 32bit string literals as their own pointer to ets_printf(). This caused eboot to crash with exception 28 on any operation that used ets_printf().


bootloaders/eboot/eboot.c
```diff
@@ -35,7 +35,7 @@ int print_version(const uint32_t flash_addr)
     uint32_t fmt[2];
     fmt[0] = S('v', '%', '0', '8');
     fmt[1] = S('x', '\n', 0, 0);
-    ets_printf((const char*) fmt, ver);
+    ets_printf((const char*) &fmt, ver);
```
earlephilhower, line 38: fmt is already a pointer, so this change should be undone.

bootloaders/eboot/eboot.c
```diff
@@ -278,7 +278,7 @@ int main()
         uint32_t e[2];
         e[0] = S('e', ':', '0' + res, '\n' );
         e[1] = 0;
-        ets_printf((const char*)e);
+        ets_printf((const char*)&e);
```
earlephilhower, line 281: Same as before, fmt already a pointer so please undo.

earlephilhower (changes requested):
I guess I flubbed the original PR big time, so thanks for noticing it.  It didn't crash in my local testing, but that was obviously not sufficient!

bootloaders/eboot/eboot.c
dirkenstein: Yup, missed that the arrays were already pointers. Oops. It's only cp that is affected.

dirkenstein:
The crash only happens when ACTION_COPY_RAW is called as part of an OTA firmware update.

bootloaders/eboot/eboot.c
```diff
@@ -22,7 +22,7 @@ extern void ets_wdt_enable(void);
 extern void ets_wdt_disable(void);
 
 // Converts bit of a string into a uint32
-#define S(a,b,c,d) ( (((uint32_t)a) & 0xff) | (((uint32_t)b) << 8) | (((uint32_t)c) << 16) | (((uint32_t)d)<<24) )
+#define S(a,b,c,d) ( (((int32_t)a) & 0xff) | (((uint32_t)b) << 8) | (((uint32_t)c) << 16) | (((uint32_t)d)<<24) )
```
d-a-v, line 25: Both versions seem fine, so I wonder why this change is needed ?

earlephilhower, line 25: Sorry, didn't notice this change before.  As @d-a-v said, can you please undo the single uint32_t->int conversion here?

earlephilhower (changes requested):

earlephilhower (approved):

d-a-v (approved):

PR #7675, opened 2020-10-24
Add Stream loaders for BearSSL

### LINKED ISSUE #7671: No support for reading client certificate and private key from LittleFS (ESP8266WiFi - WiFiClientSecure/BearSSL)
### Basic Infos

#### Platform

- Hardware:				All
- Core Version:			24-Oct-2020
- Operating System:		All

### Problem Description

Currently, when using WiFiClientSecure (BearSSL), certificate stores can be loaded from LittleFS or SD.
But there is no documented way or code to load a client certificate and private key in a similar manner.
(The X509List and PrivateKey do not take files/streams as arguments)

Old issues and examples show that older versions used to have this feature:
 - https://hackaday.io/project/12482-garage-door-opener/log/45617-connecting-the-esp8266-with-tls
 - https://github.com/esp8266/Arduino/issues/3544

Specifically it seems like the old functions `loadCertificate()` and `loadPrivateKey()` (which are deprecated) could load files.

My current solution is to copy the certificate and key to a global variable, which wastes several KB's RAM:
```cpp

#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <LittleFS.h>
#include <WiFiClientSecure.h>

#define MAX_PEM_SIZE 4096

char clientKeyStr[MAX_PEM_SIZE];
char clientCertStr[MAX_PEM_SIZE];

void setup() {
  LittleFS.begin();
  Serial.begin(115200);

  //... initialize wifi and time...

  File cert = LittleFS.open("/client-crt.pem", "r"); //can be .der file as well
  File key = LittleFS.open("/client-key.pem", "r");

  //... verify files are opened correctly and not exceed MAX_PEM_SIZE ...

  cert.readBytes(clientCertStr, cert.size()); //copy certificate from file to char array
  key.readBytes(clientKeyStr, key.size()); //same for private key
  X509List clientCert(clientCertStr); 
  PrivateKey clientKey(clientKeyStr);

 //...connect to server...
}
```
This sketch works well but wastes 4096*2 = 8192 bytes of RAM which is 10% of total RAM.
My assumption is, when using `CertStoreBearSSL.h`, the certificates are not copied to the RAM for most of the time, but loaded in a different way. 
Same thing goes when loading a certificate which is saved in PROGMEM (sketch ROM). 
Therefore, it should be possible to use a client certificates and a private key which are stored as .pem or .der in the file system, without copying the whole file content to the RAM, for the whole lifetime of the program.
I tried understanding the code in `CertStoreBearSSL.cpp` but it's too complicated for me.
Thanks in advance and Best regards!

earlephilhower:
Fixes #7671

Allows for code to do things like read certs from LittleFS or even HTTP
connections with code like:

  File cert = LittleFS.open("/client-crt.pem", "r");
  clientCert = new X509List(cert, cert.size());
  cert.close();


d-a-v (approved):
Approving even if #7671 seems to be solved

PR #7676, opened 2020-10-24
Fix long password validation in WebServer

### LINKED ISSUE #7664: Max basic authentication password length restricted by base64 encoded linefeed
While finding an answer to a question regarding the max allowed webserver password length in Tasmota (https://github.com/arendst/Tasmota/issues/9587) I stumbled across a hard limit of 47 characters.

The HTTP Basic Authorization userid password as used in ESP8266WebServer-impl.h uses function ``base64_encode_chars`` to encode the ``userid:password``. This function inserts linefeeds/carriage returns every 72 encoded characters as defined in ``cencode.h`` (``#define BASE64_CHARS_PER_LINE 72``). 

In the case of Tasmota, where the userid is fixed set to ``admin``, this allows for 47 characters in the password. Any more characters will insert a linefeed which isn't available in the HTTP request header.

As an example if a user uses a password of ``12345678901234567890123456789012345678901234567890`` it results in an base64 encoded string like
```
YWRtaW46MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIzNDU2Nzg5MDEyMzQ1Njc4
OTA=
```
so with a linefeed after ``..Njc4``. The HTTP request header contains:
```
Authorization: Basic YWRtaW46MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIzNDU2Nzg5MDEyMzQ1Njc4OTA=
```
The compare using function ``authReq.equalsConstantTime(encoded)`` will fail caused by the inserted linefeed.

As a possible solution I suggest to replace function ``authReq.equalsConstantTime(encoded)`` by a function discarding any control characters in the base64 encoded string before comparing.

EDIT: It works fine on the ESP32 as there the linefeed is not inserted in the base64 encoded data.

earlephilhower:
Use a base64 encode that doesn't add CRs to the output when comparing
username:password values for authentication.

Fixes #7664

@arendst, can you please give this patch a try?  I've tested a 100 character password with the HTTPBasicAuth example and it works for me.

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
@@ -26,6 +26,7 @@
 #include "WiFiClient.h"
 #include "ESP8266WebServer.h"
 #include "FS.h"
+#include "base64.h"
 #include "detail/RequestHandlersImpl.h"
 
 static const char AUTHORIZATION_HEADER[] PROGMEM = "Authorization";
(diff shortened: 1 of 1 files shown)


arendst:
Testing...

arendst:
Verified.  Works as expected. Thx.

d-a-v (approved):
LGTM

PR #7680, opened 2020-10-27
WiFiClientSecure: use context

d-a-v:
This PR:
- allows to pass instances of WiFiClientSecure without duplicating the secure connection
- helps #7612

How it is done:
- rename `WiFiClientSecure` class to `WiFiClientSecureCtx`
- create a new `WiFiClientSecure` with a single member as shared ptr to a `WiFiClientSecureCtx*`
- bind all virtual calls from the second to the first

Both `WiFiClientSecure` and `WiFiClientSecureCtx` derive from `WiFiClient`, but `WiFiClientSecure`'s one is not used. Instead, all its virtual functions call its context's ones.

Tested with `BearSSL_MaxFragmentLength.ino`


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -205,8 +198,112 @@ class WiFiClientSecure : public WiFiClient {
...
+    // Attach a preconfigured certificate store
+    void setCertStore(CertStore *certStore) { _ctx->setCertStore(certStore); }
+
+    // Select specific ciphers (i.e. optimize for speed over security)
+    // These may be in PROGMEM or RAM, either will run properly
+    bool setCiphers(const uint16_t *cipherAry, int cipherCount) { return _ctx->setCiphers(cipherAry, cipherCount); }
+    bool setCiphers(const std::vector<uint16_t> list) { return _ctx->setCiphers(list); }
+    bool setCiphersLessSecure() { return _ctx->setCiphersLessSecure(); } // Only use the limited set of RSA ciphers without EC
+
+    // Check for Maximum Fragment Length support for given len before connection (possibly insecure)
```
earlephilhower, line 285: FWIW, these don't need to be moved to ctx, they're static and not part of any connection.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -31,13 +31,12 @@
...
+class WiFiClientSecureCtx : public WiFiClient {
   public:
-    WiFiClientSecure();
-    WiFiClientSecure(const WiFiClientSecure &rhs);
-    ~WiFiClientSecure() override;
+    WiFiClientSecureCtx();
+    ~WiFiClientSecureCtx() override;
 
-    WiFiClientSecure& operator=(const WiFiClientSecure&) = default; // The shared-ptrs handle themselves automatically
+    WiFiClientSecureCtx& operator=(const WiFiClientSecureCtx&) = delete;
```
earlephilhower, line 39: What's with the `default`->`delete` change?  Intentional?

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -1329,17 +1324,17 @@ int WiFiClientSecure::getLastSSLError(char *dest, size_t len) {
   return err;
 }
 
-bool WiFiClientSecure::probeMaxFragmentLength(const char* name, uint16_t port, uint16_t len) {
+bool WiFiClientSecureCtx::probeMaxFragmentLength(const char* name, uint16_t port, uint16_t len) {
```
earlephilhower, line 1327: As before, these can be part of the non-ctx object

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -100,20 +100,15 @@ void WiFiClientSecure::_clearAuthenticationSettings() {
...
   _clear();
   _clearAuthenticationSettings();
   _certStore = nullptr; // Don't want to remove cert store on a clear, should be long lived
   _sk = nullptr;
   stack_thunk_add_ref();
 }
 
-WiFiClientSecure::WiFiClientSecure(const WiFiClientSecure &rhs) : WiFiClient(rhs) {
-  *this = rhs;
-  stack_thunk_add_ref();
```
earlephilhower, line 113: Did we lose an increment to the stack thunk refcnt here?  Or was the refcnt increment in error in the original code?

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -31,13 +31,12 @@
 
 namespace BearSSL {
 
-class WiFiClientSecure : public WiFiClient {
+class WiFiClientSecureCtx : public WiFiClient {
```
earlephilhower, line 34: I think the _client context is opaque, should the secure client context be opaque too?

earlephilhower (commented):
Wow, you did a lot of work!

Generally seems fine, but the extra jump for every call just grates me the wrong way...I can't think of any way around it.  Performance-wise, it should not be an issue since the SSL guts are very slow and this only adds 5-6 instructions per high-level call.  @devyte may have some ideas here.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
d-a-v: Yes, and I just added `WiFiClientSecure::WiFiClientSecure(const WiFiClientSecure &rhs) = delete;`
It is intentional to forbid copy constructor and operator for the context class.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
d-a-v: No. The copy constructor [is now disabled](https://github.com/esp8266/Arduino/pull/7680#discussion_r513378742).
That is in fact the purpose of this PR.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
d-a-v: Should I make `_cipher_list` static in `WiFiClientSecure` ?

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
d-a-v: ClientContext is indeed opaque and WiFiClient is a wrapper to the instance.
WiFiClientSecure is (now) also a wrapper to WiFiClientSecureCtx.
Making the latter opaque will need to move method bodies to the `.cpp`.
This is doable, functions are virtual but I still wonder if the compiler is able to do some optimization when code is in the declaration (?).

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
earlephilhower: Got it, thx!

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
earlephilhower: Actually, this was intended to be a comment about the MFLN calls, not setCipher.  They are not connection oriented, they're static members that just run before a context would be created.

So, leave cipher in and move the probeMFLNs back to the WiFiClientSecure is what I meant.

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
earlephilhower: Yes, good point about the compiler.  The same moving from .cpp->.h saved IROM with the WString optimization, so I'd expect the same here (w/o the code in the .h, g++ needs to generate a real function call and can't optimize out even trivial things since it doesn't know about it.).

I'm fine w/it being non-opaque as it may actually end up after compiling to not be a jump-to-a-jump-to-code and just a jump-to-code.

earlephilhower:
New changes look good.  I need to run some tests on HW locally before approving, as this is straightforward but pretty big impact...

earlephilhower (approved):
LGTM

JerryLoveCat:
where is WiFiClientSecure.verify. it only exists in 2.7.4

Issue #7684, opened 2020-10-29
CertStoreBearSSL.cpp is only file that pervents compile with -D FS_NO_GLOBALS


isadora-6th:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:				NodeMCU-V1.0 (Esp8266)
- Core Version:			espressif8266@2.6.2 , framework-arduinoespressif8266@3.20704.0 (newest platformio can download by version number)
- Development Env:		Platformio
- Operating System:		Windows

### Settings in IDE

- Module:  			Nodemcu
- Flash Mode:			dio
- Flash Size: 		        4MB
- lwip Variant:                  v2 Lower Memory
- Reset Method:	        nodemcu
- Flash Frequency:	        40Mhz
- CPU Frequency:		160MHz
- Upload Using:		SERIAL
- Upload Speed:              other

### Problem Description

Older project version used SDFat as lib needed for work.  
I updated ESP8266 core and package, and currentrly using ESP8266SdFat
Older lib had no defined namespace for it, so i need to prevent different FS in global scope.

To prevent issue with different FS difinitons i used this flag
`-D FS_NO_GLOBALS`  
So all code compiled ok except 1 file`CertStoreBearSSL.cpp`
located at path  
`C:\.platformio\packages\framework-arduinoespressif8266\libraries\ESP8266WiFi\src\CertStoreBearSSL.cpp`
pathcing file by adding `fs::` before all FS stuff usage solved issue, all compiled fine.

```ini

[env:nodemcuv2]
platform = espressif8266@2.6.2
platform_packages = 
    platformio/framework-arduinoespressif8266@3.20704.0
board = nodemcuv2
framework = arduino
build_flags = 
    -D FS_NO_GLOBALS
lib_deps =
    ESP8266WiFi
    SPI
    ESP8266SdFat

```

Sketch 
```cpp
#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <SdFat.h>
/* Compiles only with -D FS_NO_GLOBALS */
[...]

PR #7685, opened 2020-10-30
compile failure fix with FS_NO_GLOBALS flag

isadora-6th:
As was discussed at https://github.com/esp8266/Arduino/issues/7684

Added fs:: before FS stuff used in file.


earlephilhower (approved):
Thx!

PR #7691, opened 2020-11-04
Update to ESP8266HTTPClient.cpp for no Content-Length

### LINKED ISSUE #7692: http.getString() return blank
Hi
I have a simple code that read a json and display result on lcd.
After some month I resumed it but I think that some update (Arduino CLI o other) give me a problem with http.getString() that return empty string.
To do a simple test I did an empty project with only wifi connection http request.
Here my code:
```
#include <ESP8266WiFi.h>

const char* ssid = "FUNCOOL";
const char* password = "1807200505092011";


WiFiClient client;
HTTPClient http;

void setup() {

  Serial.begin(115200);
  WiFi.begin(ssid, password);

    delay(1000);
    Serial.println("Connecting...");
  }

}

void loop() {
  if (WiFi.status() == WL_CONNECTED) {

    http.begin(client, "http://192.168.1.252/user/counter.json");
    int httpCode = http.GET();
    Serial.println("Connected");
    Serial.println(httpCode);
    //Check the returning code
    if (httpCode == 200) {
      Serial.println("Connected");
      Serial.println("http string");
      Serial.println(payload);
      Serial.println("http string end");

    } else {
      Serial.println("Gateway Problem");
    }
    http.end();   //Close connection
  } else {
    Serial.println("NOT Connected! Check WiFi");
  }
  // Delay
  delay(1000);

}

```

when I trace it get 
```
Connecting...
Connecting...
Connected
200
Connected
http string

http string end
Connected
200
Connected
http string

http string end
Connected
200
Connected
http string

```

Has you can see the payload is empty/blank.
Any suggestion for me?
Thanks

drderiv:
Response bodies are ignored when _transferEncoding == HTTPC_TE_IDENTITY and there is no Content-Length header.  The added code here attempts to fix that issue by writing the response body based upon how much content is available to be read.

*edit*:
Fixes #7692


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -728,6 +728,13 @@ int HTTPClient::writeToStream(Stream * stream)
         if(len > 0) {
             ret = writeToStreamDataBlock(stream, len);
 
+            // have we an error?
```
earlephilhower, line 731: Instead of copy-pasting the error clause, can we do something like
````
if (len > 0) {
    ret = writeToStreamDataBlock(stream, len);
} else if ((len == -1 && _client->available() > 0) {
    ret = writeToStreamDataBlock(stream, _client->available() > 0);
}

// Have we an error?
if (ret < 0) {
     return returnError(ret);
}
````

**Review by earlephilhower — CHANGES_REQUESTED** [review]
Little minor refactor, please.

File: libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp (hunk 1 above)
Comment (drderiv) [reply libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp]: Sure thing.  And made a correction at the same time.  Thanks.

**Review by earlephilhower — APPROVED** [review]
Thanks for the cleanup.

File: libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp [hunk 2]
```diff
@@ -727,11 +727,13 @@ int HTTPClient::writeToStream(Stream * stream)
     if(_transferEncoding == HTTPC_TE_IDENTITY) {
         if(len > 0) {
             ret = writeToStreamDataBlock(stream, len);
+        } else if(len == -1 && _client->available() > 0) {
+            ret = writeToStreamDataBlock(stream, _client->available());
+        }
```
d-a-v, line 732: Could you please test this ?
```
        if (len > 0 || len == -1) {
            ret = writeToStreamDataBlock(stream, len);
        } // no else
```
`writeToStreamDataBlock()` is supposed to correctly handle `-1`.
`available()>0` does not garantee that every incoming data have arrived.
When `-1` is given it's either `HTTP/1.0` or `Connection: Close`. In both case, every data must be swallowed until connection is closed by peer.
I'm not sure `writeToStreamDataBlock()` handles this without a timeout though.

This will fix #7692

drderiv: I see your point about possibly not consuming all data in the response body.  The response body coming from the device I am utilizing (a Denon AV receiver) isn't large, so it's always working (and did not work prior to the change I suggested).  But you're right, it should work in all circumstances.

I tested your suggested change, and it alone is not enough to solve the original issue.  The problem appears to have something to do with the code in `writeToStreamDataBlock()`, which sets `readBytes` to `len`, and then passes that to `_client->readBytes()`, which presumably isn't expecting a negative length.

```
    // read all data from server
    while(connected() && (len > 0 || len == -1))
    {
        int readBytes = len;

        // not read more the buffer can handle
        if(readBytes > buff_size) {
            readBytes = buff_size;
        }

        // read data
        int bytesRead = _client->readBytes(buff, readBytes);

```
Thus, my original suggestion solved the problem, at least for response bodies that fit and are fully in buffer by the time the call to `writeToStreamDataBlock()` occurs in `writeToStream()`.  To solve it for all circumstances, perhaps a combination of your suggestion in `writeToStream()` and a change in the above `while` loop that utilizes `_client->available()` like my original suggestion, but does it repeatedly until there's no more to read, is appropriate?

d-a-v: I too have been confused with `bytesRead` and `readBytes`.
To summarize the function:

- `int size` goes to `int len`
- `int readBytes` takes `len`

loop:
- `int bytesRead` takes `read(readBytes)` (possible timeout when input stream is over => end of transmission)
- `write(bytesRead)`
- `len -= bytesRead`

The `if (len<0)` was introduced in #6476 between releases 2.7.1 and 2.7.2.

> it should work in all circumstances.

:+1:

> I tested your suggested change, and it alone is not enough to solve the original issue. 

Can you be more specific, does it work after a timeout or doesn't work at all ?


I am also interested to know if your issue is solved by #6979 (also included in [alpha release v0.0.2](https://d-a-v.github.io/esp8266/Arduino)) which is a complete rewrite of block transfer functions, that are currently duplicated all over the place and not necessarily efficient.

d-a-v: > perhaps a combination of your suggestion in writeToStream() and a change in the above while loop that utilizes `_client->available()` like my original suggestion, but does it repeatedly until there's no more to read, is appropriate?

When `_client` is a tcp connection (`WiFiClient`), `::available()` can be 0 at the time of testing and >0 a bit later.
So instead, the connection status needs to be checked (ie: has peer closed?) specifically when `len` is infinite (-1). That could be added in the current code.

drderiv: Thanks for the explanations.  After re-reading everything very carefully, and in an effort to answer your inquiries and upon further testing, I need to apologize and retract what I said above about your original suggestion alone not being enough to solve problem.  It does so, perfectly.

The bottom line is that, as I think you suspected, `writeToStreamDataBlock()` handles this with a timeout.  That makes sense.  If you don't have a Content-Length, you have to decide at some point when you're going to consider what you've got thus far to be _the_ response.  Turns out I hadn't proactively managed the timeout setting, and the effective timeout was so high that by the time control was returned to my sketch, it was too late to use the response body for anything constructive, causing me downstream issues that I confused with your solution not working.

I don't have a sense of how widespread missing Content-Length responses are in general, but it might be nice to have other ways to deal with it.  For example, an option to pass a (pointer to a) string of characters that if received, cause a return prior to the timeout.  In my case, the response body is fully formed XML, so it's easy to know when it's done.  Having to rely exclusively on a timeout forces one to play the game of picking a time that's long enough to most likely gather the full response, but short enough that you aren't waiting around for no reason.  But because the response time is likely to be somewhat variable, that seems like a less than ideal solution, and something more certain might be nice.

Thanks very much for your efforts.

d-a-v:
Now that it's working but with an unwilling timeout, you can update `writetostreamdatablock()` for it to stop when it can no longer read from the `_client`.
This is an old fashion function. A that time there were usually several attempts to read or write before giving up. Today one attempt and timeout checking is enough. This is addressed by #6979.
In the meantime, what is missing there is checking peer connectivity. That is answering to that question:
Is it possible to get more incoming data ? One just need to check `_client`'s `::available()==0` **and** (`::connected()==false` or `::status() != WL_CONNECTED`). When this is true, it is no longer useful to wait for the timeout, peer has closed the connection and everything that is to be read is already swallowed.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
d-a-v: As mentioned below in the thread, there is a more generic way for stopping the stream of an unknown length without having to wait for the timeout to occur.

drderiv:
With apologies, I'm not seeing how this would help, at least in the current implementations of `HTTPClient` and `Stream`.  My reading is that, in a no Content-Length scenario (so `len = -1`), once `readBytes()` is called, the timeout controls, regardless of anything else.  If that's correct, checking peer or wifi connectivity within `writeToStreamDataBlock` would not be useful.  Changes to address this would need to occur within Stream, and then `writeToStreamDataBlock` would need to be modified to check the parameters you mention.  In addition, I would still think a timeout would be needed (and it would have to be added to `writeToStreamDataBlock` as well) in the case where you remain connected but the server just decides to take a nap.

Perhaps the word 'meantime' threw me and you are asking for the `HTTPClient` mods prospectively to work with the changes to Stream implemented in [#6979](https://github.com/esp8266/Arduino/pull/6979)?

d-a-v:
In #6979, `writetostreamdatablock` is removed and replaced by `Stream::to()` - having both timeout and end of stream detection - but still lacks from maintainers's review because it was introducing `Stream::` inheritance to `String::`, now removed, that was considered by some as a blaspheme, which I still don't understand why ;) Maybe I should repost it with another name and its history expunged of the heresy.

But before #6979 is considered, a fix is needed for your issue, and #7692 too which is the same issue.

End of TCP connection cannot currently be handled by `Stream::` so it has to be `writetostreamdatablock()`'s job. While examining #7692 I ended up with the following and stopped there because I was too optimistic regarding #6979 consideration. You may try it. The logic is the following: do not ask too much to `Stream::readBytes()` because that one will wait until timeout that all requested data are received. Instead ask for what's already there and keep retrying until all data are received.
```
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -725,7 +725,7 @@ int HTTPClient::writeToStream(Stream * stream)
     int ret = 0;
 
     if(_transferEncoding == HTTPC_TE_IDENTITY) {
-        if(len > 0) {
+        if(len > 0 || len == -1) {
             ret = writeToStreamDataBlock(stream, len);
 
             // have we an error?
@@ -1184,6 +1184,9 @@ int HTTPClient::writeToStreamDataBlock(Stream * stream, int size)
         if(readBytes > buff_size) {
             readBytes = buff_size;
         }
+        int av = _client->available();
+        if (readBytes < 0 || readBytes > av)
+            readBytes = av;
 
         // read data
```
If this works for you, please update your PR.

drderiv:
That worked nicely.  And now I understand what you meant about effectively checking `_client`'s `::available()==0` and `::connected()==false`.  The idea about reading what's already there was what I was trying to do -- poorly, as I didn't include the 'until all data are received' part -- in my initial PR.  With your help, I think we've now got it right.  Thanks again.

d-a-v (approved):

PR #7699, opened 2020-11-10
Help newlib TZ parser

### LINKED ISSUE #7690: Some POSIX time zones are not handled correctly
### Basic Infos

#### Platform

- Hardware:				[ESP-12]
- Core Version:			[2.7.4]

### Settings in IDE

- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

It looks as if some POSIX time zone strings may be being interpreted incorrectly when configured using:

```cpp
configTime(const char* tz, const char* server1, const char* server2, const char* server3)
```
or

```cpp
setTZ(const char* tz)
```

The following examples have been taken from [TZ.h](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/TZ.h):

The following are some of the values which demonstrate the issue:

```cpp
#define TZ_America_Sao_Paulo	PSTR("<-03>3") // Offset by +03:00 instead of -03:00
#define TZ_Africa_Casablanca	PSTR("<+01>-1") // Offset by -01:00 instead of +01:00
#define TZ_America_Asuncion	PSTR("<-04>4<-03>,M10.1.0/0,M3.4.0/0") // Offset by +4:00 instead of -03:00
```

This pattern appears consistent for all time zones which are defined in this format.

By contrast the following (also taken from [TZ.h](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/TZ.h)) **do not** exhibit this issue:

```cpp
#define TZ_America_Los_Angeles	PSTR("PST8PDT,M3.2.0,M11.1.0") // Offset by -08:00
#define TZ_Europe_Rome	PSTR("CET-1CEST,M3.5.0,M10.5.0/3") // Offset by +01:00
#define TZ_Europe_Kaliningrad	PSTR("EET-2") // Offset by +02:00
```

All of the examples above were tested today @ ~1604351488 unix timestamp.

```cpp

#ifndef STASSID
#define STASSID "VM723402-2G"
#define STAPSK  "gmwscxnu"
#endif

#include <TZ.h>

#define MYTZ TZ_America_Sao_Paulo

#include <ESP8266WiFi.h>
#include <coredecls.h>
#include <PolledTimeout.h>

#include <time.h>
#include <sys/time.h>

extern "C" int clock_gettime(clockid_t unused, struct timespec *tp);

static time_t now;
static esp8266::polledTimeout::periodicMs showTimeNow(5000);

String toISOString(tm* time) {
  char time_string[25];
  strftime(time_string, 25, "%FT%T", time);
  return String(time_string);
}

void showTime() {
  now = time(nullptr);
  Serial.print("gmtime:     ");
  Serial.println(toISOString(gmtime(&now)));
  Serial.print("localtime:  ");
  Serial.println(toISOString(localtime(&now)));
}

void time_is_set_scheduled() {
  showTime();
}

void setup() {
  Serial.begin(115200);
  Serial.println("Starting...");

  settimeofday_cb(time_is_set_scheduled);

  // Set the timezone
  configTime(MYTZ, "pool.ntp.org");

  // start network
  WiFi.persistent(false);
  WiFi.mode(WIFI_STA);
  WiFi.begin(STASSID, STAPSK);
}

void loop() {
  if (showTimeNow) {
    showTime();
  }
}

```

### Debug Messages

One example, when using TZ_America_Sao_Paulo, local time should be offset -03:00, instead it is offset +03:00:

```
gmtime:     2020-11-02T20:36:51
localtime:  2020-11-02T23:36:51
gmtime:     2020-11-02T20:36:56
localtime:  2020-11-02T23:36:56
gmtime:     2020-11-02T20:37:01
localtime:  2020-11-02T23:37:01
gmtime:     2020-11-02T20:37:06
localtime:  2020-11-02T23:37:06
```

d-a-v:
Timezones coded with `<±nn>±nn<±nn>[±nn][,...]` are incorrectly parsed by newlib's TZ parser.
In these timezones, `<±nn>` is used because there is no matching name like in other timezones.
Example:
- `GMT0BST` for London, (GMT is known, BST is `British Summer Time`).
- `<+1030>-10:30<+11>-11` for Lord Howe Island.

These names GMT,BST,CET,... are meant to be human readable but are useless in our use case.
Replacing `<±nn>` occurences by `UNK`nown allows newlib's TZ parser to nicely interpret all timezones.
`UNK-10:30UNK-11` is then correctly parsed.

Ultimately, newlib should handle such TZ descriptors. Until then, they are handled by this script.

Fixes #7690


earlephilhower (approved):

earlephilhower:
Would you like to put in a bug on the newlib repo for this change?  https://github.com/earlephilhower/newlib-xtensa

d-a-v:
[For reference](https://github.com/earlephilhower/newlib-xtensa/issues/12)

PR #7704, opened 2020-11-12
Handle breakpoints via postmortem when GDB is not running.

mhightower83:
In the absence of GDB, install a small breakpoint handler that clears EXCM and
executes an illegal instruction. This leverages the existing exception
handling code in the SDK for `ill` instruction. At postmortem processing,
the real event is identified and reported. This allows for alerting of embedded
breakpoints. In the stack trace, `epc1` is updated with the value of `epc2` so
that the "ESP Exception Decoder" will identify the line where the BP occurred.

The SDKs default behavior was to loop on breakpoints waiting for interrupts
`>2` until the HWDT kicks in. The current "C++" compiler will embed breakpoints
when it detects a divide by zero at compile time.

Expand technique to preserve more of the current state and to also report
Kernel and Double exceptions through postmortem.

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -33,8 +33,31 @@
 #include "gdb_hooks.h"
 #include "StackThunk.h"
 #include "coredecls.h"
+#include "esp8266_undocumented.h"
+#include "core_esp8266_features.h"
 
 extern "C" {
+
+#if defined(DEBUG_ESP_PORT) || defined(DEBUG_ESP_EXCEPTIONS)
+extern void _DebugExceptionVector(void);
+extern void _KernelExceptionVector(void);
+extern void _DoubleExceptionVector(void);
+
+// Context structure used by Debug, Double, Kernel Exception and unhandled
+// exception stubs to build a stack frame with potentially useful addresses for
+// EXP Exception Decoder's use. Also used to save some special registers
+// that were overwritten by using the `ill` instruction.
+struct EXC_CONTEXT{
+    uint32_t ps;          // +0
+    uint32_t pad1;        // +4
+    uint32_t pad2;        // +8
+    uint32_t excsave2;    // +12
+    uint32_t exccause;    // +16
+    uint32_t epc1;        // +20
+    uint32_t a1;          // +24
+    uint32_t excsave1;    // +28, a0 at the time of the event
+};
+#endif
 
 // These will be pointers to PROGMEM const strings
 static const char* s_panic_file = 0;
@@ -128,6 +151,67 @@ void __wrap_system_restart_local() {
 
     ets_install_putc1(&uart_write_char_d);
 
+    /*
+      Check for Exceptions mapped into User Exception.
+    */
+    if (rst_info.reason == REASON_EXCEPTION_RST && rst_info.exccause == 0) {
+#if defined(DEBUG_ESP_PORT) || defined(DEBUG_ESP_EXCEPTIONS)
+        if ((rst_info.epc1 - 11u) == (uint32_t)_DoubleExceptionVector) {
+            struct EXC_CONTEXT *exc_context;
+            asm volatile("rsr.excsave2 %0\n\t" : "=a" (exc_context) :: "memory");
+            // Use Special Register values from before the `ill` instruction.
+            // Note, for a double exception the exccause is that of the second exception.
+            // epc1 is still the address of the 1st exception, its cause is now unknown.
+            rst_info.exccause = exc_context->exccause;
+            rst_info.epc1 = exc_context->epc1;
+            // Double Exception
+            if (rst_info.depc) {
+                ets_printf_P(PSTR("\nDouble Exception"));
+                // The current "ESP Exception Decoder" does not process the value in depc
+                // Set a fake epc1 so the decoder identifies the line of the Double Exception
+                rst_info.epc1 = rst_info.depc;
+                // The "ESP Exception Decoder" is also helped by updating the
+                // context saved on the stack.
+                exc_context->epc1 = rst_info.depc;
+            }
+            // BP - Debug Exception
+            else if (rst_info.epc2) {
+                if (rst_info.epc2 == 0x4000dc4bu) {
+                    // Unhandled Exceptions land here. 'break 1, 1' in _xtos_unhandled_exception
+                    if (20u == exc_context->exccause) {
+                        // This could be the result from a jump or call; a call
+                        // is more likely. eg. calling a null callback function
+                        // For the call case, a0 is likely the return address.
+                        rst_info.epc1 = exc_context->excsave1;
+                    }
+                    ets_printf_P(PSTR("\nXTOS Unhandled exception - BP"));
+                }
+                else if (rst_info.epc2 == 0x4000dc3cu) {
+                    // Unhandled interrupts land here. 'break 1, 15' in _xtos_unhandled_interrupt
+                    ets_printf_P(PSTR("\nXTOS Unhandled interrupt - BP"));
+                    // Don't know what should be done here if anything.
+                }
+                else {
+                    ets_printf_P(PSTR("\nDebug Exception"));
+                    // The current Exception Decoder does not process the value in epc2
+                    // Set a fake epc1 so the Exception Decoder identifies the correct BP line
+                    rst_info.epc1 = rst_info.epc2;
+                }
+            }
+            // Kernel Exception
+            else {
+                ets_printf_P(PSTR("\nKernel Exception"));
+            }
+            ets_printf_P(PSTR(" - excsave1=0x%08x excsave2=0x%08x epc1=0x%08x\n"),
+                exc_context->excsave1, exc_context->excsave2, exc_context->epc1 );
+        }
+#endif
+        // The GCC divide routine in ROM jumps to the address below and executes ILL (00 00 00) on div-by-zero
+        // In that case, print the exception as (6) which is IntegerDivZero
+        if (rst_info.epc1 == 0x4000dce5u) {
+            rst_info.exccause = 6;
+        }
+    }
     cut_here();
 
     if (s_panic_line) {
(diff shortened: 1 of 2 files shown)


earlephilhower (approved):
Very nice, as always!

Have you seen that G++ behavior (inserting BP instead of letting the FP div-0 fault) on other platforms?  It may be something we can fix in the machine definition, but I haven't the stomach to look..

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -128,6 +151,67 @@ void __wrap_system_restart_local() {
 
     ets_install_putc1(&uart_write_char_d);
 
+    /*
+      Check for Exceptions mapped into User Exception.
+    */
+    if (rst_info.reason == REASON_EXCEPTION_RST && rst_info.exccause == 0) {
+#if defined(DEBUG_ESP_PORT) || defined(DEBUG_ESP_EXCEPTIONS)
+        if ((rst_info.epc1 - 11u) == (uint32_t)_DoubleExceptionVector) {
```
devyte, line 159: Why 11? what does that magic number mean?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -128,6 +151,67 @@ void __wrap_system_restart_local() {
...
+                // The current "ESP Exception Decoder" does not process the value in depc
+                // Set a fake epc1 so the decoder identifies the line of the Double Exception
+                rst_info.epc1 = rst_info.depc;
+                // The "ESP Exception Decoder" is also helped by updating the
+                // context saved on the stack.
+                exc_context->epc1 = rst_info.depc;
+            }
+            // BP - Debug Exception
+            else if (rst_info.epc2) {
+                if (rst_info.epc2 == 0x4000dc4bu) {
```
devyte, line 179: What is this magic address? What is there?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -128,6 +151,67 @@ void __wrap_system_restart_local() {
...
+                    // Unhandled Exceptions land here. 'break 1, 1' in _xtos_unhandled_exception
+                    if (20u == exc_context->exccause) {
```
devyte, line 181: Why 20? what does the magic number mean?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -128,6 +151,67 @@ void __wrap_system_restart_local() {
...
+            else {
+                ets_printf_P(PSTR("\nKernel Exception"));
+            }
+            ets_printf_P(PSTR(" - excsave1=0x%08x excsave2=0x%08x epc1=0x%08x\n"),
+                exc_context->excsave1, exc_context->excsave2, exc_context->epc1 );
+        }
+#endif
+        // The GCC divide routine in ROM jumps to the address below and executes ILL (00 00 00) on div-by-zero
+        // In that case, print the exception as (6) which is IntegerDivZero
+        if (rst_info.epc1 == 0x4000dce5u) {
```
devyte, line 211: Per the comment above, this magic number seema to be a ROM address. What is there? a 0x0? Can the value be moved into a constexpr and commented?

mhightower83:
@earlephilhower Thanks!

> Have you seen that G++ behavior ...  on other platforms?

This is the only platform that I have significantly used G++. For me, the issue surfaced with the toolchain 10.x update.

I would expect this to be a build controllable option. I looked briefly and didn't find anything that worked in changing the outcome. At the time, it felt like the time investment for me vs. the gain was too great. GCC is a relatively new compiler for me and we do not work well together.

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -322,5 +403,199 @@ void __stack_chk_fail(void) {
...
+  calling, etc.
+
+  A problem with the `_xtos_unhandled_exception` function is that it handles the
+  event with a Debug Exception BP. If you don't have GDB running you see a HWDT
+  reset. If interrupts are at INTLEVEL 2 or higher (a BP instruction becomes a
+  NOP), you still see a HWDT reset regardless of running GDB.
+*/
+typedef void (* _xtos_handler)(void);
+static  _xtos_handler * const _xtos_exc_handler_table = (_xtos_handler *)0x3FFFC000u;
+#define ROM_xtos_unhandled_exception (reinterpret_cast<_xtos_handler>(0x4000dc44))
```
devyte, line 533: What is this magic address? Is it a ROM function? Does it make sense to move to the file where the ROM functions are mapped to addresses?
If not, does it make sense to change this define into a constexpr with a type?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -322,5 +403,199 @@ void __stack_chk_fail(void) {
...
+  can happen when a function prototype is given the weak attribute but never
+  defined, or forgetting to test a call back function pointer for NULL before
+  calling, etc.
+
+  A problem with the `_xtos_unhandled_exception` function is that it handles the
+  event with a Debug Exception BP. If you don't have GDB running you see a HWDT
+  reset. If interrupts are at INTLEVEL 2 or higher (a BP instruction becomes a
+  NOP), you still see a HWDT reset regardless of running GDB.
+*/
+typedef void (* _xtos_handler)(void);
```
devyte, line 531: Prefer using= over typedef for type definitions. In the case of a function ptr:
```cpp
using _xtos_handler = void (*)(void);
```

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -322,5 +403,199 @@ void __stack_chk_fail(void) {
...
+static  _xtos_handler * const _xtos_exc_handler_table = (_xtos_handler *)0x3FFFC000u;
```
devyte, line 532: Why this magic address? What is there? Can it be moved out to a define or constexpr with comments?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -322,5 +403,199 @@ void __stack_chk_fail(void) {
...
+    if (old_wrapper == match || NULL == match) {
+        _xtos_exc_handler_table[cause] = replacement;
+    }
+}
+
+static void install_unhandled_exception_handler(void) {
+    // Only replace Exception Table entries still using the orignal Boot ROM
+    // _xtos_unhandled_exception handler.
+    for (size_t i = 0; i < 64; i++) {
+        replace_exception_handler_on_match(
```
devyte, line 571: I understand that this 64 is the number of entries in the exception handler? Can the value be moved out to a constexpr or define?

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -322,5 +403,199 @@ void __stack_chk_fail(void) {
...
+            (_xtos_handler)postmortem_xtos_unhandled_exception);
+    }
+}
+
+void postmortem_init(void) {
+    if (!gdb_present()) {
+        // Install all three (interdependent) exception handler stubs.
+        // Length of copy must be rounded up to copy memory in aligned 4 byte
+        // increments to comply with IRAM memory access requirements.
+        uint32_t save_ps = xt_rsil(15);
```
devyte, line 583: What do 16 and 32 mean? doea it make sense to move to a define. constexpr and comments?

devyte (changes requested):
I'm not aware of the exception handler mechanism details or postmortem, so I can't offer a review based on functionality. But @earlephilhower already approved, and he's aware of that :) 
Code-wise, the one thing that bothers me is all the literal magic numbers inserted throughout the code. I suggest moving all literals to constexpr or defines, and having each one with its own comment explaining what it means. ROM function addresses, if any, should be moved to a function-address map file (there's already one with the known ROM functions, you could just add to it).

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: That is the offset into the `_DoubleExceptionVector` where the `ill` instruction was executed. I found a better way to deal with that. I'll push later,

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: That is the address in the Boot ROM for the breakpoint instruction that executes for an unhandled exception. I'll replace with a `constexpr`.

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: > EXCCAUSE value 20, InstFetchProhibitedCause - An instruction fetch referenced a page mapped with an attribute that does not permit instruction fetch.

I'll add a `constexpr` for this in the next push.

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: While `_xtos_unhandled_exception` is in the linker `.ld` file, it may have been overridden by the sketch or future SDK(not likely). We require the original Boot ROM function address to limit our override to those old values in the table.

With the new toolchain change `constexpr` doesn't work for this. It does not like an integer being cast to a function:
```cpp
constexpr _xtos_handler ROM_xtos_unhandled_exception = (reinterpret_cast<_xtos_handler>(0x4000dc44));

error: 'reinterpret_cast' from integer to pointer
```
It works if I replace `constexpr` with `const` but I was concerned it would take up DRAM. I just tried it and in this case, it doesn't appear to change DRAM usage.

I'll change it to `const` for now.

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: okay

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: okay

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: I initially wasn't sure how to generate values at compile time from the ASM blocks sizes. I think I have that resolved.

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: Yes, that address has the `ill` (illegal) instruction. Added comments and `constexpr`.

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: Dispatch table for EXCCAUSE  values. Added `constexpr` and comments.

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: I thought I had this covered, but I don't think so anymore. I think my problem stems from trying to reference xtensa sources that I can find on the Internet to fill in some of these missing function types. In the case of `_xtos_handler`, I am having trouble. They applied it to two very different function tables and it matches neither exactly. 
1) The first table is a set of ASM functions that do not comply with "C" conventions.
   * For `.cpp` files the typedef boils down to `typedef void (*_xtos_handler)(...);`
   * `extern _xtos_handler _xtos_exc_handler_table[];`
   * Of the set of ASM functions that are used in this table, there is a "C" wrapper function (`_xtos_c_wrapper_handler`) that references the 2nd table.
2) The 2nd table of "C" styled exception handling functions, defines its entries with the same typedef.
   * `extern _xtos_handler _xtos_c_handler_table[];`
   * for the functions in this table I think this would better match the function call:
     * `typedef void (_xtos_c_exception_handler_func)(struct __exception_frame *ef, int cause);`
     * `typedef _xtos_c_exception_handler_func *_xtos_c_handler;`
     * `extern _xtos_c_handler _xtos_c_handler_table[];`

My quandary is: 
 1) should I keep the typedef's I found and cast everything to fit? or
 2) Ignore the existing ones and hope those files never need to be added to this code base and define new ones?
     * for the first table I don't see a perfect fit for a non-"C" ASM. I think the current typedef would be fine.

Also affected: https://github.com/esp8266/Arduino/pull/7060#issue-370664751

mhightower83:
@devyte I think this last push, resolves your remaining issues? Let me know.

PR #7706, opened 2020-11-13
Provide String::indexOf for a char* needle

### LINKED ISSUE #7705: Unable to call String::indexOf with char* for a needle
Hey I'm trying to avoid allocating+copying to search for a needle in a haystack. So it would be great if String provided a indexOf method that receives a char*.

Ideally it would provide a method for __FlashStringHelper, but since it uses asm/intrinsics to search I don't know if there is a way to do the search without ignoring that it's from PROGMEM. Is it possible?

Anyway I'm happy with just char*.

paulocsanz:
Fixes #7705

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -628,7 +628,7 @@ int String::indexOf(const __FlashStringHelper *s2, unsigned int fromIndex) const
     if (found == NULL)
         return -1;
     return found - buffer();
-}}
+}
 
 int String::indexOf(const char *s2) const {
     return indexOf(s2, 0);
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -622,7 +622,7 @@ int String::indexOf(const __FlashStringHelper *s2) const {
 }
 
 int String::indexOf(const __FlashStringHelper *s2, unsigned int fromIndex) const {
-    return indexOf((char*) s2, fromIndex);
+    return indexOf((const char*) s2, fromIndex);
 }
 
 int String::indexOf(const char *s2) const {
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -617,6 +617,19 @@ int String::indexOf(char ch, unsigned int fromIndex) const {
     return temp - buffer();
 }
 
+int String::indexOf(const __FlashStringHelper *s2) const {
+    return indexOf(s2, 0);
+}
+
+int String::indexOf(const __FlashStringHelper *s2, unsigned int fromIndex) const {
+    if (fromIndex >= len())
+        return -1;
+    const char *found = strstr_P(buffer() + fromIndex, (char *) s2);
+    if (found == NULL)
+        return -1;
+    return found - buffer();
+}}
+
 int String::indexOf(const char *s2) const {
     return indexOf(s2, 0);
 }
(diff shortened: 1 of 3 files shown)


cores/esp8266/WString.cpp
```diff
@@ -617,6 +617,19 @@ int String::indexOf(char ch, unsigned int fromIndex) const {
...
 }
 
+int String::indexOf(const char *s2) const {
+    return indexOf(s2, 0);
+}
+
+int String::indexOf(const char *s2, unsigned int fromIndex) const {
+    if (fromIndex >= len())
+        return -1;
+    const char *found = strstr(buffer() + fromIndex, s2);
```
earlephilhower, line 635: If you change this to `strstr_P` it will work with `PROGMEM` strings natively.

earlephilhower (changes requested):
Thanks for not only filing an issue, but proposing a solution too!

You mentioned progmem in your issue and changing to `strstr_P`will make it "just work" with RAM or PROGMEM strings.

cores/esp8266/WString.cpp
paulocsanz: Perfect, thanks!

paulocsanz:
I've went ahead and implemented the changes. Should I change all `strstr` calls to `strstr_P` then? The documentation doesn't say anything about passing a s2 to `strstr` that does not live in program memory.

But again I'm not an expert on this ecosystem.

d-a-v:
> Btw did I break anything? The CI seems to have errored (I may have hurt their feelings by force pushing too much lol).

[Yes](https://github.com/paulocsanz/Arduino/blob/c2017d0c3e56b10acee6311356ddd4c2a76f314b/cores/esp8266/WString.cpp#L631)

(click on "details" on any CI line)

paulocsanz:
Thank you @d-a-v! I've just fixed it.

cores/esp8266/WString.cpp
d-a-v: I think that @earlephilhower was suggesting that only one full body of `::indexof()` was needed if `strstr_P()` is used instead of `strstr()`.

paulocsanz: I see, I'm keeping both to prevent `F(...)` to be converted to String implicitly, but I will make the char* version use `strstr_P` too then. Is that cool?

d-a-v: I think so !

paulocsanz: Done!

d-a-v (approved):
LGTM

cores/esp8266/WString.cpp
```diff
@@ -617,19 +617,35 @@ int String::indexOf(char ch, unsigned int fromIndex) const {
...
 }
 
-int String::indexOf(const String &s2) const {
+int String::indexOf(const __FlashStringHelper *s2) const {
     return indexOf(s2, 0);
 }
 
-int String::indexOf(const String &s2, unsigned int fromIndex) const {
+int String::indexOf(const __FlashStringHelper *s2, unsigned int fromIndex) const {
+    return indexOf((char*) s2, fromIndex);
```
earlephilhower, line 625: This will generate a warning when used because of the cast of a `const <blah>` to a mutable `<blah>`.

I'd recommend:  `return indexOf((const char*) s2, fromIndex);`

earlephilhower (changes requested):
Other than the single const change noted below, LGTM.  Thx!

earlephilhower (approved):
Thx!  LGTM.

PR #7707, opened 2020-11-14
Fix sntp_get_real_time prototype in header

earlephilhower:
sntp_get_real_time takes a `time_t` in the core, but in the header it
was listed as `long`.  Make them both match by changing the header.


earlephilhower:
This was spotted while doing the newlib 3.3 port where `time_t` has gone to 64-bit (i.e. a `long long`) to avoid the year 2038 problem.  On the current core, this is just a cleanup and shouldn't change any code.

PR #7708, opened 2020-11-14
BREAKING - Upgrade to upstream newlib 4.0.0 release

earlephilhower:
Includes 64 bit time_t and 5 years of updates.

Binary incompatible with libraries which use time_t (due to the size
difference).  Recompiling with the new newlib should be sufficient for
most libraries, assuming source is available.


earlephilhower:
Note that this requires a new toolchain (since GCC includes newlib), so be sure to run `get.py` if you want to try it out.

earlephilhower:
BearSSL_Validation passes, but there's a crash when running any WebServer demo while doing `strtol` on a `String.c_str()`. 

````
Exception 3: LoadStoreError: Processor internal physical address or data error during load or store
PC: 0x40211211: _strtol_l at /home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/stdlib/strtol.c line 154
EXCVADDR: 0x4024527f

Decoding stack results
0x4020a7bc: String::copy(char const*, unsigned int) at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/WString.cpp line 190
0x4020a921: String::operator=(char const*) at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/WString.cpp line 227
0x40209a74: HardwareSerial::write(unsigned char const*, unsigned int) at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/HardwareSerial.h line 164
0x40211360: strtol at /home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/stdlib/../locale/setlocale.h line 237
0x4020eb3c: atoi at /home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/stdlib/atoi.c line 52
0x40202c17: esp8266webserver::ESP8266WebServerTemplate ::_parseRequest(WiFiClient&) at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/WString.h line 323
0x4020b55e: __preloop_update_frequency() at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_main.cpp line 88
0x4010049e: millis() at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring.cpp line 193
0x4010049e: millis() at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_wiring.cpp line 193
0x4020a14c: Print::println() at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/Print.h line 57
0x40203a57: esp8266webserver::ESP8266WebServerTemplate ::handleClient() at /home/earle/Arduino/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h line 261
0x40206d68: esp8266::MDNSImplementation::MDNSResponder::_process(bool) at /home/earle/Arduino/hardware/esp8266com/esp8266/libraries/ESP8266mDNS/src/LEAmDNS_Control.cpp line 91
0x401003f8: ets_post(uint8, ETSSignal, ETSParam) at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_main.cpp line 177
0x40203b34: loop() at /home/earle/Arduino/hardware/esp8266com/esp8266/libraries/ESP8266WebServer/examples/HelloServer/HelloServer.ino line 155
0x4020b6cc: loop_wrapper() at /home/earle/Arduino/hardware/esp8266com/esp8266/cores/esp8266/core_esp8266_main.cpp line 197
````

**Comment by Jason2866** [comment]
Quiet impressed. With your latest commits Tasmota does compile successfull and it does start!!
After some time sync problems it gets NTP time and is up already for 10 minutes.

**Comment by earlephilhower** [comment]
@Jason2866 one thing to be aware of, maybe related to the NTP stuff:  If you've got a binary struct saved somewhere (say, with last time updated), the size of that struct will now change and you'll probably be interpreting garbage at that point (4->8 byte time_t).

This reminds me that I need to update the LittleFS time_t attribute.

File: .github/workflows/pull-request.yml [hunk 1]
```diff
@@ -25,12 +25,12 @@ jobs:
...
-#      uses: actions/cache@v2
-#      with:
-#        path: ./tools/dist
-#        key: key-linux-toolchain
+    - name: Cache Linux toolchain
+      id: cache-linux
+      uses: actions/cache@v2
+      with:
+        path: ./tools/dist
+        key: key-linux-toolchain
```
mcspr: iirc, changing 'key' will force it to use a different cache
it will be expired at some point, but it is not as clear as something like travis (rip)

earlephilhower: Thx.  I think generally we're OK and the un-done patch here was just a stopgap during development.

The original commenting-out was because the releases have a datestamp in them which ends up making all builds in a day conflict.  So get.py would find gcc-20121101.tgz(from 9AM) in the cached dir and not try and d/l the newer one gcc-20121101.tgz(from 11AM).  The cached version's MD5 would mismatch, so the build failed w/o a toolchain.  Easy fix, drop the caching.  W/everything coming from GH.com, anyway, the cache probably doesn't make much difference...

mcspr: True, speeds on github are really good, so cache most of the time takes longer than just fetching the sources directly...

I think it also could be this 'template', making key depend on the .json so it is always new whenever toolchain is modified
```
key: ${{ runner.os }}-${{ hashFiles('package/package_esp8266com_index.template.json') }}
```
via https://github.com/actions/cache#creating-a-cache-key

earlephilhower: That's an awesome idea!  Right now the cache will grow w/obsolete versions of the tools, but one based on the JSON would fix that problem.

Did you want the PR for this?  OTW I can throw it together quickly.

PR #7709, opened 2020-11-15
ETag support for WebServer

imwhocodes:
This pull request for ETag cache support on static files

This is done by deriving `ESP8266WebServerETag` from `ESP8266WebServer`

The hash for Etag is done with the native `MD5Builder`

This features is very useful when developing web sites on esp8266, only file actually changed during development are downloaded again

Also the existing `max-age=` can be too broad and is easy to mismanage


libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -165,6 +165,51 @@ class StaticRequestHandler : public RequestHandler<ServerType> {
...
+    using WebServerType = ESP8266WebServerTemplate<ServerType>;
+
+    public:
+    bool handle(WebServerType& server, HTTPMethod requestMethod, const String & requestUri) override {
+        if (!SRH::canHandle(requestMethod, requestUri)){
+            return false;
+        }
+
+        if(server.header("If-None-Match") == SRH::_cache_header){
+            // Serial.println("Sending 304!!!");
```
earlephilhower, line 186: Can we please drop the commented debug messages, or bracket the prints with the `DBGWS` macro based off of `DEBUG_ESP_HTTP_SERVER`

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -618,6 +619,20 @@ void ESP8266WebServerTemplate<ServerType>::collectHeaders(const char* headerKeys
   }
 }
 
+template<typename ServerType>
+void ESP8266WebServerETagTemplate<ServerType>::collectHeaders(const char* headerKeys[], const size_t headerKeysCount) {
+  WST::_headerKeysCount = headerKeysCount + 2;
```
earlephilhower, line 624: Isn't the `WST::` bit implied for the class local variables?  Others template methods seem to be fine w/o it...

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -831,4 +846,37 @@ String ESP8266WebServerTemplate<ServerType>::responseCodeToString(const int code
...
+template<typename ServerType>
+void ESP8266WebServerETagTemplate<ServerType>::serveStaticETag(const char* uri, FS& fs, const char * path) {
+
+  File toHash = fs.open(path, "r");
+
+  if(toHash){
+
+    MD5Builder calcMD5;
+    calcMD5.begin();
+
```
earlephilhower, line 863: It's minor and a taste thing, but unless you're being paid by the KLOC [old IBMer? :) ] what's the reason for the skipping every other line?  Could it be a linefeed issue in an editor?

earlephilhower (changes requested):
Thanks for the PR!  Very interesting, I see how it could be of use.

Minor little things noted below.  Also, can you please put a note in the docs or an example on how to enable this for an end-user?  Doesn't need to be a massive writeup, but a line or three in the docs would be of immense use later on.

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
imwhocodes: `ESP8266WebServerETagTemplate<ServerType>` inherit form `ESP8266WebServerTemplate<ServerType>`, the variables that I'm trying to access are from the parent class: `WTS` is a [shorthand](https://github.com/esp8266/Arduino/pull/7709/files/ae3d925e648743486cf07b7d84b8355e72c6b111#diff-799c46c81de7a8939be0ec4f4aef5ad592378592d337950447996243b78f37e0R307) for `ESP8266WebServerTemplate<ServerType>`

I went for the full baseclass templated name, I could also use `this->` to access the parent's variables but [they are not plainly accessible in the current scope](https://stackoverflow.com/questions/31231875/cant-access-variable-in-template-base-class)

Is `this->` more readable?

d-a-v:
Nice, it allows to reduce network activity !
I have a number of questions:

Is this a feature that is only useful while developing web sites, or is it also useful in production ?
If so, should it be transparently enabled ?
```
- using ESP8266WebServer = esp8266webserver::ESP8266WebServerTemplate<WiFiServer>;
+ using ESP8266WebServer = esp8266webserver::ESP8266WebServerETagTemplate<WiFiServer>;
```
or with an option (arduino menu and / or global define) ?
```
using ESP8266WebServerNoETag = esp8266webserver::ESP8266WebServerTemplate<WiFiServer>;
using ESP8266WebServerETag = esp8266webserver::ESP8266WebServerETagTemplate<WiFiServer>;
#if ETAG_ENABLED
using ESP8266WebServer = ESP8266WebServerETag;
#else
using ESP8266WebServer = ESP8266WebServerNoETag;
#endif
```
or on sketch request (as proposed by your PR) ?
```
using ESP8266WebServer = esp8266webserver::ESP8266WebServerTemplate<WiFiServer>;
using ESP8266WebServerETag = esp8266webserver::ESP8266WebServerETagTemplate<WiFiServer>;
```

Does it slow down web requets on big files since it is calculated on every request ?
Did you think of caching the results ?

imwhocodes:
Thanks for the appreciation

> Nice, it allows to reduce network activity !
> I have a number of questions:
> 
> Is this a feature that is only useful while developing web sites, or is it also useful in production ?

IMHO it depends on the situation:
- `Etag` cache control tag always do a round trip for the requested page, but the response is only a status code (`304`) if the file is unchanged
- `max-age` cache control tag most of the time do not even ask for the page, but too short of a time and you don't cache, too long and you can miss updates (or the browser can ignore the timeout and ask the page again anyway)

**But the two cache control methods could be used together for maximum caching (in the next implementation?)**

> If so, should it be transparently enabled ?
> 
> ```
> - using ESP8266WebServer = esp8266webserver::ESP8266WebServerTemplate<WiFiServer>;
> + using ESP8266WebServer = esp8266webserver::ESP8266WebServerETagTemplate<WiFiServer>;
> ```
> 
> or with an option (arduino menu and / or global define) ?
> 
> ```
> using ESP8266WebServerNoETag = esp8266webserver::ESP8266WebServerTemplate<WiFiServer>;
> using ESP8266WebServerETag = esp8266webserver::ESP8266WebServerETagTemplate<WiFiServer>;
> #if ETAG_ENABLED
> using ESP8266WebServer = ESP8266WebServerETag;
> #else
> using ESP8266WebServer = ESP8266WebServerNoETag;
> #endif
> ```
> 
> or on sketch request (as proposed by your PR) ?
> 
> ```
> using ESP8266WebServer = esp8266webserver::ESP8266WebServerTemplate<WiFiServer>;
> using ESP8266WebServerETag = esp8266webserver::ESP8266WebServerETagTemplate<WiFiServer>;
> ```

This is a good question...
Implementing it transparently would be very cool but I didn't because it would be a breaking change:
- It increase the memory footprint (a `String` of size 24) for every statically served page (a locally stored Etag string plus another possible Etag parsed from browser request header during handling) 
- For now it don't support the dynamic append of `index.htm` or `index.html` if requested page is a folder (being Etag calculated only once)

I opted for the  `on sketch request` for backward compatibility and for keeping old code excepted behaviour, but being this PR on the 3.x branch maybe can this different behaviour acceptable?

> Does it slow down web requets on big files since it is calculated on every request ?
> Did you think of caching the results ?

Actually as implemented right now the hash in only computed once when you call `ESP8266WebServerETag::serveStaticETag`, then `StaticRequestETagHandler::handle` only check if the browser request Etag is the same as the one before computed

imwhocodes:
With this last commit now there is **Transparent ETag support with backward compatibility** by splitting and specialising `StaticRequestHandler` for the `StaticDirectoryRequestHandler` and `StaticFileRequestHandler` case

Inside `StaticDirectoryRequestHandler` the code is almost the same as the one for the original `StaticRequestHandler` but the `_isFile` propriety and logic is missing being implicit of the polymorphism

On the `StaticFileRequestHandler` the hash is compute once at creation and stored as `byte` array, only during `handle` it is encoded as base64 to minimize overhaul memory usage

The choice if to use the `Directory` or `File` polymorphism is done during the call to `esp8266webserver::serveStatic` based on the existence of the argument given fs path

(I will add examples and documentation in a few days)

libraries/ESP8266WebServer/src/detail/RequestHandlersImpl.h
earlephilhower: I see this is taken care of in latest PR, thanks!

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
earlephilhower: Thanks for clearing this up!

earlephilhower (approved):
Sorry for the delay.  Looks good to me now, thanks!

d-a-v (approved):
Thanks!
Approving without doc nor example because the added feature is configuration-less and transparent for the user.

mathertel:
I have added some improvements in https://github.com/esp8266/Arduino/pull/8227
to make the solution more configurable and working with directories.

@d-a-v  || @earlephilhower: This can one be closed.

PR #7715, opened 2020-11-20
CI waveform flavour select

d-a-v:
set every other ci builder to use waveform phase lock

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -152,7 +152,7 @@ function install_ide()
     #local idever='nightly'
     #local ideurl='https://www.arduino.cc/download.php?f=/arduino-nightly'
 
-    local idever='1.8.10'
+    local idever='1.8.13'
     local ideurl="https://downloads.arduino.cc/arduino-$idever"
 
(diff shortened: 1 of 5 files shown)


tests/common.sh
```diff
@@ -66,6 +66,8 @@ function build_sketches()
     local sketches=$(find $srcpath -name *.ino | sort)
     print_size_info >size.log
     export ARDUINO_IDE_PATH=$arduino
+    local pwm_phase=""
+    [ $(( $build_rem % 2 )) -eq 0 ] && local pwm_phase="--waveform_phase"
```
earlephilhower, line 70: Bash is a monster, isn't it?  I was about to make a comment that you wanted `==` but then realized you're generating local.

Not sure if the `local` here is needed since you already made one on the prior line.

earlephilhower (approved):
LGTM after the log dump is removed.

tests/common.sh
d-a-v: I removed the second local

PR #7724, opened 2020-11-25
PROGMEM compatibility changes to String

drzony:
Changed contructor to use strlen_P
Changed copy to use memcpy_P instead of memmove_P

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -32,7 +32,7 @@
 String::String(const char *cstr) {
     init();
     if (cstr)
-        copy(cstr, strlen(cstr));
+        copy(cstr, strlen_P(cstr));
 }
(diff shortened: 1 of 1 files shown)


drzony:
@d-a-v As discussed on gitter

cores/esp8266/WString.cpp
```diff
@@ -186,7 +186,7 @@ String &String::copy(const char *cstr, unsigned int length) {
         return *this;
     }
     setLen(length);
-    memmove_P(wbuffer(), cstr, length + 1);
+    memcpy_P(wbuffer(), cstr, length + 1);
```
d-a-v, line 189: It should work with `memmove_P()`. Can you try it ?

d-a-v (approved):

PR #7727, opened 2020-11-28
Added GZipped OTA support in elf2bin and PlatformIO

drzony:


d-a-v:
Thanks @drzony ! That looks fine.
Basically from what I understand, the regular firmware file is replaced by the compressed one, and the original one renamed to `.orig`.
Is that because PIO can't handle another name than `sketch.bin` ?

drzony:
No, that's because `elf2bin.py` takes `--out` parameter, so the gzipped file needs to be named according to this parameter. The `.orig` file is just a backup for reference

earlephilhower (approved):
LGTM.  Only PIO and custom scripts will use this action, so I don't see any core impact.

d-a-v (approved):
Thanks !

PR #7733, opened 2020-12-01
Specify where Additional Boards Manager URLs is

jannst:
As a newcomer to Arduino IDE, it is not obvious where the "Additional Boards Manager URLs" field is


d-a-v (approved):
Thanks !

PR #7736, opened 2020-12-01
Clarified implementation of the signed OTA implementation

DavyLandman:
The documentation was a bit unclear what was happening, so after reading the source code I thought to update the documentation a bit.

The most stuff is clarification, but I had to remove the `INCLUDING-THIS-32-BITS` part, as the size is just the size of the signature (hard coded to `[0x0, 0x1, 0x0, 0x0]` (256 little endian)).


earlephilhower (approved):
Good catch on the layout.  LGTM.

PR #7737, opened 2020-12-01
Update to pyserial 3.5, esptool 3.0

earlephilhower:
Recent releases may work better with some newer MacOS releases
according to their commit logs.

### DIFF
--- a/doc/ota_updates/readme.rst
+++ b/doc/ota_updates/readme.rst
@@ -44,9 +44,9 @@ Make your own risk analysis and, depending on the application, decide what libra
 Advanced Security - Signed Updates
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


d-a-v:
esptool has a breaking change:
> The default `esptool.py write_flash --flash_size` argument has changed from detect to keep. This means that if no `--flash_size` argument is passed when flashing a bootloader, the value in the bootloader .bin file header is kept instead of detecting the actual flash size and updating the header. To keep the old behaviour, add the argument `esptool.py [...] write_flash [...] -fs detect`

Should we keep on with default behavior ?

earlephilhower:
Good catch, @d-a-v!  Updated to pass in the detect option automatically.
````
...
Changing baud rate to 460800
Changed.
Configuring flash size...
Auto-detected Flash size: 4MB
Compressed 318240 bytes to 230844...
...
````

d-a-v (approved):
and better flashing logs in the arduino console !
Thanks for having fixed esptool too :)

PR #7745, opened 2020-12-06
makecorever.py: Hide GIT error messages. Fixes #7744

### LINKED ISSUE #7744: Error message "fatal: not a git repository" when using the GIT snapshot
### Basic Infos

#### Platform

- Hardware:				other AI-Thinker ESP8266MOD
- Core Version:			47e7b2d
- Operating System:		Arch Linux

### Settings in IDE

- Module:  			Wemos D1 mini r2
- Flash Mode:			?
- Flash Size: 		4MB/1MB
- Reset Method:		?
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   921600

### Problem Description

If I create a toolchain based on the GIT snapshot (ZIP file), then I always get the following message prior compiling:

    fatal: not a git repository (or any of the parent directories): .git

Reason is that this script tries to fetch some info using GIT:

https://github.com/esp8266/Arduino/blob/master/tools/makecorever.py

The message does not cause any problems. It just looks ugly.

How do you plan to fix this? Hide the message? I could create a pull request if I know how you want this to be fixed.

M-Reimer:


tools/makecorever.py
```diff
@@ -26,7 +26,7 @@ def generate(path, platform_path, git_ver="ffffffff", git_desc="unspecified"):
     def git(*args):
         cmd = ["git", "-C", platform_path]
         cmd.extend(args)
-        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True)
+        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True, stderr=DEVNULL)
```
mcspr, line 29:
```suggestion
        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True, stderr=subprocess.DEVNULL)
```
try: ... except: pass below hides the `name 'DEVNULL' is not defined`

earlephilhower (approved):
LGTM, thanks!

PR #7748, opened 2020-12-09
i2s: adds i2s_rxtxdrive_begin(enableRx, enableTx, driveRxClocks, driveTxClocks)

d-a-v:
~~A `dac`~~ rx/rxDrive/tx/txDrive parameters are added to allow using only i2s in/out data pin without clock pins.
This is necessary for example in no-DAC mode when i2so-bck and i2so-ws are repurposed,
Found when using SDFS because SPI SS GPIO was overlapped.

*edit*: Related: https://github.com/earlephilhower/ESP8266Audio/pull/344

### DIFF
--- a/cores/esp8266/core_esp8266_i2s.cpp
+++ b/cores/esp8266/core_esp8266_i2s.cpp
@@ -64,7 +64,7 @@ typedef struct i2s_state {
   // Callback function should be defined as 'void ICACHE_RAM_ATTR function_name()',
   // and be placed in IRAM for faster execution. Avoid long computational tasks in this
   // function, use it to set flags and process later.
-  bool             txDAC;
+  bool             driveClocks;
 } i2s_state_t;
 
 // RX = I2S receive (i.e. microphone), TX = I2S transmit (i.e. DAC)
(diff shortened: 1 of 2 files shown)


cores/esp8266/core_esp8266_i2s.cpp
```diff
@@ -503,9 +508,11 @@ bool i2s_rxtx_begin(bool enableRx, bool enableTx) {
       // Nothing to clean up yet
       return false; // OOM Error!
     }
-    pinMode(I2SO_WS, FUNCTION_1);
     pinMode(I2SO_DATA, FUNCTION_1);
-    pinMode(I2SO_BCK, FUNCTION_1);
+    if ((tx->txDAC = txDAC)) {
```
earlephilhower, line 512: Assignment inside an `if` condition, naughty!

cores/esp8266/core_esp8266_i2s.cpp
```diff
@@ -493,6 +494,10 @@ float i2s_get_real_rate(){
 }
 
 bool i2s_rxtx_begin(bool enableRx, bool enableTx) {
+  return i2s_rxtxdac_begin(enableRx, enableTx, true);
+}
+
+bool i2s_rxtxdac_begin(bool enableRx, bool enableTx, bool txDAC) {
```
earlephilhower, line 500: I think the naming/API needs to be a bit more generic.  Maybe make it `i2s_rxtxdrive_begin(bool enRx, bool enTx, bool drRxClocks, bool drTxClocks`.  There are other reasons to use I2S w/o the other pins than DAC/NoDAC (IR blaster, video out, RGB control, etc.).  The idea here is we're potentially disabling the clocking drives.

earlephilhower (changes requested):
Thx for fixing this!

earlephilhower (approved):
LGTM, thanks!

PR #7752, opened 2020-12-10
Normalize platform.txt pattern numbers

earlephilhower:
Some tools have issues with pattern.XX numbers above 9 in platform.txt.
Adjust all pattern.XX numbering to start at 1 and remove the numbering
when there is only one step.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -96,15 +96,15 @@ compiler.elf2hex.extra_flags=
 ## needs git
 recipe.hooks.sketch.prebuild.pattern="{runtime.tools.python3.path}/python3" "{runtime.tools.signing}" --mode header --publickey "{build.source.path}/public.key" --out "{build.path}/core/Updater_Signing.h"
 # This is quite a working hack. This form of prebuild hook, while intuitive, is not explicitly documented.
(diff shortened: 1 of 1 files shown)


d-a-v (approved):
:+1:
(someone should be working on a graphical diff understanding when lines are swapped)

mhightower83:
@earlephilhower Which tool has a problem with double digits? I know there is this issue described in [Platform specification, Recipes to compile source code](https://arduino.github.io/arduino-cli/platform-specification/#recipes-to-compile-source-code)
> Warning: hooks recipes are sorted before execution. If you need to write more than 10 recipes for a single hook, pad the number with a zero, for example:

My solution was to sidestep the problem and always use two digits in case the list grew

earlephilhower:
@mhightower83  Sloeber(sp?), the Eclipse plugin, has an issue according to @hreintke in the MDNS gitter channel.

If we have more than 9 steps in a recipe we'll have to re-examine things (at that point a single Python script may be the sanest thing since Python3 is available on all platforms now).  For now, we're far enough from the limit not to worry...

jantje:
The problem in Sloeber is that sloeber assumes that the numbers starts at 1 and are consecutive.
Here the number starts from 11 so sloeber assumes nu hooks are provided

PR #7754, opened 2020-12-11
Fix: cannot build after #7060 on Win64

jjsuwa-sys3175:
* add double-quotes to `compiler.S.flags`

* fix windows-specific processes (`recipe.hooks.linking.prelink.[12].pattern.windows`)

* rewrite processing of "mkdir" and "cp" in python because of platform-independence

* make consistent with the use of quotation marks in other `*.py` files


d-a-v (approved):
Side note:
if we are ready to create external helpers,
we could use the same scripts written in python for all archs.

d-a-v (approved):

earlephilhower (approved):

dok-net:
Oops, I didn't notice this one... make my own PR #7761 fixing this same problem. Without all that Python overhead, though. I'm surprised why this PR changes so much more than I did.

PR #7761, opened 2020-12-13
Fix recent syntax error in platform.txt

dok-net:
Could not build recent master revisions on Windows, this is fixed by this PR.


earlephilhower:
Closing in preference to #7754 .  This PR would still have had an issue with spaces in paths (`"c:\Program Files (x86)..."`, etc.) and would still have different Windows/Linux+Mac paths.  Thanks anyway!

PR #7769, opened 2020-12-14
WIP - String Class Heap control

mhightower83:
Add support for String Class Heap Control by pre-setting a static class member.
Allows selecting a Heap for all String class buffer allocations.
Supports OOM Heap rollover, switch to the other heap on OOM.
Extended `irammem.ino` to show the use of String Class Heap control.

A few thoughts and questions remain:
* Because of complicated scenarios with Heap selection and roll over when the option was changeable, I switch to making it a preset static class member with weak attribute so that it can be overridden by a user module. 
* Is it safe to always get String buffer allocations from IRAM? Are all uses of the buffer accessed through `s.c_str` safe?
* lwIP, as we have it configured, will copy buffers for a TCP send. I have seen no issues in using IRAM this way with ESPAsyncTCP/ESPAsyncWebServer.
* Use with UDP is an unknown for me; however, I guess mDNS is working. Sorta, due to poor support by some devices and APPS, I find mDNS a little problematic anyway. 
* Which should be the default for the static class member `CurrentOnly` or `IramDram`?
* Heap selection API choice - which to use for implementation `ESP.set`... or direct `umm_malloc_`... ? Currently I have both implemented.

Update:
* **Side effect:** When compiled with Debug Level: "OOM", a debug message will be printed when OOM Rollover occurs.
# String Class - Heap Selection Control
Requirements: The `MMU` selection `16KB cache + 48KB IRAM and 2nd Heap (shared)` be selected from the Arduino IDE Tools menu.

Select one of these options to be used across all String class buffer allocations. These options are defined in `enum Heap`.
  * Allow OOM recovery (OOM rollover) by trying a different Heap on fail.
    * `IramDram` - Try to allocate from IRAM 1ST, then roll over to DRAM on fail.
    * `DramIram` - Try to allocate from DRAM 1ST, then roll over to IRAM on fail.
  * No OOM recovery. Use only the Heap specified.
    * `IramOnly` - Allocate only from the IRAM Heap.
    * `DramOnly` - Allocate only from the DRAM Heap.
  * Use the current umm_malloc Heap selection for new allocations.
    * `CurrentOnly` - Allocate only from the current Heap selected.
  * Use the current umm_malloc Heap selection for new allocations. On alloc failure, try the other Heap.
    * `Current` - Try to allocate from the current Heap selected, then roll over to the other Heap.

Initialize the String class static member `_preferredHeap` with the desired selection for Heap control in your sketch file somewhere before `setup()`.

Example of legacy behavior using option `CurrentOnly`:
`const String::Heap String::_preferredHeap = String::Heap::CurrentOnly;`

Example of maximizing available DRAM, by using IRAM 1st then DRAM:
`const String::Heap String::_preferredHeap = String::Heap::IramDram;`

When a Heap rollover occurs, the OOM counter for the failing Heap will increment. If both Heaps fail, both OOM counters are incremented.

Edited: for clarity, terminology, and reflect PR updates


mcspr:
I wonder if this is the case to be implemented through allocators?

As a small example: https://docs.microsoft.com/en-GB/cpp/standard-library/allocators?view=msvc-160
Plus, some more info on properties: https://en.cppreference.com/w/cpp/memory/allocator_traits

Effectively, it would be a property of the specific class, not global setting. I'll need a bit of time for the example, though

mhightower83:
@mcspr Oops, bad terminology on my part, by global I was referring to a static class member, that would hold one value used for all instantiations of the String class. Another desire was to keep the addition as small as possible for code and DRAM used by the class. The size of a static member would _not_ be multiplied across all the String class instantiations.

mcspr:
It is still a global though, only namespaced :) I was referring to the way it is done specifically for the String i.e., to create a way to allocate any objects via IramDram helper (any c++ stl container, [arduinojson also come to mind](https://arduinojson.org/v6/api/basicjsondocument/))

TD-er:
I can understand the advantages of having a fallback for allocations, but what may happen if a string has a longer lifetime and is allocated in the iRAM heap?
Does this affect execution of the rest of the application?
And what are the possible effects?
- Speed?
- Problems with ISR functions?
- Crashes?

What if a `String` object is created or does need re-allocation in a callback function (ISR call)?

mhightower83:
> what may happen if a string has a longer lifetime and is allocated in the iRAM heap?

Nothing, it just takes up space in the heap just as it would with DRAM. When configured for 48K IRAM / 16K ICACHE  and a 2nd Heap, the ratio cannot be changed until the next boot cycle. The IRAM set aside for the 2nd heap is always there.
> Does this affect execution of the rest of the application?
> * Speed?

Going will a 16K ICACHE most likely will result in added slowness. Using IRAM instead of DRAM is also slower. This option should not be taken unless you really need the additional data memory.
> * Problems with ISR functions?

`umm_malloc` will force `malloc` calls to use DRAM when called from an ISR.
> * Crashes?

None I can associate with doing this. While IRAM doesn't work everywhere, in the String class I have not seen any issues.
> What if a String object is created or does need re-allocation in a callback function (ISR call)?

If a String Object is created from an ISR, it gets DRAM
If a String Object is created from a system callback, the defined preference applies.
If a String Object needs re-allocation from an ISR context, it depends.
  * If space immediately above and/or below the allocation satisfies the new size requirement, it is resized where it is, regardless of the context of the call.
  * Otherwise, a new allocation will come from DRAM, and the string will be copied.

devyte:
> If a String Object needs re-allocation from an ISR context, it depends.

If memory serves, realloc is not interrupt safe, which means that any app where containers that use contiguous memory are used from an ISR could potentially crash. E. g  String, std::string, std::vector, etc.

PR #7773, opened 2020-12-17
Replacement for Boot ROM aes_unwrap

mhightower83:
This replacement is in and is required by the HWDT Stack Dump PR https://github.com/esp8266/Arduino/pull/7010#issuecomment-703170232.

This PR separates the fix out of the HWDT Stack Dump PR and makes the replacement `aes_unwrap` available all the time.
While I have not seen issues outside of HWDT PR, there is a concern that the AES buffer in the stack address space issue may cause a crash or data corruption in some other sketch. This may be viewed as an overabundance of caution. I cannot prove it is needed beyond my special case. However, everybody's sketch is different. It does ~add~use 304 bytes of IROM.

This description is taken from the `aes_unwrap.cpp` module:

Replacement for the ROM `aes_unwrap`() function. It uses the heap instead of the static DRAM address at `0x3FFFEA80`, which may step on the SYS stack in special circumstances such as HWDT Stack Dump.

When _not_ using WPS, the address space `0x3FFFE000` up to `0x40000000` is mostly available for the stacks. The one known exception is the ROM AES APIs. When `aes_decrypt_init` is called, it uses memory at `0x3FFFEA80` up to `0x3FFFEB30`
for a buffer. At the finish, `aes_decrypt_deinit` zeros out the buffer.

The NONOS SDK appears to have replacements for most of the ROM's AES APIs. However, the SDK still calls on the ROM's `aes_unwrap` function, which uses the ROM's AES APIs to operate. These calls can overwrite some of the stack
space. To resolve the problem, this module replaces `aes_unwrap`.

Final note, so far, I have not seen a problem when using the extra 4K heap option without the "debug HWDT". It is when combined with the HWDT Stack Dump that a problem shows. This combination adds a Boot ROM stack, which pushes up the SYS and CONT stacks into the AES Buffer space. Then the problem shows.

While debugging with painted stack space, during WiFi Connect, Reconnect, and about every hour, a block of memory `0x3FFFEA80` - `0x3FFFEB30` (176 bytes) was zeroed by the Boot ROM function `aes_decrypt_init`. All other painted
memory in the area was untouched after starting WiFi.


cores/esp8266/aes_unwrap.cpp
```diff
@@ -0,0 +1,165 @@
...
+    rijndaelKeySetupDec(rk, key);
+  }
+  return (void *)rk;
+}
+
+// This replaces the Boot ROM version just for this module
+static void aes_decrypt_deinit(void *ctx) {
+  if (ctx) {
+    ets_memset(ctx, 0, 16*11);
+    if ((uint32_t)ctx != 0x3FFFEA80ul) {
```
earlephilhower, line 60: If you're now taking over responsibility for the buffer, can this `if` ever be true?

cores/esp8266/aes_unwrap.cpp
```diff
@@ -0,0 +1,165 @@
...
+ *  pushes up the SYS and CONT stacks into the AES Buffer space. Then the
+ *  problem shows.
+ *
+ *  While debugging with painted stack space, during WiFi Connect, Reconnect,
+ *  and about every hour, a block of memory 0x3FFFEA80 - 0x3FFFEB30 (176 bytes)
+ *  was zeroed by the Boot ROM function aes_decrypt_init. All other painted
+ *  memory in the area was untouched after starting WiFi.
+ */
+
+#if defined(KEEP_ROM_AES_UNWRAP)
```
earlephilhower, line 28: What's the cost to always doing it this way?  If I understand properly, the ROM is using a fixed buffer located in where we have defined the stack (but we are normally not using this space), so the current setup is potentially an issue, no?

cores/esp8266/aes_unwrap.cpp
mhightower83: Paranoid programming on my part. 
It should be removed or may be changed to a DEV debug check.

cores/esp8266/aes_unwrap.cpp
mhightower83: > What's the cost to always doing it this way?

The cost is 304 bytes of IROM. and 176 bytes from the heap on WiFi (re)connect.
> current setup is potentially an issue, no?

Maybe

I think the extra 4K option puts us at a greater risk than the no 4K-extra.
For the no 4K-extra use case (assuming WPS), adding the define `KEEP_ROM_AES_UNWRAP` would give a little bit of code space back. Really there should not be any risk with no 4K-extra case, since WPS uses a lot of space in the area the SYS stack grows toward.

Even with the extra 4K option, the risk may be low. This is where I wish we could see the SDK code. My impression of the NONOS SDK most of its work is done via a task queue. When a new task starts the stack is near the all free mark. When a task runs the stack grows toward the `aes_buffer`. For the extra 4K option, the `cont` stack grew the SYS stack space usage and the HWDT grew the stack a little more. This left the stack at a start point with little reserve. And caused an overlap with the WiFI task stack usage and `aes_buffer`. 

Without the pressure of HWDT on the stack, there was a clearance all the time - that I saw. One concern would be, could (through bad timing) we in the midst of a WiFi event, take an interrupt and possibly a handled exception and get clobbered?

I think the risks are limited to the tasks that need the `aes_buffer`. At the moment WiFi (re)connect and something that happens every hour (I assume WiFi related) are the know users. I would think that at any other time, a SYS callback function could safely go all the way down to 0x3FFFE000 w/o issue.

I find myself neither arguing for nor against it. Just elevating the awareness of a possible issue.

cores/esp8266/aes_unwrap.cpp
mhightower83: Sorry I was a little confused. It is never needed. There should never be a problem with the buffer address being wrong in that way. Removed in the most recent push.

earlephilhower (approved):
Given the potential random/unrepeatable failures this can avoid, I think it s well worth 170 bytes of IROM and a little heap during WiFi reconnect.

d-a-v (approved):
Thanks for hacking the FW !

devyte (approved):

PR #7780, opened 2020-12-19
mmu_iram.h: Optimize non-32bit accessors

jjsuwa-sys3175:
* `mmu_get_*()`: reduced 1 insn (1 insn more if forthcoming GCC 10.3+ or patched GCC 10.2)

* `mmu_set_*()`: reduced 5 insns


earlephilhower:
@mhightower83 any thoughts on this?

cores/esp8266/mmu_iram.h
```diff
@@ -127,73 +127,66 @@ bool mmu_is_icache(const void *addr) {
 static inline __attribute__((always_inline))
 uint8_t mmu_get_uint8(const void *p8) {
   ASSERT_RANGE_TEST_READ(p8);
-  uint32_t val = (*(uint32_t *)((uintptr_t)p8 & ~0x3));
-  uint32_t pos = ((uintptr_t)p8 & 0x3) * 8;
-  val >>= pos;
-  return (uint8_t)val;
+  return *(uint32_t *)((uintptr_t)p8 & ~0x3) >> (uintptr_t)p8 * 8;
```
earlephilhower, line 130: I'm not so sure about this conversion.  Before it was `>>(p8 & 3) * 3`, now it's showing `>> p8 * 8`...i.e. the bitmask was dropped.

cores/esp8266/mmu_iram.h
```diff
@@ -127,73 +127,66 @@ bool mmu_is_icache(const void *addr) {
...
 }
 
 static inline __attribute__((always_inline))
 uint16_t mmu_get_uint16(const uint16_t *p16) {
   ASSERT_RANGE_TEST_READ(p16);
-  uint32_t val = (*(uint32_t *)((uintptr_t)p16 & ~0x3));
-  uint32_t pos = ((uintptr_t)p16 & 0x3) * 8;
-  val >>= pos;
-  return (uint16_t)val;
+  return *(uint32_t *)((uintptr_t)p16 & ~0x3) >> (uintptr_t)p16 * 8;
```
earlephilhower, line 136: Same as before, the bitmask has gone missing.

cores/esp8266/mmu_iram.h
```diff
@@ -127,73 +127,66 @@ bool mmu_is_icache(const void *addr) {
...
 }
 
 static inline __attribute__((always_inline))
 int16_t mmu_get_int16(const int16_t *p16) {
   ASSERT_RANGE_TEST_READ(p16);
-  uint32_t val = (*(uint32_t *)((uintptr_t)p16 & ~0x3));
-  uint32_t pos = ((uintptr_t)p16 & 0x3) * 8;
-  val >>= pos;
-  return (int16_t)val;
+  return *(uint32_t *)((uintptr_t)p16 & ~0x3) >> (uintptr_t)p16 * 8;
```
earlephilhower, line 142: Same

earlephilhower (changes requested):

cores/esp8266/mmu_iram.h
jjsuwa-sys3175: in GCC implementation and Xtensa ISA, rhs of shift operators is under modulo `(sizeof((lhs) + 0) * CHAR_BIT)`. and in ESP8266, `uintptr_t` and pointers are 32-bit wide,
in other words, 32-bit-lhs shift operator uses only the least significant 5 bits of rhs.

[before]
```
ptr  : 0b???????? ???????? ???????? ??????XX 
 & 3 : 0b00000000 00000000 00000000 00000011
       0b00000000 00000000 00000000 000000XX
 * 8 : 0b00000000 00000000 00000000 000XX000
                                     0bXX000
--------------------------------------------
```

[this PR]
```
ptr  : 0b???????? ???????? ???????? ??????xx
 * 8 : 0b???????? ???????? ???????? ???xx000
--------------------------------------------
                                     0bXX000
```

were your doubts cleared?

mhightower83: With what I had originally, I was going for readability and ease to follow, and a little awareness of how the compiler would handle things. 

Looking closer at the changes, I just noticed the following:

This `>> (uintptr_t)p8 * 8;` doesn't look right. You want to only multiply the lower 2-bits by 8. The full multiply will exceed 32. I am not sure this would compile into something that works.  

with `p8 & ~0x3` you lose two bits of your value on shift 0.

A simple cast on the return value leaves it up to the compiler how to best reduce, bits returned, which I think compiled to an `extui`. The simple cast gives clarity and communicates the programmer's intent for the return value.

cores/esp8266/mmu_iram.h
mhightower83: Same as above.

cores/esp8266/mmu_iram.h
mhightower83: Save as above.

cores/esp8266/mmu_iram.h
jjsuwa-sys3175: indeed, backing to `>>(p8 & 3) * 8` brings 1 more insn: `extui a2, a2, 0, 2`.

another solution: `>>(p8 * 8 & 31)` or `>>((p8 << 3) & 31)`
this has no such drawback.

mhightower83: For `>>(p8 * 8 & 31)`  the operator `&` will occur before `*`, I would expect the compiler to optimize the `8 & 31` to `8`.

I would expect `>>((p8 << 3) & 31)` to bring back the extra instruction.

mhightower83: Oops sorry I got my precedences backward.

jjsuwa-sys3175: of course, ordinary C/C++ programmers know multiplication has higher priority than bitwise-and :)

> The simple cast gives clarity and communicates the programmer's intent for the return value.

and, it's obvious what are returned from function; see prototype. abuse of explicit cast brings another kind of problem i guess.

cores/esp8266/mmu_iram.h
```diff
@@ -127,73 +127,66 @@ bool mmu_is_icache(const void *addr) {
...
-  ival &= (~valmask);
-  ival |= sval;
-  *p32 = ival;
+  uint32_t valmask, sval;
+
+  __asm__ ("ssa8b\t%2\n\t"
+           "src\t%0, %3, %3\n\t"
+           "src\t%1, %4, %4"
+           : "=&a"(valmask), "=a"(sval)
+           : "r"(p8), "r"(~0xFF), "r"(val), "r"(ival));
```
mhightower83, line 156: I don't see `ival` being used here, it would be `%5`.

jjsuwa-sys3175: > I don't see `ival` being used here, it would be `%5`.

`"r"(ival)` is used only for injecting dependency, in order to ensure `ival = *p32;` precedes the asm block.
without this, it may be placed between the asm block and `*p32 = (ival & valmask) | sval;`, thus will cause pipeline interlock stall of 1 cycle. (see Xtensa ISA refman. p608)

jjsuwa-sys3175: sorry for quite low-level machine-dependent coding ;)
added comments for the explanation.

jjsuwa-sys3175:
> reduced 1 insn (1 insn more if forthcoming GCC 10.3+ or patched GCC 10.2)

actually examined:
```
char test(void* p) {
    return mmu_get_uint8(p);
}
```

before 9de8373f1bbe885e922a36af1c1326f0e71f94f2:
```
	.file	"test.c"
	.text
	.literal_position
	.align	4
	.global	test
	.type	test, @function
test:
	movi.n	a3, -4
	and	a3, a2, a3
	l32i.n	a3, a3, 0
	slli	a2, a2, 3	; a2 = a3 >> a2 * 8;
	ssr	a2		;
	srl	a2, a3		;
	extui	a2, a2, 0, 8
	ret.n
	.size	test, .-test
	.ident	"GCC: (GNU) 10.2.0"
```

after 9de8373f1bbe885e922a36af1c1326f0e71f94f2:
```
	.file	"test.c"
	.text
	.literal_position
	.align	4
	.global	test
	.type	test, @function
test:
	movi.n	a4, -4
	and	a4, a2, a4
	l32i.n	a3, a4, 0
	ssa8l	a2		; a2 = a3 >> a2 * 8;
	srl	a2, a3		; 
	extui	a2, a2, 0, 8
	ret.n
	.size	test, .-test
	.ident	"GCC: (GNU) 10.2.0"
```

cores/esp8266/mmu_iram.h
mhightower83: I wonder if this is necessary. The compiler already appears to be pipeline aware, and often preloads values before they are needed. And, since your Extended ASM block does not use the "memory" clobber parameter. The compiler would have no need to reload the register after the ASM block.

jjsuwa-sys3175: good news: after 3498137c37ba06baf06c9823b1755701b4b803d4 (or 51bcd52bc8f99b310142cd3ba65d7a6934b472c9), such a stall seem not to be seen if without `"r"(ival)`:
```
void test(void* p, char c) {
    mmu_set_uint8(p, c);
}
```
```
	.file	"test.c"
	.text
	.literal_position
	.align	4
	.global	test
	.type	test, @function
test:
	movi.n	a4, -4
	and	a4, a2, a4
	l32i.n	a5, a4, 0
	movi	a6, -0x100
#APP
# 9 "test.c" 1
	ssa8b	a2
	src	a7, a6, a6
	src	a2, a3, a3
# 0 "" 2
#NO_APP
	and	a5, a5, a7
	or	a2, a2, a5
	s32i.n	a2, a4, 0
	ret.n
	.size	test, .-test
	.ident	"GCC: (GNU) 10.2.0"
```
however, `mmu_set_*()` are inline functions; their function bodies will be meld into another, thus there may be exist other contexts of processing that interfere with the code arrangement.
clobbering "memory" also works well but overdoing; that is non-local read/write memory barrier, so presses the other contexts of processing for non-urgent store-to-memory operations, and puts obstacles in the way of the compiler's optimizer.

in conclusion, i am sure that it is appropriate the use of unused input operand `"r"(ival)` in asm block, because it operates as read-only memory barrier on only a specific target.

mhightower83: LGTM

mhightower83:
@jjsuwa-sys3175 Something isn't right. Please try the example at `libraries/esp8266/irammem/irammem.ino`. The comparison between IRAM through exception handler, IRAM Inline, and DRAM performance have sum values for the reads. The read sum results for the updated inline functions are different from those reported for IRAM with exception handler and DRAM.

mhightower83:
@jjsuwa-sys3175 Looks like it is an unsigned 16-bit promotion to a signed 32-bit conversion problem, adding a cast uint32_t to `val` makes it work for both the 16 and 8 bit cases.

```cpp
           : "r"(p8), "r"(~0xFF), "r"((uint32_t)val), "r"(ival));
```

earlephilhower:
@mhightower83 In my my original review I was able to spot problems just visually, but I think now you're finding subtle issues here.  Do you have any specific tests we could add to device tests (and run and verify) to check the behavior?  If there are only 6-7 different conditions we need to test, it might be very simple to do.

Of course, only a device test would catch them, but running them is actually pretty straightforward now so we can run them ourselves.

Maybe something as simple as a test of byte reads at off=0,1,2,3; 16-bits are 0,2, and 32-bits at 0,4?  Might need to do both a signed/unsigned test, too, but not that large a set?

Or am I missing something more subtle we need to check for?

mhightower83:
@earlephilhower so far I have been just updating the performance part of `irammem.ino` to exercise the non32bit handling logic. I am looking at updating it again to do the DRAM test 1st then verify those results with the IRAM exception and inline. And, make a failure a little more obvious. Maybe also add  `int16` testing to the list.

Something smaller as you describe I think would work. I am not sure you would need to do 32-bits at 0,4?

earlephilhower (changes requested):
Please run and post a successful run of the test in #7812 which will give us better confidence the changes here don't affect functionality.

jjsuwa-sys3175:
closed due to lack of interest and being outdated.

PR #7790, opened 2020-12-23
HeapSelectDram for pvPortMalloc, ...

mhightower83:
The goal here is to guarantee that the SDK gets DRAM for the heap requests that are made during SDK API calls. To accomplish this I rely on the SDK to only use the portable heap functions: pvPortMalloc, pvPortCalloc, pvPortRealloc, and pvPortZalloc. I add a thin wrapper to these functions to switch to DRAM. 

I am not sure about the _lwIP_ library, it appears to use both malloc and the pvPort... APIs.


d-a-v:
lwIP can be updated to only use pvPortMalloc.
The pure lwIP malloc macro is in `cc.h` and the glue should be changed to use this macro. 

They also could be changed for lwIPMalloc.

Would you like me to make the change ?

mhightower83:
@d-a-v, Yes, that would be great!

d-a-v:
@mhightower83 You may try with #7793

mhightower83:
@d-a-v Thanks! 

I tested with `liblwip2-1460-feat.a` and `liblwip2-1460.a`. Both worked with this _bad_ example, using a HeapSelectIram across all of `setup()`. `WiFi.softAPConfig(apIP, apIP, IPAddress(255, 255, 255, 0))` was the call at this level that would lead to a HWDT.



```cpp
#include <ESP8266WiFi.h>
#include <DNSServer.h>
#include <ESP8266WebServer.h>
#include <umm_malloc/umm_heap_select.h>

const byte DNS_PORT = 53;
IPAddress apIP(172, 217, 28, 1);
DNSServer dnsServer;
ESP8266WebServer webServer(80);

String responseHTML = ""
                      "<!DOCTYPE html><html lang='en'><head>"
                      "<meta name='viewport' content='width=device-width'>"
                      "<title>CaptivePortal</title></head><body>"
                      "<h1>Hello World!</h1><p>This is a captive portal example."
                      " All requests will be redirected here.</p></body></html>";

void setup() {
  HeapSelectIram x;

  WiFi.persistent(false);
  WiFi.mode(WIFI_OFF);
  Serial.begin(115200);
  delay(15);
  Serial.println();
  Serial.println();

  WiFi.mode(WIFI_AP);
  // { // This is the WiFi call that leads to HWDT
  //   HeapSelectIram x;
  // }
  WiFi.softAP("DNSServer CaptivePortal example");

  // if DNSServer is started with "*" for domain name, it will reply with
  // provided IP to all DNS request
  dnsServer.start(DNS_PORT, "*", apIP);

  Serial.println("DNSServer CaptivePortal example running");

  // replay to all requests with same HTML
  webServer.onNotFound([]() {
    webServer.send(200, "text/html", responseHTML);
  });
  webServer.begin();
}

void loop() {
  HeapSelectIram x;
  dnsServer.processNextRequest();
  webServer.handleClient();
}
```

earlephilhower:
@d-a-v, is this good to merge?

d-a-v (approved):
Thanks !

PR #7791, opened 2020-12-23
Use smarter cache key, cache Arduino IDE

earlephilhower:
Thanks to @mcspr.  Make sure the cached toolchain is only valid when the
package.json file is identical.  Also, add the Arduino IDE to the cache
to avoid re-downloads.


mcspr:
Thanks! I missed the reply in the other pr

PR #7799, opened 2020-12-29
UART write loop could lead to wdt with low bitrates or large buffers

### LINKED ISSUE #7746: Slow serial baudrates cause exceptions / WDT
### Basic Infos

#### Platform

- Hardware:			[ESP-12]
- Core Version:			[framework-arduinoespressif8266 2.7.4]

### Settings in IDE

- Module:  			[WeMos D1 R2 and mini]
- Flash Mode:			[dio]
- Flash Size: 		        [4MB]
- lwip Variant:                    [v1.4|v2 Lower Memory|Higher Bandwidth]
- Reset Method:		[clk]
- Flash Frequency:	        [40Mhz]
- CPU Frequency:	        [80Mhz]
- Upload Using:		        [SERIAL]
- Upload Speed:                [115200] (serial upload only)
- PlatformIO - ini file:

```cpp

[env:d1_mini]
platform = espressif8266
board = d1_mini
framework = arduino
board_build.filesystem = littlefs

; COM1 or COM3
upload_port = COM[4]
upload_speed = 115200

monitor_port = COM[4]
monitor_speed = 115200
monitor_flags=
    --parity
    N

```

### Problem Description

I want to use LittleFS to save some data coming from serial port at baudrate 110, N81
I tested the provided example in:
https://github.com/esp8266/Arduino/tree/master/libraries/LittleFS/examples/LittleFS_Timestamp
 and it works fine at baudrate 115200
but when changing baudrate to 110, the code seems to retart (exception) 

I need to have a running FS at baudrate of 110, because I can not change the baudrate of the device I am connected to.

### Sketch

```cpp

https://github.com/esp8266/Arduino/blob/master/libraries/LittleFS/examples/LittleFS_Timestamp/LittleFS_Timestamp.ino

```

### Debug Messages

@ Baudrate 110
FS with 110 baudrate

@Baudrate 115200
FS with 115200 baudrate

dok-net:
EspSoftwareSerial successfully uses optimistic_yield for the same purpose already.
Fixes #7746

### DIFF
--- a/cores/esp8266/uart.cpp
+++ b/cores/esp8266/uart.cpp
@@ -506,8 +506,10 @@ uart_write(uart_t* uart, const char* buf, size_t size)
 
     size_t ret = size;
     const int uart_nr = uart->uart_nr;
-    while (size--)
+    while (size--) {
         uart_do_write_char(uart_nr, pgm_read_byte(buf++));
+        optimistic_yield(10000UL);
+    }
 
     return ret;
 }


d-a-v (approved):
LGTM !

cores/esp8266/uart.cpp
```diff
@@ -506,8 +506,10 @@ uart_write(uart_t* uart, const char* buf, size_t size)
 
     size_t ret = size;
     const int uart_nr = uart->uart_nr;
-    while (size--)
+    while (size--) {
         uart_do_write_char(uart_nr, pgm_read_byte(buf++));
+        optimistic_yield(10000UL);
```
earlephilhower, line 511: AIUI, `optimistic_yield` will `panic()` if not in the CONT context.  This would make any SYS-callback printouts (TCP, others) potentially crash.

@devyte, @d-a-v, any comments/ideas?  I like the idea, but the forced panic seems worse than a potential WDT...

earlephilhower (changes requested):
Just worried about yield() in the UART code.  Need to be very sure it's safe and not used by SYS to log/etc...

cores/esp8266/uart.cpp
devyte: optimistic_yield() checks if it can yield (i. e. in CONT), and if enough time has elapsed since the start of the current loop. If true, it yield()s, otherwise does nothing. 

So it shouldn't panic. 

However: I seem to remember there was a change to its behavior so that yield actually only happens once every blah ms instead of after blah ms since loop start.

In any case:
- At a glance, I think the argument of 10000 is way too big, I would use something like 500, i. e. yield at most every 0.5s
- this yield won't help in the case of slow comms attempted from SYS, but you could argue that comms should be done from CONT

devyte (changes requested):

cores/esp8266/uart.cpp
d-a-v: The unit is `µs` so 10k is 10ms.
It should be fine.
Maybe 1ms is better to sustain high throughput in case of a network transfer.

devyte (approved):
Oops, when I took a look before I saw ms. You're right, it's us. Maybe I need glasses...
One yield every 10ms should be ok. It might be worth testing if there's a difference vs 1ms before merging, but I don't see that as a stopper. 

Approving.

earlephilhower (approved):
Fair enough, thanks for the clarification.  LGTM!

PR #7807, opened 2021-01-03
Reduce BearSSL code size by not -O2 FSMs

earlephilhower:
The FSMs that drive the encryption logic are not performance critical,
but they are very large.  Build BearSSL using -Os on them, leaving -O2
everywhere else (i.e. encryption) for speed.

Saves 1-2KB of flash for SSL applications.


devyte:
> The FSMs that drive the encryption logic are not performance critical

What is the performance impact of this change? 
Is the connection process impacted? 
Is the encryption itself during communication impacted? 
All of the above? none? 
Is the impact negligible? ~1% slower? ~10%?

earlephilhower:
I wasn't able to measure a consistent performance difference (there is always some wiggle, this change is below that threshold) using the validation example's time-tests.

Encryption/BW is unimpacted as well, as far as I can tell (again, always some wiggle).  Only the FSM (think of it as the control plane) had `-Os` used.  Everything else is still at `-O2` (AES, Elliptic Curve, SHA, etc.)

The FSM basically was inlining a copy-small-no-of-bytes-from-a-to-b multiple times.  Saves a few register stores and call/ret, but it's only done 1-2 times per an entire connection.

The FSM itself doesn't have much option for logic optimization, it's a machine-generated, Forth-like interpreter with a byte-byte processor so each `case` is pretty minimal, and the case transitions are driven off of a big table, so GCC really can't look into things.

So, AFAICT, it's ~0 perf impact and 1-2K (depends on your calls) size savings.

devyte (approved):
Approved!

d-a-v (approved):

PR #7811, opened 2021-01-05
Protect against server hijacking error handling

paulocsanz:
If a server returns "HTTP/1.x -8 OK", for example, it can misguide an application developer into freeing less-important memory so the request can be retried and succeed, when the problem is in the server.

_returnCode is never used anywhere else, but it could still contain a negative value returned by a broken server and therefore could cause troubles in the future (if _returnCode is in fact used)

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -1125,18 +1125,18 @@ int HTTPClient::handleHeaderResponse()
                     if(transferEncoding.equalsIgnoreCase(F("chunked"))) {
                         _transferEncoding = HTTPC_TE_CHUNKED;
                     } else {
-                        return (_returnCode = HTTPC_ERROR_ENCODING);
+                        _returnCode = HTTPC_ERROR_ENCODING;
+                        return _returnCode;
                     }
                 } else {
                     _transferEncoding = HTTPC_TE_IDENTITY;
                 }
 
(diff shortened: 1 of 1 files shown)


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -1125,17 +1125,17 @@ int HTTPClient::handleHeaderResponse()
                     if(transferEncoding.equalsIgnoreCase(F("chunked"))) {
                         _transferEncoding = HTTPC_TE_CHUNKED;
                     } else {
-                        return HTTPC_ERROR_ENCODING;
+                        return (_returnCode = HTTPC_ERROR_ENCODING);
```
earlephilhower, line 1128: Style request:  Please separate assignment and return here and below, just for sanity's sake.

earlephilhower (changes requested):
Have you actually seen HTTP servers so far out of spec that they really return negative error codes? 

Minor style change, please.  OTW seems fine.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
paulocsanz: Done!

paulocsanz:
No, I've not. It's probably not a big deal. It was just something I caught in a code "audit".

earlephilhower (approved):
Thanks!  LGTM.

PR #7812, opened 2021-01-06
Add verify IRAM sum results using DRAM as reference.

mhightower83:
Make verify fail more visible.

Updated example irammem.info to show performance and test int16 handling when IRAM is used.

Removed 4K performance test. It didn't show any new information over the 1K test.


earlephilhower (approved):
Looks like a good addition!  We should probably steal bits to put in `tests/host` at some point, but for now we can ask anything related to the exception handler change to run and verify this still passes.  Thx!

PR #7815, opened 2021-01-07
EspSoftwareSerial 6.11.2

dok-net:
Delegate<> assignment operators added.
Addresses compilation errors with pending PRs due to commit 95fb104562


d-a-v:
How is the cited commit related to EspSoftwareSerial ?

dok-net:
It's not immediate related. But as it currently stands, Delegate and MultiDelegate are distributed by copy among a handful of my projects, so everytime I have to make an update in one of them, I check all the others for regressions, and if satisfied that nothing breaks, I update them all.

PR #7816, opened 2021-01-08
Improvements to non 32-bit exception handler

mhightower83:
Extended ASM got fussy when using different optimizations. ie. HWDT resets. It seems you should not use input registers for scratch registers. Fixed by adding an extra output register for temporary values. No code size increase. Now works with different optimizations.

Light refactoring for readability

Added "C" reference code for Extended ASM

Save two cycles by loading `a0` early in `exc-c-wrapper-handler.S`

Use optimization `O2`
Net change in size, 0 bytes with optimization. Save 4 bytes w/o Optimization.
With changes and `O2` save 3 cycles on write and 6 cycles on read.


earlephilhower (approved):
LGTM!  I think you found and fixed an occasional issue I was having w/the PSRAM virtual mem version of this, too.  Good debugging!

PR #7822, opened 2021-01-12
Avoid UB and abort on nullptr buffer

### LINKED ISSUE #7821: Possible UB on OOM
### Basic Infos

### Problem Description

At [libraries/Netdump.cpp](https://github.com/esp8266/Arduino/blob/7f38e141c7502032af594fdd502c1b1f6049377c/libraries/Netdump/src/Netdump.cpp#L169) `packetBuffer` is dereferenced without checking for `nullptr`.

`packetBuffer` is filled at [Netdump::tcpDump](https://github.com/esp8266/Arduino/blob/7f38e141c7502032af594fdd502c1b1f6049377c/libraries/Netdump/src/Netdump.cpp#L92), but it's with a `new (std::nothrow) ...` and `Netdump::tcpDump` is called anyway.

That means if `new (std::nothrow) ...` returns a null pointer because it's out of memory UB will happen.

I'm prepared to provide a PR to fix it, but I don't know the best way to do this. Should I just check at `Netdump::tcpDumpProcess` if `packetBuffer` is filled and return early? Is there a way to warn the user? And I'm not really sure what is the appropriate cleanup code, or if a `return;` is sufficient.

paulocsanz:
Fixes #7821.

This does not report the OOM to the user. But it already prevents UB from [Netdump::tcpDumpProcess](https://github.com/esp8266/Arduino/blob/7f38e141c7502032af594fdd502c1b1f6049377c/libraries/Netdump/src/Netdump.cpp#L169).

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClient.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClient.cpp
@@ -207,7 +207,7 @@ size_t WiFiClient::write(uint8_t b)
 
 size_t WiFiClient::write(const uint8_t *buf, size_t size)
 {
-    if (!buf || !_client || !size)
+    if (!_client || !size)
     {
         return 0;
     }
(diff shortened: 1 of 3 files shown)


paulocsanz:
I think I've found a better way. Just need to squash the commits before merge.

Is this ok to merge?

earlephilhower (approved):
LGTM, thx.

d-a-v (approved):

Issue #7826, opened 2021-01-17
Unable to override CertStore


paulocsanz:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [] I have filled out all fields below.

### Problem Description

CertStore is a cornerstone of properly using SSL in esp8266/Arduino, but it's heavily constrained, it depends on having a single archive produced by `ar` and generated in a very specific way (by a loose script in an example folder), and having a File to provide this archive to it. Meaning it's probably stored in the flash FS.

Theoretically making a overriden FS works, but it's a lot of complexity, and basically a reverse-engineering of the currently used algorithm.

That means in practice a lot of people adopt hacks like overriding the class by using the same header guard. Which is horrifying, but something we actually do in production. This link provides an alternative that is widly used. https://github.com/maakbaas/esp8266-iot-framework/blob/master/src/certStore.h

We can simply solve that need by making CertStore::installCertStore virtual, because it's the only function called by WiFiClientSecure. It's a fairly simple function but one that chooses the static callbacks to find and free a TA.

I don't think providing a built-in alternative way right now is the way to go right now. We could use this implementation https://github.com/internet-of-plants/embedded/blob/master/src/certificate_storage.cpp

But it doesn't seem to make sense to upstream and maintain another python script, to download the certificates and bundle them in the source code, and an implementation to provide certs from it, right now. Maybe it will solve a lot of people's problems, but for now I think not commiting to new behaviors is the way to go and just allowing the users to make their own.

So just making the installCertStore virtual seems enough.

Btw, is a PR to replace X509List raw pointer to a unique_ptr acceptable? Like I did in the link above.

PR #7827, opened 2021-01-17
Allow users to inherit and override CertStore

paulocsanz:
Fixes: #7826

This seems the minimal change that is enough to avoid guard headers hacks and allow users to have a more customized CertStore.


paulocsanz:
I don't get it, what happened to the CI? Did I break something? Was this just a timeout? How to re-run?

paulocsanz:
I've been thinking if it's not better to make a CertStoreBase with a virtual `installCertStore`, make CertStore inherit from and basically not modify it. And where CertStore is passed we can pass a CertStoreBase.

What are the thoughts of the community?

I can modify my PR accordingly.

earlephilhower:
> I've been thinking if it's not better to make a CertStoreBase with a virtual `installCertStore`, make CertStore inherit from and basically not modify it. And where CertStore is passed we can pass a CertStoreBase.

That seems like a reasonable thing to do and would avoid dragging unneeded class variables into subclasses.  But, I'm sure you are trying to get some other way to hold the certstore (i.e. a PROGMEM array), so if you can get that in it would really be nice.

paulocsanz:
Perfect, I've implemented it.

Pardon my lack of c++ knowledge, but should CertStoreBase have a virtual destructor? It was erroring when I put it, removing it fixed. But it seems that without it if someone with a CertStoreBase pointer tries to destroy it, it will cause problems. Although the code never does that, since it only borrows the CertStoreBase.

But Idk about c++ best practices surrounding virtual destructors.

I know CertStore should be long-lived and probably never destroyed, but just to make sure it won't cause problems one day.

earlephilhower (approved):
Looks good.  Did you want to add in your other CertStore object?

paulocsanz:
As I said in the original issue I don't think it's the best move stabilizing another CertStore alternative, with its own python script. I can provide the alternative if it's deemed that since so many people depend on the hacks and save certs to PROGMEM in source code.

But I'm not realy confortable doing it right now, but I can indeed provide it in a future PR.

maakbaas:
@paulocsanz: Thanks for adding this! It seems that this change was at least partly inspired by my hack, which indeed came about due to the lack of virtual methods.  I have updated my code with the new inheritance mechanism in this commit: https://github.com/maakbaas/esp8266-iot-framework/commit/8a752f8443a8bb770ce8165442981d1362a64422

PR #7828, opened 2021-01-17
Add a few nullptr checks

paulocsanz:
Just a few nullptr checks I found needing while perusing the library.

Technically WiFiClientSecureCtx::read doesn't need them because it calls WiFiClientSecureCtx::available before dereferencing _recvapp_buf, and available sets the buffer with a bearssl function (I'm assuming it can't ever return null). But it didn't seem to hurt.

But in WiFiClientSecureCtx::peekBytes this check is not done, so it seems you can have connected but not called available and then _recvapp_buf will be passed to memcpy. So I added a few checks to ensure a bad usage can't cause a crash (and logs what happened).

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
@@ -406,11 +406,6 @@ int WiFiClientSecureCtx::peek() {
 
 size_t WiFiClientSecureCtx::peekBytes(uint8_t *buffer, size_t length) {
   size_t to_copy = 0;
-  if (!buffer) {
-    DEBUG_BSSL("peekBytes: buffer is nullptr\n");
-    return 0;
-  }
-
   if (!ctx_present()) {
     DEBUG_BSSL("peekBytes: Not connected\n");
     return 0;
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
Thanks, but I think this PR might have missed an underlying important concepts in BSSL.

The returned pointer `_recvapp_buf` is *never* malloc'd.  It's a pointer to inside the BSSL preallocated decryption buffers.  So these checks cannot possibly ever catch anything.  The returned count will be 0 when there is nothing decrypted, which the logic now handles fine, or some value > 0 which will guarantee BSSL will retuen an index to the SSL context (allocated at object instantiation).

The whole lack of allocations in BearSSL is why it was chosen in the first place.  We used to have axTLS and had significant memory allocation issues causing tons of hard-to-repeat crashes (related to memory fragmentation, etc.).  Now, as long as the object is created and connect() succeeds, there is no memory allocation needed/used for the SSL part of the connection ever again.  So you fail early (with proper return code) or you work.  Order of magnitude fewer issues with this.

The BearSSL library author @pornin made an amazing library and some choices that work very well in the embedded world!

paulocsanz:
Of course, the `read` change can be removed, but I think the peekBytes have no check, because you can connect and peekBytes without calling available in between and it's going to crash (I may be missing something).

But I do get that this PR may not be deemed as worth merging, since it's not exactly a problem. Feel free to close it.

earlephilhower:
The `peekBytes()` method uses `available()` which also takes care of setting the recv_len to 0 on failure or to whatever really is available per the BSSL library, so the remaining check also seems not to be able to catch anything, either.  I'll close for now, but if you can find a code snippet where it does crash, I'd definitely be happy to reevaluate it.!  Thx.

PR #7832, opened 2021-01-19
Proposal: internal functions: renamings

d-a-v:
Rename internal functions:
- `esp_yield()` => ` esp_suspend_from_cont()`
  (gives control back to the underlying scheduler)
- `esp_schedule()` => `esp_request_for_cont()`
  (asks the underlying scheduler to wake our "main task" (ie: call `loop()`, not now but ASAP)
- `minimal_yield()` is `optimistic_yield()` with an internally fixed value  (1000us)
  (calling `esp_request_for_cont()`+`esp_suspend_from_cont()` with a minimum of 1ms after previous call)
- ~~Deprecate `optimistic_yield()`~~
- Ignore `optimistic_yield()`'s parameter, use the 1ms internal one (same as `minimal_yield()`)

Also:
- Don't `panic()` when `yield()` is called from SYS,
  instead emit a warning on console in debug mode

Reasons:
- Some other PRs want to add wrappers to make these functions more explicit
  I hope these names are explicit enough so they don't need to be wrapped and complex to read
- `optimistic_yield(usec)` is used with funny values all through the core
  - Make the value fixed (1000)
    choice of 1000ms: commented in the source code (1460 bytes packet inter-arrival for a 10Mbits/s BW)
  - We can always update this value in only one place without letting the user chose a random one.
- ~~I believe, without needing to be optimistic, that user facing `yield()` can just be always optimistic~~(removed, too frightening)
  

~~What would be a good reason not to rename `optimistic_yield(1000)` to `yield()` ?~~

~~I am interested if frameworks like Tasmota/EspEasy/Espurna/AnyOther could test this.~~
I ran some examples, including bearssl validation, without issue (<= that was with `yield()` internally replaced by `minimal_yield()`


TD-er:
> I am interested if frameworks like Tasmota/EspEasy/Espurna/AnyOther could test this.

In ESPEasy the `optimistic_yield` is not used, so I don't see any reason why this should fail.

TD-er:
> @TD-er `yield()` has changed in this PR and is equivalent to current `optimistic_yield(1000)`.

Ah OK, I got the impression it was the other way around.
I will later try to find a way to compile a version based on this PR....

mcspr:
Similar here, but there's only a single place that optimistic yield is mentioned - debugging function that sometimes sends data to the network, and the only reason is that it avoids panic.

No specific preference for removal / keeping it, but perhaps methods should just change the prefix from `esp_` to `cont_`?

dok-net:
After needing some time to wrap my head around the existing namings; finding out which ones to shim, IIRC also PRing additional wear refs that got merged in the recent past; and building at least the CoopTask stackful coroutine multitasking lib on top of that work, plus devising the new `esp_beak()` in a separate, still pending PR; I am not particularly enthusiastic about this round of renamings and altering the contracts. I don't know that changing the semantics of `yield()` like in this PR is still Arduino AVR compatible enough. But I am rather seriously concerned that it breaks the multi-tasking hook semantics for AVR.
Besides, does this make the API even more dissimilar from the Arduino Core for ESP32, raising addional low-level portability issues?

mcspr:
Ah, true, we do lose strong guarantee that yield() actually does something, which might be surprising...

Afaik, esp32 simply noops optimistic_yield and yield is true thread yield
https://github.com/espressif/arduino-esp32/blob/af119215353b26d944b80d7126bed68e060e6697/cores/esp32/esp32-hal-misc.c#L49
https://github.com/espressif/arduino-esp32/blob/af119215353b26d944b80d7126bed68e060e6697/cores/esp32/esp32-hal.h#L43-L45

d-a-v:
> Afaik, esp32 simply noops optimistic_yield and yield is true thread yield

On esp8266 nonos sdk, there is no real thread yield.
Current `yield()` makes internal checks about WiFi, feeds the wd.
Where relevant or when needed in the core, real functions are called (`esp_yield()` renamed to  `esp_suspend_from_cont()`).
`yield()` is the user facing API.

mcspr:
Should the internals *always* stick to esp_... then? There's still yield() in some places.

Note that my quote is about esp32 ::
esp8266 implementation behaves sort-of like threads though, even if they don't really execute at the same time. and we still have multi-threading problems...

d-a-v:
> Should the internals always stick to esp_... then? There's still yield() in some places.

Those `yield()` were or should have been former `optimistic_yield()`.
Relevant places are changed if I'm not mistaken.

> and we still have concurrency problems...

Do we ? (apart from ISRs)

dok-net:
I have a feeling that instead of essentially dropping all the uses of proper `optimistic_yield()`, based on the very low intervals in some calls, that those low intervals < 10000µs might be due to a misunderstanding of the workings of `optimistic_yield()`. 
In `libraries/ESP8266WiFi/src/WiFiClient.cpp`, `libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp`, and `libraries/ESP8266WiFi/src/WiFiUdp.cpp`, the 100µs interval is quite likely nothing less than a bug and I tend to PR an immediate fix to increase these to a reasonable value, which I consider 10000µs flat.

There's a few places using 1000µs, which would probably be fine with 10000µs a well, and then there is `WAIT_FOR_PIN_STATE` in `core_esp8266_wiring_pulse.cpp`, which uses 5000µs, which I think will be fine/better off with 10000µs as well.

Rationale: low intervals for `optimistic_yield` provide more headroom for following code to not yield for a longer time, if the `optimistic_yield` is in the code path unconditionally. Generally `optimistic_yield` is great for `available()`-like functions, that may called in loops from sketches, in order to alleviate the user from having to keep an eye on total time spent in those loops to prevent WDTs. But 10000µs would be good enough for that, right?

Why is 1000µs considered a better default interval in this PR instead of  the previous 10000µs, used in most places in the existing code?

d-a-v:
The relation between raw packet inter-arrival time and bandwidth with a given packet size is given by:

```
iaμs = 1000000 / (bwMiBitsps / 8 * 1024 * 1024 / packet-size-bytes)
bwMiBitsps = 1000000 / iaμs * 8 / 1024 / 1024 * packet-size-byte
```

We get 1114 μs, 1460 bytes, 10Mibits/s,
409 μs, 536 bytes, 10Mibits/s,
and 10'000 μs, 1460 bytes, 1.1 Mibits/s.

The best bandwidth I could ever see was never more than 10Mbits/s (7Mbits/s yesterday with a raw test).

That's the reason why 1ms - 1000 μs is a good compromise IMHO.

d-a-v:
- `yield()` is now back to unchanged, 
- This PR's title remain : internal function renaming,
- `optimistic_yield()` ignored parameter,
- no deprecation.

dok-net:
A general question, how would one go about distinguishing versions of the Arduino pre and post this patch? I would like [CoopTask](https://www.arduinolibraries.info/libraries/coop-task) to auto-detect this as much as possible, version conflicts that don't compile are a little hard on the intended user-base of the library.

d-a-v:
There's `core_esp8266_version.h` which defines `coreVersionNumeric()`. Its value changes with every commits and is numerically comparable.
However I agree that a specific C-style define like `#define ESP8266COREAPI_YIELD 2` could also be useful.
Would it suit with your constraints (or with another name) ?

dok-net:
As I believe the inspiration for this PR came in part from #7148, which also was requested to be kept "online" be @d-a-v after initial reserverations from @devyte, I have now updated PR #7148.

I believe that #7148
- has more comprehensive and consistent renamings
- has fewer renamings where I personally see no previous issues
- fixes the 'delay(0)` issue
-  maintains `esp_yield()` semantics of suspend and re-schedule
- solves compatibility issues with 3rd party MT libs by introducing the define `HAVE_ESP_SUSPEND` in `coredecls.h`

No yet there, and IMHO better suited to a separate follow-up PR:
- A `minimal_yield()`, though I disagree with 1000µs suiting all cases, in particular serial UART `available()` used to use 10000µs and that seem a better fit. Therefore, `optimistic_yield()` shall be maintained and the argument must not be disregarded.

Also, I don't know that
> Don't panic() when yield() is called from SYS,
instead emit a warning on console in debug mode

is such a great idea, by inviting new invalid calls to `yield()` from SYS, which can so simply be replaced by the proper `esp_yield()` instead.

d-a-v:
> As I believe the inspiration for this PR came in part from #7148, 

Yes, because renaming is incomplete and unclear.

> has more comprehensive and consistent renamings

This PR renames also esp_schedule along with esp_yield.

> fixes the 'delay(0)` issue

That was planned to be done in another PR.

> maintains esp_yield() semantics of suspend and re-schedule

I don't see what you're talking about, neither why you advertise your PR in another one, apart from trying to show the world how clever you are.

> solves compatibility issues with 3rd party MT libs by introducing the define HAVE_ESP_SUSPEND in coredecls.h

unrelated

> A minimal_yield(), though I disagree with 1000µs suiting all cases, in particular serial UART available() used to use 10000µs and that seem a better fit. Therefore, optimistic_yield() shall be maintained and the argument must not be disregarded.

The value of 1000µs is justified above, as an answer to your notice.
You did not reply to it and insist on your 10ms value is a "better fit", to what, I don't get that. 
This is one of the reasons of my PR: do not let users to play with this value when they clearly do not know what it is about.

> is not too easily inviting new invalid calls to yield() from SYS, which can so simply be replaced by the proper esp_yield() instead.

Yelling on the console each time this function is used when misused of the contrary of an invitation. 
I'm sorry that I have to read so many non-sense. Good luck with #7952.

dok-net:
@d-a-v I'm truly sorry that you get upset, it's not my intention and I've given my writing a few iterations to express things calmly, and based on the facts. I would hope that you can see it this way: let's get the best of everyone's ideas together. Having this PR #7148 posted ahead of yours, seems like a reasonable strategy to reference for reading, reviewing, merging and extracting.

You said,
> I don't see what you're talking about, neither why you advertise your PR in another one, apart from trying to show the world how clever you are.

Well there's nothing wrong with that, unless putting some else down and doing things exclusively for the purpose of showing off - which you can't really believe to be my intention. Mind you, I'm maintaining CoopTask and as such I hold some stake in stable APIs, isn't that sufficient reason to review and comment and propose changes to PRs that affect this?
In fact, I was surprised why you didn't review #7148 in detail and instead started a new one yourself.

Perhaps you felt my comment was trying to outline differences to your PR, no, that would be a misunderstanding, I was listing what it does, not claiming that this #7832 was doing worse overall. Sorry if that's what you thought I meant by it.

Where I said
> disagree with 1000µs suiting all cases,

it doesn't mean that 1000µs is always wrong! I only find that there are other cases where 10000µs is better suited, so while your PR would have used 1000µs everywhere regardless of the argument to `optimistic_yield()`, I agree to introduce and patch for `minimal_yield()` (if that's the final name after review), but let's keep the ability to set 10000µs where it's suitable, like `available()` functions.

I trying not to think badly of this remark you made:
> I'm sorry that I have to read so many non-sense.

I feel #7148 is changing enough for one PR now, please keep your work available such that those parts going beyond the former can be implemented by follow-ups.

PR #7835, opened 2021-01-23
Add 24-bit mode to I2S

### LINKED ISSUE #5244: supporting of different bits per sample
In https://github.com/espressif/esp-idf/blob/master/components/driver/include/driver/i2s.h , there are different type of bit width per sample supported as following.
/**
 * @brief I2S bit width per sample.
 *
 */
typedef enum {
    I2S_BITS_PER_SAMPLE_8BIT    = 8,        /*!< I2S bits per sample: 8-bits*/
    I2S_BITS_PER_SAMPLE_16BIT   = 16,       /*!< I2S bits per sample: 16-bits*/
    I2S_BITS_PER_SAMPLE_24BIT   = 24,       /*!< I2S bits per sample: 24-bits*/
    I2S_BITS_PER_SAMPLE_32BIT   = 32,       /*!< I2S bits per sample: 32-bits*/
} i2s_bits_per_sample_t;

However, in Arduino/cores/esp8266/i2s.h there is no description on how to use these.

Is there supporting of these bits_per_sample from cores for esp8266?
If yes, where can I find the usage documentation or examples?
If no, will there be a schedule to support them?

Thanks for any replying.

earlephilhower:
Add basic 24 bit mode to the I2S API with a i2s_set_bits() call.
By default 16b mode is still used, but if i2s_set_bits(24) is run
before i2s_begin() then the HW will drive 24-bits of data.  This
data must be left-aligned (i.e. bits 31..8) in 4-byte samples.

Fixes #5244 (the HW doesn't support 8 or 32 bits, only 16 or 24).


d-a-v (approved):
LGTM

PR #7841, opened 2021-01-25
allow to set pin to OUTPUT_OPEN_DRAIN in analogWriteMode

### LINKED ISSUE #7836: wiring analogWrite() function does not respect previously set pinMode()
### Basic Infos

- [X ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X ] I have tested that the issue is present in current master branch (aka latest git).
- [X ] I have searched the issue tracker for a similar issue.
- [X ] If there is a stack dump, I have decoded it.
- [X ] I have filled out all fields below.

#### Platform

- Hardware:				[ESP-12E]
- Core Version:			[master branch and 2.7.4 (3.20704.0)]

### Settings in IDE

- Module:  			[Nodemcu]
- Flash Mode:			[do not know - not relevant]
- Flash Size: 		        [4MB]
- lwip Variant:                  [do not know - not relevant]
- Reset Method:		[nodemcu]
- Flash Frequency:	        [do not know - not relevant]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:              [115200] (serial upload only)

### Problem Description

The circuit I'm using the NodeMCU in requires PWM open-drain outputs. Even when setting the outputs to open-drain via pinMode(), passing OUTPUT_OPEN_DRAIN, the analogWrite() function resets the output to push-pull mode (OUTPUT).

The problem can be found in line 65 of the source file  Arduino/cores/esp8266/core_esp8266_wiring_pwm.cpp (in the current master branch), where the pinMode() is reset to OUTPUT, regardless of what was previously set via this method.

To reproduce the problem, the sketch below can be run, with the output pin connected via a pull-down resistor (10k) to ground, and an oscilloscope connected to the output. The output should stay at 0V all the time (as pwm should either connect the output to GND, or leave it open, where the pull-down resistor would keep it at GND potential, if open-drain was respected). The observed output transitions between 0V and 3V3, as corresponds to a push-pull output configuration.

```cpp

#include <Arduino.h>

void setup() {
   pinMode(14, OUTPUT_OPEN_DRAIN);
   analogWrite(14, 128);
}

void loop() {
   // Nothing needed here
}

```

### Debug Messages

```
No debug messages
```

klugem:
I added an optional parameter in the analogWrite() function to allow open-drain PWM signals.
Also see #7836

*edit*: closes #7836


earlephilhower:
The problem w/this PR is that `Arduino.h` has all its functions as `extern "C"` which means things like different signatures and default values are not going to compile properly.

Maybe an `analogWriteEx(pin, val, setDrive)` and a corresponding `digitalWriteEx(pin, val, setDrive)` or something similar would be a better way forward?

klugem:
> The problem w/this PR is that Arduino.h has all its functions as extern "C" which means things like different signatures and default values are not going to compile properly.

Thanks, I just noticed that. 

> Maybe an analogWriteEx(pin, val, setDrive) 

I named it analogWriteMode() but we can rename it to analogWriteEx()

>and a corresponding digitalWriteEx(pin, val, setDrive)

I think digitalWrite() does not set pin mode to OUTPUT or does it?

earlephilhower:
> > Maybe an analogWriteEx(pin, val, setDrive)
> I named it analogWriteMode() but we can rename it to analogWriteEx()

No problem, your name makes sense.

> > and a corresponding digitalWriteEx(pin, val, setDrive)
> I think digitalWrite() does not set pin mode to OUTPUT or does it?

That's correct, my mistake.  I just checked upstream, too, and that's the Arduino way.

earlephilhower (approved):
LGTM, thx!  Need other maintainer to give :+1: to merge since it does add a new API.

d-a-v (approved):
LGTM

klugem:
I just saw that my editor removed a lot of whitespaces in ``doc/reference.rst`` and re-added these whitespaces with the last commit... -.-

earlephilhower:
@klugem you got the spaces thing fixed, but there's now a regression of a submodule, SoftwareSerial.

Easiest thing to do is go to the libraries/SWSerial directory and `git checkout 27477f7` and then go up to the main git repo and redo the `git commit -a` to match it to current master.  Or, if your git-fu is strong, just remove the SWSerial reference in the patch...

klugem:
This was caused because I clicked on "Update branch"...I reverted this commit.

dok-net:
@earlephilhower This commit is not OK, despite reviews. What is the intended function on an "open" analogWrite? You can't switch the mode in that case due to the "else":
```
if (analogMap & 1UL << pin) {
  // Per the Arduino docs at https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/
  // val: the duty cycle: between 0 (always off) and 255 (always on).
  // So if val = 0 we have digitalWrite(LOW), if we have val==range we have digitalWrite(HIGH)
 
    analogMap &= ~(1 << pin);
  }
  else {
    if(openDrain) {
      pinMode(pin, OUTPUT_OPEN_DRAIN);
    } else {
      pinMode(pin, OUTPUT);
    }
  }
```

Plus, the comment about on and off has drifted far from anything that it might apply to :-)

dok-net:
Give me some hint as to how this should behave and I'll PR a fix.

PR #7844, opened 2021-01-27
eboot: .RODATA, upstream uzlib, move CRC, save 112 bytes 

earlephilhower:
RODATA can be copied automatically by the bootrom, so no reason not to
allow its use for strings and constants in eboot.c

Revert to pfalcon's original uzlib since the single patch to remove
RODATA is not required.

Rationalize eboot.ld linker script, clean up BSS and init it in code.

Saves 112 bytes of space in the bootloader sector by removing the
extra code associated with literal loads.

 Move CRC out of bootload sector

We added protection to only erase the bootload sector when flashing an
image when the new sector != the old sector.  This was intended to
minimize the chance of bricking (i.e. if there was a powerfail during
flashing of the boot sector the chip would be dead).

Unfortunately, by placing the CRC inside the eboot sector *every*
application will have a unique eboot sector (due to the crc/len), so
this protection doesn't work.

Move the CRC into the first 8 bytes of IROM itself.  This frees up extra
space in the boot sector and ensures that eboot won't be reflashed
unless there really is an eboot change.


earlephilhower:
The extra 80 bytes free should help #6538 squeeze into the 4K chunk.  We can save another 16 bytes if we dump the `.bss` clearing code (safe for now, but if you have any global static default values it will cause issues).

cores/esp8266/Esp.cpp
```diff
@@ -444,21 +444,23 @@ bool EspClass::checkFlashConfig(bool needsEquals) {
...
     z[0] = z[1] = 0;
 
     // Start the checksum
-    uint32_t crc = crc32((const void*)0x40200000, 4096-8, 0xffffffff);
+    uint32_t crc = crc32((const void*)0x40200000, 4096 + 16, 0xffffffff);
     // Pretend the 2 words of crc/len are zero to be idempotent
     crc = crc32(z, 8, crc);
     // Finish the CRC calculation over the rest of flash
-    crc = crc32((const void*)0x40201000, flashsize-4096, crc);
+    crc = crc32((const void*)0x40201018, flashsize - 4096 - 16 - 8, crc);
```
mcspr, line 463: btw, we already know the 'location' via `&__crc_len` aka `0x40201010`. may these should use locations of vars + their size?
(or `&_irom0_text_start` with is also the start of the section. and, I think linker already creates `__irom0_text_start`? or something like that for each section automatically)

And I wonder if python script can somehow extract those off of the .elf instead of `4096 + 16`

earlephilhower: Good idea.  In the core I think that can be done. simply enough.  I'll throw an update.

In the Python script, accessing the symbol table would require scraping `objdump` output or something, which I think would be less clear and more prone to errors than what's there now.

d-a-v (approved):
WorksForMe!

DriekdeGadgetfreak:
> 
> Move CRC out of bootload sector
> 
The consequence of this move is that "checksum" (in write_bin) is not valid anymore, because the data is changed afterwards.

PR #7854, opened 2021-01-31
Reboot into UART download mode without external wiring or interaction

twischer:
This patch introduces the new method `Esp.reboot_uart_dwnld()`.
With this method the UART download mode can be selected by software only.

When the user calls this method the ESP8266 reboots into the UART
download mode. In this mode the user can use esptool.py to flash a new
firmware file.
This patch was tested with the following command:
```
$ esptool.py --before no_reset --after soft_reset \
    --chip esp8266 --port /dev/ttyUSB0 --baud 460800 \
    write_flash 0x0 ./firmware.bin
```

The implementation is based on the original implementation in the
boot ROM. Some parts of the original implementation can be found in
https://github.com/twischer/xtensa-subjects/tree/reversed/reversed


earlephilhower:
This PR seems to include a bunch of NONOS binaries, too.  Maybe they got pulled in by accident?  Can you pull them out of the PR?

Also, can you explain why this would be useful?  esptool is quite capable of resetting/uploading already w/o extra code in IRAM/etc., so I don't understand the rationale.

twischer:
Hello @earlephilhower,

> This PR seems to include a bunch of NONOS binaries, too. Maybe they got pulled in by accident? Can you pull them out of the PR?

This PR is not adding new binaries. It is only modifying them:
```
$ git show -- tools/sdk/lib/NONOSDK*
diff --git a/tools/sdk/lib/NONOSDK221/libmain.a b/tools/sdk/lib/NONOSDK221/libmain.a
index c0e70caf..f008bc9d 100644
Binary files a/tools/sdk/lib/NONOSDK221/libmain.a and b/tools/sdk/lib/NONOSDK221/libmain.a differ
diff --git a/tools/sdk/lib/NONOSDK22x_190313/libmain.a b/tools/sdk/lib/NONOSDK22x_190313/libmain.a
index a19b330c..66403108 100644
Binary files a/tools/sdk/lib/NONOSDK22x_190313/libmain.a and b/tools/sdk/lib/NONOSDK22x_190313/libmain.a differ
diff --git a/tools/sdk/lib/NONOSDK22x_190703/libmain.a b/tools/sdk/lib/NONOSDK22x_190703/libmain.a
index 1213ceb8..cd5fd1e8 100644
Binary files a/tools/sdk/lib/NONOSDK22x_190703/libmain.a and b/tools/sdk/lib/NONOSDK22x_190703/libmain.a differ
diff --git a/tools/sdk/lib/NONOSDK22x_191024/libmain.a b/tools/sdk/lib/NONOSDK22x_191024/libmain.a
index 58779578..cd01e1a3 100644
Binary files a/tools/sdk/lib/NONOSDK22x_191024/libmain.a and b/tools/sdk/lib/NONOSDK22x_191024/libmain.a differ
diff --git a/tools/sdk/lib/NONOSDK22x_191105/libmain.a b/tools/sdk/lib/NONOSDK22x_191105/libmain.a
index 2bc98579..0f66ab6b 100644
Binary files a/tools/sdk/lib/NONOSDK22x_191105/libmain.a and b/tools/sdk/lib/NONOSDK22x_191105/libmain.a differ
diff --git a/tools/sdk/lib/NONOSDK22x_191122/libmain.a b/tools/sdk/lib/NONOSDK22x_191122/libmain.a
index 77b7de1b..260be180 100644
Binary files a/tools/sdk/lib/NONOSDK22x_191122/libmain.a and b/tools/sdk/lib/NONOSDK22x_191122/libmain.a differ
diff --git a/tools/sdk/lib/NONOSDK3V0/libmain.a b/tools/sdk/lib/NONOSDK3V0/libmain.a
index e93201c5..c30cc88a 100644
Binary files a/tools/sdk/lib/NONOSDK3V0/libmain.a and b/tools/sdk/lib/NONOSDK3V0/libmain.a differ
```
I have changed https://github.com/esp8266/Arduino/blob/master/tools/sdk/lib/fix_sdk_libs.sh and therefore I have applied this modification already to the libraries. This modification is required because no symbol name is set for a function I need to call.

> Also, can you explain why this would be useful?

I am interested in building a quite small and chip board which is water proofed. Therefore I do not want to have any additional connectors/wiring to the outside. The only connector attached is a USB plug for power supply. But I am also connecting the UART Tx and Rx line to D+ and D- of the USB plug. Therefore UART is accessible but there is no place for any additional wiring (connected to reset and GPIO2).

In the past also others were asking for such a solution:
https://www.esp8266.com/viewtopic.php?p=38310

cores/esp8266/esp8266_undocumented.h
```diff
@@ -204,8 +218,22 @@ extern fn_c_exception_handler_t _xtos_c_handler_table[XCHAL_EXCCAUSE_NUM];
   `#ifdef __cplusplus`.
 */
 extern fn_c_exception_handler_t _xtos_set_exception_handler(int cause, fn_c_exception_handler_t fn);
+
+extern void window_spill_exc_handler(struct __exception_frame* ef);
```
earlephilhower, line 222: The core on the 8266 doesn't have register windows implemented, why is this added?

tools/sdk/ld/eagle.rom.addr.v6.ld
```diff
@@ -336,6 +338,7 @@ PROVIDE ( uart_rx_readbuff = 0x40003ec8 );
 PROVIDE ( uart_tx_one_char = 0x40003b30 );
 PROVIDE ( wepkey_128 = 0x4000bc40 );
 PROVIDE ( wepkey_64 = 0x4000bb3c );
+PROVIDE ( window_spill_exc_handler = 0x400017e0 );
```
earlephilhower, line 341: Same register window spill question

tools/sdk/ld/eagle.rom.addr.v6.ld
```diff
@@ -206,6 +207,7 @@ PROVIDE ( memset = 0x4000e190 );
 PROVIDE ( multofup = 0x400031c0 );
 PROVIDE ( pbkdf2_sha1 = 0x4000b840 );
 PROVIDE ( phy_get_romfuncs = 0x40006b08 );
+PROVIDE ( print_fatal_exc_handler = 0x40001878 );
```
earlephilhower, line 210: Not used in the code, why added?

cores/esp8266/esp8266_undocumented.h
```diff
@@ -204,8 +218,22 @@ extern fn_c_exception_handler_t _xtos_c_handler_table[XCHAL_EXCCAUSE_NUM];
...
+extern void print_fatal_exc_handler(struct __exception_frame* ef);
```
earlephilhower, line 223: Not used, why defined?

cores/esp8266/reboot_uart_dwnld.cpp
```diff
@@ -0,0 +1,112 @@
...
+	// Call _ResetVector() instead of original implementation
+	// Therefore esptool.py stub is not yet supported
+	/* moved from system_restart_core_uart_dwnld() to here
+	 * Without this bit clearing the _ResetVector() would run into
+	 * an exception
+	 */
+	CLEAR_PERI_REG_MASK(PERIPHS_DPORT_24, 0x18);
+	_ResetVector();
+
+	// TODO calling user_start_fptr() fails with an exception
```
earlephilhower, line 43: I don't quite understand the purpose of the comment or commented out code.  Remove?

cores/esp8266/reboot_uart_dwnld.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+	Wait_SPI_Idle(flashchip);
+
+	// TODO exception when calling uart_div_modify()
+	//Cache_Read_Disable();
+	//CLEAR_PERI_REG_MASK(PERIPHS_DPORT_24, 0x18);
+	main_uart_dwnld();
+}
+
+[[noreturn]] void system_restart_local_uart_dwnld()
```
earlephilhower, line 84: This is the only bit of code usable out of this file, so can we make the other functions `static` to minimize namespace pollution?

cores/esp8266/Esp.cpp
```diff
@@ -203,6 +204,15 @@ void EspClass::restart(void)
     esp_yield();
 }
 
+void EspClass::reboot_uart_dwnld()
```
earlephilhower, line 207: The class use `camelCase` for other methods, can we do the same here?  And no need to shorten names, our assembler doesn't have a limited identifier len. :)  `rebootUartDownload()` or something?

cores/esp8266/reboot_uart_dwnld.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+	if (system_func1(0x4) == -1) {
```
earlephilhower, line 86: Comments to explain what black magic you're doing would be very useful for maintaining the code, here and in the following blocks where internal bits are being twiddles.

earlephilhower (changes requested):
Thanks for the explanation in the PR.

Overall seems like a fine idea and useful in some circumstances, but a little polish would help make it maintainable later on.

cores/esp8266/esp8266_undocumented.h
twischer: Removed, now

cores/esp8266/reboot_uart_dwnld.cpp
twischer: Previously it was not working with esptool.py flasher stub (--no-stub) option was required. The commented part was at least required to support esptool.py flasher stub.

I have solved the issue and esptool.py flasher stub mode is supported, now.

cores/esp8266/reboot_uart_dwnld.cpp
twischer: I have reduced it to two methods. Unfortunately when I add static to `esp8266UartDownloadMode()` it does not work anymore. I guess this function will then inlined and there not located in IRAM.

cores/esp8266/reboot_uart_dwnld.cpp
twischer: Unfortunately even I do not know what this function is exactly doing. I have only re-implemented the original implementation of `system_restart_local()` from `libmain.a`.
I have added comments from where I have good the corresponding lines.

earlephilhower (approved):
Thanks for the add'l comments and cleanup.  LGTM!

earlephilhower (approved):
Thanks for the cleanup and add'l comments.  LGTM!

tools/sdk/lib/fix_sdk_libs.sh
```diff
@@ -2,6 +2,13 @@
 set -e
 
 export PATH=../../xtensa-lx106-elf/bin:$PATH
+VERSION=$(basename ${PWD})
+
+addSymbol() {
```
d-a-v, line 7: For clarity, I would rename this function to `addSymbol_system_func1()`.

d-a-v (approved):
LGTM.
Regular builds are unmodified according to a quick test when this new API is not used.

PR #7863, opened 2021-02-07
String: compatibility with 64 bits scalars

d-a-v:
`time_t` is now 64 bits.
This PR resolves `String(time_t)` ambiguity.


earlephilhower (approved):
LGTM!

cores/esp8266/stdlib_noniso.h
```diff
@@ -36,10 +36,14 @@ char* itoa (int val, char *s, int radix);
...
 
+char* lltoa (long long val, char* str, int slen, long long radix);
+
 char* utoa (unsigned int val, char *s, int radix);
 
 char* ultoa (unsigned long val, char *s, int radix);
- 
+
+char* ulltoa (unsigned long long val, char* str, int slen, unsigned long long radix);
+
```
mcspr, line 46: `int radix`?

cores/esp8266/WString.cpp
```diff
@@ -313,6 +339,16 @@ unsigned char String::concat(unsigned long num) {
...
+unsigned char String::concat(long long num) {
+    char buf[2 + 3 * sizeof(long long)];
+    return concat(buf, sprintf(buf, "%lld", num));
+}
+
+unsigned char String::concat(unsigned long long num) {
+    char buf[1 + 3 * sizeof(unsigned long long)];
+    return concat(buf, sprintf(buf, "%llu", num));
+}
+
```
mcspr, line 351: `4 * sizeof(...)`?

mcspr (commented):
Also note of the `charconv` in the stdlib (and it's source as well)
https://en.cppreference.com/w/cpp/utility/to_chars
https://en.cppreference.com/w/cpp/string/basic_string/to_string

lgtm as-is, but String is inherently weird with it's approach to conversions :/

d-a-v:
@mcspr I tried with `charconv` that I didn't know. That's very handy !
That one is faster (1.6x) but has a cost (+880B/flash+256B/rodata)
It has to be noted that these functions `charconv(std::to_chars)` or `lltoa` are used only when radix is not 10 (in which case `sprintf` is used). It should be quite rare.
Without further comments, I will push a newer commit keeping only `lltoa`. These functions will anyway not be linked in the general case.

cores/esp8266/WString.cpp
d-a-v: In this case, base is 10.
So it's `ceil(log(2^(s*8))/log(10)/s)` with `s=sizeof(type)`.
Result is 3 even with `long long`.

mcspr: Ah, it is base10... So, basically:
```python
>>> len(str(1 << 63))
19
```
Plus sign, plus `\0`

I remembered the to_string sources I looked some time ago, all of to_string the types used 'times 4', I'd guess for consistency and to avoid signed / unsigned slight differences

cores/esp8266/WString.cpp
```diff
@@ -98,6 +98,59 @@ String::String(unsigned long value, unsigned char base) {
...
+String::String(unsigned long long value) {
+    init();
+    char buf[1 + 8 * sizeof(unsigned long long)];
+    sprintf(buf, "%llu", value);
+    *this = buf;
+}
+
+#if 1
```
earlephilhower, line 115: I'd still drop the `#if 1` bit and just remove the `charconv` portion to avoid `ifdef` hell.  The git commit will reference this chain, so if someone really cares they can go back and find what testing you did to end up here.

Maybe just a note that "using charconv is possible but cost 880b of flash and 256b of RAM so was not acceptable" or something in the comments if you really want to be explicit.

I can't see String(int64) constructor performance being a concern to people.  If they cared about performance they wouldn't be using String in the first place...

cores/esp8266/WString.cpp
d-a-v: I kept 3 for coherency with above/below lines. Indeed 4 is used when base is a variable.

cores/esp8266/WString.cpp
d-a-v: Of course that was intended :]

PR #7873, opened 2021-02-12
Allow attribute get on root

qistoph:
This PR enables attributes on the root, such as [creation/modification time of volume](https://github.com/earlephilhower/mklittlefs/pull/15).

```cpp
  fs::Dir fs = LittleFS.openDir("/");
  Serial.print("FS creation timestamp: ");
  Serial.println(fs.fileCreationTime());
```

After quite some testing I believe the attached changed is the only one required. It didn't seem to break anything in my tests, though I'm not 100% sure, because the use and value of `_valid` in undocumented and unclear to me.

Test project available: [LittleFSTest](https://github.com/qistoph/LittleFSTest)

Effects of checking 'c' and 't' attributes on a root node, if:
1. **attribute present**: return its value (mklittlefs since [PR15](https://github.com/earlephilhower/mklittlefs/pull/15))
2. **attribute unavailable at root and sub nodes**: return 0 (older version of mklittlefs)
3. **attribute unavailable at root, but available at sub node**: return (last?) sub node's value

I'm haven't been able to figure out why situation 3. doesn't respond like 2. If this is an undesirable side effect, I'm hoping someone has a suggestion on fixing it or where to look.

### DIFF
--- a/cores/esp8266/FS.cpp
+++ b/cores/esp8266/FS.cpp
@@ -441,6 +441,13 @@ bool FS::rename(const String& pathFrom, const String& pathTo) {
     return rename(pathFrom.c_str(), pathTo.c_str());
 }
 
+time_t FS::getCreationTime() {
+    if (!_impl) {
+        return 0;
+    }
+    return _impl->getCreationTime();
+}
+
 void FS::setTimeCallback(time_t (*cb)(void)) {
     if (!_impl)
         return;
--- a/cores/esp8266/FSImpl.h
+++ b/cores/esp8266/FSImpl.h
@@ -115,7 +115,7 @@ public:
     virtual bool rmdir(const char* path) = 0;
     virtual bool gc() { return true; } // May not be implemented in all file systems.
     virtual bool check() { return true; } // May not be implemented in all file systems.
-    virtual time_t getCreationTime() { return -1; } // May not be implemented in all file systems.
+    virtual time_t getCreationTime() { return 0; } // May not be implemented in all file systems.
 
     // Filesystems *may* support a timestamp per-file, so allow the user to override with
     // their own callback for all files on this FS.  The default implementation simply
--- a/cores/esp8266/FSImpl.h
+++ b/cores/esp8266/FSImpl.h
@@ -115,6 +115,7 @@ public:
     virtual bool rmdir(const char* path) = 0;
     virtual bool gc() { return true; } // May not be implemented in all file systems.
     virtual bool check() { return true; } // May not be implemented in all file systems.
+    virtual time_t getCreationTime() { return -1; } // May not be implemented in all file systems.
 
     // Filesystems *may* support a timestamp per-file, so allow the user to override with
     // their own callback for all files on this FS.  The default implementation simply
--- a/libraries/LittleFS/src/LittleFS.h
+++ b/libraries/LittleFS/src/LittleFS.h
@@ -221,6 +221,26 @@ public:
             return false;
         }
 
+        if(_timeCallback && _tryMount()) {
+            // Mounting is required to set attributes
+
+            time_t t = _timeCallback();
+            rc = lfs_setattr(&_lfs, "/", 'c', &t, 8);
+            if (rc != 0) {
+                DEBUGV("lfs_format, lfs_setattr 'c': rc=%d\n", rc);
+                return false;
+            }
+
+            rc = lfs_setattr(&_lfs, "/", 't', &t, 8);
+            if (rc != 0) {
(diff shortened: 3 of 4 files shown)


earlephilhower (changes requested):
I'm worried that when `_valid` is false the filename is not guaranteed to be proper.  Has this been verified?

Also, the `::format` method should be setting these attributes when on-device formatting occurs.

qistoph:
I'll extend my tests and verify the workings of all available methods.

Where would you say is the right place to set the attributes on `::format`?
1. `LittleFSImpl::format` (LittleFS.h)
2. `lfs_format` (lfs.c)

The first would be most in line with the change in mklittlefs, I guess.

earlephilhower:
The `LittleFSImpl::format()` would be the spot to add the attributes.  We don't touch the upstream FS lib itself.

The _valid thing still sets off alarms.  You're assuming that dirent is all 0'd (which it is, luckily, in the constructor...but only until something writes to it which can happen on rewind() or next() or other calls).  Also, this would return the FS date no matter what the subdir it's in which is not what you're going for.  Currently, also, if I read the sprintf logic you're doing a lfs_get_attr("") and not on '\' .

qistoph:
Thanks for the feedback and insights. I have updated the code to a more proper way to get the attributes. Also included is the proposed change to set the attributes on format.

```
LittleFS.creationTime()
```

libraries/LittleFS/src/LittleFS.h
```diff
@@ -221,13 +221,48 @@ class LittleFSImpl : public FSImpl
             return false;
         }
 
+        if(_timeCallback && _tryMount()) {
+            // Mounting is required to set attributes
+
+            time_t t = _timeCallback();
+            rc = lfs_setattr(&_lfs, "/", 'c', &t, 8);
+            Serial.printf("rc1: %d\n", rc);
```
earlephilhower, line 229: CAn we get rid of this and the following debug printout, please?

cores/esp8266/FSImpl.h
```diff
@@ -115,6 +115,7 @@ class FSImpl {
     virtual bool rmdir(const char* path) = 0;
     virtual bool gc() { return true; } // May not be implemented in all file systems.
     virtual bool check() { return true; } // May not be implemented in all file systems.
+    virtual time_t creationTime() { return -1; } // May not be implemented in all file systems.
```
earlephilhower, line 118: Can we make this `getCreationTime` like the existing `File::getCreationTime`  call?

libraries/LittleFS/src/LittleFS.h
```diff
@@ -221,13 +221,48 @@ class LittleFSImpl : public FSImpl
...
+    time_t creationTime() override {
+        time_t t;
+        uint32_t t32b;
+
+        if (lfs_getattr(&_lfs, "/", 'c', &t, 8) == 8) {
+            return t;
+        } else if (lfs_getattr(&_lfs, "/", 'c', &t32b, 4) == 4) {
+            return (time_t)t32b;
+        } else {
+            return -1;
```
earlephilhower, line 262: Can we return 0 and not -1 for this?  The existing calls return 0 (Jan 1 1970)

cores/esp8266/FS.cpp
```diff
@@ -441,6 +441,13 @@ bool FS::rename(const String& pathFrom, const String& pathTo) {
     return rename(pathFrom.c_str(), pathTo.c_str());
 }
 
+time_t FS::creationTime() {
+    if (!_impl) {
+        return -1;
```
earlephilhower, line 446: Return 0 like File::getCreationTime please.

earlephilhower (changes requested):
Good way to add in the call!  Just some minor things and we should be good to go!

cores/esp8266/FS.cpp
qistoph: Sure, just thought it might be a nice way to check for errors vs valid values. I'll change it to `return 0`.

libraries/LittleFS/src/LittleFS.h
qistoph: You're absolutely right. I'm sorry for leaving these in the PR.

cores/esp8266/FSImpl.h
qistoph: Sure, that makes sense.

libraries/LittleFS/src/LittleFS.h
qistoph: Same as the other, might make it easier to differentiate between errors van valid values.

earlephilhower (approved):

PR #7874, opened 2021-02-13
Add I2S class support

### LINKED ISSUE #427: I2S Support
Now, we know how work I2S on ESP8266
https://github.com/espressif/esp8266_mp3_decoder/blob/master/mp3/include/i2s_reg.h
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21998963-i2s-support?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14245935&utm_medium=issues&utm_source=github).
</bountysource-plugin>

earlephilhower:
Fixes #427

Adds a basic I2S class based off of the Arduino-SAMD core.  The raw
i2s_xxx functions are still a better  way to use I2S due to their
flexibility, but this will allow basic Arduino sketches to work.


libraries/I2S/src/I2S.cpp
```diff
@@ -0,0 +1,209 @@
...
+  _running = false;
+  _onTransmit = nullptr;
+  _onReceive = nullptr;
+  _havePeeked = 0;
+  _peekedData = 0;
+  _bps = 0;
+  _writtenHalf = false;
+}
+
+int I2SClass::begin(int mode, long sampleRate, int bitsPerSample) {
```
d-a-v, line 36: `int mode` -> `i2s_mode_t mode` ?
Or is it that the goal is to maintain the minimum difference with upstream lib ?

libraries/I2S/src/I2S.cpp
```diff
@@ -0,0 +1,209 @@
...
+  return i2s_available();
+}
+
+void I2SClass::flush() {
+  /* No-op */
+}
+
+int I2SClass::read() {
+  if (!_running) return -1;
+  if (_havePeeked) {
```
d-a-v, line 86: To avoid recursion, would this be better, or again is it preferable to keep sync with upstream lib ?
```
if (!_havePeeked) peek();
if (!_havePeeked) return -1;
if (_bps ==16) ...
```

libraries/I2S/src/I2S.cpp
earlephilhower: Good catch.  Upstream lib has `int` IIRC, but it should be the enum.  Will fix.

libraries/I2S/src/I2S.cpp
earlephilhower: Good idea, done.

libraries/I2S/src/I2S.h
```diff
@@ -0,0 +1,86 @@
...
+  // from Print
+  virtual size_t write(uint8_t);
+  virtual size_t write(const uint8_t *buffer, size_t size);
+
+  virtual int availableForWrite();
+
+  int read(void* buffer, size_t size);
+
+  size_t write(int32_t);
+  size_t write(const void *buffer, size_t size);
```
d-a-v, line 57: I would add a comment specifying that `::write(sample)` is blocking but `::write(buffer,len)` is not.

d-a-v (approved):

PR #7888, opened 2021-02-19
String: add missing long long operator implementations

### LINKED ISSUE #7885: Missing operator +() implementation after : String: compatibility with 64 bits scalars (#7863)
### Basic Infos

### Platform

- Hardware:					[All]
- Core Version:			[483519f852d37abbd4fc15063b60e9763ccc62c2]
- Development Env:		[All]
- Operating System:		[All]

### Settings in IDE

- Module:  			[All]
- Flash Mode:			[N/A]
- Flash Size: 		[N/A]
- lwip Variant:   [N/A]
- Reset Method:		[N/A]
- Flash Frequency:	[N/A]
- CPU Frequency:		[N/A]
- Upload Using:		[N/A]
- Upload Speed:   [N/A]

### Problem Description

In WString.h there are new friend functions:

       friend StringSumHelper &operator +(const StringSumHelper &lhs, long long num);
       friend StringSumHelper &operator +(const StringSumHelper &lhs, unsigned long long num);

The function implementations seem to be missing from WString.cpp

d-a-v:
fixes #7885


earlephilhower (approved):
LGTM

Looking at this now, it seems like `StringSumHelper` is screaming for a template...

d-a-v:
> Looking at this now, it seems like `StringSumHelper` is screaming for a template...

I told the same to myself and I remembered that we have pending #7781 which would remove all of them.

PR #7891, opened 2021-02-21
EspSoftwareSerial, I keep forgetting to bump up Arduino library manager files

dok-net:
Bug fix 6.11.6 - otherwise, ESP32 drifts and people that don't read the README will complain even more than they already do :-)


dok-net:
@earlephilhower I'm terribly sorry - perhaps you should make it a habit to check the library.json and library.properties files before merging my EspSoftwareSerial PRs ... shame on me :-)

earlephilhower:
Wouldn't help.  I always forget to update my own JSON files on ESP8266Audio, too!

PR #7897, opened 2021-03-01
Add example for using ESP.rebootIntoUartDownloadMode()

mhightower83:
For the case when your ESP8266 module or IoT device does not have the nice reset and program circuitry, the proxy method shown in this example can help. At least for upgrades from stable code running with this proxy.

@twischer thank you for sharing your work. :+1: 
I forgot that I had wanted to do this proxy a long time ago; however, I could not find a start flash download API like yours to call.
And at that time, I was not ready for the deep digging to make it work as you did. This would have been/will be real handy when reflashing the Sonoff SV and Basic.


twischer:
@mhightower83 

> This would have been/will be real handy when reflashing the Sonoff SV and Basic.

This was also my main reason for this feature.
Great idea to enter UART download mode when ESP_SYNC was detected. This should be already unique enough. I was thinking about any unique string to sent but using ESP_SYNC itself is much better.

libraries/esp8266/examples/UartDownload/UartDownload.ino
```diff
@@ -0,0 +1,181 @@
...
+
+  // ...
+}
+
+void cmdLoop(Stream& oStream) {
+  if (0 >= oStream.available()) {
+    return;
+  }
+
+  switch (oStream.read()) {
```
twischer, line 136: This might eat the `slipFrameMarker`. May be also call `proxyEspSync()` in case of `slipFrameMarker`.

Beside this LGTM.

twischer (approved):

libraries/esp8266/examples/UartDownload/UartDownload.ino
mhightower83: Yep, race condition. I tried to separate the two operations in `loop()` for presentation and they really need to be integrated.

d-a-v: Would the logic be simplified if `oStream.peek()` is used instead ?

mhightower83: I refactored this earlier this morning. I pushed my changes. Let me know what you think.

d-a-v (approved):
LGTM!

PR #7899, opened 2021-03-01
Added broadcastIP method to WiFiSTA class

drzony:
As discussed with @d-a-v 
No need for IPv6 version, since IPv6 does not define broadcast.


d-a-v (approved):
LGTM

earlephilhower (approved):

PR #7901, opened 2021-03-01
Added ESP32 compatible methods for setting/getting sleep mode

drzony:
As discussed with @d-a-v


libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -77,6 +77,22 @@ class ESP8266WiFiGenericClass {
...
+        // ESP32 compatibility
+        bool setSleep(bool enable)
+        {
+            if (enable)
+            {
+                return setSleepMode(WIFI_MODEM_SLEEP);
+            }
+            else
+            {
+                return setSleepMode(WIFI_NONE_SLEEP);
```
d-a-v, line 89: The default sleep mode is WIFI_LIGHT_SLEEP in esp8266. I don't know in esp32 whether there are several sleep modes. So there are three possibilities:

| choice | sleep enabled | sleep disabled |
|--|--|--|
| 1 | WIFI_MODEM_SLEEP | WIFI_NONE_SLEEP |
| 2 | WIFI_LIGHT_SLEEP | WIFI_NONE_SLEEP |
| 3 | WIFI_MODEM_SLEEP | WIFI_LIGHT_SLEEP |

You implemented number 1. I think in the general esp8266 case, number 3 is commonly used, because of discussions like in #2330 where NONE_SLEEP was proposed to those with issues. I don't remember but current consumption in this mode is quite high, so making it a common use case may not be that good.
Is there another API call in ESP32 that would allow to select between different sleep modes ?

drzony: On ESP32:
```cpp
typedef enum {
    WIFI_PS_NONE,        /**< No power save */
    WIFI_PS_MIN_MODEM,   /**< Minimum modem power saving. In this mode, station wakes up to receive beacon every DTIM period */
    WIFI_PS_MAX_MODEM,   /**< Maximum modem power saving. In this mode, interval to receive beacons is determined by the listen_interval parameter in wifi_sta_config_t */
} wifi_ps_type_t;
```
there are two overloads:
```cpp
bool WiFiGenericClass::setSleep(bool enable);
bool WiFiGenericClass::setSleep(wifi_ps_type_t mode);
```
In the first one `enable ? WIFI_PS_MIN_MODEM : WIFI_PS_NONE`

From the docs on ESP8266 it seems that `WIFI_PS_MIN_MODEM == WIFI_LIGHT_SLEEP` and `WIFI_PS_NONE == WIFI_NONE_SLEEP`

I can copy-paste the above enum in ESP8266 and add those overrides. How about that?

d-a-v (approved):
Thanks

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -77,6 +77,26 @@ class ESP8266WiFiGenericClass {
...
+            }
+        }
+        bool setSleep(wifi_ps_type_t mode)
+        {
+            return setSleepMode((WiFiSleepType_t)mode);
+        }
+        bool getSleep()
+        {
+            return getSleepMode() == WIFI_LIGHT_SLEEP;
```
d-a-v, line 98: `getSleepMode() != WIFI_NONE_SLEEP`
(so LIGHT and MODEM are treated as sleep mode enabled)

drzony: I thought about that, but see my comment above. To maintain ESP32 compatibility I did it the same way that it's on ESP32:
https://github.com/espressif/arduino-esp32/blob/3fe7c2e8cdcc34af0f5d3c55f8b4cb79ef5318cf/libraries/WiFi/src/WiFiGeneric.cpp#L669

libraries/ESP8266WiFi/src/ESP8266WiFiType.h
```diff
@@ -48,6 +48,13 @@ typedef enum WiFiSleepType
     WIFI_NONE_SLEEP = 0, WIFI_LIGHT_SLEEP = 1, WIFI_MODEM_SLEEP = 2
 } WiFiSleepType_t;
 
+// ESP32 compatibility
+typedef enum wifi_ps_type
+{
+    WIFI_PS_NONE = WIFI_NONE_SLEEP,
+    WIFI_PS_MIN_MODEM = WIFI_LIGHT_SLEEP,
+    WIFI_PS_MAX_MODEM = WIFI_MODEM_SLEEP,
```
mcspr, line 56: Is this right though? MODEM is default, hence it should be MIN? LIGHT is 'sleepier' mode afaik

drzony: See this comment:
https://github.com/esp8266/Arduino/pull/7901#discussion_r584594164

mcspr: > From the docs on ESP8266 it seems that WIFI_PS_MIN_MODEM == WIFI_LIGHT_SLEEP and WIFI_PS_NONE == WIFI_NONE_SLEEP

https://www.espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf ?
(still applies to the 2.2.x branch)
> 3.5.54. wifi_set_sleep_type 
> - Default mode: Modem-sleep.
-  In order to lower the power comsumption, ESP8266 changes the TCP timer
tick from 250ms to 3s in light-sleep mode, which leads to increased timeout for
TCP timer. Therefore, the modem-sleep or deep-sleep mode should be used
where there is a requirement for the accurancy of the TCP timer.

Plus:
https://github.com/espressif/arduino-esp32/blob/15bae92a72c102331e7eaf7dcb439e8ad6920184/tools/sdk/include/esp32/esp_wifi_types.h#L192-L196

Meaning, NONE - power-save features are disabled, MIN - minimum amount of power-saving features, but some are enabled, MAX - maximum power-save mode

drzony: Then I don't understand the number 3 in @d-a-v comment. Why `WIFI_MODEM_SLEEP` is on enabled and `WIFI_LIGHT_SLEEP` is on disabled?

mcspr: :shrug:
Following esp32 api, 1st choice is used there. ::setSleep(enum) is used to select the MAX sleep mode
(ref. https://github.com/espressif/arduino-esp32/issues/3896 and https://github.com/espressif/arduino-esp32/issues/3900)

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
drzony: One other option would be to drop the bool version

libraries/ESP8266WiFi/src/ESP8266WiFiType.h
d-a-v: I was wrong, sorry! I swapped MODEM and LIGHT sleep modes.
MODEM is indeed the default and is between NONE and LIGHT.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
d-a-v: Indeed, the comment there is:
```cpp
// @return true if modem sleep is enabled
bool WiFiGenericClass::getSleep()
```
The naming is confusing, because "light sleep" is also an enabled sleep mode while "none sleep" has sleep mode disabled.
What about replicating the comment in esp8266 .h file, insisting on that it reflects the esp32 API ?

> One other option would be to drop the bool version

That would defeat the goal of this PR. We just don't need to add it in esp8266 documentation.
esp32 libraries will compile, esp8266 .h readers will see the comment, esp8266 api is unchanged.

d-a-v (approved):

Issue #3072, opened 2017-03-22
Can't re-enable WiFi after ESP.deepSleep(uS, WAKE_RF_DISABLED) ?


sheffieldnick:
#### Hardware
Hardware:			WeMos D1 mini
Core Version:      	v2.2.0

### Description

My sketch spends most of its time in deep sleep, just using the WiFi when it first powers up. This works fine from a full power-cycle, but not if is reset (reset button, or new sketch uploaded over serial). In that case I can't seem to switch the WiFi back on? I presume the RF is disabled from the deep sleep, even through a reset, and WiFi.mode(WIFI_STA) isn't enough to switch it back on?

I get:

```mode : sta(MAC)
add if0
f r0, 
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
....
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
scandone
reconnect
f 0, error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
....
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
scandone
reconnect
f -180, error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
....
error: pll_cal exceeds 2ms!!!
error: pll_cal exceeds 2ms!!!
scandone
del if0
usl
mode : null
```

### Settings in IDE

Module:  			WeMos D1 R2 & mini
Flash Size: 		4MB
CPU Frequency:		80Mhz
Flash Mode:			?qio?
Flash Frequency:	?40Mhz?
Upload Using:		SERIAL
Reset Method:		?ck / nodemcu?

### Sketch

```cpp

#include <ESP8266WiFi.h>
extern "C" {
#include <user_interface.h>
}

void setup() {
  Serial.begin(115200);
  Serial.setDebugOutput(true);
  // Connect to WiFi if we have just powered on (*not* if waking from deep sleep)
  rst_info *resetInfo = ESP.getResetInfoPtr();
  if (resetInfo->reason != REASON_DEEP_SLEEP_AWAKE) {
    WiFi.mode(WIFI_STA);
    WiFi.begin(WLAN_SSID, WLAN_PWD);
    WiFi.setAutoConnect(false);
[...]

PR #7902, opened 2021-03-02
[BREAKING] Disable WiFi at boot by default

### LINKED ISSUE #7303: Change autoreconnect and persistent defaults to be disabled (BREAKING)
### Basic Infos

### Platform

- Hardware:					All
- Core Version:			latest git

### Settings in IDE

- Module:  			all
- Flash Mode:			all
- Flash Size: 		all

### Problem Description

The SDK has a "feature" called persistence where the wifi credentials are "remembered" between connections. This is meant to ease setup, and reconnect faster on boot, but in practice the user sets up a connection with his own credentials.
There are reasons to use this feature, such as low power applications, but in the general use case there is a danger of switching between different wifi configs, such as with Multi, and every time a different connection is made, the new info is written to flash. It is quite easy for a new user to do the Wrong Thing and burn the flash sector due to this.
In addition, there is an autoreconnect feature that is related, which works under the hood. Given certain conditions, it will take a long time for the user to get a hint that the station has disconnected from the AP.

On the other hand, the typical use cases where these features would be most useful, such as low power applications, are generally advanced enough for the user to have Done His Homework about what they do, how they work, and what to do if something goes wrong.

This is proposal to change both defaults to disabled. The normal user should understand what is happening from looking at the code, and not have to guess about how some subsystem is trying to do things hidden away from view, especially when that logic is difficult to trace. If needed, both features can be enabled by calling the appropriate methods.

d-a-v:
SDK automatically starts WiFi at boot. This was probably intended for the Espressif AT FW that is interactive and preserves state accross reboots. This behavior is generally irrelevant with the Arduino API. Sketches usually start with `WiFi.begin()` or softAP...

This PR will change the default: ***WiFi does not anymore automatically fires up at boot***.

***This will not change anything to standard sketches***
Standard calls to `WiFi.mode()`+`WiFi.begin/WiFi.softAP(...)` will continue to enable WiFi as usual.
It may smooth current spikes at boot and also decrease DHCP stress.

Legacy behavior can be restored by calling `enableWiFiAtBootTime()` anywhere in the code (this is a weak void function intended to play with the linker). It *may be* necessary for WiFi Managers that rely on this SDK feature.

```cpp
void setup () {
#ifdef WIFI_IS_OFF_AT_BOOT
  enableWiFiAtBootTime(); // can be called from anywhere with the same effect
#endif
  ....
}
```

This is a proposal as discussed on gitter, here and there, and with maintainers.

Fixes #7303


dok-net:
@d-a-v Can you please clarify, whether `Wifi.begin()` alone is sufficient to fully re-enable everything that this PR turns off. It's not entirely clear to me due to the remark concerning Arduino library manager libraries above.

d-a-v:
`Wifi.begin()` alone is indeed sufficient to fully re-enable everything that this PR turns off.
I successfully tested some WiFiClient examples.
I haven't tried yet softAP examples but I will.

d-a-v:
AP mode with or without STA properly wakes WiFi  up too.
Removing `(WIP)`

earlephilhower (changes requested):
The code looks good to me now, but I really think this needs an entry in the docs somewhere.

d-a-v:
A fix is needed about the xref from the faq (I'm playing a local sphinx build to check that out).

dok-net:
@d-a-v Could you please correct the opening comments to this PR, it appears that now it's

`enableWiFiAtBootTime`

:-)

dok-net:
I don't understand.
If `enableWiFiAtBootTime()` were to restore the behavior prior this patch, shouldn't a persisted SSID/password suffice to reconnect after reset/restart despite a `WiFi.begin()` without arguments? Because, in my tests, I only ever get a connection to the AP if `WiFi.begin("...", "...")` with the credentials is used. Despite `enableWiFiAtBootTime()` right first in `setup()`.
I have cleared the flash memory completely during upload of the sketch.

doc/esp8266wifi/generic-class.rst
```diff
@@ -42,6 +42,33 @@ persistent
 
     WiFi.persistent(persistent)
 
+Starting from version 3 of this core, **persistence is disabled by default
+and WiFi does not anymore automatically fires up at boot** (see PR `#7902 <https://github.com/esp8266/Arduino/pull/7902>`__).
+
+Previously, SDK was automatically starting WiFi at boot.  This was probably
```
devyte, line 48: I suggest highlighting here that the prior to v3 persistent was defaulted to true, and so the wifi creds were always stored to flash when the WiFi was configured, which could lead to a reduced life of the wifi config flash sector, and hence the entire ESP. Starting with v3, persistent defaults to false, so creds aren't stored by the SDK in the flash sector and hence don't survive across reboots.

doc/esp8266wifi/generic-class.rst
```diff
@@ -42,6 +42,33 @@ persistent
...
+intended for the Espressif AT FW which is interactive and preserves WiFi
+state accross reboots.  This behavior is generally irrelevant with the
+Arduino API because sketches start with ``WiFi.begin()`` or
+``WiFi.softAP()``.
+
+This change is harmless with standard sketches: Calls to ``WiFi.mode()`` do
+enable radio as usual.  It also smooths current spikes at boot and decreases
+DHCP stress.
+
+Legacy behavior can be restored by calling ``enableWiFiAtBoot()`` from
```
devyte, line 58: To restore legacy behavior, in addition to calling enableWiFiAtBoot() the user has to call WiFi.persitent(true). This should be used with care, because use could lead to early wear of the wifi config flash sector and hence device failure.

devyte (changes requested):

doc/esp8266wifi/generic-class.rst
dok-net: @devyte I think this is not precisely correct. Settings that were once persisted, remain that way, so thereafter, `WiFi.begin()`, with no `WiFi.persistent(true)` in the sketch, as long as `enableWiFiAtBootTime()` is somewhere in the sketch, will suffice to auto-connect to the AP.

doc/esp8266wifi/generic-class.rst
dok-net: See my comment about, `WiFi.persistent(true)` is only needed to change the flash credentials cache, not in a sketch that just wants to use whatever was persisted at any time prior to it.

d-a-v: > the user has to call WiFi.persitent(true).

True !
I will add it in `enableWiFiAtBootTime()`

> could lead to early wear of the wifi config flash sector

Given the fact that when persistence is enabled, [user config is compared to stored config](https://github.com/esp8266/Arduino/blob/1b922edad127fa515dc7a00128f9cabfe0b1a67b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L171), I think that's OK.
What do you both think ?

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -360,7 +371,8 @@ extern "C" void user_init(void) {
 #if defined(MMU_IRAM_HEAP)
     umm_init_iram();
 #endif
-    preinit(); // Prior to C++ Dynamic Init (not related to above init() ). Meant to be user redefinable.
+    preinit(); // deprecated
+    __disableWiFiAtBootTime(); // default weak function disables WiFi
```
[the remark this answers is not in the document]
d-a-v: What about undeprecating preinit ?
(which is not really documented btw - apart from some examples switching wifi off)

doc/esp8266wifi/generic-class.rst
Bighoneypot: I always thought this mode was very important !!!

Bighoneypot: When the wifi is active there is a consumption of energy that it would be good to keep in case of not use, especially if you use batteries for their power supply. In case of sending data, as sometimes I do, it is good to write to the internal memory and then make only one connection per day and send all the data recorded in the preset intervals.

d-a-v: @Bighoneypot The two WiFi / low power examples should work as before:
https://github.com/esp8266/Arduino/tree/master/libraries/esp8266/examples/LowPowerDemo
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiShutdown/WiFiShutdown.ino

cores/esp8266/core_esp8266_main.cpp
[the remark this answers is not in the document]
devyte: that's also ok 😄

doc/esp8266wifi/generic-class.rst
TD-er: Maybe also good to test with a number of different ESP modules (ESP12E/F/S for example) to see if this also has an effect on deep sleep power consumption.
There have been a number of reports on deep sleep power consumption not being close to 60 uA on some nodes, with some reporting success with the most unclear assembly blobs possible and others claiming you need to do a short deep sleep first and then again call deep sleep to get true low power deep sleep and others reporting the WiFi will not start after deep sleep until you go into deep sleep again.

So I would not be surprised to see such topics pop up again if this is changed. (or this may be the final fix for it.... who knows ;) )

devyte: If you need deep sleep and low power, you need to set up a series of special things, such as early wifi init, persistent wifi creds, static IP, etc, and then your app will usually have something in setup() and an empty loop. Then you need to check power and optimize more. That's an entirely different and very specific use case compared to the basic "start wifi, connect to some router, wiggle some pins" sketch that newer users usually start with. 
In other words, deep sleep needs a special kind of sketch unlike the typical Arduino style setup - loop case addressed by this PR.

Bighoneypot: > @Bighoneypot The two WiFi / low power examples should work as before:
> https://github.com/esp8266/Arduino/tree/master/libraries/esp8266/examples/LowPowerDemo
> https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiShutdown/WiFiShutdown.ino

Thanks i watch this.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -419,7 +419,7 @@ bool ESP8266WiFiGenericClass::mode(WiFiMode_t m, WiFiState* state) {
     }
 
     if (wifi_fpm_get_sleep_type() != NONE_SLEEP_T) {
-        // wifi may have been put asleep by ESP8266WiFiGenericClass::preinitWiFiOff
+        // wifi starts asleep by default
```
mcspr, line 422: Looking at esp32's ::mode:
https://github.com/espressif/arduino-esp32/blob/46d5afb17fb91965632dc5fef237117e1fe947fc/libraries/WiFi/src/WiFiGeneric.cpp#L541

Should this be more in-line with that?
i.e.
- getMode() checks for fpm -> returns null-mode when sleeping
- mode() consecutively turns off the wifi when WIFI_OFF without an additional call to force sleep and turns it back on as well

d-a-v: :+1: thanks !

I added 1) and 3).
I'm reticent about 2) because `WiFi.mode(WIFI_OFF)` will not put the modem to sleep, or puting modem to sleep does not necessarily switches the mode to `WIFI_OFF`. 2) should come with that no ?

mcspr: I meant exactly that, to do forcesleep with WiFi.mode(WIFI_OFF) and resume with any real mode.
Not noticed the special shutdown and resume modes though, I wonder if those should be separate methods instead and not a 'mode'?

d-a-v: They are indeed "fake" mode and described as such somewhere. There are two examples making use of them.
They are designed to import the deep sleep procedure into the core and allow users to not use random fw-snippets to put the chip to sleep.

mcspr: Aaah, both are already implemented as methods. No reason examples could not use those directly?

devyte (approved):

doc/faq/readme.rst
```diff
@@ -177,3 +177,12 @@ will need to implement an additional (short) deep sleep using
 ``WAKE_RF_DEFAULT``.
 
 Ref.  `#3072 <https://github.com/esp8266/Arduino/issues/3072>`__
+
+My WiFi was previously automatically connected right after booting
```
earlephilhower, line 181: "My WiFi was previously automatically connected right after booting" => "My WiFi was previously automatically connected right after booting, but isn't anymore"?

earlephilhower (approved):
LGTM, thanks for adding documentation and comments.  I made 2 suggested edits to the wording, up to you if you want to adjust or not.

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -432,6 +426,12 @@ bool ESP8266WiFiGenericClass::mode(WiFiMode_t m, WiFiState* state) {
         return true;
     }
 
+    if (m != WIFI_OFF && wifi_fpm_get_sleep_type() != NONE_SLEEP_T) {
```
mcspr, line 429: small side note / question, will test some time later today
wifi_fpm_get_sleep_type() is MODEM_SLEEP_T by default (in the previous version, or when wifi-off preinit is disabled by user currently) since this is a setting, not the actual mode that is active. is it ok do wakeup & close without fpm active?

doc/esp8266wifi/generic-class.rst
Tech-TX: > Maybe also good to test with a number of different ESP modules (ESP12E/F/S for example) to see if this also has an effect on deep sleep power consumption.
> There have been a number of reports on deep sleep power consumption not being close to 60 uA on some nodes, with some reporting success with the most unclear assembly blobs possible and others claiming you need to do a short deep sleep first and then again call deep sleep to get true low power deep sleep and others reporting the WiFi will not start after deep sleep until you go into deep sleep again.
> 
> So I would not be surprised to see such topics pop up again if this is changed. (or this may be the final fix for it.... who knows ;) )

I didn't measure any difference in the Deep Sleep current.  I looked at startup current and different types of Sleep.  The only expected difference with other boards/modules would be due to extra chips on the boards (LDO, USB chip, battery charger, etc.), You can get several uA of difference if your GPIO outputs go the opposite direction from the pullup/pulldown resistors on those pins, ~2uA per pin in that case or 6-8uA higher total, but you pretty much have to have the pullup/pulldown resistors either on the module or in your circuit or it won't boot to Flash.

The people that needed to reset across Deep Sleep a second time to re-enable WiFi after ESP.deepSleep(uS, WAKE_RF_DISABLED) have a fix [here](https://github.com/esp8266/Arduino/issues/3072#issuecomment-741524721).  The only wart is that it ALWAYS does a full RF_CAL when you wake the modem that way.  I guess it blanks out the stored values for the modem parameters when you do WAKE_RF_DISABLED, so it has to rebuild everything from scratch.

As far as variable current during Deep Sleep, I've never seen it mentioned.  Most likely they weren't in Deep Sleep, for whatever reason.  The SDK stops WiFi and all timers before entering Sleep, and once it's asleep the CPU PLL screeches to a complete halt and the power drops to around 20uA, which is just the RTC timer running.  I get closer to 16uA since I'm frequently running at 3.0V instead of 3.3V.  If someone has an MCVE that demonstrates variable Deep Sleep currents I can figure out what's going on.

I _have_ seen numerous threads about Deep Sleep Zombie mode where it never wakes properly, but the jury is still out on the actual cause(s).  I vote for cheap Chinese Flash that doesn't always come up in full drive mode after the Flash has been slept prior to Deep Sleep, but Erriez thinks it's the NodeMCU reset circuit interfering.  We could both be right: two different causes or more.  Until I get one here I can dissect, my tools are useless at troubleshooting across the Internet.

PR #7906, opened 2021-03-04
Update mklittlefs executable

earlephilhower:
Rebuild entire toolchain instead of manually hacking the tools JSON to
ensure repeatability.

New mklittlefs sets the new FS timestamp added in #7873


d-a-v (approved):
worksforme

PR #7910, opened 2021-03-06
[BREAKING] base64::encode() compat with esp32: no newlines by default

### LINKED ISSUE #6604: base64::encode API incompatibility with arduino-esp32
There is an unfortunate behavior difference between ESP32's base64::encode and ESP8266 base64::encode. The signature of ESP8266 is:

```
        // is to add a newline every 72 (encoded) characters output.
        // This may 'break' longer uris and json variables
        String encode(const uint8_t * data, size_t length, bool doNewLines = true);
```

in esp32, the signature is:

```
static String encode(const uint8_t * data, size_t length);
```

and it doesn't introduce newlines. This is actually a much nicer behavior because the
primary purpose of these functions (which can only deal with small input sizes as they
don't implement a Stream interface) seems to be JSON APIs and http authentication headers.

Both must not have newlines. The only case where you want newlines is PEM (which
is already provided by the embedded BearSSL, so you can just use that api, which is
both better performing and more convenient) and mime64 encoding (which I think you won't be able to do as you have no stream interface).

I suggest to align the signature (remove the capability to inject newlines), which reduces code size, or at least flip the default.

d-a-v:
closes #6604


earlephilhower (approved):
LGTM.

devyte (approved):

Issue #7894, opened 2021-02-26
Please open a new issue, follow the template instructions, explain the problem with the addresses, and reference this issue.


6leonardo:
Please open a new issue, follow the template instructions, explain the problem with the addresses, and reference this issue.
Note: from a quick look at the code, I don't see any practical reason to force a single I2C bus, as long as it is understood that only one bus may work at any given time (sw implementation: timing of one would interfere with timing of the other). 

Two wire for i2c not support multi instances...

If you create multiple instance of twowire only the last one work

__Originally posted by @devyte in https://github.com/esp8266/Arduino/issues/3063#issuecomment-691589459__

6leonardo:
Now it works....it's not very beautiful code but seams to work...
I have tested with 4 gy-bme280... 2 for bus
here the patch in cores/esp8266 I
and in the library Wire
you can put the wire library in local library but for the core I must change the original files
[mod_i2c_multi-bus.zip](https://github.com/esp8266/Arduino/files/6046031/mod_i2c_multi-bus.zip)
I attach the code for test the 4 gy-bme280
[gy-bme280.ino.zip](https://github.com/esp8266/Arduino/files/6046054/gy-bme280.ino.zip)



__Originally posted by @6leonardo in https://github.com/esp8266/Arduino/issues/3063#issuecomment-786266957__

PR #7911, opened 2021-03-08
Split software I2C code into Master-only and Master-or-Slave parts, enabling multiple instances of TwoWire in master-only role

michkot:
Follow-up to the structural change to I2C core support from https://github.com/esp8266/Arduino/pull/6326

What I did is create class hierarchy, where TwiMaster (a subset of the original Twi class in) in "core" does not have any static members, callbacks or IRS, and is exposed in the twi.h header file (in C++ context only). 
TwiMasterOrSlave derives from it and adds all the remaining functionality of the original Twi class. 
The C interface (declared in twi.h) is rerouted from Twi singleton to TwiMasterOrSlave singleton.

Then, I modified the Wire library in similar manner, but extra measures had to be taken to guarantee compatibility with other Arduino platforms. The default mode of TwoWire class can be switched to master-only via TWOWIRE_MASTER_ONLY define - this implementation does not depend on any static/singleton variables, and also allocates the buffers dynamically (allowing for fine-tuning of necessary buffer size for each I2C interface).

Note: there is already an issue #7894 asking to support multiple I2C instances (with a kinda crude but probably working solution provided already - turning all the static dependencies into 5-tuples so that up-to 5 master/slave I2C interfaces can be used at a time).


michkot:
There seems to be a bug in the astyle analysis, it forces a weird style change in the initializers list, when you use braces and new keyword is part of the init expression (I changed that code to use parentheses instead).

eisnerd:
I merged over on https://github.com/eisnerd/Arduino/commit/f41e3a94a8a4834fc490e783198450420585646a, though what you probably want is a rebase, which I'm also happy to do. It's just trivial stuff with the switch from ICACHE_RAM_ATTR to IRAM_ATTR.

Not sure what the best way for libraries depending on Wire would be for accepting a TwoWire instance. If they stick with TwoWire, not TwoWireBase, you can't pass TwoWireMaster without building with ``-DTWOWIRE_MASTER_ONLY=true`` which for me hosed communication with several i2c devices. Not exactly sure why. They could switch to TwoWireBase if available (just ``#ifdef``?), but this arguably looks surprising to application developers. Could the base class be named TwoWire? This would avoid libraries needing to change. It would break a lot of application code that instantiated TwoWire, but that's presumably less as it would only affect esp8266 and code already using another TwoWire instance in some way.

michkot:
> I merged over on [eisnerd@f41e3a9](https://github.com/eisnerd/Arduino/commit/f41e3a94a8a4834fc490e783198450420585646a), though what you probably want is a rebase, which I'm also happy to do. It's just trivial stuff with the switch from ICACHE_RAM_ATTR to IRAM_ATTR.

I will look into merging onto latest;

> 
> Not sure what the best way for libraries depending on Wire would be for accepting a TwoWire instance. If they stick with TwoWire, not TwoWireBase, you can't pass TwoWireMaster without building with `-DTWOWIRE_MASTER_ONLY=true` which for me hosed communication with several i2c devices. Not exactly sure why. They could switch to TwoWireBase if available (just `#ifdef`?), but this arguably looks surprising to application developers. Could the base class be named TwoWire? This would avoid libraries needing to change. It would break a lot of application code that instantiated TwoWire, but that's presumably less as it would only affect esp8266 and code already using another TwoWire instance in some way.

Hmm.. its some time since I wrote this, I will have to check my own usage. The compatibility problem gave me a huge headache at that time :D

michkot:
> If they stick with TwoWire, not TwoWireBase, you can't pass TwoWireMaster

Could you maybe just (unsafely) down-cast the wire master-only instance at the place of passing it to library, e.g. `static_cast<TwoWire&>(yourWireMasterInstance)`?
... hmm now Master and MasterOrSlave are "siblings", not in super/sub class relation, so a double-cast would be needed `static_cast<TwoWire*>(static_cast<TwoWireBase*>(&twm)),` or the plain old `reinterpret_cast<TwoWire*>(&twm) ` :-) . 
I will definitely also look into why I did not make them Base->Master->MasterOrSlave, likely because of the typedef/using hell.

PR #7914, opened 2021-03-10
Fix link to ESP8266 NonOS SDK

ksalwin:
Topic under this link does not exist.
Link should point to ESP8266 SDK repository on GitHub.

### DIFF
--- a/doc/esp8266wifi/readme.rst
+++ b/doc/esp8266wifi/readme.rst
@@ -7,7 +7,7 @@ ESP8266 is all about Wi-Fi. If you are eager to connect your new ESP8266 module 
 Introduction
 ------------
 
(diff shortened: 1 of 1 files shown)


d-a-v:
Thanks!
The correct link is https://github.com/espressif/ESP8266_NONOS_SDK/tree/release/v2.2.x

ksalwin:
Thank you for response.
Though I'd suggest to keep link _generic_ enough not to be concerned by changes of SDK in the future.
Specific version of SDK is already in [SDK version file](https://github.com/esp8266/Arduino/blob/master/tools/sdk/include/version.h). The approach of link to _root_ not to _label_ i've found also in [README.md](https://github.com/esp8266/Arduino/blob/master/README.md).

d-a-v (approved):
I hope it is soon that next change will be for a link pointing to RTOS-SDK :)

PR #7920, opened 2021-03-12
Add setSSLVersion call to SSL object

earlephilhower:
Allow users to only allow specific TLS versions for connections with an
additional call in their app, similar to the setCiphers call.

Fixes #7918

PR #7925, opened 2021-03-15
Adds the method setSSLVersions() also to WiFiClientSecure in order to…

manolodd:
I'm not able to use new features of setSSLVersions() unless I add the method also to WifiClientSecure. I guess this method is missing as I wrote in #7920.

Hope this helps!


earlephilhower (approved):

JerryLoveCat:
where is WiFiClientSecure.verify. it only exists in 2.7.4

PR #7932, opened 2021-03-19
Update Root Certificate

virtualmix:
The root certificate used in the example to connect to "api.github.com" has expired.
A new certificate was issued on March 3, 2021 and should be valid for a year.


earlephilhower (approved):
Thanks.

PR #7935, opened 2021-03-23
StreamConstPtr: fix doc + reading flash space byte-by-byte

### LINKED ISSUE #7928: Does ESP8266WebServer and ESP8266WiFi now require non32xfer to be enabled?
Note: copied from my comment in #6979 as per @earlephilhower suggestion.

Hello, I would like to confirm my founding. It seems that the updated ESP8266WebServer (and maybe ESP8266WiFi) relies on the non32xfer service because: 

`template <typename ServerType>`
`void ESP8266WebServerTemplate<ServerType>::send_P(int code, PGM_P content_type, PGM_P content) {`
`  StreamConstPtr ref(content, strlen_P(content));`
`  return send(code, String(content_type).c_str(), &ref);`
`}` 

-----

Then, because `PGM_P` is `const char*` then this CTOR will be called:
`StreamConstPtr(const char* buffer, size_t size)`
`: _buffer(buffer), _size(size), _byteAddressable(__byteAddressable(buffer)) { }`

which mean that the buffer will be read using non PGM method.

-----

Also, there is another CTOR:
`StreamConstPtr(const __FlashStringHelper* text)`
`: _buffer(reinterpret_cast<const char*>(text)), _size(strlen_P((PGM_P)text)), _byteAddressable(false) { }`

It casts `__FlashStringHelper` to `const char*`. While `strln_P()` is correctly called, the buffer will be read using non PGM method.

-----

Then `pgmspace.h` still defines `PSTR` as:

`#ifndef PSTRN`
`    #define PSTRN(s,n) (__extension__({static const char __pstr__[] __attribute__((__aligned__(n))) __attribute__((section( "\".irom0.pstr." __FILE__ "." __STRINGIZE(__LINE__) "."  __STRINGIZE(__COUNTER__) "\", \"aSM\", @progbits, 1 #"))) = (s); &__pstr__[0];}))`
`#endif`
`#ifndef PSTR`
`  #define PSTR(s) PSTRN(s,PSTR_ALIGN)`
`#endif`

So it seems that PSTR strings still need to be accessed using the `_P()` functions.

-----

I found that if I do not enable/include the non32xfer service then calls to web server `send_P()` will cause exception 3 (EXCCAUSE_LOAD_STORE_ERROR) inside the Arduino String class when copying the characters.

It seems also the case for SSL web server (even with non32xfer enabled, my SSL web server failed to send anything to the browser despite it does not cause exception 3). I am still unsure why this behavior with SSL happened.

I want to ask if this was by design (non32xfer must be enabled/included)?

I need to inform you that until at least the new official Arduino package is released, I will not be able to produce a test sketch because like I said to @earlephilhower in my other discussion, I do not use the Arduino build system.

Because I also follow updates from Arduino ESP8266 GIT and then modify the cores a bit to match my need and style, my application code will not compile using Arduino's original core. Therefore, I am also not sure if my comment about this is actually valid when using the original core and Arduino IDE. Thank you!

d-a-v:
Should fix #7928

In the following code
```
        //if (_byteAddressable)
        //    return _peekPointer < _size ? _buffer[_peekPointer++] : -1;
        return _peekPointer < _size ? pgm_read_byte(&_buffer[_peekPointer++]) : -1;
```
I wonder if it is worth uncommenting the two first lines, which are valid.

### DIFF
--- a/cores/esp8266/StreamDev.h
+++ b/cores/esp8266/StreamDev.h
@@ -179,12 +179,16 @@ public:
 
     virtual int read() override
     {
-        return _peekPointer < _size ? _buffer[_peekPointer++] : -1;
+        //if (_byteAddressable)
+        //    return _peekPointer < _size ? _buffer[_peekPointer++] : -1;
+        return _peekPointer < _size ? pgm_read_byte(&_buffer[_peekPointer++]) : -1;
     }
 
     virtual int peek() override
     {
-        return _peekPointer < _size ? _buffer[_peekPointer] : -1;
+        //if (_byteAddressable)
+        //    return _peekPointer < _size ? _buffer[_peekPointer] : -1;
+        return _peekPointer < _size ? pgm_read_byte(&_buffer[_peekPointer]) : -1;
     }
 
     virtual size_t readBytes(char* buffer, size_t len) override
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -441,7 +441,7 @@ Stream extensions
 
     Two additional classes are provided.
 
-    - ``StreamPtr::`` is designed to hold a constant buffer (in ram or flash).
+    - ``StreamConstPtr::`` is designed to hold a constant buffer (in ram or flash).
 
       With this class, a ``Stream::`` can be made from ``const char*``,
       ``F("some words in flash")`` or ``PROGMEM`` strings.  This class makes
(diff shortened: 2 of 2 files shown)


earlephilhower:
> I wonder if it is worth uncommenting the two first lines, which are valid.

I'd say no.  Code size increase and pgm_read_byte is only 4 insns extra, anyway, so performance vs. the virtual call overhead should be nothing.

earlephilhower (approved):
LGTM

PR #7936, opened 2021-03-23
MDNS.begin() conflict? Added isRunning() test first.

jaggzh:
Please make sure this one-line change is correct.
I had only one mdns service found in my scans.
If I'm understanding, a service (library) calling MDNS.begin() will cause a reset, so multiple libs can conflict.  This will test MDNS.isRunning() to ensure it's not called again.
This change has resolved my conflict and all libs' MDNS entries are now available.

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -122,7 +122,7 @@ void ArduinoOTAClass::begin(bool useMDNS) {
   
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_MDNS)
   if(_useMDNS) {
-    MDNS.begin(_hostname.c_str());
+    if (!MDNS.isRunning()) MDNS.begin(_hostname.c_str());
 
     if (_password.length()) {
       MDNS.enableArduino(_port, true);


d-a-v:
MDNS has a lot changed since 2.7.4.
Can you please try again with [latest git master](https://arduino-esp8266.readthedocs.io/en/latest/installing.html#using-git-version) / [PlatformIO staging version](https://docs.platformio.org/en/feature-test-docs/platforms/espressif8266.html#using-arduino-framework-with-staging-version) / [arduino board manager alpha version 0.0.1](https://d-a-v.github.io/esp8266/Arduino) ?

mcspr:
I think the MCVE is just `MDNS.enableArduino(8266); WiFi.begin(...);` in setup and loop with `MDNS.update()`. After connecting (or not), loop will crash b/c UDPContext object handling UDP stuff is missing somewhere in the `_sendMDNSMessage`. Running master ofc.

```cpp
#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <ESP8266mDNS.h>

void setup() {
    Serial.begin(115200);

    MDNS.enableArduino(8266);

    WiFi.persistent(false);
    WiFi.mode(WIFI_STA);
    WiFi.begin("SSID", "passpass");

    WiFi.waitForConnectResult();
}

void loop() {
    MDNS.update();
    delay(10);
}
```

begin() fixes this, like above, but idk if this is intended behavior?

mcspr:
:oops: I misread the topic, crash above is unrelated to the patch

only thing begin() does different by calling it twice is receiving 'st' interface update multiple times. replacing `enableArduino` line above with:
```cpp
    MDNS.begin("testtest");
    MDNS.addService("something", "tcp", 12345);

    MDNS.begin("testtest");
    MDNS.enableArduino(8266);
```
```
[MDNSResponder] new Interface 'st' is UP! restarting
[MDNSResponder] _allocUDPContext

[MDNSResponder] new Interface 'st' is UP! restarting
[MDNSResponder] _allocUDPContext
```
everything is announced correctly though

d-a-v:
Closing per last comment and per checking `mDNS::begin()`

PR #7939, opened 2021-03-25
WString: return bool instead of unsigned char

mcspr:
Clean up the intent, resulting assembly stays the same as far as I could tell (objdump + vimdiff)
Follow-up #7781

`if(s)` helper is kind of redundant though, since <s>`c_str()`</s>`buffer()` may never be nullptr and always points to something. Still needed to be compatible though, if something uses it


earlephilhower (commented):
Thanks, I really like the idea, @mcspr !

That said, I'm worried about changing the signature of something as basic as `String`.  Even today, in the ArduinoAPI repo `String.h` uses `unsigned char`.  So this change would move our interface away from that, and break anything inheriting from String.
https://github.com/earlephilhower/ArduinoCore-API/blob/9da53735171f7cccbe781c66d9ee1068d94d6ec8/api/String.h

So I'm honestly not sure which way to go here.  Let's discuss and see if we can get anyone else's input.

mcspr:
I may be missing the actual issue, but since this is not virtual inheritance methods do not depend on the base signature:
https://godbolt.org/z/x9b5be8b4
Removing the implementation will use the BaseX method implicitly, *but* if method signature is actually different one needs to do a `using BaseX::method;` to make it use the base method.

Also opened a PR to the ArduinoCore-API with a similar change

earlephilhower (approved):
Solid explanation.  I'm happy to give this a shot and see if we uncover some weirdness in ancient libs, since it looks like the odds are very low.

PR #7941, opened 2021-03-25
WiFi: clean up AP SSID setter & getter, support 32 chars

mcspr:
This was updated for STA some time ago.
Plus, clean up ssid & pass copy routines similar to the #7940 

Comparison also adjusted to take into an account the ssid_len and limit itself by the passphrase length, including '\0' if I read the strncmp man correctly.
Adds note in the comments that softAp only works with passphrases (...not WEP key. Terminology used is a bit confusing in regards to password vs. psk vs. passphrase)


libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -358,25 +367,19 @@ String ESP8266WiFiAPClass::softAPmacAddress(void) {
...
     struct softap_config config;
     wifi_softap_get_config(&config);
-    char* name = reinterpret_cast<char*>(config.ssid);
-    char ssid[sizeof(config.ssid) + 1];
-    memcpy(ssid, name, sizeof(config.ssid));
-    ssid[sizeof(config.ssid)] = '\0';
-
-    return String(ssid);
+    String ssid;
+    ssid.concat(reinterpret_cast<const char*>(config.ssid), config.ssid_len);
```
mcspr, line 371: Just to note, with this test API still returns non-zero length field when user sets ssid_len=0:
```diff
diff --git a/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp b/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
index 392e142e..9fa2e514 100644
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
@@ -123,7 +123,9 @@ bool ESP8266WiFiAPClass::softAP(const char* ssid, const char* passphrase, int ch
     if (ssid_len < 32) {
         conf.ssid[ssid_len] = '\0';
     }
-    conf.ssid_len = ssid_len;
+    conf.ssid_len = 0;
+    Serial.printf("conf.ssid_len=%hhu\n", conf.ssid_len);
+    //conf.ssid_len = ssid_len;

     if(!passphrase || pass_len == 0) {
         conf.authmode = AUTH_OPEN;
```
```cpp
#include <Arduino.h>
void setup() {
    Serial.begin(115200);
    auto timeout = esp8266::polledTimeout::oneShotFastMs(1000);
    while (!timeout) {
        delay(100);
    }

    WiFi.persistent(false);
    WiFi.mode(WIFI_AP);
    String ssid;
    while (ssid.length() < 31) {
        ssid += 'A';
    }
    String pass("testtesttest");

    WiFi.softAP(ssid, pass, 6);
    delay(100);
    
    softap_config config;
    wifi_softap_get_config(&config);
    Serial.printf("retrieved ssid_len=%hhu\n", config.ssid_len);
}
void loop() {
}
```
```
conf.ssid_len=0
retrieved ssid_len=31
```

arihantdaga:
@mcspr I am facing a problem where Calling `WiFi.softAP(SSIDS.c_str(), PASSS.c_str());` Isnt working, and not updating the ssid and password. 
It is only taking for the first time, next time if i upload code with different SSID and PASS, it is not updated(The wifi discovered by phone is still have the same older name). I even tried this pr, but still the same. 

Is it a related problem ? 
Sketch
```

#include <ESP8266WiFi.h>
#include <WiFiClient.h>
#include <ESP8266WebServer.h>


String SSIDS = "NewSSID"; // If next time i change this and upload the code again, ssid displayed doesnt actually change. 
String PASSS = "NewPassword";


/* Just a little test message.  Go to http://192.168.4.1 in a web browser
   connected to this access point to see it.
*/


void setup() {
  delay(1000);
  WiFi.persistent(false);
  Serial.begin(115200);
  Serial.println();
  Serial.print("Configuring access point...");
  /* You can remove the password parameter if you want the AP to be open. */
  WiFi.softAP(SSIDS.c_str(), PASSS.c_str());

  IPAddress myIP = WiFi.softAPIP();
  Serial.print("AP IP address: ");
  Serial.println(myIP);
 
}

void loop() {
  server.handleClient();
}
```

mcspr:
@arihantdaga not likely. If you are using Android, just restart the WiFi via toggle or Airplane mode, or just filter out by bssid if this is some app. Scan results are cached. e.g.
https://gist.github.com/mcspr/430ba0b302faf9497c9492a41bad7dd6
Will probably display both at the same time.

arihantdaga:
@mcspr . I am sure, i can verify this is some bug, not problem with my mobile, I have tried with laptop and I have tried setting up different wifi names etc. I startee seeing it a while ago, today I verified multiple times.

mcspr:
Open an issue? This is not related to the PR here.
Note the example above also toggles wifi on -> off -> on, which is to showcase the simulations display

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -99,13 +102,15 @@ bool ESP8266WiFiAPClass::softAP(const char* ssid, const char* passphrase, int ch
...
-    if(!ssid || strlen(ssid) == 0 || strlen(ssid) > 31) {
+    size_t ssid_len = strlen(ssid);
+    if(!ssid || ssid_len == 0 || ssid_len > 32) {
         // fail SSID too long or missing!
         DEBUG_WIFI("[AP] SSID too long or missing!\n");
         return false;
     }
 
-    if(passphrase && strlen(passphrase) > 0 && (strlen(passphrase) > 63 || strlen(passphrase) < 8)) {
+    size_t pass_len = strlen(passphrase);
```
d-a-v, line 112: passphrase nullptr case is checked after strlen()

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -358,25 +367,19 @@ String ESP8266WiFiAPClass::softAPmacAddress(void) {
...
     struct softap_config config;
     wifi_softap_get_config(&config);
     char* pass = reinterpret_cast<char*>(config.password);
-    char psk[sizeof(config.password) + 1];
-    memcpy(psk, pass, sizeof(config.password));
-    psk[sizeof(config.password)] = '\0';
-
-    return String(psk);
+    pass[sizeof(config.password) - 1] = '\0'; // it is impossible to set PSK through the Arduino API because we only support WPA2
+    return pass;
```
d-a-v, line 384: Can't the password be 64 bytes long without `\0` (like with the SSID ?)

mcspr: Not from the Arduino API as-is, pass_len > 63 fails the WiFi.softAP() call
Re-googling the WPA-Personal, my mistake here... passphrase limit is indeed 63, *but* HEX key is exactly 64

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
mcspr: do you mean to just use
```cpp
if(pass_len > 0 && (pass_len > 63 || pass_len < 8)) {
```
?

libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -85,50 +88,57 @@ static bool softap_config_equal(const softap_config& lhs, const softap_config& r
...
     if(!WiFi.enableAP(true)) {
         // enable AP failed
         DEBUG_WIFI("[AP] enableAP failed!\n");
         return false;
     }
 
-    if(!ssid || strlen(ssid) == 0 || strlen(ssid) > 31) {
-        // fail SSID too long or missing!
-        DEBUG_WIFI("[AP] SSID too long or missing!\n");
+    size_t ssid_len = strlen(ssid);
```
d-a-v, line 105: `ssid` was tested against `nullptr` before calling `strlen()`.
same with `psk` below.
I'm not sure why this is removed and how `strlen()` would behave.

mcspr: Depends on the strlen implementation, as it turns out. idk what it is though, it is from sdk since the newlib file is commented out.
It is good to know though that newlib is a basic `while (*str) { ++len; }` :/, as it will break with null

mcspr: updated with null check which sets ...\_len to 0

arihantdaga:
@mcspr  Thank you. I got it, the actual problem was because i was setting a password that was less than 8 characters long. 
Found this issue. https://github.com/xoseperez/espurna/issues/1151
Thank you.

d-a-v (approved):

earlephilhower (approved):

PR #7949, opened 2021-03-28
HW Serial swap and pin setting work only on a few pins

dok-net:
Change return type from void to bool on affected functions, return false on failure.


d-a-v (approved):
LGTM

devyte (approved):
Checked the AVR and Due cores: no signature clash: they don't have these methods.
LGTM!

devyte:
This is actually a breaking change due to signature change.

PR #7952, opened 2021-03-31
Improve optimistic_yield intervals for performance gain in sketches

dok-net:
The purpose of optimistic_yield is to be used in library functions that get called in tight loops from CONT, in order to avoid unnecessarily frequent yields to SYS, and the performance impact that entails. This PR tries to identify and fix a few places where very low intervals were used for no obvious purpose.


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -479,7 +479,7 @@ int WiFiClientSecureCtx::_run_until(unsigned target, bool blocking) {
   esp8266::polledTimeout::oneShotMs loopTimeout(_timeout);
 
   for (int no_work = 0; blocking || no_work < 2;) {
-    optimistic_yield(100);
+    optimistic_yield(1000);
```
dok-net, line 482: I am undecided on this one. I don't think a forced plain `yield()` is in order here, but I don't quite dare to shot for the 10000µs interval either. If anyone with better insight into what `_run_until` actually does could comment, please?

earlephilhower: This loop pumps the BSSL engine.  It really should be a small optimistic_yield because the way things work we may initially be able to go thru everything and get a result in 1 pass, in which case the smallest delay is best.  Or, we might need to wait for data to come in over the WiFi to feed in/out of the BSSL engine, in which case we want to `yield()` as there's nothing we can do.

I would not increase the optimistic to 1ms, as when it's waiting and you don't have a yield then you're really just wasting time and no progress can be made until that `yield()` happens and gets some data.

In fact, a `yield()` here would work as well, if you want to scrub out optimistic yields.

dok-net: I'm reverting this spot to the time-proven 100µs. Thanks for the comprehensive input.

cores/esp8266/flash_hal.cpp
```diff
@@ -41,7 +41,7 @@ int32_t flash_hal_read(uint32_t addr, uint32_t size, uint8_t *dst) {
 }
 
 int32_t flash_hal_write(uint32_t addr, uint32_t size, const uint8_t *src) {
-    optimistic_yield(10000);
+    optimistic_yield(1000);
```
earlephilhower, line 44: For here and the hal_read I'm not sure the original thinking behind the yield.  These operations are fixed duratin and IRQs/etc. are disabled 100% during their operation, so nothing happens during the actual transfer.  The transfer is, however, <1ms in time.

Given that, I'm not sure I'd increase the frequency of the yield in these two calls...

cores/esp8266/flash_hal.cpp
```diff
@@ -60,7 +60,7 @@ int32_t flash_hal_erase(uint32_t addr, uint32_t size) {
     const uint32_t sector = addr / SPI_FLASH_SEC_SIZE;
     const uint32_t sectorCount = size / SPI_FLASH_SEC_SIZE;
     for (uint32_t i = 0; i < sectorCount; ++i) {
-        optimistic_yield(10000);
+        optimistic_yield(1000);
```
earlephilhower, line 63: Here things are different because erasing is sloooow (many x the write or read times) and this call can do multiple sectors.  No issue decreasing yield time here, I suppose.

cores/esp8266/uart.cpp
```diff
@@ -509,7 +509,10 @@ uart_tx_fifo_full(const int uart_nr)
 static void
 uart_do_write_char(const int uart_nr, char c)
 {
-    while(uart_tx_fifo_full(uart_nr));
+    while(uart_tx_fifo_full(uart_nr))
+    {
+        optimistic_yield(1000UL);
+    }
```
earlephilhower, line 515: Busywaiting on a fifo full has a mean expectation time of (10/2) bits @ baud.  For 115.2k that's 4.34us. I think this is a no-op in general.  Even at 9600bps it's only 520us, so same conclusion.  So this will generally be a noop...

libraries/ESP8266WiFi/src/WiFiClient.cpp
```diff
@@ -256,7 +256,7 @@ int WiFiClient::available()
     int result = _client->getSize();
 
     if (!result) {
-        optimistic_yield(100);
+        optimistic_yield(10000);
```
earlephilhower, line 259: Here the user has asked if there's more data and the answer is no.  There can't be more data until a yield() happens, so by increasing the delay I think you'll decrease performance.

libraries/ESP8266WiFi/src/WiFiUdp.cpp
```diff
@@ -118,7 +118,7 @@ int WiFiUDP::available() {
     if (!result) {
         // yielding here will not make more data "available",
         // but it will prevent the system from going into WDT reset
-        optimistic_yield(1000);
+        optimistic_yield(10000);
```
earlephilhower, line 121: Same comment as above.  I think the existing //comments are somewhat misleading...the return from this call to `available` won't change but the context might actually have gotten some data due to letting the wifi/etc. stack run.

I'd make the delay consistent with the TCP one, though (100).

libraries/Wire/Wire.cpp
```diff
@@ -226,7 +226,7 @@ int TwoWire::available(void)
     {
         // yielding here will not make more data "available",
         // but it will prevent the system from going into WDT reset
-        optimistic_yield(1000);
+        optimistic_yield(10000UL);
```
earlephilhower, line 229: :) Here's some old cut-n-paste, I think.  Same comment as UDP and TCP.  Stick w/100.

cores/esp8266/core_esp8266_i2s.cpp
```diff
@@ -313,7 +313,7 @@ static bool _i2s_write_sample(uint32_t sample, bool nb) {
         if (tx->slc_queue_len > 0) {
           break;
         } else {
-          optimistic_yield(10000);
+          optimistic_yield(1000);
```
earlephilhower, line 316: Expectation wait time for a new buffer to be freed up by the HW DMA is (64/2) 32-bit samples at 44.1khz, so ~726us.

So the prior yield would very rarely occur, but this new one has a very high chance of hitting especially if the app has done other stuff before I2S write.

Because I2S is time critical (you underflow, you click and pop) I think increase the number of unpredictable calls to SYS context could increase issues and would not change the timeout here or in the other sections of I2S.

earlephilhower (changes requested):
As we discussed, I think the existing constants in optimistic_yield were chosen as a best-guess with no real "100% correct answer" possible.  So I looked at the reasoning as to why the oy was there in the 1st place and how the change would affect it (considering expectation value of wait times for busywait loops).

cores/esp8266/flash_hal.cpp
dok-net: But no compelling reason to do it, either - just slows down things even more, doesn't it? I'll revert to the previous 10000µs.

cores/esp8266/uart.cpp
dok-net: @earlephilhower Please look at this again from this perspective: `uart_do_write_char()` gets called from `uart_write_char()`, where before there is an `optimistic_yield(10000UL);` and `uart_write()`, where there is none, despite the latter being the longer running block write.
Might be good to do the same thing in both places. I'm going for 10000µs, perhaps it's better unconditionally after the register write in `uart_do_write_char()`, what do you think?

libraries/ESP8266WiFi/src/WiFiUdp.cpp
```diff
@@ -118,7 +118,7 @@ int WiFiUDP::available() {
     if (!result) {
         // yielding here will not make more data "available",
         // but it will prevent the system from going into WDT reset
-        optimistic_yield(1000);
+        optimistic_yield(100UL);
```
earlephilhower, line 121: (here and other spots)
In #6869 you'd suggested a happy medium of `10000` us.  I think it needs to be consistent here and elsewhere.  The `100` that's in these spots seems like it will always fire because the delay is so small, and because they're not "yield to the OS" but instead "don't WDT" it seems like we'd benefit from the longer offset (i.e. higher throughput).

cores/esp8266/uart.cpp
earlephilhower: OK, sounds reasonable.

cores/esp8266/core_esp8266_wiring_pulse.cpp
```diff
@@ -43,10 +50,16 @@ unsigned long pulseIn(uint8_t pin, uint8_t state, unsigned long timeout)
...
+    if (!waitForPinState(pin, !state, timeout_cycles, start_cycle_count)) {
+        return 0;
+    }
+    if (!waitForPinState(pin, state, timeout_cycles, start_cycle_count)) {
+        return 0;
+    }
     const uint32_t pulse_start_cycle_count = xthal_get_ccount();
-    WAIT_FOR_PIN_STATE(!state);
+    if (!waitForPinState(pin, !state, timeout_cycles, start_cycle_count)) {
+        return 0;
```
earlephilhower, line 61: This is wrong per my reading of the (not so hot docs)  https://www.arduino.cc/reference/en/language/functions/advanced-io/pulsein/?setlang=it

Return 0 on no pulse beginning edge detected.  OTW, it should return the full timeout as the pulse width  (i.e. get rid of the `return 0` and just fallthru tho the return below).

libraries/ESP8266WiFi/src/WiFiUdp.cpp
earlephilhower: Also, can we be consistent on the `UL` or not `UL`?  I think in this PR 80% of the calls have `UL`, but IIRC none in #6869 have `UL`.

earlephilhower (changes requested):

cores/esp8266/core_esp8266_wiring_pulse.cpp
dok-net: @earlephilhower Can you agree this year; the change is just same effective execution path as with the macros before?

PR #7953, opened 2021-03-31
Sanitizing HW serial's uart.cpp just a little more.

dok-net:


cores/esp8266/uart.cpp
```diff
@@ -832,14 +826,18 @@ uart_set_tx(uart_t* uart, int tx_pin)
     case UART0:
         if(uart->tx_enabled)
         {
-            if (uart->tx_pin == 1 && tx_pin == 2)
+            if (uart->tx_pin == tx_pin)
+            {
+                return true;
+            }
```
d-a-v, line 832: This is a fix, in the sense that this case was previously returning false, right ?

d-a-v (approved):
LGTM

cores/esp8266/uart.cpp
```diff
@@ -832,14 +826,18 @@ uart_set_tx(uart_t* uart, int tx_pin)
...
+            else if (tx_pin == 2)
             {
                 pinMode(uart->tx_pin, INPUT);
-                uart->tx_pin = 2;
+                uart->tx_pin = tx_pin;
                 pinMode(uart->tx_pin, FUNCTION_4);
                 return true;
             }
-            else if (uart->tx_pin == 2 && tx_pin != 2)
+            else if (tx_pin != 2)
```
devyte, line 840: Shouldn't this really be == 1? As it is (both master's code and this PR's code), a change with any pin different than the current will apparently succeed, but the real pin set is hardcoded to 1 below.

devyte (changes requested):

devyte (approved):

PR #7956, opened 2021-04-02
[BREAKING] wifi: remove pseudo-modes for shutdown, expose `::[resumeFrom]shutdown()`

mcspr:
coming from #7902 discussion:
https://github.com/esp8266/Arduino/pull/7902#discussion_r604525282

make `shutdown()` and `resumeFromShutdown()` public
removes extra code from the `mode()` related to shutdown state
include force sleep and shutdown method descriptions in the docs


d-a-v:
The pseudo modes are removed, so if it has to be merged it must be soon.
I understand the need to make the methods public.
I don't really mind with the pseudo-modes I added that you removed, they were marked as are experimental.
Most important is the goal: store the sleep&wake functionalities in the core API so they can be improved there for everyone.

doc/esp8266wifi/generic-class.rst
```diff
@@ -170,6 +148,40 @@ getPhyMode
...
+
+.. code:: cpp
+
+    bool  shutdown (uint32_t sleepUs, WiFiState* state)
+    bool  resumeFromShutdown (WiFiState* state)
+
+Stores the STA interface IP configuration in the specified state struct and calls ``forceSleepBegin(sleepUs)``.
+Restores STA interface configuration from the ``state`` and calls ``forceSleepWake()``.
+
+These method is intended to be used in low-power scenarios, e.g. where ESP.deepSleep is used between actions to preserve battery power. It is the user's responsibility to preserve the WiFiState between ``shutdown()`` and ``resumeFromShutdown()`` by storing it in the RTC user data and/or flash memory.
```
d-a-v, line 181: These methods *are*

d-a-v (approved):

doc/esp8266wifi/generic-class.rst
mcspr: thanks! also updated the double just above when describing forceSleepWake

dok-net (approved):

doc/esp8266wifi/generic-class.rst
```diff
@@ -170,6 +148,40 @@ getPhyMode
 
 Gets the WiFi radio phy mode that is currently set.
 
+forceSleepBegin
+~~~~~~~~~~~~~~~
+
+.. code:: cpp
+
+    bool  forceSleepBegin (uint32_t sleepUs=0)
+    bool  forceSleepWake ()
```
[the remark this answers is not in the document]
mcspr: already noticed in the https://github.com/esp8266/Arduino/pull/7956/commits/03aaa4f3b9373e8dc6bd6523e13bc17f0da7c0eb

d-a-v:
```cpp
bool shutdown(WiFiState* stateSave) { return shutdown(0, stateSave); }
```
Do you think this would simplify the clarity of the API and the examples ?

mcspr:
Maybe rotate the arguments?
```cpp
bool shutdown(WiFiState* state = nullptr, uint32 time = 0);
```
Since the default arguments are already there

d-a-v:
Anything that would make the API clearly readable at the first glance.

The same suggestion would arise anyway, if `shutdown(nullptr, 500000)` were to be used.
I'd then suggest
```cpp
bool shutdown(uint32_t us) { return shutdown(nullptr, us); }
```

d-a-v:
And I'd also suggest to rename `::shutdown()` to `::shutdown_us()` in both case.

mcspr:
btw, are we still operating under the assumption mode(WIFI_OFF) *will* turn off the modem in some patch before v3?
what is the point of shutdown_us() when there's forceSleepBegin() doing... the same thing?

Considering that, is there a point in default args and pointers then? Should it be
```cpp
bool mode(WiFiMode_t opmode);
bool forceSleepBegin(uint32_t time);
bool forceSleepWake();
bool shutdown(WiFiState& state);
bool shutdown(WiFiState& state, uint32_t whatForceSleepBeginGets);
```
?

State is not optional, time is

d-a-v:
>  are we still operating under the assumption mode(WIFI_OFF) will turn off the modem in some patch before v3?

I'm not sure. OFF is not the same state as modem-off and we still want to have the API for this mode right ?

> what is the point of shutdown_us() when there's forceSleepBegin() doing... the same thing?

`shutdown()` was private. Now it is exposed, let the API be coherent by unifying names. We can either
- call it `forceSleepBegin()` (keep current exposed name) and add optional state parameter
- remove default value in the first parameter

and in all case, I think it's good to add the time unit where it is needed, and allow to use the function with only one of the two parameters (time and state).

dok-net:
... `forceSleepBegin()` - functions ending in `Begin()` are reminiscent of the asynchronous programming model and I would like to ask such naming to be reserved to that, on any platform, please.

mcspr:
I don't really have a better argument than 'why not' :) There's no use for the modem, so might as well put it to sleep?
(btw also just noticed the wifi_fpm_auto_sleep_set_in_null_mode)

And I still don't really get how the time plays out with the shutdown(). fpm wake up callback is never set, thus how do we know when to restore the state if time is not infinite? Should there be a callback arg then as well? Or is it expected to check for the sleep? (...but there's no get sleep state method?)

So I guess this is the desired model / naming:
```
bool mode(WiFiMode_t);
bool forceSleep(uint32_t uSec);
bool forceWake();
bool shutdown(WiFiState& state);
bool resumeFromShutdown(WiFiState& state);
bool shutdown(WiFiState& state, uint32_t time, SomeCb cb); // state can't be null ptr, 
```

d-a-v:
@mcspr please go ahead !

dok-net:
@mcspr I've submitted PR #7979 that adds the part of sleep functionality that is not directly associated with WiFi connection setup and restore to the ESP class, among a few other things. I believe our code for that part is mostly identical.

I would like to propose that the `forceSleep()` and `forceSleepWake()` in the WiFi class, as begin implemented by this PR, make use of that new API. I feel that greater care needs to be taken to align with the language used in Espressif documentation, otherwise it adds to the confusion surrounding the various power efficiency modes. My hunch is that there is no such thing as a "forced sleep" without mentioning modem or light - where modem is actually a lighter sleep than "light" :-)

`shutdown` and `resumeFromShutdown` are not only useful for deep sleep, but also for forced light sleep, as opposed to modem sleep modes.

dok-net:
@mcspr For it all to make sense, I've completed the sleep support in PR #7979, ESP class, reviewed this PR and your issue #7975 such that I have not commited any regressions in my work, and would kindly like to ask you to review my PR and consider rebasing your work here on #7979. Any input is highly welcome by me. Thanks!

d-a-v (approved):
The API is cleaner, thanks !
(side note: the one-thing-to-learn-everyday was `strnlen()` for me)

PR #7957, opened 2021-04-02
Ticker callbacks must be in IRAM

dok-net:
Unless I'm mistaken, the Ticker callbacks are performed by an ISR, and belong into IRAM to prevent flash memory races.
IIRC, lambdas are only safe, that is, in IRAM, if they are without closures, while bind is OK.

### DIFF
--- a/libraries/Ticker/examples/TickerFunctional/TickerFunctional.ino
+++ b/libraries/Ticker/examples/TickerFunctional/TickerFunctional.ino
@@ -19,12 +19,12 @@ class ExampleClass {
     int _pin, _duration;
     Ticker _myTicker;
 
-    void classBlink() {
+    IRAM_ATTR void classBlink() {
       digitalWrite(_pin, !digitalRead(_pin));
     }
 };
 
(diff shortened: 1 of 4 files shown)


mcspr:
os_timer_arm is indeed scheduled through the timer ISR, but execution seems happens in the SYS context?

quoting sdk docs
> 3.1 Software timer
> Please note that os_timer APIs listed below are software timers executed in tasks, thus
> timer callbacks may not be precisely executed at the right time; it depends on priority. If you
> need a precise timer, please use a hardware timer which can be executed in hardware
> interrupt. For details please refer to hw_timer.c. 

and this reference code from Sming
https://github.com/SmingHub/Sming/blob/6bbcba9f9ef36f9fecc2099113d772b516be44c1/Sming/Arch/Esp8266/Components/driver/os_timer.cpp
> The timer alarm interrupt simply queues a task to handle the event.

devyte:
Previous comment is correct: the non-scheduled user callbacks are executed in SYS.

Issue #7903, opened 2021-03-02
Feature Request: simplify PolledTimeout namespace


d-a-v:
`PolledTimeout`'s namespace is large.

example: `esp8266::polledTimeout::periodicFastMs ledBlink(1000);`
It is, among other use-cases, aimed at replacing timed events calling `millis()` and is thus used at several places in the core and examples.
It is also the preferred solution proposed to users when they're facing delay issues in their main loop ("do not use delay(), prefer PolledTimeout, check the example").

There is at least [an aborted proposal](https://github.com/esp8266/Arduino/pull/7356#discussion_r436376219) to make it simpler with a mention to specific ideas for addressing this issue.

Kindly requesting to:
- remove the necessity of including `<something.h>` to use it (like with `millis()` and `delay()` which require nothing)
- reduce the namespace size, or remove it

d-a-v:
closing as wontfix.

PR #7960, opened 2021-04-04
new polledtimeout

devyte:
- Restructure polledtimeout
- simplify namespaces
- rename across code

### DIFF
--- a/cores/esp8266/PolledTimeout.h
+++ b/cores/esp8266/PolledTimeout.h
@@ -1,309 +1,85 @@
-#ifndef __POLLEDTIMING_H__
-#define __POLLEDTIMING_H__
+#ifndef __POLLEDTIMEOUT_H__
+#define __POLLEDTIMEOUT_H__
 
 
-/*
- PolledTimeout.h - Encapsulation of a polled Timeout
 
- Copyright (c) 2018 Daniel Salazar. All rights reserved.
- This file is part of the esp8266 core for Arduino environment.
-
- This library is free software; you can redistribute it and/or
- modify it under the terms of the GNU Lesser General Public
- License as published by the Free Software Foundation; either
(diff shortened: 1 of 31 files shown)


d-a-v:
Thanks !
Is it a strong breaking change or do you plan to add something like this, for users who would have made the effort to use it ?
```cpp
namespace esp8266
{
    using  polledTimeout = ::PolledTimeout [[ deprecated, "use PolledTimeout instead of esp8266::polledTimeout" ]];
}
```
This fixes #7903

devyte:
> Is it a strong breaking change or do you plan to add something like this, for users who would have made the effort to use it ?
> 
> ```c++
> namespace esp8266
> {
>     using  polledTimeout = ::PolledTimeout [[ deprecated, "use PolledTimeout instead of esp8266::polledTimeout" ]];
> }
> ```

I intend this to be a strong breaking change. I don't want the legacy naming to be dragged into the future.

dok-net:
@devyte The copyright header is AWOL in `PolledTimeout.h`, the three newly added files seem OK in this regard.

cores/esp8266/PolledTimeout.h
```diff
@@ -1,309 +1,85 @@
...
-#include <core_esp8266_features.h>
 
-namespace esp8266
+namespace PolledTimeout
 {
+  //Construction of types using the policies
 
+  // legacy type names, deprecated (unit is milliseconds)
+  using oneShot = TimeoutTemplate<false, YieldPolicy::DoNothing, TimePolicy::TimeMillis>;
+  using periodic = TimeoutTemplate<true, YieldPolicy::DoNothing, TimePolicy::TimeMillis>;
```
d-a-v, line 23: Why are these kept in this breaking change commit ?

PR #7978, opened 2021-04-12
Resolve "PWM-locked" / "phase-locked" waveform merge leftover in Servo lib

### LINKED ISSUE #7977: PWM-locked support not properly working after "PWM/Phase"-locked merge
There's a serious functionality, probably also performance, issue caused by the runtime selection patch that did away with the preprocessor #define based switch between "PWM locked" and "phase locked" waveform modes.

See `libraries/Servo/src/Servo.cpp`, in `Servo::attach`, `Servo::detach`, `Servo::writeMicroseconds` the conditional compilation cannot occur.

dok-net:
Fixes #7977.
@earlephilhower Please review carefully, there's a caveat for the PWM-locked mode that I can't decide on.


earlephilhower (changes requested):
Can you please add in an check for `|| (timeHighCycles==0)` in :
https://github.com/esp8266/Arduino/blob/209e467120ad0cd84db6bdf8dadd93ea55924f8c/cores/esp8266/core_esp8266_waveform_pwm.cpp#L359

libraries/Servo/src/Servo.cpp
```diff
@@ -95,11 +90,8 @@ void Servo::detach()
 {
   if (_attached) {
     _servoMap &= ~(1 << _pin);
-#ifdef WAVEFORM_LOCKED_PHASE
+    // TODO - timeHigh == 0 is illegal in _PWM code branch. And now what?
```
earlephilhower, line 93: Sorry, forgot to remind you do drop this comment.  IT's done w/your last commit.

earlephilhower (changes requested):

libraries/Servo/src/Servo.cpp
dok-net: oh sure, my bad

earlephilhower (approved):

Issue #7055, opened 2020-02-02
Implement low power mode api


devyte:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

### Platform

- Hardware:					All
- Core Version:			latest git today
- Development Env:		All
- Operating System:		All

### Problem Description

PR #6989 provides a low power usage reference example, and it has been merged. Per [@d-a-v 's comment](https://github.com/esp8266/Arduino/issues/6642#issuecomment-575822115), it should be investigated how to unify the current experimental pseudo modes with the reference code in the example, and how to provide a consistent api to the users. Once that api is available, the example in #6989 should be migrated to make use of that api without changing the current behavior of the tests in that example sketch. 

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

See the example merged in #6989.

devyte:
CC @Tech-TX

devyte:
Related to discussion in #6642

Tech-TX:
(laughs) That's not exactly an MCVE, but it'll do.  Timed Light Sleep (timer or GPIO wake) and Forced Light Sleep (GPIO wake) would benefit from API wrappers; the rest of the modes are already encapsulated.  Both of those Light Sleep modes need the WiFi off, so it'd be more friendly if it were an ESP API and not a WiFi API.  Similarly, an ESP API for Forced Modem Sleep would be nice if you're not using WiFi (the last section in the README).

Additionally I'm still researching the low-power boot to see if I can get that working again.  I saw something last night in [another thread](url=https://github.com/esp8266/Arduino/issues/6007#issuecomment-542235707) that doesn't solve it exactly, but it certainly helps.  It does _something weird_ to the modem and current only peaks at 53mA then drops to 30mA, but I can't drop it to the 15-18mA 'Forced Modem Sleep' range after that's been done (yet).  It's zeroing something undocumented in the RTC to _mostly disable_ the modem.  On the plus side, the current only hits ~53mA for 650us which is filterable by an electrolytic cap.

[...]

PR #7979, opened 2021-04-13
WiFi default is off now, don't force sleep in setup() of examples

dok-net:
@d-a-v Do you think it's safe to remove the call to `WiFi.forceSleepBegin();`, in fact any reference to WiFi, for the purpose of a terse example?


dok-net:
Fixes #7055 - thanks to @Tech-TX for pointing this out.

cores/esp8266/Esp.cpp
```diff
@@ -140,6 +140,117 @@ uint64_t EspClass::deepSleepMax()
...
+    wifi_fpm_set_wakeup_cb([]() {
+        if (saved_wakeupCb) {
+            saved_wakeupCb();
+            saved_wakeupCb = nullptr;
+        }
+        esp_schedule();
+        });
+    {
+        esp8266::InterruptLock lock;
+        saved_timer_list = timer_list;
```
mcspr, line 212: Why grab the timer list from the SDK?

cores/esp8266/core_esp8266_phy.cpp
```diff
@@ -288,20 +289,10 @@ static const uint8_t ICACHE_FLASH_ATTR phy_init_data[128] =
...
 
-
-// These functions will be overriden from C++ code.
-// Unfortunately, we can't use extern "C" because Arduino preprocessor
-// doesn't generate forward declarations for extern "C" functions correctly,
-// so we use mangled names here.
-#define __get_adc_mode _Z14__get_adc_modev
-#define __get_rf_mode _Z13__get_rf_modev
-#define __run_user_rf_pre_init _Z22__run_user_rf_pre_initv
-
```
mcspr, line 299: Does removal mean the comment is addressed for the sketch .ino having  `void __get_adc_mode() { return myvalue; }` func that is without the `extern "C"`?

cores/esp8266/Esp.cpp
```diff
@@ -140,6 +140,117 @@ uint64_t EspClass::deepSleepMax()
...
+    auto ret_do_sleep = wifi_fpm_do_sleep(duration_us);
+    if (ret_do_sleep != 0)
+    {
+#ifdef DEBUG_SERIAL
+        DEBUG_SERIAL.printf("core: error %d with wifi_fpm_do_sleep: (-1=sleep status error, -2=force sleep not enabled)\n", ret_do_sleep);
+#endif
+        return false;
+    }
+    // SDK turns on forced modem sleep in idle task
+#ifdef HAVE_ESP_SUSPEND
```
mcspr, line 172: Why mix other PR code when it's not yet in the Core?

cores/esp8266/Esp.h
```diff
@@ -98,6 +95,25 @@ class EspClass {
         void deepSleepInstant(uint64_t time_us, RFMode mode = RF_DEFAULT);
         uint64_t deepSleepMax();
 
+        bool forcedModemSleep(uint32_t duration_us = 0, void (*wakeupCb)() = nullptr);
```
mcspr, line 98: `using WakeupCb = void(*)()`?

cores/esp8266/Esp.cpp
devyte: I agree. The yield/delay will be getting a PR from me, including renaming.

cores/esp8266/Esp.cpp
devyte: This is necessary, it's the only known way to get the mode to work.

mcspr: ...then there should be comment as to why it is saved? do timers expire incorrectly since rtc is stopped, are they not resumed, or does sdk simply kills them all? original issue only provides some suspicions of the problem, not some specific example when it breaks (or what breaks, exactly)

cores/esp8266/core_esp8266_phy.cpp
dok-net: These functions are now properly prototyped in `coredecls.h`, as such, yes, external C linkage.

dok-net: non-void `int __get_adc_mode();` - smile ;-)

cores/esp8266/Esp.cpp
dok-net: @devyte why start new work instead of reviewing PR #7148?

cores/esp8266/Esp.h
dok-net: There is an existing type for this, see same function in Esp.cpp, plus void(*)() is trivial.

mcspr: general reasoning - this is more verbose. plus, it's repeated twice and it's the same callback both times.
(*mega pedantic note*: `using` is just a code hint aka alias, not something enforcing a specific type as it would with `struct { ... }`. So, just IMO, I'd prefer to have some distinct name rather than re-using something else)

dok-net: But your `using WakeupCb = void(*)();`suggestion relies upon no other definition for `WakeupCb` preexisting in the given namespace or there would a name clash, isn't that right? I respect your opinion, but I am not going to agree .

mcspr: `fpm_wakeup_cb`?

dok-net: Typedefing pointers is considered bad practice in general, even worse for function pointers. Opinions may vary, but I repeat that the function equivalent of 'void*' does not deserve an alias, this is of no utility to anyone. Please stop this.

dok-net: Seeing you probably suggested exposing the Expressif API includes, this is undesirable by breaking the concept of encapsulation/information hiding. For  a void pointer in particular not warranted.

Tech-TX:
I've been doing some testing for Dirk for a week or two, and have re-verified the correct operation of this last commit 637f5d6.  The timers are correctly re-attached after Forced Light Sleep, which I never could get to.  Thanks, Dirk!  The three examples function as expected, and all power saving modes hit the right current levels.  No surprises, no warts that I've seen.  For people using servos I'd stop them _before_ Forced Light Sleep, else they may slam the stops when PWM halts.  You can restart them afterwards.

In order to maintain WiFi across Forced Light Sleep you **must** sleep the modem cleanly before ESP.forcedLightSleepBegin().  I tried it the dirty way and WiFi wouldn't recover after Sleep.
I tested WiFi.forceSleepBegin(0xFFFFFFF) (sleep forever) before putting it to sleep, and WiFi.forceSleepWake() after it had recovered from ESP.forcedLightSleepEnd().  No problems restoring the connection that way.

I need to revisit https://github.com/esp8266/Arduino/pull/6989 and clean that program up with the new API calls.

Tech-TX:
Utterly side comment, not required for this PR: I think I just figured out how the NodeMCU folks implemented something like 'wake from Forced Light Sleep with a CHANGE level interrupt, not just the HI LEVEL and LO LEVEL specified in the API Reference.

If you read the pin level of the interrupt pin _just_ before going to Forced Light Sleep and set the interrupt to the opposite level, it accomplishes the same thing.  Going in with a HIGH on the pin, set the interrupt to LOW and vice versa.  Then any change during Forced Light Sleep wakes it, and the user doesn't care what state the pin had been in immediately prior to Sleep; any change wakes it up.

cores/esp8266/core_esp8266_phy.cpp
```diff
@@ -314,19 +305,43 @@ extern int IRAM_ATTR __wrap_spi_flash_read(uint32_t addr, uint32_t* dst, size_t
...
+extern int __get_rf_powerup_disable_mode(void) __attribute__((noinline, weak));
+extern int __get_rf_powerup_disable_mode(void)
+{
+    // Starting from arduino core v3: wifi is disabled at boot time
+    // mode == 2: RF initialization only calibrates VDD33 which will take about 2 ms;
+    // this has the least current consumption.
+    return 2;
+}
+
+extern int __get_rf_mode(void)  __attribute__((noinline, weak));
```
mcspr, line 326: Are these numbers mentioned somewhere in the docs, or is the rtos sdk the source of these?
(...citation needed?)

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -340,10 +340,7 @@ extern "C" void __disableWiFiAtBootTime (void)
 {
     // Starting from arduino core v3: wifi is disabled at boot time
     // WiFi.begin() or WiFi.softAP() will wake WiFi up
-    wifi_set_opmode_current(0/*WIFI_OFF*/);
-    wifi_fpm_set_sleep_type(MODEM_SLEEP_T);
-    wifi_fpm_open();
-    wifi_fpm_do_sleep(0xFFFFFFF);
+    ESP.forcedModemSleep();
```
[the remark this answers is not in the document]
dok-net: Like so #8006?

cores/esp8266/Esp.cpp
dok-net: > problem created by the #7148 in the first place

Please, no problem created, improvement/fix provided instead.
Give me a break, I'm testing all my PRs in a local global merge and I need this - I can well revert these particular #ifdefs just before merge to master when time comes, thank you.

cores/esp8266/Esp.cpp
```diff
@@ -140,6 +140,168 @@ uint64_t EspClass::deepSleepMax()
...
+    wifi_fpm_close();
+    saved_sleep_type = wifi_get_sleep_type();
+    wifi_set_sleep_type(LIGHT_SLEEP_T);
+}
+
+void EspClass::autoSleepOff() {
+    wifi_set_sleep_type(saved_sleep_type);
+    saved_sleep_type = NONE_SLEEP_T;
+}
+
```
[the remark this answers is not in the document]
dok-net: Please elaborate further.

cores/esp8266/core_esp8266_phy.cpp
Tech-TX: That's 3.3.12. system_phy_set_powerup_option from the API Reference.

doc/libraries.rst
```diff
@@ -71,12 +71,26 @@ An ESP8266 port of SoftwareSerial library done by Peter Lerup (@plerup) supports
...
+
+``ESP.forcedModemSleepOff()`` immediately returns the chip to the automatic sleep mode in effect before the preceeding call to ``ESP.forcedModemSleep``.
+
+``ESP.forcedLightSleepBegin(microseconds, callback)`` works in tandem with ``ESP.forcedLightSleepEnd(cancel)`` to put the chip into forced ``LIGHT_SLEEP``. A microseconds duration after which the sleep mode returns to the automatic sleep mode that was effective before this call can be given, a value of 0 or 0xFFFFFFF turns off that timeout. The optional callback function will be invoked when the forced light sleep ends. Forced light sleep halts the CPU, in addition to the timeout, it can be awakened via GPIO input. Between the calls to ``ESP.forcedLightSleepBegin`` and ``ESP.forcedLightSleepEnd`` any GPIOs except GPIO16 to use for wakeup can be set up.
+
+``ESP.forcedLightSleepEnd(cancel)`` causes the chip to enter the forced light sleep mode that was set up as described for ``ESP.forcedLightSleepBegin``. The optional cancel argument, if true, prevents the sleep mode transition from occuring and returns to the automatic sleep mode that was effective before ``ESP.forcedLightSleepBegin``.
```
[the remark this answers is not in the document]
dok-net: Done, I think, thanks for pointing out what was not easy to understand.
> Light, modem fpm sleep somehow change wifi driver sleep setting?

Please be more specific, what is the nature of your question?

cores/esp8266/core_esp8266_phy.cpp
dok-net: The latter. Please propose how to do this.

doc/libraries.rst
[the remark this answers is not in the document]
Tech-TX: >Since you have already tested this... does this mean:
    light sleep is entered at the end of SYS idle task
    light sleep is entered immediately with wifi_fpm_do_sleep()

Forced Light Sleep is entered sometime _during_ the idle task, and it's a fuzzy number between 2 and 8ms.  Most chips entered Sleep right around 2ms, but I had one chip that needed 8 myself last year.  In my version I bumped that idle to 10ms in case there was a chip that needed even longer than the one weird chip I had.  I didn't want to increase it too much as the whole point of Forced Light Sleep is that it wakes much faster than Deep Sleep, albeit at a much higher power consumption during Sleep (460uA vs 20uA for Deep Sleep).

cores/esp8266/Esp.cpp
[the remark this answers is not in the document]
dok-net: @mcspr I've researched based on your open-ended question and I can honestly say that until now, we haven't measured the effects of the SDK version 3 calls,  that `setSleepMode` and friends provide hints for. Plus I honestly have no opinion.
> Why sync this up with forced sleep though, since both do different things either when WiFi is ON or OFF?

I don't understand why this is a question and not an expression of opinion, please explain what you prefer and be specific.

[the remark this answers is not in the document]
dok-net: The APIs in question are `wifi_.et_sleep_level()` and `wifi_.et_sleep_interval()` (regex notation) . They were updated in this repo by 8ef21ca3aeff76d1265a72ac106999ed9c631a7e on 2018-10-09, which speaking in release numbers places them in core version 2.5.0 and onward. I don't know enough about SDK version numbering to make any sense of v2.2 in the commit message, the online SDK docs, but `NONOSDK3V0` in the code's `#ifdef`s, but there was #5919 .
I don't know that there is a need for all this mixing and matching via preprocessor defines years back into the past going to core milestone 3.0.0, so perhaps those code variants could be cleaned up, I'm sure @d-a-v  is the right person to get involved in deciding that.

Also, it's getting out of control for a single PR. I'm glad for the hints, we are investigation the effects on power from altering sleep level and interval for the ESP class functions proposed here, but any discussion of further integration with ESP8266WiFi should probably occur in a separate thread dedicated to that library.

cores/esp8266/core_esp8266_phy.cpp
dok-net: I can't really find a pre-existing and successful pattern for this in the sources' comments - sometimes SDK references end up pointing at documents that are no longer there - or the one's in the right place (Espressif server?) make no mention of what previously people were referring to. So, "the numbers" are explained in this code section and that's it :-)

cores/esp8266/Esp.cpp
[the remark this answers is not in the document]
mcspr: Perhaps... #7956 is still on hold though

This is not the question about the effects though, but the API usage and why it is used that way. Right now there is a single value storing both WiFi sleep and FPM sleep, which may be saved / restored at different points throughout the ESP method calls. Why set / get both interchangeably? Shouldn't there be a separate value for fpm and another one for modem sleep?

cores/esp8266/core_esp8266_phy.cpp
mcspr: Direct quote of the 3.3.12?
Window to fetch 2.2.2 docs from espressif is already lost (?), but I wonder why that document was not saved as a reference directly in the repo

dok-net:
With latest commit, completed migration of ESP8266Wifi recently PR'ed to the new Esp class API proposed here.

dok-net:
Reinstated `#define RF_PRE_INIT()` due to its use in https://github.com/esp8266/Arduino/issues/8783

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+void EspClass::forcedModemSleepOff()
+{
+    const sleep_type_t sleepType = wifi_fpm_get_sleep_type();
+    if (sleepType != NONE_SLEEP_T) {
+        if (sleepType == MODEM_SLEEP_T) wifi_fpm_do_wakeup();
+        wifi_fpm_close();
+    }
+    wifi_fpm_set_sleep_type(saved_sleep_type);
+    saved_sleep_type = NONE_SLEEP_T;
+}
```
mcspr, line 187: Combine with begin() as a single operation?
If begin() called once and then off(), sort-of works (if fpm was not enabled externally)
If begin() called once and then again, sleep mode gets overwritten

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+#ifdef DEBUG_SERIAL
+        DEBUG_SERIAL.printf("core: error %d with wifi_fpm_do_sleep: (-1=sleep status error, -2=force sleep not enabled)\n", ret_do_sleep);
+#endif
+        return false;
+    }
+    // SDK turns on forced modem sleep in idle task
+#ifdef HAVE_ESP_SUSPEND
+    esp_delay(10);
+#else
+    delay(10);
```
mcspr, line 173: Are we waiting for certain condition to happen? Should it check something instead of assuming 10ms delay does something?
also not necessary ifdef

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+    if (!duration_us || duration_us > 0xFFFFFFF) {
+        duration_us = 0xFFFFFFF;
+    }
+    wifi_fpm_close();
+    saved_sleep_type = wifi_fpm_get_sleep_type();
+    wifi_set_opmode(NULL_MODE);
+    wifi_fpm_set_sleep_type(MODEM_SLEEP_T);
+    wifi_fpm_open();
+    saved_wakeupCb = nullptr;
+    if (wakeupCb) wifi_fpm_set_wakeup_cb(wakeupCb);
```
mcspr, line 160: > fpm_wakeup_cb_func will be called after system wakes up only if the force sleep
time out (wifi_fpm_do_sleep and the parameter is not 0xFFFFFFF).
fpm_wakeup_cb_func will not be called if wake-up is caused by
wifi_fpm_do_wakeup from MODEM_SLEEP_T type force sleep.

Also, perhaps, not necessary to expose to user? Isn't return signal enough of a result?

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+            saved_wakeupCb = nullptr;
+        }
+        esp_schedule();
+        });
+#ifdef DEBUG_SERIAL
+    walk_timer_list();
+#endif
+    {
+        esp8266::InterruptLock lock;
+        saved_timer_list = timer_list;
```
mcspr, line 249: -//- as modem - if we hide timer list, it should not be possible for user to overwrite this accidentally. Single operation instead of separate begin() -> end()?

mcspr, line 249: Repeating the q from above - what is the difference between GPIO-activated sleep and timed one here? Is CPU really sleeping or we just simulate that it does?

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+        timer_list = nullptr;
+    }
+    return wifi_fpm_do_sleep(duration_us) == 0;
+}
+
+void EspClass::forcedLightSleepEnd(bool cancel)
+{
+    if (!cancel) {
+        // SDK turns on forced light sleep in idle task
+#ifdef HAVE_ESP_SUSPEND
```
mcspr, line 259: > HAVE_ESP_SUSPEND

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+        esp8266::InterruptLock lock;
+        timer_list = saved_timer_list;
+    }
+    saved_wakeupCb = nullptr;
+    wifi_fpm_close();
+    wifi_fpm_set_sleep_type(saved_sleep_type);
+    saved_sleep_type = NONE_SLEEP_T;
+    if (cancel) {
+        // let the SDK catch up in idle task
+#ifdef HAVE_ESP_SUSPEND
```
mcspr, line 278: > HAVE_ESP_SUSPEND

cores/esp8266/Esp.cpp
dok-net: Your opinion may differ, even for a reason, but I implement to the public specification and as my comment `// SDK turns on forced modem sleep in idle task` explains, there shall be a delay for the idle task to take over.
Thanks for picking up the quite dated HAVE_ESP_SUSPEND, it's been so long, I don't even recall where and when it was ever defined.

cores/esp8266/Esp.cpp
dok-net: If there's no hurt in exposing it as an option, let's keep it this way, please.

cores/esp8266/Esp.cpp
```diff
@@ -138,6 +138,186 @@ uint64_t EspClass::deepSleepMax()
...
+    if (!duration_us || duration_us > 0xFFFFFFF) {
+        duration_us = 0xFFFFFFF;
+    }
+    wifi_fpm_close();
+    saved_sleep_type = wifi_fpm_get_sleep_type();
+    wifi_set_opmode(NULL_MODE);
+    wifi_fpm_set_sleep_type(LIGHT_SLEEP_T);
+    wifi_fpm_open();
+    saved_wakeupCb = wakeupCb;
+    wifi_fpm_set_wakeup_cb([]() {
```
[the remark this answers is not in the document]
dok-net: All things the same, same answer from me?

cores/esp8266/Esp.cpp
dok-net: This way, it worked/works, other ways, it didn't work/does work, i.e. doesn't enter the power saving modes properly. It was all carefully measured by the other guy at the time. Whatever happened to him :-(

> Single operation instead of separate begin() -> end()?

I don't understand that sentence, I am sorry.

dok-net: OMG, you are asking questions that I cannot remember the answer to, except that I am 100% certain that we literally did hours of repetitive tests to show that power consumption dropped to the expected levels, and it didn't or the MCU became unreliable or the tables got corrupted if it wasn't done in exactly this - documented by Espressif - way. We really should not change a single line.

cores/esp8266/Esp.cpp
dok-net: Done.

cores/esp8266/Esp.cpp
dok-net: Done, too.

cores/esp8266/Esp.cpp
mcspr: > This way, it worked/works, other ways, it didn't work/does work, i.e. doesn't enter the power saving modes properly. It was all carefully measured by the other guy at the time. Whatever happened to him :-(

Sadly, yeah, might be difficult to test at this point from our baseline of knowledge :/
As I understand it, we have issue where SDK **processes** timer list. So, that is what we try to avoid in code. If there is an ets_post to a task, does it still work? Are timers waking up  early, or SDK never sleeps / idles to begin with?

> I don't understand that sentence, I am sorry.

Something like InterruptLock. Imagine FpmThingy class that enables sleep on begin or construction, and disables on destruction. It may or may not be applicable to our internal usage, but seems like a better API for user side of things. Plus, it would not be necessary to expose too much methods to control it

mcspr: Right, but now we have a half-understood feature that maybe works?
Reading our issues simply repeats the above - we copy nodemcu and test how it did. Would have to read that yet again if you don't have an explanation :/
(https://github.com/nodemcu/nodemcu-firmware/pull/1231#issuecomment-314497558 and below)

cores/esp8266/Esp.cpp
dok-net: Any calls to the SDK or modifying sleep modes by sidelining this API is undefined behavior. Simple as that.
The `saved_sleep_type`, if that is what you are focused on, is rather a convenience than a full-blown requirement for this to be useful. I don't see any real malfunction if its used the way you described.

cores/esp8266/Esp.cpp
dok-net: @mcspr I have added an RAII class and modified `libraries\esp8266\examples\ForcedLightSleep\ForcedLightSleep.ino`. Please review that and let me know if that matches what you had in mind. Thanks for the suggestion!

dok-net: I'm not sure we are not talking past each other. The sleep isn't actived by the GPIO, but cancelled. Same for the timer. I'm sure that's what you meant. As far as it got measure two years ago, yes we were sure it did as advertised - sleeping to save power, resuming either by GPIO trigger or timeout.

dok-net: @mcspr Is this now stalled again due to to unrelated workload, or have my changes and explanations been unconvincing so far? I would be totally relieved if this finally got merged, I am sure it will help people a lot, at least it might allow more specific reports about power-saving mode issues?

PR #7993, opened 2021-04-18
Fixes #7988: reverse dependency core on libraries: `enableWiFiAtBootTime()`

### LINKED ISSUE #7988: Arduino library detection works differently than "enableWiFiAtBootTime()" expects
PR #7902 suggests that only adding one call, `enableWiFiAtBootTime()`, is sufficient to restore previous behavior. This is underlined by the commit log:
> * mv enableWiFiAtBootTime() to core_esp8266_features.h

suggesting that no additional includes from any library, in particular `#include <ESP8266WiFi.h>` is required for this to work.

Unfortunately, this is not how the Arduino library resolver works, at least in my installation.

MCVE does not pass the link phase:
```
void setup()
{
#ifdef WIFI_IS_OFF_AT_BOOT
	enableWiFiAtBootTime();
#endif
}

void loop()
{
}
```
Result:
```
Compiling 'mcve' for 'LOLIN(WEMOS) D1 mini Pro'
ld.exe: mcve.cpp.o:(.text.setup+0x0): undefined reference to enableWiFiAtBootTime
 
ld.exe: mcve.cpp.o: in function setup
mcve.ino:4: undefined reference to enableWiFiAtBootTime
 
collect2.exe*: error: ld returned 1 exit status

Error linking for board LOLIN(WEMOS) D1 mini Pro
Build failed for project 'mcve'
```

Adding `#include <ESP8266WiFi.h>` helps, but in that case please
```
--- a/cores/esp8266/core_esp8266_features.h
+++ b/cores/esp8266/core_esp8266_features.h
@@ -105,8 +105,6 @@ uint64_t micros64(void);
 void delay(unsigned long);
 void delayMicroseconds(unsigned int us);
 
-void enableWiFiAtBootTime (void) __attribute__((noinline));
-
 #if defined(F_CPU) || defined(CORE_MOCK)
 #ifdef __cplusplus
 constexpr
```
and
```
--- a/libraries/ESP8266WiFi/src/ESP8266WiFi.h
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFi.h
@@ -53,6 +53,7 @@ extern "C" {
 #define DEBUG_WIFI(...) do { (void)0; } while (0)
 #endif
 
+extern "C" void enableWiFiAtBootTime (void) __attribute__((noinline));
 
 class ESP8266WiFiClass : public ESP8266WiFiGenericClass, public ESP8266WiFiSTAClass, public ESP8266WiFiScanClass, public ESP8266WiFiAPClass {
     public:
```
because compiler errror messages are usually more easily understood by users than linker errors.

In the line of business finding: The `#define WIFI_HAS_EVENT_CALLBACK` from commit d9a7a816d51535b21fec38cb5db30cc8511075c3 seems to have no known use, as such, could it be (re)moved, because it intertwines ESP8266WiFi with the core without need?

dok-net:
Fixes #7988

### DIFF
--- a/cores/esp8266/core_esp8266_features.h
+++ b/cores/esp8266/core_esp8266_features.h
@@ -105,8 +105,6 @@ uint64_t micros64(void);
 void delay(unsigned long);
 void delayMicroseconds(unsigned int us);
 
-void enableWiFiAtBootTime (void) __attribute__((noinline));
-
 #if defined(F_CPU) || defined(CORE_MOCK)
 #ifdef __cplusplus
 constexpr
--- a/doc/esp8266wifi/generic-class.rst
+++ b/doc/esp8266wifi/generic-class.rst
@@ -61,7 +61,7 @@ linker).
 
 .. code:: cpp
 
-    #include <ESP8266WiFi.h>    // necessary to avoid a linker error
+    #include <ESP8266WiFi.h>
 
     void setup () {
     #ifdef WIFI_IS_OFF_AT_BOOT
--- a/libraries/ESP8266WiFi/src/ESP8266WiFi.h
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFi.h
@@ -53,6 +53,7 @@ extern "C" {
 #define DEBUG_WIFI(...) do { (void)0; } while (0)
 #endif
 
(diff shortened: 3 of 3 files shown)


devyte (approved):

d-a-v:
Thanks

PR #7996, opened 2021-04-22
Make `enablePhaseLockedWaveform()` visible from `Arduino.h`

### LINKED ISSUE #7976: enablePhaseLockedWaveform is hidden
@earlephilhower In order to make default PWM generator selection in sketch and library source code properly available, it should be exposed without additional internal includes - Arduino.h or at least one of the commonplace includes for ESP8266 Core.

dok-net:
Move the prototype for `enablePhaseLockedWaveform()` linker magic into header that's included by default instead of particular internal core header.
Fixes #7976 .


earlephilhower (approved):
Thanks

PR #8008, opened 2021-04-29
Waveform: int/uint8_t inconsistency and implied stop PWM

dok-net:
@earlephilhower This is basically a cleanup, and I'm asking for you to review this, because it's your code.
I'm head scratching about the weak-ref magic you performed, it looks convoluted, but appears to be the only working variant - albeit I am worried that `enablePhaseLockedWaveform();` does not remove the PWM bits from the final BSP like it was with the previous config menu?


cores/esp8266/core_esp8266_waveform_pwm.cpp
```diff
@@ -96,7 +96,7 @@ static WVFState wvfState;
 static IRAM_ATTR void timer1Interrupt();
 static bool timerRunning = false;
 
-static __attribute__((noinline)) void initTimer() {
+static void initTimer() {
```
earlephilhower, line 99: The noinline attribute was added here to save a few bytes of program space...

cores/esp8266/core_esp8266_waveform_phase.cpp
```diff
@@ -308,7 +305,7 @@ static IRAM_ATTR void timer1Interrupt() {
   uint32_t now = ESP.getCycleCount();
   uint32_t isrNextEventCcy = now;
   while (busyPins) {
-    if (static_cast<int32_t>(isrNextEventCcy - now) > IRQLATENCYCCYS) {
+    if (static_cast<int32_t>(isrNextEventCcy - now) > IRQLATENCYCCYS + DELTAIRQCCYS) {
```
earlephilhower, line 308: Slight code change here.  Intended?

cores/esp8266/core_esp8266_waveform_pwm.cpp
```diff
@@ -157,9 +156,10 @@ static IRAM_ATTR void disableIdleTimer() {
 // Wait for mailbox to be emptied (either busy or delay() as needed)
 static IRAM_ATTR void _notifyPWM(PWMState *p, bool idle) {
   p->pwmUpdate = nullptr;
+  MEMBARRIER();
   pwmState.pwmUpdate = p;
   MEMBARRIER();
-  forceTimerInterrupt();
+  // The ISR pulls updates on the next PWM interval
```
earlephilhower, line 162: That could be 10 **milli**seconds away.  forceTimerInterrupt would make it happen in 10 **micro**seconds.  I think this needs to be undone.

cores/esp8266/core_esp8266_waveform_pwm.cpp
```diff
@@ -264,54 +262,56 @@ IRAM_ATTR bool _stopPWM(uint8_t pin) {
...
+    // Stash the val and range so we can re-evaluate the fraction
+    // should the user change PWM frequency.  This allows us to
+    // give as great a precision as possible.  We know by construction
+    // that the waveform for this pin will be inactive so we can borrow
+    // memory from that structure.
   wvfState.waveform[pin].desiredHighCycles = val;  // Numerator == high
   wvfState.waveform[pin].desiredLowCycles = range; // Denominator == low
 
-  uint32_t cc = (_pwmPeriod * val) / range;
+    uint32_t cc = (_pwmPeriod * val) / range;
```
earlephilhower, line 274: Everything but line 271/272 was reindented here.  Can you make them all the same indent?

cores/esp8266/Tone.cpp
```diff
@@ -30,11 +30,6 @@ static void _startTone(uint8_t _pin, uint32_t high, uint32_t low, uint32_t durat
     return;
   }
 
-  // Stop any analogWrites (PWM) because they are a different generator
-  _stopPWM(_pin);
-  // If there's another Tone or startWaveform on this pin
-  // it will be changed on-the-fly (no need to stop it)
-
```
earlephilhower, line 37: LGTM

cores/esp8266/core_esp8266_waveform_phase.cpp
```diff
@@ -267,7 +264,7 @@ static IRAM_ATTR void timer1Interrupt() {
   const bool isCPU2X = CPU2X & 1;
   if ((waveform.toSetBits && !(waveform.enabled & waveform.toSetBits)) || waveform.toDisableBits) {
     // Handle enable/disable requests from main app.
-    waveform.enabled = (waveform.enabled & ~waveform.toDisableBits) | waveform.toSetBits; // Set the requested waveforms on/off
+    waveform.enabled = (waveform.enabled | waveform.toSetBits) & ~waveform.toDisableBits; // Set the requested waveforms on/off
```
earlephilhower, line 267: I suppose there is a slight logic change here if the "toset" and "todisable" bitmasks overlap (which seems like a programming error somewhere higher in the stack).  Intended?  Fixing something you found?

cores/esp8266/core_esp8266_waveform_pwm.cpp
```diff
@@ -529,39 +532,39 @@ static IRAM_ATTR void timer1Interrupt() {
...
+              GPOS = pwmState.mask; // Set all active pins high
+              if (pwmState.mask & (1 << 16)) {
+                GP16O = 1;
+              }
+              pwmState.idx = 0;
+            }
+            else {
+              do {
+                // Drop the pin at this edge
+                GPOC = 1 << pwmState.pin[pwmState.idx];
```
earlephilhower, line 553: This bit was bracketed inside an `if` in the original code, so the logic here is different and I'm not sure correct.

cores/esp8266/core_esp8266_wiring_digital.cpp
```diff
@@ -82,8 +82,7 @@ extern void __pinMode(uint8_t pin, uint8_t mode) {
 }
 
 extern void IRAM_ATTR __digitalWrite(uint8_t pin, uint8_t val) {
-  stopWaveform(pin); // Disable any Tone or startWaveform on this pin
-  _stopPWM(pin);     // and any analogWrites (PWM)
+  stopWaveform(pin); // Disable any Tone, startWaveform, or analogWrite on this pin
```
earlephilhower, line 85: LGTM

cores/esp8266/core_esp8266_wiring_pwm.cpp
```diff
@@ -44,7 +44,14 @@ extern void __analogWriteFreq(uint32_t freq) {
 }
 
 extern void __analogWrite(uint8_t pin, int val) {
-  analogWriteMode(pin, val, false);
+  if (pin > 16) {
```
earlephilhower, line 47: This changed standard behavior.  `analogWrite()` in  Arduino-land doesn't allow open drain in the cores I've looked at, so this should be undone.  The user has to explicitly use the overloaded call we added in this repo a while back (it's for a specific closed bug).

cores/esp8266/core_esp8266_wiring_pwm.cpp
```diff
@@ -59,27 +66,35 @@ extern void __analogWriteMode(uint8_t pin, int val, bool openDrain) {
   }
 
   if (analogMap & 1UL << pin) {
-  // Per the Arduino docs at https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/
-  // val: the duty cycle: between 0 (always off) and 255 (always on).
-  // So if val = 0 we have digitalWrite(LOW), if we have val==range we have digitalWrite(HIGH)
-
     analogMap &= ~(1 << pin);
+    const bool isOpenDrain = GPC(pin) & (1 << GPCD);
```
earlephilhower, line 70: Same comment about openDrain.  Please undo

cores/esp8266/core_esp8266_wiring_pwm.cpp
```diff
@@ -59,27 +66,35 @@ extern void __analogWriteMode(uint8_t pin, int val, bool openDrain) {
...
   // Find the first GPIO being generated by checking GCC's find-first-set (returns 1 + the bit of the first 1 in an int32_t)
   int phaseReference = __builtin_ffs(analogMap) - 1;
+  // Per the Arduino docs at https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/
+  // val: the duty cycle: between 0 (always off) and 255 (always on).
+  // So if val = 0 we have digitalWrite(LOW), if we have val==range we have digitalWrite(HIGH)
   if (_setPWM(pin, val, analogScale)) {
-    analogMap |= (1 << pin);
-  } else if (startWaveformClockCycles(pin, high, low, 0, phaseReference, 0, true)) {
-    analogMap |= (1 << pin);
+    if (val > 0 && val < analogScale) analogMap |= (1 << pin);
```
earlephilhower, line 86: We try not to use 1-line `if`s, preferring the bracketed version because it's bitten us before.

cores/esp8266/core_esp8266_wiring_pwm.cpp
```diff
@@ -59,27 +66,35 @@ extern void __analogWriteMode(uint8_t pin, int val, bool openDrain) {
...
+  } else {
+    const bool detach = (val == 0 || val == analogScale);
+    // To go steady LOW or HIGH, let the waveform run into next duty cycle, if any. Then stop.
+    if (startWaveformClockCycles(pin, high, low, static_cast<uint32_t>(detach), phaseReference, 0, true)) {
+      if (detach) {
+        delay((1000 + analogFreq) / analogFreq);
+        stopWaveform(pin);
```
earlephilhower, line 93: Do you mean to clear analogMap(1<<pin) here on detach?

earlephilhower (changes requested):
Some logic changes I don't think are correct here.

Also, please try not to mix naming/indentation/style changes with logic changes.  It makes reviewing the PR very difficult because multiple unrelated things are all changing at once.  There are probably 30 real logic/type change lines in this PR, but it's a few hundred lines due to formatting and renaming changes (which GH diff struggles to match, making it a real pain to examine).  Simple change 1-thing PRs are way faster/easier for 3rd parties to look at and review.

cores/esp8266/core_esp8266_waveform_pwm.cpp
dok-net: It's in a `#pragma GCC optimize ("Os")` code section anyway. Let the compiler decide?

cores/esp8266/core_esp8266_waveform_phase.cpp
dok-net: On purpose.

cores/esp8266/core_esp8266_waveform_pwm.cpp
dok-net: I think: the ISR does NOT act on the request immediately when the ISR is triggered, but only as scheduled previously, anyway, because of `!pwmState.cnt`:
```
  } else if (!pwmState.cnt && pwmState.pwmUpdate) {
    // Start up the PWM generator by copying from the mailbox
    pwmState.cnt = 1;
    pwmState.idx = 1; // Ensure copy this cycle, cause it to start at t=0
    pwmState.nextServiceCycle = GetCycleCountIRQ(); // Do it this loop!
    // No need for mem barrier here.  Global must be written by IRQ exit
  }
```

dok-net: So, it's a tradeoff, when it's the first PWM starting, you're right, if anything else is running already, it's wasting resources and may very badly affect PWM/waveform timings.

cores/esp8266/core_esp8266_waveform_pwm.cpp
dok-net: Blech, completely different presentation in my environment, lines 273-274 are mis-indented here - good catch of course, I'm fixing this, thanks.

cores/esp8266/core_esp8266_wiring_pwm.cpp
dok-net: Case in point - GitHub review is completely unreadable here, I'm have the hardest time figuring out which line you are referring to with your remark :-)

dok-net: The intent is a bit different - `analogWrite` before PR always resets the open drain mode, this change simply prevents that. Otherwise, the overloaded call sows more confusion than just fixing an issue.

cores/esp8266/core_esp8266_wiring_pwm.cpp
dok-net: Please reconsider.

cores/esp8266/core_esp8266_wiring_pwm.cpp
dok-net: Fair enough.

cores/esp8266/core_esp8266_wiring_pwm.cpp
dok-net: Yes, it's stopped to permanently HIGH or LOW, so drop the bit from analogMap.

cores/esp8266/core_esp8266_wiring_pwm.cpp
dok-net: I explained in https://github.com/esp8266/Arduino/pull/7841

dok-net:
@s-hadinger I am curious if you would be willing to spend some time on giving an opinion with regard to this PR and/or #8011 performing with LED strips. Real-life use cases are the most interesting, after all :-)

s-hadinger:
Sorry I'm not familiar with this code. I'm afraid I can't help but

dok-net:
@s-hadinger I am sorry, too, I was just going by things like https://github.com/esp8266/Arduino/pull/7231#issuecomment-616758135 and #7057 and all your much appreciated feedback to #7022

cores/esp8266/core_esp8266_waveform_phase.cpp
dok-net: It's been a while now, but I still remember actually doing this change to fix a bug.

cores/esp8266/core_esp8266_waveform_pwm.cpp
dok-net: Re-added the noinline attribute.

PR #8010, opened 2021-04-30
Smartconfig begin takes an optional user callback

arihantdaga:
In the current state of the smart config using Arduino core, mobile has to broadcast the ssid and password unencrypted. And the esp will connect to the wifi credentials received directly. 
By adding support for this optional callback, mobile can send an encrypted password string which can be handled/decrypted by user callback and then connect the esp to the network using the parsed credentials. This does not make it fully secure, but at least a pre-shared key based encryption can be implemented(up to the user).

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -604,10 +604,19 @@ int8_t ESP8266WiFiSTAClass::RSSI(void) {
 
 bool ESP8266WiFiSTAClass::_smartConfigStarted = false;
 bool ESP8266WiFiSTAClass::_smartConfigDone = false;
+sc_callback_t ESP8266WiFiSTAClass::_smartConfigUserCallback;
 
 /**
  * Start SmartConfig
+ * @param userCallback
+ * @return smart config start success
  */
+
+bool ESP8266WiFiSTAClass::beginSmartConfig(sc_callback_t userCallback) {
+    _smartConfigUserCallback = userCallback;
+    return beginSmartConfig();
+}
+
 bool ESP8266WiFiSTAClass::beginSmartConfig() {
     if(_smartConfigStarted) {
         return false;
@@ -664,11 +673,22 @@ void ESP8266WiFiSTAClass::_smartConfigCallback(uint32_t st, void* result) {
     sc_status status = (sc_status) st;
     if(status == SC_STATUS_LINK) {
         station_config* sta_conf = reinterpret_cast<station_config*>(result);
-
-        wifi_station_set_config(sta_conf);
-        wifi_station_disconnect();
-        wifi_station_connect();
-
+        if(_smartConfigUserCallback) {
+            if(WiFi._persistent) {
+                wifi_station_set_config(sta_conf);
+            } else {
+                wifi_station_set_config_current(sta_conf);
+            }
+            _smartConfigUserCallback(status, sta_conf);
+        } else {
+            if(WiFi._persistent) {
+                wifi_station_set_config(sta_conf);
+            } else {
+                wifi_station_set_config_current(sta_conf);
+            }
+            wifi_station_disconnect();
+            wifi_station_connect();
+        }
         _smartConfigDone = true;
     } else if(status == SC_STATUS_LINK_OVER) {
         WiFi.stopSmartConfig();
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
@@ -28,6 +28,7 @@
 #include "ESP8266WiFiGeneric.h"
 #include "user_interface.h"
 #include "LwipIntf.h"
+#include "smartconfig.h"
 
 
 class ESP8266WiFiSTAClass: public LwipIntf {
@@ -96,6 +97,7 @@ class ESP8266WiFiSTAClass: public LwipIntf {
 
         bool beginWPSConfig(void);
         bool beginSmartConfig();
+        bool beginSmartConfig(sc_callback_t userCallback);
         bool stopSmartConfig();
         bool smartConfigDone();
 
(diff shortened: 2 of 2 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -664,11 +673,22 @@ void ESP8266WiFiSTAClass::_smartConfigCallback(uint32_t st, void* result) {
...
+            } else {
+                wifi_station_set_config_current(sta_conf);
+            }
+            _smartConfigUserCallback(status, sta_conf);
+        } else {
+            if(WiFi._persistent) {
+                wifi_station_set_config(sta_conf);
+            } else {
+                wifi_station_set_config_current(sta_conf);
+            }
```
d-a-v, line 688: This block is duplicated and always executed.

d-a-v:
Can you please also add an example showing the use of your proposal ?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
arihantdaga: Fixed

d-a-v:
Thanks for the example !  I see several improvements that could be addressed in this proposal.

First, the API should be kept simple:
- For example the first variable `st` is unused in the user callback, could it be removed if it is useless ?
- The callback needs to be casted when calling `WiFi.beginSmartConfig()`.

Is it needed to follow `sc_callback_t` type in this *user* callback ?
Would it be possible for the user to only provide `decryptSmartConfigPassword()` and leave `smartConfigCallback()` hidden in the internal API ?

Second, the API should be kept simple :)
As an Arduino example meant to teach about the API, `decryptSmartConfigPassword()` is not clear enough (variable name, memory allocation, why not using the equivalent `strcpy()`)
```cpp
void decryptSmartConfigPassword(char * buffer, const char* encString){
  // currently returning same string
  // but we can throw in simple aes encryption. 
  memcpy(buffer,encString, strlen(encString));
}
```
What about this:
```cpp
String decryptSmartConfigPasswordTrivial(const char* encString){
  // no decoding
  return encString;
}

String decryptSmartConfigPasswordAES(const char* encString){
  String decrypted;
  decrypted.reserve(nnn);
  /* show an example on how to decode with AES */
  return decrypted;
}
```

arihantdaga:
@d-a-v  Thank you for guiding. I'll improve the example.

libraries/ESP8266WiFi/examples/WiFiSmartConfig/WiFiSmartConfig.ino
```diff
@@ -0,0 +1,69 @@
+/*
+    This sketch demonstrates how to use Smart Config.
+    It calls a user callback on recieving credentials from ESPTouch mobile app.
```
dok-net, line 3: Spelling mistake - `receiving`

libraries/ESP8266WiFi/examples/WiFiSmartConfig/WiFiSmartConfig.ino
```diff
@@ -0,0 +1,69 @@
...
+#include <WiFiClient.h>
+#include "smartconfig.h"
+
+void decryptSmartConfigPasswordTrivial(char * decrypted, const char* encString){
+  // No encryption. Password is sent as plan text
+  // If password is encrypted with AES(CBC/ECB). Decrypt here
+  // Note: The format of encrypted string and choice of encryption algoright is upto the user.
+  // It has to be chosen with constraint that maxium length of encrypted string should not be above 64. 
+    
+  strcpy(decrypted, encString);
```
dok-net, line 20: Inherently unsafe, please provide buffer size in the API and use strncpy?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -664,11 +673,18 @@ void ESP8266WiFiSTAClass::_smartConfigCallback(uint32_t st, void* result) {
     sc_status status = (sc_status) st;
     if(status == SC_STATUS_LINK) {
         station_config* sta_conf = reinterpret_cast<station_config*>(result);
-
-        wifi_station_set_config(sta_conf);
+        if(_scDecryptFn) {
+            char decryptedPass[64] = "";
+            _scDecryptFn(decryptedPass, (const char *)sta_conf->password);
+            memcpy(sta_conf->password,decryptedPass, 64);
```
dok-net, line 679: Allocation of intermediate buffer `decryptedPass` would be obsoleted by providing buffer size to _scDecryptFn, done like  this it's particularly prone to malicious buffer overruns, or is it?

libraries/ESP8266WiFi/src/ESP8266WiFiType.h
```diff
@@ -156,5 +156,6 @@ struct WiFiEventSoftAPModeProbeRequestReceived
     uint8 mac[6];
 };
 
+typedef void (*sc_pass_decrypt_fn_t)(char* decrypted, const char* encString);
```
dok-net, line 159: A matter of preference, I'm just suggesting: perhaps use `Cb` instead of `Fn` and stick to C++ camelCase for the function type as well? Could even use `std::function`.

libraries/ESP8266WiFi/examples/WiFiSmartConfig/WiFiSmartConfig.ino
```diff
@@ -0,0 +1,69 @@
...
+  WiFi.softAPdisconnect();
+  WiFi.enableAP(false);
+  WiFi.enableSTA(true);
+  WiFi.mode(WIFI_STA);
+  delay(1000);
+
+  // If sending encrypted password. Pass an optional decrypt password function
+  WiFi.setSmartConfigDecryptFn(decryptSmartConfigPasswordTrivial);
+
+  if(!WiFi.beginSmartConfig() {
```
dok-net, line 39: Please make sure that examples actually build ;-) Plus, CI insists on astyle.

libraries/ESP8266WiFi/examples/WiFiSmartConfig/WiFiSmartConfig.ino
```diff
@@ -0,0 +1,69 @@
...
+    delay(100);
+  }
+  Serial.print("IP address: ");
+  Serial.println(WiFi.localIP());
+
+}
+
+void loop() {
+  
+}
```
dok-net, line 69: Newline at end of file is demanded by quite a few tools.

PR #8025, opened 2021-05-10
Update Arduino.h

alihakimtaskiran:
As I attempt to make the pin HIGH, it becomes LOW; and as I attempt to make the pin LOW, it becomes HIGH. This leaves a contradiction in the code. Therefore, I reversed the logic HIGH and logic LOW names

### DIFF
--- a/cores/esp8266/Arduino.h
+++ b/cores/esp8266/Arduino.h
@@ -41,8 +41,8 @@ extern "C" {
 #include "core_esp8266_features.h"
 #include "core_esp8266_version.h"
 
-#define HIGH 0x1
-#define LOW  0x0
+#define HIGH 0x0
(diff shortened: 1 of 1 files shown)


alihakimtaskiran:
@earlephilhower In the Arduino boards LOW turns down the LED and HIGH glowes the LED. But in the esp8266 it's reversed.

earlephilhower:
`digitalWrite(pin, LOW)` puts GND on a pin, while `digitalWrite(pin, HIGH)` puts VCC.  The ESP isn't 5V tolerant and has low current source capabilities, so all boards use a active low signal to control LEDs so the chip only has to sink current to ground, not source it from internal VCC.  There is nothing wrong with the core here or in AVR.  They're just different board designs.

PR #8037, opened 2021-05-16
update draft release github action

d-a-v:
ditto on the draft action file

### DIFF
--- a/.github/workflows/tag-to-draft-release.yml
+++ b/.github/workflows/tag-to-draft-release.yml
@@ -18,13 +18,14 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: true
+        fetch-depth: 0
     - uses: actions/setup-python@v2
       with:
         python-version: '3.x'
     - name: Set GIT tag name
       run: |
         # Sets an environment variable used in the next steps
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-to-draft-release.yml
```diff
@@ -18,13 +18,14 @@ jobs:
...
         submodules: true
+        fetch-depth: 0
     - uses: actions/setup-python@v2
       with:
         python-version: '3.x'
     - name: Set GIT tag name
       run: |
         # Sets an environment variable used in the next steps
-        echo "::set-env name=TRAVIS_TAG::$(git describe --exact-match --tags)"
+        echo "TRAVIS_TAG=$(git describe --exact-match --tags)" >> $GITHUB_ENV
```
mcspr, line 28: Note that there is $GITHUB_REF set to `refs/tags/1.2.3` when it's tag, so bash also can do `${GITHUB_REF:10}`
ref. https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables and https://github.com/letscontrolit/ESPEasy/pull/3642/commits/4082a3fc847151f74937ecb9609a085805719ee7#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L91 (but I gave up on using the repo for tag retrieval)

PR #8056, opened 2021-05-19
Add writeToPrint to ESP8266HTTPClient

TheNitek:
Currently ESP8266HTTPClient only has writeToStream(Stream *), but interally only uses the Stream* as a Print*. Since Print* is more lightweight, adding a writeToPrint(Print *) seems to make sense. E.g. this allows to use SdFile directly.
wirteToStream is kept for backward compatibility


d-a-v (approved):
Thanks !

earlephilhower (approved):

PR #8073, opened 2021-05-23
ESP8266WiFiScan::scanNetworks signature and capabilities enhancement for ESP32 compatibility

### LINKED ISSUE #8053: suggestion to change parameters of scanNetworks() to match the ones of ESP32
The ESP32, in header WiFiScan.h, defines this function:
int16_t scanNetworks(bool async = false, bool show_hidden = false, bool passive = false, uint32_t max_ms_per_chan = 300, uint8_t channel = 0);

The ESP8266, in header ESP8266WiFiScan.h, defines this function:
int8_t scanNetworks(bool async = false, bool show_hidden = false, uint8 channel = 0, uint8* ssid = NULL);

My question is why you do not modify the ESP8266 code this to be as close match as possible with the ESP32?
It would be nice for code compatibility.
Or at least parameters async, show_hidden, passive, channel, and max_ms_per_chan.

Thank you

dok-net:
Fixes #8053


libraries/ESP8266WiFi/src/ESP8266WiFiScan.h
```diff
@@ -34,7 +34,7 @@ class ESP8266WiFiScanClass {
 
     public:
 
-        int8_t scanNetworks(bool async = false, bool show_hidden = false, uint8 channel = 0, uint8* ssid = NULL);
```
d-a-v, line 37: This is a breaking change.
What about leaving the former API without the default values ?
```cpp
+ int8_t scanNetworks(bool async = false, bool show_hidden = false, uint8 channel, uint8* ssid);
```

dok-net: Looks fishy, and indeed, it leads to an ambiguity error for the case of default ssid argument, because the compiler cannot distiguish the actual parameter for `bool passive` from `uint8_t channel`. Better break once, for good, in the breaking 3.0 release.

PR #8075, opened 2021-05-23
platformIO: import MMU defines

d-a-v:
documentation: https://arduino-esp8266.readthedocs.io/en/latest/mmu.html
@ivankravets @valeros

### DIFF
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -266,13 +266,6 @@ else:
 
 
 #
-# Waveform
-#
-if "PIO_FRAMEWORK_ARDUINO_WAVEFORM_LOCKED_PHASE" in flatten_cppdefines:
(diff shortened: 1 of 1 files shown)


valeros:
Hi @d-a-v ! Thanks for bringing it up. We already have such functionality https://docs.platformio.org/en/latest/platforms/espressif8266.html#mmu-adjusting-icache-to-iram-ratio but haven't pushed it to the upstream yet. We plan to do this once we publish a new release of the `espressif8266` platform.

d-a-v:
Nice ! I guess I'll just have to close this PR then.
We'll remove the obsolete PWM definition in another PR.
One day I'll try to understand more precisely what definitions go where, between `tools/platformio-build.py` and PIO repo.

PR #8076, opened 2021-05-23
LwipIntfDev - static IP config for eth interfaces

jandrassy:
There are some problems with Ethernet LwIP interfaces static IP configuration. I want to address them in this PR, but I need directions.
I tested the Ethernet interfaces for W5500 and ENC28j60 with WiFi off and eth.setDefault().

Problem one: DNS setting was not applied. The first commit in this PR has a commit with a proposal how to fix it. Additionally I added default value for dns parameters in `config`.

Problem two: static config only works if eth.config is followed by eth.setDefault(). I am new to LwIP. Maybe eth.config should have `netif_set_up`?


d-a-v:
Problem1: The current changes are OK with me.
(I'll check on problem2 later)

d-a-v:
@jandrassy Thanks!
Can you please open a new issue for your problem2 ?
> static config only works if `eth.config` is followed by `eth.setDefault()`. I am new to LwIP. Maybe `eth.config` should have `netif_set_up`?

PR #8091, opened 2021-05-31
Update toolchain to fix pgm_read_float_unaligned

### LINKED ISSUE #8083: `pgm_read_float_unaligned()` ignores the second word
### Basic Infos

- [ ] I have searched the issue tracker for a similar issue.

#### Platform

- Hardware:					[other (ESP-12 compatible)] (AE-ESP-WROOM-02: see https://akizukidenshi.com/catalog/g/gK-09758/)
- Core Version:			[bf6f3e7c536176cf6c6263746de57d0202c17124]

### Settings in IDE

- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[80Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

`pgm_read_float_unaligned()` in 'tools/xtensa-lx106-elf/xtensa-lx106-elf/include/sys/pgmspace.h' should use `pgm_read_dword_with_offset()` instead of `pgm_read_with_offset()`, as per `pgm_read_dword_unaligned()`.

```cpp
void setup() {

  static const float SPECIMEN = 3.14159265359F;
  static uint8_t __attribute__((aligned(4))) buffer[8];
  memset(buffer, 0, sizeof(buffer));

  uint8_t* unaligned_location = &buffer[1];
  memcpy(unaligned_location, &SPECIMEN, sizeof(SPECIMEN));

  Serial.begin(115200);
  Serial.print("\n\n"
               "** pgm_read_float_unaligned() test **\n"
               "specimen: ");
  Serial.println(SPECIMEN, 6);
  Serial.print("readout:  ");
  Serial.println(pgm_read_float_unaligned(unaligned_location), 6);
}

void loop() {

}
```

### Debug Messages

```
** pgm_read_float_unaligned() test **
specimen: 3.141593
readout:  0.000000
```

With the abovementioned way,
```
** pgm_read_float_unaligned() test **
specimen: 3.141593
readout:  3.141593
```

earlephilhower:
Fixes #8083


dok-net:
@earlephilhower I can't quite make up a sketch where accessing unaligned data directly fails, with PROGMEM and all, it just always gives the right values :-( Do you have some code to test that pgm_read_xxx even makes a difference?

earlephilhower:
Make sure the exception handler is disabled.  The code in the original bug fails w/master and passes w/this.  Also use pgm_read_xxx_aligned or pgm_read_xxx_unaligned to ensure you pull in the expected version of the macros.

dok-net:
Exception handler is disable, if option "Exceptions: disabled (new abort on oom)" is what you are referring to.
The code in the issue is too constructed for my liking, pgm_read* is for reading from unaligned IRAM or PROGMEM. And either I can't get it to run (crash? unintelligible gibberish on serial) at all, master or this PR, or I works without pgm_read for totally unaligned floats in PROGMEM. Something like a useful MCVE is needed here!

dok-net:
Equally made up, useless test, but it proves the point:
```
static const float PROGMEM SPECIMEN = 3.14159265359F;
static const uint8_t __attribute__((aligned(4))) PROGMEM buffer[8] = {
    0,
    ((uint8_t*)&SPECIMEN)[0],
    ((uint8_t*)&SPECIMEN)[1],
    ((uint8_t*)&SPECIMEN)[2],
    ((uint8_t*)&SPECIMEN)[3],
    0,
    0,
    0
};

void setup() {
    const uint8_t* unaligned_location = &buffer[1];

    Serial.begin(74880);
    delay(500);
    Serial.print("\n\n"
        "** pgm_read_float_unaligned() test **\n"
        "specimen: ");
    Serial.println(SPECIMEN, 6);
    Serial.print("readout:  ");
    Serial.println(pgm_read_float_unaligned(unaligned_location), 6);
}

void loop() {
}
```
Fails spectacularly verbatim. If you remove the PROGMEM attribute from buffer, works properly!

dok-net:
Here is one that works with this PR, and proves that master doesn't work. It's less synthetic in that it doesn't fabricate an array in RAM, but puts the constants into PROGMEM. If one adds `const` to the `constants` structure instance, the compiler optimizations outsmart the test and then it proves nothing.
```
static const uint8_t pi_array[] PROGMEM = { 42, 219, 15, 73, 64 };
static struct __attribute__((__packed__)) {
	uint8_t pad = 42;
	float pi = 3.1415926535897932384626433832795F;
	double pi_d = 3.1415926535897932384626433832795D;
} constants PROGMEM;

void setup()
{
	Serial.begin(74880);
	while (!Serial);
	delay(500);
	Serial.println("pgmread");
	float pi = PI;
	Serial.println(pi, 6);
	for (int i = 0; i < 4; ++i)
		Serial.println(*((uint8_t*)&pi + i));
	Serial.println(pgm_read_float_unaligned((float*)&pi_array[1]), 6);
	Serial.println(pgm_read_float_unaligned(&constants.pi), 6);
	//Serial.println(pgm_read_double_unaligned(&constants.pi_d), 11);
}

void loop()
{

}
```

earlephilhower:
By exceptions I meant HW exceptions, not C++ exceptions.  The menu name is "Non 32-bit access" or something similar.  The default now may include a HW exception handler to (very, very) slowly emulate the misaligned reads.

The examples in the main bug cover the failure case, even though they use RAM.  ROM or not, the misalignment is the real problem (since float = 32b = natural read from anywhere, double = 64 bit = 2x32b natural reads from anywhere).  You can't do non-32b aligned `float` reads (w/o the HW exception handler) from anywhere.

d-a-v (approved):

Issue #8085, opened 2021-05-30
Proxy support for ESP8266HTTPClient


ChrisSmith:
### Problem Description

I would be nice if the http client supported http proxies. Adding proxy support is relatively straightforward as its done using headers. 

Regular HTTP request (connection to target 1.1.1.1)
```
GET / HTTP/1.1
Host: 1.1.1.1:8080
User-Agent: curl/7.68.0
```

Request to a proxy (connection to proxy 2.2.2.2)
```
GET http://1.1.1.1:8000/ HTTP/1.1
Host: 1.1.1.1:8000
User-Agent: curl/7.68.0
```

The primary difference is that the tcp connection doesn't match the host header and the full uri is used in the request line. Authorization could be manually performed using the current `addHeader()` method to add the `Proxy-Authorization` header.

I have working changes locally that I'd be happy to contribute if the maintainers think this is a worthwhile feature.

d-a-v:
Your Pull-Request is welcome !

FAAAAT:
Hello ChrisSmith,

I have a proxy request similar as yours, but it's for https requests through http proxy. Is your PR support https request?

Thanks.

ChrisSmith:
@FAAAAT are you referring to 8086? My implementation doesn't support the TCP tunneling required for CONNECT

I don't think thats required for a https proxy though, an alternative implementation would perform tls termination at the proxy which would be possible to support in my implementation but I have not tried it yet.

PR #8094, opened 2021-06-02
Add support for http proxy to esp8266 http client

ChrisSmith:
Fixes https://github.com/esp8266/Arduino/issues/8085

I was able to verify this works with my proxy setup, it would be good for others to test as well, I'm no expert on how proxies are configured. 

I do very little c++ so don't be afraid to give me lots of feedback :)

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -858,17 +858,72 @@ bool HTTPClient::connect(void)
 
     _client->setTimeout(_tcpTimeout);
 
-    if(!_client->connect(_host.c_str(), _port)) {
-        DEBUG_HTTPCLIENT("[HTTP-Client] failed connect to %s:%u\n", _host.c_str(), _port);
+    String host = _host;
+    uint16_t port = _port;
+    if (_useProxy)
+    {
+        host = _proxyHost;
+        port = _proxyPort;
+        DEBUG_HTTPCLIENT("[HTTP-Client] using proxy %s:%u\n", host.c_str(), port);
+    }
+
+    if(!_client->connect(host.c_str(), port)) {
+        DEBUG_HTTPCLIENT("[HTTP-Client] failed connect to %s:%u\n", host.c_str(), port);
         return false;
     }
 
-    DEBUG_HTTPCLIENT("[HTTP-Client] connected to %s:%u\n", _host.c_str(), _port);
+    DEBUG_HTTPCLIENT("[HTTP-Client] connected to %s:%u\n", host.c_str(), port);
 
 #ifdef ESP8266
     _client->setNoDelay(true);
 #endif
     return connected();
+}
+
+bool HTTPClient::setProxyHost(const String &proxyUrl)
+{
+    String url(proxyUrl);
+
+    DEBUG_HTTPCLIENT("[HTTP-Client][proxy] url: %s\n", url.c_str());
+
+    // check for : (http: or https:
+    int index = url.indexOf(':');
+    if (index < 0) {
+        DEBUG_HTTPCLIENT("[HTTP-Client][proxy] failed to parse protocol\n");
+        return false;
+    }
+
+    _proxyProtocol = url.substring(0, index);
+    url.remove(0, (index + 3)); // remove http:// or https://
+
+    if (_proxyProtocol == "http") {
+        // set default port for 'http'
+        _port = 80;
+    } else if (_proxyProtocol == "https") {
+        // set default port for 'https'
+        _port = 443;
+    } else {
+        DEBUG_HTTPCLIENT("[HTTP-Client][proxy] unsupported protocol: %s\n", _proxyProtocol.c_str());
+        return false;
+    }
+
+    index = url.indexOf('/');
+    String host = url.substring(0, index);
+    url.remove(0, index); // remove host part
+
+    // get port
+    index = host.indexOf(':');
+    if (index >= 0) {
+        _proxyHost = host.substring(0, index); // hostname
+        host.remove(0, (index + 1));           // remove hostname + :
+        _proxyPort = host.toInt();             // get port
+    } else {
+        _proxyHost = host;
+    }
+
+    DEBUG_HTTPCLIENT("[HTTP-Client][proxy] host: %s port: %d\n", _proxyHost.c_str(), _proxyPort);
+    _useProxy = true;
+    return true;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -858,19 +858,74 @@ bool HTTPClient::connect(void)
...
-    DEBUG_HTTPCLIENT("[HTTP-Client] connected to %s:%u\n", _host.c_str(), _port);
+    DEBUG_HTTPCLIENT("[HTTP-Client] connected to %s:%u\n", host.c_str(), port);
 
 #ifdef ESP8266
     _client->setNoDelay(true);
 #endif
     return connected();
 }
 
+bool HTTPClient::setProxyHost(const String &proxyUrl)
```
ChrisSmith, line 883: most of this is copied from beginInternal. Would it be worthwhile to pull the shared logic out into a URI parser?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
```diff
@@ -177,6 +177,7 @@ class HTTPClient
     void setAuthorization(const char * user, const char * password);
     void setAuthorization(const char * auth);
     void setTimeout(uint16_t timeout);
+    bool setProxyHost(const String &proxyUrl);
```
ChrisSmith, line 180: Sample usage 

```
WiFiClient client;
HTTPClient http;

http.setProxyHost("http://192.168.1.1:8080");
http.begin(client, "http://foo.bar/");
  
// If the proxy requires auth
http.addHeader("Proxy-Authorization", "Basic aaaaaaaaaaaa=");
http.GET();
```

ChrisSmith: sure, can you point me to where those are located?

d-a-v: Here's the HTTPClient [examples location](https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266HTTPClient/examples).
As you can see, the file hierarchy is `myExampleName/myExampleName.ino`.

Documentation is ...
... there's absolutely no official documentation on `HTTPClient`.
So your addition will be added in the documentation when it will be written for the whole class, or if we add doxygen at some point.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
ChrisSmith: I gave it a shot in this commit https://github.com/esp8266/Arduino/pull/8094/commits/178495725b3db7bf3f1cf24f6965255b3e556b5c

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
```diff
@@ -31,6 +31,10 @@
 #include <StreamString.h>
 #include <WiFiClient.h>
 
+
+#define DEBUG_ESP_HTTP_CLIENT 1
+#define DEBUG_ESP_PORT Serial
```
d-a-v, line 36: These debug lines should be removed

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.h
```diff
@@ -231,21 +244,21 @@ class HTTPClient
...
     uint16_t _tcpTimeout = HTTPCLIENT_DEFAULT_TCP_TIMEOUT;
     bool _useHTTP10 = false;
 
-    String _uri;
-    String _protocol;
     String _headers;
     String _userAgent;
-    String _base64Authorization;
+    
+    bool _useProxy = false;
```
d-a-v, line 260: Can't `_useProxy` be replaced by `(_proxyUri != nullptr)` ?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -858,19 +803,108 @@ bool HTTPClient::connect(void)
 
     _client->setTimeout(_tcpTimeout);
 
-    if(!_client->connect(_host.c_str(), _port)) {
-        DEBUG_HTTPCLIENT("[HTTP-Client] failed connect to %s:%u\n", _host.c_str(), _port);
+    String host = _uri->host;
```
d-a-v, line 806: `host` can be a reference here?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -884,15 +918,27 @@ bool HTTPClient::sendHeader(const char * type)
...
-    header.reserve(_headers.length() + _uri.length() +
-            _base64Authorization.length() + _host.length() + _userAgent.length() + 128);
+    header.reserve(_headers.length() + _uri->path.length() +
+            _uri->base64Authorization.length() + _uri->host.length() + _userAgent.length() + 128);
     header += type;
     header += ' ';
-    if (_uri.length()) {
-        header += _uri;
+
+    String target;
```
d-a-v, line 926: `target` should have its `.reserve(sufficient_space)` like `header` has above

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -884,15 +918,27 @@ bool HTTPClient::sendHeader(const char * type)
...
+    if (_useProxy) {
+        // If using proxy, use full uri in request line (http://foo:123/bar)
+        target = _uri->protocol + F("://") + _uri->host;
```
d-a-v, line 929: When a `String` is `::reserve(sufficient_space)`, and to avoid ram fragmentation, it is generally preferable to do
```cpp
target = _uri->protocol;
target += F("://");
target += _uri->host;
```
to avoid use of temporaries when building the pieces.

d-a-v (changes requested):
Thank you for your time on this factorisation effort !
Little changes, the most important are the debug lines that were mistakingly added I believe

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
ChrisSmith: I don't believe so, its reassigned right below. ` host = _proxyUri->host;` If it were a reference like `String &host = _uri->host;` it would reassign the value of `_uri->host` which is not what is intended

d-a-v: True :/ Using a pointer instead would avoid a useless allocation

ChrisSmith: done

d-a-v:
Thanks for all your updates !
CI is still complaining:
https://github.com/esp8266/Arduino/pull/8094/checks?check_run_id=2744634954#step:5:117

d-a-v (approved):
Thanks for your patience !

d-a-v:
I have to explain my last comment.
I realized later that this file had to contain gcc cmdline options `-Dblah=123` or `-fdo-some-optimization`.
So `@` is good, not `-include`.
But here's what I got when trying with this file:
```
-DBLAH
-DBLOH=1
```
It was copied to the tmp dir with this content:
```
#line 1 "/home/gauchard/dev/proj/arduino/8095/build_opt.h"
-DBLAH
-DBLOH=1
```
And that gave this error message when compiling:
```
xtensa-lx106-elf-g++: error: #line: No such file or directory
xtensa-lx106-elf-g++: error: 1: No such file or directory
```

d-a-v:
@ChrisSmith Do you think you can solve the conflict, or do you need help ?
I/someone will merge right after the fix

ChrisSmith:
Hey @d-a-v I don't have time to resolve the conflict and test at the moment. You should be able to resolve + commit to this branch if you'd like. Otherwise I'll get to it when I am free

mcspr:
https://github.com/esp8266/Arduino/pull/8237 also broke this even more, so I don't mind fixing this instead of @d-a-v

PR #8095, opened 2021-06-02
Add prebuild script to use additional build options

brainelectronics:
Additional compile arguments can be given by a file named "build_opt.h" and shall be placed next to the Sketch.ino file.
This enables users to easily extend their arguments on a central place.
The script has been taken from [stm32duino's Arduino_Core_STM32 package](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/system/extras/prebuild.sh)


tools/extras/prebuild.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
earlephilhower, line 1: This needs to be a small Python3 script, not bash, to support Windows.  We include Python in our toolchain distribution on all hosts just for such small scripting needs.

platform.txt
```diff
@@ -85,6 +85,16 @@ compiler.size.cmd=xtensa-lx106-elf-size
 # This can be overridden in boards.txt
 build.extra_flags=-DESP8266
 
+# Pre and post build hooks
+build.opt.name=build_opt.h
```
earlephilhower, line 89: `build_opt.h` is not going to be a valid C or C++ header.  Can we use a different extension and call it something like `build.opt` just to make it blindingly obvious?

platform.txt
```diff
@@ -107,13 +117,13 @@ recipe.hooks.linking.prelink.3.pattern="{compiler.path}{compiler.c.cmd}" -CC -E
...
 
 ## Compile S files
-recipe.S.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.flags} {compiler.S.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.S.extra_flags} {build.extra_flags} {includes} "{source_file}" -o "{object_file}"
+recipe.S.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.flags} {compiler.S.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.S.extra_flags} {build.extra_flags} "@{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
```
earlephilhower, line 126: Won't you also need to include the options in the linker stages as well?  Not sure of the exact use case here...

platform.txt
```diff
@@ -85,6 +85,16 @@ compiler.size.cmd=xtensa-lx106-elf-size
 # This can be overridden in boards.txt
 build.extra_flags=-DESP8266
 
+# Pre and post build hooks
+build.opt.name=build_opt.h
+build.opt.path={build.path}/sketch/{build.opt.name}
+
+# extras.path={build.core.path}/extras
+extras.path={runtime.platform.path}/tools/extras
```
earlephilhower, line 93: Can we just place prebuild.py in the tools directory and avoid yet another subdir with a single file, please?

earlephilhower (changes requested):
I like the idea, but there are a few changes needed to get it going.  Thanks for the PR!

platform.txt
brainelectronics: I've tried to use something else, but only .h files have been successful. I guess this is connected on how Arduino detects files, but I'm not sure

platform.txt
brainelectronics: If you would like to handover linkerargs as well, yes. But to "simply" tweak the code, those locations are sufficient. I'm open for further improvements

platform.txt
earlephilhower: Ah, good point.  The Arduino IDE won't copy every file to INO directory the build dirs.  I remember hitting this with some other things a while back.

platform.txt
earlephilhower: Fair enough.  This will get you easy access to `#define`s, etc. which would cover the vast majority of use cases.

earlephilhower (approved):
Thanks, LGTM!  I was thinking we should add something to the docs about this new option, but I can't find anywhere that it fits in well presently in the `doc/*.rst` directory.  If you think it might make sense to add a couple lines in the `faq.rst` about this, that would be icing on the cake.

d-a-v (approved):
I personnally like very much this addition.
It is a long waited feature in Arduino IDE that is still not addressed, I will cite https://github.com/arduino/Arduino/issues/421 and https://github.com/arduino/arduino-cli/issues/846. I also [made a proposal](https://github.com/arduino/arduino-cli/pull/1117) that was not dismissed, nor closed nor commented yet.
It's a pity that a global solution is not available. This forces platforms to adopt local solution like this one.
Approving, and thanks for bringing this.

doc/faq/readme.rst
```diff
@@ -152,6 +152,41 @@ Artificially clearing them is a workaround to help saving precious heap.
...
+      digitalWrite(LED_BUILTIN, LOW);
+      delay(WAIT_TIME);
+    }
+
+Place the following content in a file named ``build_opt.h`` at the same
+location as the ``sketch.ino`` file to define the ``WAIT_TIME`` to 500ms.
+
+.. code:: cpp
+
+    -DWAIT_TIME=500
```
earlephilhower, line 189: 👍  Awesome, thanks!

mhightower83:
@brainelectronics - This is great! 

I think, when you create an empty `build_opt.h`, you need to backdate the file to avoid triggering a rebuild of libraries.

For a while now, I  have been doing something similar to this PR. I have noticed that core '.S' files are not rebuilt when, for your case, the `build_opt.h` is changed. You may want to confirm your results and if necessary add a comment to the docs.
Look for something like this:
```
Using previously compiled file: /tmp/arduino_build_327471/core/cont.S.o
```
while everything else in core is rebuilt.

d-a-v:
> I think, when you create an empty build_opt.h, you need to backdate the file to avoid triggering a rebuild of libraries.

True!
It should take users's `build_opt.h` date (it may have been updated and in this case, recompilation is necessary) or, if it does not exist, be created to the same date as any other existing source file (sketch/sketch.ino? core/core_version.h? )

brainelectronics:
PR is ready for review again and could be merged if alpha testing has been passed

d-a-v:
I'm using this sketch
```cpp
#ifndef BLAH
#error
#endif
```
and this `build_opt.h`
```cpp
#define BLAH toto
```
I made a build with `arduino-builder` with linux and got this:
```
Using board 'd1_mini' from platform in folder: /local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266
Using core 'esp8266' from platform in folder: /local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266
/local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266/tools/python3/python3 /local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266/tools/prebuild.py /tmp/build_opt_test-d1 /home/gauchard/dev/proj/arduino/build_opt_test
  File "/local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266/tools/prebuild.py", line 121, in <module>
    sys.exit(main())
  File "/local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266/tools/prebuild.py", line 117, in main
    create_build_options_file(source_path, build_path)
  File "/local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266/tools/prebuild.py", line 62, in create_build_options_file
    build_opt_ctime = get_creation_time(build_opt_build_path)
  File "/local/users/gauchard/arduino/arduino-210516/hardware/esp8266com/esp8266/tools/prebuild.py", line 108, in get_creation_time
    return os.path.getctime(file_path)
  File "/usr/lib/python3.8/genericpath.py", line 65, in getctime
    return os.stat(filename).st_ctime
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/build_opt_test-d1/sketch/build_opt.h'
exit status 1
```
Windows alpha version 0.0.2 contains your PR (thanks to the tag `alpha`) and I get the same error (`sketch/build_opt.h` not found) using this time Arduino IDE.

brainelectronics:
@d-a-v That's strange. As you placed the `build_opt.h` next to the `your-sketch.ino` file, Arduino should take care of copying that build options file to the tmp build directory. This seems not to be the case. Could you check the content of the build directory? It should be similar to this content
```
./build.options.json
./your-sketch.ino.bin
./your-sketch.ino.elf
./your-sketch.ino.hex
./your-sketch.ino.map
./sketch/build_opt.h
./sketch/your-sketch.ino.cpp
./sketch/your-sketch.ino.cpp.d
./sketch/your-sketch.ino.cpp.o
```

earlephilhower (dismissed):
LGTM, still.  Thanks!  Slated for 3.1.0, after the 3.0.x bugfixes are all settled.

earlephilhower:
Argh, just saw @d-a-v's comment.  I will look at it.  I think maybe adding a build_opt.h file in one of the examples would be a good thing to catch this.

earlephilhower:
I can repro @d-a-v's failure.  Made a sketch:

sketch.ino
````void setup() {
Serial.begin(115200);
delay(5000);
#ifdef MYDEFINE
Serial.println(MYDEFINE);
#else
Serial.println("ERROR:  MYDEFINE not present");
#endif
}

void loop() {

}
````

build_opt.h
````
-DMYDEFINE="Chimichangas do not exist"
-O3
-fanalyzer
````


And builds fail with
````
/usr/local/bin/arduino-1.8.15/arduino-builder -dump-prefs -logger=machine -hardware /usr/local/bin/arduino-1.8.15/hardware -hardware /home/earle/.arduino15/packages -hardware /home/earle/Arduino/hardware -tools /usr/local/bin/arduino-1.8.15/tools-builder -tools /usr/local/bin/arduino-1.8.15/hardware/tools/avr -tools /home/earle/.arduino15/packages -built-in-libraries /usr/local/bin/arduino-1.8.15/libraries -libraries /home/earle/Arduino/libraries -fqbn=esp8266com:esp8266:d1_mini:xtal=80,vt=flash,exception=disabled,stacksmash=disabled,ssl=all,mmu=3232,non32xfer=fast,eesz=4M2M,ip=lm2f,dbg=Serial,lvl=SSLTLS_MEMHTTP_CLIENTHTTP_SERVERCOREWIFIHTTP_UPDATEUPDATEROTAOOMMDNSHWDT_NOEXTRA4K,wipe=none,baud=921600 -vid-pid=1A86_7523 -ide-version=10815 -build-path /tmp/arduino_build_231337 -warnings=all -build-cache /tmp/arduino_cache_721404 -prefs=build.warn_data_percentage=75 -verbose /home/earle/Arduino/sketch_jun17a/sketch_jun17a.ino
/usr/local/bin/arduino-1.8.15/arduino-builder -compile -logger=machine -hardware /usr/local/bin/arduino-1.8.15/hardware -hardware /home/earle/.arduino15/packages -hardware /home/earle/Arduino/hardware -tools /usr/local/bin/arduino-1.8.15/tools-builder -tools /usr/local/bin/arduino-1.8.15/hardware/tools/avr -tools /home/earle/.arduino15/packages -built-in-libraries /usr/local/bin/arduino-1.8.15/libraries -libraries /home/earle/Arduino/libraries -fqbn=esp8266com:esp8266:d1_mini:xtal=80,vt=flash,exception=disabled,stacksmash=disabled,ssl=all,mmu=3232,non32xfer=fast,eesz=4M2M,ip=lm2f,dbg=Serial,lvl=SSLTLS_MEMHTTP_CLIENTHTTP_SERVERCOREWIFIHTTP_UPDATEUPDATEROTAOOMMDNSHWDT_NOEXTRA4K,wipe=none,baud=921600 -vid-pid=1A86_7523 -ide-version=10815 -build-path /tmp/arduino_build_231337 -warnings=all -build-cache /tmp/arduino_cache_721404 -prefs=build.warn_data_percentage=75 -verbose /home/earle/Arduino/sketch_jun17a/sketch_jun17a.ino
Using board 'd1_mini' from platform in folder: /home/earle/Arduino/hardware/esp8266com/esp8266
Using core 'esp8266' from platform in folder: /home/earle/Arduino/hardware/esp8266com/esp8266
/home/earle/.arduino15/packages/esp8266/tools/python3/3.7.2-post1/python3 /home/earle/Arduino/hardware/esp8266com/esp8266/tools/prebuild.py /tmp/arduino_build_231337 /home/earle/Arduino/sketch_jun17a
  File "/home/earle/Arduino/hardware/esp8266com/esp8266/tools/prebuild.py", line 121, in <module>
    sys.exit(main())
  File "/home/earle/Arduino/hardware/esp8266com/esp8266/tools/prebuild.py", line 117, in main
    create_build_options_file(source_path, build_path)
  File "/home/earle/Arduino/hardware/esp8266com/esp8266/tools/prebuild.py", line 62, in create_build_options_file
    build_opt_ctime = get_creation_time(build_opt_build_path)
  File "/home/earle/Arduino/hardware/esp8266com/esp8266/tools/prebuild.py", line 108, in get_creation_time
    return os.path.getctime(file_path)
  File "/usr/lib/python3.6/genericpath.py", line 65, in getctime
    return os.stat(filename).st_ctime
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/arduino_build_231337/sketch/build_opt.h'
exit status 1
Error compiling for board LOLIN(WEMOS) D1 R2 & mini
````

Builds without a `build_opt.h` succeed fine.

**Comment by earlephilhower** [comment]
The contents of the arduino_buid/sketch directory are unpopulated at the time of failure.  Maybe the python is running too soon in the process and needs to be run elsewhere?
````
earle@server:~/Arduino/hardware/esp8266com/esp8266$ ls -lR /tmp/arduino_build_231337/
/tmp/arduino_build_231337/:
total 8
-rw-r--r-- 1 earle earle  910 Jun 17 17:57 build.options.json
drwxrwxr-x 2 earle earle 4096 Jun 17 17:57 sketch

/tmp/arduino_build_231337/sketch:
total 0
````

**Comment by brainelectronics** [comment]
I'll try to check & fix the issue on the weekend

**Comment by d-a-v** [comment]
I had to do this change to make it work under linux:

(`@file` => `-include file`)

I did not try with windows, but without it, I get the same error as with linux.

Then it seems to work well and be used for every single file from core and external libs :)

```diff
diff --git a/platform.txt b/platform.txt
index 555171d4..551e1a14 100644
--- a/platform.txt
+++ b/platform.txt
@@ -115,13 +115,13 @@ recipe.hooks.linking.prelink.3.pattern="{compiler.path}{compiler.c.cmd}" -CC -E
 recipe.hooks.linking.prelink.4.pattern="{compiler.path}{compiler.c.cmd}" -CC -E -P {build.vtable_flags} {build.mmuflags} "{runtime.platform.path}/tools/sdk/ld/eagle.app.v6.common.ld.h" -o "{build.path}/local.eagle.app.v6.common.ld"
 
 ## Compile c files
-recipe.c.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.flags} {compiler.c.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.c.extra_flags} {build.extra_flags} "@{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
+recipe.c.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.flags} {compiler.c.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.c.extra_flags} {build.extra_flags} -include "{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
 
 ## Compile c++ files
-recipe.cpp.o.pattern="{compiler.path}{compiler.cpp.cmd}" {compiler.cpreprocessor.flags} {compiler.cpp.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.cpp.extra_flags} {build.extra_flags} "@{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
+recipe.cpp.o.pattern="{compiler.path}{compiler.cpp.cmd}" {compiler.cpreprocessor.flags} {compiler.cpp.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.cpp.extra_flags} {build.extra_flags} -include "{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
 
 ## Compile S files
-recipe.S.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.flags} {compiler.S.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.S.extra_flags} {build.extra_flags} "@{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
+recipe.S.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.flags} {compiler.S.flags} -D{build.sdk}=1 -DF_CPU={build.f_cpu} {build.lwip_flags} {build.debug_port} {build.debug_level} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch} -DARDUINO_BOARD="{build.board}" {build.led} {build.flash_flags} {compiler.S.extra_flags} {build.extra_flags} -include "{build.opt.path}" {includes} "{source_file}" -o "{object_file}"
 
 ## Create archives
 recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{archive_file_path}" "{object_file}"
```

d-a-v:
I have to explain my last comment.
I realized later that this file had to contain gcc cmdline options `-Dblah=123` or `-fdo-some-optimization`.
So `@` is good, not `-include`.
But here's what I got when trying with this file:
```
-DBLAH
-DBLOH=1
```
It was copied to the tmp dir with this content:
```
#line 1 "/home/gauchard/dev/proj/arduino/8095/build_opt.h"
-DBLAH
-DBLOH=1
```
And that gave this error message when compiling:
```
xtensa-lx106-elf-g++: error: #line: No such file or directory
xtensa-lx106-elf-g++: error: 1: No such file or directory
```

d-a-v:
I had a try with your PR.
`.h` files are preprocessed by `arduino-builder` so I updated your script to:
- use the filename without the `.h` extension,
- copy the file if it exists,
- create/truncate the destination file if source does not exist.

It seems to work locally and this is a great feature.

There is a caveat: the project must be cleared and rebuilt when the option file is changed. That was anyway already the case (everything had to be compiled and not only the sketch). Maybe the python script can do something about that.

Another caveat is that it becomes incompatible with the STM32 version.
Why is the `.h` file not updated by arduino-cli within the STM32 platform ?

Here's my proposal against your current patch.
I removed your date functions but they should be kept if implementing something about the above caveat.

```diff
diff --git a/platform.txt b/platform.txt
index 223c2cf5..48d06a7a 100644
--- a/platform.txt
+++ b/platform.txt
@@ -87,7 +87,7 @@ compiler.size.cmd=xtensa-lx106-elf-size
 build.extra_flags=-DESP8266
 
 # Pre and post build hooks
-build.opt.name=build_opt.h
+build.opt.name=build_opt
 build.opt.path={build.path}/sketch/{build.opt.name}
 
 # Create empty {build.opt} if it does not exist in the output sketch dir
diff --git a/tools/prebuild.py b/tools/prebuild.py
index 3b074ca6..1db9dc20 100644
--- a/tools/prebuild.py
+++ b/tools/prebuild.py
@@ -12,7 +12,7 @@
 
 import os
 import sys
-
+from shutil import copyfile
 
 def create_sketch_dir(build_path):
     """
@@ -26,96 +26,21 @@ def create_sketch_dir(build_path):
         os.makedirs(sketch_build_path)
 
 
-def create_build_options_file(source_path, build_path, build_opt_name="build_opt.h"):
+def create_build_options_file(source_path, build_path, build_opt_name="build_opt"):
     """
-    Create the build options file in the build directory.
-    The modification time of the build options file is set to the sketch
-    modification time in case the file did not exist or the users build options
-    modification time to reflect changes which require a recompilation
-    :param      source_path:     The path to the source directory
-    :type       source_path:     str
-    :param      build_path:      The path to the build directory
-    :type       build_path:      str
-    :param      build_opt_name:  The build option file name
-    :type       build_opt_name:  str, optional
+    Copy the build options file to the build directory,
+    or create an empty file there if source does not exist.
+    Creation or modification time does not matter.
     """
     build_opt_source_path = os.path.join(source_path, build_opt_name)
     build_opt_build_path = os.path.join(build_path, "sketch", build_opt_name)
 
     # check for an existing build options file in the source directory
     if os.path.exists(build_opt_source_path):
-        # user does have/use an build options file in it's sketch directory
-
-        if os.path.exists(build_opt_build_path):
-            # set build options file modification time to the same time as the
-            # sketch modification time to avoid rebuilding libraries
-            set_file_time_to_sketch_time(source_path, build_opt_build_path)
-        else:
-            # the build options file does not yet exist in the build directory
-            # just continue as it will be copied by the IDE automatically on
-            # time. This section is only entered on the very first run
-            pass
+        copyfile(build_opt_source_path, build_opt_build_path)
     else:
-        # user does not have a build options file, create empty one to avoid
-        # compilation error due to missing file for compilation
-        open(build_opt_build_path, 'a').close()
-
-        # set build options file modification time to the same time as the
-        # sketch modification time to avoid rebuilding libraries
-        set_file_time_to_sketch_time(source_path, build_opt_build_path)
-
-
-def set_file_time_to_sketch_time(source_path, destination_path):
-    """
-    Set the file ctime and mtime to the ctime and mtime of the sketch.
-    :param      source_path:       The source path to the sketch
-    :type       source_path:       str
-    :param      destination_path:  The destination path to the file to modify
-    :type       destination_path:  str
-    """
-    sketch_path = get_full_sketch_path(source_path)
-    sketch_mtime = get_modification_time(sketch_path)
-    destination_ctime = get_creation_time(destination_path)
-
-    os.utime(destination_path, (destination_ctime, sketch_mtime))
-
-
-def get_full_sketch_path(source_path):
-    """
-    Get the full sketch path.
-    :param      source_path:  The source path
-    :type       source_path:  str
-    :returns:   The full sketch path including the '.ino' extension.
-    :rtype:     str
-    """
-    base_name = os.path.basename(source_path)
-
-    sketch_path = os.path.join(source_path, base_name + '.ino')
-
-    return sketch_path
-
-
-def get_modification_time(file_path):
-    """
-    Get the modification time of a file.
-    :param      file_path:  The file path
-    :type       file_path:  str
-    :returns:   The modification time.
-    :rtype:     float
-    """
-    return os.path.getmtime(file_path)
-
-
-def get_creation_time(file_path):
-    """
-    Gets the creation time if a file.
-    :param      file_path:  The file path
-    :type       file_path:  str
-    :returns:   The creation time.
-    :rtype:     float
-    """
-    return os.path.getctime(file_path)
-
+        # create empty file
+        open(build_opt_build_path, 'w').close()
 
 def main():
     if len(sys.argv) == 3:
```

mcspr:
About the

> ```
> #line 1 "/home/gauchard/dev/proj/arduino/8095/build_opt.h"
> -DBLAH
> -DBLOH=1
> ```
> 
> And that gave this error message when compiling:
> 
> ```
> xtensa-lx106-elf-g++: error: #line: No such file or directory
> xtensa-lx106-elf-g++: error: 1: No such file or directory
> ```

see https://github.com/arduino/arduino-cli/issues/1338

earlephilhower:
Copying the file directly like @d-a-v suggests may be the best way to get around this.

Alternatively, would it be possible to just post-process the pre-processed header file to do an equivalent  `sed 's/#.*$//` and remove anything on any line after a `#`?

d-a-v:
Both would work for us.
If some sort of unification with STM32 is seeked, then python is needed on their side, or arduino-cli should be fixed regarding @mcspr's comment.
@brainelectronics what do you think ?

fpistm:
>  [that one](https://github.com/stm32duino/Arduino_Core_STM32/pull/1408) is also very nice. It seems stm32 platform relies on bash, so on WSL2 ?

No not on WSL2, we simply have a busybox.exe for windows provided with the tools :wink:

Just FYI, I'm waiting Arduino feedback about this. I would really like they allows some other extension like `.opt` which will be less confusing than `.h`.

d-a-v:
@fpistm 
Did you see the [above comment](https://github.com/esp8266/Arduino/pull/8095#issuecomment-870793783) ?
Would it fit with your platform, or is the `.h` extension old enough to be a breaking change for you ?

fpistm:
> 
> 
> @fpistm
> Did you see the [above comment](https://github.com/esp8266/Arduino/pull/8095#issuecomment-870793783) ?
> Would it fit with your platform, or is the `.h` extension old enough to be a breaking change for you ?

I did not check your proposal. About breaking change well, I always try to avoid it but sometimes we have no choice.
anyway simply change file name could easily be managed without breaking anything.

earlephilhower:
One more vote for not using the .h extension.  It’s not a real header file, so don’t make users think it is.  .opt or something would be very obviously different from the end user perspective.

I’d say there is no backward compatibility issue at all because the compiler flags between different architectures will almost always be different…

d-a-v:
Then @brainelectronics @fpistm @earlephilhower, would `build.opt` be a good compromise ?
@brainelectronics I understand this PR is a way to uniformize this feature accross stm32 and esp8266 arduino platforms, so we do need to reach to an agreement until Arduino provides a more "official" way for doing so when using the IDE, don't we ?

brainelectronics:
@d-a-v fully agree. Naming the file `build.opt` would be a good idea. Copying that file maybe as `build.opt.h` to the sketch compilation folder would be done by the script and thereby not confusing the user with a header file name. 
Of course I would like to reach a unification with STM32duino package, so users can use the build options in both worlds.
I'll also update the docs about @mcspr's issue of using multiple arg lines in the file

brainelectronics:
> I had a try with your PR.
> `.h` files are preprocessed by `arduino-builder` so I updated your script to:

Basically that described issues should have been solved with the timestamp manipulation. At least that was the intension. I'll try to check again next week as I'm currently quite busy with another project

fpistm:
@d-a-v 
I've made a PR to fix the issue brought by arduino-cli. See https://github.com/stm32duino/Arduino_Core_STM32/pull/1442
Very simple and backward compatible. 😉

d-a-v:
@brainelectronics 
Regarding your current PR, do you agree with @fpistm 's changes in STM32 repo and adopt the new naming `build.opt` instead of `build_opt.h` for the file containing gcc cmdline options ?
It would be more easy to understand (`.h` files usually contains C source code) and would be a nice workaround for the issue introduced by `arduino-builder` (and the future ones regarding source files `.h` management).

d-a-v:
@fpistm 's https://github.com/stm32duino/Arduino_Core_STM32/pull/1442 is merged,
so what's the status of @brainelectronics 's PR - this PR ?

mhightower83:
@brainelectronics If you are stuck and are open to some new ideas, here are my rambling thoughts.

I am probably missing some things. I'll try and summarize and meld together various ideas that I have.

I noted that `.h` files are easy to edit with the Arduino IDE and are handled well. When a save-as is performed, all project files are copied. However, `build.opt` would be missing in the new saved sketch.

The idea I have is to add support for a `sketch.ino.globals.h` file, using @d-a-v 's file naming suggestion.
This file would have an optional "C" comment block that would encapsulate the `build.opt` details.
Something like
```cpp
/*@create-file:build.opt@
 *
-DMYDEFINE="\"Chimichangas do not exist\""
-O3
-fanalyzer
 *
 */
```
The comment block encapsulation idea is not mine, I read the suggestion somewhere, several years ago.

To reduce interference from the IDE, store `build.opt` and globals files in a unique directory like `custom`.
The python script will copy/update `sketch.ino.globals.h` when the source directory version changes.
The python script would extract the embedded block into the `build.opt` file.
Use the `-include ...` to create a dependency for the `sketch.ino.globals.h` file.
Any time `sketch.ino.globals.h` is changed a new `build.opt` would be extracted by the script.

An adaptation of your script with these ideas can be found here:
https://gist.github.com/mhightower83/0a26fcbf4683aee4b67c87ba56829366
If it works for you, you are more than welcome to use it in part or whole for this PR.

You can ignore the `sketchbook global` stuff, that is extra stuff I needed while I was using it.

mhightower83:
Never mind, I seem to be out of sync with all the other threads out there, enough so that I am confused as to what they have decided and where it is implemented. In any event, the example I did works; however, it may not be the direction everyone is going. Sorry for the noise.

mhightower83:
## Focusing on PR#1442
@brainelectronics  If you are stuck or have just been too busy, here is my interpretation of the immediate contents of https://github.com/stm32duino/Arduino_Core_STM32/pull/1442#issue-942145032. And, how to resolve some of the shortfalls. 
>I say "immediate contents" because I get lost trying to see beyond that PR. There are a lot of additional thoughts and discussions out there concerning expanding support for build options on the Arduino IDE; however, those discussions have been ongoing for many years. My guess is they will continue for many more.

Observations of `build_opt.h` in PR#1442
* To hold the command-line options, the sketch folder optionally contains the `build_opt.h` file.
* It must not contain any "C" code and must not be referenced by any `#include` statements. 
* At build time, `build_opt.h` in the _source_ sketch folder is copied to the _build_ sketch folder as **`build.opt`**. 
* There is no global dependency to respond to `build_opt.h` changes.
* If you change `build_opt.h`, you may need to restart the Arduino IDE to get a good build.
* After global dependencies are addressed, need to address the issue of the default, "Aggressively cache compiled core".
* While having non "C" code in a `.h` is a bit odd, the file having a `.h` extension allows you to edit it with the Arduino-IDE.

An easy solution for a global dependency is to create a proxy .h file and use @d-a-v 's `-include ...` suggestion. Updated the proxy file each time a new version of `build.opt` is written.

To confirm, that the ideas expressed above work, I updated your script, expanded on the thoughts in PR#1442, and added dependency support.

<details>
  <summary>Revised script:</summary>

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
#
# prebuild.py — This script manages updates to build.opt in the build directory
# with changes from build_opt.h in the Sketch directory.
#
# Copy build_opt.h from the source sketch folder to build.opt in the build
# sketch folder. If source sketch folder does not have a build_opt.h file,
# create am empty build_opt.h files in the build sketch folder.
#
# For dependencies to work, maintain timestamp on a proxy .h file for build.opt.
#
# Written by brainelectronics, 2021.
#
# Update by M Hightower, 2022.
#
"""
"Aggressively cache compiled core" must be turned off for a reliable build process.
In ~/.arduino15/preferences.txt, to disable the feature:
  compiler.cache_core=false

Reference:
https://forum.arduino.cc/t/no-aggressively-cache-compiled-core-in-ide-1-8-15/878954/2
"""

"""
# Updates or Additions for platform.txt or platform.local.txt

runtime.tools.prebuild={runtime.platform.path}/tools/prebuild.py

build.opt.source.path={build.source.path}/build_opt.h
build.opt.path={build.path}/sketch/build.opt
build_opt.proxy.path={build.path}/sketch/PROXY.build_opt.h

recipe.hooks.prebuild.2.pattern="{runtime.tools.python3.path}/python3" "{runtime.tools.prebuild}" "{build.opt.source.path}" "{build.opt.path}" "{build_opt.proxy.path}"

compiler.cpreprocessor.flags=-D__ets__ -DICACHE_FLASH -U__STRICT_ANSI__ -D_GNU_SOURCE @{build.opt.path} -include "{build_opt.proxy.path}" "-I{compiler.sdk.path}/include" "-I{compiler.sdk.path}/{build.lwip_include}" "-I{compiler.libc.path}/include" "-I{build.path}/core"
"""
import os
import sys
from shutil import copyfile

def main():
    if len(sys.argv) == 4:
        src_build_opt_path = sys.argv[1]
        dst_build_opt_path = sys.argv[2]
        dst_build_opt_proxy_path = sys.argv[3]

        build_path_sketch, not_used = os.path.split(dst_build_opt_path)

        if not os.path.exists(build_path_sketch):
            os.makedirs(build_path_sketch)

        # Conditionally copy a newer build_opt.h from source directory to the
        # build/sketch directory as build.opt. When build_opt.h does not exist
        # in the source  directory, create an empty build.opt file in the
        # build/sketch directory.
        if os.path.exists(src_build_opt_path):
            if (os.path.exists(dst_build_opt_path)) and \
            (os.path.getmtime(dst_build_opt_path) >= os.path.getmtime(src_build_opt_path)):
                # only copy newer file - do nothing, all is good
                print("Using sketch compiler command-line options in " + dst_build_opt_path)
            else:
                # The new copy gets stamped with the current time, just as other
                # files copied by `arduino-builder`.
                copyfile(src_build_opt_path, dst_build_opt_path)
                open(dst_build_opt_proxy_path, 'w').close()
                print("Updated sketch compiler command-line options in " + dst_build_opt_path)
        else: # The Sketch currently does not have a build_opt.h.
            if os.path.exists(dst_build_opt_path) and \
            os.path.getsize(dst_build_opt_path) == 0:
                # All done, the Sketch has not changed.
                pass
            else:
                # Place holder - Create an empty `build.opt` to satisfy the
                # `@file` parameter. Create proxy .h to track changes to
                # `build.opt`. This will help trigger dependency builds.
                open(dst_build_opt_path, 'w').close()
                open(dst_build_opt_proxy_path, 'w').close()

    else:
        print("Wrong argument count")
        print("  All arguments FQFN. Source path build_opt.h, Build path build.opt, Build path PROXY.build_opt.h")
        sys.exit(1)

if __name__ == '__main__':
    sys.exit(main())
```
</details>

## Aggressive Caching of `core.a`
A compiler command-line option file and the **Aggressively cache compiled core** feature do not work well together.
For the case of `build.opt` in the build sketch folder, `build.opt` can contain defines that require `core.a` to be recompiled. When multiple Sketches are open, they can no longer reliably share one cached `core.a`. (Looks like) The 1st Sketch to be built has its version of `core.a` cached. Other sketches will get this cached version for their builds.

To see the issue:
1. Close all Arduino IDE windows and do a fresh start of Arduino IDE.
2. Do _File->New_, then click on _Verify_. Wait for the build to complete.
3. Create another new sketch add this example code:
```cpp
void setup() {
  Serial.begin(115200);
  delay(200);
  Serial.printf("\r\nRunning on empty\r\n\r\n");
  Serial.printf("Heap Low Watermark %u\r\n", umm_free_heap_size_min());
}
void loop() {}
```
4. Add _New Tab_ `build_opt.h` with the following:
```
-DUMM_STATS_FULL=1
```
5. Click on _Verify_. The build will fail.

To use a compiler command-line option file reliably, you need to set `compiler.cache_core=false`. Then each sketch will maintain its _own_ copy of `core.a`. The alternative when using `compiler.cache_core=true`, is to close all Arduino IDE sketch windows. Start and run _only_ one instance of the IDE, while building a Sketch that uses `build.opt`.

PR #8104, opened 2021-06-06
Update toolchain to gcc 10.3 w/patches

earlephilhower:


d-a-v (approved):
Like the previous one: under windows second build is faster.
This time I also flashed an esp from windows and it works.

Also tested with linux (compile twice + flash, running OK)

dok-net:
A project using WiFi, which before took really long to compile for any small change in the sketch, and an external lib.
With 3.0.0-17-g114a7267 : first clean build 1:05m, second without any changes 0:40m
With 3.0.0-19-ge8b411fa : first clean build 1:05m, second without any changes 0:14m
(Older octacore Windows 10 PC with HDD)
Way to go!

PR #8116, opened 2021-06-11
Prevent abs() macro from sign-flipping 0.0F

### LINKED ISSUE #8115: abs() macro in Arduino.h inverts signedness of (positive) 0.0F
With regard to https://github.com/arduino/ArduinoCore-avr/pull/406,
I found out that the C macro defined in our `Arduino.h`, flips the signedness of 0.0F to -0.0F. I think it would be more natural, if one has to make a choice, to let the error (if it is considered as such) to let abs() return (negative) -0.0F for the input (negative) -0.0F but not flip the signedness of the positive 0.0F.

MCVE, abstest is an identical copy of the `abs()` macro in `Arduino.h`:
```
float one = 1.0F;
float minusone = -1.0F;
float zero = 0.0F;
float minuszero = -0.0F;

#define abstest(x) ({ __typeof__(x) _x = (x); _x > 0 ? _x : -_x; })

void setup()
{
	Serial.begin(74880);
	delay(500);
	Serial.println("abstest");
	Serial.print("one = "); Serial.println(*reinterpret_cast<uint32_t*>(&one));
	Serial.print("minusone = "); Serial.println(*reinterpret_cast<uint32_t*>(&minusone));
	Serial.print("zero = "); Serial.println(*reinterpret_cast<uint32_t*>(&zero));
	Serial.print("minuszero = "); Serial.println(*reinterpret_cast<uint32_t*>(&minuszero));

	float res;
	res = abstest(one);
	Serial.print("abs(one) = "); Serial.println(*reinterpret_cast<uint32_t*>(&res));
	res = abstest(minusone);
	Serial.print("abs(minusone) = "); Serial.println(*reinterpret_cast<uint32_t*>(&res));
	res = abstest(zero);
	Serial.print("abs(zero) = "); Serial.println(*reinterpret_cast<uint32_t*>(&res));
	res = abstest(minuszero);
	Serial.print("abs(minuszero) = "); Serial.println(*reinterpret_cast<uint32_t*>(&res));
}

void loop()
{
}
```

Output:
```
abstest
one = 1065353216
minusone = 3212836864
zero = 0
minuszero = 2147483648
abs(one) = 1065353216
abs(minusone) = 1065353216
abs(zero) = 2147483648
abs(minuszero) = 0
```

dok-net:
Fixes #8115


dok-net:
This variant solves the `abs(-0.0F) := -0.0F` problem, at the expense of a few CPU cycles but also 32 bytes code:
```
#define abstest(x) ({ __typeof__(x) _x = (x); _x >= 0 ? _x + 0 : -_x; })
```

earlephilhower:
In the original PR in the AVR, I fail to see the actual bug they're solving.  If `-0.0f == 0.0f` then why would one care?

Since this is a macro (thanks, Arduino) it seems like it will add 32 bytes per call, everywhere, with the change, which seems large.

dok-net:
@earlephilhower Within a single compilation unit, the compiler seems to optimize that overhead. But as you can tell from me not putting into the PR, instead just giving it a showing in a remark, I don't see much need for that myself. As for this PR, I find some merit in having all-zero bits representation at least for the positive zero, such that (uint32_t)0 and (float)0 are the same in memory. For what it's worth :-)

mcspr:
> @mcspr This is not as much about a more perfect ab() function, as it is about being Arduino-alike.

then, I'd agree with the comment above - no point in fixing anything, as neither abs or round macros are provided in the current ArduinoCore API headers in favour of math.h:
https://github.com/arduino/ArduinoCore-API/issues/76#issuecomment-781132117

and current macro can stay as-is, until the issue above is definitely closed and there's a clear answer what arduino side want to do with them

Issue #8049, opened 2021-05-18
Incomplete core_version.h from 3.0.0 release .zip


mcspr:
Current archive contains these:
```cpp
#define ARDUINO_ESP8266_GIT_VER 0xefb0341a
#define ARDUINO_ESP8266_GIT_DESC 3.0.0
#define ARDUINO_ESP8266_RELEASE_
#define ARDUINO_ESP8266_RELEASE ""
```

which should probably have been `ARDUINO_ESP8266_RELEASE_3_0_0` and `ARDUINO_ESP8266_RELEASE "3_0_0"`?

d-a-v:
This has been fixed by #8126 and #8127

Next 3.0.1 release will have
```cpp
#define ARDUINO_ESP8266_GIT_VER   0xA11900D
#define ARDUINO_ESP8266_GIT_DESC  3.0.1
#define ARDUINO_ESP8266_VERSION   3.0.1

#define ARDUINO_ESP8266_MAJOR     3
#define ARDUINO_ESP8266_MINOR     0
#define ARDUINO_ESP8266_REVISION  1

#define ARDUINO_ESP8266_RELEASE   "3.0.1"
#define ARDUINO_ESP8266_RELEASE_3_0_1
```

PR #8126, opened 2021-06-15
Release process fix & major/minor/rev macro addition

d-a-v:
- Release process update is based on #8034.
- `makecorever.py` is now used as part of the release process script.
- `MAJOR`/`MINOR`/`REVISION` are added in `core_version.h` based on `platform.txt`'s `version` when building with Arduino-CLI, or git tag when releasing. PlatformIO will have to be updated too (this is not addressed here).

This is WIP but is going to be merged anyway in order to try the release script executed remotely by CI.
A tag will be added (3.0.1) and removed right after that (this gymnastic happens quite often while releasing these days).
This will allow to check generated logs and archive to address #8049

### DIFF
--- a/package/build_boards_manager_package.sh
+++ b/package/build_boards_manager_package.sh
@@ -110,11 +110,12 @@ $SED -E "s/name=([a-zA-Z0-9\ -]+).*/name=\1(${ver})/g"\
 
 # Put core version and short hash of git version into core_version.h
 ver_define=`echo "${plain_ver}" | tr "[:lower:]." "[:upper:]_"`
-echo "ver_define: ${ver_define}  (plain_ver: ${plain_ver})"
-echo "#define ARDUINO_ESP8266_GIT_VER 0x`git rev-parse --short=8 HEAD 2>/dev/null`" >${outdir}/cores/esp8266/core_version.h
-echo "#define ARDUINO_ESP8266_GIT_DESC `git describe --tags 2>/dev/null`" >>${outdir}/cores/esp8266/core_version.h
-echo "#define ARDUINO_ESP8266_RELEASE_${ver_define}" >>${outdir}/cores/esp8266/core_version.h
-echo "#define ARDUINO_ESP8266_RELEASE \"${ver_define}\"" >>${outdir}/cores/esp8266/core_version.h
+#echo "ver_define: ${ver_define}  (plain_ver: ${plain_ver})"
+#echo "#define ARDUINO_ESP8266_GIT_VER 0x`git rev-parse --short=8 HEAD 2>/dev/null`" >${outdir}/cores/esp8266/core_version.h
+#echo "#define ARDUINO_ESP8266_GIT_DESC `git describe --tags 2>/dev/null`" >>${outdir}/cores/esp8266/core_version.h
+#echo "#define ARDUINO_ESP8266_RELEASE_${ver_define}" >>${outdir}/cores/esp8266/core_version.h
+#echo "#define ARDUINO_ESP8266_RELEASE \"${ver_define}\"" >>${outdir}/cores/esp8266/core_version.h
+python3 ${srcdir}/tools/makecorever -b ${outdir} -i cores/esp8266 -p ${srcdir} -v ${ver_define}
 
 # Zip the package
 pushd package/versions/${visiblever}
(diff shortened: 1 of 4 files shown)


tools/makecorever.py
```diff
@@ -20,23 +20,40 @@
...
 
-def generate(path, platform_path, git_ver="ffffffff", git_desc="unspecified"):
+def generate(path, platform_path, git_ver="ffffffff", platform_version="unspecified"):
     def git(*args):
         cmd = ["git", "-C", platform_path]
         cmd.extend(args)
         proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True, stderr=subprocess.DEVNULL)
         return proc.stdout.readlines()[0].strip()
 
+    git_desc = platform_version;
```
mcspr, line 33: Following the above, should this have additional param to skip git-describe?
Meaning, this still could still work with a shallow copy or with some broken tagging (like github actions checkout does)

tools/makecorever.py
```diff
@@ -20,23 +20,40 @@
...
+    text += "#define ARDUINO_ESP8266_MINOR     {}\n".format(version[1])
+    # revision can be ".n" or ".n-dev"
+    revision = re.split("-", version[2])
+    text += "#define ARDUINO_ESP8266_REVISION  {}\n".format(revision[0])
+    text += "\n"
+    if len(revision) > 1:
+        text += "#define ARDUINO_ESP8266_DEV       1 // developpment version\n"
+    else:
+        text += "#define ARDUINO_ESP8266_RELEASE   \"{}\"\n".format(git_desc)
+        text += "#define ARDUINO_ESP8266_RELEASE_{}\n".format(re.sub("[-\.]", "_", git_desc))
```
mcspr, line 56: also, re. `re`
python has string methods like `"1.2.3".split(".")` and `"1.2.3".replace(".", "_")`
also note of the `"1.2.3-dev".partition("-")`

package/build_boards_manager_package.sh
```diff
@@ -109,12 +109,13 @@ $SED -E "s/name=([a-zA-Z0-9\ -]+).*/name=\1(${ver})/g"\
...
-echo "#define ARDUINO_ESP8266_GIT_DESC `git describe --tags 2>/dev/null`" >>${outdir}/cores/esp8266/core_version.h
-echo "#define ARDUINO_ESP8266_RELEASE_${ver_define}" >>${outdir}/cores/esp8266/core_version.h
-echo "#define ARDUINO_ESP8266_RELEASE \"${ver_define}\"" >>${outdir}/cores/esp8266/core_version.h
+#ver_define=`echo "${plain_ver}" | tr "[:lower:]." "[:upper:]_"`
+#echo "ver_define: ${ver_define}  (plain_ver: ${plain_ver})"
+#echo "#define ARDUINO_ESP8266_GIT_VER 0x`git rev-parse --short=8 HEAD 2>/dev/null`" >${outdir}/cores/esp8266/core_version.h
+#echo "#define ARDUINO_ESP8266_GIT_DESC `git describe --tags 2>/dev/null`" >>${outdir}/cores/esp8266/core_version.h
+#echo "#define ARDUINO_ESP8266_RELEASE_${ver_define}" >>${outdir}/cores/esp8266/core_version.h
+#echo "#define ARDUINO_ESP8266_RELEASE \"${ver_define}\"" >>${outdir}/cores/esp8266/core_version.h
+python3 ${srcdir}/tools/makecorever.py -b ${outdir} -i cores/esp8266 -p ${srcdir} -v ${plain_ver}
```
mcspr, line 118: Should the script have an explicit option to toggle the release mode, so we don't have to auto-detect?

tools/makecorever.py
d-a-v: Tagging works OK including when used by github-action. Released are based on tagging.
I updated the script per your 3 reviews and added comments.

mcspr: broken == checkout doing `git fetch origin +$commit-sha$:refs/tags/$tag-name$`, which is worked around with fetch-depth:0... which could be avoided

my only problem with that is that version string depends on too much stuff, where we already know it from the tag name anyway; and commit sha comes from rev-parse regardless, shallow copy or not.

PR #8136, opened 2021-06-17
WiFiClientSecure owns context, fix stopAllExcept

### LINKED ISSUE #8079: [ESP8266HTTPClient] Updates from HTTPS servers fail before transferring new binary begins
http update not working on this arduino-core version

HTTP_UPDATE_FAILD Error (-106): Verify Bin Header Failed;
I just instaled arduino core 2.7.4 and it worked.

NodeMCU 1.0, Arduino 1.8.15, code

[httpUpdateSecure](https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266httpUpdate/examples/httpUpdateSecure)

earlephilhower:
Fixes #8079

Because WiFiClientSecure inherits WiFiClient, and WiFiClientSecureCtx also
inherits WiFiClient, they both end up in the list of TCP connections that
are used for WiFiClient::stopAllExcept().  This would cause the underlying
SSL connection to be closed whenever you attempted to
stopAllExcept(WiFiClientSecure)

Fix by adding a "_owner" pointer in the WiFiClient object which points to
nullptr (default case) or to the associated upper-layer connection.
When stopping all connections except one, only look at the uppermost
connections.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClient.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiClient.cpp
@@ -77,14 +77,14 @@ WiFiClient* SList<WiFiClient>::_s_first = 0;
 
 
 WiFiClient::WiFiClient()
-: _client(0)
+: _client(0), _owner(0)
 {
     _timeout = 5000;
     WiFiClient::_add(this);
 }
 
 WiFiClient::WiFiClient(ClientContext* client)
-: _client(client)
+: _client(client), _owner(0)
 {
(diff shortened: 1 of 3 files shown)


earlephilhower:
@d-a-v, I think I need to reverse the sense here and look at the lowest-level, not uppermost one.  So hold off on any review or merge, please.

earlephilhower:
Okay, good to go @d-a-v.  Logic reversed and tested locally.

PR #8137, opened 2021-06-17
Change protocol detection so uppercase or lowercase works

idolpx:
Sometimes a url is specified in all upper case and HTTPClient rejects it when it would be perfectly fine.  This just makes the protocol detection case insensitive so the url is accepted if the protocol is uppercase.


earlephilhower (approved):
Thanks, LGTM

Issue #8139, opened 2021-06-17
ESP8266NetBIOS library only works in wifi STA mode, and not in wifi AP


pabloandresm:
Title says it all. The problem is the same as the NetBIOS library for esp32 (https://github.com/espressif/arduino-esp32/issues/5300)

If you look at the code, file ESP8266NetBIOS.cpp, line 215, in the netbios answer, you can see:
	nbnsa.NBNSA_NODEADDRESS = WiFi.localIP(); // ulozime nasi IP adresu

Which works only if WiFi is in STA mode, but not AP mode.

The fix would be something like this:
	if (WiFi.getMode()==WIFI_STA)
		nbnsa.NBNSA_NODEADDRESS = WiFi.localIP(); // ulozime nasi IP adresu
	else
		nbnsa.NBNSA_NODEADDRESS = WiFi.softAPIP(); // ulozime nasi IP adresu

earlephilhower:
@pabloandresm you've found a problem and potential solution.  Why not try your hand at a [pull request](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests) so your work would be preserved in the `git` logs?

pabloandresm:
I'll do that right now. Thanks.

PR #8140, opened 2021-06-17
fix 8139 netbios in AP mode

pabloandresm:
fix for NetBIOS only working if WIFI in STA mode and not AP mode

https://github.com/esp8266/Arduino/issues/8139

### DIFF
--- a/libraries/ESP8266NetBIOS/ESP8266NetBIOS.cpp
+++ b/libraries/ESP8266NetBIOS/ESP8266NetBIOS.cpp
@@ -212,7 +212,10 @@ void ESP8266NetBIOS::_recv(udp_pcb *upcb, pbuf *pb, const ip_addr_t *addr, uint1
                     nbnsa.NBNSA_TIMETOLIVE = LWIP_PLATFORM_HTONL(300000UL);// Time to live (30000 sekund)
                     nbnsa.NBNSA_LENGTH = LWIP_PLATFORM_HTONS(6);
                     nbnsa.NBNSA_NODEFLAGS = LWIP_PLATFORM_HTONS(0);
-                    nbnsa.NBNSA_NODEADDRESS = WiFi.localIP(); // ulozime nasi IP adresu
+					if (WiFi.getMode()==WIFI_STA)
+						nbnsa.NBNSA_NODEADDRESS = WiFi.localIP(); // ulozime nasi IP adresu
+					else
+						nbnsa.NBNSA_NODEADDRESS = WiFi.softAPIP(); // ulozime nasi IP adresu
 
                     pbuf* pbt = pbuf_alloc(PBUF_TRANSPORT, sizeof(nbnsa), PBUF_RAM);
                     if(pbt != NULL) {


earlephilhower:
@pabloandresm, I don't think any maintainers use the NetBIOS lib.  Have you personally tried this change with AP mode?

libraries/ESP8266NetBIOS/ESP8266NetBIOS.cpp
```diff
@@ -212,7 +212,10 @@ void ESP8266NetBIOS::_recv(udp_pcb *upcb, pbuf *pb, const ip_addr_t *addr, uint1
                     nbnsa.NBNSA_TIMETOLIVE = LWIP_PLATFORM_HTONL(300000UL);// Time to live (30000 sekund)
                     nbnsa.NBNSA_LENGTH = LWIP_PLATFORM_HTONS(6);
                     nbnsa.NBNSA_NODEFLAGS = LWIP_PLATFORM_HTONS(0);
-                    nbnsa.NBNSA_NODEADDRESS = WiFi.localIP(); // ulozime nasi IP adresu
+					if (WiFi.getMode()==WIFI_STA)
+						nbnsa.NBNSA_NODEADDRESS = WiFi.localIP(); // ulozime nasi IP adresu
+					else
+						nbnsa.NBNSA_NODEADDRESS = WiFi.softAPIP(); // ulozime nasi IP adresu
```
d-a-v, line 218: I just realized that it is an old and partial reimplementation of UdpContext, not aware of IPv6.
That is OK because [NetbiosNS/udp137 does not support IPv6](https://en.wikipedia.org/wiki/NetBIOS).

This code is better but still improvable.
Can you please try with
```
nbnsa.NBNSA_NODEADDRESS = ip4_addr_get_u32(ip_2_ip4(ip_current_dest_addr())); // `ip_current_dest_addr()` is usable inside lwIP's UDP receive callback 
```
This address is the IPv4 destination address of the received packet that is replied to, so it is the ESP address on the right interface.
Hence it should work on AP, STA, AP+STA when coming from AP, AP+STA when coming from STA, ethernet, PPP.

pabloandresm: ok. I will give it a try and let you know...both in esp32 and esp8266 bugs

d-a-v: The proposed fix uses lwIP-v2 API that is also used in esp32 Arduino core, and valid whether IPv6 is enabled or not.

d-a-v: @pabloandresm Did you have a chance to give it a try ?

d-a-v:
pushing to v4 per still in discussion

libraries/ESP8266NetBIOS/ESP8266NetBIOS.cpp
pabloandresm: Hello there.

I've tried the fix with arduino ide esp32 v2.0.2 and v2.0.3 based on ESP-IDF 4.4.1, and it works perfectly.

To sum up, 1st: the fix returns the IP of STA if in STA mode. If not, then it returns the IP of AP mode.

Without the fix, the IP of AP mode is always returned regardless in which mode we are.

Pablo

d-a-v (approved):

mcspr:
> pushing to v4 per still in discussion
> 
> _edit_: Approving per feedback. Changes look good, maybe not fully aware of all network states (ap+sta?), also not ethernet aware. Netbios users will report back in case of further issues.

why not `upcb->local\_ip` though? we do write a low level code interacting with the lwip here, why mix in WiFi class at all?

d-a-v:
@mcspr As no-one can make and test a more robust change, I suggest, as part of the current freezing process, to merge this bugfix before of after adding this comment
```cpp
// Note from maintainers: for further improvements, see
// see https://github.com/esp8266/Arduino/pull/8140#discussion_r654549861
// and https://github.com/esp8266/Arduino/pull/8140#issuecomment-1159587969
if (WiFi.getMode()==WIFI_STA)
...
```

mcspr:
@d-a-v let me check the change and get back to you... it also seems to generate a malformed packet

PR #8150, opened 2021-06-20
Fix deprecation of bzero, prefer POSIX over GNU newlib extensions

### LINKED ISSUE #8122: strptime dissapeared on 3.0.0
Arduino last IDE, esp8266 just updated to 3.0.0, and I'm getting undefined error for strptime. After playing with and without  XSI_VISIBLE, searching the includes, trying with and without time.h and timelib.h, I'm doing a workaround (parsing the old way) and letting you know.

dok-net:
Fixes #8122.
Bzero was once deprecated, apparently compiler optimizations could remove the zeroing, which for obvious reasons is a serious security issue.


dok-net:
ESP32 build settings are generated from IDF, and there _GNU_SOURCE is used, too. So the whole point is moot. Retracting.

PR #8153, opened 2021-06-21
Add missing menu.UploadTool name definition

silvanocerza:
Hello!

We recently merged Arduino/arduino-cli#1326 on the Arduino CLI, it introduces an enforcement of the platform specifications for [custom board options][custom-board-options]. Boards that define some custom options without their name defined somewhere else in the `boards.txt` won't be loaded, the rest will be loaded without issues if they respect the specifications.

A user reported in Arduino/arduino-cli#1329 that `esp8266:esp8266:espduino` is not loaded because of this change, this PR will fix that issue.

That change has not been released but is only available in the nightly, so most users are not affected as of now but will in the next release. 

[custom-board-options]: https://arduino.github.io/arduino-cli/latest/platform-specification/#custom-board-options

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -6,6 +6,7 @@
 
 menu.BoardModel=Model
 menu.ESPModule=Module
+menu.UploadTool=Upload Tool
 menu.led=Builtin Led
 menu.baud=Upload Speed
 menu.xtal=CPU Frequency
--- a/tools/boards.txt.py
+++ b/tools/boards.txt.py
@@ -1621,6 +1621,7 @@ def all_boards ():
     # With Arduino IDE 1.8.7 the order of the menu items will be honored from the tools pull down list.
     print('menu.BoardModel=Model')
     print('menu.ESPModule=Module')
+    print('menu.UploadTool=Upload Tool')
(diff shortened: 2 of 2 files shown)


d-a-v:
You need to update [this file](https://github.com/esp8266/Arduino/blob/45e7976c50ef33d059e7b78384d0ffd519dbae25/tools/boards.txt.py#L1623) then run the script:
```
./tools/board.txt.py --allgen
```
and commit the changed files (in this case it should only be `boards.txt` *and* `tools/boards.txt.py`.

Or we can do it for you, as you prefer.

CI did not pass because the same process is applied (which removes your change) and compared.

silvanocerza:
@d-a-v no worries, I fixed it. Didn't notice the `boards.txt` was generated automatically.

d-a-v (approved):
Thanks!

patoberli:
Would it be possible to backport this to 2.7.4? It seems this change is required for Arduino IDE 2.x support. For more information see the thread in the Arduino forum: https://forum.arduino.cc/t/ide-2-0-beta-9-not-working-wemos-d1-mini-esp8266/891349/4

PR #8165, opened 2021-06-23
Add errors on invalid/missing function return type

earlephilhower:
GCC 10.x seems to have a knack for crashing when a function which is declared
to return a value does not.  Add a warning, present on all builds, when this
is the case.  For more info see
https://github.com/esp8266/Arduino/discussions/8160

Thanks to @hreintke for the tip

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -21,11 +21,11 @@ runtime.tools.mkdir={runtime.platform.path}/tools/mkdir.py
 runtime.tools.cp={runtime.platform.path}/tools/cp.py
 runtime.tools.eboot={runtime.platform.path}/bootloaders/eboot/eboot.elf
 
-compiler.warning_flags=-w
-compiler.warning_flags.none=-w
-compiler.warning_flags.default=
-compiler.warning_flags.more=-Wall
-compiler.warning_flags.all=-Wall -Wextra
+compiler.warning_flags=-w -Werror=return-type
+compiler.warning_flags.none=-w -Werror=return-type
+compiler.warning_flags.default=-Werror=return-type
+compiler.warning_flags.more=-Wall -Werror=return-type
+compiler.warning_flags.all=-Wall -Wextra -Werror=return-type
 
 build.lwip_lib=-llwip_gcc
 build.lwip_include=lwip/include
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -62,6 +62,7 @@ env.Append(
         "-std=gnu17",
         "-Wpointer-arith",
         "-Wno-implicit-function-declaration",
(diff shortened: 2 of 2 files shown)


d-a-v (approved):

dok-net:
This warning is basically a must-have on the ESPs. Is there a way to treat it as an error? Because that what it really is on this architecture, returning from non-void function without a return statement.

mcspr:
> This warning is basically a must-have on the ESPs. Is there a way to treat it as an error? Because that is what it really is on this architecture, returning from non-void function without a return statement.

Notice it is not just -W, but -Werror. Turns out gcc can selectively pick warnings to be treated as errors

dok-net:
@me-no-dev In reference to the now closed discussion about https://github.com/espressif/arduino-esp32/pull/5310, would you see any value in this and get it inserted into the right places, instead of the simple warning, that I am already getting for such misuses (`warning: no return statement in function returning non-void [-Wreturn-type]`)?

me-no-dev:
@dok-net I will accept the same change in the ESP32 repo 😄

dok-net:
@me-no-dev Before moving over there (sorry ESP8266 guys for talking about ESP32 here), you recently said:
> all of the changes that you have made are to files that are auto generated by the lib builder. They will not stick. The lib builder get's the flags from IDF's compile process.

The files were `platform.txt`, and `tools/platformio-build-esp32*.py`.
Predictably, now I am confused, shall I PR a change to these files in the ESP32 repo or not? Mind you, I am not familiar with "the lib builder" or the IDF.

me-no-dev:
just `-Werror=return-type` can be added to platform.txt. Nothing else from your previous PR

dok-net:
@earlephilhower This is from ESP32,
```
compiler.warning_flags.more=-Wall -Werror=all
compiler.warning_flags.all=-Wall -Werror=all -Wextra
```

What do you think about `-Werror=all`and `-Wextra` for `flags.more` and `flags.all`, respectively? I can't image there is any difference between ESP8266 and ESP32 that requires this difference? Of course this is before this PR, but I'm just about to duplicate that to ESP32, so I could use some guidance on whether to make this the same.

mcspr (commented):
Also need to update tools/platformio-build.py
https://github.com/esp8266/Arduino/blob/d3136738ad2eb96ba40675f4b9d83f6da8cf32a1/tools/platformio-build.py#L61-L87

I guess CCFLAGS is the correct choice, based on underlying build system docs:
https://scons.org/doc/latest/HTML/scons-user/apa.html
> CCFLAGS
> General options that are passed to the C and C++ compilers.

PR #8168, opened 2021-06-23
github action: enforce issue detection with bash 

d-a-v:
Per discussion with @mcspr.

When not specifiying `bash` as running shell, default options do not include `-o pipefail` [as it should](https://github.com/actions/runner/issues/353).
It may not prevent a script to stop when one component of a pipe chain fails (ex: `false | true` won't fail but it should).
As a consequence in this PR, `bash` is [explicitely specified as default shell](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-2).

Also enforcing issue detection by breaking down commands
`echo "$(command)" >> file`  =>  `var=$(command)` then `echo ${var} >> file`

### DIFF
--- a/.github/workflows/tag-to-draft-release.yml
+++ b/.github/workflows/tag-to-draft-release.yml
@@ -28,8 +28,8 @@ jobs:
     - name: Set GIT tag name
       run: |
         # Sets an environment variable used in the next steps
-        TRAVIS_TAG=$(git describe --exact-match --tags)"
-        echo "TRAVIS_TAG=${TRAVIS_TAG}" >> $GITHUB_ENV
+        TRAVIS_TAG="$(git describe --exact-match --tags)"
+        echo "TRAVIS_TAG=\"${TRAVIS_TAG}\"" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/tag-to-draft-release.yml
```diff
@@ -25,7 +28,8 @@ jobs:
     - name: Set GIT tag name
       run: |
         # Sets an environment variable used in the next steps
-        echo "TRAVIS_TAG=$(git describe --exact-match --tags)" >> $GITHUB_ENV
+        TRAVIS_TAG=$(git describe --exact-match --tags)"
```
mcspr, line 31:
```suggestion
        TRAVIS_TAG=$(git describe --exact-match --tags)
```
(I typed this wrong)

d-a-v: Thanks for reviewing (I didn't make a copy-paste, It's simply that I made the same mistake)

earlephilhower (approved):
Silly GH CI bug, good catch.

PR #8172, opened 2021-06-25
Update debug level list link

MatthewSabre:
The link to the debug levels was broken in that the file changed and it no longer points to the correct lines. Update the line numbers and also include the commit hash in the URL to prevent it from breaking in the future.


d-a-v (approved):
Thanks !

PR #8175, opened 2021-06-25
platformio: use CCFLAGS for -Werror

mcspr:
Fixup https://github.com/esp8266/Arduino/pull/8165, as it only changed behaviour for .c files

Also adds some comments referencing SCons documentation (literally)


earlephilhower (approved):
Thanks for catching this and adding comments so it's unlikely to happen again!

PR #8182, opened 2021-06-26
Copy released JSON to board URL, not new one

### LINKED ISSUE #8180: Package script should take pre-gen'd JSON from draft, not make a new one
The GHCI publish script should not regenerate the ZIP and JSON  file before uploading it to the server because the `sha256sum` of the newly generated ZIP file will not match the one in the release (because of timestamp changes for the source files).

Instead, the package script should just pull the generated JSON file from the release and push it to the boards URL w/o adjustment.

earlephilhower:
Fixes #8180

The draft release generates a ZIP and JSON.  Instead of rebuilding the ZIP
on the publish step (which may result in a different SHA256 due to file time
differences in the new ZIP), just copy the one from the published release
directly.

Also clean up and remove unneeded environment variables and CI steps.

### DIFF
--- a/.github/workflows/release-to-publish.yml
+++ b/.github/workflows/release-to-publish.yml
@@ -36,26 +36,6 @@ jobs:
       run:
         shell: bash
     steps:
-    - uses: actions/checkout@v2
-      with:
-        submodules: true
-        fetch-depth: 0
-    - uses: actions/setup-python@v2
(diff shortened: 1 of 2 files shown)


d-a-v (approved):
LGTM but we never know until this is tested IRL.
Thanks for digging into this !

PR #8184, opened 2021-06-27
Fix NO_GLOBAL_INSTANCES

### LINKED ISSUE #8183: -DNO_GLOBAL_INSTANCES breaks ESP8266HttpClient
### Basic Infos

#### Platform

- Hardware:				ESP-12
- Core Version:			master
- Operating System:		Ubuntu

### Problem Description

Defining `NO_GLOBAL_INSTANCES` breaks `ESP8266HttpClient` because it depends on `devnull` being defined. But it seems that if it's needed, it shouldn't be disable by that option. And since it's one byte it seems acceptable.

```cpp
/src/ESP8266HTTPClient.cpp:850:32: error: 'devnull' was not declared in this scope
  850 |         _client->sendAvailable(devnull); // clear _client's output (all of it, no timeout)
      |                                ^~~~~~~
*** [.pio/build/release/libc18/ESP8266HTTPClient/ESP8266HTTPClient.cpp.o] Error 1
```

paulocsanz:
Fixes #8183 

It's not clear to me if this is the best solution. This is just a first draft so the community can debate on a better solution.

But it seems a fairly simple thing to fix. But something needed, if something is used internally it should be declared + defined.

It's not clear to me if we should create it when `NO_GLOBAL_STREAMDEV` is specifically defined.


paulocsanz:
Is it better as static or as a regular stack variable? Since it only occupies 1 byte.

devyte (approved):

paulocsanz:
@dok-net Can you review it? I fear it may cause breaking changes since it will force people to define Serial when they didn't have to before.

dok-net (commented):
ArduinoCore-avr does `void serialEventRun(void)` differently, I suggest adopting that implementation:
Implement `void serialEventRun(void)` in `HardwareSerial.cpp` unconditionally, that is, outside the `#if` block. In turn, enclose the code inside the function body in the same preprocessor conditional directive.
One could potentially add `serialEvent1`, too, in the same way.

paulocsanz:
I've undone the Serial changes, because it seems is a more complex issue that doesn't need to block this PR. We can talk about this more in the issue and if it makes sense, doesn't break compatibility and we find all the other declaration that are not defined we can make a new PR.

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -849,6 +849,10 @@ bool HTTPClient::connect(void)
 {
     if(_reuse && _canReuse && connected()) {
         DEBUG_HTTPCLIENT("[HTTP-Client] connect: already connected, reusing connection\n");
+
+#if defined(NO_GLOBAL_INSTANCES)
```
d-a-v, line 853: This is the negation of https://github.com/esp8266/Arduino/blob/29c63506f7146c9fe85dc9286f63b85b421ff573/cores/esp8266/StreamSend.cpp#L354


Should it be this instead ? 
```cpp
#if defined(NO_GLOBAL_INSTANCES) || defined(NO_GLOBAL_STREAMDEV)
```

paulocsanz: You are right, I thought that because the person explicitly disabled it it would be better for them to define. But moving it to the stack solves the issue.

dok-net:
> I've undone the Serial changes, because it seems is a more complex issue that doesn't need to block this PR. We can talk about this more in the issue and if it makes sense, doesn't break compatibility and we find all the other declaration that are not defined we can make a new PR.

Declaring symbols that are not defined is an error, don't you think so? Your reversal simply causes the compile phase to succeed, but then linking will fail. Not the proper way of doing this.

As to my remark involving the `extern void serialEventRun(void) __attribute__((weak));`, that can indeed be a separate matter, if you think it's outside the scope of this PR.

paulocsanz:
I added it again

earlephilhower (changes requested):
Thanks for trying to fix multiple things, but it’s really neat if you do one change per PR, please.    Can you remove the unrelated hardware serial stuff?

dok-net:
> Thanks for trying to fix multiple things, but it’s really neat if you do one change per PR, please. Can you remove the unrelated hardware serial stuff?

@earlephilhower If it were a mistake, it would be on me, I've asked the submitter to add it back in. You would have to try and see it from a different angle: The issue is not to create the devnull, but the fact that `NO_GLOBAL_INSTANCES` is not handled correctly. This applies to both HardwareSerial and  HttpClient. Plus, the comment that got removed is non-applicable to ESP8266's implementation and confusing, which turns up when doing this change.

There are NO further issues in this core regarding `NO_GLOBAL_INSTANCES`!

So, isn't it best to rename this PR, instead of splitting it up into extraordinarily small chunks? Besides, it's a simple fix for HardwareSerial, nothing else, or what could possible be the point in declaring something `extern` (no weak or anything) in the core headers that is not defined? It's just going to produce a linker error, which are harder on the average coder to understand.

paulocsanz:
Awaiting further instructions. I'm OK with both approaches, just ping me and I will edit the PR accordingly (or make a new PR to split the changes). But I don't have much to contribute to the discussion. I agree that this could become a PR to fix all the problems related to NO_GLOCAL_INSTANCES, but I haven't ensured those are the only ones, they are the ones I found while maintaining our system.

dok-net:
@paulocsanz I grepped all `cores` and `libraries` and found no other uses, that's all.

d-a-v (approved):

earlephilhower (approved):
OK, I'm convinced.  LGTM.

PR #8185, opened 2021-06-27
Get header from name stored in progmem

### LINKED ISSUE #8199: HttpClient::header doesn't support names stored in PROGMEM
### Basic Infos

### Problem Description

Most (if not all) names of headers to be collected  are stored in flash, but we have to allocate just to compare two char*. HttpClient::header should provide an API that supports flash stored names.

paulocsanz:
Currently you need a `const char*` stored in RAM to get a header from `ESP8266HTTPClient`, this adds an API to support `const __FlashStringHelper*` . And since most of the times headers needed are known at compile time, it seems better to store their names in flash.

Fixes #8199


paulocsanz:
Fixes #8199

earlephilhower (commented):
This seems okay, but do you think you can achieve the same thing, for more cases and with less code, by simply using the strcmp-p call way of string equality checking in the existing signature?  That’s exactly what is going on under the hood, anyway…

paulocsanz:
My bad I thought this was causing a crash in our code. There were a lot of changes to make our code stop crashing when updating from 2.7 to 3.0 that I ended up misdiagnosing a few.

PR #8187, opened 2021-06-28
Add a PostHttpsClient example in the ESP8266HTTPClient library.

Alba0404:
This PR add an example for the **ESP8266HTTPClient** library.
This example is a POST request using HTTPS protocol.
It was made using the PostHttpClient and GetHttpsClient examples.

### DIFF
--- a/libraries/ESP8266HTTPClient/examples/PostHttpsClient/PostHttpsClient.ino
+++ b/libraries/ESP8266HTTPClient/examples/PostHttpsClient/PostHttpsClient.ino
@@ -1,0 +1,87 @@
+/**
+   PostHttpsClient.ino
+    Created on: 28.06.2021
+*/
+
+#include <Arduino.h>
+#include <ESP8266WiFi.h>
+#include <ESP8266WiFiMulti.h>
+#include <WiFiClientSecureBearSSL.h>
+#include <ESP8266HTTPClient.h>
+
+// Fingerprint for demo URL, expires on September 5, 2021, needs to be updated well before this date
+const uint8_t fingerprint[20] = {0xbb, 0x88, 0x7f, 0x7c, 0x77, 0xc2, 0x59, 0x97, 0xb7, 0x00, 0x35, 0x74, 0x50, 0x47, 0x7e, 0x67, 0x42, 0x02, 0x2f, 0xf0};
+const char* URL = "https://www.reseau-astuce.fr/fr/horaires-a-larret/28/StopTimeTable/NextDeparture";
+const char* REQUEST = "destinations=%7B%221%22%3A%22Technop%C3%B4le+SAINT-ETIENNE-DU-ROUVRAY%22%7D&stopId=102154&lineId=175&sens=1";
+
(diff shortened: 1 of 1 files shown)


libraries/ESP8266HTTPClient/examples/PostHttpsClient/PostHttpsClient.ino
```diff
@@ -0,0 +1,86 @@
...
+
+void loop() {
+
+  if ((WiFiMulti.run() == WL_CONNECTED)) {
+    Serial.println("[HTTPS] begin...");
+
+    std::unique_ptr<BearSSL::WiFiClientSecure>client(new BearSSL::WiFiClientSecure);
+    client->setInsecure();                  // Ignore SSL certificate
+    //client->setFingerprint(fingerprint);  //Use SSL
+    HTTPClient https;
```
d-a-v, line 59: For coherency,
- `https` is already global
- why is `client` in heap and `https` in stack ?

Alba0404: - For the `https` var, it was a mistake, I remove the global var.
- For the difference heap and stack. I don't know, it was in other examples and it works

libraries/ESP8266HTTPClient/examples/PostHttpsClient/PostHttpsClient.ino
```diff
@@ -0,0 +1,86 @@
...
+
+
+void loop() {
+
+  if ((WiFiMulti.run() == WL_CONNECTED)) {
+    Serial.println("[HTTPS] begin...");
+
+    std::unique_ptr<BearSSL::WiFiClientSecure>client(new BearSSL::WiFiClientSecure);
+    client->setInsecure();                  // Ignore SSL certificate
+    //client->setFingerprint(fingerprint);  //Use SSL
```
d-a-v, line 57: It is prefered to use secure connections in examples. 
You may leave the `setInsecure()` commented though.

libraries/ESP8266HTTPClient/examples/PostHttpsClient/PostHttpsClient.ino
d-a-v: Stack (cons: too small or precious), Heap (cons: fragmentation), or global (cons: cannot be released), the debate is never closed.
What about using the same allocation scheme for both client and https ?

earlephilhower: You really want to allocate and leave allocated a WiFiClientSecure either globally or very near the beginning of setup().  This is because the client will allocate memory for itself (~17kb IIRC) but also allocate ~7kb for the SSL stack.  Later in the app the heap may have enough space, but be fragmented such that you might have 30kb free but no single block of 17kb or 7kb for the stack or the SSL buffers.

d-a-v:
pushing to v4 per still in discussion

PR #8190, opened 2021-07-01
Add wifi kit 8 to boards

gneiss15:
I use a "Wifi Kit 8" from Heltec.
This board is heavily offered at amazon right now.
This board is not included in the current tool chain for the ESP8266.
It is very similar to one that is offered by AZ-Delivery (not sure which is the original one).

I modified boards.txt.py and added the variant, so this kit is selectable using the normal "board selection menu".
There is a separate tool chain that supports this board (by installing a library from Aron Lee), but this uses a far outdated version of the ESP8266 tool chain. By adding this board to the official too chain it can be used without installing an extra (outdated) tool chain.


earlephilhower (approved):
Thanks

earlephilhower:
For the CI failure,  I think you just need to run `./tools/boards.py --allgen` to rebuild the docs with this new board.

gneiss15:
I had only done a --boardsgen ... 
So let's do a --allgen...
.. that produces two additional changes.
Hopefully the test did pass now..

gneiss15:
OK, Tests passed....
So I hope this request will be acepted.

Because I am new to this kind of versioning (I use subversion with full access to the repo), I am not sure which is the best way to make a 2. Pull request (with a total different context).
This request make the "mkdir.py" (under tools) compatible with Python Versions < 3_5.
The current mkdir.py will produces an error, because in versions < 3.5 there is no argument "exist_ok" for
pathlib.Path( p ).mkdir( parents = parents_, exist_ok = True )

As You see that should be a total seperate request, but how to generate this ?
Must I make a 2. fork of the master and do the changes there, or is it possible to add the changes here and generate a pull request that only contains the changes done for this "issue" ?

I hope I could explain what my problem is. I am not sure if this is clear in English, I am a German.

d-a-v:
Your PR seems to be good. I have two questions:
- Why didn't you select `'serial': '921',` like with the wemos/lolin boards
  The other boards don't have it and it's a shame (but leaving as is per history)
- [Per schematics](https://github.com/HelTecAutomation/Heltec_ESP8266#hardware-reference) there is no onboard led, so the LED_BUILTIN definition is wrong. Did Heltec try to not define it in the variant file ?
  (I was so sad when I tried the Blink test sketch with your board definition on my wifikit8 board (version from deek-robot "DK-W8266") hoping to discover an hidden microled :)

gneiss15:
@d-a-v:
> Why didn't you select 'serial': '921',
That's because I don't know how to do that. As the comment in "boards.txt" stated this file should be created by "boards.txt.py".
So I try to extract the relevant parts I found inside the repro from Aron Lee and insert them into this official repro.

Because I didn't know how to modify this part (adding a special baud rate) I haven't done that.
Now, that You mention the wemos/lolin boards, I found, that a simple additional line will do the trick.
I may add this tomorrow, if no one disagrees.
My experience is, that this baud rate does not work on every PC (it works on only one of 3 older PCs I had tested it).

> ... LED...
Yes, that's something I although had my problems with, because before having studied the schematic, I had assumed that the "power led" is the one that should blink ;-)
But Here again: If I know how to remove the led from the settings I will do. In Aron Lee's repro it present too (assuming he didn't know either).

gneiss15:
@d-a-v:
OK, I made some tests..

regarding LED:
By removing the line
#define LED_BUILTIN 0
inside variants/wifi_kit_8/pins_arduino.h
trying to build the blink sketch will result in:
error: 'LED_BUILTIN' was not declared in this scope
So that possible gave further users the idea that the board has no build in LED.
I will commit this change immediately

regarding 'serial': '921':
Even without the addition of 
 'serial': '921'
to the board definition, I am able to select an upload baud rate of 921600 Baud !?!
So, what is the affect of this ?
Can You please try if You are able to select this speed with the current state of the repro ?

d-a-v:
Yes, serial/921 can be selected from the menu.
Adding the option makes it the default choice.
But if it does not work for you then that's OK.

d-a-v (approved):

PR #8194, opened 2021-07-04
Make mkdir.py work under python3 versions below 3.5

gneiss15:
Because early versions of python3 did not have the optional arg "exist_ok" for "pathlib.Path(...).mkdir(...)" a build under this versions will abort with an error message. This PR will modify the python script so that it works even under python 3.4 (and below).


earlephilhower (changes requested):
Please don't change indents with patches.  This and the other PY scripts are 4-space indented.  Spacing changes like this make it hard to review/sync up.

Also, this seems like a spot where cascaded `try (3.5+ version)` `except all (<3.5 version)` would be better than examining python version numbers.  Or, if the <3.5 version of `Path.mkdir` works on >3.5, then just change it to use that API no matter what.

gneiss15:
1. Idents:
I corrected this. It's something that has changed by using my favorite editor.

2. Cascaded try
I don't thing that a nested try will be better here (because the function that fails is inside a loop, which will imply an exception on every iteration).
A try will have its advantage if it can't be precisely determined when the function will fail.
Here it is a well know language feature that was added with version 3.5, so the (original) call to:
pathlib.Path( ..).mkdir(.., exist_ok = True )
will fail on all versions below 3.5

3. Or, if the <3.5 version of...
OK, that was (IMHO) a better approach.
I had leave the original inside, because I would have changed a less as possible.

OK, I will upload a version that only use the <3.5 version.

earlephilhower (approved):
Thanks, this is much cleaner!

d-a-v (approved):

PR #8201, opened 2021-07-08
add comments and corrections

mhightower83:


d-a-v (approved):
LGTM

earlephilhower (approved):
Looking at the ROM disassembly, it does seem that there is no UART param in `ets_install_uart_printf`, so LGTM.  Thx!

PR #8204, opened 2021-07-09
feat: adding possibility to manually set MD5 checksum for HTTP update

vlastahajek:
I've written [a library](https://github.com/bonitoo-io/ESP-Github-Updater) that allows updating from Github releases. To verify firmware integrity HTTPUpdater already supports MD5 checksum, but is only read from the server response header. This is not possible if there is no control of the server.
If a checksum is known from another source, it is not possible to set it. This PR allows this.


earlephilhower (changes requested):
Looking at this change, it seems to only affect the debug output and nothing more.  The new ‘md5’ variable isn’t used outside the print, so I don’t see what the intent is here.  The description leads me to believe you wanted to override the check on some way.

vlastahajek:
@earlephilhower, thanks for the quick and detailed review. I've fixed the wrong merge of locally modified file.

d-a-v (approved):
LGTM

earlephilhower (approved):

PR #8206, opened 2021-07-09
Fixes failing https connections to HelloServerBearSSL

mhightower83:
Fixes failing https connections to HelloServerBearSSL when using the MMU option with 48K IRAM shared. This happened after build changes to the example that increased IRAM code size that caused free IRAM for Heap to fall below ~16K, then "new" would OOM out in WiFiClientSecureBearSSL. This was observed when used with https://github.com/esp8266/Arduino/pull/8196#issue-684769650

Added private function to try IRAM first then switch to DRAM on fail, to WiFiClientSecureBearSSL for iobuff allocations.


d-a-v (approved):
LGTM

earlephilhower (approved):
Thanks!

PR #8207, opened 2021-07-09
call umm_init just before starting SDK

mhightower83:
This fixes a problem I saw on a post to esp8266/arduino :point_right: : [June 30, 2021 11:40 AM](https://gitter.im/esp8266/Arduino?at=60dcba869cf317173002a00f), where the BearSSL stack trace is all zero's after an HWDT. I was too slow to recognize I had a problem around the time of his post.

The problem was `umm_init` was being done too early and the Heap was zeroed out before the stack dump was performed.

Added code to HWDT example to use StackThunk. This should make the problem more visible in the future should it return.


d-a-v (approved):
LGTM. 
I understand it is a bugfix that should be merged now, right ?

mhightower83:
@d-a-v - Yes this is a bug fix ready for merge. I also added a few missing comments, removed an unused build path, and updated the example code to make the zeros more evident should they return.

earlephilhower (approved):
LGTM

PR #8215, opened 2021-07-13
Netdump printf fix

hreintke:
Fixes https://github.com/esp8266/Arduino/discussions/8210

I also added a second commit which updates several printf to printf_P.
If you want that as a separate PR, just let me know.


d-a-v (approved):
Thanks !

earlephilhower (approved):

PR #8218, opened 2021-07-16
certificate and public keys automatic updater

d-a-v:
Release process is updated to use a new python script (`tools/cert.py` via `tools/certsUpdater`) which updates certificates, public keys and fingerprints used in examples.


d-a-v:
[It should pass](https://github.com/esp8266/Arduino/pull/8218/checks?check_run_id=3089335465#step:4:942)
```
The server certificate can be completely ignored and its public key
hardcoded in your application. This should be secure as the public key
needs to be paired with the private key of the site, which is obviously
private and not shared.  A MITM without the private key would not be
able to establish communications.
Trying: api.github.com:443...[hostByName] request IP for: api.github.com
[hostByName] Host: api.github.com IP: 140.82.114.5
BSSL:_connectSSL: start connection
BSSL:_wait_for_handshake: failed
BSSL:Couldn't connect. Error = 'Invalid signature in ServerKeyExchange or CertificateVerify message.'
*** Can't connect. ***
```

Should the host test fail when this happens ?

earlephilhower (approved):
LGTM and solves a constant niggle with the releases and stale certs.

PR #8222, opened 2021-07-17
Fix PRxxx printf format macros

### LINKED ISSUE #8220: Compile error when using fixed size integer macros from inttypes.h
### Basic Infos

#### Platform

- Hardware:					[other]
- Core Version:			[2021-07-16]
- Operating System:		[Ubuntu]

### Settings in IDE

- Flash Mode:			[dout]
- Flash Size: 		[1MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]

### Problem Description

Compile error when using the fixed size integer macros from inttypes.h with library versions 3.0.0+ but was OK in previous versions (e.g. v2.7.4). The code example also compiles OK on GNU GCC v7.1.1 

```cpp

#include <Arduino.h>
#include <stdint.h>
#include <inttypes.h>
#include <stdio.h>

void setup() {
  char buffer[100] = {0};
  uint32_t v = 12345;
  
  sprintf(buffer,"%" PRIu32, v);
}

void loop() {
}

```

### Debug Messages
Error message is:
```
MCVE.ino: In function 'void setup()':
MCVE:10:21: error: expected ')' before '__INT32'
   10 |   sprintf(buffer,"%" PRIu32, v);
      |          ~          ^
      |                     )
exit status 1
expected ')' before '__INT32'
```

earlephilhower:
Update toolchain (newlib) to supply proper definitions for PRxxx macros.
Fixes #8220

Added a PRxxx macro to an example to ensure CI will catch any problem like
this in the future.


earlephilhower:
Failure was due to a GCC-internal flag which was set to use `long int` for 32-bits when compiling GCC for `newlib`.  Yet another new toolchain build enroute.

d-a-v (approved):
Works for me, tested the linux taste of gcc, lwIP in debug mode (using PRxxx too)
Thanks

PR #8225, opened 2021-07-18
Allow control over authorization String allocation

### LINKED ISSUE #8224: HTTPClient::setAuthorization needs two allocations to support non-null terminated string views
### Basic Infos

### Problem Description

We store our fixed size data in std::array<char, SIZE>. This brings a few problems when dealing with APIs that only receive a starting char pointer, but not the length. So we endup having to allocate a string just so we have a null-terminated string to pass to those APIs.

With HttpClient::setAuthorization this means we have to allocate twice. I'm here asking for support for non-null terminated views as first class. Be it by passing the size, or a String to be moved, so we control the string allocation.

paulocsanz:
Fixes #8224

It seems one allocation is inevitable here, but allowing the user to control it solves the non-null terminated string-view problem. That currently needs two allocations to fix.

This replace(String('\n'), ...) doesn't seem ideal, should I implement String::replace(char, String&), or String::remove(char) to avoid allocating here needlessly? Of course this virtually has no impact since it's immediately freed, it won't cause heap fragmentation, so it seems overkill. But this kind of API bothers me a lot, where allocation is not needed but happens. It makes everything less deterministic.


paulocsanz:
Nevermind forgot about SSO for a second. Ignore the String::replace part.

earlephilhower (approved):
LGTM, thx!

earlephilhower (approved):

PR #8226, opened 2021-07-18
Use only installed Python on Windows

### LINKED ISSUE #8096: Fatal Python Error when using any ESP8266 board with Arduino IDE.
Hi,

my arduino ide installation is throwing up a strange problem when I change the board to any ESP8266... following is the error... unable to understand what the issue is.

```
Fatal Python error: initfsencoding: unable to load the file system codec
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00000e14 (most recent call first):
exit status 3
C:\Program Files (x86)\Arduino\arduino-builder returned 3
Error compiling for board NodeMCU 1.0 (ESP-12E Module).

```

earlephilhower:
When Python is installed on Windows separately (i.e. Windows Store,
Python.org MSI, etc.) it set the PYTHONHOME environment variable to
the installed path.

When we call our own portable Python, it tries to use the support
PYC files in PYTHONHOME.  If they're from a different version of
Python, however, they won't work.  Even though we're running our
own distributed Python.exe, we are crashing on the system-wide
Python installation.

Add -I to all python3 calls, ignore PYTHONHOME/etc. environment vars.

Tested under Ubuntu 18.04 Linux w/o any ill effects (we ship pyserial
ourselves and add it manually to the Python search path).

Fixes #8096, or should as I understand it.


mcspr:
btw there's also `-I` aka isolated mode
https://docs.python.org/3/using/cmdline.html#id2

platform.txt
```diff
@@ -158,9 +158,9 @@ tools.esptool.upload.params.quiet=
 # First, potentially perform an erase or nothing
 # Next, do the binary upload
 # Combined in one rule because Arduino doesn't support upload.1.pattern/upload.3.pattern
-tools.esptool.upload.pattern="{cmd}" "{runtime.platform.path}/tools/upload.py" --chip esp8266 --port "{serial.port}" --baud "{upload.speed}" "{upload.verbose}" {upload.erase_cmd} {upload.resetmethod} write_flash 0x0 "{build.path}/{build.project_name}.bin"
+tools.esptool.upload.pattern="{cmd}" "{runtime.platform.path}/tools/upload.py" -E --chip esp8266 --port "{serial.port}" --baud "{upload.speed}" "{upload.verbose}" {upload.erase_cmd} {upload.resetmethod} write_flash 0x0 "{build.path}/{build.project_name}.bin"
 
-tools.esptool.upload.network_pattern="{network_cmd}" "{runtime.platform.path}/tools/espota.py" -i "{serial.port}" -p "{network.port}" "--auth={network.password}" -f "{build.path}/{build.project_name}.bin"
+tools.esptool.upload.network_pattern="{network_cmd}" "{runtime.platform.path}/tools/espota.py" -E -i "{serial.port}" -p "{network.port}" "--auth={network.password}" -f "{build.path}/{build.project_name}.bin"
```
mcspr, line 163: `{cmd} espota.py -E` -> `{cmd} -E -- espota.py`?
assuming cmd expands to python.exe, arg is intended for python
(and probably also applies above, where python cmd is spelled out)

earlephilhower: Yes, I will update soon. That’s what I get for submitting PRs at 2am…

earlephilhower: Done with latest push.  I was worried about `-I` and the scripts directory, but we don't have any cross-script dependencies so it does work fine.

For PlatformIO, it seems the builder script doesn't use the packages Python and relies on `$PYTHONEXE` which I assume is part of/required for Platform.IO.

mcspr: True, lgtm. (although, I'd still add `--` for future's sake)

PIO itself needs system / bundled python to run, so it is slightly different there. e.g. with esptool, it just downloads a standalone script, since it already includes pyserial as it's core dependency

d-a-v (approved):
Build and flash is OK under ubuntu

d-a-v (approved):
Compilation and flashing works under windows and arduino IDE.
tested using the alpha release.

Issue #8228, opened 2021-07-21
[Question]: return enum WRONG_Passowrd (ESP8266WIFISTA.cpp)


hasenradball:
Hello is it here better to return the Status 6 == WRONG_PASSWORD?

https://github.com/esp8266/Arduino/blob/8a42163a50e97fddd2dd1d7522fbe6bfc5eeb929/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp#L136

d-a-v:
Yes. Would you be ready to create a pull-request for this ?

hasenradball:
I will do a PR for this 😉

hasenradball:
Have You seen my post in Gitter chat?

d-a-v:
> Have You seen my post in Gitter chat?

You have many :) There's always a reason when no one answer to a particular post. If the issue is real/persistent, posting a new entry in the issue tracker like you did is OK and better because is stays visible and opened until one has time to understand and eventually answer.
One could say that you deleted the issue template, and you did, but providing a solution to an issue is, I dare to think, a valid exception.

hasenradball:
Hi sorry added the the template and filled out :-)

hasenradball:
Hi I also recognized if using `waitForConnectResult` --> it returns also no `WL_WRONG_PASWORD`;

```
int8_t ESP8266WiFiSTAClass::waitForConnectResult(unsigned long timeoutLength) {
    //1 and 3 have STA enabled
    if((wifi_get_opmode() & 1) == 0) {
        return WL_DISCONNECTED;
    }
    // if probing doesn't trip, this yields
    using oneShotYieldMs = esp8266::polledTimeout::timeoutTemplate<false, esp8266::polledTimeout::YieldPolicy::YieldOrSkip>;
    oneShotYieldMs timeout(timeoutLength); // number of milliseconds to wait before returning timeout error
    while(!timeout) {
        if(status() != WL_DISCONNECTED) {
            return status();
        }
    }
    return -1; // -1 indicates timeout
}
```

hasenradball:
I tried to get the status `WL_WRONG_PASSWORD` by using the folllowing test code (see below), but giving a wrong password does not result in `Wifi.status() == WL_WRONG_PASSWORD` why???

Does the NOTE has an impact on this?

![grafik](https://user-images.githubusercontent.com/26808969/126552379-68278572-6165-4506-8574-b3cc8de12509.png)


```
#include <Arduino.h>
[...]

PR #8229, opened 2021-07-21
return `WL_WRONG_PASSWORD` - Update ESP8266WiFiSTA.cpp - issue #8228

hasenradball:
[change] return `WL_WRONG_PASSWORD` if there is an issue with the password!

See issue: #8228


d-a-v (approved):
LGTM !

hasenradball:
Hi,

thanks for review.
can I contact you with a special question on a private chat?

hasenradball:
Hi,
I am actually checking some WLAN states....
Lets assume you have an connected ESP which tells:
`WiFI.status =  CONNECTED`

Then you switch `WiFi.mode(WIFI_OFF);`

What status would you expect?

PR #8235, opened 2021-07-23
Make multiple FS begin calls noops() SDFS/LittleFS

earlephilhower:
When LittleFS.begin() or SDFS.begin() is called after the filesystem is
already mounted, don't unmount/remount.  When an unmount happens, all old
Files become invalid (but the core doesn't know this), so you would end
up with random crashes in FS code.

Now, check for _mounted, and if so just return immediately from begin().
This mimics the original SPIFFS code.

Fixes https://github.com/earlephilhower/ESP8266Audio/issues/407


d-a-v (approved):
Looks good and coherent even with spiffs

PR #8236, opened 2021-07-23
Implement move operations for HTTPClient

### LINKED ISSUE #8231: Unable to set HTTPClient to a std::optional
### Basic Infos

#### Platform

- Hardware:					ESP-12
- Core Version:			3.0.1
- Operating System:		Ubuntu

### Problem Description

We are unable to set a HTTPClient to a std::optional<HTTPClient> that already exists. I have no idea why, the compiler generates hundreds of lines of error all related to the fact that `std::optional<HTTPClient>(std::optional<HTTPClient>&&)` is implicitly deleted. So right now it's impossible to set a HTTPClient to a std::optional, and I have no idea how to make it work, or provide a PR for this (although I'm willing if pointed in the right direction). But this seems a bug. And is something we actually need (we memset the unused 4kb sys stack to zero and cast it to a struct that contains a bunch of classes, and the easier way to properly initialize them is to wrap them in an optional).

```cpp

#include <optional>

void setup() {
    std::optional<HTTPClient> http;
    http = std::make_optional(HTTPClient());
}

void loop() {

}

```

### Debug Messages

It's a huge compiler error: https://pastebin.com/2NZzvrgY

The core of it seems this:
```
include/loop.hpp:189:57: error: use of deleted function 'std::optional<HTTPClient>& std::optional<HTTPClient>::operator=(std::optional<HTTPClient>&&)'
  189 |       this->data->http = std::make_optional(HTTPClient());

.platformio/packages/toolchain-xtensa@2.100200.0/xtensa-lx106-elf/include/c++/10.2.0/optional:659:11: note: 'std::optional<HTTPClient>& std::optional<HTTPClient>::operator=(std::optional<HTTPClient>&&)' is implicitly deleted because the default definition would be ill-formed:
  659 |     class optional
      |           ^~~~~~~~

.platformio/packages/toolchain-xtensa@2.100200.0/xtensa-lx106-elf/include/c++/10.2.0/optional:659:11: error: use of deleted function 'std::_Enable_copy_move<false, false, false, false, _Tag>& std::_Enable_copy_move<false, false, false, false, _Tag>::operator=(std::_Enable_copy_move<false, false, false, false, _Tag>&&) [with _Tag = std::optional<HTTPClient>]'

.platformio/packages/toolchain-xtensa@2.100200.0/xtensa-lx106-elf/include/c++/10.2.0/bits/enable_special_members.h:306:5: note: declared here
  306 |     operator=(_Enable_copy_move&&) noexcept                         = delete;

.platformio/packages/toolchain-xtensa@2.100200.0/xtensa-lx106-elf/include/c++/10.2.0/optional: In instantiation of 'constexpr std::optional<typename std::decay<_Tp>::type> std::make_optional(_Tp&&) [with _Tp = HTTPClient; typename std::decay<_Tp>::type = std::decay<HTTPClient>::type]':
include/loop.hpp:189:57:   required from here .platformio/packages/toolchain-xtensa@2.100200.0/xtensa-lx106-elf/include/c++/10.2.0/optional:1207:62: error: no matching function for call to 'std::optional<HTTPClient>::optional(<brace-enclosed initializer list>)'
 1207 |     { return optional<decay_t<_Tp>> { std::forward<_Tp>(__t) }; }

```

And it goes on and on.

paulocsanz:
Fixes #8231

This implements move constructor and assignment for HTTPClient, intended to support std::optional<HTTPClient> proper usage. It's not clear to me if this is the right place to test for compilation errors.

### DIFF
--- a/tests/device/test_sw_http_client/test_sw_http_client.ino
+++ b/tests/device/test_sw_http_client/test_sw_http_client.ino
@@ -212,17 +212,6 @@ TEST_CASE("HTTPS GET request", "[HTTPClient]")
     }
 }
 
-TEST_CASE("Move constructible", "[HTTPClient]")
-{
-   {
-       std::optional<HTTPClient> maybe;
-       maybe = std::make_optional(HTTPClient());
-   }
-   {
-       HTTPClient http(std::move(HTTPClient()));
-   }
-}
-
 void loop()
 {
 }
--- a/tests/device/test_sw_http_client/test_sw_http_client.ino
+++ b/tests/device/test_sw_http_client/test_sw_http_client.ino
@@ -3,6 +3,7 @@
 #include <ESP8266HTTPClient.h>
 #include <BSTest.h>
 #include <pgmspace.h>
+#include <optional>
 
 BS_ENV_DECLARE();
 
@@ -209,6 +210,12 @@ TEST_CASE("HTTPS GET request", "[HTTPClient]")
             }
         }
     }
+}
+
+TEST_CASE("Move constructible", "[HTTPClient]")
+{
+   std::optional<HTTPClient> maybe;
+   maybe = std::make_optional(HTTPClient());
 }
 
 void loop()
--- a/tests/device/test_sw_http_client/test_sw_http_client.ino
+++ b/tests/device/test_sw_http_client/test_sw_http_client.ino
@@ -214,8 +214,13 @@ TEST_CASE("HTTPS GET request", "[HTTPClient]")
 
 TEST_CASE("Move constructible", "[HTTPClient]")
 {
-   std::optional<HTTPClient> maybe;
-   maybe = std::make_optional(HTTPClient());
+   {
+       std::optional<HTTPClient> maybe;
+       maybe = std::make_optional(HTTPClient());
+   }
+   {
+       HTTPClient http(std::move(HTTPClient()));
+   }
 }
 
 void loop()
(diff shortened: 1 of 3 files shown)


mcspr:
Late sorry for not being clear, since I did not literally mean to implement the move yourself with every member, and it should be apparent it is not really... nice looking and there is a better way to handle this.

Just for consideration: https://github.com/esp8266/Arduino/compare/master...mcspr:ptr-wrap-test
(and strictly solving the problem at hand, idk if I'd call it something nice either. real change I'd like would touch much more stuff)
With the idea to have a protocol for certain members handled through the member class itself, and not to manually micro-manage everything.

tests/device/test_sw_http_client/test_sw_http_client.ino
```diff
@@ -211,6 +212,17 @@ TEST_CASE("HTTPS GET request", "[HTTPClient]")
...
+{
+   {
+       std::optional<HTTPClient> maybe;
+       maybe = std::make_optional(HTTPClient());
+   }
+   {
+       HTTPClient http(std::move(HTTPClient()));
+   }
+}
+
```
mcspr, line 225: And another note for this, I think it would be sufficient to have:
```cpp
static_assert(std::is_default_constructible_v<HTTPClient>, "");
static_assert(!std::is_copy_constructible_v<HTTPClient>, "");
static_assert(std::is_move_constructible_v<HTTPClient>, "");
static_assert(std::is_move_assignable_v<HTTPClient>, "");
```
in the library's .cpp?

paulocsanz:
Your changes seem way better! Since you already implemented it, wouldn't you prefer to provide a PR and I close this one?

Just a suggestion, a name like ObserverPtr/NonOwningPtr seem better than ClientWrapper.

It seems a std::unique_ptr<Client> with a no-op deleter would work as well, but seem hackish.

My first instinct was using `std::optional<std::reference_wrapper<Client>>`, but it may be overkill to bring all those dependencies for this small feature.

PR #8251, opened 2021-07-30
Add functionality to copy generated binaries into sketch folder

gneiss15:
Because I need signed OTA updates, I found that the current IDE (1.8.15) does not export the generated signed binaries, nor does it show where these binaries are created (the current temporary folder).
IMHO it is a good idea to allow a user to save generated files.
My approach shown here is to copy these files into a subdir (binaries) of the sketch.
Inside this dir there will be another subdir with the current board name, here the files <sketchname>.ino.bin and <sketchname>.ino.bin.signed will be copied to.

My current "solution" works for me. For general usage, there should be a configuration method (inside the IDE), but that is something that I am not able to do.

So hopefully some that does a review can assist me here.


gneiss15:
As mentioned before I currently didn't know how to make this "configurable", but I think I had found the principal way...

There is a dir "tools" directly in the root folder of arduino.
There are already the tools, that supply the menu entries "ESP8266 ... Data Upload" and a description "howto.txt".
I think this is the way to add new entries to the tools-menu of arduino.
But this requires "java" what isn't my favorite language :-)

jandrassy:
https://arduino.github.io/arduino-cli/latest/platform-specification/#recipes-to-export-compiled-binary

gneiss15:
@jandrassy:
Thanks for pointing me to this direction. Unfortunately I couldn't figure out how to specify more than one file :-(

Inside the java classes it seams that ths entry is handled as a list, but whatever I tried (using two name seperated by):

- space
recipe.output.tmp_file={build.project_name}.bin {build.project_name}.bin.signed
recipe.output.save_file={build.project_name}.{build.variant}.bin {build.project_name}.{build.variant}.bin.signed

- ","
recipe.output.tmp_file={build.project_name}.bin, {build.project_name}.bin.signed
recipe.output.save_file={build.project_name}.{build.variant}.bin, {build.project_name}.{build.variant}.bin.signed

- as a list: "[<name1>, <name2>]
#recipe.output.tmp_file=[{build.project_name}.bin, {build.project_name}.bin.signed]
#recipe.output.save_file=[{build.project_name}.{build.variant}.bin, {build.project_name}.{build.variant}.bin.signed]

This will result in no file copied!

So the foreseen method for "export binary" doesn't seams to be prepared for copying more than one file :-(

gneiss15:
So another idea was to create an entry for the tools menu (by cloneing/modifying the tools for littleFs/SPIFFS) that performs that action. But even this seams to be a bit ugly, because the tools menu didn't alow the use of nested entries.
I had thing about something like:
ESP-Tools
+- littleFs
    +- create & upload
    +- create & export
+- SIFFFS
    +- create & upload
    +- create & export
+- Sketch
   +- export binary
   +- export signed binary

jandrassy:
> 
> 
> Thanks for pointing me to this direction. Unfortunately I couldn't figure out how to specify more than one file :-(
> So the foreseen method for "export binary" doesn't seams to be prepared for copying more than one file :-(

sorry, I thought it can do more. 
Maybe you could try `;` or ':'

gneiss15:
OK, now I found it!
I have to put this onto two lines:
```
recipe.output.tmp_file.1={build.project_name}.bin
recipe.output.save_file.1={build.project_name}.{build.variant}.bin
recipe.output.tmp_file.2={build.project_name}.bin.signed
recipe.output.save_file.2={build.project_name}.{build.variant}.bin.signed
```
This will do it!

So, this pull request is now void... How to cancel it ?

jandrassy:
perfect solution. 
maybe delete this PR and create a new one with a single simple commit

gneiss15:
You are right (delete this PR and....).
I will do so

PR #8256, opened 2021-08-05
Update LEAmDNS_Control.cpp

folkertvanheusden:
sendParameter.m_bUnicast is checked in _sendMDNSMessage() (src/LEAmDNS_Transfer.cpp, line 84) so should be initialized.


folkertvanheusden:
Please ignore:  MDNSResponder::stcMDNSSendParameter::clear() already does this.

PR #8258, opened 2021-08-06
Place deprecated attribute in front of function

oxan:
This is the standards-compliant way to mark functions as deprecated (see https://en.cppreference.com/w/cpp/language/attributes/deprecated).


earlephilhower (approved):
There are just so many ways to get from A to B in C++ it kind of makes one's head spin.  This change LGTM.

PR #8290, opened 2021-08-24
Update mmu_get... and mmu_set... 

mhightower83:
Updated performance inline functions `mmu_get_uint8()`, ... and `mmu_set_uint8()`, ...  to comply with strict-aliasing rules. 

Changes to avoid over-optimization of 32-bit wide transfers from IRAM, turning into 8-bit or 16-bit transfers by the new GCC 10.3 compiler. This has been a reoccurring/tricky problem for me with the new compiler. 

So far referencing the 32-bit value loaded by way of an Extended ASM R/W output register will stop the compiler from optimizing down to an 8-bit or 16-bit transfer.

Example:
```cpp
  uint32_t val;
  __builtin_memcpy(&val, v32, sizeof(uint32_t));
  asm volatile ("" :"+r"(val)); // inject 32-bit dependency
  ...
```

Updated example `irammem.ino`
* do a simple test of compliance to strict-aliasing rules
* For `mmu_get_uint8()`, added tests to evaluate if 32-bit wide transfers were converted to an 8-bit transfer.

### DIFF
--- a/cores/esp8266/mmu_iram.h
+++ b/cores/esp8266/mmu_iram.h
@@ -127,8 +127,26 @@ bool mmu_is_icache(const void *addr) {
 static inline __attribute__((always_inline))
 uint8_t mmu_get_uint8(const void *p8) {
   ASSERT_RANGE_TEST_READ(p8);
-  uint32_t val = (*(uint32_t *)((uintptr_t)p8 & ~0x3));
-  uint32_t pos = ((uintptr_t)p8 & 0x3) * 8;
+  // https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8#how-do-we-type-pun-correctly
+  // Comply with strict-aliasing rules. Using memcpy is a Standards suggested
+  // method for type punning. The compiler optimizer will replace the memcpy
+  // with an `l32i` instruction.  Using __builtin_memcpy to ensure we get the
+  // effects of the compiler optimization and not some #define version of
+  // memcpy.
+  void *v32 = (void *)((uintptr_t)p8 & ~(uintptr_t)3u);
+  uint32_t val;
+  __builtin_memcpy(&val, v32, sizeof(uint32_t));
+  // Use an empty ASM to reference the 32-bit value. This will block the
+  // compiler from immediately optimizing to an 8-bit or 16-bit load instruction
+  // against IRAM memory. (This approach was inspired by
+  // https://github.com/esp8266/Arduino/pull/7780#discussion_r548303374)
+  // This issue was seen when using a constant address with the GCC 10.3
+  // compiler.
+  // As a general practice, I think referencing by way of Extended ASM R/W
+  // output register will stop the the compiler from reloading the value later
+  // as 8-bit load from IRAM.
+  asm volatile ("" :"+r"(val)); // inject 32-bit dependency
+  uint32_t pos = ((uint32_t)p8 & 3u) * 8u;
   val >>= pos;
   return (uint8_t)val;
 }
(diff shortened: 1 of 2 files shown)


libraries/esp8266/examples/irammem/irammem.ino
```diff
@@ -14,6 +14,204 @@
 #define ETS_PRINTF ets_uart_printf
 #endif
 
+/*
+  Verify mmu_get_uint16()'s compliance with strict-aliasing rules under
+  different optimizations.
+*/
```
mcspr, line 20: It's already here, but would this and the rest of the file make sense as a device test file at `/tests/device/test_irammem/` like there's already a umm malloc one? And given the structure of the things below with success and failures

libraries/esp8266/examples/irammem/irammem.ino
```diff
@@ -14,6 +14,204 @@
...
+  for (size_t i = 0; i < sz; i++) {
+    Serial.printf(" %3u", x[i]);
+  }
+  Serial.printf("\r\n");
+}
+
+bool testPunning() {
+  bool result = true;
+  // Get reference result for verifing test
+  alignas(alignof(uint32_t)) uint16_t x_ref[] = {1, 2, 3, 0};
```
mcspr, line 171: nit: `alignas(uint32_t)` already handles this, just to be less noisy

https://en.cppreference.com/w/cpp/language/alignas
> aliagas(type-id) is equivalent to alignas(alignof(type-id))

cores/esp8266/mmu_iram.h
```diff
@@ -204,13 +245,13 @@ extern void _text_end(void);
...
-  uint32_t sec_heap = (uint32_t)_text_end + 32;
-  return (void *)(sec_heap &= ~7);
+  uintptr_t sec_heap = (uintptr_t)_text_end + (uintptr_t)32u;
+  return (void *)(sec_heap &= ~(uintptr_t)7u);
 }
 
 static inline __attribute__((always_inline))
 size_t mmu_sec_heap_size(void) {
-  return (size_t)0xC000UL - ((size_t)mmu_sec_heap() - 0x40100000UL);
+  return (size_t)0xC000ul - ((uintptr_t)mmu_sec_heap() - (uintptr_t)0x40100000ul);
```
mcspr, line 254: I wonder if this was mentioned before, but... isn't it IRAM_START? Could be easier to track named consts instead of just a number?
Also note that the rest of `#define ...` inside of the func blocks above is kind-of misleading, as it is global for anything including this header

mcspr (approved):
lgtm, besides the comments
(and it is broken as-is, so I'd rather push this sooner than later)

libraries/esp8266/examples/irammem/irammem.ino
mhightower83: I don't know much about how the optimizations might vary across processor vendors, my thinking was this way it is tested with a matching vendor/version of the ESP8266 GCC.

libraries/esp8266/examples/irammem/irammem.ino
mhightower83: done.

cores/esp8266/mmu_iram.h
mhightower83: I found a #define for IRAM start, I'll use it.

For #define inside of a function. I did that as an indication that it was the only place it was being used. I see that was not a clear thing to do. I wanted to use `constexpr`, however, that would not work for `.c` modules. I see that the optimizer to works very well in absorbing const into the code. I'll replace the #define(s) with const instead.

libraries/esp8266/examples/irammem/irammem.ino
mcspr: The basic idea is to have a separate set of 'tests' that have a clearly defined success and failures, which is useful to have grouped together instead of remembering to run the example code to sanity-check certain Core implementation details. 'Device' is just the board running the test program, and... we are fighting the c++ object lifetime assumptions done by the compiler, not the machine code specifics?

Just to note, idk if it actually plays correctly with something supposed to be hw-only and not mixed with MOCK tests

cores/esp8266/mmu_iram.h
```diff
@@ -26,13 +26,24 @@
...
+#define XCHAL_INSTRAM1_VADDR		0x40100000
+#define XCHAL_INSTROM0_VADDR		0x40200000
+#else
+#include <sys/config.h> // For config/core-isa.h
+/*
+  Cautiously use XCHAL_..._VADDR values where possible.
+  While XCHAL_..._VADDR values in core-isa.h may define the Xtensa processor
+  CONFIG options, they are not always an indication of DRAM, IRAM, or ROM
+  size or position in the address space.
+*/
```
mcspr, line 46: I guess make sense with the available definitions, naming would also simplify migration to anything else.

Looking at the RTOS, there is esp8266/eagle_soc.h declaring various IRAM_..., DRAM_..., RTC_... preprocessor tokens and funcs related to ranges. This is out-of-scope here, but I wonder if something like this is warranted here as well (and since even the 3.x NONOS does not have anything of sorts)

PR #8297, opened 2021-08-31
MOD-WIFI-ESP8266(-DEV) update

Stanimir-Petev:
Added Flash mode and reset method selection options for Olimex board MOD-WIFI-ESP8266(-DEV)


earlephilhower (changes requested):
This isn't going to pass CI.  You need to update tools/boards.txt.py and regenerate the boards.txt file from it,, not manually edit boards.txt.

Plus, it seems like you're doing both normal definitions and menu overridden ones which is not correct.  See line 5832-5833 and 5849-5850.

Stanimir-Petev:
Thank you for the advice! I wasn't aware of the python script inside the "/tools" folder.
After you mentioned about the double definitions of the FlashMode I noticed that I did the same mistake with the ResetMethod.
Now as it's done via the boards.txt.py script is there anything else that needs to be done in order to be valid?

The whole point of adding these options are that we have new flash memory chips on the board that works in those modes: ResetMethod: no dtr (aka ck); and FlashMode: QOUT.

d-a-v (approved):

earlephilhower (approved):
LGTM, thanks!

PR #8317, opened 2021-09-20
introduce `yieldUntil(untilFonction, timeoutMs)` 

d-a-v:
`yieldUntil(untilFonction, timeoutMs)` is a delay interruptible under a provided condition.

It is useful for ethernet drivers using recurrent scheduled functions.

replaces #6212

### DIFF
--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -249,3 +249,15 @@ void run_scheduled_recurrent_functions()
 
     fence = false;
 }
+
+
+bool yieldUntil (std::function<bool()> until, uint32_t maxMs) {
+    esp8266::polledTimeout::oneShotFastMs timeOut(maxMs);
+    while (true) {
+        if (until())
+            return true;
+        if (timeOut)
+            return false;
+        yield();
+    }
+}
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
@@ -26,6 +26,7 @@
 #include <string.h>
 #include <coredecls.h>
 #include <PolledTimeout.h>
+#include <Schedule.h>
 #include "ESP8266WiFi.h"
 #include "ESP8266WiFiGeneric.h"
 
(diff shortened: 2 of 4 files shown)


cores/esp8266/Schedule.cpp
```diff
@@ -249,3 +249,15 @@ void run_scheduled_recurrent_functions()
 
     fence = false;
 }
+
+
+bool yieldUntil (std::function<bool()> until, uint32_t maxMs) {
```
mcspr, line 254: std::function is not necessary for this use-case as the function obj is destroyed right after (it is relatively cheap though, but it's still a type pun hidden in the .cpp implementation)
plus this should probably be inside of esp8266 ns?

should this be a schedule.h template instead?
```cpp
namespace esp8266 {
namespace utils {

template <typename T>
bool yieldUntil(T&& untilCallback, unsigned long uint32_t) {
...
}

} // namespace utils
} // namespace esp8266
```

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -621,22 +622,19 @@ int ESP8266WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResul
         aResult = IPAddress(&addr);
     } else if(err == ERR_INPROGRESS) {
         _dns_lookup_pending = true;
-        delay(timeout_ms);
-        // will resume on timeout or when wifi_dns_found_callback fires
-        _dns_lookup_pending = false;
-        // will return here when dns_found_callback fires
-        if(aResult.isSet()) {
+        if (yieldUntil([pIp = &aResult](){ return pIp->isSet(); }, timeout_ms))
```
mcspr, line 625: why not just `[&]() { return aResult.isSet(); }`?

cores/esp8266/LwipIntfDev.h
```diff
@@ -36,7 +36,7 @@ class LwipIntfDev: public LwipIntf, public RawDev
         _mtu(DEFAULT_MTU),
         _intrPin(intr),
         _started(false),
-        _default(false)
+        _default(true)
```
mcspr, line 39: was this intended?

cores/esp8266/Schedule.cpp
d-a-v: `esp8266::utils::yieldUntil()` is moved in `cores/esp8266/yieldUntil.h`

cores/esp8266/LwipIntfDev.h
mcspr: makes sense for one instance, true. ok if not a problem
(but, I wonder if default=false would be safer when there are more)

d-a-v: If there are more than one interface, only the last "defaulted" one becomes the default route.
I will investigate further on why this is necessary, but last time I checked packets are not routed without it.
I could have delegated the directive to "userland" but I think it is better to let the library deal with defaults so we don't need to change the API or the guidelines / examples when something changes or is fixed.

d-a-v: see also [WiFi](https://github.com/d-a-v/esp82xx-nonos-linklayer/blob/5919b910580265252ac1980b9fe936ece5839e41/glue-lwip/lwip-git.c#L326)

mcspr: so the instance variable does not make sense and it should be a static (and shared between the wifi and the lwipintf class), if every interface wants to become the default on callback?

d-a-v: Fair enough.
I have moved the default value back to `false` and updated examples to explicitly use the new interface as default route.

mcspr: Should the lwip part also be made aware of default modifications? (with another issue / pr, I gather, not to pollute the changes too too much)

d-a-v: Modification for "default" is reverted to what it was before the PR no ?
If the question is about "lwip2 - the driver adapter",  it is very delicate to update the WiFi interface management code per possible hidden side effect. Maybe a new gh-issue describing the API question might be indeed opened for that purpose.

Issue #8327, opened 2021-10-01
client.connected() return true when internet lost connection


ruem2802:
----------------------------- Delete below -----------------------------

If your issue is a general question, starts similar to "How do I..", is related to 3rd party libs, or is related to hardware, please discuss at a community forum like esp8266.com.

INSTRUCTIONS
============
If you do not follow these instructions, your issue may be dismissed.

1. Follow the checklist under Basic Infos and fill in the [ ] spaces with an X.
2. Fill in all the fields under Platform and Settings in IDE marked with [ ] (pick the correct option for you in each case, delete the others).
3. If you haven't already done so, test your issue against current master branch (aka latest git), because it may have been already fixed.
4. Describe your problem.
5. If you have a STACK DUMP decode it:

https://arduino-esp8266.readthedocs.io/en/latest/Troubleshooting/stack_dump.html

6. Include a Minimal Complete Reproducible Example sketch that shows your issue. Do not include your entire project, or a huge piece of code.
7. Include debug messages:

https://arduino-esp8266.readthedocs.io/en/latest/Troubleshooting/debugging.html 

8. Use markup (buttons above) and the Preview tab to check what the issue will look like.
9. Delete these instructions from the above to the below marker lines before submitting this issue.

----------------------------- Delete above -----------------------------

### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12|ESP-01|ESP-07|ESP8285 device|other]
- Core Version:			[latest git hash or date]
- Development Env:		[Arduino IDE|Platformio|Make|other]
- Operating System:		[Windows|Ubuntu|MacOS]

### Settings in IDE

- Module:  			[Generic ESP8266 Module|Wemos D1 mini r2|Nodemcu|other]
- Flash Mode:			[qio|dio|other]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v1.4|v2 Lower Memory|Higher Bandwidth]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
[...]

PR #8329, opened 2021-10-04
documentation: add description for connected() and status()

mcspr:
ref. the original PR and the discussion
https://github.com/esp8266/Arduino/pull/4626
https://github.com/esp8266/Arduino/issues/6701

as mentioned in the https://github.com/esp8266/Arduino/issues/8327#issuecomment-933369314

(I have yet to check the readthedocs rendering)

### DIFF
--- a/doc/esp8266wifi/client-class.rst
+++ b/doc/esp8266wifi/client-class.rst
@@ -16,7 +16,17 @@ Methods documented for `Client <https://www.arduino.cc/en/Reference/WiFiClientCo
 9.  `flush() <https://www.arduino.cc/en/Reference/WiFiClientFlush>`__
 10. `stop() <https://www.arduino.cc/en/Reference/WiFIClientStop>`__
 
-Methods and properties described further down are specific to ESP8266. They are not covered in `Arduino WiFi library <https://www.arduino.cc/en/Reference/WiFi>`__ documentation. Before they are fully documented please refer to information below.
+Methods and properties described further down below are specific to ESP8266. Some of them behave differently from the reference `Arduino WiFi library <https://www.arduino.cc/en/Reference/WiFi>`__ , or are only implemented for this Core.
+
+connected
+~~~~~~~~~
+
+Unlike the reference implementation, ``connected()`` means that the client is available for both reads and writes. Please use ``status()`` for only the connection information, and ``available()`` if you mean to check whether there's unread data.
+
+status
+~~~~~~
+
+Current implementation returns ``0`` (``CLOSED``) when the client is disconnected and ``4`` (``ESTABLISHED``) when connected. At the time of writing these refer to the ``enum tcp_state`` values that can be found at the `lwip/tcpbase.h <https://github.com/esp8266/Arduino/blob/master/tools/sdk/lwip2/include/lwip/tcpbase.h>`
 
 flush and stop
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


mcspr:
And as #6701 already mentioned, this is somewhat different for the secure variant:
https://github.com/esp8266/Arduino/blob/9d024d17fd18dffa569fa646b5326b5e8f58b0d5/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L259-L264
But, it's unclear whether this refers to the WiFiClient interface or the basic-socket WiFiClient class

d-a-v:
To summarize the need for this change:
In the case where `connected()` *was* returning true when the connection was closed by peer but data were still available for read in receive buffers, sketches found themselves allowed to write something which obviously leaded to an error when they did.
Removing the `available()` test from `connected()` could stop this behavior.
The downside of this change is that `available()` has to be checked even if `connected()` is false.
... and if this (half-closed by peer) connection is fully closed by sketch before reading remaining data, peer complains because its data will never be read. The server example in doc allows to reproduce this, the matching posix error on the peer side shows a 'Broken pipe' if I remember well.

So indeed I think this PR should also remove the `available()` part of TLS's `connected()` function for coherency with `WiFiClient` implementation. Good catch.

>  But, it's unclear whether this refers to the WiFiClient interface or the basic-socket WiFiClient class

The TLS context is supposed to match the WiFiClient class.

mcspr:
> The TLS context is supposed to match the WiFiClient class.

...true, but also note that it is used internally. can update, but it may be better suited as a separate patch

Will update with the 'connected' part with the reasoning, seems like a good place to also explain things

d-a-v:
> but also note that it is used internally.

Right!
In those places, one must check whether testing `available()` must be added next to `connected()`, or replace it.

doc/esp8266wifi/client-class.rst
```diff
@@ -16,7 +16,21 @@ Methods documented for `Client <https://www.arduino.cc/en/Reference/WiFiClientCo
 9.  `flush() <https://www.arduino.cc/en/Reference/WiFiClientFlush>`__
 10. `stop() <https://www.arduino.cc/en/Reference/WiFIClientStop>`__
 
-Methods and properties described further down are specific to ESP8266. They are not covered in `Arduino WiFi library <https://www.arduino.cc/en/Reference/WiFi>`__ documentation. Before they are fully documented please refer to information below.
+Methods and properties described further down below are specific to ESP8266. Some of them behave differently from the reference `Arduino WiFi library <https://www.arduino.cc/en/Reference/WiFi>`__ , or are only implemented for this Core.
+
+connected
+~~~~~~~~~
+
+Unlike the reference implementation, ``connected()`` means that the client is available for both reads and writes. It will return ``true`` when the client is ``status() == ESTABLISHED`` or ``available() == true``, but will return ``false`` when the ``status() == CLOSED``. Please use ``status()`` for the connection information, ``availableForWrite()`` to check whether it is possible to write, and ``available()`` if you mean to check whether there's unread data.
```
d-a-v, line 24: > It will return ``true`` when the client is ``status() == ESTABLISHED`` or ``available() == true``, but will return ``false`` when the ``status() == CLOSED``.

According to current code, it will return false if the connection is closed, regardless `available()` result,
meaning "writing is not allowed".

`available()` can still be checked to tell if something has been received and has been kept unread.

https://github.com/esp8266/Arduino/blob/f401f08aba74f6dd0f6dbeb9b1b6b6d94b174b31/libraries/ESP8266WiFi/src/WiFiClient.cpp#L329-L335

PR #8330, opened 2021-10-07
Don't return `true` with `WiFiClientSecureBearSSL::connected()` when really disconnected

mcspr:
ref. https://github.com/esp8266/Arduino/pull/8329#issuecomment-933913998
not tested (yet)

Apply the same condition as with normal WiFiClient - we are not connected
when it's not possible to both write and read.

Implement separate methods for actual connection status and the internal
ssl engine status and update methods that were previously using available()
for this purpose

Update examples to check available() when the intent is to only read the
data and not interact with the client in any other way. Also, use connect()
as a way to notify errors, no need to check things twice


libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp
```diff
@@ -253,19 +253,24 @@ void WiFiClientSecureCtx::_freeSSL() {
...
 
 uint8_t WiFiClientSecureCtx::connected() {
-  if (available() || (_clientConnected() && _handshake_done && (br_ssl_engine_current_state(_eng) != BR_SSL_CLOSED))) {
-    return true;
+  if (!_clientConnected()) {
+    return false;
   }
-  return false;
+
+  return (_updateRecvBuffer() || _engineConnected());
```
d-a-v, line 268: `_updateRecvBuffer()` is the alternate name of both `client` and `clientCtx`'s `available()`.
The logic is now
* can't write => not connected
* otherwise: same code as before: `available()` || "SSLworking"

I was thinking that the purpose of this PR was to remove `available()` from `connected()`,
It looks like now, and maybe on purpose:
- `connected()` => false when can't write on TCP socket (even if data is available for read)
- otherwise: `connected()` => `available()` or SSL engine is enabled

If the SSL engine is not working (`!_engineConnected()`), then `connected()` still returns true if data are available for read, allowing user to call `write()` which is not possible at that time (I'm still wondering if I miss something, like `_engineConnected()` can later switch to true)

mcspr: This still goes around the issue with the socket - we definitely know the connection state from the `status()`. Internals *only* check for WiFiClient::available() when reading, and state() when writing anyway, only externals are affected.

The original idea with `available()` / `_updateRecvBuf()` is to advance the state via `_run_until()`, and maybe process some other stuff. If the engine closes while inside available(), it is returning 0. In case the underlying socket closes and there's no pending data or it reaches timeout, it is also 0.

So maybe the better idea is to:
- only use internal methods with definitive results instead of ref Arduino API
- fix Arduino API back to the reference state
- add smth like lwipConnected()

?

d-a-v: Yes I do mean that.

But your ideas deserve better consideration:
> - only use internal methods with definitive results instead of ref Arduino API
> - add smth like lwipConnected()

This point is delicate
> - fix Arduino API back to the reference state

because the change (`connected()` not including `available()`) fixes network errors (#4626) when a user sketch closes without reading all received data. The notions of having the right to read and the right to write must be distinct and available to user. If you consider this API change (4y ago) is still harmful regarding official Arduino API, I suggest to open a new discussion and tag it for 3.1 so it is not forgotten, as I know other users still may have comments about that (@jandrassy at least).

libraries/ESP8266WiFi/examples/BearSSL_MaxFragmentLength/BearSSL_MaxFragmentLength.ino
```diff
@@ -80,9 +79,7 @@ int fetchMaxFragmentLength() {
   bool mfln = client.probeMaxFragmentLength("tls.mbed.org", 443, 512);
   Serial.printf("\nConnecting to https://tls.mbed.org\n");
   Serial.printf("MFLN supported: %s\n", mfln ? "yes" : "no");
-  if (mfln) { client.setBufferSizes(512, 512); }
```
d-a-v, line 83: Has it been falsely dropped during this fixing process ?

d-a-v (approved):

Issue #7969, opened 2021-04-09
WiFiScan.ino ported to CoopTask libraries doesn't find any networks


dok-net:
### Preparation (MCVE)
Let's trivially port the `libraries\ESP8266WiFi\examples\WiFiScan\WiFiScan.ino` to the Arduino [CoopTask](https://www.arduino.cc/reference/en/libraries/cooptask/) library, like so:
```
/*
    This sketch demonstrates how to scan WiFi networks.
    The API is almost the same as with the WiFi Shield library,
    the most obvious difference being the different file you need to include:
*/

#include <ESP8266WiFi.h>
#include <CoopTask.h>

void loop1() {
    String ssid;
    int32_t rssi;
    uint8_t encryptionType;
    uint8_t* bssid;
    int32_t channel;
    bool hidden;
    int scanResult;

    for (;;) {
        Serial.println(F("Starting WiFi scan..."));

        scanResult = WiFi.scanNetworks(/*async=*/false, /*hidden=*/true);

        if (scanResult == 0) {
            Serial.println(F("No networks found"));
        }
        else if (scanResult > 0) {
            Serial.printf(PSTR("%d networks found:\n"), scanResult);

            // Print unsorted scan results
            for (int8_t i = 0; i < scanResult; i++) {
                WiFi.getNetworkInfo(i, ssid, encryptionType, rssi, bssid, channel, hidden);

                Serial.printf(PSTR("  %02d: [CH %02d] [%02X:%02X:%02X:%02X:%02X:%02X] %ddBm %c %c %s\n"),
                    i,
                    channel,
                    bssid[0], bssid[1], bssid[2],
                    bssid[3], bssid[4], bssid[5],
                    rssi,
                    (encryptionType == ENC_TYPE_NONE) ? ' ' : '*',
                    hidden ? 'H' : 'V',
                    ssid.c_str());
                yield();
            }
        }
        else {
            Serial.printf(PSTR("WiFi scan error %d"), scanResult);
        }

        // Wait a bit before scanning again
        delay(5000);
    }
}

BasicCoopTask<CoopTaskStackAllocatorAsMember<1600>> task1("l1", loop1);

void setup() {
  Serial.begin(115200);
  Serial.println(F("\nESP8266 WiFi scan example"));

  // Set WiFi to station mode
  WiFi.mode(WIFI_STA);

  // Disconnect from an AP if it was previously connected
  WiFi.disconnect();
  delay(100);

  task1.scheduleTask();
}

[...]

PR #8332, opened 2021-10-09
PR #7148 w/ lwIP Ethernet merged

dok-net:
This is for testing whether the delay_* metaphors proposed in PR #7148 also work as required for he LwIP Ethernet library from PR #8317.
@mcspr I'm posting this for you, if you should like to do some testing. I have done the same as to whether `yieldUntil` works for issue #7969, and I must report that sorry, no, it does not at all.


mcspr:
Thanks, but I already have 'git checkout'ed the required lib files straight from the #8317 branch

The only issue is - I can't seem to get the board to work with d1mini. Running the 'basic' esp-idf esp32(orig/s2/c3) example - it works, without a hitch. However, the lwip_ethernet lib just dumps that there's 'no hardware found' :/
Set up like it is suggested - SCLK <-> GPIO14, MOSI <-> GPIO13, MISO <-> GPIO12 and CS <-> GPIO5 (per the GPIO15 comment and b/c it is causing a boot mode change, but it also does not find the hardware with that pin anyway)

PR #8255, opened 2021-08-02
Add functionality to copy generated SIGNED binaries into sketch folder

gneiss15:
The current platform.txt doesn't define the entries to copy signed binaries when performing a "Export compiled Binary".
This pull request adds the nessesary enties.

Whit this (extra) entries) both binaries are copied to the sketch folder (.bin & .bin.signed)

PR #8338, opened 2021-10-14
"Integrated handling for filesystem and gzipped Binaries" (Restart with fresh Fork)

gneiss15:
ATENTION: This is tested on Linux only! See below for open questions/prerequisites.

Currently there are some manual steps necessary to prepare the files necessary for OTA updates (especially for ESP8266HTTPUpdateServer)

The steps are:

Compile the sketch
Export the generated sketch-binary
Generate the filesystem
Export the fs-binary
Generate gzipped versions of binaries
Generate signed versions of gzipped binaries
Which of the above steps are necessary depends on the sketch (does it use a fs?) and the kind of OTA chosen (signed binaries necessary?) or preferred usage (use of gzipped binaries for faster upload?).

There are some problems/inconveniences:
2. The signed sketch-binary is not exported when using "Export compiled Binary".
(This point will be solved with my Pull-Request #8255)
3. & 4. AFAIK, the only “integrated” way to generate the fs is using one of the tools:
“arduino-esp8266littlefs-plugin” or “arduino-esp8266fs-plugin”.
Both tools are designed to create & upload the fs, but not export them.
So if preparing the files for an OTA when no board is connected, you have to:

ignore errors shown because upload is not possible
find & navigate to the “build.path” and copy the fs-binary manually
& 6. AFAIK, there is no “integrated” way to perform these steps.
This Pull-Request will integrate all of the above steps into the “compile and/or upload” action of Arduino.

This is done by adding 3 new entries (“Upload”, "Filesystem" and "Export") to the tools-menu.
These entries are presented just like any other options for esp8266 (“Upload speed” ... “Non-32-Bit Access”) and placed right after the last.

Below is the text that is already integrated into "doc/filesystem.rst"
--- schnipp ---

Menu-Entry “Upload” lets You choose between:
- Sketch
- Filesystem
- Both
This Menu controls which parts are uploaded when "Sketch->Upload" (Ctrl-U)
is selected from the Arduino Menu.

Menu-Entry “Filesystem” lets You choose between:
- Off
- LitteFs
- SPIFFS
This Menu controls which Fs will be created when "Sketch->Compile" (Ctrl-R)
is selected from the Arduino Menu. The Fs will always be created inside the
"export" dir (see below). 

Menu-Entry “Export” lets You choose between:
- Off
- .bin & .bin.signed)
- Create & Export gzipped Binaries too
This Menu controls export & generation of (extra) Sketch-Binaries.

The "export" dir
................
After a (successful) compile, the files will be exported to a subdir
“bin/{variant}” of Your sketch directory
(“{variant}” replaced with the name of the board as shown behind
“Tools -> Board:”).
Signed variants of all Binaries are generated similar to the
“automatic signing” done for normal sketch binary.

As valid for the "old" tools mentioned below, it is nessesary to place
files you want to be inside the generated file system into a directory
named ``data`` inside Your sketch directory.
--- schnapp ---

Open questions/prerequisites:
1. As stated above I had tested this on Linux only.
Hopefully some cant test it under windows..

2. tkinter should be available.
if it is not, and creating of fs is requested AND the data dir is empty, I didn't know how to ask the user: “Are you sure you want to create an empty ... image?”.
So in this case a message is given and the fs will not be created.
(So its not a real problem if tkinter isn't installed)

An additional question for developers:
As mentioned inside function main of “postbuild.py”, there is a strange behavior of the arg "{build.path}" supplied via “platform.txt”. If someone can explain me why this “string” behave as described in the comments, I would be happy ;-)


d-a-v:
Sorry for taking so long to check on your work, I'm just not used to use filesystems flashing (I prefer to use network protocols and download files on first run). I have been trying the ["alpha" release](https://d-a-v.github.io/esp8266/Arduino/index.html) on a dummy linux account that I use for testing.
It appears that `mklittlefs` is not found after installing. 
mklittlefs
In `platform.txt` python3 for example is accessed through `{runtime.tools.python3.path}/python3` so `mklittlefs` should be accessible with `{runtime.tools.mklittlefs.path}/mklittlefs`. Maybe this variable could be passed to `upload.py`?

gneiss15:
@d-a-v: I am not sure what You are telling me with: "It appears that mklittlefs is not found after installing"...
1. mklittlefs will be "pulled in" on a install "Using git version" (see https://github.com/esp8266/Arduino) of the esp8266 toolchain only after You "Initialize the submodules" as described here:
2. https://arduino-esp8266.readthedocs.io/en/latest/installing.html#using-git-version
3. The screenshot You present above does not indicate, that mklittlefs wasn't installed, it indicates that You try to compile a sketch without having supplied the necessary "data" dir!
4. As written inside (my extendend) documentation, you have to provide the data that should be placed into the FS inside a subdir of your sketch. This dir has to be named "data"!
5. Here is my error message window, when I rename the dir data to data2:
6. 
image

What looks a bit seldom is, that in Your window there is no "path" to the (missing) data dir is shown ??
Currently I have no idea what may be the reason.

gneiss15:
Sorry, Commit text is wrong it should have been "Improved..." not "Aproved..."

gneiss15:
@d-a-v
@earlephilhower 
Almost 30 Days without any reaction ?

d-a-v:
#6690 is mine, is more than 2 years old, and hasn't received any public comment since 9 months.

Please be patient, we don't have more free time than you do.

gneiss15:
@d-a-v That sound like You haven't enough manpower. May I help anywhere ?

Issue #7253, opened 2020-04-29
Unable to upload sketch for ESP8266 due to some elf2bin.py error


BenniKain:
> Hello,
i am trying for hours to get a sensor sending data over wifi to my flask server.
however due to some reason of esp82 error link below i cannot upload the sketch.

it worked 2 times in 4 hours for me seemingly random.

if anyone could help me i d be very thankful

```

Arduino: 1.8.12 (Windows Store 1.8.33.0) (Windows 10), Board: "LOLIN(WEMOS) D1 mini Lite, 80 MHz, Flash, Legacy (new can return nullptr), All SSL ciphers (most compatible), 1MB (FS:64KB OTA:~470KB), v2 Lower Memory, Disabled, None, Only Sketch, 921600"

C:\Users\User\Documents\Arduino\testung_httpclient\testung_httpclient.ino: In function 'void loop()':

C:\Users\User\Documents\Arduino\testung_httpclient\testung_httpclient.ino:56:47: warning: 'bool HTTPClient::begin(String)' is deprecated (declared at C:\Users\User\Documents\ArduinoData\packages\esp8266\hardware\esp8266\2.6.3\libraries\ESP8266HTTPClient\src/ESP8266HTTPClient.h:155) [-Wdeprecated-declarations]

    http.begin("http://192.168.8.105:8090/post");      //Specify request destination

                                               ^

Traceback (most recent call last):

  File "C:\Users\User\Documents\ArduinoData\packages\esp8266\hardware\esp8266\2.6.3/tools/elf2bin.py", line 169, in <module>

    sys.exit(main())

  File "C:\Users\User\Documents\ArduinoData\packages\esp8266\hardware\esp8266\2.6.3/tools/elf2bin.py", line 159, in main

    write_bin(out, args.app, ['.irom0.text', '.text', '.text1', '.data', '.rodata'], 0, args.flash_mode, args.flash_size, args.flash_freq, args.path)

  File "C:\Users\User\Documents\ArduinoData\packages\esp8266\hardware\esp8266\2.6.3/tools/elf2bin.py", line 74, in write_bin

    [size, addr] = get_segment_size_addr(elf, segment, path)

  File "C:\Users\User\Documents\ArduinoData\packages\esp8266\hardware\esp8266\2.6.3/tools/elf2bin.py", line 55, in get_segment_size_addr

    raise Exception('Unable to find size and start point in file "' + elf + '" for "' + segment + '"')

[...]

PR #8356, opened 2021-10-31
Fix for invalid elf file when converting to bin

anilsoni85:
On windows 10 xltensa-lx106-elf-objdump sporadically returns no output and results into build errors. Adding logic to retry to read segment from elf.

Ref: #7253


tools/elf2bin.py
```diff
@@ -44,15 +44,20 @@ def get_elf_entry(elf, path):
...
+        with subprocess.Popen([path + '/xtensa-lx106-elf-objdump', '-h', '-j', segment,  elf], stdout=subprocess.PIPE, universal_newlines=True ) as p:
+            lines = p.stdout.readlines()
+            if (len(lines) > 0): 
+                for line in lines:
+                    if segment in line:
+                        words = re.split('\s+', line)
+                        size = int(words[3], 16)
+                        addr = int(words[4], 16)
+                        return [ size, addr ]
+                raise Exception('Unable to find size and start point in file "' + elf + '" for "' + segment + '" path "' + path + '"')
```
earlephilhower, line 60: Is this the right level for the `raise`?  Shouldn't it be back at the main `def` level (same level as the `while` block) and not inside the `if (len)...` block?

earlephilhower (changes requested):

tools/elf2bin.py
anilsoni85: Yes, moved raised to def level

d-a-v:
Would it be possible to print the exception details so this weird behavior is understood ?

earlephilhower:
> Would it be possible to print the exception details so this weird behavior is understood ?

I'm not sure that would be helpful.  The exception now manually raised when the output is empty.

Python's `subprocess.popen()` will throw an `OSError` exception (which is not caught and will dump to console/IDE) if the app can't start for some reason.  But, it's not in the case they're reporting because they're just showing the manually created "no start found" exception we make.

So we really don't have anything to go on here.  I would maybe add a note to the top saying `# This is a hack to fix intermittent Windows silent failures, see #7253` or something because it's the only spot in a whole bunch of calls we do this.

I'd also like to get a report or two that this fixes things from someone who has been having failures before merging.

My gut says there's something in these users' Windows installs that's blocking the app from running (malware scanner? antivirus? race condition w/OneDrive online?  no idea) every now and then.  I, too, have never had it fail in my (admittedly limited) Windows testing, nor has it crapped out *ever* in Windows CI.

d-a-v:
I was indeed thinking of a race condition without further clue but 

> malware scanner? antivirus? race condition w/OneDrive online?

Considering that we are dealing with executable files, I think you put your finger on it @earlephilhower .

anilsoni85:
> Would it be possible to print the exception details so this weird behavior is understood ?
> Python's `subprocess.popen()` will throw an `OSError` exception

The process doesn't print any output to STDOUT, while debugging the issue I dumped each line inside the for-loop and noticed that when it fails there is no output. Perhaps the process writes on STDERR but returns with exist code 0 because there was no OSError exception on console. The process silently goes away sporadically without printing a single line. I will revert my local script change and put screenshot later when I will get chance. The error was pretty consistent to reproduce but it was failing randomly for one of the segment of elf.

> My gut says there's something in these users' Windows installs that's blocking the app from running (malware scanner? antivirus? race condition w/OneDrive online? no idea)

I was running it on local isolated development environment which is excluded from OneDrive sync therefore I don't think its related to one drive/Antivirus/Malware.

anilsoni85:
Look at the attached log lines from build output I generated this log file locally after adding print in elf2bin.py. Notice how there were 2 attempts to load .text segment at line#6 and line#7.

`            lines = p.stdout.readlines()
            print("Segment: {}. Attempt: {}/{} STDOUT Lines: {}".format(segment, attempts, maxAttempts, lines))`

[elf2bin.txt](https://github.com/esp8266/Arduino/files/7457289/elf2bin.txt)

anilsoni85 (commented):
Fixed raise exception level

mcspr:
> The process silently goes away sporadically without printing a single line.

Have you tried wait() / communicate() after Popen() call, and then read the stdout?
What is the stderr=subprocess.PIPE result though, is there anything there? (unlikely, but nonetheless... and I assume arduino ide would've printed it as we don't capture it?)

earlephilhower:
> Have you tried wait() / communicate() after Popen() call, and then read the stdout?

The `readlines()` call should be blocking until the process exits, as I understand the Python libs.  So, it's already effectively doing this.

Seeing if `STDERR` has anything would be useful, though, like you suggested!

The logs seem to show it failing after succeeding several times, so the race condition thing seems harder to believe now (unless AV/etc. is non-deterministic and only checks things after a few 10s of ms).  And, still, no Windows CI failures (running on a Win2K21 Server VM I believe).

anilsoni85:
I cannot reproduce it anymore . I have modified the reattempt logic and added print statement to dump STDOUT and STDERR on console may be it will help in future.

panakos:
Citrix Workspace was causing the issue. I updated Citrix and  problem gone!

PR #8358, opened 2021-10-31
Remove warnings, errors during host tests in CI

earlephilhower:
Debug strings often included format parameters which did not exactly match
the passed in format parameters, resulting in warnings in the host test build
process like
````
/home/runner/work/Arduino/Arduino/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp:107:20: warning: format ‘%u’ expects argument of type ‘unsigned int’, but argument 3 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
  107 |         DEBUG_WIFI("[AP] SSID length %u, too long or missing!\n", ssid_len);
      |                                                                   ~~~~~~~~
      |                                                                   |
      |                                                                   size_t {aka long unsigned int}
````

Fix by applying casting or PRxxx macros as appropriate.

Also, fix one debug message which was trying to use a `String` as a `char *`:
````
/home/runner/work/Arduino/Arduino/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp: In member function ‘wl_status_t ESP8266WiFiMulti::connectWiFiMulti(uint32_t)’:
/home/runner/work/Arduino/Arduino/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp:331:34: warning: format ‘%s’ expects argument of type ‘char*’, but argument 3 has type ‘String’ [-Wformat=]
  331 |                 DEBUG_WIFI_MULTI("[WIFIM] Connecting %s\n", ssid);
````

### CODE REVIEW

File: libraries/ESP8266mDNS/src/LEAmDNS_Structs.cpp [hunk 1]
```diff
@@ -1666,7 +1666,7 @@ bool MDNSResponder::stcMDNSServiceQuery::stcAnswer::stcTTL::finalTimeoutLevel(vo
 unsigned long MDNSResponder::stcMDNSServiceQuery::stcAnswer::stcTTL::timeout(void) const
 {
 
-    uint32_t    u32Timeout = esp8266::polledTimeout::oneShotMs::neverExpires;
+    uint32_t    u32Timeout = (uint32_t)esp8266::polledTimeout::oneShotMs::neverExpires;
```
mcspr, line 1669: this actually seems wrong, we strip 64bit type into 32bit one.
neverExpires is `decltype(millis())` aka unsigned long (which is also a return type)
```cpp
static_assert(sizeof(unsigned long) == 8, ""); // on host
```

multiply below should also be fixed then?

libraries/SD/src/SD.h
```diff
@@ -136,7 +136,7 @@ class SDClass {
         uint64_t sz = size64();
 #ifdef DEBUG_ESP_PORT
 	if (sz > (uint64_t)SIZE_MAX) {
-            DEBUG_ESP_PORT.printf_P(PSTR("WARNING: SD card size overflow (%lld>= 4GB).  Please update source to use size64().\n"), sz);
+            DEBUG_ESP_PORT.printf_P(PSTR("WARNING: SD card size overflow (%lld>= 4GB).  Please update source to use size64().\n"), (long long)sz);
```
mcspr, line 139: `std::numeric_limits<uint32_t>::max()` as rhs?
plus host `(u64)val > (u64)(SIZE_MAX)` never works since `SIZE_MAX` is u64 already

libraries/SDFS/src/SDFS.h
```diff
@@ -117,7 +117,7 @@ class SDFSImpl : public FSImpl
 #ifdef DEBUG_ESP_PORT
         if (i.totalBytes > (uint64_t)SIZE_MAX) {
             // This catches both total and used cases, since used must always be < total.
-            DEBUG_ESP_PORT.printf_P(PSTR("WARNING: SD card size overflow (%lld>= 4GB).  Please update source to use info64().\n"), i.totalBytes);
+            DEBUG_ESP_PORT.printf_P(PSTR("WARNING: SD card size overflow (%lld>= 4GB).  Please update source to use info64().\n"), (long long)i.totalBytes);
```
mcspr, line 120: -//-

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -556,7 +556,7 @@ int HTTPClient::sendRequest(const char * type, Stream * stream, size_t size)
     size_t transferred = stream->sendSize(_client.get(), size);
     if (transferred != size)
     {
-        DEBUG_HTTPCLIENT("[HTTP-Client][sendRequest] short write, asked for %d but got %d failed.\n", size, transferred);
+        DEBUG_HTTPCLIENT("[HTTP-Client][sendRequest] short write, asked for %u but got %u failed.\n", (unsigned int)size, (unsigned int)transferred);
```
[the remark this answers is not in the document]
earlephilhower: Hehehe, this is actually very funny.  I started with the `%zu` everywhere and then "remembered" that the nano-printf we use doesn't understand `%z`.  I went and actually checked the code and lo and behold, it not only does but I was the one who added it.  D'oh!  I'll update these accordingly.
https://github.com/earlephilhower/newlib-xtensa/blob/85c33ba4523820f1525aa4045cb5d7393b7b1e79/newlib/libc/stdio/nano-vfprintf.c#L629-L633

libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
```diff
@@ -328,7 +328,7 @@ wl_status_t ESP8266WiFiMulti::connectWiFiMulti(uint32_t connectTimeoutMs)
 
             // Check SSID
             if (ssid == entry.ssid) {
-                DEBUG_WIFI_MULTI("[WIFIM] Connecting %s\n", ssid);
+                DEBUG_WIFI_MULTI("[WIFIM] Connecting %s\n", ssid.c_str());
```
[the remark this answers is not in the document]
earlephilhower: It actually is, but AFAICT this specific code path is not built in CI.  A `String` in a printf does generate the expected warning:
````
void setup() {
  String happy="123";
  Serial.printf("String %s\n", happy);
}
void loop() {
}
````
gives
````
/home/earle/Arduino/sketch_nov01b/sketch_nov01b.ino: In function 'void setup()':
/home/earle/Arduino/sketch_nov01b/sketch_nov01b.ino:4:15: warning: format '%s' expects argument of type 'char*', but argument 3 has type 'String' [-Wformat=]
    4 | Serial.printf("String %s\n", happy);
      |               ^~~~~~~~~~~~~
````

libraries/ESP8266mDNS/src/LEAmDNS_Structs.cpp
earlephilhower: I ended up essentially there.  This is an error/invalid condition per the comments, anyway, so it's not feeling excessively critical.

libraries/SD/src/SD.h
earlephilhower: Good catch, cleaned up!

libraries/SDFS/src/SDFS.h
earlephilhower: Updated in latest push

libraries/ESP8266mDNS/src/LEAmDNS_Structs.cpp
d-a-v: I suggest leaving it as it were (still with the warning) and fix in another PR.

Or, if you have time, it looks like this `unsigned long` / `uint32_t` should be a `decltype(esp8266::polledTimeout::oneShotMs::getTimeout())` (which is an internal `timeType` and in the end a `decltype(millis())` = `unsigned long`... without warning or suspicious casting)
see LEAmDNS.h's `esp8266::polledTimeout::oneShotMs m_TTLTimeout;`

d-a-v (approved):
Thanks !
Compilation on host is now clean !
🎉

PR #8376, opened 2021-11-16
Little update for lwIP_enc28j60

Nourbakhsh-Rad:
A little update for lwIP_enc28j60 ...

### DIFF
--- a/libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
+++ b/libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
@@ -55,92 +55,116 @@ void serial_printf(const char *fmt, ...)
 
 #define DEBUG 0
 #if DEBUG
-#define PRINTF(...) printf(__VA_ARGS__)
+    #define PRINTF(...)     printf(__VA_ARGS__)
 #else
-#define PRINTF(...) do { (void)0; } while (0)
+    #define PRINTF(...)     do { (void)0; } while (0)
 #endif
 
-#define EIE   0x1b
-#define EIR   0x1c
-#define ESTAT 0x1d
-#define ECON2 0x1e
-#define ECON1 0x1f
+#define EIE                 0x1b
+#define EIR                 0x1c
+#define ESTAT               0x1d
+#define ECON2               0x1e
+#define ECON1               0x1f
 
-#define ESTAT_CLKRDY 0x01
-#define ESTAT_TXABRT 0x02
+#define ESTAT_CLKRDY        0x01
+#define ESTAT_TXABRT        0x02
 
-#define ECON1_RXEN   0x04
-#define ECON1_TXRTS  0x08
+#define ECON1_RXEN          0x04
+#define ECON1_TXRTS         0x08
 
-#define ECON2_AUTOINC 0x80
-#define ECON2_PKTDEC  0x40
+#define ECON2_AUTOINC       0x80
+#define ECON2_PKTDEC        0x40
 
-#define EIR_TXIF      0x08
+#define EIR_TXIF            0x08
 
-#define ERXTX_BANK 0x00
+#define ERXTX_BANK          0x00
 
-#define ERDPTL 0x00
-#define ERDPTH 0x01
-#define EWRPTL 0x02
-#define EWRPTH 0x03
-#define ETXSTL 0x04
-#define ETXSTH 0x05
-#define ETXNDL 0x06
-#define ETXNDH 0x07
-#define ERXSTL 0x08
-#define ERXSTH 0x09
-#define ERXNDL 0x0a
-#define ERXNDH 0x0b
-#define ERXRDPTL 0x0c
-#define ERXRDPTH 0x0d
+#define ERDPTL              0x00
+#define ERDPTH              0x01
+#define EWRPTL              0x02
+#define EWRPTH              0x03
+#define ETXSTL              0x04
+#define ETXSTH              0x05
+#define ETXNDL              0x06
+#define ETXNDH              0x07
+#define ERXSTL              0x08
+#define ERXSTH              0x09
+#define ERXNDL              0x0a
+#define ERXNDH              0x0b
+#define ERXRDPTL            0x0c
+#define ERXRDPTH            0x0d
 
-#define RX_BUF_START 0x0000
-#define RX_BUF_END   0x0fff
+#define RX_BUF_START        0x0000
+#define RX_BUF_END          0x0fff
 
-#define TX_BUF_START 0x1200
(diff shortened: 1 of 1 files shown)


libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
```diff
@@ -742,3 +766,106 @@ ENC28J60::readFrameData(uint8_t *buffer, uint16_t framesize)
...
+    
+    if(LEDAcfg)             temp &= LEDA_MASK;
+    if(LEDBcfg)             temp &= LEDB_MASK;
+	
+    phyWrite(PHLCON, (temp |(uint16_t)(LEDAcfg <<4) |LEDBcfg |0x3006));
+}
+
+//--------------------------------------------------------
+bool
+ENC28J60::resetEther(void)
```
d-a-v, line 845: Why is it necessary ?

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
```diff
@@ -742,3 +766,106 @@ ENC28J60::readFrameData(uint8_t *buffer, uint16_t framesize)
...
+{
+    return reset();
+}
+
+uint8_t
+ENC28J60::hardwareStatus(void)
```
d-a-v, line 851: Is it really `hardwareStatus()` or hardwareRevision/hardwareVersion() ?
We already have [`::readrev()`](https://github.com/esp8266/Arduino/blob/bf2882d8e003c278017639fa964248fedb942454/libraries/lwIP_enc28j60/src/utility/enc28j60.cpp#L308L322) which seems to be identical.

libraries/lwIP_enc28j60/src/utility/enc28j60.h
```diff
@@ -41,6 +41,35 @@
...
 #include <SPI.h>
 
+// PHLCON: PHY module LED control register
+//
+//      R R R R    L  L  L  L    L  L  L  L             R
+//      E E E E    A  A  A  A    B  B  B  B    L  L  S  E
+//      S S S S    C  C  C  C    C  C  C  C    F  F  T  S
+//      E E E E    F  F  F  F    F  F  F  F    R  R  R  E
+//      R R R R    G  G  G  G    G  G  G  G    Q  Q  C  R
+//      V V V V    3  2  1  0    3  2  1  0    1  0  H  V
```
d-a-v, line 51: Can this table receive more comments on what it is for ?

d-a-v (commented):
Beside the led API addition, some methods are aliased or duplicated.
Can you please enlighten us about your proposal ?

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
Nourbakhsh-Rad: Sometimes the ENC28J60 does not work properly (freezes) and needs to be reconfigured

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
Nourbakhsh-Rad: - This function is similar to readrev function (it is private)
- If this function returns 0xFF, that means "Hardware Error" (incorrect hardware revision)

libraries/lwIP_enc28j60/src/utility/enc28j60.h
Nourbakhsh-Rad: This is LED control register that described in the ENC28J60 document (page 11) and used with LEDsConfig function

usage:
`LEDsConfig(LED_LINK_STATUS, LED_TRANS_RECV);                     // LEDA, LEDB`
`LEDsConfig(LED_BLINK_SLOW, LED_NO_CHANGE);                     // LEDA, LEDB`

http://ww1.microchip.com/downloads/en/devicedoc/39662a.pdf

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
Nourbakhsh-Rad: This function can be changed this way
```
uint8_t
ENC28J60::hardwareStatus(void)
{
	return readrev();		/* 0xFF = Hardware Error */
}
```

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
```diff
@@ -742,3 +766,106 @@ ENC28J60::readFrameData(uint8_t *buffer, uint16_t framesize)
...
+    // set the PHY register address
+    writereg(MIREGADR, address);
+    setregbitfield(MICMD, 0x01);                // MII Read Enable [MIIRD]
+    
+    setregbank(REG_BANK_3);
+	
+    // wait until the PHY read completes
+    while(readreg(MISTAT) & 0x01){
+        delay(10);
+        wdt_reset();
```
dok-net, line 787: IMHO, delay() takes care of resetting the watchdog.

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
```diff
@@ -742,3 +766,106 @@ ENC28J60::readFrameData(uint8_t *buffer, uint16_t framesize)
...
+    // write the PHY data
+    writereg(MIWRL, data & 0xff);
+    writereg(MIWRH, data >> 8);
+    
+    setregbank(REG_BANK_3);
+    
+    // wait until the PHY write completes
+    while(readreg(MISTAT) & 0x01){
+        delay(10);
+        wdt_reset();
```
dok-net, line 818: delay implies a WDT reset.

dok-net:
Please, while you are updating this lib, could you check about the fate of
> src/utility/enc28j60.h:    void clock_delay_usec(uint16_t dt);

I don't see any definition of the function anywhere.

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
d-a-v: Ah you need to call it from "userland".
Instead of duping it, why don't you move it to the public section ?

libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
d-a-v: Or it could be, as for `reset()`, moved to the public section without changing its name ?

libraries/lwIP_enc28j60/src/utility/enc28j60.h
d-a-v: Do you plan to add this informations as source-code comments in your proposal ?

PR #8379, opened 2021-11-20
adding esp32's ArduinoOTAClass::end function to esp8266

Phil1pp:
This adds the end() function for the ArduinoOTA class.

ESP32 has this implemented, however ESP8266 was lacking this.


earlephilhower (approved):
LGTM, thanks!

PR #8386, opened 2021-11-26
LittleFS: add overrides for Stream::send

### LINKED ISSUE #8384: LittleFS missing availableForWrite()
### Basic Infos

#### Platform

- Hardware:					[ESP-12F]
- Core Version:			[3.0.2]
- Development Env:		[VSMicro]

### Settings in IDE

- Module:  			[LOLIN(WEMOS) D1 D2 & mini]
- Flash Mode:			[DIO]
- Flash Size: 		[4MB]
- lwip Variant:   [Higher Bandwidth]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600] (serial upload only)

### Problem Description

Attempting to use HTTPClient to stream directly to a LittleFS `File` times out with error code -11. This is due to the `LittleFSFileImpl` not implementing `availableForWrite()`, so the base implementation returns 0 and the streaming can't write and times out. I added this implementation but I do not know if this is the right thing to return. It probably isn't but hey it works! It would probably work even if I returned 1 so I'm not going to throw myself a parade or anything for getting it working.
```
// in LittleFS.h LittleFSFileImpl
    int availableForWrite() override {
      if (!_opened || !_fd) {
        return 0;
      }
      return _fs->_lfs_cfg.block_size;
    }
```

```cpp
void taskcbIcsCheck(void)
{
  Serial.println(F("Performing ICS check..."));

  WiFiClient *client;
  if (strncmp(ICS_URL, "https://", 8) == 0)
  {
    client = new BearSSL::WiFiClientSecure();
    ((BearSSL::WiFiClientSecure *)client)->setInsecure();
  }
  else
    client = new WiFiClient();

  HTTPClient *https = new HTTPClient();
  int httpCode = -1;
  if (https->begin(*client, ICS_URL))
  {
    https->setUserAgent("Google-Calendar-Importer");
    https->setFollowRedirects(HTTPC_FORCE_FOLLOW_REDIRECTS);
    httpCode = https->GET();
    Serial.print(F("Response code ")); Serial.println(httpCode, DEC);

    if (httpCode == HTTP_CODE_OK) {
      File f = LittleFS.open("cal.ics", "w");
      int n = https->writeToStream(&f);
      f.close();
      Serial.print("Download complete n="); Serial.println(n, DEC);

      g_IcsLastCheck = time(nullptr);
      taskIcsRead.restartDelayed(100);
    }
    https->end();
  }
  else
    Serial.print("Connect failed");

  delete https;
  delete client;
}
```

### Debug Messages

```
[HTTP-Client][handleHeaderResponse] size: 8718
// and then nothing, return code -11
```

d-a-v:
fixes #8384 

`::availableForWrite()` could unconditionnally return a full sector as free space like proposed in #8384.

This proposal aims at returning a non-0 size fitting the current file block.
If the block is full and the filesystem is not full, it returns a full-FS-block size.
This way, the API allows users to take proper actions without losing data from the input stream when the filesystem is full, even if there is generally little to do in such case.

### DIFF
--- a/cores/esp8266/FS.h
+++ b/cores/esp8266/FS.h
@@ -118,6 +118,18 @@ public:
     time_t getCreationTime();
     void setTimeCallback(time_t (*cb)(void));
 
+    // Stream::send configuration
+
+    bool inputCanTimeout () override {
+        // unavailable data can't become later available
+        return false;
+    }
+
+    bool outputCanTimeout () override {
+        // free space for write can't increase later
+        return false;
+    }
+
 protected:
     FileImplPtr _p;
     time_t (*_timeCallback)(void) = nullptr;
--- a/libraries/LittleFS/examples/SpeedTest/SpeedTest.ino
+++ b/libraries/LittleFS/examples/SpeedTest/SpeedTest.ino
@@ -132,6 +132,16 @@ void DoTest(FS *fs) {
   f.close();
   stop = millis();
   Serial.printf("==> Time to read 64KB in 1b chunks = %lu milliseconds = %s\n", stop - start, rate(start, stop, 65536));
+
+
+  start = millis();
+  auto dest = fs->open("/test1bw.bin", "w");
+  f = fs->open("/test1b.bin", "r");
+  auto copysize = f.sendAll(dest);
+  dest.close();
+  stop = millis();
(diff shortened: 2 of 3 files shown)


earlephilhower (approved):
Looks good to me.  I didn't know about the `lfs_file_t.off` but it does seem to be what you're looking for.  Nice!

PR #8388, opened 2021-11-27
Allow configuring Wire.h buffer size

flaviut:
This is needed since in some applications, we need much much larger buffers.

See https://github.com/flaviut/emporia-vue2-reversing/issues/1#issuecomment-978174665

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -28,8 +28,11 @@
 #include "Stream.h"
 
 
-
+#ifndef ARDUINO_WIRE_BUFFER_LENGTH
 #define BUFFER_LENGTH 128
+#else
+#define BUFFER_LENGTH ARDUINO_WIRE_BUFFER_LENGTH
+#endif
 
 class TwoWire : public Stream
 {


d-a-v (approved):

flaviut:
After further investigation prompted by @Maelstrom96's comments elsewhere, this isn't sufficient to allow for the 284 byte payloads we need to handle.

We'd also need `#if`s to make sure that the various types aren't `uint8_t`, but are promoted to `uint16_t` when the size of the payload is increased. I'd be happy to do that work, if the maintainers would chime in on if that'd be an acceptable solution, or if it'd be better to fork the Wire library.

flaviut:
Whoops! I shouldn't be programming before fully waking up! I'm actually looking to do this on the ESP32, not ESP8266.

I'd be happy to port any changes here too, please re-open this if you'd like me to do that.

mcspr:
See https://github.com/esp8266/Arduino/pull/8388
Might be a good idea to share the name of the setting

d-a-v:
Using int16 or even int32 sometimes generate less code / is faster than when using int8, on this architecture.

PR #8392, opened 2021-11-28
compiler.libraries.ldflags= was missing, generating an error when trying to link precompiled libraries

marcel-licence:
This line is required to be compatible with the Arduino library specification

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -120,7 +120,7 @@ recipe.S.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.cpreprocessor.fla
 recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{archive_file_path}" "{object_file}"
 
 ## Combine gc-sections, archives, and objects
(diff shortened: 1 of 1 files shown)


mcspr (changes requested):
Should this also be added to the combine pattern here? I have noticed esp32 platform.txt using this property
https://github.com/espressif/arduino-esp32/blob/399f4ecbb3a4cef21e2bffa37adb6190356dfb76/platform.txt#L151
https://github.com/esp8266/Arduino/blob/d5444c4aa38bff01269cfbd98a13a1454d0c62df/platform.txt#L122

mcspr:
Also, some references
https://arduino.github.io/arduino-cli/0.20/platform-specification/#recipes-for-linking

Turns out this was reported to the Arduino repo last year
https://github.com/arduino/Arduino/issues/10473
https://github.com/arduino/arduino-cli/pull/840

marcel-licence:
I compiled with and without the requested change and couldn't spot a difference. I am not so familar with arduino to understand what this line should do.
If you like I can just add this to the push request but it was not required to link the library.

mcspr:
> I compiled with and without the requested change and couldn't spot a difference. I am not so familar with arduino to understand what this line should do.
If you like I can just add this to the push request but it was not required to link the library.

Please do. We should also fix the comment to reference the reason why we need this?

---
tldr; this function of the arduino-cli builder uses the combine pattern:
https://github.com/arduino/arduino-cli/blob/e31a717ebc34fff80892dbbe25a9d4581906e13a/legacy/builder/phases/linker.go#L66-L123

If your project happens to have a lot of files and / or many external libraries, not including the compiler.libraries.ldflags would cause compilation error when building that 'combined' archive file. I have rebuild the arduino-cli with condition changed from 30000 to 1, and grabbed https://github.com/BoschSensortec/BSEC-Arduino-library to reproduce this.

marcel-licence:
Okay, I've added this to the pattern group

PR #8393, opened 2021-11-28
GCC 10.3 Bugfix 1 - Constant literal address fix

earlephilhower:
Fixes a hard-to-track bug in GCC 10.x.
https://github.com/earlephilhower/newlib-xtensa/issues/19

GCC 10.3 had an issue with addressing constant literals which would result
in crazy offsets being used and random crashes in production.  Update
with an upstream GCC 11 bugfix.


mcspr (approved):

bwjohns4:
@mcspr , I see your PIO package for this (mcspr/toolchain-xtensa @ ~5.100300.211127). Would it be possible to upload the binaries for 'darwin_arm64'?

mcspr:
@bwjohns4 PIO registry are just repackages files from https://github.com/earlephilhower/esp-quick-toolchain/releases. [Plus, I don't think we could build that GCC right now without some manual patching?](https://github.com/iains/gcc-darwin-arm64)

bwjohns4:
@mcspr , So it's the exact same files from the release packaged for PIO? Wonder how I tell my M1 MAC to just use the x64 intel version and virtualize it with Rosetta??

Do you know why the default pckage works fine on the M1 MAC but this toolchain does not and instead errors out on could not find darwin_arm64?

mcspr:
Hmm... The registry payload returns x86_64 version for both, I suppose package.json was modified to include it
```json
{
  "checksum": {"sha256": "0a40e1cc5f6f9cedd85d3d4dac548bfb5535c015c40cb10a7755f7dbd3f1db4c"},
  "download_url": "https://dl.registry.platformio.org/download/platformio/tool/toolchain-xtensa/2.100300.210717/toolchain-xtensa-darwin_x86_64-2.100300.210717.tar.gz",
  "name": "toolchain-xtensa-darwin_x86_64-2.100300.210717.tar.gz",
  "size": 75089523
}
```

mcspr:
For a follow-up, does darwin_arm64 arch work now?
(ref. [registry api response](https://api.registry.platformio.org/v3/packages/mcspr/tool/toolchain-xtensa), there is now a special version with a +darwin suffix. I assume that will work, but I wonder if it needed to be a real version bump)

bwjohns4:
It does work now! Did that GCC build require manual patching? How should I regard the integrity of this compiler vs some of the other OS compilers? Won't they all generate slightly different firmware binary since they're different GCC binaries themselves? Is any one system more reliable/trusted than another?

Issue #8391, opened 2021-11-28
TwoWire: use 32 bits for buffer size and offset


d-a-v:
For the following reasons, TwoWire's offset and size members should be enhanced to more than 8 bits:

https://github.com/esp8266/Arduino/pull/8388#issuecomment-980780082
https://github.com/esp8266/Arduino/pull/8390#pullrequestreview-817212903

PR #8398, opened 2021-12-05
Wire buffer length improvments.

flaviut:
## Enable I2C_BUFFER_LENGTH definition for Wire lib
 
Based on
https://github.com/paclema/arduino-esp32/commit/375a89ed58f341a8aae26dab505713f3ef0edb3d
 
We should have been very careful when #defining things to not use a name
that could conflict with the user's own code, so this marks that old
define as deprecated.
 
If you opt-into the new behavior, you do not get the old BUFFER_LENGTH
constant.

## Increase buffer indexing variable size
 
I looked over the users of these variables and they should be fine with
no additional changes. The existing methods already have an option to
use size_t rather than uint8_t.
 
There's a few methods which return int instead of size_t, which isn't
great from a portability perspective but will be fine since this only is
designed to run on the ESP8266.'

Fixes https://github.com/esp8266/Arduino/issues/8391

------

Feel free to use this patchset in full or in part.

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -28,11 +28,10 @@
 #include "Stream.h"
 
 
-
 #ifndef I2C_BUFFER_LENGTH
+// DEPRECATED: Do not use BUFFER_LENGTH, prefer I2C_BUFFER_LENGTH
 #define BUFFER_LENGTH 128
-#else
-#define BUFFER_LENGTH I2C_BUFFER_LENGTH
+#define I2C_BUFFER_LENGTH BUFFER_LENGTH
 #endif
 
 class TwoWire : public Stream
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -38,13 +38,13 @@ class TwoWire : public Stream
 {
 private:
     static uint8_t rxBuffer[];
-    static uint8_t rxBufferIndex;
-    static uint8_t rxBufferLength;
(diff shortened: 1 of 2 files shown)


libraries/Wire/Wire.h
```diff
@@ -28,24 +28,23 @@
 #include "Stream.h"
 
 
-
 #ifndef I2C_BUFFER_LENGTH
+// DEPRECATED: Do not use BUFFER_LENGTH, prefer I2C_BUFFER_LENGTH
```
flaviut, line 32: I expect this may be a bit controversial. My idea is to use the new `I2C_BUFFER_LENGTH` functionality as a carrot to convince people to switch to the new define.

flaviut:
cc @d-a-v

d-a-v (approved):
LGTM

d-a-v:
It has to be noted that the resulting flash occupation is decreased with this PR.
I don't know whether a note about that in OP would be useful.

So for example when one needs an `uint8_t`, it may be always good in internal code to use [`uint_fast8_t`](https://en.cppreference.com/w/c/types/integer) instead
(except when code must ensure that `255+1==0`).

flaviut:
> It has to be noted that the resulting flash occupation is decreased with this PR.

That's interesting, and not something I expected. I've updated the commit message with this information.

earlephilhower (approved):
LGTM, thanks!

PR #8406, opened 2021-12-14
Backported getLocalTime from ESP32 Arduino

h1aji:
Backporting getLocalTime from ESP32 for better compatibility.
Also fixed configTzTime


cores/esp8266/Arduino.h
```diff
@@ -269,16 +269,15 @@ void setTZ(const char* tz);
 void configTime(int timezone, int daylightOffset_sec, const char* server1,
     const char* server2 = nullptr, const char* server3 = nullptr);
 
-void configTime(const char* tz, const char* server1,
```
earlephilhower, line 272: Why would you delete the alternative API?  Restore, please.

cores/esp8266/time.cpp
```diff
@@ -187,26 +187,44 @@ void configTime(int timezone_sec, int daylightOffset_sec, const char* server1, c
...
+void setTZ(const char* tz)
+{
     char tzram[strlen_P(tz) + 1];
     memcpy_P(tzram, tz, sizeof(tzram));
     setenv("TZ", tzram, 1/*overwrite*/);
     tzset();
 }
 
-void configTime(const char* tz, const char* server1, const char* server2, const char* server3)
+void configTzTime(const char* tz, const char* server1, const char* server2, const char* server3)
```
earlephilhower, line 198: Don't rename a well-known function.  That breaks backwards compatibility for no reason.

cores/esp8266/time.cpp
```diff
@@ -187,26 +187,44 @@ void configTime(int timezone_sec, int daylightOffset_sec, const char* server1, c
...
 }
 
+/*
+ * Backported from ESP32 Arduino
+ */
+bool getLocalTime(struct tm * info, uint32_t ms)
+{
+    uint32_t start = millis();
+    time_t now;
+    while((millis()-start) <= ms)
```
earlephilhower, line 215: Indent has gone wacky in this function.

earlephilhower (changes requested):

cores/esp8266/time.cpp
d-a-v: FYI, it is already [there](https://github.com/esp8266/Arduino/blob/55ef3e73978ad010d44ad7e4c9bd4b4f32455989/cores/esp8266/Arduino.h#L276-L280)

PR #8407, opened 2021-12-14
Backported getLocalTime from ESP32 Arduino

h1aji:
New PR. Removed previosly added modifications.
This time just importing new getLocalTime from ESP32 Arduino core

### DIFF
--- a/cores/esp8266/time.cpp
+++ b/cores/esp8266/time.cpp
@@ -187,8 +187,8 @@ void configTime(int timezone_sec, int daylightOffset_sec, const char* server1, c
     sntp_init();
 }
 
-void setTZ(const char* tz){
-	
+void setTZ(const char* tz)
+{
     char tzram[strlen_P(tz) + 1];
     memcpy_P(tzram, tz, sizeof(tzram));
     setenv("TZ", tzram, 1/*overwrite*/);
@@ -198,13 +198,31 @@ void setTZ(const char* tz){
 void configTime(const char* tz, const char* server1, const char* server2, const char* server3)
 {
     sntp_stop();
-
     setServer(0, server1);
     setServer(1, server2);
     setServer(2, server3);
-	setTZ(tz);
-	
+    setTZ(tz);
     sntp_init();
+}
+
+/*
+ * Backported from ESP32 Arduino
+ */
+bool getLocalTime(struct tm * info, uint32_t ms)
+{
+    uint32_t start = millis();
+    time_t now;
+    while ((millis()-start) <= ms)
+    {
+      time(&now);
+      localtime_r(&now, info);
+      if (info->tm_year > (2016 - 1900))
+      {
+        return true;
+      }
+      delay(10);
+    }
+    return false;
 }
 
 static BoolCB _settimeofday_cb;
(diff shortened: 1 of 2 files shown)


d-a-v:
It seems that you have to rename or remove [this one](https://github.com/esp8266/Arduino/blob/55ef3e73978ad010d44ad7e4c9bd4b4f32455989/libraries/LittleFS/examples/LittleFS_Timestamp/LittleFS_Timestamp.ino#L22-L42)

h1aji:
the only part that appears different is -  ` uint32_t count = ms / 10;`

I assume we can modify https://github.com/esp8266/Arduino/blob/55ef3e73978ad010d44ad7e4c9bd4b4f32455989/libraries/LittleFS/examples/LittleFS_Timestamp/LittleFS_Timestamp.ino#L156 and make it 
`  getLocalTime(&tmstruct, 500);
`
then delete **getLocalTime** from that sketch

@earlephilhower what do you think?

h1aji:
@d-a-v does this look ok now?

cores/esp8266/time.cpp
```diff
@@ -198,15 +198,43 @@ void setTZ(const char* tz){
...
+ 
+  time(&now);
+  localtime_r(&now, info);
+ 
+  if (info->tm_year > (2016 - 1900)) 
+  {
+    return true;
+  }
+ 
+  while (count--)
```
d-a-v, line 225: With this loop, maximum timeout is 10 times the value of `ms`, which is why Earle divided it by 10.
Original and probably later updated / fixed esp32 code [is there](https://github.com/espressif/arduino-esp32/blob/34125cee1d1c6a81cd1da07de27ce69d851f9389/cores/esp32/esp32-hal-time.c#L91) with a more accurate loop.
Do you think you can improve your proposal ?

d-a-v:
About the CI error, `time.cpp` is not included in host tests because it holds standard functions already available on host's system.
I suggest moving the function into a new file (`timehelper.cpp`?) and adding this file in the [host build Makefile](https://github.com/esp8266/Arduino/blob/42aa0e6d253603a136cd813b7b834364791b89ad/tests/host/Makefile#L99-L100).

cores/esp8266/time.cpp
h1aji: Okay then, do you have an idea how LittleFS_Timestamp.ino sketch should be modified?

h1aji:
It looks like this PR causing lots of troubles. Shall I close it?

d-a-v:
> Shall I close it?

No :) if you want to do it.

cores/esp8266/time.cpp
d-a-v: I think it is OK to use the esp32 implementation globally and remove littlefs test's one.

h1aji:
I dont mind if you add it yourself, otherwise I am happy with this commit, I wasnt sure how to change that sketch file, that why i though asking you

d-a-v:
> otherwise I am happy with this commit

I used updated esp32 code in #8413, moved the function into another file so it can be used for the esp8266, emulation on host, and host tests, and removed the old local copy from the littleFS test.

You can get inspiration from it to fix the CI builds if you like.

h1aji:
@d-a-v nice. tthank you.
I will close this PR

PR #8409, opened 2021-12-15
Emulation on host: updates

d-a-v:
Two enhancements for Emulation on host:

- allow scheduled functions
- include Hash functions


tests/host/common/Arduino.cpp
```diff
@@ -42,6 +44,8 @@ extern "C" unsigned long micros()
 
 extern "C" void yield()
 {
+    run_scheduled_functions();
+    run_scheduled_recurrent_functions();
```
mcspr, line 48: shouldn't yield() only run run_scheduled_recurrent_functions() to be consistent with core_esp8266_main.cpp?
(also note of `loop_end()`)

tests/host/common/ArduinoMain.cpp
```diff
@@ -309,6 +309,7 @@ int main (int argc, char* const argv [])
 			usleep(1000); // not 100% cpu, ~1000 loops per second
 		loop();
 		check_incoming_udp();
+		yield(); // call scheduled functions
```
mcspr, line 312: `loop_end()`?
https://github.com/esp8266/Arduino/blob/42aa0e6d253603a136cd813b7b834364791b89ad/cores/esp8266/core_esp8266_main.cpp#L231-L235

PR #8413, opened 2021-12-16
import getLocalTime() from esp32/Arduino

d-a-v:
follows and closes  #8407


cores/esp8266/timehelper.cpp
```diff
@@ -0,0 +1,19 @@
+
```
earlephilhower, line 1: This might better be put into the existing `time.cpp` file instead of making a new one.

Also, a new file would need the GPL headers/etc.

earlephilhower (changes requested):

cores/esp8266/timehelper.cpp
d-a-v: > This might better be put into the existing time.cpp file instead of making a new one.

`time.cpp` is not compiled in host mode. This new file would.

d-a-v: The function is now in `time.cpp`, the other new file is removed.

earlephilhower (approved):
LGTM!

PR #8418, opened 2021-12-19
WiFiServer - don't inherit from Server class

jandrassy:
The base type Server is useless because it doesn't declare any method to get a client so it can't be used as base type reference or pointer. 
Server inherits from Print, but ESP8266WiFi library's WiFiServer doesn't implement the print-to-all-clients functionality, inheriting from Print only adds virtual methods overhead.
My addition the ArduinoWiFiServer class implements the print-to-all-clients functionality so it needs to inherit from Print.

I created and maintain two networking libraries (WiFiEspAT and EnternetENC) and both don't use the Server class and have a separate class for server with print-to-all-clients functionality. There are no issues with this.


jandrassy:
the failed check looks like a problem of the check

d-a-v (approved):
Looks good.
It is a non-sense for WiFiServer to inherit from Print without implementing properly Print functions.
OK to assume it is not a breaking change.

PR #8419, opened 2021-12-19
WiFiServer - 'rename' available() to accept()

jandrassy:
If this PR gets positive reviews, I will add commits with examples and documentation changes.

The available() method in WiFiServer and WiFiServerSecureBearSSL is not implemented as the available() in the Arduino WiFi library and other WiFi libraries by Arduino. The implementation of available() matches the Ethernet library's accept() method added in 2018. 

This PR marks available() in WiFiServer as deprecated and adds the accept() method with implementation identical to available().  

My addition to the ESP8266WiFi library, the ArduinoWiFiServer class has available() implemented the Arduino way.


jandrassy:
the checks treat (deprecated) warning as errors so I had to replace available() to accept() in libraries and examples.

now only the mock test fails with ` /home/runner/work/Arduino/Arduino/tests/host/../../libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h:284: undefined reference to `WiFiServer::accept()'`, but I don't know how to fix that

d-a-v:
> but I don't know how to fix that

```cpp
diff --git a/tests/host/common/MockWiFiServer.cpp b/tests/host/common/MockWiFiServer.cpp
index 132ca5e8..7b449845 100644
--- a/tests/host/common/MockWiFiServer.cpp
+++ b/tests/host/common/MockWiFiServer.cpp
@@ -55,9 +55,8 @@ WiFiServer::WiFiServer (uint16_t port)
        _port = port;
 }
 
-WiFiClient WiFiServer::available (uint8_t* status)
+WiFiClient WiFiServer::accept ()
 {
-   (void)status;
        if (hasClient())
                return WiFiClient(new ClientContext(serverAccept(pcb2int(_listen_pcb))));
        return WiFiClient();
```

d-a-v (approved):
LGTM, it is not a breaking change but a deprecation.
Per [`Ethernet.accept()`](https://www.arduino.cc/reference/en/libraries/ethernet/accept/) it seems that our implementation for `server::available()` does not comply with Arduino. It was the `::accept()` implementation.

doc/esp8266wifi/server-class.rst
```diff
@@ -12,13 +12,24 @@ Methods documented for the `Server Class <https://www.arduino.cc/en/Reference/Wi
...
+For most use cases the basic WiFiServer class of the ESP8266WiFi library is suitable.
+
 Methods and properties described further down are specific to ESP8266. They are not covered in `Arduino WiFi library <https://www.arduino.cc/en/Reference/WiFi>`__ documentation. Before they are fully documented please refer to information below.
 
+accept
+~~~~~~
+
+Method ``accept()`` returns a waiting client connection. `accept() <https://www.arduino.cc/en/Reference/EthernetServerAccept?'__ is documented for the Arduino Ethernet library.
+
+Note: Method ``available`` in the WiFiServer class is deprecated. It doesn't work as documented for the Arduino WiFi library. (It works the same way as ``accept``.)
```
mcspr, line 26: should the paragraph at the top and this be combined?

> In ESP8266WiFi library, the ArduinoWiFiServer class implements ``available()`` and the write-to-all-clients functionality as described in the Arduino WiFi library reference, while the WiFiServer class implements ``available()`` as ``accept()`` (see below). The PageServer example shows how ``available`` and the write-to-all-clients works. 
> 
...
> available
> \~~~~~~~~~
> Deprecated since Core 3.1.0, see :ref:`accept` (or however sphinx links work in the same doc)

ref.
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

jandrassy: sorry, but "while the WiFiServer class implements available() as accept() (see below). " is only confusing there, I think. 
I had something similar there and I removed it. I thought about all these doc changes many hours.

Why document `available` for WiFiServer now when it is deprecated?

If somebody wonders about 'accept()' they find the explanation in the accept section.
If somebody searches for 'true' available() they will find it next to the links to Arduino doc.

mcspr: I asked exactly b/c I was confused what reference is referencing what funtionality :) Considering we mention both, shouldn't we mention the difference just in one place to avoid jumping contexts? Similarly to the reason why those are different, perhaps a brief explanation of the expected behaviour of the method also helps?

I was suggesting to document the deprecation, and sphinx allows a pretty nice syntax. When the time comes, perhaps it could be moved somewhere at the bottom

jandrassy: is there a way to preview the documentation from PR or locally?

mcspr:
```
$ pip install -r doc/requirements.txt # or use venv
$ make -C doc/ html # just make sure sphinx-build is in PATH
```

result will be in the doc/_build/html

jandrassy: > ```
> $ pip install -r doc/requirements.txt # or use venv
> $ make -C doc/ html # just make sure sphinx-build is in PATH
> ```
> 
> result will be in the doc/_build/html

thank you, that was simple.

it is not possible to use :ref: without tagging the headings, but that would better done with a separate PR.

PR #8429, opened 2022-01-03
Fix umm_blocks()

mhightower83:
Extracted fix from upstream for umm_blocks() - On allocations that were too
large, umm_blocks() could return an incorrectly truncated value when the result
is cast to uint16_t.


d-a-v (approved):
LGTM

PR #8432, opened 2022-01-05
platform.txt - added espota as programmer

jandrassy:
as a solution for IDE 'network port' detection problems or delays.

espota.py is configured in platform.txt as a programmer. 
In programmers.txt then the user can create configurations for specific projects with fixed IP address.  
(Unfortunately programmers.local.txt is not supported by Arduino builder.)

the 'fake programmer' solution is used with my generic ArduinoOTA library for networking libraries without multicast UDP support or as workaround for 'network port' problems or simply for convenience.


jandrassy:
the failed check doesn't seem to be related to this PR

PR #8451, opened 2022-01-17
added weak fnc add_custom_offer_options

bwjohns4:
Here is an attempt to incorporate add_custom_offer_options(). This is a weak function that is called at the end of add_offer_options() that allows another library to redefine the function to hook into and add custom DHCP options.

One question I have is what restrains the size of optptr? How much can it grow (via *optprt++) before causing problems? Is it in an allocated space already?

### DIFF
--- a/cores/esp8266/LwipDhcpServer.cpp
+++ b/cores/esp8266/LwipDhcpServer.cpp
@@ -367,6 +367,20 @@ uint8_t* DhcpServer::add_msg_type(uint8_t *optptr, uint8_t type)
     *optptr++ = type;
     return optptr;
 }
+
+extern "C" uint8_t* __add_custom_offer_options(uint8_t *optptr){
+    /* Example showing how to add Captive Portal Option to DHCP Server
+    *optptr++ = 114;
+    *optptr++ = 19;
+    char captive_uri[] = "http://192.168.4.1/";
+    for(int i = 0; i<19; i++){
+    *optptr++ = captive_uri[i];
+    }
+    */
+    return optptr; //do nothing by default. Libraries can then redefine this weak function in order to hook into DHCP server options
+}
+extern "C" uint8_t* add_custom_offer_options(uint8_t *optptr) __attribute__ ((weak, alias("__add_custom_offer_options")));
+
 ///////////////////////////////////////////////////////////////////////////////////
 /*
     DHCP msg offer
@@ -469,6 +483,8 @@ uint8_t* DhcpServer::add_offer_options(uint8_t *optptr)
     *optptr++ = 2;
 #endif
 
+    optptr = add_custom_offer_options(optptr); //this allows external code to redefine this weak function and hook into DHCP Offer.
+    
     return optptr;
 
 #undef ipadd


earlephilhower:
@bwjohns4 the CI failed due to indent/spacing issues.  Run `tests/restyle.sh` and commit/push the changes and it should clear up.

bwjohns4:
@earlephilhower , do I just run that script from the root of the repo directory? I ran it there and nothing appears to have changed to the file. Do I need to pass any arguments into the script? Here are the error messages that I get:

image

earlephilhower:
Your screenshot looks like Ubuntu terminal.  `sudo apt install astyle` or equivalent will be required to get the style tool installed, then you can re-run the script to do the fixes.

d-a-v:
> One question I have is what restrains the size of optptr? How much can it grow (via *optprt++) before causing problems? Is it in an allocated space already?

https://github.com/esp8266/Arduino/blob/9fcf14f81fa9be589530e9596b7c5a264dc81ee8/cores/esp8266/LwipDhcpServer.cpp#L72

https://github.com/esp8266/Arduino/blob/9fcf14f81fa9be589530e9596b7c5a264dc81ee8/cores/esp8266/LwipDhcpServer.cpp#L532

This code is an as-is-copy of the original espressif dhcp server from the initial lwip-1.4.
As we can see there's not much control.
After a second look, what about holding a trivial linked list of callbacks that would be callable from `::send_offer()` ?
These callbacks would pass the option pointer and the remaining available size (`end - &m->options[4]`).

bwjohns4:
Where in Arduino is the DCHP server instantiated and where in user/library code would this happen? What about adding a setter to the DHCPserver class to pass in custom options as well as a bool pointer that can be used to pass back true or false success of adding in the custom options.

Then when it's time to add the custom options just check that `end - &m->options[0]` is less than 312 - 1 (subtract one for the 255 byte appended to the end). If it's too big then fail and pass the status back to the user/library. I don't think I'm 100% clear on the best way to communicate the success/fail back to the user/lib.

How does this align with your suggestion of the linked list callable from `send_offer()`

d-a-v:
> Where in Arduino is the DCHP server instantiated and where in user/library code would this happen?

It is [globally instantiated](https://github.com/esp8266/Arduino/blob/f4ef6fd7d82973895e8c1460faef5ce1693610d0/cores/esp8266/LwipDhcpServer-NonOS.cpp#L34). Nonosdk firmware initially calls `dhcps_start()` which is defined in the same file.

> [...] How does this align with your suggestion of the linked list callable from send_offer()

`send_offer()` is called much later than when dhcp server is initialized, and that, on user request after explicitly installing AP mode.
So user sketch has time to set something up to be taken into account.
I was mentioning a linked list. A simple mechanism may be sufficient:
In the dhcp server class, a `std::function<bool(uint8_t* option, size_t remain)>` member, and a setter.

This API can be overloaded if one need to *separately* add more options, which I doubt.

bwjohns4:
What would be the use case for the linked list here? Build a linked list of options and their associated parameters to add?

d-a-v:
> What would be the use case for the linked list here? Build a linked list of options and their associated parameters to add?

No. It was in case dhcp options would have to be added more than once and from different locations in user code.
This is unlikely and I shouldn't have been mentioning it.

Lets keep it simple. I think your PR will be fine if you add a size parameter to the weak function.

d-a-v (approved):
Thanks
This probably should be merged before #8546

bwjohns4:
So I've got a draft after redoing this to use class methods and move away from the weak function idea. I'll do some testing and push that here asap.

bwjohns4:
@d-a-v  I've created a class member function as shown below. It basically adds to a running `dhcpCustomOffers[]` that keeps track of all the requested options. I've added the Serial.print's just to check if this function ever gets called. I cannot see this ever printing. Am I doing something wrong in my user code to trigger this to get called? I'm doing `dhcpSoftAP.add_dhcps_custom_options(114, "This is a test");` but it doesn't seem that this my function is getting called. Any insight?

```
uint16 DhcpServer::add_dhcps_custom_options(uint8 offerCode, char *offerContent)
{
    Serial.print("OfferCode: ");
    Serial.println(String(offerCode)+offerContent);
    int sizeOfCustomOptions = strlen(dhcpCustomOffers);
    dhcpCustomOffers[sizeOfCustomOptions] = offerCode;
    dhcpCustomOffers[sizeOfCustomOptions +1] = strlen(offerContent);
    for(int i = 0; i<(strlen(offerContent)); i++){
    dhcpCustomOffers[sizeOfCustomOptions + 2 + i] = offerContent[i];
    }
    return strlen(dhcpCustomOffers);
}
```

PR #8454, opened 2022-01-18
emulation on host: minor updates

d-a-v:
- Merge MockDigital.cpp and HostWiring.cpp
  one is used for CI, the other for emulation on host
- other minor fixes


mcspr:
Formatting broke somewhere? Some things are tabs, some are 2sp, some are 4sp.
Just to avoid another extern, can't it install a certain func that either does nothing or prints things?
i.e. mock_verbose_func(nullptr) to disable log, mock_verbose_func([](const char* fmt, ...) { ... do stuff ... }); to enable

d-a-v:
> Formatting broke somewhere? Some things are tabs, some are 2sp, some are 4sp.

#8455 will fix this.

PR #8458, opened 2022-01-22
Heap init code improvements and updates

mhightower83:
Moved secondary heap init code to flash.
```
  External -24 IRAM, +32 IROM
  IRAM     -76 IRAM, +64 IROM
```

General updates to `umm_init` call path and defines to better align with upstream. Name changes: `UMM_INIT_HEAP` with `UMM_CHECK_INITIALIZED`, `umm_init_stage_2` with `_umm_init_heap`, and `umm_init_common` with `umm_init_heap`.

Add file `umm_cfgport`.h to hold port-specific values. Stay focused on heap initialization only move-related defines.

Improved comments.

Created a wrapper function for running pre-SDK code from flash.
Updated `hwdt_app_entry` to use it.

Update `umm_init` with the option to run from ICACHE.
~Added build define `UMM_INIT_USE_ICACHE` to move `umm_init` call path to flash.~
Added build define `UMM_INIT_USE_IRAM` to move `umm_init` call path back to IRAM.
Using ICACHE frees up 160 bytes of IRAM at a cost of 208 bytes of IROM.

~Defaults to no change, `umm_init` call path will be in IRAM.~
Defaults with `umm_init` call path in ICACHE.


d-a-v:
Would it be more efficient to [change default settings](https://github.com/esp8266/Arduino/pull/8458/files#diff-2c0065f95ae5033cb3f917b32b1bde87b39916cf727321aab4b54b1a497b8529R30-R32) ?

mhightower83:
@d-a-v I was being over-cautious. The experience we have with using switching cache on and off before the SDK initializes is with HWDT Stack Dump, which for me has worked fine without issue. I don't know how often it gets used. I also used it in one of the methods in my "ideas for config erase" PR. So I played it safe and left it as an option to turn on. 

Hmm, if we have a while before the next dot release and you guys feel comfortable with the change, I could switch the default around?

d-a-v:
> Hmm, if we have a while before the next dot release and you guys feel comfortable with the change, I could switch the default around?

We think it is OK to go this way !

d-a-v (approved):
Thanks!

PR #8463, opened 2022-01-25
Avoid writing to const storage

### LINKED ISSUE #8462: Avoid writing in const functions, as it can technically be UB
### Basic Infos

#### Platform

- Hardware:			        ESP-12
- Core Version:			current master
- Operating System:		MacOS

### Problem Description

This isn't really reproducible, it's more of a theoterical problem. Technically if a variable value is declared as const the compiler can assume it will never change, so changing may trigger UB. Which means doing that in some platforms causes problems, currently to the CPU we target, with the compiler we target, there is no side-effect. And there is no statical analyzer to catch it, it's a silent UB.

Because of SSO the sketch below technically is UB. As described by https://en.cppreference.com/w/cpp/language/const_cast

There also are many real life examples of programs being broken by this, like: https://stackoverflow.com/questions/8954260/const-cast-doesnt-work-c

Of course the circunstances are different, but it's not ideal to live with UB just because it works for now.

`String::substring` and `String::lastIndexOf` both do this. Ideally wbuffer() shouldn't be const to avoid this problem happening by accident. 

```cpp
#include <Arduino.h>
void setup() {
  Serial.begin(115200);
  delay(5000);
  const String a(F("Abc"));
  Serial.println(a.substring(2, 3));
  Serial.println(a);
}
void loop() {}
```

paulocsanz:
Fixes #8462 

This avoids the null termination requirement of both `String::substring` and `String::lastIndexOf` by using APIs that don't require it. So we can stop writing to the buffer inside of const functions.

There might be a off by one error there, I need to take a better look at the code and would love if someone could review.

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -682,14 +682,9 @@ int String::lastIndexOf(char ch) const {
 int String::lastIndexOf(char ch, unsigned int fromIndex) const {
     if (fromIndex >= len())
         return -1;
-    char *writeTo = wbuffer();
-    char tempchar = writeTo[fromIndex + 1]; // save the replaced character
-    writeTo[fromIndex + 1] = '\0';
-    char *temp = strrchr(writeTo, ch);
-    writeTo[fromIndex + 1] = tempchar; // restore character
-    if (temp == NULL)
-        return -1;
-    return temp - writeTo;
+    int index = fromIndex + 1;
+    while (index-- > 0 && buffer()[index] != ch);
+    return index;
 }
 
 int String::lastIndexOf(const String &s2) const {
@@ -732,11 +727,7 @@ String String::substring(unsigned int left, unsigned int right) const {
         return out;
     if (right > len())
         right = len();
-    char *writeTo = wbuffer();
-    char tempchar = writeTo[right]; // save the replaced character
-    writeTo[right] = '\0';
-    out = writeTo + left; // pointer arithmetic
-    writeTo[right] = tempchar; // restore character
+    out.concat(buffer() + left, right - left);
     return out;
 }
 
--- a/cores/esp8266/WString.h
+++ b/cores/esp8266/WString.h
@@ -263,8 +263,8 @@ class String {
         void setCapacity(int cap) { if (!isSSO()) ptr.cap = cap; }
         void setBuffer(char *buff) { if (!isSSO()) ptr.buff = buff; }
         // Buffer accessor functions
-        const char *buffer() const { return wbuffer(); }
-        char *wbuffer() const { return isSSO() ? const_cast<char *>(sso.buff) : ptr.buff; } // Writable version of buffer
+        const char *buffer() const { return isSSO() ? sso.buff : ptr.buff; }
+        char *wbuffer() { return const_cast<char *>(buffer()); } // Writable version of buffer
 
         // concatenation is done via non-member functions
         // make sure we still have access to internal methods, since we optimize based on capacity of both sides and want to manipulate internal buffers directly


earlephilhower:
Seems to have broken some host tests:

````
-------------------------------------------------------------------------------
String SSO works
-------------------------------------------------------------------------------
core/test_string.cpp:322
...............................................................................

core/test_string.cpp:420: FAILED:
  REQUIRE( s.length() == 5 )
with expansion:
  15 == 5
````

**Comment by earlephilhower** [comment]
https://github.com/esp8266/Arduino/blob/30b0450d2b0350ee0bcf58120793d21e4b414958/tests/host/core/test_string.cpp#L417-L420

**Comment by paulocsanz** [comment]
Will fix it, thanks! How can I run the host tests locally so I don't clutter the project's CI and can iterate over the code faster?

**Comment by earlephilhower** [comment]
> Will fix it, thanks! How can I run the host tests locally so I don't clutter the project's CI and can iterate over the code faster?

In `tests/host` just run `make test`.  You'll probably need valgrind and a couple other dependencies which will show up.

**Comment by paulocsanz** [comment]
Hmm, I'm getting a bunch of compilation errors because of `std::enable_if_t`, `std::is_same_v`, `std::decay_t` and `#include "../lib/littlefs/lfs.h`, but it seems I fixed the test.

**Comment by mcspr** [comment]
Were you intending to fix wbuffer() as well? It seems the const_cast originates b/c of the const qualifier(s), where the compiler did warn about the const misnomer but it was circumvented

File: cores/esp8266/WString.cpp [hunk 1]
```diff
@@ -682,14 +682,13 @@ int String::lastIndexOf(char ch) const {
...
-    char *temp = strrchr(writeTo, ch);
-    writeTo[fromIndex + 1] = tempchar; // restore character
-    if (temp == NULL)
-        return -1;
-    return temp - writeTo;
+    for (unsigned int index = fromIndex; index > 0; --index) {
+        if (buffer()[index] == ch)
+            return index;
+    }
+    if (buffer()[0] == ch)
```
mcspr, line 689: note that the loop could include this condition as well (...also seen in std::string::rfind...)
```cpp
#include <array>
#include <iostream>

const char x[] {
    'a', 'b', 'c', 'd', 'e'
};

int main() {
    auto size = std::size(x);
    std::cout << "size is " << size << '\n';
    if (size) {
        --size;
        for (++size; size--;) {
            std::cout << x[size] << ':' << size << '\n';
        }
    }
}
```
```
size is 5
e:4
d:3
c:2
b:1
a:0
```

d-a-v: :+1: then the not-found case can also be integrated in the loop
```cpp
    int index = fromIndex + 1;
    while (--index >= 0 && buffer()[index] != ch);
    return index;
```

paulocsanz:
@d-a-v Done, your trick worked, just had to change a tiny detail!

I also changed wbuffer to make it non const.

d-a-v (approved):
LGTM

cores/esp8266/WString.h
```diff
@@ -263,8 +263,8 @@ class String {
         void setCapacity(int cap) { if (!isSSO()) ptr.cap = cap; }
         void setBuffer(char *buff) { if (!isSSO()) ptr.buff = buff; }
         // Buffer accessor functions
-        const char *buffer() const { return wbuffer(); }
-        char *wbuffer() const { return isSSO() ? const_cast<char *>(sso.buff) : ptr.buff; } // Writable version of buffer
+        const char *buffer() const { return isSSO() ? sso.buff : ptr.buff; }
+        char *wbuffer() { return const_cast<char *>(buffer()); } // Writable version of buffer
```
mcspr, line 267: just c/p? I though the idea was to also remove the const_cast, not just fix the const-correctness of the class itself

core guidelines also suggest the same approach https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es50-dont-cast-away-const
(and since we can't have a specific member that will hold the pointer)

paulocsanz: Oh ok, but the const correctness is still there, since wbuffer is a non const function it ensures the internal buffer actually is writable, so it just avoids copy paste without any risk. But yes I can copy and paste.

paulocsanz: @mcspr should I copy and paste or are you ok with my reasoning?

mcspr: > The problem wasn't the const_cast, it was that wbuffer was a const function so other const functions were using it to write to String, which could be in const storage. Now if wbuffer is called, String is ensured to not be in const storage.

I concur.

Re-reading the paragraph on the order, it's the correct thing to do.
But don't really follow the reason behind the explicit `static_cast<const Foo&>(*this).get_bar()` (ref. core guidelines) to call the method instead of just get_bar()

mcspr (approved):

PR #8464, opened 2022-01-25
Astyle to clang format

mcspr:
replaces / amends #8455 
- temporary .clang-format replaced with a -style arg with yaml -> yaml conversion replacing 'blocks' with 'flow' (i.e. removing indentation and wrapping things instead, { for objects, [ for lists)
- clean-up restyle.sh to use git for relpath, verbose mode to match what astyle & bash echo did. plus, shows the actual command with an additional '-x' option

there does not seem to be an option for one-line brace addition for if() { ... } while() { ... } etc. :/
clang-tidy might help with that, but it needs a [build command line info via compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html) so it can handle macros, build options, know the language context etc. For target build of the Core, arduino-cli or platformio can already do that, but need some kind of project file and there might be problems with dependencies not being picked up. Host files need something else, either manually create the .json or use some [makefile generator like cmake to build mock files and retrieve build cmd that way](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).


mcspr:
would a different author help filtering when blame / log-search is happening?
order or PRs wouldn't really matter, I guess, might commit this first in the queue

d-a-v:
Indeed FWIW and given the huge umber of changes, in the past and in my previous attempt I used the separate "allman" account in order to get a better commit history.

PR #8465, opened 2022-01-27
Heap panic / abort cleanup

mhightower83:
Isolate NULL/panic test of `_context` to dev debug assert macro.

Use abort instead of panic for the case of the caller providing a non-heap address pointer.
*  Added debug print.
* Improved `get_unpoisoned_check_neighbors` to print file/line when available.


d-a-v (approved):
Nothing jumps at me in this code simplification.
Thank you for keeping improving the undergrounds and UMM

mcspr:
Should be ok to merge then?

PR #8467, opened 2022-01-28
Fix WiFiEventHandlerOpaque, now implemented correctly in header and i…

EliaTomaselli:
I was trying out the library and noticed that when calling

`WiFiEventHandler handler = WiFi.onStationModeConnected(
        [](WiFiEventStationModeConnected e) { ... });`

the variable "handler" (which is typedef'd as a std::shared_ptr<WiFiEventHandlerOpaque>) was basically a shared pointer with an empty struct inside, since WiFiEventHandlerOpaque in ESP8266WiFiGeneric.h was just declared like this.

`struct WiFiEventHandlerOpaque;`

This leads to a "pointer to incomplete class type is not allowed" compile time error, because the interface is missing and no method or member variable can be accessed from "handler".

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
@@ -53,29 +53,23 @@ extern "C" {
 // ------------------------------------------------- Generic WiFi function -----------------------------------------------
 // -----------------------------------------------------------------------------------------------------------------------
 
-struct WiFiEventHandlerOpaque
+
+WiFiEventHandlerOpaque::WiFiEventHandlerOpaque(WiFiEvent_t event, std::function<void(System_Event_t*)> handler)
+: mEvent(event), mHandler(handler)
 {
-    WiFiEventHandlerOpaque(WiFiEvent_t event, std::function<void(System_Event_t*)> handler)
-    : mEvent(event), mHandler(handler)
-    {
-    }
+}
 
-    void operator()(System_Event_t* e)
-    {
-        if (static_cast<WiFiEvent>(e->event) == mEvent || mEvent == WIFI_EVENT_ANY) {
-            mHandler(e);
-        }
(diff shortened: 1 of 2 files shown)


mcspr:
What is the exact error message you are seeing, and what is the example code that triggers the error?
We are allowed to return the pointer, but yes we can't access it's members from user code as it is only available from inside of ESP8266WiFiGeneric.cpp

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -53,29 +53,23 @@ extern "C" {
...
+    if (static_cast<WiFiEvent>(e->event) == mEvent || mEvent == WIFI_EVENT_ANY) {
+        mHandler(e);
     }
+}
 
-    WiFiEvent_t mEvent;
-    std::function<void(System_Event_t*)> mHandler;
-    bool mCanExpire = true; /* stopgap solution to handle deprecated void onEvent(cb, evt) case */
-};
+bool WiFiEventHandlerOpaque::canExpire()
```
TD-er, line 69: While you're at it, this function should be marked `const`

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.h
```diff
@@ -37,7 +37,15 @@
...
 
-struct WiFiEventHandlerOpaque;
+struct WiFiEventHandlerOpaque {
+    WiFiEventHandlerOpaque(WiFiEvent_t event, std::function<void(System_Event_t*)> handler);
+    void operator()(System_Event_t* e);    
+    bool canExpire();
+
+    WiFiEvent_t mEvent;
+    std::function<void(System_Event_t*)> mHandler;
+    bool mCanExpire = true; /* stopgap solution to handle deprecated void onEvent(cb, evt) case */
```
TD-er, line 47: Just curious, why is there a `canExpire()` function to access a member which is public?

mcspr:
*ping*

PR #8472, opened 2022-01-31
Permit using the Updater _hash function, even if we don't have a signature appended to the image

einglis:
The _hash and _verify functionality of the Updater class are pretty much entwined.  But it might be useful to calculate the hash, even without a signature present in the image itself.  This change permits that by allowing a zero-length signature.

For reference, one possible use case is where the expected hash is provided separately to the uploaded image.  Another is to provide generic post-upload validation of the image: the hash function permits a convenient way of inspecting the complete image against arbitrary conditions; this is particularly useful after HTTP client OTA updates. 

(It's fair that reading the whole image back out of flash is not very efficient, but that's not the concern of this PR.)


cores/esp8266/Updater.cpp
```diff
@@ -255,20 +259,24 @@ bool UpdaterClass::end(bool evenIfRemaining){
...
 #endif
-    uint8_t *sig = (uint8_t*)malloc(sigLen);
-    if (!sig) {
-      _setError(UPDATE_ERROR_SIGN);
-      _reset();
-      return false;
-    }
-    ESP.flashRead(_startAddress + binSize, sig, sigLen);
+
+    uint8_t *sig = 0; // Safe to free if we don't actually malloc
```
earlephilhower, line 263: Could we make this `nullptr` and not `0`,  just for consistency?

einglis:
Hold on.  This line:

```int binSize = _size - sigLen - sizeof(uint32_t) /* The siglen word */;```

still ends up subtracting an extra uint32_t from the total image length, so the hash function doesn't get to see all the data.  That looks like double-accounting of sigLen to me, but I'll look closely and revise my PR accordingly.

Sorry I didn't catch this sooner.

PR #8478, opened 2022-02-04
Fix compilation of ArduinoOTA when no global MDNS is available

drzony:
Fix for https://github.com/esp8266/Arduino/pull/8379


d-a-v (approved):
Thanks !

Issue #8481, opened 2022-02-08
umm_calloc undetected multiply overflow/wrap


mhightower83:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[3027acaf11db93e877fca943eb815216d0507744]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Adafruit Feather HUZZAH ESP8266]
- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [460800] (serial upload only)

### Problem Description

umm_calloc(16777217u, 256u) should fail because the buffer requested is too large; however, it succeeds and returns a buffer pointer. The problem is caused by an undetected overflow/wrap using unsigned multiply.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

void setup() {
  Serial.begin(115200);
  delay(200);

  Serial.printf("\r\nCalloc Multiply Overflow Test\r\n");

  size_t a = 16777217u, b = 256u;
  void *p = calloc(a, b);
  if (p) {
    Serial.printf("  Failed: calloc(%u, %u) returned pointer: %p - Unexpected success on failure case\r\n", a, b, p);
    free(p);
  } else {
    Serial.printf("  Passed: calloc(%u, %u); - Expected failure NULL pointer returned.\r\n", a, b);
  }
}

void loop() {}

```

### Debug Messages

```
Calloc Multiply Overflow Test
  Failed: calloc(16777217, 256) returned pointer: 0x3ffef3bc - Unexpected success on failure case
```

PR #8482, opened 2022-02-08
Fix, calloc now fails on extra-large request.

mhightower83:
Fixes https://github.com/esp8266/Arduino/issues/8481#issue-1126658395

umm_calloc, now fails on extra-large request.

Added code to handle multiply overflow in umm_calloc.
Added code to handle add overflow in umm_poison_*


mhightower83:
@d-a-v  I have one more push coming. I missed the calloc entry that libc uses. I am not sure if this is just overkill.

PR #8484, opened 2022-02-10
Moved EEPROM Library functions from header to CPP file

AddioElectronics:
EEPROM library contained function's in the header, which when including from multiple files gives the "redefinition of class" error.
Functions were moved from header to CPP file so that you are able to include EEPROM.h from multiple files.


mcspr:
Functions in the class body *are* inline, we don't have to do what you are proposing.

You sure that is the reason you are seeing this issue?
>  which when including from multiple files gives the "redefinition of class" error.
```
$ ~/d/arduino8484> cat src/one.cpp
#include <Arduino.h>
#include <EEPROM.h>

void two();

void setup() {
    EEPROM.begin(512);
    EEPROM.put(0, "4321");
    two();
}

void loop() {
    delay(1000);
}
~/d/arduino8484> cat src/two.cpp
#include <Arduino.h>
#include <EEPROM.h>

void two() {
    EEPROM.put(0, "1234");
}
$ pio project init -s -b d1_mini
$ pio run -s ; echo $status
0
```
Is successfully building with PlatformIO, at the very least

AddioElectronics:
Ok I found out the real issue.
I had https://github.com/jwrw/ESP_EEPROM installed which was colliding with the standard library.
Even though I didn't #include <ESP_EEPROM.h>, just having it in my libraries folder was causing it.
Just tested it again, and it doesn't have that problem in regular Arduino IDE, only in Visual Studio/Visual Micro.

I guess I must have removed the library at the same time that I moved the functions to CPP without realizing, and attributed the "fix" to the moving of the functions.
Being more familiar with C, if you define a function or variable that is not static, in a header that is included more than once, it gives a redefinition error, which is why I thought it was that.

Anyways, my apologies for wasting your time. Closed.

PR #8486, opened 2022-02-10
Free String memory on assign of "" or clear()

### LINKED ISSUE #8485: String clear() does not free memory, only set length(0)
### Basic Infos

#### Platform

### Problem Description

The String::clear() function does only set the length to 0, it does not free the memory.
The same for assigning an empty string, as this only checks if the reserved capacity is enough.

This means it is quite tricky when clearing a string that's no longer needed.
For example a buffer I'm using, which is an array of String.

The only way to really free the memory is by assigning a new string which will then be copied using the move operator, like this:

```c++
Message[read_idx] = String();
```

It does not really make sense to only set the length of the string to 0, as that's not what `clear()` suggests.

earlephilhower:
Fixes #8485

Whenever the empty C string is assigned or the .clear() method
called, free any heap allocated with the String.

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -258,7 +258,7 @@ String &String::operator =(String &&rval) noexcept {
 }
 
 String &String::operator =(const char *cstr) {
-    if (cstr)
+    if (cstr && cstr[0])
         copy(cstr, strlen(cstr));
     else
         invalidate();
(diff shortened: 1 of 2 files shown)


erkr (commented):
I consider this not desirable. Use case  for reserve() is to avoid memory fragmentation. To reuse such a String , cleaning without freeing the menus is required.
Strings can be freed this way already:
``` 
String reserved=String():
```

Issue #8475, opened 2022-02-02
All Wemos D1 mini code stopped working after esp8266 Boards upgrade from 2.5.2 to 3.0.2


TSBrownie:
### Basic Info
- [x ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ?] I have tested that the issue is present in current master branch (aka latest git).
- [ x] I have searched the issue tracker for a similar issue.
- [ x] If there is a stack dump, I have decoded it.
- [ x] I have filled out all fields below.

#### Platform
- Hardware:			Wemos D1 mini  (reports: ESP8266EX) w/ RTC Shield/SD
- Core Version:			[latest git hash or date]
- Development Env:		1.8.16
- Operating System:		Win 10

### Settings in IDE

- Module:  			Wemos D1 mini ESP8266 3.0.2 (from 2.5.2)
- Flash Mode:			qio
- Flash Size: 		        4MB
- lwip Variant:                   v2 lower mem 
- Reset Method:		nodemcu, hard, power
- Flash Frequency:	        ?? Mhz
- CPU Frequency:		Crystal is 26MHz
- Upload Using:		SERIAL
- Upload Speed:               460800 (auto set)

### Problem Description
Working code (plus 10 past versions) stopped working immediately after updating from Boards esp8266 by ESP8266 Community version 2.5.2 to 3.0.2.
Code still complies w/o error.  (Takes take 3x longer.) 
I have tried all the Wemos LOLIN boards on the 3.0.2, include generic esp8266. Similar chksum errors from all.

Output before:
========= 20220201,Tue,07:54:28
Device Altitude(abs): 6.00 meters AMSL
Temperature: 28.47C; 83.25F  Falling
Barometric Pressure(abs): Rising-Fair Weather
  955.597 hPa(mbar); 95559.720 Pa
  950.916 hPa, Short Term Past Average 
  716.757 mmHg
  0.943101 Atm 
Barometric Pressure-Comp'ed To Sealevel
  956.247 hPa(mbar); 95624.692 Pa
  717.244 mmHg
  0.943742 Atm 

Output after:
 ets Jan  8 2013,rst cause:4, boot mode:(3,6)
wdt reset
load 0x4010f000, len 3460, room 16 
tail 4
chksum 0xcc
[...]

Issue #8421, opened 2021-12-24
Problem with int function which doesn't return a value 


thierry7100:
----------------------------- Delete below -----------------------------

If your issue is a general question, starts similar to "How do I..", is related to 3rd party libs, or is related to hardware, please discuss at a community forum like esp8266.com.

INSTRUCTIONS
============
If you do not follow these instructions, your issue may be dismissed.

1. Follow the checklist under Basic Infos and fill in the [ ] spaces with an X.
2. Fill in all the fields under Platform and Settings in IDE marked with [ ] (pick the correct option for you in each case, delete the others).
3. If you haven't already done so, test your issue against current master branch (aka latest git), because it may have been already fixed.
4. Describe your problem.
5. If you have a STACK DUMP decode it:

https://arduino-esp8266.readthedocs.io/en/latest/Troubleshooting/stack_dump.html

6. Include a Minimal Complete Reproducible Example sketch that shows your issue. Do not include your entire project, or a huge piece of code.
7. Include debug messages:

https://arduino-esp8266.readthedocs.io/en/latest/Troubleshooting/debugging.html 

8. Use markup (buttons above) and the Preview tab to check what the issue will look like.
9. Delete these instructions from the above to the below marker lines before submitting this issue.

----------------------------- Delete above -----------------------------

### Basic Infos

- [ X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ ]X I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ X] I have tested that the issue is present in current master branch (aka latest git).
- [ X] I have searched the issue tracker for a similar issue.
- [ X] If there is a stack dump, I have decoded it.
- [ X] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12]
- Core Version:			[3.0.2]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Wemos D1 mini r2]
- Flash Mode:			[?]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

[...]

PR #8492, opened 2022-02-20
Docs: add an entry about compiler warnings

mcspr:
per https://github.com/esp8266/Arduino/issues/8421 and https://github.com/esp8266/Arduino/issues/8475
after checking HW, suggest to check SW as well at least by enabling compiler warnings
(and also note of the IDE weirdest defaults causing issues we expected to stay solved)

### DIFF
--- a/doc/faq/a02-my-esp-crashes.rst
+++ b/doc/faq/a02-my-esp-crashes.rst
@@ -79,9 +79,8 @@ Enable compilation warnings
 
 Most common issues may be resolved by enabling compilation warnings and fixing them.
 
-For Arduino IDE:
+For Arduino IDE, select ``File -> Preferences``:
 
-- Select ``File -> Preferences``
 - Make sure ``Show verbose output during: compilation`` is enabled
 - Set ``Compiler warnings`` to ``More`` or ``All``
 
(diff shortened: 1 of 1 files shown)


d-a-v (approved):
This was missing.

earlephilhower:
Don't we *always* generate a warning/error on this case, even with `None`?  If not, I think we can fix it instead of documenting it, by changing the `platform.txt` options...

mcspr:
only 3 out of 4 options produce error:
https://github.com/esp8266/Arduino/blob/15e7d35d6e7c430915a5a86d854ae855d10a8da5/platform.txt#L24-L28

'None' never does, `-w` overrides any `-W` flags since they apply all at once and not like 'disable all warnings *but* still enable this one', order does not matter. Removing `-w` from 'None' kind of defeats the purpose of the option, and there is 'Default' for that.

We do still need to mention the warnings, which even in 'Default' will give some useful advice. Specifically with C++, shouldn't we mention that it is undefined behaviour when trying to code like that?

btw low-level routines that really do not return may still be coded like this
```cpp
int func() {
  __builtin_unreachable();
}
```

earlephilhower:
Argh, you are correct.  Even trying to explicitly enable a warning like `-w -Wreturn-type -Werror=return-type` doesn't work.

It seems the only way to make this automatically catch is to not use `-w` and specifically list all ~100 `-Wno-XXX` in an `@option` file.  While doable, that's more involved than I have time for this AM...

earlephilhower (approved):

Issue #8483, opened 2022-02-09
Build size on Linux significantly larger than on Windows


TD-er:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-12|ESP-01|ESP-07|ESP8285 device|other]
- Core Version:			[Any core version since at least 2.6.x]
- Development Env:		[Platformio]
- Operating System:		[Windows|Ubuntu|MacOS]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[dio]
- Flash Size: 		[4MB/1MB]
- lwip Variant:   [v2 Lower Memory|Higher Bandwidth]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz|160MHz]
- Upload Using:		[OTA|SERIAL]
- Upload Speed:   [115200] (serial upload only)

### Problem Description

For ESPEasy I do develop on Windows.
The builds are made using GitHub Actions, which is running on Linux.
I also tested it locally on Linux and a user tested it on MacOS too.
The builds made on Windows are significantly smaller (order of 35k - 40k) compared to those made on Linux/MacOS.

I do use the same PlatformIO config on both.
However, I do not know whether the toolchain on both may be different.

I did try a lot of extra compiler flags to get the build size down as it is an ever ongoing battle to get as much as possible in the small max. sketch size of an ESP8266.
For example `-Os` to optimize for size and `-s` to force extra effort to strip unused functions.
However, this does not seem to have any effect on Windows builds and hardly noticable effect (if any) on Linux/MacOS.

Not sure if this is a problem for this repository, or PlatformIO or even Gcc, but I have to start somewhere.

A few thoughts of what might be the issue here:
- Ignoring the flag to allow for flash string alignments other than 4 (or using even more?)
- Not performing Flash string duplicate checks?
- Using different alignments between structs?
- Using different alignments for switch statements using "smaller" enums like an enum class of size char or switch statements over a char?
[...]

PR #8493, opened 2022-02-20
Correctly using fs:: namespace in SD & SDFS

mcspr:
Remove `using namespace fs;` from SDFS.h
Fix everything that depended on it

As noticed in the #8483


d-a-v (approved):
Looks good !

PR #8504, opened 2022-03-06
Add support for global build defines and options

mhightower83:
A script that manages the use of a file with a unique name, like `Sketch.ino.globals.h`, in the sketch source directory, to provide compiler command-line options (build options) and sketch global `#define` macros. The build option data is encapsulated in a unique "C" comment block and extracted into the build tree during prebuild.


d-a-v (approved):
Thanks a lot for this proposal.
It is an excellent replacement to a feature-request which is still not available upstream.

d-a-v:
Please ask them directly (arduino/esp32).

Best is to ask Arduino themselves. There are lots en entry-points about this subject.
I will not flood, so I'll give only one: https://github.com/arduino/arduino-cli/issues/846 . Please note the "criticality: high" tag they added recently.

... 

Well, I will add [this one](https://github.com/arduino/tooling-rfcs/pull/14) because they still have not closed it yet despite the fact that they should have after months of inactivity.


PS: @mhightower83 Thanks for this addition

PR #8507, opened 2022-03-11
Permit using the Updater _hash function, even if we don't have a signature appended to the image

einglis:
NOTE: Re-submission of PR #8472 due to user error :(

The _hash and _verify functionality of the Updater class are pretty much entwined. But it might be useful to calculate the hash, even without a signature present in the image itself. This change permits that by allowing a zero-length signature.

For reference, one possible use case is where the expected hash is provided separately to the uploaded image. Another is to provide generic post-upload validation of the image: the hash function permits a convenient way of inspecting the complete image against arbitrary conditions; this is particularly useful after HTTP client OTA updates.

(It's fair that reading the whole image back out of flash is not very efficient, but that's not the concern of this PR.)

### DIFF
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -223,19 +223,29 @@ bool UpdaterClass::end(bool evenIfRemaining){
     _size = progress();
   }
 
-  uint32_t sigLen = 0;
   if (_verify) {
-    ESP.flashRead(_startAddress + _size - sizeof(uint32_t), &sigLen, sizeof(uint32_t));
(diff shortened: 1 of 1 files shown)


d-a-v (approved):
LGTM

einglis:
Ty

einglis:
Good spot, @blue-wind-25.  I think you're right.

blue-wind-25:
You are welcome :-)

blue-wind-25:
Great, thank you for the confirmation :-)

Issue #8510, opened 2022-03-15
ArduinoOTA retrieves incorrect IP for TCP transfer


krisrok:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:				ESP-8266EX
- Core Version:			3.30002.0
- Development Env:		Platformio
- Operating System:		Windows

### Settings in IDE

- Module:  			Wemos D1 mini

### Problem Description

When trying to use OTA updates via internet [this call](https://github.com/esp8266/Arduino/blob/3027acaf11db93e877fca943eb815216d0507744/libraries/ArduinoOTA/ArduinoOTA.cpp#L178) retrieves a wrong, local IP.
The UDP communication works just fine but because of the wrong IP the TCP connection (pushing the actual firmware/payload) fails:

```
10:57:53 [INFO]: Sending invitation to: x.x.x.x
10:58:03 [ERROR]: No Answer
```

Network setup it is: Internet -> Router A -> Router B (in Acces Point mode) -> ESP. The call above yields the local IP address of Router A instead of the correct remote one.

I thought I could use the `--host_ip` argument to set the correct IP but it tries to bind the socket to it which fails. And the cpp side does not use this supplied address, it relies on the remote address of the UDP connection.

So I've added some code to `espota.py` to NOT bind to the to `localAddr` but to [`INADDR_ANY`](https://docs.python.org/3/library/socket.html#socket-families)
```python
server_address = ("", localPort)
```
and append the correct IP localAddr with the message:
```python
message = '%d %d %d %s\n%s\n' % (command, localPort, content_size, file_md5, localAddr)
```

and to read it on the cpp side:
```cpp
String ipstring = readStringUntil('\n');
_ota_ip2.fromString(ipstring);
```

and then I use this for the TCP connection:
```cpp
if (!client.connect(_ota_ip2, _ota_port)) {
```

With this now the OTA update works when explicitly setting the IP, e.g. like this:
```
[...]

PR #8512, opened 2022-03-16
Ota explicit host ip

krisrok:
This PR is a cleaned up version of what I outlined in #8510. 

The python script explicitly transfers the host's IP to the client because the UdpContext did not reliably return the correct remote address.

I changed the meaning of the `host_ip` argument. Previously this was the IP the host's TCP socket binds to. Now it always binds to [`INADDR_ANY`](https://docs.python.org/3/library/socket.html#socket-families) because why not -- I'm no expert in networking regards so this change has to be discussed.
Instead of chaning the existing argument we could also introduce a new `host_ip_remote` or 

Other than the `host_ip` it should be fully backwards compatible so you can use the new python script with the old cpp implementation and vice versa.

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -170,6 +170,7 @@ String ArduinoOTAClass::readStringUntil(char end){
 void ArduinoOTAClass::_onRx(){
   if(!_udp_ota->next()) return;
   IPAddress ota_ip;
+  IPAddress ota_ip2;
 
   if (_state == OTA_IDLE) {
     int cmd = parseInt();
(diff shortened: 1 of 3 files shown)


tools/espota.py
```diff
@@ -94,8 +94,8 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
   file_md5 = hashlib.md5(f.read()).hexdigest()
   f.close()
   logging.info('Upload size: %d', content_size)
-  message = '%d %d %d %s\n' % (command, localPort, content_size, file_md5)
 
+  message = '%d %d %d %s\n%s\n' % (command, localPort, content_size, file_md5, localAddr)
```
earlephilhower, line 98: I am worried that this will break existing deployments since the wire protocol is now different.  The existing code will not read the extra IP address sent afterwards, but instead interpret it as part of the actual update and fail.

Can someone confirm that a build of an app w/the main branch code works fine with this new updater (i.e. an update of a deployed system).?

krisrok: It’s backwards compatible because the (old and new) client luckily only reads the md5 hash until encountering `\n`.

So the newly added ip adress would sit in the udp message buffer but that‘s it. 

The update itself happens via a secondary tcp connection and is not affected.

libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -170,6 +170,7 @@ String ArduinoOTAClass::readStringUntil(char end){
 void ArduinoOTAClass::_onRx(){
   if(!_udp_ota->next()) return;
   IPAddress ota_ip;
+  IPAddress ota_ip2;
```
d-a-v, line 173: I kindly ask whether it possible, for readbility, to rename this to `ota_ip_provided_by_cmdline` or `ota_ip_finally_retained` or anything describing what it is and different from "name2".

tools/espota.py
d-a-v: Thanks for this clarification :+1:

libraries/ArduinoOTA/ArduinoOTA.cpp
d-a-v, line 173: I kindly ask whether it possible, for readbility, to rename this `to ota_ip_provided_by_cmdline` or `ota_ip_finally_retained` or anything describing what it is and different from "name2".

d-a-v (changes requested):

krisrok:
It is a workaround around the local network limitation, yes :)

If there are other, more viable ways to enable OTA via web, that‘d be great of course.
I just needed the functionality at the time, was able to hack it together and thought I‘d share.

I did not see a real use case for the `host-ip` argument. Maybe there are some, I don‘t know.

PR #8526, opened 2022-03-30
WString: unify numeric conversion and fix assignments

### LINKED ISSUE #8430: String: API must be fixed
Per [this comment](https://github.com/esp8266/Arduino/issues/8404#issuecomment-1003521423),
`String` API has changed since release 2.7.4 and should be fixed.

(cc: @Ciapas-Linux)

mcspr:
Restore the pre-3.0.0 behaviour when we could assign numeric values to
the string object. After introducing `operator =(char)`, everything was
converted to `char` instead of the expected 'stringification' of the
number (built-in int, long, unsigned int, unsigned long, long long,
unsigned long long, float and double)

Add `toString()` that handles conversions, re-use it through out the class

Fix #8430


mcspr:
I suppose there should be tests for assignment... Since there are for constructors.

Issue #8082, opened 2021-05-28
Panic on WPA2 Enterprise Connection


Flole998:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:				ESP-01
- Core Version:			3.0.0
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			dio
- Flash Size: 		1M
- lwip Variant:   v2 Lower Memory
- Reset Method:		ck
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

Immediately after connecting to a WPA2 Enterprise encrypted network I receive an exception. This seems to be related to the way free is called from a function in eap.c (which seems to be part of the SDK). I am not sure if this is a bug here or in the SDK, but apparently an address that is unmapped was passed to free. I am not sure if maybe the stack got corrupted somewhere as that function hierarchy doesn't really make sense to me.....

I have everything working on the super old version 2.3.0 (which uses completely different functions for setting WPA Enterprise up) but I want to update.

I have reproduced this on all Espressif Firmware Versions that are available.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

void setup() {

	Serial.begin(115200);
	Serial.println("Startup!");

	Serial.print("Heap Free: ");
	Serial.println(system_get_free_heap_size());

	enableWiFiAtBootTime();

	wifi_set_opmode_current(STATION_MODE);

	struct station_config wifi_config;

	memset(&wifi_config, 0, sizeof(wifi_config));
	strcpy((char*)wifi_config.ssid, ssid);
	wifi_station_set_config_current(&wifi_config);

	wifi_station_set_wpa2_enterprise_auth(1);

	wifi_station_set_enterprise_identity((uint8_t*)identity, strlen(identity));
[...]

PR #8529, opened 2022-04-04
Fix double-free when connecting to WPA2-Enterprise networks

Flole998:
Fixes: #8082

This patches the callx0 instruction to a nop in eap.o which is part of libwpa2.a.
It looks like espressif fixed the Bug in newer SDK versions, so if we update to the latest NONOS-SDK it is most likely not necessary to add/adapt this patch.
Also modifies the fix_sdk_libs.sh script as it even changed files if no changes were necessary, for example adding multiple system_func1 exports.


tools/sdk/lib/fix_sdk_libs.sh
```diff
@@ -1,36 +1,67 @@
...
+}
 
 # Remove mem_manager.o from libmain.a to use custom heap implementation,
 # and time.o to fix redefinition of time-related functions:
 xtensa-lx106-elf-ar d libmain.a mem_manager.o
 xtensa-lx106-elf-ar d libmain.a time.o
 
+# Patch WPA2-Enterprise double-free
+xtensa-lx106-elf-ar x libwpa2.a eap.o
+espcs=$(sha256sum eap.o | awk '{print $1}')
```
d-a-v, line 36: Should it be `eapcs=` ?

Flole998: Sorry, not entirely sure how that could happen. Of course... Fixed....

mcspr (commented):
what's the status here btw? everything was tested and works after these?

tools/sdk/lib/fix_sdk_libs.sh
```diff
@@ -1,36 +1,67 @@
 #!/bin/bash
 set -e
 
-export PATH=../../xtensa-lx106-elf/bin:$PATH
+export PATH=../../../xtensa-lx106-elf/bin:$PATH
```
mcspr, line 4: this was correct though? from the tools/sdk/lib, we expect get.py to put toolchain there
```
> realpath -m ../../xtensa-lx106-elf/bin
/home/builder/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/tools/xtensa-lx106-elf/bin
```

Flole998:
It does work, but there is still a memory leak upon reconnection (which was there before aswell, so no regression). That is not easily fixable though (I am not sure it can be fixed at all at the moment). If you can live with that leak then this is perfectly usable.

tools/sdk/lib/fix_sdk_libs.sh
Flole998: This script must be run from one of the subdirectories though as it does patch the contents in the subdirectory. So how I used it (and how I think it should be used) is like this:

```
cd tools/sdk/lib/
cd NONOSDK221
../fix_sdk_libs.sh
```

d-a-v (approved):
Thanks !
Next step is to add a basic example in the WiFi example directory
and maybe a WiFi API around the espressif API.

tools/sdk/lib/fix_sdk_libs.sh
```diff
@@ -1,36 +1,67 @@
...
 VERSION=$(basename ${PWD})
 
 addSymbol_system_func1() {
-	ADDRESS=$1
-	xtensa-lx106-elf-objcopy --add-symbol system_func1=.irom0.text:${ADDRESS},function,global user_interface.o
+	if ! xtensa-lx106-elf-nm  user_interface.o | grep -q " T system_func1"; then # Don't add symbol if it already exists
```
mhightower83, line 8: Sorry, I just noticed that `grep -q` will always return 0.

Flole998: Even if it's not in there yet? That's strange (be aware that it got already added twice or so...)

mhightower83: Yes, that's what I noticed when I tried some modifications.  grep -q appears to be an exception on returning non-zero exit status.

Flole998: For me it works:
```
$ echo ThisIsATest | grep -q Test
$ echo ?$
0
$ echo ThisIsATest | grep -q Fail
$ echo $?
1
```

mhightower83: Well, I am confused. It wasn't working for me and now it is. I had duplicate entries and the man page says `grep -q` exits with 0. However, I just tried everything again and it worked. 

Sorry, I don't know what happened. Just ignore me, I am going to take a break.

Flole998: The current version as it is in the repo at the moment is having duplicate entries, those should be removed at some point. So maybe that caused the confusion (and that's why I added that check)

mhightower83: Hmm, strange. I only picked up the script and tried it and had duplicate entries. As well as other problems from running the script at the wrong directory which creates a `libmain.a` which creates a lot of grief.

d-a-v: @Flole998 what do you mean "duplicate entries in the current version" ?
Is it in master or in this pull request ?

Flole998: In master (and also in this pull request because I didn't remove/fix it as that would be something for another PR I figured)

PR #8535, opened 2022-04-09
Delete operator=(Self&) when copy constructor is deleted

### LINKED ISSUE #8534: Some classes delete the copy constructor, as it's unsound, but allow operator=(Self&)
### Basic Infos

#### Platform
- Core Version:			684156d211b21af6e0303275b9ceccc06f132f8f
- Operating System:		Windows-WSL2

### Problem Description

Seeing this issue/comment [#8522](https://github.com/esp8266/Arduino/issues/8522#issuecomment-1086440477) inspired me to check for other places where we were forgetting to delete some constructor/assignment. And it turned out I found a few places with problems that would lead to invalid logic.

This should compile, but it doesn't

```cpp
#include <WiFiClientSecureBearSSL.h>

void setup() {
  static_assert(!std::is_copy_constructible<BearSSL::WiFiClientSecureCtx>());
  static_assert(!std::is_copy_assignable<BearSSL::WiFiClientSecureCtx>());
  static_assert(!std::is_copy_constructible<BearSSL::PublicKey>());
  static_assert(!std::is_copy_assignable<BearSSL::PublicKey>());
  static_assert(!std::is_copy_constructible<BearSSL::PrivateKey>());
  static_assert(!std::is_copy_assignable<BearSSL::PrivateKey>());
  static_assert(!std::is_copy_constructible<BearSSL::X509List>());
  static_assert(!std::is_copy_assignable<BearSSL::X509List>());
}
void loop() {}

```

paulocsanz:
Fixes #8534

This brings two questions tho, should we break backwards compatibility with a usage that is literally broken, so nobody should be actually depending on this? As these changes technically make some broken code to stop compiling. I'm out of the loop on how the community decides on situations like this.

And what is with the fake `WiFiClientSecureCtx::clone`? With my changes it actually always fails to compile, so I moved to causing a runtime failure that will be more explicit. What do you all think? I am open to suggestions.

I see that there is a different PR tackling a similar issue (targeting the original issue), and it might expand the functionalities, but I see this as orthogonal, if we decide to support copy operations in some of those classes in the future we can remove the two deletes, but it doesn't seem a decided issue.

### DIFF
--- a/libraries/ESP8266WiFi/src/BearSSLHelpers.h
+++ b/libraries/ESP8266WiFi/src/BearSSLHelpers.h
@@ -58,6 +58,7 @@ class PublicKey {
 
     // Disable the copy constructor, we're pointer based
     PublicKey(const PublicKey& that) = delete;
+    PublicKey& operator=(const PublicKey& that) = delete;
 
   private:
     brssl::public_key *_key;
@@ -86,6 +87,7 @@ class PrivateKey {
 
     // Disable the copy constructor, we're pointer based
     PrivateKey(const PrivateKey& that) = delete;
+    PrivateKey& operator=(const PrivateKey& that) = delete;
 
   private:
     brssl::private_key *_key;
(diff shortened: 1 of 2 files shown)


d-a-v (approved):
Thanks

`WiFiClientSecureCtx::clone()` is indeed supposed to be never called

PR #8545, opened 2022-04-19
Fix Updater non-zero _verify->length() once again

mcspr:
As mentioned in the https://github.com/esp8266/Arduino/pull/8507#issuecomment-1096565166
I took the liberty to also do some refactoring; specifically, fixing signed vs. unsigned mismatch in len, using pointer object vs. the original manual malloc & free, try to have named constants for certain addresses and lengths, plus localize printing of u8 arrays.

The suggested test to have a 'dummy' verifier works just fine. (...how it actually works and gets the hash to compare with is a whole other question, though)

Another issue noticed while testing, in the underlying bearssl api there's an actual limit for hash length.
https://github.com/earlephilhower/bearssl-esp8266/blob/6105635531027f5b298aa656d44be2289b2d434f/inc/bearssl_rsa.h#L257


libraries/ESP8266WiFi/src/BearSSLHelpers.cpp
```diff
@@ -938,9 +938,12 @@ uint32_t SigningVerifier::length()
 extern "C" bool SigningVerifier_verify(PublicKey *_pubKey, UpdaterHashClass *hash, const void *signature, uint32_t signatureLen) {
   if (_pubKey->isRSA()) {
     bool ret;
```
d-a-v, line 940: Maybe the 64 reason could be mentionned here ?
```
// 64: https://github.com/earlephilhower/bearssl-esp8266/blob/6105635531027f5b298aa656d44be2289b2d434f/inc/bearssl_rsa.h#L257
```

cores/esp8266/Updater.cpp
```diff
@@ -224,71 +227,87 @@ bool UpdaterClass::end(bool evenIfRemaining){
   }
 
   if (_verify) {
+    // If expectedSigLen is non-zero, we expect the last four bytes of the buffer to
+    // contain a matching length field, preceded by the bytes of the signature itself.
+    // But if expectedSigLen is zero, we expect neither a signature nor a length field;
+    static constexpr uint32_t sigSize = 4;
```
d-a-v, line 233: `sigSize = 4` was `sigSize = sizeof(uint32_t)` or better should be `sigSize = sizeof(sigLen)` no ?
(I must mention it, because I'd feel bad to go on with teaching students that constants are forbidden in the middle of a source code)

mcspr: It was, just all over the place. `using SigType = uint32_t;` could also work? Changed to sizeof right now
(but... scope constants are extremely useful, I guess am a rebel like that)

d-a-v: I guess anything but a scalar would do it :)

PR #8546, opened 2022-04-19
Initialize SoftAP DhcpServer object on demand

### LINKED ISSUE #7871: Wifi.softAP DHCP issue - ERROR: send_offer (error -13)
### Basic Infos

#### Platform

- Hardware:				ESP8266 node MCU V3
- Core Version:			2.7.2
- Operating System:		Windows

### Settings in IDE

- Module:  			nodemcuv2
- Flash Size: 		        4MB
- CPU Frequency:		[80Mhz
- Upload Using:		SERIAL]

### Problem Description

When I run WIFI in soft AP mode I can't connect to my board using DHCP. I've tried from different devices both Windows PC and Android phones failed during obtaining IP address. When I assign static IP everything works fine. I've checked network communication using Wireshark and DHCP offer message wasn't send from ESP.  When I enabled debug I see following errors:
ERROR: send_offer (error -13)
ERROR: dhcps send ack (error -13)

I've tried different devices and play with a lot of different settings like: WiFi.setSleepMode. IP configuration, dhcps_offer_option, WiFi.mode but without success. I've also check a lot of similar issues (one with mDNS, too short password etc) but none of them fix mine.  I will be really glad if somebody give me some advice.

```cpp

#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <ESPAsyncTCP.h>
#include <ESPAsyncWebServer.h>
#include <AsyncJson.h>
#include <ArduinoJson.h>
#include <DHT.h>
#include <LittleFS.h>

extern "C" {
  #include "user_interface.h"
}

void setup() {
     Serial.begin(115200);
     Serial.setDebugOutput(true);

     WiFi.disconnect(true);
     WiFi.mode(WIFI_AP);

     uint8 mode = 1;
     wifi_softap_set_dhcps_offer_option(OFFER_ROUTER, &mode);

     WiFi.softAP(systemConfig.wifiName, systemConfig.wifiPassword, 9);
     delay(500); //wait for SYSTEM_EVENT_AP_START 
     WiFi.softAPConfig(IPAddress{10, 0, 0, 1}, IPAddress{10, 0, 0, 1}, IPAddress{255, 255, 255, 0});

     Serial.println("Connected:");
     
     Serial.println("DHCP status:");
     Serial.println(wifi_softap_dhcps_status());

    //starting web server etc
}

void loop() {
   //basic reading sensor operation nothing heavy
}

```

### Debug Messages
SDK:2.2.2-dev(38a443e)/Core:2.7.2=20702000/lwIP:STABLE-2_1_2_RELEASE/glue:1.2-30-g92add50/BearSSL:5c771be
sleep disable
[AP] softap config unchanged
[APConfig] local_ip: 10.0.0.1 gateway: 10.0.0.1 subnet: 255.255.255.0
[APConfig] DHCP IP start: 10.0.0.100
[APConfig] DHCP IP end: 10.0.0.200
Connected:
10.0.0.1
DHCP status:
1
wifi evt: 7
aid 1
station: 28:b2:bd:cf:c0:f1 join, AID = 1
wifi evt: 5
ERROR: send_offer (error -13)
wifi evt: 7
....
wifi evt: 7
wifi evt: 9
wifi evt: 7
ERROR: dhcps send ack (error -13)
wifi evt: 9
wifi evt: 7
wifi evt: 7
...

### LINKED ISSUE #8307: DHCPS included in the Core and the old SDK implementation
### Basic Infos

#### Platform

- Hardware: any
- Core Version: 3.0.x

### Problem Description

Something noticed while building lwip2 [here](https://github.com/d-a-v/esp82xx-nonos-linklayer/pull/46)

Since #6680 (https://github.com/esp8266/Arduino/commit/51c2a1437b720d584d2f8dd7e01efeb608f991a4#diff-39ab7a280ef46486a19b45c179a5a160a836d57ab164f627ade75cb92ee94d54), old lwip-bundled dhcps was extracted and rewritten into a class

Current example for SoftAP replaced the original calls with the Arduino-API-like global object:
https://github.com/esp8266/Arduino/blob/211606fe9cc781d576a1ae074607e172050b7d64/libraries/ESP8266WiFi/examples/StaticLease/StaticLease.ino#L74-L80
And the internal WiFi code is also forced to work through the wrapper:
https://github.com/esp8266/Arduino/blob/211606fe9cc781d576a1ae074607e172050b7d64/libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp#L247

While the original functions are still in the SDK header. Since the lwip1 code is removed, this is just a linking error for the existing programs pre 3.0.0 that use any of `wifi_softap_..._dhcps_...` funcs
https://github.com/esp8266/Arduino/blob/211606fe9cc781d576a1ae074607e172050b7d64/tools/sdk/include/user_interface.h#L385-L395

I'd propose *keeping* the SDK API, but still have a custom class for an netif when actually warranted (maybe in cases like Ethernet?). This will also remove the need for the WiFi class code to reference the SoftAP DHCPS through an Arduino sketch-like API (...which also means, there will be no need to have headers with carefully placed externs)

cc @d-a-v

also, there is slightly more up-to-date rtos sdk implementation (which also has english comments)
https://github.com/espressif/ESP8266_RTOS_SDK/blob/master/components/lwip/apps/dhcpserver/dhcpserver.c

but, it is interesting that the esp32 variant of idf does not include any alternative to add/remove reserved mac feature
https://github.com/espressif/esp-idf/blob/master/components/lwip/include/apps/dhcpserver/dhcpserver.h

### LINKED ISSUE #8471: broken wifi_softap_set_dhcps_lease function in 3.x.x version
How does DHCP server set IP address of clients ?

Before I did that:

```
    struct ip_info info;
    wifi_set_opmode(STATIONAP_MODE);
    wifi_softap_dhcps_stop();
    IP4_ADDR(&info.ip, 10, 0, 0, 1);
    IP4_ADDR(&info.gw, 10, 0, 0, 1);
    IP4_ADDR(&info.netmask, 255, 255, 255, 0);
    wifi_set_ip_info(SOFTAP_IF, &info);
    
    struct dhcps_lease dhcp_lease;
    IP4_ADDR(&dhcp_lease.start_ip, 10, 0, 0, 100);
    IP4_ADDR(&dhcp_lease.end_ip, 10, 0, 0, 105);
    wifi_softap_set_dhcps_lease(&dhcp_lease);
```

But since version 3.x.x I'm getting compilation exception: undefined reference to `wifi_softap_set_dhcps_lease'

mcspr:
Remove dependency on global ctor, and just construct the object when
someone asks us to do it. Only dependency right now is netif_git, which
is expected to be initialized by the lwip code some time before
dhcps_start happens.

Removing ip_info from begin(), since we never reference later on.
Also removing the specific check for netif id and simplify the ctors.
Fixing _netif->state check left over from the old code.

Update tests and recover old nonos-sdk dhcps functions that were not implemented.

resolve #7871
resolve #8307
resolve #8471


libraries/ESP8266WiFi/src/ESP8266WiFiAP.cpp
```diff
@@ -244,21 +247,22 @@ bool ESP8266WiFiAPClass::softAPConfig(IPAddress local_ip, IPAddress gateway, IPA
...
+    if(!server.set_dhcps_lease(&dhcp_lease))
     {
         DEBUG_WIFI("[APConfig] wifi_set_ip_info failed!\n");
         ret = false;
     }
 
     // set lease time to 720min --> 12h
-    if(!dhcpSoftAP.set_dhcps_lease_time(720))
+    if(!server.set_dhcps_lease_time(720))
     {
```
d-a-v, line 259: Lease time is by default 120 seconds, which was a ridiculous default lease time.
Setting it to 720 here is better, but it would be better to make it a default value.

Also these self-`set_`functions always return true.
Would there be a static-constexpr way to avoid this code and messages ?

mcspr: 720 is new default, no need to call it then.

true, but that would be a lot of loc changes. redundant isRunning, various sanity checks may as well happen in init? since I am pretty sure I'll also will write options code, keeping changes here at a minimum

d-a-v (approved):
Thanks!

PR #8548, opened 2022-04-21
WebServer: use String when working with Basic authentication

mcspr:
Avoid blowing up user code when `$user:$password` string is longer than
127 bytes. Use String to both manage the memory and handle concatenation.

Also clean-up historical quicks such as
- `if(StringObject)` that is always true since we implemented SSO
- `authReq = "";` / `authReq = String();`, which will happen anyway
- `(String)...` casts that happen anyway, implicitly

### DIFF
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
@@ -108,13 +108,12 @@ bool ESP8266WebServerTemplate<ServerType>::authenticate(const char * username, c
 
       String raw;
       raw.reserve(username_len + password_len + 1);
-      if(!raw.length()) {
-        return false;
-      }
-
       raw.concat(username, username_len);
       raw += ':';
       raw.concat(password, password_len);
+      if(!raw.length()) {
+        return false;
+      }
 
       String encoded = base64::encode(raw, false);
       if(!encoded.length()){
--- a/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
+++ b/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
@@ -102,31 +102,32 @@ bool ESP8266WebServerTemplate<ServerType>::authenticate(const char * username, c
     if(authReq.startsWith(F("Basic"))){
       authReq = authReq.substring(6);
       authReq.trim();
-      char toencodeLen = strlen(username)+strlen(password)+1;
-      char *toencode = new (std::nothrow) char[toencodeLen + 1];
-      if(toencode == NULL){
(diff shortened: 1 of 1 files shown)


libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -102,31 +102,31 @@ bool ESP8266WebServerTemplate<ServerType>::authenticate(const char * username, c
...
-      char toencodeLen = strlen(username)+strlen(password)+1;
-      char *toencode = new (std::nothrow) char[toencodeLen + 1];
-      if(toencode == NULL){
-        authReq = "";
+
+      const size_t username_len = strlen(username);
+      const size_t password_len = strlen(password);
+
+      String raw;
+      raw.reserve(username_len + password_len + 1);
```
earlephilhower, line 110: I think we need  to add one for the `\0` because we keep `String`s 0-terminated (no stinky Pascal here!).  `::reserve` allocates only the size requested (modulo 16):
https://github.com/esp8266/Arduino/blob/02c1a502ca475d1bc202ba881fc5a2eb07a532d5/cores/esp8266/WString.cpp#L220
https://github.com/esp8266/Arduino/blob/02c1a502ca475d1bc202ba881fc5a2eb07a532d5/cores/esp8266/WString.cpp#L234

earlephilhower (changes requested):

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
mcspr: Already accounted for? e.g. we ask for 255, and get 256; ask for 256, get 272; and etc.
```python
def fits(length, capacity):
    return (length + 1) <= capacity

for asked in range(12, (2 << 15)):
    assert(fits(asked, (asked + 16) & (~0xf)))
```

earlephilhower: I'm talking about the %16==0 case.  Then you ask for, say, 32, and get 32, but need to realloc on the last assignment.

All others, sure, are fine.

mcspr: It is what I am saying though, you get a larger capacity when that happens
```python
if length == 32:
    print(f'{length=} {capacity=')
    break
```
```
length=32 capacity=48
```

earlephilhower: Ah, definitely my bad.  I didn't read the code thoroughly, sorry!

earlephilhower (approved):

PR #8564, opened 2022-05-12
Update RFC 5246 URL

christopolise:
The old link to RFC 5246 has been moved to https://tools.ietf.org/search/rfc5246


earlephilhower (approved):
Thanks!

PR #8565, opened 2022-05-12
Fix utf-8 encoding

mhightower83:
Issue posted to commit
https://github.com/esp8266/Arduino/commit/d1d4212e8ba799fb16530f2f97a821aa2ce0e2fb#r73517358

Should resolve UnicodeDecodeError: multibyte-byte.
I don't have an environment to test this; however, this version still works on my machine.

The python docs say "UTF-8 is one of the most commonly used encodings, and Python often defaults to using it."
Hmm, "often" wasn't often enough.  I set encoding to "utf-8" for all `open()` calls. Is there a chance that doing all of them will cause a problem?


jjsuwa-sys3175:
LGTM

IMHO touching utime should open by encoding-independent manner eg. binary mode

PR #8566, opened 2022-05-13
Patch eap.o memory leak

mhightower83:
WiFi Enterprise option can leak up to 3 allocations per connect/disconnect cycle (configuration dependent): anonymous Identity, password, and some unidentified allocation.

This solution patches `eap.o` from `libwpa2.a` to call a special 2-part wrapper instead of `vPortFree` at cleanup.

This is supplemental to https://github.com/esp8266/Arduino/pull/8529#issue-1192106505 and should not be merged prior.

There may be a better way of doing this. My knowledge of `xtensa-lx106-elf-objcopy` is near zero and everything that sounded promising didn't work the way I needed. 

I don't have a use case for WiFi Enterprise so my testing has been limited. More testing is needed.

To minimize resources used, the patch will only be linked in when you have a call to `enable_wifi_enterprise_patch()`.

Use something like this to enable patch:
```cpp
#include <ESP8266WiFi.h>
#include <coredecls.h>
// ...
void setup() {
  enable_wifi_enterprise_patch();
// ...
}
```


cores/esp8266/wpa2_eap_patch.cpp
```diff
@@ -0,0 +1,118 @@
...
+    ".size z2EapFree, .-z2EapFree\n\t"
+);
+
+/*
+ * While some insight can be gained from the ESP32 repo for this structure.
+ * It does not match exactly. This alternate structure focuses on correct offset
+ * rather than trying to exactly reconstruct the original labels.
+ */
+struct StateMachine { // size 200 bytes
+    void* befoeConfig[16];
```
Flole998, line 70: Not that it matters, but you probably meant `beforeConfig` (you're missing the r here).

Flole998:
Excellent job, I am really impressed. I tried to do something similar but failed miserably.

cores/esp8266/wpa2_eap_patch.cpp
mhightower83: Thanks! I need to get a new keyboard. I think the keys bind a bit and I don't notice, due to my brain's tendency to auto-complete. I think it is a feature of dyslexia where I cannot see things that are missing. :frowning_face:

mhightower83:
@earlephilhower Thanks! I started using this bug to test some code I was working on to catch a double free. I ended out making several more debug enhancements to umm_malloc. These helped reduce the amount of assembly code I was reading.

PR #8568, opened 2022-05-14
feat(ESP8266HTTPUpdateServer): add option `_rebootOnUpdate`

ruleechen:
Add this option just like what ESP8266httpUpdate does. 

Now system restaerts after updating Firmware and also FileSystem. Sometimes new Firmware to update is rely on a new FileSystem, only when we update both Firmware and FileSystem then safe to perform restart.


ruleechen:
I am closing this since there are issue as my testing.

PR #8569, opened 2022-05-16
WString: remove operator==(const __FlashStringHelper*)

d-a-v:
fixes https://gitter.im/esp8266/Arduino?at=627fee0ed30b6b44ebeaa803

@bwjohns4 @jandrassy ~~there will still be a warning, because `nullptr` should be used instead of `NULL`~~.

### DIFF
--- a/cores/esp8266/WString.h
+++ b/cores/esp8266/WString.h
@@ -227,9 +227,6 @@ class String {
         bool operator ==(const char *cstr) const {
             return equals(cstr);
         }
-        bool operator ==(const __FlashStringHelper *rhs) const {
-            return equals(rhs);
-        }
         bool operator ==(std::nullptr_t) const {
             return length() == 0;
         }
@@ -238,9 +235,6 @@ class String {
         }
         bool operator !=(const char *cstr) const {
             return !equals(cstr);
-        }
-        bool operator !=(const __FlashStringHelper *rhs) const {
-            return !equals(rhs);
         }
         bool operator !=(std::nullptr_t) const {
             return length() != 0;
--- a/cores/esp8266/WString.h
+++ b/cores/esp8266/WString.h
@@ -211,9 +211,10 @@ class String {
             return *this;
         }
 
-        // checks whether the String is empty
+        // checks whether the internal buffer pointer is set.
+        // (should not be the case for us, since we always reset the pointer to the SSO buffer instead of setting it to nullptr)
         explicit operator bool() const {
-            return length() != 0;
+            return buffer() != nullptr;
         }
 
         int compareTo(const String &s) const;
--- a/cores/esp8266/WString.h
+++ b/cores/esp8266/WString.h
@@ -230,6 +230,33 @@ class String {
         bool operator ==(const __FlashStringHelper *rhs) const {
             return equals(rhs);
         }
+        bool operator ==(std::nullptr_t) const {
+            return length() == 0;
+        }
+        bool operator ==(int num) const {
+            return equals(String(num));
+        }
+        bool operator ==(unsigned int num) const {
+            return equals(String(num));
+        }
+        bool operator ==(long num) const {
+            return equals(String(num));
+        }
+        bool operator ==(unsigned long num) const {
+            return equals(String(num));
+        }
+        bool operator ==(long long num) const {
+            return equals(String(num));
+        }
+        bool operator ==(unsigned long long num) const {
+            return equals(String(num));
+        }
+        bool operator ==(float num) const {
+            return equals(String(num));
+        }
+        bool operator ==(double num) const {
+            return equals(String(num));
+        }
         bool operator !=(const String &rhs) const {
             return !equals(rhs);
         }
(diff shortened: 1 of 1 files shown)


mcspr:
We don't want to overload comparisons though? Check out operator calls you are proposing with older versions, or the esp32 Core, or even the ArduinoCore API test suite.
```
// with 2.7.4
src/main.cpp: In function 'void setup()':
src/main.cpp:7:16: error: invalid conversion from 'int' to 'const char*' [-fpermissive]
     if (tmp == 5) {
                ^
```

Main ambiguity comes from fpstr vs. const char, that wasn't there in the older Core versions.
`== NULL` is still an `equals(int)`, even with the overloaded nullptr you are adding
it is a warning, but not something critical from the gcc pov.

Yes, this is another case of older code breaking, but I'd rather WiFiManager fixes this issue b/c we add another heap of issues on top.


```
src/main.cpp: In function 'void setup()':
src/main.cpp:7:16: warning: passing NULL to non-pointer argument 1 of 'bool String::operator==(int) const' [-Wconversion-null]
    7 |     if (tmp == NULL) {
      |                ^~~~
In file included from /home/builder/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/Arduino.h:291,
                 from src/main.cpp:1:
/home/builder/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/WString.h:234:30: note:   declared here
  234 |         bool operator ==(int val) const {
      |
```

cores/esp8266/WString.h
```diff
@@ -230,6 +230,33 @@ class String {
         bool operator ==(const __FlashStringHelper *rhs) const {
             return equals(rhs);
         }
+        bool operator ==(std::nullptr_t) const {
+            return length() == 0;
+        }
```
mcspr, line 235: re. comment, only this part would make sense for a fix. the rest is even more ambigious issues later down the line

d-a-v:
Removing `==(FlashStringHelper*)` overload still allows comparison with `P("test")` which will use an implicit temporary `String(P("test"))`,
and it also restores warning-less & correct comparison with NULL through `==(const char*)` overload.

mcspr:
note that #8106 used this as a way to shrink ESPeasy binary size due to a temporary String obj

d-a-v:
> note that #8106 used this as a way to shrink ESPeasy binary size due to a temporary String obj

True.
... after checking, the NULL comparison error is existing from this pull request, which happenned after few months 3.0.2 release.

d-a-v:
With this last commit, `NULL` is correctly managed, with a deprecating warning so user is invited to switch to `nullptr`

cores/esp8266/WString.h
```diff
@@ -230,6 +229,13 @@ class String {
...
+        [[deprecated("use nullptr instead of NULL")]]
+        bool operator ==(decltype(NULL)) const {
```
mcspr, line 236: just note that the trade-off is *someone will write* `obj == 5;` returning false positive :)
(...and iirc ide default to 'no warnings', which does not help...)

cores/esp8266/WString.h
```diff
@@ -211,10 +211,9 @@ class String {
             return *this;
         }
 
-        // checks whether the internal buffer pointer is set.
-        // (should not be the case for us, since we always reset the pointer to the SSO buffer instead of setting it to nullptr)
+        // checks whether the String is empty
         explicit operator bool() const {
-            return buffer() != nullptr;
+            return length() != 0;
```
mcspr, line 216: ?
I thought the idea was to check whether the string is a usable object. Original string (and anyone using ArduinoCore api stuff, which retains this behaviour) need this check since it *may* return null when calling c_str()
(...but, might as well have checked that directly...)

d-a-v: We are currently incorrectly using it [by checking a String length](https://github.com/esp8266/Arduino/blob/1bb041bb6c488dd5e6462ab8b84a01fd2ebae168/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L324) which is the natural meaning of this operator no ?

mcspr: Sort of. Arduino api wrote it as a check for buffer() ptr being set, not length
Plus, operator[] includes both length and buffer() != null checks anyway

cores/esp8266/WString.h
mcspr: I think they refer to some old version of GCC (pre 4.6?) with the `__null` stuff.
And I'd like to find some examples of `#define NULL std::nullptr`, as it seems like something for the system header to have

d-a-v: The [only one](https://github.com/unicode-org/icu/blob/242ad4723ed579bd1f317846b2c04d96b05750cd/icu4c/source/common/unicode/utypes.h#L179-L190) I could find that is installed on my workstation and probably rarely used.

cores/esp8266/WString.h
d-a-v: I had missed the [official API updated version](https://github.com/arduino/ArduinoCore-avr/blob/bc792a9a716289249fa5f2f67be3ee8ba6507d6c/cores/arduino/WString.h#L138) (and https://www.artima.com/articles/the-safe-bool-idiom). 
This is far beyond this PR (and simplicity). I have reverted this change.

d-a-v:
This change proposal ends up with simply removing `operator==(FlashStringHelper)`.

cores/esp8266/WString.h
mcspr: So we could've had this
https://github.com/gcc-mirror/gcc/blob/1815462a6e53465c404f8a5f6116891492d4b50b/gcc/ginclude/stddef.h#L396

But, newlib overrides it
https://github.com/earlephilhower/newlib-xtensa/blob/ebc967552ce827f21fc579fd8c437037c1b472ab/newlib/libc/include/stdlib.h#L60-L62

mcspr: Also, glibc code requesting it
https://code.woboq.org/userspace/glibc/stdlib/stdlib.h.html#_M/__need_NULL

cores/esp8266/WString.h
mcspr: Just to note, that was pre C++11. Article issue is resolved by `explicit operator bool() const`

mcspr:
*renamed*
`str == F(...)` alternative would be `str == PSTR(...)` accepting progmem'ed strings
pointer coercion will drive us mad otherwise

PR #8570, opened 2022-05-16
platform.txt: Override python3 anon pipe encoding to utf-8 for 'tools/sizes.py'

jjsuwa-sys3175:
see https://github.com/esp8266/Arduino/commit/da4a19fdacd7a859da395e014c9e0fded99aa985#commitcomment-73793209


d-a-v (approved):
no change w/linux

jjsuwa-sys3175:
> no change w/linux

thx

a little concern: `-X utf8` requires python 3.7 or later

mcspr:
Thanks. Also a good reminder to double-check any python scripts to force text encoding :)
BTW, what happens with legacy cmd.exe when invoking the script manually?

jjsuwa-sys3175:
> BTW, what happens with legacy cmd.exe when invoking the script manually?

scrnshot

PR #8572, opened 2022-05-17
tools/sizes.py: Restore to the prior behavior regarding output destination

jjsuwa-sys3175:
once `tools/sizes.py` outputted to `stderr` rather `stdout` before da4a19fdacd7a859da395e014c9e0fded99aa985

(Arduino IDE 1.8.19 for Windows doesn't capture `stdout`, 2.0.0-rc6 does)

### DIFF
--- a/tools/sizes.py
+++ b/tools/sizes.py
@@ -20,7 +20,9 @@
 import argparse
 import os
 import subprocess
+import sys
 
+sys.stdout = sys.stderr
 
 def get_segment_sizes(elf, path, mmu):
     iram_size = 0


mcspr:
Should we, though? This is not an error, as the coloring might suggest, but a simple informative message.

---

BTW I thought we had an option to replace `Sketch uses...` lines (ref. commit comments), but...
https://github.com/esp8266/Arduino/blob/1bb041bb6c488dd5e6462ab8b84a01fd2ebae168/platform.txt#L157-L161

It is actually something hard-coded
https://github.com/arduino/arduino-cli/blob/18cb00c25bcb5c7dfc2dc733231ff391a281899e/legacy/builder/phases/sizer.go#L97

Only future 2.0.0 IDE allows to have a custom output (by reading certain JSON field)
https://github.com/arduino/arduino-cli/blob/18cb00c25bcb5c7dfc2dc733231ff391a281899e/legacy/builder/phases/sizer.go#L52

jjsuwa-sys3175:
> This is not an error, as the coloring might suggest, but a simple informative message.

i understand what you mean, but no sizes shown actually... only in my environment?
scrnshot

better than not seeing anything, even if that color may suggest an error.

tools/sizes.py
```diff
@@ -158,14 +158,14 @@ def main():
     for group, (segments, total) in sizes:
         number, used, segments = filter_segments(segments)
 
-        print(f". {group:<8}, used {used} / {total} bytes ({percentage(used, total)})")
-        print("|   SEGMENT  BYTES    DESCRIPTION")
+        sys.stderr.write(f". {group:<8}, used {used} / {total} bytes ({percentage(used, total)})\n")
+        sys.stderr.write("|   SEGMENT  BYTES    DESCRIPTION\n")
```
mcspr, line 162: `sys.stdout = sys.stderr` just before print() will effectively do the same thing

mcspr:
Not really, based on the builder's code :/
https://github.com/arduino/arduino-cli/search?q=utils.Show+%2F*+stderr+*%2F
(also assuming current arduino-cli code is the same as the included builder one)

I have tried `.postbuild`, but the same rules apply
```ini
recipe.hooks.sketch.postbuild.1.pattern="{runtime.tools.python3.path}/python3" -I "{runtime.tools.sizes}" --elf "{build.path}/{build.project_name}.elf" --path "{runtime.tools.xtensa-lx106-elf-gcc.path}/bin" --mmu "{build.mmuflags}"
```

Oh well.

per1234:
>Arduino IDE 1.8.19 for Windows doesn't capture stdout, 2.0.0-rc6 does

I happened to take a look at this today, and was confused by the quoted statement. Since the "why is it printed in red in the IDE" question is likely to be raised by the users periodically, I'll add some information here:

When the "**Show verbose output during: ☐ compilation**" Arduino IDE preference is disabled, the IDE only prints the output from the stderr stream of the compilation commands. The stdout stream output of the commands is only printed when that preference is enabled.

This handling of compilation output is the same in Arduino IDE 1.x and 2.x so I can only think of two possible causes of the claimed different behavior between the two versions:

- The "**Show verbose output during: ☐ compilation**" Arduino IDE preference was set differently from one IDE to another (IDE 2.x uses a different preference store from 1.x)
- This bug: https://github.com/arduino/arduino-ide/issues/374 that caused verbose compilation output on upload when verbose upload output was enabled, which has been fixed since Arduino IDE 2.0.0-rc9.1

PR #8579, opened 2022-05-22
Move _GNU_SOURCE from CCFLAGS to CPPDEFINES

maxgerhardt:
For more than [2 years](https://community.platformio.org/t/identifier-is-undefined-setenv-tzset/16162) people were complaining that in VSCode, certain functions were not recognized as existing, but compiled just fine. For example, `tzset()`, `setenv` or [strlcpy](https://community.platformio.org/t/intellisense-error-identifier-strlcpy-is-undefined/27873).

The cause of building fine but not intellisense visible in regards to these functions is because they're GNU / BSD functions whose declaration is only enabled in headers if e.g. `__BSD_VISBLE` is defined, which gets activated through one possible chain through `_GNU_SOURCE`. PlatlformIO gives the `-D_GNU_SOURCE` flag via `CCFLAGS` (like [Arduino platform.txt does](https://github.com/esp8266/Arduino/blob/master/platform.txt#L73)), which causes that flag to be given in compilation correctly, but not showing up in Intellisense (`.vscode/c_cpp_properties`), which only shows `CPPDEFINES`. The usage of `-D...` in CCFLAGS or the macro itself in CPPDEFINES is equivalent.

This PR fixes that `_GNU_SOURCE` is given via `CPPDEFINES` instead and is thus visible in intellisense and fixes all these "Intellisense can't find function but it still compiles" errors. Compilation is logically unaffected. Tested fine on my local machine.


mcspr:
Doesn't it also mean `-U__STRICT_ANSI__` is not being processed? See 7e6ee374a85bef237e286536d74d823d7b20de02 for the reasoning
Just curious, is there a standalone way to generate c_cpp_properties from our sources, or it is something internal to vscode, it's cpptools or platformio extension?

maxgerhardt:
PlatformIO generates the `c_cpp_properties.json` by executing the template logic [here](https://github.com/platformio/platformio-core/tree/develop/platformio/project/tpls/vscode/.vscode) and [here](https://github.com/platformio/platformio-core/blob/develop/platformio/project/generator.py). 

Indeed, the `-U__STRICT_ANSI__` does not appear in this file (see e.g. [here](https://pastebin.com/cKZRPDC3) for after PR state). And looking at the [microsoft docs](https://code.visualstudio.com/docs/cpp/c-cpp-properties-schema-reference), they have `defines`, but not `undefines`, so not sure how to inform Intellisense about that all.

But of course PlatformIO still gives that `-U` flag into the compilation process. It's just the VSCode Intellisense that knows nothing about it -- and it doesn't seem to be causing display issues with missing functions.

maxgerhardt:
The function referenced in the STRICT ANSI comment displays nicely with my PR.

grafik

What I am noticing though is that maybe a more general fix in the PlatformIO could would be better, so that if `CCFLAGS` contains `-D` flags for defines, it correctly transfers them to the `defines` in the IDE intellisense files.

maxgerhardt:
Let's wait a bit so that we might resolve this issue in the core.

PR #8582, opened 2022-05-27
DHCP custom option

mcspr:
@bwjohns4
re comments in #8571


d-a-v:
To the risk of being pedantic, If there a way to retrieve user option from flash ?

mcspr:
Sure, but what is the use-case anyway?
Playing around with the code here... We could have static opts as static vars, sure, but then we can't have anything dynamic that way. e.g. based on the netif IP, or some other service info we would want to add.

Suppose, we have https://github.com/esp8266/Arduino/pull/8451 with a pointer attached to the instance instead
```cpp
    dhcpSoftAP.custom_offer_options = [](auto& options, const DhcpServer& server) {
        options.add(/* code = */ 43, /* data = */ {0xca, 0xfe, 0xca, 0xfe, 0xfe});

        char clientId[32];
        ... somehow generate it from server._netif data ...
        options.add(93, reinterpret_cast<const uint8_t*>(&clientId[0]), len);
    };
```
Where `auto& options` is the magic handler with size + code checks from above.

d-a-v:
My fingers always hurt when storing an array in a source code, which is copied on boot from flash to ram, then by code from ram to ram. Data are there three times, so this was my question. Of course some dhcp configuration blocks are dynamic, so it should not be prevented to deal with them.
I was thinking of something like `memcpy_P(optptr, &option.data.begin(), option.data.length())` instead of this loop
```
        for (auto it = option.data.begin(); it != option.data.end(); ++it) {
            *optptr++ = *it;
```
But yes, dhcp data blocks are often dynamic and among the static ones not many are long enough to get a real ram gain.
So yes, it was a pedantic question :stuck_out_tongue:

mcspr:
With the above, add() could accept u8 pointer then and do memcpy.
Let me fight the clang-format out first... It does some weird stuff with the built-in options when processing them as function arguments.

libraries/ESP8266WiFi/examples/RangeExtender-NAPT/RangeExtender-NAPT.ino
```diff
@@ -53,9 +53,10 @@ void setup() {
   }
   Serial.printf("\nSTA: %s (dns: %s / %s)\n", WiFi.localIP().toString().c_str(), WiFi.dnsIP(0).toString().c_str(), WiFi.dnsIP(1).toString().c_str());
 
-  // give DNS servers to AP side
-  dhcpSoftAP.dhcps_set_dns(0, WiFi.dnsIP(0));
-  dhcpSoftAP.dhcps_set_dns(1, WiFi.dnsIP(1));
+  // by default, DNS option will point to the interface IP
+  // instead, point it to the real DNS server
+  // (notice that only a single IP is supported)
```
d-a-v, line 58: [This is not totally true](https://www.rfc-editor.org/rfc/rfc2132.html#section-3.8) and thanks for this comment.
Multiple option 6 (DHCP_OPTION_DNS_SERVER) seems to be allowed by RFC and it was not supported by our current code. I'm not sure it is useful to change anything but this comment, here or there.

mcspr: ...supported by the *implementation*. should there be `add(code, ip4_addr_t[])` and list of dns servers instead of just one?

d-a-v: Nope, I guess, until someone demonstrates the need for this :]

cores/esp8266/LwipDhcpServer.cpp
```diff
@@ -329,145 +338,50 @@ bool DhcpServer::add_dhcps_lease(uint8* macaddr)
...
 
-#if 0  // already set (DHCP_OPTION_SUBNET_MASK==1) (??)
-    *optptr++ = 0x01;
-    *optptr++ = 4;
-    *optptr++ = 0;
-    *optptr++ = 0;
-    *optptr++ = 0;
-    *optptr++ = 2;
-#endif
+        options.add(DHCP_OPTION_BROADCAST_ADDRESS, &broadcast);
```
d-a-v, line 371: I think [`broadcast = (addr & mask) | ~mask`](https://github.com/esp8266/Arduino/blob/9e2103f27e9eacba39d25697bacc5142a7aa5e66/cores/esp8266/LwipDhcpServer.cpp#L1001-L1003)

d-a-v (approved):
Nice one !

blue-wind-25:
Hello,

In 'LwipDhcpServer.cpp' line 444 I found:
`std::memcpy((u8_t*)q->payload, reinterpret_cast<u8_t*>(&m), q->len);`

shouldn't that be:
`std::memcpy((u8_t*)q->payload, (u8_t*)m, q->len);`

(use 'm' instead of '&m')?


Also, in 'ESP8266WiFiAP.cpp', the call to `wifi_softap_dhcps_stop()` in `ESP8266WiFiAPClass::softAPConfig()` has been removed. Shouldn't that be kept there so that the DHCPS from the SDK is kept turned off? 

In previous cores, I can use custom IP such as '192.168.0.XXX' using DHCP in access point mode, but if  `wifi_softap_dhcps_stop()` is removed then that custom IP will not work.


Could you verify those two? Thank you!

mcspr:
> shouldn't that be:
std::memcpy((u8_t*)q->payload, (u8_t*)m, q->len);

it should. just a bad c/p

> Also, in 'ESP8266WiFiAP.cpp', the call to wifi_softap_dhcps_stop() in ESP8266WiFiAPClass::softAPConfig() has been removed. Shouldn't that be kept there so that the DHCPS from the SDK is kept turned off?

probably? iirc, there's no difference between `server.end()` and `wifi_softap_dhcps_stop()` besides  `wifi_softap_dhcps_status()` getting changed. it ends up calling `dhcps_stop()` which is proxied to `server.end()` anyway, while also doing some extraneous checks we don't really care about

PR #8589, opened 2022-05-30
Update broken links

gatesry:
https://www.arduino.cc/en/main/software - broken
https://www.arduino.cc/en/software - working


gatesry:
Oops, misclick. Thanks for confirming.

PR #8590, opened 2022-05-31
added an example to switch LED  using ESP8266WebServer

nsisong:


cziter15:
Why changing led pin state in main loop function?

PR #8597, opened 2022-06-12
Avoid reading past end of non-zero terminated char arrays

paulocsanz:
Some char* passed may be non-zero terminated slices, we could both endup with a non-zero terminated `String` and read past the end of the allocation, so on unmapped memory (or even possibly flash memory).

This fixes the issue.

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -260,7 +260,8 @@ String &String::copy(const char *cstr, unsigned int length) {
         return *this;
     }
     setLen(length);
-    memmove_P(wbuffer(), cstr, length + 1);
+    memmove_P(wbuffer(), cstr, length);
+    wbuffer()[length] = 0;
     return *this;
 }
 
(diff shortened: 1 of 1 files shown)


mcspr (approved):

cores/esp8266/WString.cpp
```diff
@@ -411,8 +413,9 @@ bool String::concat(const __FlashStringHelper *str) {
     unsigned int newlen = len() + length;
     if (!reserve(newlen))
         return false;
-    memcpy_P(wbuffer() + len(), (PGM_P)str, length + 1);
+    memcpy_P(wbuffer() + len(), (PGM_P)str, length);
```
d-a-v, line 416: `strlen_P()` is called on this one,
`\0` presence is then ensured ?

paulocsanz: You are right in this case. I can revert it back, I just ended up replacing everywhere that copied length + 1 as it seemed error prone.

d-a-v: That's OK, no need for a further update :+1:

PR #8598, opened 2022-06-12
do not call tcp/udp ::stopAll() when it is not required to do so

### LINKED ISSUE #6123: stopAll() in WiFiGeneric causes currently in progress webServer transmission to stop.
This line: https://github.com/esp8266/Arduino/blob/eea9999dc5eaf464a432f77d5b65269f9baf198d/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp#L224 causes a currently in session webServer send to stop. Before I make an MCVE, can anyone share why this stopAll exists?

If, while the web server is sending a chunked response, a WiFi event occurs, the chunked response is abandoned and the web browser gives an error. Commenting it out appears to fix the problem but maybe there was a reason for it and something else would be broken after removing it?

It may also have some bearing on an older issue: https://github.com/esp8266/Arduino/issues/1554

d-a-v:
fixes #6123

### DIFF
--- a/libraries/ArduinoOTA/ArduinoOTA.cpp
+++ b/libraries/ArduinoOTA/ArduinoOTA.cpp
@@ -265,8 +265,6 @@ void ArduinoOTAClass::_runUpdate() {
   delay(100);
 
   Update.setMD5(_md5.c_str());
-  WiFiUDP::stopAll();
-  WiFiClient::stopAll();
 
   if (_start_callback) {
     _start_callback();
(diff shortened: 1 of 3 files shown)


mcspr:
ArduinoOTA as well? Even a failed update kills every client
Also note of WebUpdate example

Continuing the idea, `UpdaterClass` should probably handle that
https://github.com/esp8266/Arduino/blob/8bfc2e9eabbd33f53f67c736e322a9bd88788fc5/libraries/ArduinoOTA/ArduinoOTA.h#L50-L61
Where the user might as well do the
```cpp
WiFiClient::stopAll();
WiFiUDP::stopAll();
```
```cpp
// or, where we want some fine grained control
WiFiClient::stopAllExcept(&server.client());
WiFiUDP::stopAllExcept(&myPreciousCnC);
```

mcspr (approved):

PR #8603, opened 2022-06-14
Do not show python traceback when esptool fails

mcspr:
https://github.com/espressif/esptool/blob/57bd869b64c4eeaaad74791b2f7795264561f06e/esptool/__init__.py#L993-L998
With addition of a `finally:` branch to handle tmpfile. Otherwise, we get a full traceback for (mostly) no reason.

`os.path.join` knows what to do
`if str:` is implicitly checking for length in strings in containers.


tools/upload.py
```diff
@@ -45,25 +45,30 @@
...
-esptool.main(cmdline)
-
-if len(erase_file):
-    os.remove(erase_file)
+try:
+    esptool.main(cmdline)
+except esptool.FatalError as e:
+    sys.stderr.write('\nA fatal esptool.py error occurred: %s' % e)
+finally:
+    if len(erase_file):
```
earlephilhower, line 71: Should this just be `if erase_file:` to mirror the other changes you've done to Python-ify the script?

earlephilhower (approved):

PR #8604, opened 2022-06-14
Check ESP8285 at runtime

### LINKED ISSUE #7340: Runtime ESP8285 detection and isFlashInterfacePin(x)
### Basic Infos

#### Platform

- Hardware:				ESP8286 / ESP8285
- Core Version:			2.7.1
- Development Env:		Any

### Settings in IDE

- Module:  			Generic ESP8266 Module / ESP8285

### Problem Description

This is more of a feature proposal issue.

Currently, a way to distinguish between esp8266 and esp8285 is by selecting build variant manually, hard-coding certain constants, macros and pin values. Main example is `isFlashInterfacePin(X)` macro:
https://github.com/esp8266/Arduino/blob/27ef03fc057c91f66c32316eec1c5a79838a3979/variants/esp8285/common.h#L34

Which happens to be used defensively across the Core:
https://github.com/plerup/espsoftwareserial/blob/594904c7e4eb0ddb23969765ec55f0b91c5b2ca9/src/SoftwareSerial.cpp#L51
https://github.com/esp8266/Arduino/blob/36e047e908cfa6eafaaf824988070b49f2c2ff2a/cores/esp8266/core_esp8266_waveform.cpp#L120
https://github.com/esp8266/Arduino/blob/a04c3244d5bca7434a8825ad855c1af62489a762/cores/esp8266/core_esp8266_wiring_digital.cpp#L240

However, based on current RTOS code and snippets from esptool.py, we can actually replace macro with a real function detecting pin configuration at runtime (e.g., like one can notice when uploading stuff via esptool.py to the esp8285 device):
https://github.com/espressif/esptool/blob/f04d34bcab29ace798d2d3800ba87020cccbbfdd/esptool.py#L1060-L1070
https://github.com/espressif/ESP8266_RTOS_SDK/blob/3c055779e9793e5f082afff63a011d6615e73639/components/esp8266/include/esp8266/efuse_register.h#L20-L21
Which becomes this piece of code somewhere:
```cpp
const uint32_t efuse_blocks[4] {
    READ_PERI_REG(0x3ff00050),
    READ_PERI_REG(0x3ff00054),
    READ_PERI_REG(0x3ff00058),
    READ_PERI_REG(0x3ff0005c)
};

bool is_esp8285 = (
    (efuse_blocks[0] & (1 << 4))
    || (efuse_blocks[2] & (1 << 16))
);
```

The intent of that is that Generic variant can be merged with esp8285, thus any binary could be built and used with both types of devices.

Although, I wonder if I am missing documentation piece where this is described more clearly, instead of referencing 2-3 lines of code from *somewhere*.

mcspr:
resolves #7340

not really sure about `isFlashInterfacePin(x)`, should there be `isFlashInterfacePinEsp8285(x)`?
main issue is every variant having `#include "../generic/common.h"`


mcspr:
> Should we also change [this one](https://github.com/esp8266/Arduino/blob/7d9abbb60597e0e14abe7ed77c21de5dc63a7d30/variants/esp8285/common.h#L34) or leave it explicitly as statically defined esp8285 ?
> In the latter case, should we disable dynamic detection when esp8285 is selected by user ?

Either if we change board= to esp8285, right now it adds `-DARDUINO_ESP8266_ESP01` when building with esp8285 as variant. Or, a custom flag inside of the variant header

And yes, there's no point in dynamic detection in the `isFlashInterfacePin` macro with specific variant.
Plus, only generic variant may add `esp_is_8285()` call, keeping the rest of variants with the original one.

d-a-v:
> Either if we change board= to esp8285, right now it adds `-DARDUINO_ESP8266_ESP01` when building with esp8285 as variant. Or, a custom flag inside of the variant header

Given that `esp_is_8285()` is always defined, let's not introduce breaking changes by changing macros

d-a-v (approved):

Issue #8372, opened 2021-11-13
SPIFFS println weird behavior with long strings


fabianoriccardi:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					Wemos D1 mini
- Core Version:			commit: bf2882d
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			Wemos D1 mini
- Flash Mode:			dio
- Flash Size: 		        4MB
- lwip Variant:                  v2 Lower Memory
- Reset Method:		default
- Flash Frequency:	        40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   916200

### Problem Description

Sporadically, File::println("message") fails, and instead of recording "\r\n" it prints 0xFFFF. In this sketch, this happens only with long messages (i.e. >350 chars). However, in more complicated sketches (where multiple write per lines occurs), it happens even with short strings. It happens with Arduino Core 3.x.x, as well as the latest commit, but it works fine in previous releases (2.7.4 and lower).

I know that SPIFFS is deprecated, but this seems a regression and it should be mentioned. 

### Sketch

```cpp
# print 10 lines on file and then print back to Serial.

#include <FS.h>

void setup() {
  Serial.begin(115200);
  while(!Serial);
  Serial.println();
  
  // You mays need to format the flash before using it
  //SPIFFS.format();

  if(SPIFFS.begin()){
    Serial.println("Filesystem mounted successfully");
  }else{
    Serial.println("Filesystem NOT mounted. System halted");
    while(1) delay(100);
  }
}

void loop() {
  for(int i=0; i<10; ++i){
    Serial.println(i);
    File f = SPIFFS.open("/mytext.txt", "a");
    if(f){
[...]

PR #8605, opened 2022-06-16
Handle unaligned address in EspClass::flashWrite u8 overload

mcspr:
another approach at fixing #8372
resolves #8588 @fabianoriccardi

instead of changing the branch, just break the flow on a lower level when actually writing things
both unaligned src and dest are placed into buffer. this copies behaviour of RTOS functions that I mentioned in the issue

example sketch seems to work. I just hope I did not break something else while fixing spiffs
*(...and geez, how much unaligned writes it does. no wonder it's slow...)*


fabianoriccardi:
I guessed that this issue could be solved with a good refactor of the whole *spi_flash_write_* methods. Sure I will vote for your pull request, my intention was proposing a *punctual* fix, modifying as little as possible, that the overall refactor is the way to go!

cores/esp8266/Esp.cpp
```diff
@@ -673,7 +673,48 @@ bool EspClass::flashEraseSector(uint32_t sector) {
...
+    }
+
+    // otherwise, write the initial part and continue writing breaking each page interval
+    SpiFlashOpResult result = SPI_FLASH_RESULT_ERR;
+    if ((result = spi_flash_write(offset, data, page_size)) != SPI_FLASH_RESULT_OK) {
+        return result;
+    }
+
+    uint32_t page_offset = PageSize;
+    for (uint32_t page = 0; page < (size - page_size) / PageSize; ++page) {
```
TD-er, line 700: While you're at it optimizing speed, better move the `(size - page_size) / PageSize` part of the for loop call into a const declared variable before calling the for loop.

cores/esp8266/Esp.cpp
```diff
@@ -673,7 +673,48 @@ bool EspClass::flashEraseSector(uint32_t sector) {
...
+        if ((result = spi_flash_write(offset + page_offset, data + (page_offset >> 2), PageSize)) != SPI_FLASH_RESULT_OK) {
+            return result;
+        }
+
+        page_offset += PageSize;
+    }
+
+    if ((offset + page_offset) < (offset + size)) {
```
TD-er, line 708: looks to me like `offset +` can be removed from both sides of the compare?

cores/esp8266/Esp.cpp
```diff
@@ -720,195 +761,127 @@ static SpiFlashOpResult spi_flash_write_puya(uint32_t offset, uint32_t *data, si
...
-    if (!flashReplaceBlock(address, data, pageLeft)) {
-        return false;
-    }
-    offset += pageLeft;
-    sizeLeft -= pageLeft;
-    // We replaced last 4-byte block of the page, now we write the remainder in next page
-    if (!flashReplaceBlock(address + offset, data + offset, sizeLeft)) {
-        return false;
+    while (size > 0) {
+        auto len = std::min(size, sizeof(buf));
```
TD-er, line 834: Maybe have a double check on what is returned by `sizeof(buf)` here?
I always find it tricky to see for myself whether the `sizeof` is returning the size of the array or the size of a pointer.
In this instance, it should be `FLASH_PAGE_SIZE`

mcspr: No point guessing when IDE / language server in any editor will display this b/c it's a compile time value :)
But I suppose this could be `std::size(buf);` instead

TD-er: Well my "code review IDE" (also called "brain") is lacking this feature. ;)  It is only highlighting code which looks like code I have spent numerous hours on to fix in the past where I missed such a mistake.
Apart from that you still need to hover your mouse on those parts of the code in an IDE that does show you what it will do and that was my suggestion, to actually do that ;)

mcspr:
re. does it work with u8 data, we do have a test cases for various offsets
https://github.com/esp8266/Arduino/blob/master/tests/device/test_spi_flash/test_spi_flash.ino

PR adds an additional case, which fails with the current master

cores/esp8266/Esp.cpp
TD-er: I did just now test ESPEasy with the code changes of this PR.
It does seem to work fine.

d-a-v (approved):
device test started & passed

cores/esp8266/Esp.cpp
mcspr: Please do.

fabianoriccardi:
I've tested the new commits and they solve https://github.com/esp8266/Arduino/issues/8372. Hoping they will be integrated soon in the main line.

PR #8606, opened 2022-06-16
Allow to use temporary strings in configTime

### LINKED ISSUE #7056: SNTP: configTime server cannot reference temporary char pointer
### Basic Infos

#### Platform

- Hardware:				ESP8285
- Core Version:			2.6.3 / [63b73bf181bfd495ed8d8c45742ed1a117842220]
- Development Env:		PlatformIO
- Operating System:		Linux

### Settings in IDE

- Defaults

### Problem Description
With this piece of code below:
https://github.com/esp8266/Arduino/blob/63b73bf181bfd495ed8d8c45742ed1a117842220/cores/esp8266/time.cpp#L61
If user does something like:
```c++
const String server = retrieveMyServerSettings();
configTime("UTC0", server.c_str());
```
Because lwip only stores a pointer
https://git.savannah.nongnu.org/cgit/lwip.git/tree/src/apps/sntp/sntp.c?h=STABLE-2_1_x#n876
```
void
sntp_setservername(u8_t idx, const char *server)
{
  LWIP_ASSERT_CORE_LOCKED();
  if (idx < SNTP_MAX_SERVERS) {
    sntp_servers[idx].name = server;
  }
}
```
Pointer becomes garbage after some time.

[Library example is not affected](https://github.com/esp8266/Arduino/blob/63b73bf181bfd495ed8d8c45742ed1a117842220/libraries/esp8266/examples/NTP-TZ-DST/NTP-TZ-DST.ino#L215)

```cpp

#include <ESP8266WiFi.h>
#include <Arduino.h>
#include <sntp.h>
#include <PolledTimeout.h>

static esp8266::polledTimeout::periodicFastMs poll(1000);

void setup() {
    Serial.begin(115200);
    String value("testtesttest");
    configTime("UTC0", value.c_str(), "domainname");
    Serial.printf("before: 0 -> %s, 1 -> %s\n", sntp_getservername(0), sntp_getservername(1));
    WiFi.persistent(false);
    WiFi.mode(WIFI_STA);
    WiFi.begin("blah", "blah");
}

void loop() {
    if (poll) {
        Serial.printf("1 -> %s\n", sntp_getservername(1));
        Serial.flush();
        Serial.printf("0 -> %s\n", sntp_getservername(0));
        Serial.flush();
    }
}
```

### Debug Messages

```
// 2.6.3 even crashes. git manages to stay online

before: testtesttest domainname
1 -> domainname
0 -> ␔��?�:
1 -> domainname
0 -> ␔��?�:
1 -> domainname
0 -> ␔��?�:

Exception (28):
epc1=0x40207c2e epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000
...
```

mcspr:
resolves #7056 

in addition to the existing function, adds an overload accepting String
adds warning in the comment. not sure about older versions, but in ide 2.0 this will be displayed when hovering over the function call

although... there may be a confusion between F() and PSTR(). F() will be loaded into a static string, while PSTR() is const char ptr and uses the original function. at least with DNS, there's is a strlen call which will obviously fail


cores/esp8266/time.cpp
```diff
@@ -208,6 +208,19 @@ void configTime(int timezone_sec, int daylightOffset_sec, const char* server1, c
     sntp_init();
 }
 
+void configTime(int timezone_sec, int daylightOffset_sec, String server1, String server2, String server3)
+{
+    static String servers[3];
+    servers[0] = std::move(server1);
+    servers[1] = std::move(server2);
+    servers[2] = std::move(server3);
+
```
d-a-v, line 217: I suspect there is one copy and one move
- copy: at user calling level
- move: here

It can be opposed to this case where there'd be only one copy:
```cpp
void configTime(int timezone_sec, int daylightOffset_sec, const String& server1, const String& server2, const String& server3)
{
    servers[0] = server1; ...
```


This is more of a question than anything else.
I was used to using `std::move()` only with rvalue references which prevent from passing temporaries (read: my understanding stopped there).
```cpp
void configTime(int timezone_sec, int daylightOffset_sec, String&& server1, String&& server2, String&& server3)
{
    servers[0] = std::move(server1); ...
```
but in this case, user string would be cleared/zeroed after this call.

cores/esp8266/time.cpp
```diff
@@ -228,6 +241,19 @@ void configTime(const char* tz, const char* server1, const char* server2, const
     sntp_init();
 }
 
+void configTime(const char* tz, String server1, String server2, String server3)
+{
+    static String servers[3];
```
d-a-v, line 246: :+1:
I assume these `static` arrays are not linked in when these `configTime()` overloads are not used.

mcspr: In this case, yes.

cores/esp8266/time.cpp
mcspr: depends on how it's called. here it is rvalue from the `String()` pov, so it is moved all the way
```cpp
String getSomeServer();

void configure() {
  configTime("UTC0", getSomeServer());
}
```

this will be a const ref, meaning a copy plus move
```cpp
String blah("best.server.ntp.org");
configTime("UTC0", best);
```

`std::move()` is just a `&&` cast on the original type. in this case, we force a `String::operator=(String&&)` overload, otherwise it would use `String::operator=(const String&)`
since we are making a copy anyway, might as well avoid writing `const String&` and `String&&` overloads

d-a-v (approved):

PR #8613, opened 2022-06-21
Update to toolchain 3.1.0-gcc10.3

earlephilhower:


d-a-v (approved):
LGTM !
Flash occupation for a random ~500KB flash sketch was reduced by ~500 bytes

earlephilhower:
Can you try again?  BearSSL was rebuilt using one of the new optimization settings.

We also need to try the generated code out.  I'm not 100% confident of all the changes at this point.  There were a lot of machine definition changes...

jjsuwa-sys3175:
the patches that seem to have remarkable effects with this update IMHO:
- "Defer storing integer constants into litpool until reload"
until now, early pooling of constants hindered integer optimizations.
- "Add support for sibling call optimization"
  it reduces funccall overhead and stack usage.
- "Improve constant synthesis for both integer and floating-point"
 new synthesis pattern, and early synthesis hindered integer optimizations until now.
- and a couple of mods related to logical AND with constants...

earlephilhower:
Thanks @d-a-v !  I'd suggest we merge this ASAP to get as many user hours on it as possible before a release now that some complicated tests are proven good...

PR #8620, opened 2022-06-27
Update actions

wrt54g:
I updated the GitHub actions to later versions.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -65,7 +65,7 @@ jobs:
         python-version: '3.x'
     - name: Cache Linux toolchain
       id: cache-linux
-      uses: actions/cache@v3.0.2
+      uses: actions/cache@v3.0.4
       with:
         path: ./tools/dist
         key: ${{ runner.os }}-${{ hashFiles('package/package_esp8266com_index.template.json', 'tests/common.sh') }}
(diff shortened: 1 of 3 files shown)


d-a-v (approved):
Thanks,
I see no objection but I still wonder what's hiding behind such changes

mcspr:
Why not v3 for checkout?
Can't we use dependabot to avoid human-dependabot dependency?
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

wrt54g:
@d-a-v When will this get merged?
@mcspr This PR updates all actions/checkout to v3 already.

.github/workflows/pull-request.yml
```diff
@@ -22,15 +22,15 @@ jobs:
       matrix:
         chunk: [0, 1, 2, 3, 4, 5, 6, 7]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3.0.2
```
mcspr, line 25: ^ this and cache. why not v3?

wrt54g: @mcspr Because 3.0.0 is an earlier version than 3.0.2?

mcspr: no? :)
action tags like v3 are updated during the action lifetime, they always point to the latest version available

mcspr: https://github.com/actions/cache/tags

wrt54g:
Suspended by #8624

PR #8622, opened 2022-06-27
Use current netif address in NetBIOS response

### LINKED ISSUE #8139: ESP8266NetBIOS library only works in wifi STA mode, and not in wifi AP

mcspr:
fix for NetBIOS only working if WIFI in STA mode and not AP mode

resolve #8139
resolve #8140


mcspr:
yet, wireshark seems to think we have a malformed packet when asking for name via ip
not related to the issue at hand thoug

d-a-v (approved):

PR #8623, opened 2022-06-28
Handle recent python versions in device tests

mcspr:
Update soon-to-be deprecated code in the runner.py
https://docs.python.org/3/library/imp.html is deprecated since 3.4, will be removed in 3.12
https://docs.python.org/3/library/configparser.html readfp is replaced with read_file since 3.2, will be removed in 3.12

Use venv instead of virtualenv. We don't really use any of the extended features, and might as well  simplify installation requirements.
virtualenv/bin/python can execute runner.py inside of venv, no need to activate beforehand.


d-a-v (approved):
Thanks !
Some device test-on-device and some test-on-host passed without python warnings.

PR #8625, opened 2022-07-02
Longer delays for Ticker and some internal updates

### LINKED ISSUE #8066: Ticker.h - Too-large timeouts should generate an error, not silently fail (was: big float and unt32_t numbers)
#### Platform

- Hardware:					[ESP-12]
- Core Version:			[2.7.4] & [3.0.0]

### Settings in IDE

- Module:  			[Generic ESP8266]
- Flash Mode:			[qio]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Higher Bandwidth]
- Reset Method:		[nodemcu]
- Flash Frequency:	[80Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[OTA|SERIAL]
- Upload Speed:   [115200] 

### Problem Description

Any Ticker tasks fire OK if time is short (tested up to 60 seconds (60 000ms)). However, when I try to control the brightness after sunset, the time gets big (for example 44944 sec (44 944 000ms)) & the same task no longer start. The code below is not exactly what I have in my project, but close.

```cpp

#include <Arduino.h>
#include <Ticker.h>

Ticker ticker_autoBrightness;

uint32_t timeSec = 44079; //12h:14m:39s

void setup() {
pinMode(2, OUTPUT);

task_clockDayBrightness();
}

void loop() {
//empty
}

void task_clockDayBrightness()
{
  digitalWrite(2, LOW); //turn LED on

  ticker_autoBrightness.once_scheduled(timeSec, task_clockNightBrightness);
}

void task_clockNightBrightness()
{
  digitalWrite(2, HIGH); //turn LED off

  ticker_autoBrightness.once_scheduled(timeSec, task_clockDayBrightness);
}

mcspr:
resolves #8066

Adds max duration check. In case it is over SDK limit, enable 'repeat'ing timer with a duration proportional to the original one and count until it executes N times, only then run the callback.
Code with durations less than that executes as usual. Original proposal was to not create anything or create some kind of error state... which seems counter-productive to not help out with this pretty solvable use-case.

Three additional updates, while refactoring the class
- Stronger types for internal time management using `std::chrono::duration`. Works the same, `std::chrono::duration` handles seconds <-> milliseconds conversion.
- `::detach()` 'once' timer when it finishes. Fixes (unintentional?) side-effect that we remain `::active()`. Plus, this destroys any lambda-bound variables that will persist with the Ticker object. And, since we can't re-arm with the same function (`Ticker::attach_ms(uint32_t just_the_time)` and etc.)
- `std::variant` aka union for internal callback storage (kind-of similar to #6918). Instead of having two separate code paths, **always** attach our static function and dispatch using type info. Also helps with the issue described above, since it will call `std::function` dtor when ptr + arg is attached instead of doing nothing.

### DIFF
--- a/libraries/Ticker/src/Ticker.cpp
+++ b/libraries/Ticker/src/Ticker.cpp
@@ -46,7 +46,7 @@ void Ticker::_attach(Ticker::Milliseconds milliseconds, bool repeat)
     _tick.repeat = repeat;
 
     if (milliseconds > DurationMax) {
-        _tick.total += 1;
+        _tick.total = 1;
         while (milliseconds > DurationMax) {
             _tick.total *= 2;
             milliseconds /= 2;
@@ -62,13 +62,7 @@ void Ticker::detach()
     if (_timer) {
         os_timer_disarm(_timer);
         _timer = nullptr;
-
-        _tick = callback_tick_t{
-            .total = 0,
-            .count = 0,
-            .repeat = false,
-        };
-
+        _tick = callback_tick_t{};
         _callback = std::monostate{};
     }
 }
--- a/libraries/Ticker/src/Ticker.cpp
+++ b/libraries/Ticker/src/Ticker.cpp
@@ -23,49 +23,84 @@
 #include "eagle_soc.h"
 #include "osapi.h"
 
+#include <Arduino.h>
 #include "Ticker.h"
 
-Ticker::Ticker()
-    : _timer(nullptr) {}
-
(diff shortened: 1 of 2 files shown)


mcspr:
Speaking of `void(*)(char)` to `void(*)(void*)` conversion... Does it actually work correctly for arguments *less* than 4 bytes? Don't we have some undefined behaviour here?

PR #8626, opened 2022-07-04
Fix hostname management

### LINKED ISSUE #8143: Change hostname with ethernet W5500lwIP library
Hello, turning this into an issue since i could not find a way to get this working (it is, with a wifi connection)

### Basic Infos

- [x] I have tested that the issue is present in current master branch (aka latest git). (using later version through platformio)
- [ ] If there is a stack dump, I have decoded it. - No stack dump

#### Platform

- Hardware:				Esp8266
- Core Version:			latest core through platformio
- Operating System:		Windows

### Settings in IDE

- Module:  			Wemos D1 mini
- Flash Mode:			[qio|dio|other]?
- Flash Size: 		        4MB
- lwip Variant:                  v2 Lower Memory (platformio default)
- Reset Method:		[ck|nodemcu]?
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL and OTA

### Problem Description
Using a W5500 module with new lwIP library included in core 3.0.0
I can't get a hostname to be assigned to the device. 
Not a problem assigning a hostname with a wireless connection, but with a wired one the setHostname is always failing, everywhere i put it in the script, and obviously my router sees the device with just a "-" as hostname.
This is not a mDNS issue, i'm not looking for a name.local, i'm trying to give a name to the device that my router can resolve, like i'm doing with all my other wifi esp8266 devices.
As suggested by @d-a-v , i also tried explicitly enabling the wifi with enableWiFiAtBootTime(), with and without the WiFi.mode(WIFI_OFF)
It keeps failing and i don't know how to debug this further
Thanks

### Discussed in https://github.com/esp8266/Arduino/discussions/8129

<div type='discussions-op-text'>

<sup>Originally posted by **MassiPi** June 16, 2021</sup>
Hello,
i searched both the open issues and the pull requests for setHostname() and i could not find anything, please excuse me if i lost something :)
i'm trying to change the hostname with a wired ethernet connection using the "new" W5500lwIP library.
I'm trying to use both the eth.setHostname() and the WiFi.setHostname() functions (that should be the very same and both based on the wifi_station_set_hostname())
I tried to change the hostname in various places, getting no result at all.
This is the sample code:
```cpp
#include <ESP8266WiFi.h>
#include <SPI.h>
#include <W5500lwIP.h>

byte mac[] = {0x00, 0xAA, 0xBB, 0xCC, 0xDE, 0x02};

const char hostname[] = "NAMETEST";

Wiznet5500lwIP eth(D1);


void setup() {
  WiFi.mode(WIFI_OFF);
  // Open serial communications and wait for port to open:
  Serial.begin(115200);
  while (!Serial) {
    ; // wait for serial port to connect. Needed for native USB port only
  }

  SPI.begin();
  SPI.setBitOrder(MSBFIRST);
  SPI.setDataMode(SPI_MODE0);
  SPI.setFrequency(4000000);

  Serial.println("First: before setDefault");
  Serial.println("eth.setHostname()");
  eth.setHostname(hostname);
  Serial.println("WiFi.setHostname()");
  WiFi.setHostname(hostname);

  eth.setDefault(); // use ethernet for default route

  Serial.println("Second: before begin");
  Serial.println("eth.setHostname()");
  eth.setHostname(hostname);
  Serial.println("WiFi.setHostname()");
  WiFi.setHostname(hostname);

  int present = eth.begin(mac);
  if (!present) {
    Serial.println("no ethernet hardware present");
    while(1);
  }

  Serial.println("Third: before connected");
  Serial.println("eth.setHostname()");
  eth.setHostname(hostname);
  Serial.println("WiFi.setHostname()");
  WiFi.setHostname(hostname);

  Serial.print("connecting ethernet");
  while (!eth.connected()) {
    Serial.print(".");
    delay(100);
  }
  Serial.println("Connected");

  Serial.println("Fourth: after connection");
  Serial.println("eth.setHostname()");
  eth.setHostname(hostname);
  Serial.println("WiFi.setHostname()");
  WiFi.setHostname(hostname);

}

void loop() {

  Serial.println("Fifth: in loop");
  Serial.println("eth.setHostname()");
  eth.setHostname(hostname);
  Serial.println("WiFi.setHostname()");
  WiFi.setHostname(hostname);

  delay(5000);
}
```

This is what i get:
```
First: before setDefault
eth.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
WiFi.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
Second: before begin
eth.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
WiFi.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
Third: before connected
eth.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
WiFi.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
connecting ethernet.Connected
Fourth: after connection
eth.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
WiFi.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
Fifth: in loop
eth.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
WiFi.setHostname()
WiFi.hostname(NAMETEST): wifi_station_set_hostname() failed
```

what am i missing?
is there a way to further debug this?
thanks!</div>

d-a-v:
This proposal addresses two issues
- Ethernet hostname change was not working as intended because hostname management relies on wifi_**station**_set_hostname() and STA can be down when ethernet is used,
- Since WiFi is off at boot time, for the same reason, hostname can't be changed until STA is up.

For both issue, solution is to play with hostname location in ram, reserved by SDK, and filled with `ESP-baffed` when WiFi boots but which can be now overwritten back when user had already set it to something.

fixes #8143

### DIFF
--- a/cores/esp8266/LwipIntf.cpp
+++ b/cores/esp8266/LwipIntf.cpp
@@ -16,6 +16,9 @@ extern "C"
 #include "debug.h"
 #include "LwipIntf.h"
 
+extern "C" char* wifi_station_hostname; // sdk's hostname location
+extern "C" char* wifi_station_default_hostname; // sdk's hostname location
+
 // args      | esp order    arduino order
 // ----      + ---------    -------------
 // local_ip  | local_ip     local_ip
@@ -66,7 +69,7 @@ bool LwipIntf::ipAddressReorder(const IPAddress& local_ip, const IPAddress& arg1
 */
 String LwipIntf::hostname(void)
 {
-    return wifi_station_get_hostname();
+    return wifi_station_hostname; //wifi_station_get_hostname();
 }
 
 /**
@@ -75,7 +78,7 @@ String LwipIntf::hostname(void)
 */
 const char* LwipIntf::getHostname(void)
 {
-    return wifi_station_get_hostname();
+    return wifi_station_hostname; //wifi_station_get_hostname();
 }
 
 /**
@@ -136,20 +139,26 @@ bool LwipIntf::hostname(const char* aHostname)
         DEBUGV("hostname '%s' is not compliant with RFC952\n", aHostname);
     }
 
+#if 0
     bool ret = wifi_station_set_hostname(aHostname);
     if (!ret)
     {
         DEBUGV("WiFi.hostname(%s): wifi_station_set_hostname() failed\n", aHostname);
         return false;
     }
+#else
+    bool ret = true;
 
+    strcpy(wifi_station_hostname, aHostname);
+    strcpy(wifi_station_default_hostname, aHostname);
+#endif
     // now we should inform dhcp server for this change, using lwip_renew()
     // looping through all existing interface
     // harmless for AP, also compatible with ethernet adapters (to come)
     for (netif* intf = netif_list; intf; intf = intf->next)
     {
         // unconditionally update all known interfaces
-        intf->hostname = wifi_station_get_hostname();
+        intf->hostname = wifi_station_hostname; //wifi_station_get_hostname();
 
         if (netif_dhcp_data(intf) != nullptr)
         {
--- a/cores/esp8266/LwipIntf.cpp
+++ b/cores/esp8266/LwipIntf.cpp
@@ -16,8 +16,16 @@ extern "C"
 #include "debug.h"
 #include "LwipIntf.h"
 
-extern "C" char* wifi_station_hostname; // sdk's hostname location
-extern "C" char* wifi_station_default_hostname; // sdk's hostname location
+// wifi_station_hostname is SDK's station(=global) hostname location
+// - It is never nullptr but wifi_station_get_hostname()
+//   can return nullptr when STA is down
+// - Because WiFi is started in off mode at boot time,
+//   wifi_station_set/get_hostname() is now no more used
+//   because setting hostname firt does not work anymore
+// - wifi_station_hostname is overwritten by SDK when wifi is
+//   woken up in WiFi::mode()
+//
+extern "C" char* wifi_station_hostname;
 
 // args      | esp order    arduino order
 // ----      + ---------    -------------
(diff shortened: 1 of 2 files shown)


mcspr:
Running the example from Matrix, everything works correctly.

Is the idea to allow to set netif->hostname later on, or it will stay global hostname setting?
And I guess, replacing SDK approach by removing wifi_station_... func from the SDK blobs, and using our own func setting netif->hostname?

tests/host/common/MockEsp.cpp
```diff
@@ -39,6 +39,8 @@
 #include <user_interface.h>
 struct rst_info resetInfo;
 
+char wifi_station_hostname [33]; // exists in nonosdk
+
```
mcspr, line 43: common.cpp still has `char wifi_station_get_hostname_str[128]`
mock should just use this array instead?

d-a-v:
> Is the idea to allow to set netif->hostname later on, or it will stay global hostname setting?

Both are possible, I think keeping a common global name is better.
Like when you log in a computer with several network cards (ethernet, wifi) or a router: It's much easier when a single name designates a single place. 

> And I guess, replacing SDK approach by removing wifi_station_... func from the SDK blobs, and using our own func setting netif->hostname?

This is an interesting idea. For now I believe a sketch using the arduino API won't link these functions in.
We can open a new issue for core v4 about that.

> Running the example from Matrix, everything works correctly.

Which matrix is it ?

mcspr:
> Which matrix is it ?

Capital M one! Where you said it would be enough to check ./libraries/ESP8266WiFi/examples/IPv6/IPv6.ino example :shrug:

> This is an interesting idea. For now I believe a sketch using the arduino API won't link these functions in.
We can open a new issue for core v4 about that.

True, but it's the same issue as with dhcps funcs - NONOS SDK api stops working / interferes with the existing code. SDK part could also be `--weaken-symbol=symbolname`'d with the objcopy script, and we could have `{get,set}hostname()` doing this global setting. I will stop theorizing now as well, until the ethernet part works as intended on all fronts :)

PR #8627, opened 2022-07-05
emulation on host: fix incorrect lwIP DNS implementation

d-a-v:
- fix mismatched header of lwIP's `dns_getserver()` implementation
- correctly initialize emulated netif0
- make ipv6 example unfail (even if emulation on host is ipv4 only)

### DIFF
--- a/tests/host/common/MocklwIP.cpp
+++ b/tests/host/common/MocklwIP.cpp
@@ -3,6 +3,9 @@
 
 #include "MocklwIP.h"
 
+#include <lwip/dns.h>
+#include <arpa/inet.h>
+
 esp8266::AddressListImplementation::AddressList addrList;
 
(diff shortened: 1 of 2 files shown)


tests/host/common/MocklwIP.cpp
```diff
@@ -57,4 +60,18 @@ extern "C"
...
+    {
+        (void)numdns;
+        (void)dnsserver;
+    }
+
+    const ip_addr_t* dns_getserver(u8_t numdns)
+    {
+        (void)numdns;
+        static ip_addr_t addr;
+        addr.addr = htonl(0x7f000001);  // 127.0.0.1
```
mcspr, line 73: irrc lwip has macros like IP(A,B,C,D)?

PR #8630, opened 2022-07-09
add missing license headers

d-a-v:


earlephilhower:
Man, clang beautifier is picky about empty newlines!

PR #8634, opened 2022-07-13
Board ID as build flag

### LINKED ISSUE #7759: mDNS discovery with IDE cannot detect the board name
- x[ ] I have searched the issue tracker for a similar issue.

#### Platform

- Hardware:					ESP-12
- Core Version:			[latest git ]

### Settings in IDE

- Module:  			[Generic ESP8266 ]
- Flash Mode:			[dio]
- Flash Size: 		[4MB]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[OTA]
- Upload Speed:   [115200|] 

### Problem Description
OTA description is incomplete - variant is missing  in description in  Arduino OTA menu unlike for ESP32

image
it is due to property board which should contain the variant name, but ESP8266 put another string 
image
 string is put here (thanks @tobozo to find out)
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266mDNS/src/LEAmDNS.cpp#L1278
and also https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266mDNS/src/LEAmDNS2Host.cpp#L792
So arduino IDE cannot match the variant with one present in board.txt

if I hardcode "generic" for the variant on board property - the result is correct

image
image

mcspr:
Use `{_id}` in mDNS response instead of the `{build.board}`, adds extra build flag `-DARDUINO_BOARD_ID="{_id}"`

For example
```
$ arduino-cli.exe board list
Port          Protocol Type              Board Name                FQBN                       Core
xxx.xxx.x.xxx network  Network Port      LOLIN(WEMOS) D1 R2 & mini esp8266com:esp8266:d1_mini esp8266com:esp8266
```

resolve #7759


mcspr:
...or, just replace the `{build.board}` with `{_id}`. without adding yet another flag
just not sure whether changing its value matters anywhere but the IDE discovery, or if any user code might expect / use it in any way

d-a-v (approved):

PR #8642, opened 2022-07-18
update example certs

d-a-v:
This is normally done during release process,
but changing them now because variable names are also updated.

Examples passed on WiFi hw.


mcspr:
looking at host steps, verification example for gitlab was intended to fail? key is sni.cloudflaressl.com, but it connects to gitlab.com

PR #8645, opened 2022-07-20
make WiFi/Ethernet interface compatible with Arduino Ethernet API

d-a-v:
While helping a colleague with ethernet on a portenta H7,
I noticed that it was poorly implemented with respect to Official Ethernet Arduino API,
and it does not help foreigners to understand official examples.
Before trying to fix there, looking at esp8266's own backyard first.

Some basic legacy examples are also imported with minimal changes.
They are ethernet examples but they could also run on WiFi.


JAndrassy:
`using EthernetServer = ArduinoWiFiServer;` I like that

everywhere I recommend to use SDK NTP functions on esp8266. so maybe don't add the NTP example. even it is a good example for UDP same arguments against it apply as for WebClient example.

d-a-v:
> `using EthernetServer = ArduinoWiFiServer;` I like that

And thank you for that :)

> everywhere I recommend to use SDK NTP functions on esp8266. so maybe don't add the NTP example. even it is a good example for UDP same arguments against it apply as for WebClient example.

You are totally right. I'll replace it with another UDP example. Suggestions are welcome (even for TCP examples).

mcspr:
Shouldn't ethernet thing go to the ethernet class? I have noticed `using Ethernet = WiFi;`, but now we have yet another set of `enum` tokens in a global namespace.

d-a-v:
> Shouldn't ethernet thing go to the ethernet class? I have noticed `using Ethernet = WiFi;`, but now we have yet another set of `enum` tokens in a global namespace.

Global enums are now moved to the Ethernet class.
These arduino definitions were allowing maximum compatibility with legacy sketch when using WiFi as network.
But I get it ;)

NTP example is removed and replaced by UDP server example.

libraries/lwIP_Ethernet/examples/LegacyDhcpAddressPrinter/LegacyDhcpAddressPrinter.ino
```diff
@@ -0,0 +1,101 @@
...
+
+    case 4:
+      // rebind success
+      Serial.println("Rebind success");
+      // print your local IP address:
+      Serial.print("My IP address: ");
+      Serial.println(Ethernet.localIP());
+      break;
+
+    default:
```
mcspr, line 97: no need

libraries/lwIP_Ethernet/examples/LegacyUDPSendReceiveString/LegacyUDPSendReceiveString.ino
```diff
@@ -0,0 +1,153 @@
...
+
+  // Open serial communications and wait for port to open:
+  Serial.begin(9600);
+  while (!Serial) {
+    ;  // wait for serial port to connect. Needed for native USB port only
+  }
+
+  // start the Ethernet
+  // Ethernet.begin(mac, ip);
+  if (Ethernet.begin(mac) == 0) {
```
mcspr, line 56:
```suggestion
  if (!Ethernet.begin(mac)) {
```

mcspr (approved):
lgtm then :)
do we care about code in these legacy examples? like, `Ethernet.begin(...) == 0` where it is `bool`, or switch-case on integers instead of `DHCP_CHECK...`, or the pre-connection link checks that are slightly different between .ino setups

d-a-v:
> do we care

The goal is to keep them the closest possible to the original ethernet examples.

```diff
--- ../../../Ethernet/examples/ChatServer/ChatServer.ino	2022-07-18 22:17:20.550672427 +0200
+++ LegacyChatServer.ino	2022-07-20 03:12:30.293107744 +0200
@@ -7,7 +7,7 @@
  Using an Arduino Wiznet Ethernet shield.
 
  Circuit:
- * Ethernet shield attached to pins 10, 11, 12, 13
+ * Ethernet Wiznet5500/Wiznet5100/ENC28J60 on esp8266
 
  created 18 Dec 2009
  by David A. Mellis
@@ -16,14 +16,22 @@
 
  */
 
-#include <SPI.h>
-#include <Ethernet.h>
+
+// specific to esp8266 w/lwIP
+#include <EthernetCompat.h>
+ArduinoWiznet5500lwIP Ethernet(/*SS*/ 16);  // <== adapt to your hardware
+// ArduinoWiznet5100lwIP Ethernet(/*SS*/ 16);  // <== adapt to your hardware
+// ArduinoENC28J60lwIP Ethernet(/*SS*/ 16);  // <== adapt to your hardware
 
 // Enter a MAC address and IP address for your controller below.
 // The IP address will be dependent on your local network.
 // gateway and subnet are optional:
-byte mac[] = {
-  0xDE, 0xAD, 0xBE, 0xEF, 0xFE, 0xED };
+byte notNeededButAllowed_mac[] = {
+  0xDE, 0xAD, 0xBE, 0xEF, 0xFE, 0xED
+};
+byte* mac = nullptr;  // automatic mac
 IPAddress ip(192, 168, 1, 177);
 IPAddress myDns(192, 168, 1, 1);
 IPAddress gateway(192, 168, 1, 1);
@@ -36,12 +44,13 @@
 
   // Ethernet.init(33);  // ESP32 with Adafruit Featherwing Ethernet
+  //                     // esp8266 w/lwIP: SS set in Ethernet constructor
 
   // initialize the ethernet device
   Ethernet.begin(mac, ip, myDns, gateway, subnet);
```

PR #8653, opened 2022-07-30
Updater lifetime callbacks

mcspr:
follow-up of https://github.com/esp8266/Arduino/pull/8598
we removed some WiFiClient stopAll(s) from updaters, make some lifetime callbacks similar to ArduinoOTA ones so it can be handled by the user

arduino ota part is not 'proxied' to updater counterparts, since the idea there is to track things specific to arduino ota and not anything global

not tested


cores/esp8266/Updater.h
```diff
@@ -51,7 +51,9 @@ class UpdaterVerifyClass {
 
 class UpdaterClass {
   public:
-    typedef std::function<void(size_t, size_t)> THandlerFunction_Progress;
+    using THandlerFunction_Progress = std::function<void(size_t, size_t)>;
+    using THandlerFunction_Error = std::function<void(uint8_t)>;
+    using THandlerFunction = std::function<void()>;
```
mcspr, line 56: While ArduinoOTA callbacks arent doing that... I wonder if the 1st argument should be the instance / `this`
But, I suppose Arduino way is to always assume there is a global object, somewhere, doing the thing you want and there's even no need to capture anything. Just feels weird.

Issue #8657, opened 2022-08-14
"Heap init code improvements" doesn't boot on ESP-01 with PUYA flash


avian2:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-01]
- Core Version:			[313b3c07ecccbe6fee24aa9fa447c4aed16ca499]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[dio]
- Flash Size: 		[1MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

Any program compiled and uploaded using the latest git master does not run at all. After a successful compile and upload Arduino serial monitor only shows garbage (using a scope on the serial line shows data with a non-standard baudrate that I can't decode). The LED is constantly on. 

Simplest sketch that reproduces this is the "Blink" example. However the result is the same regardless of what program is uploaded.

Using git bisect from a known-good release (2.4.2) shows that the commit that introduced this issue is:

7356cd1ef13c7a7da4631be01060f6f40806fa23
Heap init code improvements and updates (#8458)

Reverting just this commit on the latest git master fixes the problem.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch
```cpp
void setup() {
  pinMode(LED_BUILTIN, OUTPUT);
}

void loop() {
  digitalWrite(LED_BUILTIN, HIGH);
  delay(1000);
  digitalWrite(LED_BUILTIN, LOW);
  delay(1000);
}
```

### Debug Messages

```
Picked up JAVA_TOOL_OPTIONS: 
. Variables and constants in RAM (global, static), used 27984 / 80192 bytes (34%)
║   SEGMENT  BYTES    DESCRIPTION
[...]

PR #8658, opened 2022-08-20
Fix pre-SDK Cache_Read_Enable for PUYA flash

mhightower83:
This resolves the exception 0 issue with PUYA flash during the pre-SDK context. This Issue showed when using flash/ICACHE for `umm_init` and/or using `HWDT` Stack Dump .

The `SPI_CS_SETUP` parameter has been observed set in RTOS SDK and NONOS SDK as part of flash init/configuration. It may be necessary for some flash chips to perform correctly. Turning on and leaving it on should be okay.

`phy_get_bb_evm` is the key function, called from `fix_cache_bug` in the NONOS SDK. This addition resolves the exception 0 issue with PUYA Flash when early `Cache_Read_Enable` is used.

Changed `umm_init` back to default in IRAM. Define `UMM_INIT_USE_ICACHE` to move `umm_init` to flash and free up IRAM. For the non-debug build, it frees 136 bytes of IRAM while increasing IROM by 224 bytes.

Resolves https://github.com/esp8266/Arduino/issues/8657#issue-1338167822 (has PUYA flash)

Unable to recreate the problem with the PUYA P25Q80H flash chip on ESP-01S.

Updated to summarize current PR.


d-a-v:
Per [testing](https://github.com/esp8266/Arduino/issues/8657#issuecomment-1234543293), is this still a draft ?

mhightower83:
The approach has been tested; however, this PR with the approach has not been verified. Otherwise, it is ready.

Also for umm_init, do you agree with the change back to using IRAM as default and ICACHE requiring the override?

PR #8661, opened 2022-08-23
Stack decoder script

### LINKED ISSUE #6310: Add a CLI version of ESPExceptionDecoder
Seems like Platform.IO doesn't have JAVA plugins, so running the stack dump decoder is not an option for them.

It would be nice to have a CLI version of the exception decoder that would take the ELF file and stack dump as input and produce the same output as the java plugin.

mcspr:
*subj*
Needs toolchain path, elf path, and the postmortem either as a file or piped into stdin
Not as colorful as java decoder, though. Like monitor filter, tries to 'update' already existing data with our knowledge of addresses and source locations.

Based on the code from
* https://github.com/me-no-dev/EspExceptionDecoder/blob/master/src/EspExceptionDecoder.java
* https://github.com/platformio/platform-espressif8266/blob/develop/monitor/filter_exception_decoder.py

Defaults to addr2line, which seems to be working just fine.
Optional GDB `info` output
(could be `list`, but I hope if there's an easier way than to parse it yet again like java version does)

resolves #6310 , but probably needs some mention in the documentation


mcspr:
```cpp
#include <Arduino.h>

int* somewhere { nullptr };

void setup() {
    delay(5000);
    Serial.begin(115200);
    Serial.printf("%d\n", *somewhere);
}

void loop() {
}
```
```

Exception (28):
epc1=0x4020105f epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffe00 end: 3fffffd0 offset: 0190
3fffff90:  0001c200 0000001c 00000000 402018d9
3fffffa0:  3fffdad0 00000000 3ffee4bc 40201055
3fffffb0:  feefeffe feefeffe 3ffee510 40201954
3fffffc0:  feefeffe feefeffe 3ffe85d8 40100c39
<<<stack<<<

```
```shell
> decoder.py --toolchain-path ~/.platformio/packages/toolchain-xtensa/ --elf-path .pio/build/d1_mini/firmware.elf stack.txt

Exception (28) - LoadProhibited: A load referenced a page mapped with an attribute that does not permit loads
epc1=0x4020105f: setup at /home/runner/dev/arduino8661/src/main.cpp:8


0x402018d9: esp_delay at /home/runner/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/core_esp8266_main.cpp:158
0x40201055: setup at /home/runner/dev/arduino8661/src/main.cpp:8
0x40201954: loop_wrapper() at /home/runner/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/core_esp8266_main.cpp:244
0x40100c39: cont_wrapper at ??:?
```

PR #8663, opened 2022-08-25
[String] Fix float assignment using wrong type

TD-er:
Using 3.0.2 assigning a float to a string resulted in undefined behaviour.

```c++
String str;
float myfloat[12][4];
myfloat[0][0] = 1.0f;
str = myfloat[0][0];
```

`str` does contain a single char.

Not sure if a 2D float array is needed to reproduce it, but that's how it was used in my project.
These were part of a larger struct, so maybe the assignment should be done in a separate function to make sure the compiler is not optimizing the code too much.

Worked fine in 2.7.4
Looks like it might have been introduced here: https://github.com/esp8266/Arduino/pull/8526  @mcspr 

Change is mainly to enforce `explicit` behavior when calling the `toString` static functions by using const reference as their argument.

Also added `bool operator ==(const __FlashStringHelper *s) const` and its counter part, which saved roughly 3k in build size on my project ESPEasy.

### DIFF
--- a/cores/esp8266/WString.cpp
+++ b/cores/esp8266/WString.cpp
@@ -37,7 +37,7 @@
 /*  Conversion helpers                       */
 /*********************************************/
 
-static String toString(unsigned char value, unsigned char base) {
+static String toString(const unsigned char& value, unsigned char base) {
     String out;
 
     char buf[1 + std::numeric_limits<unsigned char>::digits];
@@ -46,7 +46,7 @@ static String toString(unsigned char value, unsigned char base) {
     return out;
 }
 
-static String toString(int value, unsigned char base) {
+static String toString(const int& value, unsigned char base) {
     String out;
 
     char buf[2 + std::numeric_limits<int>::digits];
@@ -55,7 +55,7 @@ static String toString(int value, unsigned char base) {
     return out;
 }
 
-static String toString(unsigned int value, unsigned char base) {
+static String toString(const unsigned int& value, unsigned char base) {
     String out;
 
     char buf[1 + std::numeric_limits<unsigned int>::digits];
@@ -64,7 +64,7 @@ static String toString(unsigned int value, unsigned char base) {
     return out;
 }
 
-static String toString(long value, unsigned char base) {
+static String toString(const long& value, unsigned char base) {
     String out;
 
     char buf[2 + std::numeric_limits<long>::digits];
@@ -73,7 +73,7 @@ static String toString(long value, unsigned char base) {
     return out;
 }
 
-static String toString(unsigned long value, unsigned char base) {
+static String toString(const unsigned long& value, unsigned char base) {
     String out;
 
     char buf[1 + std::numeric_limits<unsigned long>::digits];
@@ -84,7 +84,7 @@ static String toString(unsigned long value, unsigned char base) {
 
 // TODO: {u,}lltoa don't guarantee that the buffer is usable directly, one should always use the returned pointer
 
-static String toString(long long value, unsigned char base) {
+static String toString(const long long& value, unsigned char base) {
     String out;
 
     char buf[2 + std::numeric_limits<long long>::digits];
@@ -93,7 +93,7 @@ static String toString(long long value, unsigned char base) {
     return out;
 }
 
-static String toString(unsigned long long value, unsigned char base) {
+static String toString(const unsigned long long& value, unsigned char base) {
     String out;
 
     char buf[1 + std::numeric_limits<unsigned long long>::digits];
@@ -102,7 +102,7 @@ static String toString(unsigned long long value, unsigned char base) {
     return out;
 }
 
-static String toString(double value, unsigned char decimalPlaces) {
+static String toString(const double& value, unsigned char decimalPlaces) {
     String out;
 
     char buf[33];
@@ -111,7 +111,7 @@ static String toString(double value, unsigned char decimalPlaces) {
     return out;
 }
 
-static String toString(float value, unsigned char decimalPlaces) {
+static String toString(const float& value, unsigned char decimalPlaces) {
     return toString(static_cast<double>(value), decimalPlaces);
 }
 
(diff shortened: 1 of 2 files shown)


TD-er:
Hmm @tonhuisman mentioned to me that the operator== for flash strings was removed in another commit, I started looking for it in the history.

It was indeed removed here: https://github.com/esp8266/Arduino/pull/8569
Does it still need to be removed?

mcspr:
You sure about the source of the issue? What about a device / host test? Notice the CI suite and the `tests/` dir related to strings (...and let me check the issue as well :)
FlashStringHelper was intentionally removed, iirc pointer overload is a source of trouble. And pretty sure you could resort to using const char as-is anyway.

mcspr:
> Using 3.0.2

...and are these tests done on git installation?

TD-er:
> > Using 3.0.2
> 
> ...and are these tests done on git installation?

Used this in the PlatformIO config:
```
[core_3_0_2]
extends                   = esp82xx_3_0_x
platform                  = espressif8266@3.2.0
platform_packages = 
    platformio/framework-arduinoespressif8266 @ https://github.com/esp8266/Arduino.git
    mcspr/toolchain-xtensa @ ~5.100300.211127
build_flags               = ${esp82xx_3_0_x.build_flags}
                            -DPIO_FRAMEWORK_ARDUINO_ESPRESSIF_SDK3
lib_ignore                = ${esp82xx_defaults.lib_ignore}
                            IRremoteESP8266
                            HeatpumpIR
                            LittleFS(esp8266)
                            ServoESP32
                            TinyWireM
```

I did for sure a clean build.
@tonhuisman did test it on several builds and also on the build I made with these changes.

This is the section which was our use case:
https://github.com/letscontrolit/ESPEasy/blob/ab9fa55bc9e7d5b457e69a685b9d4b243e3ee341/src/_P021_Level.ino#L127-L136
Line 132, where `string` got assigned a float.

N.B. the operator== doesn't have anything to do with this fix, it was just added as way to reduce build size, but I guess if it does cause issues, I will remove it and go over my code to replace all == and != compares with flash strings to the equals function.

mcspr:
> I did for sure a clean build.

With the `toString()` changes included in the latest commits? PIO does not keep latest branch HEAD automatically. That PR fixed the integer coercion like here, where it was incorrectly assumed to be a `char` type when it fits.

> This is the section which was our use case:
https://github.com/letscontrolit/ESPEasy/blob/ab9fa55bc9e7d5b457e69a685b9d4b243e3ee341/src/_P021_Level.ino#L127-L136
> Line 132, where string got assigned a float.
>
> By explicitly creating a string and assigning it to the string object, this issue was fixed in ESPEasy, but of course then we don't have a clue where else it may cause issues.

The idea is to have some small test code we could run without the whole ESPeasy becoming a test case

TD-er:
> With the `toString()` changes included in the latest commits?

I did edit the WString.cpp/.h files in my local branch of this repo and then copied those 2 files to the tree in the .pio/packages folder.
So it was for sure using the same files as I have here in the PR.

> The idea is to have some small test code we could run without the whole ESPeasy becoming a test case

Will try to make a suitable test case later this evening.

mcspr:
Why I am asking, I don't see any issues with overload itself (that this PR proposes to change)
Also note this will break *some* libs that use bit-field members and serialize them won't work via references.

PIO repo for me is in `~/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910`. No need to manually copy anything, just 'checkout' / 'switch' into the master branch and try to build some small project. e.g.
```cpp
#include <Arduino.h>

void setup() {
    Serial.begin(115200);
}

void loop() {
    String a(1.0f);
    printf("String ctor -> \"%s\"\n", a.c_str());

    String b;
    float num = 1.0f;
    b = num;
    printf("String assignment -> \"%s\"\n", b.c_str());

    delay(1000);
}
```
```
String ctor -> "1.00"
String assignment -> "1.00"
```

TD-er:
OK, made a quick MCVE to show what's wrong.
Also included the PIO config so you can see what exact env is used.
See: https://github.com/TD-er/MCVE_ESPxx/tree/bug/ESP8266_Arduino_8663_Str_assignment

TD-er:
OK, I did check the content of WString.h/.cpp and indeed these are not using the latest versions.
I did manually update the WString files with the current ones on GitHub and it works with your code.
But this leaves the question, how to get the latest code from this repo while using PlatformIO?
Apparently I don't now.

mcspr:
The idea here is to use `framework-arduinoespressif8266 @ git+https://...` package spec so the original version gets 'substituted' with the a different one
https://github.com/platformio/platform-espressif8266/blob/d2f3b915a416c300e521cb76149a40c72acd7264/platform.json#L48-L53 
No relation to the 'platform's git , it could use the usual versioned one

It should show up in the PACKAGES list when `run`ing, too
```
> pio pkg list -e d1_mini
Resolving d1_mini dependencies...
Platform espressif8266 @ 4.0.1 (required: espressif8266)
├── framework-arduinoespressif8266 @ 3.1.0-dev+sha.313b3c0 (required: git+https://github.com/esp8266/Arduino.git) <<<<< this is a git version
...
```

But, it is up to the user to manually keep it up to date when no commit or version (tag) are specified in the repo URI 
```
> pio pkg update -e ...
```
idk how GUI / IDE interacts with updates, and whether there are any more commands or command line flags that help out with the internal package structure

While I simply get by and edit package files and maintain development branches right there at the path I mentioned above, it is also possible to use 'local folders / symlinks' (https://docs.platformio.org/en/latest/core/userguide/pkg/cmd_install.html?highlight=symlink#local-folder)
```ini
platform_packages =
    framework-arduinoespressif8266 @ symlink://c:/Users/you/Documents/PlatformIO/esp8266-Arduino
```

TD-er:
@mcspr 
Thanks, I changed it to this one:
https://github.com/TD-er/MCVE_ESPxx/blob/800099a4fd2fe724d42f2c4417713ce208a42979/platformio.ini#L220-L225

Now it is working in the MCVE code without the needed changes of this PR.
So it can be closed as it was indeed already fixed.

N.B. my suggested fix may also cause issues for other types like `volatile int` as those differ from the types added, so const reference will not match.

Issue #6551, opened 2019-09-23
eeprom no longuer working since 2.5.1


viknet365:
----------------------------- Delete above -----------------------------

### Basic Infos

- [X ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [X ] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:					[wemos d1 r1]
- Core Version:			2.5.1
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			[Wemos D1 R1|
- Flash Mode:			unknown
- Flash Size: 		[4MB]
- lwip Variant:   v2 Lower Memory
- Reset Method:		board button
- Flash Frequency:	[unknown]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600] (serial upload only)

### Problem Description

The eeprom.read functions no longer work on my board with my IDE since 2.5.1
it works if using core 2.5.0.
The following sketch read eeprom (emulated eeprom), then write random data to it. and commit at the end.
The sketch read 0 if compiled with core 2.5.1 and later
the sketch dump correct eeprom value if run with core 2.5.0

Other project using eeprom no longuer work either (in my environment).

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
#include <EEPROM.h>

int addr = 0;
int readMode = 1; 
int writeMode = !readMode;
int eepromLength = 100;

void setup()
{
  EEPROM.begin(eepromLength); 
  Serial.begin(9600);
  delay(2000);
  Serial.println();
  Serial.println();
  Serial.printf("Core version        : %s\n"     , ESP.getCoreVersion().c_str() );
  Serial.printf("SDK  version        : %s\n"     , ESP.getSdkVersion() );
  Serial.printf("eeprom length       : %i\n"     , eepromLength );
}

void loop()
{
  if (addr % 20 == 0)
  {
    Serial.println();
    Serial.print(addr);
    Serial.print(" : ");
    delay(250);
  }

  int val = random(50);

[...]

PR #8664, opened 2022-08-30
boards.txt.py: add Wemos ESP-WROOM-02 board

rotulet:
Get fixes from https://github.com/esp8266/Arduino/issues/6551#issuecomment-534737671


mcspr:
Is it really a wemos board?
Shouldn't this be esp wroom-02 instead?

rotulet:
@mcspr good question! I have got the "official" Wemos logo on it. In many places on internet I found the same board branded Wemos but also clones. But I did not find any references on the Wemos web site!

tools/boards.txt.py
```diff
@@ -621,6 +621,21 @@
...
+    ( 'd1_wroom_02', {
+        'name': 'LOLIN(WEMOS) D1 ESP-WROOM-02',
+        'opts': {
+            '.build.board': 'ESP8266_WEMOS_D1WROOM02',
+            '.build.variant': 'd1_mini',
+            },
+        'macro': [
+            'resetmethod_nodemcu',
+            'flashmode_dio',
+            'flashfreq_26',
```
AronHetLam, line 633: `flashfreq_26` doesn't exist in macro list, only in a sub element. This causes boardsgen to fail. 
https://github.com/esp8266/Arduino/blob/c681587895c08944bcb4e84ba9ba86a24461f4a6/tools/boards.txt.py#L1092-L1113

AronHetLam: Adding this to the macros list fixes it:
```
'flashfreq_26': collections.OrderedDict([
        ( '.build.flash_freq', '26' ),
]),
```

rotulet: thanks @AronHetLam, I pushed a fix.

AronHetLam:
@rotulet, I'm not sure what exactly are you unsure about, but I'll explain the problem i had.
When I ran the script for the ESP-wroom-02 board i got an exception because there was no entry in the macros dict for `flashfreq_26`. Adding it, as you've done one the fix you pushed, fixed the issue for me, and i was able to generate a boards.txt with the entries needed for ESP-wroom-02.
I ran the scrip as standalone, and didn't look to deep into how it's all put together. As you can see in the issue I referenced this issue from, I had someone with a ESP-wroom-02 board that couldn't get my code compiled correctly.

mcspr:
Simply run the boards script and commit changes to boards.txt and documentation.

Our Check expects that we have the same filesystem contents before and after running it
https://github.com/esp8266/Arduino/actions/runs/3099983428/jobs/5202829401
(unfold the crossed out section, it shows a git-diff output of what boards.txt.py generated and what changed in the repo)

rotulet:
Sorry for the delay...

Thanks @mcspr for the hint! 

@AronHetLam I updated the board.txt file and rebased on master.

PR #8667, opened 2022-09-06
Add httpupdate lib import in ota documentation

amjed-ali-k:
Docs doesn't include any reference to include <ESP8266httpUpdate.h> library in Advanced OTA Updater section. It took me some time to find out while reading the docs


DarioGHub:
@amjed-ali-k I may not understand your motivation, but a lesson I learned was that "docs" include the examples. Given that the examples use the line `#include <ESP8266httpUpdate.h>`, how is it helpful to add that line twice more to readme.rst? Perhaps you would like to clarify the need.

Issue #8666, opened 2022-09-02
Evaluate stack guard implementation(s)


mcspr:
Right now we have at least three ways to ensure stack-smash does not happen.
- CONT guard check, with two u32 at both start and the end our struct. Additionally, we implement stack size check and fill our local array with the same value. ref. https://gitter.im/esp8266/Arduino?at=630eddaff4d7a323dead0fcf and https://github.com/esp8266/Arduino/blob/master/cores/esp8266/cont_util.cpp `cont_...` functions related to checks and 're-painting'
- `-fstack-protector` that instruments functions (also [`__attribute__((stack_protect))`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes) to manually inject such checks) with guard variables and a special `__stack_chk_{guard,fail()}`; ref. https://gcc.gnu.org/onlinedocs/gccint/Stack-Smashing-Protection.html and https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_postmortem.cpp implementations
- StackThunk also implements guards for its stack through a similar mechanisms (ref. `stack_thunk_fatal_overflow()` and main implementation in https://github.com/esp8266/Arduino/blob/master/cores/esp8266/StackThunk.h)

Could GCC implementation supplement both StackThunk and CONT? Or, replace it? (see `__attribute__` above)
Should CONT checks randomize its guard value to separate stack contents themselves from structure guard members?
Could we add something like [address sanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizerAlgorithm) that verifies that we don't go over the stack boundaries, not just protect us from writing things *over* one specific u32 value?
w

earlephilhower:
The CONT and StackThunk guards and  `-fstack-protector` are protecting against two different problems, I believe.

AIUI `stack-protector` checks for things like writing a local stack array beyond bounds to, say, inject a different return address or a higher-level call's locals.  It has no idea what the real total stack limits are, only where the current function call's place on the stack it.

The CONT and StackThunk guards check for total stack overflow, after the fact.  Too many levels of calls or too large locals.  (i.e. every funciton was well behaved, but they needed, say, 5K of stack)

GCC never knows the real total stack limits, that's more of the OS's job to handle.

[...]

PR #8670, opened 2022-09-09
Nice stack smashing postmortem message

mcspr:
Wire everything that relies on stack smashing detection to call `__stack_chk_fail()` (aka what libssp / ssp / stack-protector uses) Expose it in our debugging header

Rename overflow -> smashing, as these are different things we are trying to detect (meaning, that we check for things writing there, not some kind of `alloca` issue or the way `-fstack-check` would have worked) ref. #8666

`-fstack-protector` continues to work as it always did CONT replaces `abort()`, also moves its check to the loop wrapper to avoid dumping otherwise useless SYS context memory StackThunk replaces a similar `abort()` call

### DIFF
--- a/cores/esp8266/StackThunk.cpp
+++ b/cores/esp8266/StackThunk.cpp
@@ -142,11 +142,12 @@ void stack_thunk_dump_stack()
   ets_printf("<<<stack<<<\n");
 }
 
-/* Called when the stack overflow is detected by a thunk.  Main memory is corrupted at this point.  Do not return. */
-void stack_thunk_fatal_overflow()
+/* Called when the stack overflow is detected by a thunk.  Main memory is corrupted at this point.
+ * Do not return, use libssp-compatible function to notify postmortem and immediately reboot. */
+void stack_thunk_fatal_smashing()
 {
-    ets_printf("FATAL ERROR: BSSL stack overflow\n");
-    abort();
+    ets_printf("FATAL ERROR: BSSL stack smashing detected\n");
+    __stack_chk_fail();
 }
 
-};
+}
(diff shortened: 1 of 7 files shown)


mcspr:
btw, something to reproduce failures
```cpp
#include <Arduino.h>
#include <StackThunk.h>

#include <cont.h>
#include <user_interface.h>

#include <array>

// reproduce crash for all three methods, in sequence
constexpr uintptr_t RtcMem { 0x60001200 + 128 };
auto* count = reinterpret_cast<volatile uint32_t*>(RtcMem);

extern "C" rst_info resetInfo;

// there is only a single thunk in the app
extern "C" void thunk_failme();
extern "C" void failme() {
    printf("will ravage %p\n", stack_thunk_ptr);
    *stack_thunk_ptr = 0xfafafafa;
}

make_stack_thunk(failme)

// needs -fstack-protector to work
void ssp_failme() __attribute__((stack_protect));
void ssp_failme() {
    char buf[4];
    strcpy(buf, "ABCDE");
    puts(buf);
}

void setup() {
    Serial.begin(115200);

    switch (resetInfo.reason) {
    case REASON_DEFAULT_RST:
    case REASON_EXT_SYS_RST:
        *count = 0;
        break;
    }
}

void loop() {
    const auto last = *count;
    *count = ((last + 1) % 3);

    puts("");
    printf("%u - wait for it...\n", last);
    delay(3000);

    switch (last) {
    case 0:
        g_pcont->stack_guard1 = 0xfafafafa;
        g_pcont->stack_guard2 = 0xfafafafa;
        break;
    case 1:
        stack_thunk_add_ref();
        thunk_failme();
        break;
    case 2:
        ssp_failme();
        break;
    }
}
```

Issue #7965, opened 2021-04-06
no more 802.11n connection since 2.6.0


stef-ladefense:
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [ ] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Hardware:				ESP-12
- Core Version:			2.51 , 2.6.0, 2.7.4
- Development Env:		Arduino IDE 1.8.13
- Operating System:		Windows

### Settings in IDE

- Module:  			Wemos D1 mini r2

### Problem Description

sketch is "Udp NTP Client" included with release core.

I recently noticed that I had 802.11n and 802.11g devices on my network.
and I noticed that those which were connected in 802.11n are old modules on which I did not make any change in their codes and which were compiled with a core version lower than 2.5.0 (2.4.2 for the majorities)

I therefore put the supplied program "Udp NTP Client" and I tested with different versions of the cores, starting from 2.4.2.

up to version 2.5.1 included, it appears in 802.11n on the wifi router of my box.
except since version 2.6.0 until the last 2.7.4, it appears in 802.11g ...

do you have any idea why?

![image](https://user-images.githubusercontent.com/43587190/113735158-45171e00-96fc-11eb-9876-a37c5437e5fe.png)

fsommer1968:
Really a firmware problem? My 8266 modules connect well with 2.7.4:
![grafik](https://user-images.githubusercontent.com/17525558/114013033-62b4c680-9867-11eb-9ce3-9cf80df176cd.png)

stef-ladefense:
that's why I'm asking for advice!
I don't change anything, it's the same Lolin module (original V3.10)
it's the same code (delivered with the core, I didn't write anything more)
no change of config on my router
I install the 2.42 core, I am recognized in 802.11n on the router, ditto in 2.5.1.
as soon as I install the 2.6.0 core I am recognized in 802.11g, ditto until the current version 2.7.4.

for me there is something that has changed with the passage of 2.6.0

d-a-v:
You can try with other/older firmware version in the IDE menu:
[...]

PR #8683, opened 2022-10-08
Expand access to SDK's struct bss_info

mhightower83:
The NONOS SDK's `struct bss_info` in `user_interface.h` has grown since the beginning of this project. The additional elements are not accessible. Add a method for R/O access to full `struct bss_info`.
```cpp
const bss_info *ESP8266WiFiScanClass::getScanInfoByIndex(int i);
```

Update example WiFiScan.ino

Requested here https://github.com/esp8266/Arduino/issues/7965#issuecomment-1266744892. I have also needed this feature.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiScan.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiScan.cpp
@@ -149,12 +149,12 @@ void ESP8266WiFiScanClass::scanDelete() {
 
 /**
  * returns const pointer to the requested scanned wifi entry for furthor parsing.
- * @param networkItem uint8_t
+ * @param networkItem int
  * @return struct bss_info*, may be NULL
  */
-const struct bss_info *ESP8266WiFiScanClass::getNetworkInfo(uint8_t i) {
-    return reinterpret_cast<const struct bss_info*>(_getScanInfoByIndex(i));
-}
+const bss_info *ESP8266WiFiScanClass::getScanInfoByIndex(int i) {
+    return reinterpret_cast<const bss_info*>(_getScanInfoByIndex(i));
+};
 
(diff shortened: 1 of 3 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiScan.cpp
```diff
@@ -147,6 +147,14 @@ void ESP8266WiFiScanClass::scanDelete() {
...
 }
 
+/**
+ * returns const pointer to the requested scanned wifi entry for furthor parsing.
+ * @param networkItem uint8_t
+ * @return struct bss_info*, may be NULL
+ */
+const struct bss_info *ESP8266WiFiScanClass::getNetworkInfo(uint8_t i) {
+    return reinterpret_cast<const struct bss_info*>(_getScanInfoByIndex(i));
+}
```
mcspr, line 157: should we match the current esp32 api by name?
https://github.com/espressif/arduino-esp32/blob/a5f03a86512b270be1080d7e7cb82e4fd71c9a1a/libraries/WiFi/src/WiFiScan.h#L48
imo I'd prefer to still use a hard type. I don't see much benefit of either ambiguous void or creating our own type for whatever (...encapsulation?) reason. but, whatever approach works

minor nit: I suppose the idea is to refer to global bss info type, but we can totally lose the 'struct' since c++ does not care about that and it could be just `const bss_info*`. plus we will see definite warning if we ever move headers and forget user interface, `struct bss_info*` will just warn about something undefined instead of hard-failing

mhightower83: Thanks, I meant to look at how the ESP32 API did it and forgot. I'll do an update.

PR #8688, opened 2022-10-10
Rework CI workflows

mcspr:
- split workflows into separate files to trigger by path
  this should help out documentation and boards / eboot / pkg files updates, since those *wont* trigger usual build stuff anymore
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore
- build*.sh whatever moved into common.sh and build.sh
  trigger different parity builds, mod % rem and allow to set .ino list through the environment variable
- make use of the sketch env variable to explicitly give Windows and macOS file to build
- make use of the sketch env variable to have PlatformIO build random .ino files each ci invocation
- detect root of the repo through git cli, not base{name,dir} or relative paths
- remove legacy TRAVIS env vars, use ESP8266_ARDUINO prefix instead


mcspr:
'Required' workflows never going to work with path filtering, though. Github suggests something **very** weird with name substitution - https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks - which also brings up the fact that workflows match **by name**. ^ matched with Windows build, for some reason

I think those are somewhere in the org settings, so need to edit that
(...probably, by removing them completely...)

mcspr:
Experimenting with folds / groups, using them may result in almost unreadable error messages b/c Github *defers* log page loading.

External example
https://github.com/earthly/earthly/actions/runs/3213430189/jobs/5254151335#step:17:21
(nice and simple game of 'try to find what caused the error' by scrolling and scrolling and scrolling... :)

Issue #8692, opened 2022-10-15
WiFiClientSecure returns 0 for remoteIP(), remotePort(), localIP(), localPort() calls


JiriBilek:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [n/a ] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					ESP-01 (but doesn't matter)
- Core Version:			latest git (d3eddeb5014d04aa9e61b97474ef98d0214eab1a)
- Development Env:		Arduino IDE
- Operating System:		Windows

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			DOUT
- Flash Size: 		1MB
- lwip Variant:   v2.1 Lower Memory
- Reset Method:		dtr
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200

### Problem Description

The functions `WiFiClientSecure.remoteIP(), remotePort(), localIP(), localPort()` return 0. This is caused by the fact, that `WiFiClientSecure` is derived from `WiFiClient` but doesn't use its `ClientContext* _client` variable. It uses `WiFiClient` class that is the base of `WiFiClientSecureCtx* _ctx` instead, as is written in the documentation of the `WiFiClientSecure` class.

The problem is, that `WiFiClientSecure.remoteIP()` calls `WiFiClient.remoteIP()` with `_client=nullptr` instead of calling something like `WiFiClientSecure._ctx->remoteIP()`. The same holds with the other 3 function mentioned in the title.

The solution is to make these functions virtual and override them in `WiFiClientSecure` class.
I am creating a PR in no time.

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

The sketch is a modified WiFiClient example

```cpp
/*
    This sketch establishes a TCP connection to a "quote of the day" service.
    It sends a "hello" message, and then prints received data.
*/

#include <ESP8266WiFi.h>

#ifndef STASSID
#define STASSID "xxx"
#define STAPSK  "xxx"
#endif

[...]

PR #8693, opened 2022-10-15
Fix WiFiClientSecure remoteIP(), remotePort(), localIP(), localPort() functions

JiriBilek:
Fixes the https://github.com/esp8266/Arduino/issues/8692

Fixes the incorrect behavior of WiFiClientSecure.remoteIP(), .remotePort(), .localIP(), .localPort().

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClient.h
+++ b/libraries/ESP8266WiFi/src/WiFiClient.h
@@ -91,10 +91,10 @@ public:
   virtual uint8_t connected() override;
   virtual operator bool() override;
 
-  IPAddress remoteIP();
-  uint16_t  remotePort();
-  IPAddress localIP();
-  uint16_t  localPort();
+  virtual IPAddress remoteIP();
+  virtual uint16_t  remotePort();
+  virtual IPAddress localIP();
+  virtual uint16_t  localPort();
 
   static void setLocalPortStart(uint16_t port) { _localPort = port; }
 
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
@@ -278,6 +278,11 @@ class WiFiClientSecure : public WiFiClient {
     void flush() override { (void)flush(0); }
     void stop() override { (void)stop(0); }
 
+    IPAddress remoteIP() override { return _ctx->remoteIP(); }
+    uint16_t  remotePort() override { return _ctx->remotePort(); }
+    IPAddress localIP() override { return _ctx->localIP(); }
+    uint16_t  localPort() override { return _ctx->localPort(); }
+
     // Allow sessions to be saved/restored automatically to a memory area
     void setSession(Session *session) { _ctx->setSession(session); }
 


mcspr:
Notice that keep alive also has this problem. Could we go about fixing the inheritance / class access instead of proxying these methods or adding even more virtual calls?

JiriBilek:
I find the multiple inheritance of WiFiClient (in WiFiClientSecure and WiFiClientSecureCtx) confusing. Having read the comment https://github.com/esp8266/Arduino/blob/d3eddeb5014d04aa9e61b97474ef98d0214eab1a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h#L237
I thought the change was made on purpose and I didn't think of it more.

IMO changing the inheritance would result in lot of work quite close to the rule 'if it ain't broken don't fix it'. Although, I know, it _is_ a bit broken.

mcspr:
> I find the multiple inheritance of WiFiClient (in WiFiClientSecure and WiFiClientSecureCtx) confusing
> IMO changing the inheritance would result in lot of work quite close to the rule 'if it ain't broken don't fix it'. Although, I know, it is a bit broken.

Main issue is, bit by bit it becomes even more broken and not really fixed. Plus, anyone reading this now or in the future either thinks C++ is insane or this is the right way to do this kind of thing (whatever the thing is). So am leaning more to fixing the overall situation instead of patching it here and there. With the same idea as #7680, just doing it a bit different

More comments were added after the fact, when the clone() was introduced to fix object copies, but that was more of a sidenote thing to understand what's happening there. And lack of time was another factor of not really fixing anything

JiriBilek:
Hi,
thanks for finding the https://github.com/esp8266/Arduino/pull/7680, I thought this must have been done on purpose.
I am going to read the PR and related info to learn why exactly the change was done this way.

I agree with you that doing multiple fixes makes the code bad and hard to maintain. 
I don't know of a case where this erratic behavior of WiFiClientSecure.remoteIP() would cause problems. I came to it by chance while testing my code. So this is not a priority issue, I suppose.

JAndrassy:
I see only two options how to solve remoteIP & co. if they should work for WiFiClientSecure executed over WiFiClient pointer.
1) make remoteIP & co. in WiFiClient virtual as this PR does
2) make remoteIP of WiFiClient work for WiFiClientSecure meaning WiFiClientSecureCtx implementing ClientContext, which would require virtual methods in ClientContext

mcspr:
> I agree with you that doing multiple fixes makes the code bad and hard to maintain.

What about the time we can't do it and need to revise the whole underlying class architecture? We are just stuck b/c noone wants to read more that 5 lines of code change? Or the way to do that is to write WiFiClient2 (and 3, and 4, etc.)? I appreciate the time taken for bug & PR, but I'd really prefer to fix the core issue here. And I don't think we are in a rush atm.

> make remoteIP & co. in WiFiClient virtual as this PR does

only for the reasons described above - because the way it was written originally!

> make remoteIP of WiFiClient work for WiFiClientSecure meaning WiFiClientSecureCtx implementing ClientContext, which would require virtual methods in ClientContext

nothing of sorts, just making it possible to copy wificlient and manage clientcontext with something. i.e. handle refs, unrefs, and removing another weird quirk of self-deleting itself on refcnt zero.

So far my understanding is no-one had time to do this, so I will PR my changes instead as soon as I figure out where to put keep alive consts and flush timeout. Meaning, we revert to the old WiFiClientSecure : WiFiClient inheritance without an extra class and just use the original methods to get remote & local address info.

mcspr:
... and also ref. https://github.com/esp8266/Arduino/pull/7612#issuecomment-700942104 
Meaning, TLS state needs to be the same between copies. Not necessarily in the ClientContext, perhaps in a second helper object (and which is what I am adapting atm)

JAndrassy:
but in a perspective of the 3.1.0 release? maybe merge this PR for 3.1.0 and make the larger rewrite for 3.2.0?

mcspr:
Likely, not closing anything yet.

PR #8695, opened 2022-10-19
Allow passing beacon interval to SoftAP config 

piotrcurious:
Allow passing beacon_interval to SoftAP config.
Beacon interval affects DTIM and affects overall energy efficiency of AP (and responsiveness).
Decrease to make AP more realtime
Increase to make AP (and it's nodes) more energy and spectrum efficient. 

Afffects waking up of nodes from light sleep too.


libraries/ESP8266WiFi/src/ESP8266WiFiAP.h
```diff
@@ -37,8 +37,8 @@ class ESP8266WiFiAPClass {
 
     public:
 
-        bool softAP(const char* ssid, const char* psk = NULL, int channel = 1, int ssid_hidden = 0, int max_connection = 4);
-        bool softAP(const String& ssid,const String& psk = emptyString,int channel = 1,int ssid_hidden = 0,int max_connection = 4);
+        bool softAP(const char* ssid, const char* psk = NULL, int channel = 1, int ssid_hidden = 0, int max_connection = 4, int beacon_interval = 100);
+        bool softAP(const String& ssid,const String& psk = emptyString,int channel = 1,int ssid_hidden = 0,int max_connection = 4,int beacon_interval = 100);
```
d-a-v, line 41: [Same comment](https://github.com/esp8266/Arduino/pull/8758#discussion_r1048280597) may apply (to avoid possible breaking change)

```cpp

        bool softAP(const char* ssid, const char* psk = NULL, int channel = 1, int ssid_hidden = 0, int max_connection = 4);
        bool softAP(const String& ssid,const String& psk = emptyString,int channel = 1,int ssid_hidden = 0,int max_connection = 4);
+       bool softAP(const char* ssid, const char* psk, int channel, int ssid_hidden, int max_connection, int beacon_interval);
+       bool softAP(const String& ssid,const String& psk,int channel,int ssid_hidden,int max_connection,int beacon_interval);
```

d-a-v: I don't know if I was clear and sorry for that.
I meant it is preferable to do this, in case of subclassing / overloading.
```cpp
        bool softAP(const char* ssid, const char* psk = NULL, int channel = 1, int ssid_hidden = 0, int max_connection = 4)
        {
                return softAP(ssid, psk, channel, hidden, max_connection, 100);
        }

+       bool softAP(const char* ssid, const char* psk, int channel, int ssid_hidden, int max_connection, int beacon_interval);
```

piotrcurious:
I think whole point of preinitalizing variables by compiler is that it is code designed for embedded system, so this leaves chance for compiler to cut some slack during optimization phase f.e. by reusing constants.
Leaving them uninit and adding extra code to return defaults would make sense in bigger environment, but it will just add bloat on embedded syste

-------- Original Message --------

libraries/ESP8266WiFi/src/ESP8266WiFiAP.h
mcspr: I don't believe this matters in our case too too much?

Ambiguous overload is resolved at build time, everything is ok on that part.
If we worry about faulty default method used, arguments are expanded left-to-right so user will magically use the new method after upgrading our version.
If we worry about binary stability and preserving existing function signature in the .o file... idk how does one need to build its software for it to matter :)

piotrcurious:
For binary stability one can use immediate binary wrapper functions, solving incompatibility, bit like LDPRELOAD mechanism in linux.
For embedded system with opensource environment expecting binary compatibility is quite high aim, certainly not achieveable for such a volatile target like 8266 which will most likely disappear once trade sanctions will take traction. Even worse, i guess to crumble sanctioned entities, entire architecture will be convoluted leaving recompile as only option, so caring about binary compatibility in this very moment of history would be unethical really.

-------- Original Message --------

libraries/ESP8266WiFi/src/ESP8266WiFiAP.h
d-a-v: That's right. What I was saying seems to be only relevant when the method is virtual.

d-a-v (approved):

PR #8708, opened 2022-11-03
Correction on Heap API available from ISR

mhightower83:
The call path for "new" implemented in `abi.cpp` is not in IRAM and should not be called from an ISR.

### DIFF
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -27,15 +27,18 @@ and have several limitations:
   or use a scheduled function (which will be called outside of the interrupt
   context when it is safe) to do long-running work.
 
-* Memory operations can be dangerous and should be avoided in interrupts.
-  Calls to ``new`` or ``malloc`` should be minimized because they may require
-  a long running time if memory is fragmented.  Calls to ``realloc`` and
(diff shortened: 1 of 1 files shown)


doc/reference.rst
```diff
@@ -27,15 +27,18 @@ and have several limitations:
...
-  ``free`` or ``realloc`` themselves is also forbidden for the same reason.
-  This means that ``String``, ``std::string``, ``std::vector`` and other
-  classes which use contiguous memory that may be resized must be used with
-  extreme care (ensuring strings aren't changed, vector elements aren't
-  added, etc.).
+* Heap API operations can be dangerous and should be avoided in interrupts.
+  Calls to ``malloc`` should be minimized because they may require a long
+  running time if memory is fragmented.  Calls to ``realloc`` and ``free``
+  must NEVER be called. Using any routines or objects which call ``free`` or
+  ``realloc`` themselves is also forbidden for the same reason. This means
```
mcspr, line 34: While we are here, should we explain the reason for realloc & free?
malloc can take long time, new / delete are in flash. No apparent reason here, though, just cause

Issue #8707, opened 2022-11-03
ESPhttpUpdate Error[12] Signature verification failed


seism0saurus:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[D1 Mini]
- Core Version:			[git fff12e30686eadd7f9bb0efb8e08da7764a651d1]
- Development Env:		[Arduino IDE 1.8.13]
- Operating System:		[Debian stable]

### Settings in IDE

- Module:  			[Lolin(Wemos) D1 mini r2 & mini]
- Flash Size: 		[4MB (2MB FS 1MB OTA)]
- lwip Variant:   [v2 Lower Memory]
- CPU Frequency:		[80Mhz]
- Upload Using:		[First SERIAL than OTA]
- Upload Speed:   [921600]
- Debug Level: [HTTP_UPDATE]
- They are the default settings after selecting the D1 mini as board

### Problem Description

Since I used the latest version from the master branch the autosigning of the bin files does no longer work correctly. It worked with the last release. I switched to the git version because a bugfix for a component is not released yet.

I made a simplified MCVE with only WLAN and ESP8266httpUpdate to test that problem.
* create a project folder in your Arduino IDE folder
* You need a keypair in the project folder to enable the signing:
  ```bash
  openssl genrsa -out private.key 2048
  openssl rsa -in private.key -outform PEM -pubout -out public.key
   ```
* You have to set your WLAN credentials and the ip of your development pc and compile the scratch
* Upload it to the D1 mini
* Export the scratch as bin with the Arduino IDE feature
* Start a small server to deliver the bin file 
  ```bash
  { echo -ne "HTTP/1.0 200 OK\r\nContent-Length: $(wc -c <MCRE.ino.d1_mini.bin)\r\n\r\n"; cat MCRE.ino.d1_mini.bin; } | nc -l 9090
  ```
* Restart the D1 mini and watch the serial output. The update will fail because it can't verify the signature
* In the export log was also a signed file mentioned.  Copy it 
  ```bash
[...]

PR #8709, opened 2022-11-05
Only check the actual used lenght of the hash.

seism0saurus:
This fixes the error in #8707 but needs a professional review, since I am no C programmer.

The signature check fails, if the `hash->len()` is only 32 bytes instead of the assumed 64 bytes.
Since the variable `HashLengthMax` has a max in its name, I assume that it is allowed to have shorter hash lengths.
I assume that `memcmp` did an buffer over read here and read stuff after the 32 bytes of hash. But that's my narrow understanding of the internals of C as a non-C developer.


mcspr:
Thanks!

mcspr:
> I assume that memcmp did an buffer over read here and read stuff after the 32 bytes of hash.

Right. No function called to fill it (memset, std::fill), no zero-init (`= {0};`, `= {};`) or explicit member init (`u8 buf[] {1,2,3}`), so we end up with some random stack contents. Also, accidentally read 32bytes past hash object real buffer contents.

PR #8723, opened 2022-11-22
Fix attach servo with initial value

om2804:
functions
```
uint8_t attach(int pin);
uint8_t attach(int pin, uint16_t min, uint16_t max);
```
always sets 0 (zero) as the initial value. This is not obvious. It does not allow to attach servo without movement.

I changed the body of functions, so that the movement was caused only when transmitting the value to the function.
`uint8_t attach(int pin, uint16_t min, uint16_t max, int value);`

### DIFF
--- a/libraries/Servo/src/Servo.cpp
+++ b/libraries/Servo/src/Servo.cpp
@@ -81,7 +81,7 @@ uint8_t Servo::attach(int pin, uint16_t minUs, uint16_t maxUs)
 
 uint8_t Servo::attach(int pin, uint16_t minUs, uint16_t maxUs, int value)
 {
-  attach(pin, minUs, maxUs, _valueUs);
+  attach(pin, minUs, maxUs);
   write(value);
   return pin;
 }
--- a/libraries/Servo/src/Servo.cpp
+++ b/libraries/Servo/src/Servo.cpp
@@ -63,11 +63,6 @@ uint8_t Servo::attach(int pin)
 
 uint8_t Servo::attach(int pin, uint16_t minUs, uint16_t maxUs)
 {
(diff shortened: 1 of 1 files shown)


dok-net:
What is wrong with using this overload, as it is intended to serve your use case:
```
    // attach the given pin to the next free channel, sets pinMode, min, and max values for write(),
    // and sets the initial value, the same as write().
    // returns channel number or 0 if failure.
    uint8_t attach(int pin, uint16_t min, uint16_t max, int value);
```

om2804:
@dok-net  In the current implementation, it is impossible attach without moving a servo drive. This is not obvious. This violates compatibility with https://github.com/arduino-libraries/servo.

dok-net:
> [...] violates compatibility with https://github.com/arduino-libraries/servo.

I have checked the implementations for AVR and and SAMD in that lib. The Servo constructor sets a default position, `attach` starts the PWM on the set pin. To be sure, `write` does nothing more but update the position. I can't see how your claim that in that library `attach` would not cause the servo to move unless it is the default position to begin with already.

om2804:
Moving to zero will be at each attach after deattach without a reconstruction of a class instance.

example:
servo = new Servo();
servo .attach(9); // move to 0, first attach, it's ok
servo.write(value);  // move to value
servo.detach(); 
servo.attach(9);  // current implementation moves servo to 0, it's not ok
servo.write(value);  // move to value

dok-net:
@om2804 No offense, but I am certain I have a simpler solution that does what you requested. Please review.

om2804:
@ dok-net ok. I close this PR.

Issue #8716, opened 2022-11-16
exception.what() not a string


mhightower83:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					[ESP-01S]
- Core Version:			[80bf71662551fb5e6579713dc15e541ac48ccf98]
- Development Env:		[Arduino IDE]
- Operating System:		[Ubuntu]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[dout]
- Flash Size: 		[1MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [921600 (serial upload only)

### Problem Description

For the case of a missing `try/catch` around `new`, a secondary crash within postmortem occurs while processing the unhandled C++ Exception.

The second crash is caused when `e.what()` returns a value of `0x12`.
https://github.com/esp8266/Arduino/blob/80bf71662551fb5e6579713dc15e541ac48ccf98/cores/esp8266/core_esp8266_main.cpp#L296
resulting in a new ESP exception in postmortem at
https://github.com/esp8266/Arduino/blob/80bf71662551fb5e6579713dc15e541ac48ccf98/cores/esp8266/core_esp8266_postmortem.cpp#L144

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp
/*
   Build with Tools->C++ Exceptions: "enable"
*/
char *p = NULL;

void setup() {
  Serial.begin(115200);
  delay(100);
  Serial.printf("\nUp and running ...\n");

  // loops - crashing in postmortem
  p = new char[128000];

  if (p) {
    Serial.printf("\n%p, \"new\" worked, it should not have!\n", p);
  }
}

void loop() {}
```

### Debug Messages

```
Up and running ...

User exception (panic/abort/assert)
[...]

PR #8726, opened 2022-11-23
Fix typo in section list

mhightower83:
For GCC LD scripts, using commas in a section expression sometimes cause problems.

With this change,  `.text.gdbstub*` moved from IROM to IRAM where it was intended to be. 
@earlephilhower please correct me or confirm. Thanks!

Resolves issue #8716

I could not find an explanation of how a "," was to be interpreted. It looks like a simple `*(.somthing*, .something2*)` worked; however, a more complex `*(.text.something*, .text.something2*)` did not - _most of the time_.

I don't have much experience with LD scripts. Anyone with more insight please help.

### DIFF
--- a/tools/sdk/ld/eagle.app.v6.common.ld.h
+++ b/tools/sdk/ld/eagle.app.v6.common.ld.h
@@ -138,7 +138,7 @@ SECTIONS
 
     *(.text.app_entry*)  /* The main startup code */
 
-    *(.text.gdbstub*, .text.gdb_init)    /* Any GDB hooks */
+    *(.text.gdbstub* .text.gdb_init)    /* Any GDB hooks */
 
     /* all functional callers are placed in IRAM (including SPI/IRQ callbacks/etc) here */
     *(.text._ZNKSt8functionIF*EE*)  /* std::function<any(...)>::operator()() const */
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
From https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html it looks like the comma should work, but it's also legal to just use spaces.

````
* (section)
* (section, section, ...)
* (section section ...)
    Instead of explicitly naming particular input files in a link control script, you can refer to all files from the ld command line: use `*' instead of a particular file name before the parenthesized input-file section list. If you have already explicitly included some files by name, `*' refers to all remaining files--those whose places in the output file have not yet been defined. For example, to copy sections 1 through 4 from an Oasys file into the .text section of an a.out file, and sections 13 and 14 into the .data section:

    SECTIONS {
      .text :{
        *("1" "2" "3" "4")
      }
      
      .data :{
        *("13" "14")
      }
    }

    `[ section ... ]' used to be accepted as an alternate way to specify named sections from all unallocated input files. Because some operating systems (VMS) allow brackets in file names, that notation is no longer supported. 
````

So, LGTM, especially since this was proven as fixing something broken!

PR #8736, opened 2022-12-01
Added NONOS SDK 3.0.0 ~ 3.0.5 (again)

metarutaiga:
I push the PR again.
I used the 3.0.5, it seems stable.

In Addition, the patch i don't know it needed after 3.0.0
// # WPA2-Enterprise patch which replaces a double-free with nop, see #8082
// # special vPortFree to recover leaked memory


cores/esp8266/core_esp8266_main.cpp
```diff
@@ -404,6 +404,73 @@ extern "C" void flashinit (void);
...
+    const partition_item_t at_partition_table[] =
+    {
+        { SYSTEM_PARTITION_RF_CAL,           rf_cal,           0x1000 },
+        { SYSTEM_PARTITION_PHY_DATA,         phy_data,         0x1000 },
+        { SYSTEM_PARTITION_SYSTEM_PARAMETER, system_parameter, 0x3000 },
+    };
+    system_partition_table_regist(at_partition_table, sizeof(at_partition_table) / sizeof(at_partition_table[0]), system_get_flash_size_map());
+
+    extern void user_rf_pre_init();
+    void user_rf_pre_init();
```
mcspr, line 469:
```suggestion
    user_rf_pre_init();
```
(if you intended to call it)

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -404,6 +404,73 @@ extern "C" void flashinit (void);
...
+    extern uint32_t user_rf_cal_sector_set(void);
+    user_rf_cal_sector_set();
+
+    const partition_item_t at_partition_table[] =
+    {
+        { SYSTEM_PARTITION_RF_CAL,           rf_cal,           0x1000 },
+        { SYSTEM_PARTITION_PHY_DATA,         phy_data,         0x1000 },
+        { SYSTEM_PARTITION_SYSTEM_PARAMETER, system_parameter, 0x3000 },
+    };
+    system_partition_table_regist(at_partition_table, sizeof(at_partition_table) / sizeof(at_partition_table[0]), system_get_flash_size_map());
```
mcspr, line 466: don't we run into a conflict with our .ld scripts? also note cores/esp8266/FlashMap.h (although, that one is mostly experimental and could be changed at any point)

boards.txt
```diff
@@ -365,6 +365,18 @@ generic.menu.sdk.nonosdk221=nonos-sdk 2.2.1 (legacy)
...
+generic.menu.sdk.nonosdk301=nonos-sdk 3.0.1
+generic.menu.sdk.nonosdk301.build.sdk=NONOSDK301
+generic.menu.sdk.nonosdk302=nonos-sdk 3.0.2
+generic.menu.sdk.nonosdk302.build.sdk=NONOSDK302
+generic.menu.sdk.nonosdk303=nonos-sdk 3.0.3
+generic.menu.sdk.nonosdk303.build.sdk=NONOSDK303
+generic.menu.sdk.nonosdk304=nonos-sdk 3.0.4
+generic.menu.sdk.nonosdk304.build.sdk=NONOSDK304
+generic.menu.sdk.nonosdk305=nonos-sdk 3.0.5
+generic.menu.sdk.nonosdk305.build.sdk=NONOSDK305
```
mcspr, line 379: ...do we actually need 3.0.0...3.0.4?

cores/esp8266/core_esp8266_main.cpp
metarutaiga: It cannot run with our partition.
It need its default partition.
We hack RF CAL partition, so the partition is not important.

cores/esp8266/core_esp8266_main.cpp
metarutaiga: I fixed it.

boards.txt
metarutaiga: Yeah, I am not sure all 3.0 are stable for everyone.

cores/esp8266/core_esp8266_main.cpp
mcspr: Don't really understand where `system_get_flash_size_map()` size comes from :)
Do we read hard-coded value like `esptool write_flash --flash_size <SIZE>` that is part of the .bin? Or is it generating something on the fly based off of the real size?

metarutaiga: the system_get_flash_size_map is in user_interface.h of NONOS SDK.

I don't know that also, but system_partition_table_regist is must be called.
I written 6, but system is said 4, and it's not match and system initialize failed.

So I think called system_get_flash_size_map to match system_get_flash_size_map inside system_partition_table_regist.

mcspr: OK. Looking at RTOS https://github.com/espressif/ESP8266_RTOS_SDK/issues/122 there is a similar confusion, does not seem like this API was really thought through :/
I guess - whatever, it is broken in some cases. Thanks for checking out, we dragged 3.0.0 beta for long enough at this point

Issue #8731, opened 2022-11-26
Should the C++ constructor and destructor tables move into FLASH?


metarutaiga:
https://github.com/esp8266/Arduino/blob/fe451a814f0d219c8cc8cbac7735e981f8d46f9c/tools/sdk/ld/eagle.app.v6.common.ld.vtables.h#L15-L26

and

https://github.com/esp8266/Arduino/blob/fe451a814f0d219c8cc8cbac7735e981f8d46f9c/tools/sdk/ld/eagle.app.v6.common.ld.vtables.h#L35-L39

These are all 32bit pointers.
It should move into flash to reduce memory usage.

I change it in my repo.
https://github.com/metarutaiga/ESP8266-Arduino/commit/3bc7ce4c93659546fc5bdc9ac41932f6d3998491

mcspr:
We could just copy-paste those .{c,d}tor entries from the vtables .h .rodata into the .irom0 one? No relation to vtables, just move those to flash at all times

earlephilhower:
IIRC they're in RAM now because of IRQ safety.  If any classes are used in IRQ handlers, you can't be sure the flash is accessible so risk crashes.  There was a bug I can't find opened about this after the original memory map shuffling...

mcspr:
Arent dtors & ctors (and the whole init array) are supposed to be used for global objects? IRQ code will inject ctor / dtor call when needed, where the actual section can be changed with the IRAM attr

earlephilhower:
Good catch, that sounds right.

The `__init_array_start/end` are only used in the CRT initialization (now handled by us in `main.cpp`) https://github.com/esp8266/Arduino/blob/fe451a814f0d219c8cc8cbac7735e981f8d46f9c/cores/esp8266/core_esp8266_main.cpp#L271-L278

BSS start/end, I *think* is used by the ROM code to do the `bzero` but I can't find where we stuff it for the ROM to find right this morning.

I'd say move it at this point.  We're not saving all that much space at this point, but every byte counts...

PR #8737, opened 2022-12-02
Move C++ constructor and destructor tables and .bss to .irom0 (flash)

metarutaiga:
the PR is from the issue https://github.com/esp8266/Arduino/issues/8731

The FREE_MORE_DRAM is force move some string literal into flash.
It's not stable.

for example the __get_global_locale is global, I didn't modify it ever.
But it has 364 bytes.


tools/sdk/ld/eagle.app.v6.common.ld.h
```diff
@@ -229,6 +247,35 @@ SECTIONS
...
+    *libc.a:lib_a-nano-vfprintf_float.o (.rodata .rodata.*)
+    *libc.a:lib_a-nano-vfprintf.o       (.rodata .rodata.*)
+    *libc.a:lib_a-nano-vfscanf_float.o  (.rodata .rodata.*)
+    *libc.a:lib_a-nano-vfscanf_i.o      (.rodata .rodata.*)
+    *libc.a:lib_a-rand.o                (.rodata .rodata.*)
+    *libc.a:lib_a-strtod.o              (.rodata .rodata.*)
+    *libc.a:lib_a-tzset_r.o             (.rodata .rodata.*)
+    *libc.a:lib_a-tzvars.o              (.rodata .rodata.*)
+    *liblwip2-536.a:                    (.rodata .rodata.memp_pools .rodata.tcp_pcb_lists .rodata.dns_mquery_v4group .rodata.ip_addr_broadcast .rodata.ip_addr_any)
+    *liblwip2-1460.a:                   (.rodata .rodata.memp_pools .rodata.tcp_pcb_lists .rodata.dns_mquery_v4group .rodata.ip_addr_broadcast .rodata.ip_addr_any)
```
mcspr, line 273: ?
no -feat, no ipv6 variant. unless intentional.
not really sure every .rodata is readable, though. doesn't this also need exception trap for unaligned reads e.g. when we try to read u16 or u8 / strings?

metarutaiga: I added feat and v6

the FREE_MORE_DRAM is not stable for everyone.

metarutaiga: I removed FREE_MORE_DRAM because its need to trace from non 32bit load handler.

tools/sdk/ld/eagle.app.v6.common.ld.h
```diff
@@ -162,6 +162,24 @@ SECTIONS
...
+    KEEP (*crtbegin.o(.dtors))
+    KEEP (*(EXCLUDE_FILE (*crtend.o) .dtors))
+    KEEP (*(SORT(.dtors.*)))
+    KEEP (*(.dtors))
+    . = ALIGN(4);       /* this table MUST be 4-byte aligned */
+    _bss_table_start = ABSOLUTE(.);
+    LONG(_bss_start)
+    LONG(_bss_end)
+    _bss_table_end = ABSOLUTE(.);
+
```
mcspr, line 182: Was this something from old SDK? Don't see anything referring to this

tools/sdk/ld/eagle.app.v6.common.ld.h
```diff
@@ -229,6 +247,9 @@ SECTIONS
     /* Fundamental type info */
     *(.rodata._ZTIPKc .rodata._ZTIc .rodata._ZTIv .rodata._ZTSv .rodata._ZTSc .rodata._ZTSPKc .rodata._ZTSi .rodata._ZTIi)
 
+    /* std::make_shared */
+    *(.rodata._ZZNSt19_Sp_make_shared_tag5_S_tiEvE5__tag)
+
```
mcspr, line 252: Looking at shared ptr bits
https://github.com/gcc-mirror/gcc/blob/fee53a3194c0d8b747486b23980f1214cc1355b9/libstdc%2B%2B-v3/include/bits/shared_ptr_base.h#L566

feels weird leaving stdlib static here of a implementation-dependant obj. I'd rather remove this one

tools/sdk/ld/eagle.app.v6.common.ld.h
mhightower83: SDK 2.2.1 will zero that range during `user_start()`. I think the globals without values are grouped there.

mhightower83: This range is zeroed 1 byte at a time, so flash storage will not work.

mcspr: I mean `_bss_table_start` and `end`. It injects addresses of bss\_start and end between the two, but those are still in RAM
(see LONG(addr) -> https://sourceware.org/binutils/docs/ld.html#Output-Section-Data)

PR #8753, opened 2022-12-10
Servo: re-attach to same value (position)

### LINKED ISSUE #8752: Re-attaching servo causes erratic movement
In #8723, there is a report of an incompatibility with AVR Arduino.
I don't think the PR is correct, the issue is for real.
I am attaching a PR that I believe fixes the issue.

dok-net:
The current implementation of the Servo lib always resets the position when `detaching`.
In AVR Servo, this isn't the case, instead, it doesn't move the servo but leaves it as it was before getting detached.
Fixes #8752.
Should obviate #8723

### DIFF
--- a/libraries/Servo/src/Servo.cpp
+++ b/libraries/Servo/src/Servo.cpp
@@ -58,7 +58,7 @@ Servo::~Servo() {
 
 uint8_t Servo::attach(int pin)
 {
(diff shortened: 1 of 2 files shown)


mcspr:
I guess, approved by @om2804?
By AVR Servo, you mean https://github.com/arduino-libraries/Servo?
(also includes sam,samd,nrf52 and stm)

dok-net:
@mcspr Yes, the original Arduino Servo lib.

PR #8755, opened 2022-12-12
Resolve flash address issues with SDK v3.0.0

mhightower83:
Fix EEPROM vs `RF_CAL` flash address conflict. The `EEPROM` address and `RF_CAL` address were the same.

Add support for Flash size: "Mapping defined by Hardware and Sketch" aka autoconfig

Change `at_partition_table` from dynamic to static.

When the Arduino IDE flash size selected is correct, the flash sector positions for EEPROM, RF_CAL, and SYSTEM_PARAMETER  are kept constant across SDK selections; however, when you change SDK versions, you should erase WiFi settings. The file system should be portable across SDK selections; however, if you have valued data backup before changing. Data may be lost if the flash size setting and device mismatch.

Status: Ready


cores/esp8266/core_esp8266_main.cpp
```diff
@@ -408,71 +409,97 @@ uint32_t __flashindex;
...
+      ets_uart_printf(PSTR("\nSystem partition table registration failed!\n"));
+
+      //C What is the best action on failure?
+      //C * The above printf maybe redundant. `system_partition_table_regist`
+      //C   appears to print specific failure messages. most of the time.
+      //C * Things will not get better by rebooting.
+      //C * Most likely the error message was not readable - mismatched console speed
+      //C * Expressif example do a `while(true){};` on fail. HWDT appears to be
+      //C   off and the device freezes.
+      abort();
```
mhightower83, line 493: What is the preferred action for `system_partition_table_regist` failure?
* The above printf may be redundant. `system_partition_table_regist`
  appears to print specific failure messages. most of the time.
* Things will not get better by rebooting.
* Most likely the error message was not readable - mismatched console speed
* Expressif example do a `while(true){};` on fail. HWDT appears to be
  off and the device freezes.

mcspr: Preferably, we should be able to handle partition(s) size and address and also validate them at build time to avoid this path altogether. But, for this set of choices, while(true) loop seems most sensible here.

mhightower83: My final solution was to force the printing speed to 11500bps. (special case) Print messages, block on input character read in a `do {} while(true);` loop. Message repeats on any character received.

mhightower83:
I think things are covered with "non-auto-size .ld files"; however, I plan to change RF_CAL and parameters to be the last 4 sectors of the flash. Get rid of the EEPROM relative approach. That will be more consistent with `ESP.eraseConfig()`. I'll keep the PHY_DATA read spoofing over the EEPROM. Of course, my thinking may change when I go to do it.

With `auto-size flash mode` the data in PROGMEM is used before `flashinit()` has run. With `user_pre_init` it runs before any C runtime code runs and before `flashinit()` in `user_init()`. I think we will need a version that can be called from `user_pre_init` and `__flashindex` will need to move to the .noinit section. And, `panic()` will need to be dealt with differently.

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -400,79 +400,230 @@ extern "C" void __disableWiFiAtBootTime (void)
...
+        // precalibration rate.
+        if (F_CRYSTAL != 40000000) {
+            uart_div_modify(0, F_CRYSTAL * 2 / 115200);
+            ets_delay_us(150);
+        }
+        do {
+            ETS_PRINTF("\n\n");
+            // Because SDK v3.0.x always has a non-32-bit wide exception handler
+            // installed, we can use PROGMEM strings with Boot ROM print functions.
+#if defined(DEBUG_ESP_CORE) | defined(DEBUG_ESP_PORT) // DEBUG_ESP_CORE => verbose
```
mcspr, line 557:
```suggestion
#if defined(DEBUG_ESP_CORE) || defined(DEBUG_ESP_PORT) // DEBUG_ESP_CORE => verbose
```
btw whats the reason for ESP PORT dep? another worry is this way we depend on it being properly constructed, which may not be the case if called before c++ ctors init

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -400,79 +400,230 @@ extern "C" void __disableWiFiAtBootTime (void)
...
+
+        // All the examples I find, show the partition table in the global address space.
+        static partition_item_t at_partition_table[] =
+        {
+            { SYSTEM_PARTITION_PHY_DATA,         phy_data,         0x1000 }, // type 5
+            { SYSTEM_PARTITION_RF_CAL,           rf_cal,           0x1000 }, // type 4
+            { SYSTEM_PARTITION_SYSTEM_PARAMETER, system_parameter, 0x3000 }, // type 6
+        };
+        _at_partition_table = at_partition_table;
+        _at_partition_table_sz = sizeof(at_partition_table) / sizeof(at_partition_table[0]);
```
mcspr, line 516:
```suggestion
        _at_partition_table_sz = std::size(at_partition_table);
```

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -400,79 +400,230 @@ extern "C" void __disableWiFiAtBootTime (void)
...
+        #if FLASH_MAP_SUPPORT & defined(DEBUG_ESP_PORT)
+        // I don't think this will ever fail. Everything traces back to the results of spi_flash_get_id()
+        if (flash_size != flashchip->chip_size) {
+            chip_sz_str = PSTR("Flash size mismatch, check that the build setting matches the device.\n");
+            continue;
+        }
+        #endif
+
+        // All the examples I find, show the partition table in the global address space.
+        static partition_item_t at_partition_table[] =
```
mcspr, line 509:
```suggestion
        static const partition_item_t at_partition_table[] =
```

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -503,8 +654,10 @@ extern "C" void user_init(void) {
 #if defined(MMU_IRAM_HEAP)
     umm_init_iram();
 #endif
-#if FLASH_MAP_SUPPORT
-    flashinit();
+#if FLASH_MAP_SUPPORT & (NONOSDK < 0x30000)
```
mcspr, line 657:
```suggestion
#if FLASH_MAP_SUPPORT && (NONOSDK < 0x30000)
```

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -400,79 +400,230 @@ extern "C" void __disableWiFiAtBootTime (void)
...
+        // `core_esp8266_phy.cpp`.
+        extern void sdk3_begin_phy_data_spoof();
+        sdk3_begin_phy_data_spoof();
+
+        ld_config_chip_size = phy_data + 4096 * 5;
+
+        // -DALLOW_SMALL_FLASH_SIZE=1
+        // Allows for small flash-size builds targeted for multiple devices,
+        // commonly IoT, of varying flash sizes.
+        #if !defined(FLASH_MAP_SUPPORT) & !defined(ALLOW_SMALL_FLASH_SIZE)
```
mcspr, line 482:
```suggestion
        #if !defined(FLASH_MAP_SUPPORT) && !defined(ALLOW_SMALL_FLASH_SIZE)
```

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -400,79 +400,230 @@ extern "C" void __disableWiFiAtBootTime (void)
...
+            continue;
+        }
+        #elif defined(ALLOW_SMALL_FLASH_SIZE) && !defined(FLASH_MAP_SUPPORT)
+        // Note, while EEPROM is confined to a smaller flash size, we are still
+        // placing RF_CAL and SYSTEM_PARAMETER at the end of flash. To prevent
+        // this, esptool or its equal needs to not update the flash size in the
+        // .bin image.
+        #endif
+
+        #if FLASH_MAP_SUPPORT & defined(DEBUG_ESP_PORT)
```
mcspr, line 500:
```suggestion
        #if FLASH_MAP_SUPPORT && defined(DEBUG_ESP_PORT)
```
(would also suggest moving all `#...` to the beginning of line? things are pretty dense here)

cores/esp8266/core_esp8266_main.cpp
mhightower83: I use `defined(DEBUG_ESP_PORT)` strictly as a debug build flag. It's value is never used for selecting the serial device. All printing is handled through Boot ROM `ets_uart_printf`.

cores/esp8266/core_esp8266_main.cpp
mhightower83: I have gone back and forth on this. I kept getting lost with all the `#if`s in column one. The density of the `#if`s is the reason I started indenting the innermost `#if`s to reduce the visual disruption to a block of "C" code. When nested the outer `#if` are at the beginning of the line.  I don't know if doing this is a style violation.

cores/esp8266/core_esp8266_main.cpp
mhightower83: I'll correct all the `&` vs `&&` a little bit of language confusion. I took one wrong instance and multiplied it

cores/esp8266/core_esp8266_main.cpp
mhightower83: I assume this will happen at C++ compile time and we can count on not needing C++ runtime init.

cores/esp8266/core_esp8266_main.cpp
mhightower83: done

cores/esp8266/core_esp8266_main.cpp
mcspr: It is, number of elements is part of the type info.
`template <typename T, size_t Size> constexpr size_t size(const T (&)[Size]) { return Size; }`

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -400,79 +400,230 @@ extern "C" void __disableWiFiAtBootTime (void)
...
+        // 0x3000" are positioned in the last five sectors of flash memory.
+        // PHY_INIT_DATA is special. It is a one time read of 128 bytes of data
+        // that is provided by a spoofed flash read.
+        #if FLASH_MAP_SUPPORT
+        flash_size = __flashdesc[__flashindex].flash_size_kb * 1024u;
+        #else
+        // flashchip->chip_size is updated by the SDK. The size is based on the
+        // value patched into the .bin header by esptool.
+        // system_get_flash_size_map() returns that patched value.
+        flash_size = flashchip->chip_size;
```
mhightower83, line 456: Rethinking things, should `flash_size` simply be set with `1 << ((spi_flash_get_id() >> 16) & 0xff)` and avoid the possible config selection errors? However, use these config selections in debug builds to verify against the `spi_flash_get_id()` based value.

mcspr: Maybe later down the line, also reducing the number of .ld files
I'd also call debugging as a pretty good reason to keep it as-is

amrlsayed:
Hello, sorry for commenting on an old thread, but I thought your comment will help me out.
I have a project using the SDK `3.0.1` just before this merge, and I have the problem that the EEPROM is overwritten randomly that I am not able to reproduce. 
My question is does this issue relate to that ?  as the EEPROM address and RF_CAL address were the same, so may be RF_CAL data overwrites EEPROM data ?
Thanks

mhightower83:
> ... EEPROM is overwritten randomly that I am not able to reproduce.

Yes, EEPROM writes may write over RF_CAL data. And SDK RF_CAL writes may write over EEPROM data. I suspect/expect the SDK implements wear leveling on their writes. This means the location of the writing might not be the same each time. This could contribute to the reproducibility problem.

PR #8758, opened 2022-12-12
Add eraseap option to WiFi disconnect method.

cziter15:
The problem is that WiFi.disconnect function always set SSID and Password to NULL. On ESP32's version of Arduino, user can define if the config should be ever touched or not. When manually implementing reconnection, current state requires to store credentials externally between disconnect and connect method calls.

This PR aims to fix this problem by introducing second variable (eraseap), which defaults to false to maintain compatibility.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -357,7 +357,7 @@ bool ESP8266WiFiSTAClass::reconnect() {
  * @param wifioff
  * @return  one value of wl_status_t enum
  */
-bool ESP8266WiFiSTAClass::disconnect(bool wifioff, bool eraseap) {
+bool ESP8266WiFiSTAClass::disconnect(bool wifioff) {
 {
     return disconnect(wifioff, false);
 }
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -353,28 +353,29 @@ bool ESP8266WiFiSTAClass::reconnect() {
 }
 
 /**
- * Disconnect from the network
- * @param wifioff
+ * Disconnect from the network with clearing saved credentials
+ * @param wifioff Bool indicating whether STA should be disabled.
  * @return  one value of wl_status_t enum
  */
 bool ESP8266WiFiSTAClass::disconnect(bool wifioff) {
-    return disconnect(wifioff, false);
+    // Disconnect with clearing saved credentials.
+    return disconnect(wifioff, true);
 }
 
 /**
  * Disconnect from the network
- * @param wifioff
- * @param eraseap
+ * @param wifioff Bool indicating whether STA should be disabled. 
+ * @param eraseCredentials Bool indicating whether saved credentials should be erased.
  * @return  one value of wl_status_t enum
  */
-bool ESP8266WiFiSTAClass::disconnect(bool wifioff, bool eraseap) {
+bool ESP8266WiFiSTAClass::disconnect(bool wifioff, bool eraseCredentials) {
     bool ret = false;
 
     // Read current config.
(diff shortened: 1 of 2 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -48,7 +48,7 @@ class ESP8266WiFiSTAClass: public LwipIntf {
         bool config(IPAddress local_ip, IPAddress gateway, IPAddress subnet, IPAddress dns1 = (uint32_t)0x00000000, IPAddress dns2 = (uint32_t)0x00000000);
 
         bool reconnect();
-        bool disconnect(bool wifioff = false);
+        bool disconnect(bool wifioff = false, bool eraseap = false);
```
d-a-v, line 51: Thanks !
I'd approve but I remember @devyte saying that in case of an API addition it should be a principle to leave the original declaration as-is and add a new one for the API extension, in order to not break any overload that would exist in other projects using this one.
This would give:
```cpp
  bool disconnect(bool wifioff = false);
+ bool disconnect(bool wifioff, bool eraseap);
```

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -357,11 +357,28 @@ bool ESP8266WiFiSTAClass::reconnect() {
  * @param wifioff
  * @return  one value of wl_status_t enum
  */
-bool ESP8266WiFiSTAClass::disconnect(bool wifioff) {
+bool ESP8266WiFiSTAClass::disconnect(bool wifioff, bool eraseap) {
+{
+    return disconnect(wifioff, false);
+}
```
d-a-v, line 363: There's a typo, it is calling itself ?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -48,7 +48,9 @@ class ESP8266WiFiSTAClass: public LwipIntf {
         bool config(IPAddress local_ip, IPAddress gateway, IPAddress subnet, IPAddress dns1 = (uint32_t)0x00000000, IPAddress dns2 = (uint32_t)0x00000000);
 
         bool reconnect();
+
         bool disconnect(bool wifioff = false);
+        bool disconnect(bool wifioff = false, bool eraseap = false);
```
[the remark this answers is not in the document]
cziter15: Yeah, my mistake :) I wanted to be too fast.

d-a-v (approved):

cziter15:
Looks like mustr emove defaults also from header

cziter15:
Do you know why its failing? Is it missing header?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -358,10 +358,27 @@ bool ESP8266WiFiSTAClass::reconnect() {
  * @return  one value of wl_status_t enum
  */
 bool ESP8266WiFiSTAClass::disconnect(bool wifioff) {
+{
```
d-a-v, line 361: Two opening braces ?

cziter15:
Shame on me, will fix today. Didnt see that 😀

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
```diff
@@ -358,10 +358,26 @@ bool ESP8266WiFiSTAClass::reconnect() {
  * @return  one value of wl_status_t enum
  */
 bool ESP8266WiFiSTAClass::disconnect(bool wifioff) {
+    return disconnect(wifioff, false);
```
d-a-v, line 361: Default behaviour was always `true` no ?

cziter15: Yes, it was. Good point.

d-a-v (approved):

Issue #8761, opened 2022-12-16
Panic in flashinit fails to generate stack dump


mhightower83:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					ESP-01S
- Core Version:			da48a52d7a8854363946fe0c05f02ec029c852c5
- Development Env:		Arduino IDE
- Operating System:		Ubuntu

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			dio
- Flash Size: 		1MB
- lwip Variant:   v2 Lower Memory
- Reset Method:		nodemcu
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   115200 (serial upload only)

### Problem Description
When using FLASH_MAP_SUPPORT with a custom table and the real flash size is missing from the table `panic()` is called.
However, no stack dump is printed.

The problem is all the GPIO pins are reset turning off the serial interface before `flashinit()` is called,

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#if FLASH_MAP_SUPPORT
#include <flash_hal.h>
#define FLASH_MAP_bad \
    { \
        { .eeprom_start = 0x403fb000, .fs_start = 0x403c0000, .fs_end = 0x403fb000, .fs_block_size = 0x1000, .fs_page_size = 0x100, .flash_size_kb = 2040 }, \
        { .eeprom_start = 0x405fb000, .fs_start = 0x40400000, .fs_end = 0x405fa000, .fs_block_size = 0x2000, .fs_page_size = 0x100, .flash_size_kb = 4090 }, \
        { .eeprom_start = 0x402fb000, .fs_start = 0x402eb000, .fs_end = 0x402fb000, .fs_block_size = 0x1000, .fs_page_size = 0x100, .flash_size_kb = 1020 }, \
[...]

PR #8762, opened 2022-12-16
fix flashinit panic not printing

mhightower83:
Resolves
* https://github.com/esp8266/Arduino/issues/8761#issue-1499638772


cores/esp8266/flash_hal.h
```diff
@@ -33,21 +33,32 @@ extern "C" {
...
 
+#ifndef QUOTE
+#define QUOTE(a) __STRINGIFY(a)
+#endif
+
+#ifdef DEBUG_ESP_CORE
+#define DEBUG_FLASH_MAP_NOT_FOUND " flashinit: configuration not found"
+#else
+#define DEBUG_FLASH_MAP_NOT_FOUND
+#endif
```
mcspr, line 48: we always print exc text though, why hide?
could also just have `static const char DEBUG_FLASH_MAP_NOT_FOUND[] PROGMEM = ...` right here at the top?

mhightower83: Was concerned about growing space used up in non-debug builds.

I was thinking I wanted the line # reported to be in the function body; however, I later realized it is the macro line number that is being reported. Also, if I place it in the scope of the function I avoid the potential problem of a hidden variable name conflict in the module.

I have no strong preferences.

cores/esp8266/core_esp8266_main.cpp
```diff
@@ -480,6 +480,11 @@ extern "C" void user_init(void) {
 
     uart_div_modify(0, UART_CLK_FREQ / (115200));
 
+#if FLASH_MAP_SUPPORT
+    const char *err_msg = flashinit();
+    if (err_msg) __unhandled_exception(err_msg);
```
mhightower83, line 485: The message text looks like this when printed. Is there any objection to the "Unhandled C++ exception" choice in the postmortem? I didn't see any others that would work well for a single string.
```
Unhandled C++ exception: autoconfig.ino.ino:15 flashinit: configuration not found
```

d-a-v: could `__panic__func()` or `__assert_func()` help ?

mhightower83: If I change postmortem to print panics when `s_panic_line` is set to 0. That seems to work. I'll push the change.

cores/esp8266/flash_hal.h
d-a-v: Maybe file and line are unnecessary in this case, and only the string is relevant and useful even in non-debug mode.

mhightower83: That was my first attempt then I realized the user could provide a broken override `flashinit`. 
Providing the file and line gives a strong indication of where the fault lives. And that the override won during the linking process.

mcspr:
Just as a thought, don't we want to return something like this?

```cpp
FlashInitResult flashinit();
```
```cpp
struct FlashInitResult {
    bool ok { false };
    const char* err_msg;
};
```

d-a-v:
`flashinit()` was not designed to fail, so it is 'panic'ing instead

mhightower83:
Too many merge conflicts were corrected with web edit, I need to correct some stuff and do a few real test.

mcspr:
> flashinit() was not designed to fail, so it is 'panic'ing instead

I just mean the return value. Struct member name tells us what we mean right there, no need to remember triple meaning of a char pointer

d-a-v (approved):

mhightower83:
> `FlashInitResult flashinit();`

FWIF: I tried this method and it increases "IROM code in flash" by 16 bytes.

PR #8770, opened 2022-12-21
NONOS SDK clean-up

### LINKED ISSUE #5784: SDK issues (SDK reverted from pre3 to 2.2.1)
This issue keeps track of underlying espressif SDK questions

Current SDK in master branch is nonos-sdk-2.2.1, as shipped in core-2.4.2.

This espressif nonos-sdk 2.2.1 has a WiFi sleep bug (ref: #2330) which is partly solved with a pre-version of espressif nonos-sdk-v3. Workarounds exist with this issue, like regularly sending gratuitous ARP.

The unofficial nonos-sdk-pre-v3.0.0 espressif firmware shipped with arduino-core 2.5.0 has two issues that leaded us to revert back to 2.2.1: 
* Some boards show erratic behavior (radio connection is quickly lost), with an unknown cause.
  These boards work well with previous nonos-sdk-2.2.1 firmware (#5736)
* Overall performances have decreased (#5513)

Why not migrating to official nonos-sdk-v3 now ?
- arduino core uses unofficial hacks and API with nonos-sdk.
  These hacks currently do not work with latest versions (v3.0.0+) of this sdk.
- nonos sdk is claimed to be EOL by Espressif, meaning that we can never officially ask for feature support like full WPA2-Enterprise ([rtos](https://github.com/espressif/ESP8266_RTOS_SDK/issues/442), [nonos](https://github.com/espressif/ESP8266_NONOS_SDK/issues/133)) (refs: [1](https://github.com/espressif/ESP8266_NONOS_SDK/issues/192#issuecomment-445102901) [2](https://github.com/espressif/ESP8266_NONOS_SDK/search?q=RTOS&state=open&type=Issues))
- Migrating to RTOS-SDK is being under serious consideration
  disclaimer: **not saying** that esp8266 arduino-core will be RTOS - we can always run in a single task / single stack while being run by an RTOS (I personally think this is also the nonos-sdk model) - That, because we are short in main RAM

mcspr:
Remove nonos-sdk-v2.2.0-28-g89920dc aka 3v0 (per matrix proposal), since we have a real v3
Remove 3.0.0...3.0.4 to reduce overhead in maintaining a bunch or binary patches
Update our docs and menu opts to mention 'experimental' status

I believe 3v0 support was treated as 'experimental', so this is a *non breaking* change?
Not sure why we refer to NONOS as 'Espressif Fw / Firmware', maybe also rename it?
Old versions still remains in git history, so anyone wanting to play around with 3.0.x could still make use of that work

(*edit* by @d-a-v)
Closes #5784


d-a-v (approved):
LGTM

+ 2.5K lines made unnecessary is always good

d-a-v:
> I believe 3v0 support was treated as 'experimental', so this is a non breaking change?

Yes

>  Not sure why we refer to NONOS as 'Espressif Fw / Firmware', maybe also rename it?

Agreed

PR #8778, opened 2022-12-29
Fix typo in the example of ESP8266HttpClient library

farhan2077:
Basically what the title says


earlephilhower:
That is not a typo.  Using a ref avoids copying potentially large strings around.

PR #8785, opened 2023-01-03
fix Stream's parseFloat()

### LINKED ISSUE #8782: parseFloat is not able to parse positive float numbers
### Basic Infos

#### Platform

- Hardware:				Adafruit Feather HUZZAH ESP8266
- Core Version:			[latest git hash or date]
- Operating System:		Windows

### Settings in IDE

- Flash Size: 		[4MB/1MB]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[160MHz]
- Upload Using:		[SERIAL]
- Upload Speed:   11520

### Problem Description
parseFloat is not able to parse positive float numbers

I want to send float numbers from Arduino UNO to ESP8266. I2c was my first choice but unfortunately ESP8266 does not support receiver mode and I can not make ESP8266 master as there are time critical threads running on UNO and when it is free, it sends telemetry data to ESP8266. 

Hence, I am attempting to receive data using serial and parseFloat. However, the parser is not able to parse the positive float numbers and also does not allow me to modify the SKIP parameters. 

Any help would be greatly appreciated.

**Sender Code :**  (Arduino UNO)
```cpp
void setup()
{
  Serial.begin(115200);
}

float x = 157.50;

void loop()
{
  // Start the packet
  Serial.print(x);
  x++;
  delay(100);
}
```
**Receiver Code : (ESP8266)**
```cpp
#include <Arduino.h>
#include <Stream.h>

void setup() {
  Serial.begin(115200);
}

void loop() {
  if (Serial.available() > 0) {
    float myFloat = Serial.parseFloat();

    //prints the received float number
    Serial.print("I received: ");
    Serial.println(myFloat);
  }
}
```

d-a-v:
cherry-pick updates for `Stream::peekNextDigit()` from AVR implementation: decimal dot is optionally allowed

fixes #8782

### DIFF
--- a/cores/esp8266/Stream.cpp
+++ b/cores/esp8266/Stream.cpp
@@ -58,16 +58,16 @@ int Stream::timedPeek() {
 
 // returns peek of the next digit in the stream or -1 if timeout
 // discards non-numeric characters
-int Stream::peekNextDigit() {
+int Stream::peekNextDigit(bool detectDecimal) {
     int c;
     while(1) {
         c = timedPeek();
-        if(c < 0)
-            return c;  // timeout
-        if(c == '-')
+        if( c < 0 || // timeout
+            c == '-' ||
+            ( c >= '0' && c <= '9' ) ||
+            ( detectDecimal && c == '.' ) ) {
             return c;
-        if(c >= '0' && c <= '9')
-            return c;
+        }
         read();  // discard non-numeric
     }
 }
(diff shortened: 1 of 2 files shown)


earlephilhower:
Apropos of nothing, it's somewhat amusing that the `decimal separator` is hardcoded to a `.` by Arduino.cc even though in the Italian locale they use commas for the decimal separator.  😆

d-a-v:
It's without any regret that I copied this patch.
Decimal commas are hurting computing.

(Note: French decimal separator is also a comma)

PR #8796, opened 2023-01-09
Pointer<->Int casts are not constexpr

mcspr:
Gcc is more lenient with this, Clang is not

### DIFF
--- a/cores/esp8266/esp_priv.h
+++ b/cores/esp8266/esp_priv.h
@@ -23,9 +23,8 @@
 
 #if defined(CORE_MOCK)
 
-constexpr bool __byteAddressable(const void* addr)
+constexpr bool __byteAddressable(const void*)
 {
-    (void)addr;
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


cores/esp8266/esp_priv.h
```diff
@@ -34,7 +33,7 @@ constexpr bool __byteAddressable(const void* addr)
 #include <sys/config.h>
 
 // returns true when addr can be used without "pgm_" functions or non32xfer service
-constexpr bool __byteAddressable(const void* addr)
+inline bool __byteAddressable(const void* addr)
 {
     return addr < (const void*)(XCHAL_DATARAM0_VADDR + XCHAL_DATARAM0_SIZE);
```
mcspr, line 38: [`-Winvalid-constexpr`](https://clang.llvm.org/docs/DiagnosticsReference.html#winvalid-constexpr)
> error: constexpr function never produces a constant expression [-Winvalid-constexpr]
> note: cast that performs the conversions of a reinterpret_cast is not allowed in a constant expression

(where ^ points at XCHAL expression)

PR #8818, opened 2023-01-18
Properly handle u8 pointers when assigning and comparing

### LINKED ISSUE #8817: IPAddress::operator= needs 32-bit aligned pointer
### Basic Infos

#### Platform

- Core Version:			[Current master / 2.7.4]
- Operating System:		[Windows|Ubuntu]

### Settings in IDE

- Flash Size: 		[4MB/1MB]
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	[40Mhz]
- Upload Using:		[OTA|SERIAL]

### Problem Description

```
LoadStoreAlignmentCause: Load or store to an unaligned address
  epc1=0x402a1358 in IPAddress::operator=(unsigned char const*) at ??:?
en deze stacktrace:
0x4029280d in String::operator=(char const*) at ??:?
0x401012bb in umm_free_core at umm_malloc.cpp:?
0x401015a2 in malloc at ??:?
0x402d1866 in wifi_get_opmode at ??:?
0x402a2568 in spiffs_cache_page_get_by_fd at ??:?
0x402a2568 in spiffs_cache_page_get_by_fd at ??:?
0x402a2568 in spiffs_cache_page_get_by_fd at ??:?
0x402a2568 in spiffs_cache_page_get_by_fd at ??:?
0x4024821f in setConnectionSpeed() at ??:?
0x40248dc2 in prepareWiFi() at ??:?
0x40241468 in WiFiEventData_t::markWiFiBegin() at ??:?
0x40248f29 in AttemptWiFiConnect() at ??:?
0x40205e24 in ESP8266WiFiSTAClass::status() at ??:?
0x40248084 in wifiAPmodeActivelyUsed() at ??:?
0x4024909c in WiFiConnectRelaxed() at ??:?
0x40245ce9 in NetworkConnectRelaxed() at ??:?
0x4024bf23 in ESPEasy_setup() at ??:?
0x40215084 in setup at ??:?
0x40293ddb in loop_wrapper() at core_esp8266_main.cpp:?
0x40100469 in cont_wrapper at ??:?
```

See the assignment operator and other functions where the `const uint8_t*` is cast to an `const uint32_t*`

https://github.com/esp8266/Arduino/blob/f06710eb6e1559b8a0035d02c0c0fc0c860b821f/cores/esp8266/IPAddress.cpp#L118-L122

https://github.com/esp8266/Arduino/blob/f06710eb6e1559b8a0035d02c0c0fc0c860b821f/cores/esp8266/IPAddress.cpp#L130-L132

The call is simply like this:

```c++
  const IPAddress ip     = Settings.IP;
  const IPAddress gw     = Settings.Gateway;
  const IPAddress subnet = Settings.Subnet;
  const IPAddress dns    = Settings.DNS;
```

Where all these are part of a larger struct, which begins like this:

```c++
  unsigned long PID = 0;
  int           Version = 0;
  int16_t       Build = 0;
  uint8_t       IP[4] = {0};
  uint8_t       Gateway[4] = {0};
  uint8_t       Subnet[4] = {0};
  uint8_t       DNS[4] = {0};
```

As can be seen, these are indeed not 32-bit aligned.

I think the build toolchain may recently have been updated, as this struct and the code that's now failing has not been changed for at least 4 years.

The same code in `IPAddress` has not been changed for a while as the current master branch has the same code for these functions as the 2.7.4 we're using.

mcspr:
* Simplify ctors and operator=, use a common code paths instead of special handling here and there
* Fix #8817 u8<->u32 assignment with pointers, copy before using u8 data
* Do not use raw\_address() internally

raw6() is one thing left over, we sometimes have `u8 scope` at the end so it is not techically valid code to reinterpret like that. alignment works out, though
tbh I'd prefer we return a real value instead of pointer for our internal usage, and leave ptr access for Arduino compatibility (meaning, std::array<uint8_t, 4> and std::array<uint16_t, 8> / std::array<uint8_t, 16> and operate on that vs. weird overloads all over the place and no size checks)

### DIFF
--- a/cores/esp8266/Client.h
+++ b/cores/esp8266/Client.h
@@ -38,7 +38,13 @@ class Client: public Stream {
         virtual void stop() = 0;
         virtual uint8_t connected() = 0;
         virtual operator bool() = 0;
-
+    protected:
+        uint8_t* rawIPAddress(IPAddress& addr) {
+            return addr.raw_address();
+        }
+        const uint8_t* rawIPAddress(const IPAddress& addr) {
+            return addr.raw_address();
+        }
 };
 
 #endif
--- a/cores/esp8266/Client.h
+++ b/cores/esp8266/Client.h
@@ -38,13 +38,7 @@ class Client: public Stream {
         virtual void stop() = 0;
         virtual uint8_t connected() = 0;
         virtual operator bool() = 0;
-    protected:
-        uint8_t* rawIPAddress(IPAddress& addr) {
-            return addr.raw_address();
-        }
-        const uint8_t* rawIPAddress(const IPAddress& addr) {
-            return addr.raw_address();
-        }
+
 };
 
 #endif
--- a/cores/esp8266/IPAddress.cpp
+++ b/cores/esp8266/IPAddress.cpp
@@ -41,24 +41,14 @@ bool IPAddress::isSet () const {
 }
 
 IPAddress::IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet) {
-    setV4();
-    (*this)[0] = first_octet;
-    (*this)[1] = second_octet;
-    (*this)[2] = third_octet;
-    (*this)[3] = fourth_octet;
-}
+    uint8_t addr[] {
+        first_octet,
+        second_octet,
+        third_octet,
+        fourth_octet,
+    };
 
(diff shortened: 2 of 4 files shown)


d-a-v (approved):

TD-er:
Just as a compliment:
This was an extremely swift colaboration. Within 1h after reporting, a possible fix was already known and now already being merged.
Thums up!

PR #8828, opened 2023-01-24
A new approach for erasing WiFi Settings

mhightower83:
Add hardware reset function - simulates `EXT_RST` via accelerated HWDT.

Add reset selection to `ESP.eraseConfig()` for calling hardware reset after erasing the WiFi Settings.

Update ArduinoOTA to use `ESP.eraseConfig(true)`

Externalized a function to Stop WiFi, Erase WiFi Settings, and Reset, `ArduinoOTA.eraseConfigAndReset()`.

Add OTA examples to illustrate using changes.


libraries/ArduinoOTA/ArduinoOTA.cpp
```diff
@@ -320,18 +314,27 @@ void ArduinoOTAClass::_runUpdate() {
...
+      if (_eraseConfig) {
+        eraseConfigAndReset();  // returns on failure
+        //C What is the best action to take on failure?
+        //C  1) On failure, we could invalidate eboot_command buffer -
+        //C     aborting the flash update.
+        //C  2) Just ignore it and restart.
+        //C  3) Retry forever
+        if (_error_callback) {
+          _error_callback(OTA_ERASE_SETTINGS_ERROR);
+        }
```
mhightower83, line 334: At failure, there are no good choices. 
I picked 3 to retry forever. 
If they power cycle, the download is aborted.

mcspr: Why a separate condition & callback though? We can retrieve error id

mcspr (commented):
Can you expand *why* we want sys reset happen that way vs. `__builtin_trap()` or similar? (which iirc works in this case. or, why it is not appropriate to trap vs. using this handler)
Do we want trick SDK into thinking we pressed RST?

mcspr:
And... why not in Updater? Don't we want this feature across *all* updater users?

libraries/ArduinoOTA/ArduinoOTA.cpp
mhightower83: I don't understand your reference. I am using the existing error callback with a new error number to identify the situation.

mcspr: (review UI sent a draft :/)

Why even mention it? Since we want this to be done, probably unroll is a better option than leave things in an indeterminate state?
Another part - it is the only state that continues execution. We still have eboot part like you said, even after ::end()ing the instance (suppose that from user pov it is an option to recover).

mhightower83:
> And... why not in Updater? Don't we want this feature across all updater users?

If I followed this right, the updater action has to be driven from the library side. Where in the updater are you thinking this can be placed?

> Can you expand _why_ we want sys reset happen that way vs. `__builtin_trap()` or similar? (which iirc works in this case. or, why it is not appropriate to trap vs. using this handler) Do we want trick SDK into thinking we pressed RST?

Yes. I didn't want to give the SDK the opportunity to do the wrong thing. I am trying to create a situation where the SDK will have no reason and minimum opportunity to write to the freshly erase data area. (Which it rightfully thinks it owns.) We are possibly going to load a different SDK.  I didn't want to go down the path of a graceful shutdown and reboot with the SDK in full control. I know it is very heavy-handed to force a reboot this way. I felt it was the safer path.

Until I found a method to force a quick hardware reset through software, I rejected this approach. Without it, we would need to force a regular HWDT.  

`__builtin_trap()` compiles to a `break   1, 15`. The NON-OS SDK's default breakpoint handler is to loop on a `waiti 2` until the HWDT kinks in after 8 seconds. I guess this is an option (or a simple `while(true);`) if the 8-second delay is not an issue; it would save a little code space.

Let me know if you prefer the 8-second HWDT method.

mcspr:
I was thinking of something like existing setAsync or installSignature, just another Updater method to set a boolean flag
```cpp
Update.setEraseConfigBeforeReboot(true);`
```

Explicit code is probably better then, lets keep the hardware\_reboot as-is

mhightower83:
There is an odd behavior with `eboot` copy flash and load when HWDT is involved in any way. 
`eboot` plays with disable/enable WDT in the copy flash function. This leaves the WDT running when it was originally off. After a HWDT the WDT does not appear to be configured; however, a soft restart has it configured.
What I see is that there is an extra HWDT event after the flash copy has finished. After the extra HWDT event, the WD is off again and the flash can load without obstruction.

I need to verify more. I think the WDT is always off at the start of `eboot`

mhightower83:
> I was thinking of something like existing setAsync or installSignature, just another Updater method to set a boolean flag
> 
> ```c++
> Update.setEraseConfigBeforeReboot(true);`
> ```

I started to do something like that. I had trouble putting stuff in `UpdaterClass` that needed to call library methods. 
To complete an `Update.setEraseConfigBeforeReboot(true)`, I would need to call two library functions: `WiFi.mode(WIFI_OFF)` and `ESP.eraseConfig(true)` from the `UpdaterClass`.

mcspr:
Out of scope here, but I see https://github.com/esp8266/Arduino/pull/7979 as a possible solution to that
(it was neglected for some time, though, but overall idea of moving sleep modes into ESP should be merged at some point)

3.1.2 is slightly stretched, so holding for a bit.

libraries/ArduinoOTA/ArduinoOTA.h
```diff
@@ -45,7 +52,12 @@ class ArduinoOTAClass
     void setPasswordHash(const char *password);
 
     //Sets if the device should be rebooted after successful update. Default true
-    void setRebootOnSuccess(bool reboot);
+    //"eraseConfig" selects to erase WiFi Settings in conjunction with a reboot
+    //after a successful update. Default OTA_ERASE_CFG_NO - Legacy behavior
+    void setRebootOnSuccess(bool reboot, ota_erase_cfg_t eraseConfig = OTA_ERASE_CFG_NO);
+
+    //Sets flag to erase WiFi Settings at reboot/reset.
+    void setEraseConfig(bool eraseConfig = true);
```
mcspr, line 60: ref. previous discussion
`setRebootOnSuccess(bool)` remains as is, `setEraseConfig(ota_erase_cfg_t)` setting erase flag

cores/esp8266/Esp.h
```diff
@@ -210,7 +210,20 @@ class EspClass {
...
+
+          As a precaution, since this operation erases the running SDKs flash
+          configuration space, use reset flag "true" with eraseConfig. Also, for
+          additional protection, call "WiFi.mode(WIFI_OFF)" before calling.
+
+          If you need to erase "WiFi Settings" and reboot consider using
+          "ArduinoOTA.eraseConfigAndReset()" it handles shutting down WiFi
+          before the erase.
+        */
+        static bool eraseConfig(bool reset = false);
```
mcspr, line 226: Same sentiment as in ArduinoOTA, two separate funcs?
`eraseConfig()` and `eraseConfigAndReset()`

mcspr (commented):
Pending FPM change, going to rewrite into Updater part some time later

libraries/ArduinoOTA/ArduinoOTA.cpp
mhightower83: https://github.com/esp8266/Arduino/blob/0e1059d60f3f3799589b92469609cf485e2b275c/libraries/ArduinoOTA/ArduinoOTA.cpp#L328-L346

When enabling erase WiFi settings, the user does so by selecting the fail option ignore or abort. Does that satisfy your concerns?

libraries/ArduinoOTA/ArduinoOTA.h
mhightower83: done.

cores/esp8266/Esp.h
mhightower83: done

cores/esp8266/hardware_reset.h
```diff
@@ -0,0 +1,12 @@
+#ifndef HARDWARE_RESET_H
+#define HARDWARE_RESET_H
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+[[noreturn]] extern void hardware_reset(void);
```
mcspr, line 7: No need for `extern`

---

`__attribute__((noreturn))`?
at the end, once in header https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-noreturn-function-attribute

In case of `__cplusplus`, were are ok and can use `[[noreturn]]`
Without `__cplusplus`, there is no `noreturn` but `_Noreturn` (or, `noreturn` w/ `stdreturn.h`, that disables itself in C++ mode).

But, I am still having warnings when using either one, so not really sure whats the deal here
> warning: 'noreturn' attribute ignored [-Wattributes]
> warning: '_Noreturn' attribute ignored [-Wattributes]

>  7 | [[noreturn]] void testme(void);
      | ^

mcspr (approved):

PR #8837, opened 2023-01-29
Update documentation and arduino debug options

mhightower83:
Add a section in FAQ for `Improving Exception Decoder Results`
Add a description for the `Debug Optimization` menu to `ideoptions.rst`
Added some connecting links between related topics.

Update to `boards.txt.py` to generate `boards.txt` with a `Debug Optimization` menu.
`platform.txt` has optimization defaults set by `build.debug_optim=-Os`
The `Debug Optimization` menu selection overrides that.
Then an `-O` option in `build.opt` takes priority over those.

No changes to `platformio-build.py`.

Status: ready
edited to sync with the current state of the PR


jjsuwa-sys3175:
Hmm, maybe the compiler should disable both leaf function handling and sibling call optimization by default when '-Og' is specified.  Because the GCC manual entry for that option says first:
> Optimize debugging experience.  '-Og' should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience.

d-a-v:
With `-Og` instead of `-Os` (without other debugging option), tried with the `FSBrowser` example, I get
- +600 Bytes in IRAM text
- +24KB in flash

With `-Os -fno-optimize-sibling-calls` I get
- +18 bytes in IRAM text
- +500 bytes in flash

Is `-Og` preferred when debug option on Serial are enabled ?
Or is `-fno-optimize-sibling-calls` sufficient and document instead how to enable `-Og` in the gdb help section ?

tools/platformio-build.py
```diff
@@ -205,6 +205,14 @@ def scons_patched_match_splitext(path, suffixes=None):
     LIBPATH=[NONOSDK_LIBPATH],
 )
 
+#
+# debug
+#
+for define in env["CCFLAGS"]:
```
mcspr, line 211: [definitions are in cppdefines](https://scons.org/doc/latest/HTML/scons-user.html#cv-CPPDEFINES)
could use the flattened ones, or loop env[...] but note of possible types (tuple or string)

jjsuwa-sys3175:
@d-a-v :
> Is `-Og` preferred when debug option on Serial are enabled ?
> Or is `-fno-optimize-sibling-calls` sufficient and document instead how to enable `-Og` in the gdb help section ?

Well, just thought that using '-Og' would generally make debugging less disturbing.

From GCC manual again:
> -Og
> - Optimize debugging experience.
> '-Og' should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience.
> It is a better choice than '-O0' for producing debuggable code because some compiler passes that collect debug information are disabled at '-O0'.
> Like '-O0', '-Og' completely disables a number of optimization passes so that individual options controlling them have no effect.
> Otherwise '-Og' enables all '-O1' optimization flags **except for those that may interfere with debugging**:
>   - -fbranch-count-reg -fdelayed-branch -fdse -fif-conversion -fif-conversion2 -finline-functions-called-once -fmove-loop-invariants -fssa-phiopt -ftree-bit-ccp -ftree-dse -ftree-pta -ftree-sra

mhightower83:
With other compilers, I have used you get a better matchup between code addresses and the source code when you build with debug option. Would that be true here with -Og?

jjsuwa-sys3175:
I haven't compared the output with 'other compilers'... but I can say that with '-Og' GCC disables some optimizations that make debugging difficult (both machine-independent and dependent, eg. 'constant synthesis' in Xtensa)
'-Og' basically implies '-O1', so optimization options not included in '-O1', such as '-foptimize-sibling-calls', are disabled, but register saving of leaf functions still can be omitted (there is no option to control this separately).
Fortunately, forcing **all** functions to save register A0 on '-Og' can be easily implemented with a few lines of patch to the Xtensa port.

mhightower83:
> Fortunately, forcing all functions to save register A0 on '-Og' can be easily implemented with a few lines of patch to the Xtensa port.

Sounds good to me; however, I have to defer to others for such a change. @d-a-v @earlephilhower

d-a-v:
I am personally looking forward for such improvement with the debugging experience.
However, I think that `-Og` *by default* when serial log is enabled makes a too large change (similar to what enabling IPv6 brings and this is one of the main reason why is it not enabled by default).
What can be proposed is to add two more debugging port options with the following:
```diff
diff --git a/tools/boards.txt.py b/tools/boards.txt.py
index 89380d20..52300a11 100755
--- a/tools/boards.txt.py
+++ b/tools/boards.txt.py
@@ -1353,11 +1353,15 @@ def all_debug ():
     options += [ listsingle ]
     debugmenu = collections.OrderedDict([
             ( '.menu.dbg.Disabled', 'Disabled' ),
-            ( '.menu.dbg.Disabled.build.debug_port', '' ),
+            ( '.menu.dbg.Disabled.build.debug_port', '-Os' ),
             ( '.menu.dbg.Serial', 'Serial' ),
-            ( '.menu.dbg.Serial.build.debug_port', '-DDEBUG_ESP_PORT=Serial' ),
+            ( '.menu.dbg.Serial.build.debug_port', '-DDEBUG_ESP_PORT=Serial -Os -fno-optimize-sibling-calls' ),
             ( '.menu.dbg.Serial1', 'Serial1' ),
-            ( '.menu.dbg.Serial1.build.debug_port', '-DDEBUG_ESP_PORT=Serial1' ),
+            ( '.menu.dbg.Serial1.build.debug_port', '-DDEBUG_ESP_PORT=Serial1 -Os -fno-optimize-sibling-calls' ),
+            ( '.menu.dbg.Serial-dbg', 'Serial (full debug)' ),
+            ( '.menu.dbg.Serial-dbg.build.debug_port', '-DDEBUG_ESP_PORT=Serial -Og' ),
+            ( '.menu.dbg.Serial1-dbg', 'Serial1 (full debug)' ),
+            ( '.menu.dbg.Serial1-dbg.build.debug_port', '-DDEBUG_ESP_PORT=Serial1 -Og' ),
             ( '.menu.lvl.None____', 'None' ),
             ( '.menu.lvl.None____.build.debug_level', '' ),
         ])
```
Giving this Arduino IDE 1.8 menu:
image

**additionally**: options `-Os` from three lines in `platform.txt` have to be removed


With these changes, we have 
`-Os` by default
`-Os -fno-optimize-sibling-calls` when one of the serial ports is enabled
`-Og` when one of the serial ports w/`full debug` option is enabled



**additionnally**: `platformio-build.py` would have yet to be updated too

platform.txt
```diff
@@ -70,24 +70,24 @@ compiler.path={runtime.tools.xtensa-lx106-elf-gcc.path}/bin/
 compiler.sdk.path={runtime.platform.path}/tools/sdk
 
 compiler.libc.path={runtime.platform.path}/tools/sdk/libc/xtensa-lx106-elf
-compiler.cpreprocessor.flags=-D__ets__ -DICACHE_FLASH -U__STRICT_ANSI__ -D_GNU_SOURCE -DESP8266 {build.opt.flags} "-I{compiler.sdk.path}/include" "-I{compiler.sdk.path}/{build.lwip_include}" "-I{compiler.libc.path}/include" "-I{build.path}/core"
+compiler.cpreprocessor.flags=-D__ets__ -DICACHE_FLASH -U__STRICT_ANSI__ -D_GNU_SOURCE -DESP8266 -Os {build.opt.flags} "-I{compiler.sdk.path}/include" "-I{compiler.sdk.path}/{build.lwip_include}" "-I{compiler.libc.path}/include" "-I{build.path}/core"
```
mhightower83, line 73: @d-a-v I would like to suggest a slight modification. Moving/consolidating the three `-Os` to here permits override from `build.opt.flags` as well as `build.debug_port. From what I have read, GCC uses the last optimization specified.

With the above change, I would leave `( '.menu.dbg.Disabled.build.debug_port', '' )` as is so that `build.opt` would have more control for defining global optimization in the absence of the `Debug port` specification.

doc/faq/a06-global-build-options.rst
```diff
@@ -1,5 +1,3 @@
-:orphan:
-
```
mcspr, line 2: > /home/runner/work/Arduino/Arduino/doc/faq/a06-global-build-options.rst: WARNING: document isn't included in any toctree

mhightower83: I maintain this file as a `.md` on my machine and use `pandoc` to turn it into a `.rst` for a PR. So far I have noticed the `:orphan:` line seemed to magically appear later after a merge, 

Are you indicating that this `:orphan:` line should always be added?

mcspr: Missed c/p
https://github.com/esp8266/Arduino/actions/runs/4097689448/jobs/7066241098#step:4:222

Sphinx build thinks this is something worth warning about
#8686 opted out to use this tag, but using this page in a toctree somewhere would fix it
(also see other faq/*.rst)

mhightower83:
```cpp
// Without "-fno-ipa-pure-const", a complete caller list is not on the stack for decoding.
#pragma GCC optimize("Og,no-ipa-pure-const")

// Or this set for -Os
//#pragma GCC optimize("Os,no-inline,no-optimize-sibling-calls,no-ipa-pure-const")

void hideaway() {
    asm volatile ("" ::: "a0", "memory");
    while(true);
}

void setup() {
  Serial.begin(115200);
  delay(200);  
  Serial.printf("\r\n\r\nNow waiting forever ...\r\n");
  hideaway();
}
void loop() {}
```
I don't have a lot of experience with the new optimization options to know if this would be a commonly useful option or just a wack-a-mole one-off. Unless someone has some insight into this flag, I am inclined to not change the PR. I think as experience is gained with changing optimization for improved decoding, the refinement list will expand. Maybe later add an `Optimum-Plus` with all the extras flags that prove useful.

PR #8839, opened 2023-01-30
Postmortem improvements

mhightower83:
Improve divide by 0 reporting to point back at the divide function caller

Improve stack offset adjustments to be independent of `__wrap_system_restart_local()`'s  stack frame size.

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -110,10 +110,27 @@ static void cut_here() {
     ets_putc('\n');
 }
 
-void __wrap_system_restart_local() {
-    register uint32_t sp asm("a1");
-    uint32_t sp_dump = sp;
+static void postmortem_report(uint32_t sp_dump) __attribute__((used));
+/*
+  Add some assembly to grab the stack pointer and pass it as an argument before
+  it grows for the target function. Should stabilize the stack offsets, used to
+  find the relevant stack content for dumping.
+*/
+extern "C" void __wrap_system_restart_local(void);
+asm(
+    ".section     .text.__wrap_system_restart_local,\"ax\",@progbits\n\t"
+    ".literal_position\n\t"
+    ".align       4\n\t"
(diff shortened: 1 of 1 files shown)


cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -110,10 +110,27 @@ static void cut_here() {
     ets_putc('\n');
 }
 
-void __wrap_system_restart_local() {
-    register uint32_t sp asm("a1");
-    uint32_t sp_dump = sp;
-
+static void postmortem_report(uint32_t sp_dump) __attribute__((used));
```
mcspr, line 113: Is it dropped, somehow?
I don't see any errors on my side after removing this and adding `static` below

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -174,17 +199,32 @@ void __wrap_system_restart_local() {
...
+        // 256 Exception vector reserves stack space
+        //     filled in by "C" wrapper handler
+        //  16 Handler level 1 - enable icache
+        //  64 Handler level 2 - exception report
+        offset = 64 + 16 + 256;
     }
     else if (rst_info.reason == REASON_WDT_RST) {
         offset = 0x10;
     }
+    else if (rst_info.reason == REASON_USER_SWEXCEPTION_RST) {
```
mcspr, line 225: Also update ^ `REASON_WDT_RST` to decimal?

mcspr (commented):
lgtm based on repeating crashes from #8670

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: done

cores/esp8266/core_esp8266_postmortem.cpp
mhightower83: Updated. There is some scenario where I need this, but not here. :confused:

PR #8845, opened 2023-02-02
Wrappers allowing to compile and run esp32-arduino basic examples

d-a-v:
Tested with following esp32-arduino [examples](https://github.com/espressif/arduino-esp32/tree/master/libraries/WebServer/examples):
- AdvancedWebServer 
- FSBrowser  
- HelloServer  
- SimpleAuthentification

Compatibility files are stored into a new `WiFi/` library because of name collision with legacy Arduino-avr WiFi library which priority would otherwise take over.


d-a-v:
@JAndrassy your #9044 reminds me of this.
Is your work around Arduino networking libraries portability also extends to higher level APIs ?

JAndrassy:
> Is your work around Arduino networking libraries portability also extends to higher level APIs ?

no.
for networking libraries by Arduino there is no WebServer, mDNS .... 
There is [something](https://blog.arduino.cc/2023/11/10/introducing-arduinos-new-storage-libraries-streamline-data-management-for-your-projects/) very new for storage

d-a-v:
It seems we can make an adapter for SDFat

PR #8850, opened 2023-02-05
make http-server less verbose in debug mode

d-a-v:


libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -296,12 +296,30 @@ void ESP8266WebServerTemplate<ServerType>::handleClient() {
...
-    _currentStatus==HC_WAIT_CLOSE?"wait-close":
-    "??");
+#ifdef DEBUG_ESP_HTTP_SERVER
+
+  struct compare_s
+  {
+    uint8_t connected;
+    int available;
+    HTTPClientStatus status;
+    bool operator != (const compare_s& o) { return memcmp(this, &o, sizeof(o)) != 0; }
```
mcspr, line 306: Padding? Just write every member instead of comparing raw mem?
*(or someone copies this into their virtual class accidentally :>)*

d-a-v: I wonder why c++ does not define a default operator by implicitly calling it on every members.

mcspr: https://en.cppreference.com/w/cpp/language/default_comparisons (c++20)
https://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0515r3.pdf

PR #8851, opened 2023-02-07
Have Postmortem report address interrupted by Soft WDT

mhightower83:
Update postmortem to report location interrupted by Software WDT.

In the NON-OS SDK, Software WDT is implemented as a level-1 interrupt. 
In the ISR handler `pp_soft_wdt_feed_local`, various exception registers are saved to System RTC memory.

I have hesitated and forgotten about doing this PR several times. 
I keep asking the question what am I missing?
Is there a situation where this will not be accurate?


d-a-v (approved):
Valuable info with random examples doing soft WDT

Issue #8856, opened 2023-02-13
Boards Manager upgrade ESP8266 Core 3.0.2 to 3.1.0/3.1.1 fails to compile - error missing CommonHFile.h


TobbeG:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [X] I have filled out all fields below.

#### Platform
- Hardware:			       ESP-01
- Core Version:		       All works ok with 3.0.2 but fails to compile with 3.1.0/3.1.1
- Development Env:	          Arduino IDE 1.8.16
- Operating System:		Windows11

### Settings in IDE

- Module:  			Generic ESP8266 Module
- Flash Mode:			qio
- Flash Size: 		       4MB or 1MB
- lwip Variant:                  Higher Bandwidth
- Reset Method:		[ck|nodemcu]
- Flash Frequency:	        40Mhz
- CPU Frequency:		160MHz
- Upload Using:		SERIAL
- Upload Speed:              921600

### Problem Description
Cannot compile code with core version 3.1.0 or 3.1.1
Core 3.0.2 works fine

Detailed problem description goes here.
When compling I get error message of missing file:       CommonHFile.h

Cannot even compile simples Blink.c example.

### Debug Messages
ERROR MESSAGE
cc1plus.exe: fatal error: C:\Users\Thorbjörn Geiser O\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\3.1.0\cores\esp8266\CommonHFile.h: No such file or directory
compilation terminated.
exit status 1
Error compiling for board Generic ESP8266 Module.

Rainbow-Cai:
Hi @igrr  , Could you please update this question?

Rainbow-Cai:
Hi  @TobbeG   This is a third party project, not something that is supported by Espressif.  esp8266/Arduino is a third party project, no one in Espressif is working on it.

mcspr:
Duplicate of #8811?
More build log lines please? Is it python error, compiler one, arduino-{builder,cli}?

mcspr:
Please show the full build log

[...]

PR #8860, opened 2023-02-17
Resolve Windows path problems

mhightower83:
Moved up `os.makedirs` for `./core/` directory.
Strange problem creating files with file paths with spaces. 
Strange that "create file" would work when the path did not contain spaces and the last folder of the path hadn't been created.

Added try/except on main to commit print buffer on traceback for context.

Additional issues with diacritics and locale character encoding for shell vs source code.
`build.opt` is written with the same encoding as the shell; however, the data read from the `Sketch.ino.global.h` is `UTF-8`.

should resolve https://github.com/esp8266/Arduino/issues/8856#issue-1582158038

### DIFF
--- a/tools/mkbuildoptglobals.py
+++ b/tools/mkbuildoptglobals.py
@@ -192,6 +192,8 @@ import sys
 import textwrap
 import time
 
+import locale
+
 # Need to work on signature line used for match to avoid conflicts with
 # existing embedded documentation methods.
 build_opt_signature = "/*@create-file:build.opt@"
@@ -296,14 +298,22 @@ def copy_create_build_file(source_fqfn, build_target_fqfn):
                 pass
     return True     # file changed
 
-
 def add_include_line(build_opt_fqfn, include_fqfn):
+    # Given that GCC will insert (or process as if interted) these lines into
+    # the command line. I assume that the contents of @file need to be encoded
+    # to match that of the shell running GCC runs. I am not 100% sure this API
+    # gives me that, but it appears to work.
+    #
+    # However, elsewhere when dealing with source code we continue to use 'utf-8',
+    # ref. https://gcc.gnu.org/onlinedocs/gcc-4.1.2/cpp/Character-sets.html
+    shell_encoding = locale.getdefaultlocale()[1]
     if not os.path.exists(include_fqfn):
         # If file is missing, we need an place holder
-        with open(include_fqfn, 'w', encoding="utf-8"):
+        with open(include_fqfn, 'w', encoding=shell_encoding):
             pass
-        print("add_include_line: Created " + include_fqfn)
-    with open(build_opt_fqfn, 'a', encoding="utf-8") as build_opt:
+        print_msg("add_include_line: Created " + include_fqfn)
+
+    with open(build_opt_fqfn, 'a', encoding=shell_encoding) as build_opt:
         build_opt.write('-include "' + include_fqfn.replace('\\', '\\\\') + '"\n')
 
 def extract_create_build_opt_file(globals_h_fqfn, file_name, build_opt_fqfn):
@@ -610,6 +620,9 @@ def main():
     global docs_url
     global debug_enabled
     num_include_lines = 1
+
+    def_locale = locale.getdefaultlocale()
+    print_msg(f'default locale: {def_locale}')
 
     args = parse_args()
     debug_enabled = args.debug
--- a/tools/mkbuildoptglobals.py
+++ b/tools/mkbuildoptglobals.py
@@ -612,14 +612,13 @@ def parse_args():
     # ref no '--n' parameter, https://stackoverflow.com/a/21998252
 
 
-def get_encoding():
-    try:
-        from locale import getencoding
-    except ImportError:
-        # https://stackoverflow.com/a/23743499
-        return locale.getpreferredencoding(False) or locale.getdefaultlocale()[1]
-
-    return locale.getencoding()
+# retrieve *system* encoding, not the one used by python internally
+if sys.version_info >= (3, 11):
+    def get_encoding():
+        return locale.getencoding()
+else:
+    def get_encoding():
+        return locale.getdefaultlocale()[1]
 
 
 def show_value(desc, value):
--- a/tools/mkbuildoptglobals.py
+++ b/tools/mkbuildoptglobals.py
@@ -187,6 +187,7 @@ from shutil import copyfile
 import glob
 import os
 import platform
+import traceback
 import sys
 import textwrap
 import time
@@ -304,7 +305,6 @@ def add_include_line(build_opt_fqfn, include_fqfn):
         print("add_include_line: Created " + include_fqfn)
     with open(build_opt_fqfn, 'a', encoding="utf-8") as build_opt:
         build_opt.write('-include "' + include_fqfn.replace('\\', '\\\\') + '"\n')
-
 
 def extract_create_build_opt_file(globals_h_fqfn, file_name, build_opt_fqfn):
     """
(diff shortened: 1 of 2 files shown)


mcspr:
```python
>>> locale.getdefaultlocale()
<stdin>:1: DeprecationWarning: Use setlocale(), getencoding() and getlocale() instead
('en_US', 'UTF-8')
```

For sizes .py, I thought POpen univeral\_newlines should already default to locale setting by wrapping input with TextIOWrapper. 
https://docs.python.org/3/library/io.html#text-encoding

I guess we do need to source some fixed locale / encoding value by external means, but what are values for the rest of stdlib functions? Why it breaks on buildopts we understand, but where does the size .py issue come from? What are default locale & encoding values for your environment from sys / locale?

---

Some external references, since I am pretty sure both manage to work ok with *most* installations. PIP change was a pretty recent one, though

https://github.com/platformio/platformio-core/blob/ec5bf1b5e78cf6baa180e8efb51db4e9a30e5546/platformio/compat.py, hover 'get_filesystem_encoding' and 'get_locale_encoding' and see 'References'
https://github.com/pypa/pip/search?q=sys.getfilesystemencoding
https://github.com/pypa/pip/search?q=locale.getpreferredencoding

tools/mkbuildoptglobals.py
```diff
@@ -609,8 +616,19 @@ def main():
...
+    global default_locale
     num_include_lines = 1
 
+    # Given that GCC will handle lines from an @file as if they were on
+    # the command line. I assume that the contents of @file need to be encoded
+    # to match that of the shell running GCC runs. I am not 100% sure this API
+    # gives me that, but it appears to work.
+    #
+    # However, elsewhere when dealing with source code we continue to use 'utf-8',
+    # ref. https://gcc.gnu.org/onlinedocs/gcc-4.1.2/cpp/Character-sets.html
```
mcspr, line 628: should link to latest?
https://gcc.gnu.org/onlinedocs/cpp/Character-sets.html

pending locale explanation from comment. afaik, we just have to use what system uses for fs, why I mentioned sys.getfilesystemencoding(); mingw -> gcc layer is expectedly using fopen() as-is, so there have to be some kind of raw bytes in the correct encoding for sys api to work

mhightower83:
```
locale.getdefaultlocale()                ('en_US', 'cp1252')
sys.getfilesystemencoding()              utf-8
sys.getdefaultencoding()                 utf-8
locale.getpreferredencoding(False)       cp1252
sys.stdout.encoding                      cp1252
locale.setlocale(locale.LC_ALL, None)    C
locale.getlocale()                       (None, None)
locale.setlocale(locale.LC_ALL, '')      English_United States.1252
locale.getlocale()                       ('English_United States', '1252')
```

> \<stdin\>:1: DeprecationWarning: Use setlocale(), getencoding() and getlocale() instead

You must be using a different version of python than the release.
For   `locale.getencoding()`
I get: `AttributeError: module 'locale' has no attribute 'getencoding` 

It is a bit frustrating. Any solution I find on the Internet either doesn't work or will shortly be deprecated and stop working.
I guess `setlocale(), getencoding() and getlocale() ` is supposed to be the right way. I couldn't find any examples using that set. And `getencoding()` doesn't appear to exist in 3.10.6. The docs say they added it in 3.11. They don't leave much time before deprecating stuff.

mcspr: Ugh. So, looking at patch that adds it
https://github.com/python/cpython/commit/677320348728ce058fa3579017e985af74a236d4
Surprising side-effect from `-X utf-8` which we use for sizes.py, that (sort of) explains the difference with the encoding='locale' for TextIO <-> subprocess
(also casually mentioned in docs)

Suppose this difference could just we wrapped with sys.version check at a global level to avoid possible future warnings. It is the right way, just heavily modified version to version :/
```python
>>> sys.version_info
sys.version_info(major=3, minor=7, micro=2, releaselevel='final', serial=0)
>>> sys.version_info >= (3, 7)
True
>>> sys.version_info >= (3, 11)
False
```

tools/mkbuildoptglobals.py
```diff
@@ -605,12 +611,64 @@ def parse_args():
...
 
+
+def get_encoding():
+    try:
+        from locale import getencoding
+    except ImportError:
+        # https://stackoverflow.com/a/23743499
+        return locale.getpreferredencoding(False) or locale.getdefaultlocale()[1]
+
+    return locale.getencoding()
```
mcspr, line 622:
```suggestion
# retrieve *system* encoding, not the one used by python internally
if sys.version_info >= (3, 11):
    def get_encoding():
        return locale.getencoding()
else:
    def get_encoding():
        return locale.getdefaultlocale()[1]
```
?

no block scope in python
getpreferredencoding() would prefer (hm) utf-8 if we happen to use `-X utf-8`, behaviour changes depending on version. new one does not change with mode setting. both can return None aka default for us

tools/mkbuildoptglobals.py
```diff
@@ -187,10 +187,14 @@
 import glob
 import os
 import platform
+import traceback
 import sys
 import textwrap
 import time
 
+import locale
+import codecs
```
mcspr, line 196:
```suggestion
```

tools/sizes.py
```diff
@@ -72,8 +72,10 @@ def get_segment_sizes(elf, path, mmu):
         (".bss", "BSS"),
     )
 
+    import locale
+    shell_encoding = locale.getdefaultlocale()[1]
```
mcspr, line 76: -//- from mkbuildopts
should be the same `get_encoding()` and import at top

tools/mkbuildoptglobals.py
```diff
@@ -295,16 +299,16 @@ def copy_create_build_file(source_fqfn, build_target_fqfn):
                 pass
     return True     # file changed
 
-
 def add_include_line(build_opt_fqfn, include_fqfn):
+    global default_encoding
     if not os.path.exists(include_fqfn):
         # If file is missing, we need an place holder
-        with open(include_fqfn, 'w', encoding="utf-8"):
+        with open(include_fqfn, 'w', encoding=default_encoding):
```
mcspr, line 306: missed one odd encoding=
```suggestion
        with open(include_fqfn, 'w'):
```

mhightower83: What you are suggesting is I didn't need any of this `get_encoding()` logic - tangent I went on. In the absence of the `-X utf8`, I just needed to selectively remove all the `, encoding="utf-8"` options where I needed the system encoding to dominate.

Since merged - do I push an update or need a new PR?

mcspr: > I just needed to selectively remove all the , encoding="utf-8" options where I needed the system encoding to dominate.

nothing is written to file, so no need for encoding= to possibly confuse future reader(s)
python won't re-encode stuff, this is just the thing for read / write through text io wrapper

in absence of -Xutf-8, text io selects something default aka depends on version
https://github.com/python/cpython/blob/cbd192b6ff137aec25913cc80b2bf7bf1ef3755b/Modules/_io/textio.c#L1102-L1141 (our bundled version)
https://github.com/python/cpython/blob/ddb65c47b1066fec8ce7a674ebb88b167f417d84/Modules/_io/textio.c#L1126-L1132 (3.11 on any recent installation)

utf-8 output is ok, reading cli / writing stuff for external tools should have system encoding

mhightower83: > nothing is written to file, so no need for encoding= to possibly confuse future reader(s)

Just for clarification when creating `build.opt`, we are reading `utf-8` from dot-h files and writing with system encoding to `build.opt` otherwise I can get a file not found for a `-include ...` if the file path contains diacritics.

mcspr: also nit:

> diacritics

any time we have characters from codepage different from ansii e.g. `C:\Users\Максім` in cp1251
just my guess that GCC probably does not use win32 `_wfopen` and widechar, so we have to use exact codepage bytes for filepath representation so system exec for GCC gets proper paths

path and text have different code sections dealing with them, just to denote the difference

(I am gonna go read [Python PEP 529](https://peps.python.org/pep-0529/), https://vstinner.github.io/python36-utf8-windows.html series and cpython [unicodeobject.c](https://github.com/python/cpython/blob/main/Objects/unicodeobject.c))

PR #8868, opened 2023-02-24
Update compiler toolchain to 3.2.0

earlephilhower:
Includes compiler updates from @jjsuwa-sys3175


earlephilhower:
Just to be clear, I have not yet tried any built applications with this tool chain.  So don't merge until we get a few positive responses, please.

I'll try and do some testing tonight my time and report back as well.

jjsuwa-sys3175:
Thanks for building the toolchain :)

For the time being, regression tests have been done when merging to gcc mainline, so it shouldn't be completely unusable...

I checked the build of "fsbrowser" and the actual chip operation check for each patch submission.

Notable changes:
- Preliminary -mlra option added (not ready for use yet)
- Always save register A0 (function return address) to stack when -Og is specified, even for leaf functions
- Improved register allocation efficiency by removing RA tweak remnants that no longer worked
- Related to the above, the order of register allocation has changed
- Once register A0 has been saved to the stack in the function prologue, the compiler may assign register A0 to variables or values being computed in the function body

earlephilhower:
I've run about a dozen examples (LittleFS, WiFi, HTTPClient, Arduino core, WebServer) and had no issues, so I think this is good to go whenever you guys want to update the toolchain.

TD-er:
Is LTO also supported with this updated toolchain?
If not, what is needed to support it?
On my ESP32 builds, it makes a difference of roughly 5 - 7% in binary size. (with debug symbols still included)

earlephilhower:
Not supported.  If I remember correctly many years ago igrr tried LTO and had serious issues with the way it handled weak functions.

This is just the latest patches back ported from the upstream GCC later versions with some size and speed optimizations.

TD-er:
> with some size and speed optimizations.

OK, fair enough. Let's hope those size optimizations are already helping out then...

PR #8869, opened 2023-02-25
EspSoftwareSerial 8.0.1: onReceive bug fix and new namespace

dok-net:
Fixed onReceive IRAM cache issues.
Example for power-saving with wakeup from suspend by SW serial RX event.

### DIFF
--- a/libraries/ESP8266WiFi/examples/WiFiTelnetToSerial/WiFiTelnetToSerial.ino
+++ b/libraries/ESP8266WiFi/examples/WiFiTelnetToSerial/WiFiTelnetToSerial.ino
@@ -31,7 +31,7 @@
     SWAP_PINS:
    0: use Serial1 for logging (legacy example)
    1: configure Hardware Serial port on RX:GPIO13 TX:GPIO15
-      and use SoftwareSerial for logging on
+      and use EspSoftwareSerial for logging on
       standard Serial pins RX:GPIO3 and TX:GPIO1
 */
 
@@ -59,7 +59,7 @@
 
 #if SWAP_PINS
 #include <SoftwareSerial.h>
-SoftwareSerial::UART* logger = nullptr;
+EspSoftwareSerial::UART* logger = nullptr;
 #else
 #define logger (&Serial1)
 #endif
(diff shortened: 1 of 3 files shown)


dok-net:
In order to be able to move forward  with the development of this lib, like implementing issue resolutions for strapping pins usage as serial pins, the code is refactored to use the namespace SoftwareSerial, and rename the class SoftwareSerial to UART.
@mcspr There is a conflict with ESP32 Core's outmoded use of #defines for the SERIAL_8N1 etc; I have posted a PR there, but you never know if and when... Perhaps the SoftwareSerial::SERIAL_8N1 should be renamed to SoftwareSerial::CONFIG_8N1, etc. What do you think? It's all a breaking change anyway, but there's no need to make it any harder to seek&replace than necessary. The rename to UART is in the same category, do you think that's still idiomatic?

mcspr:
As previously stated, I'd be very wary with simply dropping stuff without providing any backwards compatibility.

Suppose we have `softwareserial::SoftwareSerial`, global namespace then has `using SoftwareSerial = softwareserial::SoftwareSerial;`

Same goes with config; `using Foo = softwareserial::non_class_enum;` would import things into global ns, see our WiFi overrides for lwip tcp state. Or, another option is to simply have `constexpr int`s, then `constexpr auto SWCONFIG_8N1 = softwareserial::.*::Whatever8N1;`.

(claiming SoftwareSerial for namespace feels weird tbh)

dok-net:
@mcspr You kind of mostly told me what I was hoping to hear :-) I've added the using to map class `SoftwareSerial` to the renamed `UART`. I've renamed the namespace to EspSoftwareSerial. Upper case or no, opinions vary, even I use in turn. About the other identifiers, I have renamed the enums from SWSERIAL_8N1 etc. to plain SERIAL_8N1, just must have missed that. I expect it's not used that often, so our users are invited to update their code if they do.

mcspr:
Any name should be fine, just not the old class one.
Why no `using SWSERIAL_8N1 = EspSoftwareSerial::CONFIG_8N1;` (and the rest)?

We'd wanna a patch release approach here for most identifiers, 3.2.0 come you can remove those if you want.

dok-net:
> Why no `SWSERIAL_8N1 = EspSoftwareSerial::CONFIG_8N1;` (and the rest)?

I haven't explicitly tried now, but I suspect the parity would need the same. So I've reverted to SWSERIAL_8N1 etc. and added a using namespace EspSoftwareSerial. Unless now somebodies code has "UART" in conflict, it seems to be fine. I have undone most of the changes to the ESP8266 core's examples as well.

dok-net:
I bumped up the bug fix release number just to make sure that Platformio doesn't keep using the wrong commit for 8.0.0 - it scans the main branch instead of looking at Github release tags. I like the Arduino library manager better :-)

Issue #8872, opened 2023-02-27
'Reallocating large String(... -> ...)' warning when using WebServer


hmueller01:
Hi,

since PR #8821 I get this output from ESP8266WebServer:
```
[String] 'Content-Ty ... ires: -1
': Reallocating large String(140 -> 162 bytes)
```
It's from this line
https://github.com/esp8266/Arduino/blob/be02af0504869e2bf119a013fe8fd6cf5ed8e411/libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h#L433
if `sendHeader()` is used more than once. Like here:
https://github.com/khoih-prog/ESP_WiFiManager_Lite/blob/085afdbd7ccfb7e46b1ed9f1cf5c50be29cc3223/src/ESP_WiFiManager_Lite.h#L2771

Is there something we can do against it?

d-a-v:
If you are using the git version of this core, can you try with #8873 ?

hmueller01:
As written in #8873 the change in ESP8266WebServer looks good to me.

There is similar issue in [ESP8266HTTPClient](https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266HTTPClient)
https://github.com/esp8266/Arduino/blob/6ad9bf3b07ecbe8dd34623c8f81db3b7614cffbf/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L820
if you add multiple headers ...

mathertel:
The output appears also in the published version 3.1.2 when DEBUG_ESP_PORT is enabled even when Debug Level == none.

Example:

```
[String] 'Content-Ty ... rigin: *
': Reallocating large String(159 -> 183 bytes)
[String] 'Content-Ty ... ep-alive
': Reallocating large String(183 -> 209 bytes)
```

PR #8873, opened 2023-02-28
Avoid large reallocations in webserver's header response

d-a-v:


earlephilhower:
CI seems to be aborting (not code related) frequently right about now.  Let's try to re-run later...

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -420,7 +420,8 @@ void ESP8266WebServerTemplate<ServerType>::stop() {
 }
 
 template <typename ServerType>
-void ESP8266WebServerTemplate<ServerType>::sendHeader(String&& name, String&& value, bool first) {
+template <typename S1, typename S2>
+void ESP8266WebServerTemplate<ServerType>::sendHeader(S1 name, S2 value, bool first) {
```
mcspr, line 424: minor nit

at a glance, we have a bunch of typed pairs that generate multiple emplace\_... for every sendHeader variant
I'd consider using perfect forward and a specific type we already have

```cpp
auto foo(S1&& name, S2&& value, bool condition) {
    auto pair = std::make_pair(
        String(std::forward<S1>(name)),
        String(std::forward<S2>(value)));

    if (condition) {
        foo(std::move(pair));
    } else {
        bar(std::move(pair));
    }
}
```

hmueller01:
@d-a-v I have pulled your branch and did a quick test. Looks good. Thanks for the quick fix!

What I found when more that one client is connecting to the WebServer is this:
```
[String] 'User-Agent ... ome/110.0.': Reallocating large String(143 -> 144 bytes)
[String] 'User-Agent ... 53 Mobile ': Reallocating large String(159 -> 160 bytes)

```
It's surely not releated to this change and I even do not know if it comes from ESPWebServer or our code ...

libraries/ESP8266WebServer/src/ESP8266WebServer-impl.h
```diff
@@ -439,44 +432,80 @@ void ESP8266WebServerTemplate<ServerType>::setContentLength(const size_t content
...
     if (_keepAlive) {
-      sendHeader(String(F("Keep-Alive")), String(F("timeout=")) + HTTP_MAX_CLOSE_WAIT);
+        if (!_streamHeader(F("Keep-Alive"), String(F("timeout=")) + HTTP_MAX_CLOSE_WAIT))
+            return false;
     }
 
+    for (const auto& kv: _userHeaders)
+        if (!_streamHeader(kv.first, kv.second))
+            return false;
+    _userHeaders.clear();
```
TD-er, line 503: When will these be cleared if streaming failed?

N.B. I got the impression the current implementation of the webserver does hold some data way longer than needed as opening a web page may sometimes free up quite a bit of memory. (maybe POST data of a previous request???)

hmueller01:
I think the same happens in the [ESP8266HTTPClient.cpp#addHeader](https://github.com/esp8266/Arduino/blob/7f2deb14a254d0d1b0903e0bb15a87815833f579/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L695) ...
```c++
        if (first) {
            _headers = headerLine + _headers;
        } else {
            _headers += headerLine;
        }
```
No one knows in advance how big `_headers` grows. Resulting in:
```
[String] 'x-ESP8266- ...  1478656
': Reallocating large String(134 -> 165 bytes)
[String] 'x-ESP8266- ... : 548896
': Reallocating large String(165 -> 221 bytes)
[String] 'x-ESP8266- ... afd6bfc5
': Reallocating large String(221 -> 251 bytes)
[String] 'x-ESP8266- ...  2097152
': Reallocating large String(251 -> 290 bytes)
[String] 'x-ESP8266- ... b29dcd3)
': Reallocating large String(290 -> 315 bytes)
[String] 'x-ESP8266- ...  version
': Reallocating large String(315 -> 338 bytes)
[String] 'x-ESP8266- ... sion: 41
': Reallocating large String(338 -> 357 bytes)
```

hmueller01:
@d-a-v Is there any progress on this PR / issue #8872?

PR #8885, opened 2023-03-10
various minor web fixes

d-a-v:
Single PR about various minor fixes and additions for webserver in order to help debugging.

- httpclient: use refs
- httpserver: add chunks in examples
- BearSSL `read(void)`: debug log: read() returning -1 is usual and means "nothing to read"
- emulation on host: SSL server has never been and is now working
- move SSL server cert from multiple examples to a single location and add appropriate warnings
- http-client: add `getString(pre-reservation)`


libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -631,20 +631,20 @@ WiFiClient* HTTPClient::getStreamPtr(void)
...
     _payload.reset(new StreamString());
 
-    if(_size > 0) {
-        // try to reserve needed memory
-        if(!_payload->reserve((_size + 1))) {
-            DEBUG_HTTPCLIENT("[HTTP-Client][getString] not enough memory to reserve a string! need: %d\n", (_size + 1));
-            return *_payload;
-        }
+    if (_size > 0 && _size >= reserve)
+        reserve = _size + 1;
```
mcspr, line 643: Already size+1 when string doing buffer allocation?

libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
```diff
@@ -631,20 +631,20 @@ WiFiClient* HTTPClient::getStreamPtr(void)
...
+
+    if (reserve && !_payload->reserve(reserve)) {
```
mcspr, line 645: This is int, `reserve > 0`
(unless negative is ok)

PR #8888, opened 2023-03-13
Support user_rf_pre_init() for SDK v3.0

mhightower83:
For SDK v3.0+, early system calls that were called from user_rf_pre_init (SDK v2.2) need to be called from user_pre_init.

Moved call to `user_rf_pre_init()`, from `user_init()` to the end of `user_pre_init()`, so we can still perform early calls like system_phy_set_rfoption(rf_mode), system_phy_set_powerup_option(2), etc.

The hint for where to place `user_rf_pre_init()` came from [ESP-FAQ : 2.2.4 How to modify the default method of RF calibration in ESP8266?](https://docs.espressif.com/_/downloads/espressif-esp-faq/en/latest/pdf/#page=14) (Side note, I think the FAQ has a typo. When they refer to byte 115, I think they meant 114.)
The specific sentence of interest: "Call system_phy_set_powerup_option(3) in function __user_pre_init__ or user_rf_pre_init"

Edited for clarity


RomanNechaev:
пр

PR #8902, opened 2023-03-31
Document git version install, upkeep and PR management

mcspr:


d-a-v (approved):
We also need to document on how to use git version with Arduino IDE 2.x

Issue #3494, opened 2017-08-04
No Poison after the block Error in Smartconfig


romuye123:
#### Hardware
Hardware:			ESP-12E
Core Version:      	2.1.0-rc2

### Description
Downloaded the latest Smartconfig SDK on iOS/android. The smartconfig works perfectly fine for the first few times. During the testing i was repeatedly trying the smartconfig even if it is connected to WIFI and the ESP got a crash with poison block error mentioned below after 4th or 5th attempt. It happens randomly. Is it not advisable to repeatedly use smartconfig even if the device is connected to WIFI ? or is there any workaround for this ?

### Settings in IDE

Module:  			Generic ESP8266 Module
Flash Size: 		4MB/1MB
CPU Frequency:		80Mhz
Flash Mode:			qio
Flash Frequency:	40Mhz
Upload Using:	 SERIAL
Reset Method:		ck

### Sketch

```cpp
void setup()
{
Serial.begin(115200);
startSmartConfig();
}
void startSmartConfig(){

    WiFi.mode(WIFI_STA);
    WiFi.beginSmartConfig();
    delay(200);
  
    //Wait for SmartConfig packet from mobile
    Serial.println("Waiting for SmartConfig.");
    while (!WiFi.smartConfigDone()) {
      delay(500);
      Serial.print(".");
    }
  
    Serial.println("");
    Serial.println("SmartConfig received.");
  
    //Wait for WiFi to connect to AP
    int counter = 0;
    Serial.println("Waiting for WiFi");
    while (WiFi.status() != WL_CONNECTED) {
      delay(500);
      Serial.print(".");
      counter++;
      if(counter == 60){
        break;  
      } 
    }
    counter = 0;

    if(WiFi.status() != WL_CONNECTED)
    {
      Serial.println("WiFi not Connected.");      
    }
    else
    {
      Serial.println("WiFi Connected.");
      Serial.print("IP Address: ");
      Serial.println(WiFi.localIP());
    }
}
void loop(){}
```

### Debug Messages

```
Smart config activated...
Waiting for SmartConfig...................there is no poison after the block. Expected poison address: 0x3fff6e91, actual data: 0xa5 0xa5 0xa5 0x0
block start: 3fff6ab4
```

romuye123:
@me-no-dev need your help on this.

tablatronix:
I got this a few times, I also get wdt resets using 2.4.0

```
Waiting for SmartConfig.
scandone
scandone
[...]

PR #8914, opened 2023-04-22
umm_poison false positive from ISR

mhightower83:
The umm_poison logic runs outside the UMM_CRITICAL_* umbrella. When interrupt routines do alloc calls, it is possible to interrupt an in-progress allocation just before the poison is set, with a new alloc request resulting in a false "poison check fail" against the in-progress allocation. The SDK does mallocs from ISRs. SmartConfig can illustrate this issue.

This should resolve https://github.com/esp8266/Arduino/issues/3494#issue-247919259

### DIFF
--- a/cores/esp8266/umm_malloc/umm_local.c
+++ b/cores/esp8266/umm_malloc/umm_local.c
@@ -138,12 +138,16 @@ static void *get_unpoisoned_check_neighbors(void *vptr, const char *file, int li
 void *umm_poison_realloc_fl(void *ptr, size_t size, const char *file, int line) {
     void *ret;
 
+    POISON_CRITICAL_ENTRY();
+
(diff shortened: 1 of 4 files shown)


mcspr:
Any reason not to move currently separate `umm_malloc` and `get_poisoned` into a single operation that happens inside critical enter & exit blocks? 
Adding a different false positive, other way around by ignoring the check with some specific counter value, feels kind of weird.

mhightower83:
The current upstream umm_malloc has umm_poison as an outer wrapper. Without searching through all the old stuff, it used to be more mixed together. I have already started to blur that separation with `POISON_CHECK_NEIGHBORS()` in `umm_malloc()`. I'll look at moving `get_poisoned()` in as well. I will need to think about this a bit. `realloc()` gets tricky.

Issue #8913, opened 2023-04-22
light sleep mode & WiFI & millis() overflow


Bruno-Muller:
Hi,

**Issue:**
WiFi stops working after a few hours when using the light sleep mode.

**Description:**
I am not sure if the Arduino library is doing well with the **time drifting** caused by the  **Light Sleep Mode** because the **clock is idle**. And if the **millis() overflow** can cause any **dead lock** issue in the timeout condition to exit loops.

I'm not an expert, I suspect that the WiFi API does not renew and negotiate some WiFi token in a timely manner because millis() is not a good time indicator when we use the light sleep mode.

**Steps to reproduce:**
Periodically put the ESP8266 in light sleep mode.
Here in my tests, the wake time is around 5s, and sleep time is around 15s.

In setup():
```
WiFi.mode(WIFI_STA);

// I've seen that the library will override the value to MODEM_SLEEP later.
// but I will go around to avoid the situation with some custom sleep functions.
WiFi.setSleepMode(WiFiSleepType::WIFI_LIGHT_SLEEP);

WiFi.begin(WIFI_SSID, WIFI_PASSWORD);
```

Go to light sleep and wake up functions:

```
volatile bool wifi_sleeping = false;

// call this function to go to light sleep mode
void start_light_sleep(uint32_t time_ms) {

  // Not sure if this is needed, but I've read that all timer interrupts should be disabled to not interfere
  // So maybe it makes sense:
  extern os_timer_t *timer_list; 
  timer_list = nullptr; 

  wifi_station_disconnect();
  wifi_set_opmode(NULL_MODE);
  wifi_fpm_set_sleep_type(LIGHT_SLEEP_T);
  wifi_fpm_open(); 

  Serial.println("ENTER LIGHT SLEEP MODE");
  Serial.flush();
  Serial.end();

  wifi_sleeping = true;
  
  wifi_fpm_set_wakeup_cb(light_sleep_wakeup);
  wifi_fpm_do_sleep(time_ms*1000);  // in us, finally we are going to light sleep mode.
  
  // So I've read on the internet that the delay should be sleep_time_ms + 1 ms.
  // But I'm not sure if this is exact arithmancy.
  // I've measured that the ESP8266 needs around 460 ms to go to light sleep.
  // So I assume that the delay has to be at least 460ms.
  esp_delay(time_ms+1, [](){ return wifi_sleeping; }); 
}

// this function is called by the wake up callback interrupt. Do not call it yourself.
[...]

PR #8915, opened 2023-04-23
Bug fix EspSoftwareSerial 8.0.3:: Arduino library repo didn't pick up 8.0.2 proper

dok-net:
@mcspr I'm sorry, my fault.

PS: do you still intend to review and merge #7979  in the near future? There's #8913 now with assumptions about stopping the timers etc, that would make most sense to revisit in the context of my PR.


mcspr:
Just use braces to init?
```cpp
#include <atomic>
struct Foo {
  std::atomic<bool> bar { false };
};
```

gcc8 stdlib should be ok with that
(`~/toolchain-xtensa-esp32@8.4.0+2021r2-patch5/bin/xtensa-esp32-elf-g++ -c atomic.cpp` works with the above)

dok-net:
@mcspr Too short, I dont' have a clue what you are talking about. Anyway, since 8.0.3 has passed the Arduino library release checks and all now, is this a blocking issue you are referring me to? Again, please let me know what you are talking about, I'll gladly put in the sources, ready for the next release :-)

dok-net:
@mcspr This release 8.1.0 is ready for ESP32 IDF 5.1 and fixes a few other compilation problems as well. Please merge at your discretion.

d-a-v:
This one is still tagged as draft.

dok-net:
Some careful consideration by the core team is required here: the new ghostl library apparently does get pulled in just fine by the Arduino library dependency framework, but this is breaking with the traditional model of the ESP8266 core that has all required libs as copies or git submodules.

mcspr:
Arduino library dependency framework would pick things up during interactive session, true. Not the case here, library must be present in search path.

Note that if you'd do that, now you have 2 libs to sync to here :> And Core suddenly has an external dependency for no reason. Bundling is not an option?

dok-net:
"Bundling" like previously, or are suggesting something I haven't thought of?

Anyway:

```
libraries/esp8266/examples/SerialStress/Release/board.buildinfo
libraries/esp8266/examples/SerialStress/SerialStress.ino
libraries/esp8266/examples/SerialStress/SerialStress.vcxproj
libraries/ESP8266WiFi/examples/WiFiTelnetToSerial/WiFiTelnetToSerial.ino
libraries/lwIP_PPP/examples/PPPServer/PPPServer.ino
```

are the only place where the EspSoftwareSerial lib is used directly.
My focus has moved from the ESP8266 to the ESP32 family, like most users seem to have as well. The inclusion as a submodule in ESP8266 core has been a source of pain for everyone using both ESP8266 and ESP32s and me far too long already.
I'm adding the ghostl library as a new submodule in this PR now, but I am not convinced. Perhaps do this and merge once now, then sometime soon remove BOTH submodules and either document or change the few places that use EspSoftwareSerial in the ESP8266 tree, for everyone's benefit.
I don't have much of a preference, except that I will not be able to spend much effort - plus everyone complaining about issues with multiple versions of the libs between ESP8266's included and any Arduino / platformio downloaded ones in ./libraries.

dok-net:
I've checked the aforementioned sketches. The all appear to use EspSoftwareSerial in order to have two UARTs available, one for logging. IIRC the second UART of ESP8266 is TX only, but perfectly suited to that task. Therefore, somebody could rewrite the sketches just fine and free the EspSoftwareSerial hard dependency. Everyone needing are real duplex SW UART can use the Arduino library manager way.

mcspr:
> I've checked the aforementioned sketches. The all appear to use EspSoftwareSerial in order to have two UARTs available, one for logging. IIRC the second UART of ESP8266 is TX only, but perfectly suited to that task. Therefore, somebody could rewrite the sketches just fine and free the EspSoftwareSerial hard dependency. Everyone needing a real duplex SW UART can use the Arduino library manager way.

Agreed, examples should prefer real uart here.

> "Bundling" like previously, or are you suggesting something I haven't thought of?

Just put it inside of the lib folder as previously, but with an added namespace. Meaning
```cpp
#include <circular_queue.h>
```
Becomes this
```cpp
#include <ghostl/circular_queue.h>
```

(but, I think here it must also drop any mention of ghostl from library .json and .properties)

> I don't have much of a preference, except that I will not be able to spend much effort - plus everyone complaining about issues with multiple versions of the libs between ESP8266's included and any Arduino / platformio downloaded ones in ./libraries.

True, but so is the case for WiFi libs or anything coming from the overly-simplified library manager. IDE / CLI that causes a lot of headache when using multiple development platforms. Specifically the fact that $SKETCHLIB/libraries is overriding Core lib(s) for some reason. PIO actually has it solved with project-private libs folders, but it is more of an Arduino-adjacent solution and an altogether different tool.

I'm still on the 'keep it here' side, so repo has a version of swserial that definitely works. Whether it is up-to-date is whole other question.
(I can always just cherry-pick the version to sync here if you don't want to :)

dok-net:
I don't understand, going backward to

`#include <ghostl/circular_queue.h>`

and including ghostl as a copy in EspSoftwareSerial is not an option now - just to state it categorically once more.
Especially that now somebody was kind enough to submit ghostl to the platformio library repository.

Is there anything I am missing with this PR, putting ghostl side-by-side in the submodules list? This works and finally solves one on the update headaches. If you're puzzled about the project ownership - I am the original author of ghostl, AND I am the acting maintainer of EspSoftwareSerial, too - so there is conflict of interest or potential hazard that one or the other would be handled differently with regards to ESP8266 Arduino core's needs.

Are considering merging this PR at all?

mcspr:
Fixing CI might be a good start? Can't merge if 'Squash and merge' button does not work.
https://github.com/esp8266/Arduino/blob/47327e8449a061dcce540ba60e4341cefbabac50/tests/common.sh#L203
`skip_sketch`, followed by `skip_ino` must skip impossible-to-build sketches, which includes ghostl coroutine examples. Either by explicitly mentioning them, or creating `SKETCH_DIR/.test.skip`

To reiterate - the idea is to have '#include' path contain all of the required files in one step, **just** for the Core distribution. And all while inside of the single lib directory, possibly using a separate branch that have these already merged. If you don't think that is viable, ok, we'll ship both.
(btw aren't the even more problems b/c of something else depending on ghostl causing out-of-sync versions yet again?)

Namespacing of library includes would be helpful regardless? Seems like a commonplace practice for more-or-less complex libs. `#include <ghostl.hpp>` for everything, `#include <ghostl/foo.h>` for specific stuff. Otherwise this is yet another problem where header names can clash with each other.

dok-net:
Ah, CI, right.
Would you consider adding this to the default?

I'm currently using a platform.local.txt:
```
build.stdcpp_level=-std=gnu++20
compiler.cpp.extra_flags=-fcoroutines -std=gnu++20
```

mcspr:
See #8916 and #8990. Main concerns are with new -std=... build defaults applied to every lib and sketch besides our Core files, (sometimes useless) new warnings, and also incomplete support and its unknown quality in gcc10 branch compared to gcc11 and gcc12. Maybe next version, not really feeling adventurous just yet

dok-net:
About translating c++ namespace into directory stucture ... good point... I'm working that out. Bear with me.

mcspr:
OK to merge this?

PR #8918, opened 2023-04-30
Soft WDT: detect deliberate infinite loop at Postmortem

mhightower83:
A popular method of handling an unrecoverable state is to reboot. The SDK does this in many places by printing a cryptic debug message followed by something equivalent to `while(true){}`, which compiles down to `loop: j loop`, creating a Soft WDT reset. 

This change calls attention to these deliberate WDT resets at Postmortem. In the one case that I examined closer, the SDK responded this way after an OOM event.

For SDK 3.0.5, I found this many "deliberate infinite loops":
| Libary | count |
| ---- | ----:|
| libpp.a |81
| libmain.a |8
| libespnow.a | 6

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -181,7 +181,13 @@ static void postmortem_report(uint32_t sp_dump) {
             exccause, epc1, rst_info.epc2, rst_info.epc3, rst_info.excvaddr, rst_info.depc);
     }
     else if (rst_info.reason == REASON_SOFT_WDT_RST) {
-        ets_printf_P(PSTR("\nSoft WDT reset\n"));
+        ets_printf_P(PSTR("\nSoft WDT reset"));
+        const char infinite_loop[] = { 0x06, 0xff, 0xff };  // loop: j loop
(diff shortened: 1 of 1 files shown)


cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -181,7 +181,13 @@ static void postmortem_report(uint32_t sp_dump) {
             exccause, epc1, rst_info.epc2, rst_info.epc3, rst_info.excvaddr, rst_info.depc);
     }
     else if (rst_info.reason == REASON_SOFT_WDT_RST) {
-        ets_printf_P(PSTR("\nSoft WDT reset\n"));
+        ets_printf_P(PSTR("\nSoft WDT reset"));
+        const char infinite_loop[] = { 0x06, 0xff, 0xff };  // loop: j loop
+        if (0 == memcmp_P((void*)rst_info.epc1, infinite_loop, 3u)) {
```
mcspr, line 186: args other way around? 1st should allow byte access, 2nd is read via pgm macro
https://github.com/earlephilhower/newlib-xtensa/blob/ebc967552ce827f21fc579fd8c437037c1b472ab/newlib/libc/sys/xtensa/string_pgmspace.c#L144

mhightower83: Thanks!

PR #8919, opened 2023-05-03
Update cert.py

hafidhh:
fix TypeError: hex() takes no arguments (1 given)


mcspr:
https://docs.python.org/3/library/stdtypes.html#bytes.hex
update python to at least 3.8? we need *at least* version 3.7 for most scripts, not exactly 3.7

PR #8931, opened 2023-05-19
Change host from tls.mbed.org to api.my-ip.io

Lan-Hekary:
https://tls.mbed.org/ host does not support MFLN and Also Redirects the Client.
https://api.my-ip.io/ip is a better alternative, supporting MFLN and allowing the user to get a simple text with a useful information.


d-a-v (approved):
Thanks !

Issue #8939, opened 2023-06-08
Feature Request: TCP Keepalive in WiFiClientSecure


mobizt:
### Basic Infos

- [ ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ ] I have tested that the issue is present in current master branch (aka latest git).
- [ ] I have searched the issue tracker for a similar issue.
- [ ] If there is a stack dump, I have decoded it.
- [ ] I have filled out all fields below.

#### Platform

- Core Version:			latest (3.1.2)

### Problem Description

The TCP Keepalive was already implemented in WiFiClient class but it was missing from WiFiClientSecure class. 

Can you add support TCP Keepalive in WiFiClientSecure class?

https://github.com/esp8266/Arduino/blob/57fa6cdc9295c30cbde7aa2e4205f8d52dbaba33/libraries/ESP8266WiFi/src/WiFiClient.cpp#L417-L440

earlephilhower:
WiFiClientSecure already implements this through the magic of C++ inheritance, no?  The following snippet builds fine:
````
#include <ESP8266WiFi.h>

void setup() {
  WiFiClientSecure s;
  s.keepAlive(1,2,3);
}

void loop() { }
````

**Comment by mobizt** [comment]
I test with your simple code and device crashed with this error in Serial monitor.

```
--------------- CUT HERE FOR EXCEPTION DECODER ---------------

Exception (28):
epc1=0x40207056 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffe30 end: 3fffffd0 offset: 0150
3fffff80:  40207558 00000000 00001388 feefeffe  
3fffff90:  00000000 00000000 00000000 3ffef504  
3fffffa0:  3ffef504 3fff0e1c feefeffe feefeffe  
3fffffb0:  feefeffe feefeffe 3ffee6f8 40204510  
3fffffc0:  feefeffe feefeffe 3fffdab0 40100c39  
<<<stack<<<

--------------- CUT HERE FOR EXCEPTION DECODER ---------------

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 3424, room 16 
tail 0
chksum 0x2e
[...]

PR #8940, opened 2023-06-09
Add support WiFiClientSecure TCP KeepAlive

mobizt:
This PR will fix this issue https://github.com/esp8266/Arduino/issues/8939

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
+++ b/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
@@ -358,20 +358,20 @@ class WiFiClientSecure : public WiFiClient {
     // consume bytes after use (see peekBuffer)
     virtual void peekConsume (size_t consume) override { return _ctx->peekConsume(consume); }
   
-    void keepAlive(uint16_t idle_sec = TCP_DEFAULT_KEEPALIVE_IDLE_SEC, uint16_t intv_sec = TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC, uint8_t count = TCP_DEFAULT_KEEPALIVE_COUNT)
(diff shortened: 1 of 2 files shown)


earlephilhower (approved):
Thanks!

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
```diff
@@ -357,6 +357,21 @@ class WiFiClientSecure : public WiFiClient {
 
     // consume bytes after use (see peekBuffer)
     virtual void peekConsume (size_t consume) override { return _ctx->peekConsume(consume); }
+  
+    void keepAlive(uint16_t idle_sec = TCP_DEFAULT_KEEPALIVE_IDLE_SEC, uint16_t intv_sec = TCP_DEFAULT_KEEPALIVE_INTERVAL_SEC, uint8_t count = TCP_DEFAULT_KEEPALIVE_COUNT)
```
earlephilhower, line 361: Thinking over about this, I would request one change.  Can we make the `WiFiClass` implementation of these as `virtual` and then make these methods `override`?

That way we can guarantee the issue you hit in https://github.com/esp8266/Arduino/issues/8939#issuecomment-1584737239 will be avoided if someone passes a WiFiClientSecure around by reference and it gets consumed as a WiFiClient.

earlephilhower (changes requested):

earlephilhower (approved):

libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.h
earlephilhower: Thanks!

mobizt: Yes, I done it.

Issue #8947, opened 2023-06-26
delay(0) behavior changed


mhightower83:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					ESP-12
- Core Version:			521ae60a89e64bb0d1eb7a0b7addf620ced5cad3
- Development Env:		Arduino IDE 1.8.19
- Operating System:		Ubuntu

### Settings in IDE

- Module:  			Adafruit Feather HUZZAH ESP8266
- Flash Mode:			qio
- Flash Size: 		4MB
- lwip Variant:   v2 Higher Bandwidth
- Reset Method:		nodemcu
- Flash Frequency: 40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   460800

### Problem Description

A blocking library function that uses `delay(0)` will fail with _Software WDT Reset_.

This last worked in Core version 3.1.0

Changes made in commit 39080e317e0dc5a182feb0f593b084647be53a2a altered this behavior

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

```cpp

#include <Arduino.h>

// Last worked with Core v3.1.0

void setup() {
  Serial.begin(115200);
  delay(200);
}

void loop() {
  Serial.printf("\r\n\r\ndelay(0) test - should never WDT reset\r\n");

  while(true) {
    delay(0);
  }
}

```

### Debug Messages

```

--------------- CUT HERE FOR EXCEPTION DECODER ---------------

Soft WDT reset

Exception (4):
epc1=0x40100249 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffe20 end: 3fffffd0 offset: 0160
3fffff80:  00000000 3fffff60 3fffff90 3ffee5a8 <
3fffff90:  3fffdad0 00000000 00000000 40201f3c  
3fffffa0:  3fffdad0 00000cc1 3ffee57c 40201069  
[...]

PR #8949, opened 2023-07-01
Fix 8947: delay(0) must yield(); recurrent scheduled functions max delay fixed

dok-net:


cores/esp8266/core_esp8266_main.cpp
```diff
@@ -166,18 +163,20 @@ extern "C" void __esp_delay(unsigned long ms) {
 extern "C" void esp_delay(unsigned long ms) __attribute__((weak, alias("__esp_delay")));
 
 bool esp_try_delay(const uint32_t start_ms, const uint32_t timeout_ms, const uint32_t intvl_ms) {
+    if (!timeout_ms) {
+        esp_yield();
+        return true;
+    }
```
d-a-v, line 169: This alone allows to fix #8947.

cores/esp8266/Schedule.cpp
```diff
@@ -49,6 +47,7 @@ struct recurrent_fn_t
 
 static recurrent_fn_t* rFirst = nullptr;
 static recurrent_fn_t* rLast = nullptr;
+static uint32_t rTarget;
```
d-a-v, line 50: maybe a comment stating that `rTarget` will be initialized before the first call to `schedule_function()`

cores/esp8266/Schedule.cpp
```diff
@@ -122,6 +121,13 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
 
     esp8266::InterruptLock lockAllInterruptsInThisScope;
 
+    // prevent new item overwriting an already expired rTarget.
+    const int32_t remaining = rTarget - micros();
+    if (!rFirst || (remaining > 0 && static_cast<uint32_t>(remaining) > item->callNow.remaining()))
+    {
+        rTarget = micros() + item->callNow.remaining();
```
d-a-v, line 128: `item->callNow.remaining()` is called twice.
Should it be cached ?

cores/esp8266/Schedule.cpp
```diff
@@ -132,37 +138,20 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
...
-#ifdef DEBUG_ESP_CORE
-        static uint32_t last_grain = 0;
-        if (recurrent_max_grain_mS != last_grain)
-        {
-            ::printf(":rsf %u->%u\n", last_grain, recurrent_max_grain_mS);
-            last_grain = recurrent_max_grain_mS;
-        }
-#endif
-    }
+    if (!rFirst) return ~static_cast<decltype(micros())>(0) >> 1;
```
d-a-v, line 146: `~static_cast<decltype(micros())>(0) >> 1` should be usefully replaced by `std::numeric_limits<decltype(micros())>::max()`

Why not using `using micros_t = decltype(micros())`
and `using micros_signed_t = std::make_signed_t<micros_t>` ?

cores/esp8266/Schedule.cpp
```diff
@@ -132,37 +138,20 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
...
+    // handle already expired rTarget.
+    const int32_t remaining = rTarget - micros();
+    return (remaining > 0) ? static_cast<uint32_t>(remaining) : 0;
+}
 
-    return recurrent_max_grain_mS;
+uint32_t get_scheduled_delay_us()
+{
+    return sFirst ? 0 : ~static_cast<decltype(micros())>(0) >> 1;
```
d-a-v, line 154: `std::numeric_limits<decltype(micros())>::max()`

cores/esp8266/Schedule.cpp
```diff
@@ -225,11 +214,18 @@ void run_scheduled_recurrent_functions()
...
 
+    decltype(rLast) stop;
     recurrent_fn_t* prev = nullptr;
-    // prevent scheduling of new functions during this run
-    auto stop = rLast;
-
     bool done;
+
+    {
+        esp8266::InterruptLock lockAllInterruptsInThisScope;
```
d-a-v, line 222: Why is this needed here ?

d-a-v (changes requested):
While I find it funny that in *your humble opinion* you think the logic is incorrect, I took time to review your proposal.
The idea to recalculate next call date at every call is interesting and could be indeed more efficient than relying on the durations gcd to check on them all (while I'm not certain yet, because gcd is calculated once).

Also, this PR's title is outdated.

cores/esp8266/Schedule.cpp
dok-net: I have added a comment about the purpose of this object, C++ rules are clear on that it is zero initialized by default.

cores/esp8266/Schedule.cpp
dok-net: Done.

cores/esp8266/Schedule.cpp
dok-net: > `~static_cast<decltype(micros())>(0) >> 1` should be usefully replaced by `std::numeric_limits<decltype(micros())>::max()`

The results are dramatically different, unless also `>> 1`. I feel that the purpose becomes less "obvious" when working with "max", which had given thought and discarded due to that reason. The wrap-over logic is a very technical detail.

> Why not using `using micros_t = decltype(micros())` and `using micros_signed_t = std::make_signed_t<micros_t>` ?

I am really not a fan of aliasing. If your change is a prerequiste, re-request, please.

cores/esp8266/Schedule.cpp
dok-net: see comment above, please.

cores/esp8266/Schedule.cpp
dok-net: I don't see any reason either. Changed.

cores/esp8266/Schedule.cpp
d-a-v: > The wrap-over logic is a very technical detail.

Really?

The fact is that this code does not compile without warnings in 64 bit host mode.

How would your logic be different without `>>1` ?
Or from `rTarget = micros() + (~static_cast<decltype(micros())>(0) >> 1)` to `rTarget = micros() - 1`

I tried this and it seems to works with a trivial example.
Demixing `decltype(micros())` and `uint32` appears to be necessary.
```diff

--- a/cores/esp8266/Schedule.cpp
+++ b/cores/esp8266/Schedule.cpp
@@ -146,7 +146,7 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
 
 uint32_t get_scheduled_recurrent_delay_us()
 {
-    if (!rFirst) return ~static_cast<decltype(micros())>(0) >> 1;
+    if (!rFirst) return std::numeric_limits<uint32_t>::max();
     // handle already expired rTarget.
     const int32_t remaining = rTarget - micros();
     return (remaining > 0) ? static_cast<uint32_t>(remaining) : 0;
@@ -154,7 +154,7 @@ uint32_t get_scheduled_recurrent_delay_us()
 
 uint32_t get_scheduled_delay_us()
 {
-    return sFirst ? 0 : ~static_cast<decltype(micros())>(0) >> 1;
+    return sFirst ? 0 : std::numeric_limits<uint32_t>::max();
 }
 
 void run_scheduled_functions()
@@ -223,7 +223,7 @@ void run_scheduled_recurrent_functions()
 
     // prevent scheduling of new functions during this run
     stop = rLast;
-    rTarget = micros() + (~static_cast<decltype(micros())>(0) >> 1);
+    rTarget = micros() - 1;
```

cores/esp8266/Schedule.cpp
```diff
@@ -122,6 +123,15 @@ bool schedule_recurrent_function_us(const std::function<bool(void)>& fn,
...
 
+    // prevent new item overwriting an already expired rTarget.
+    const auto now = micros();
+    const auto itemRemaining = item->callNow.remaining();
+    const int32_t remaining = rTarget - now;
+    if (!rFirst || (remaining > 0 && static_cast<decltype(micros())>(remaining) > itemRemaining))
+    {
+        rTarget = now + itemRemaining;
+    }
+
```
d-a-v, line 134: Without needing `remaining` and `static_cast`:
```
if (!rFirst || (rTarget > now && rTarget > now + itemRemaining))
```
?

dok-net:
Set to draft: I'm working out the int32_t uses. Will be back shortly.

cores/esp8266/Schedule.cpp
dok-net: I would like to think that the new comment at HALF_MAX_MICROS explains everything. If it doesn't, let's improve the wording together.

cores/esp8266/Schedule.cpp
dok-net: I have changed the remaining int32_t uses and the respective casts.

d-a-v (commented):
I run this test:
```c
#include <Schedule.h>

void setup ()
{
    Serial.begin(74880);
    Serial.println();
    Serial.println("result in usec:");
    
    for (int i = 0; i < 4; i++)
        schedule_recurrent_function_us([]()
        {
            return true;
        }, 1000 + (300 * i) /*us*/);
}

uint64_t between;
uint32_t count;
decltype(esp_get_cycle_count()) last, now;

void loop ()
{
    now = esp_get_cycle_count();
    
    if (!(count & 0xfffff))
        Serial.printf("%f\n", between / (1.0 * F_CPU * count / 1000000));

    between += now - last;
    count++;

    last = esp_get_cycle_count();
}
```

It is designed to count time passed by core between loops.
I get 12us with master and 23us with this PR.

dok-net:
@d-a-v I'm beginning to wonder if the test methodology is off target: we are going the extra mile to extract a good delay duration - that is what this PR is all about. But the test makes no use of that at all. I know it becomes more complicated to extract a metric that matters, but without delaying, and perhaps measuring the effect on power consumption, it seems without much value to the overall picture. What do you think?

d-a-v:
What I understand from your message

> the test methodology is off target
> the test doesn't focus on the "quality" at all
> measure the effect on power consumption (<= no difference after measurement)

Can you explain what is "quality" if it is not speed of execution, which is directly related to the number of instructions executed between loops. As you found no difference on power consumption we can deduce that internal work between loops consume twice more energy for the same purpose. I am aware that your changes deal with usecs where current code is only msecs-aware. Current code uses a cache, yours recalculate bounds at every loop. Can't you change this ? Or should we update the current code to use usecs instead ?

> we should explore the effects of computing a maximum delay vs. not computing one at all and using a fixed value in the first place

This looks like missing the target, or I don't understand this point.

dok-net:
@d-a-v unfortunately I have to focus my attention on other projects. While I think there's still something strange about using gcd, my current proposal has a significant performance impact and would need fixing.
Should we leave this MR standing (as draft?) or shall I withdraw it completely?

Issue #8952, opened 2023-07-06
No poison after block with current git version


hmueller01:
### Basic Infos

- [x] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [x] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [x] I have tested that the issue is present in current master branch (aka latest git).
- [x] I have searched the issue tracker for a similar issue.
- [x] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:				[ESP8266 device]
- Core Version:			[framework-arduinoespressif8266 @ 3.2.0-dev+sha.521ae60]
- Development Env:		[Platformio]
- Operating System:		[MacOS]

### Settings in IDE

- Module:  			[Shelly 1]
- Flash Mode:			[qio]
- Flash Size: 			[2MB]
- lwip Variant:			[default, not set]
- Reset Method:		[nodemcu]
- Flash Frequency:		[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:		[115200] (serial upload only)

### Problem Description

After connecting to the AP the module panics
```
add 1
aid 1
station: a4:83:e7:d2:34:b7 join, AID = 1
No poison after block at: 0xe5a4e455, actual data: 0x80 0x0 0x0 0x0

User exception (panic/abort/assert)
--------------- CUT HERE FOR EXCEPTION DECODER ---------------

Panic umm_malloc.cpp:751 umm_malloc_core

>>>stack>>>
```
Decoded stack dump see below.

**The same code works well with Arduino Release 3.1.2.**

### [MCVE](https://stackoverflow.com/help/mcve) Sketch

There is a lot of code I wrote, so it is not really possible to show it here. But showing the lines noted in the stack dump might help.

/Users/hmueller/src/PlatformIO/Projects/rhts-platformio/src/rhts-arduino.cpp:142
is a global definition of https://github.com/hmueller01/ESP_WiFiManager_Lite
```cpp
ESP_WiFiManager_Lite m_wifi_manager;
```

0x4020dc73: ESP_WiFiManager_Lite::run() at /Users/hmueller/src/PlatformIO/Projects/rhts-platformio/lib/ESP_WiFiManager_Lite.hmueller01/src/ESP_WiFiManager_Lite.h:921
```cpp
[...]

PR #8953, opened 2023-07-08
Fixes occasional UMM_POISON failure

mhightower83:
Bug introduced with PR fix #8914.
When a reallocated pointer could not grow in place, a replacement allocation was created. Then UMM_POISON was written to the wrong block.

This should resolve:
 - https://github.com/esp8266/Arduino/issues/8952#issue-1792056809


d-a-v (approved):
I encountered this issue too, thanks for the fix !

hmueller01:
As written in #8952, this fixes my problems. Pls merge soon. Tx.

Issue #8938, opened 2023-06-06
Map function - IntegerDivideByZero


FernandoGarcia:
Hi!

Someone did a regression in map function and I'm getting this error.

![Captura de tela de 2023-06-03 19 03 44](https://github.com/esp8266/Arduino/assets/2778501/1da1a635-a174-473b-a560-d2dc1795377e)

This issue has been fixed [here](https://github.com/esp8266/Arduino/pull/2397) before.

Best regards.

mcspr:
So https://github.com/esp8266/Arduino/pull/7027 is the culprit. But code above does not check the value regardless, so our exception is replaced with a run-time error.

Looking at esp32 where PR was also merged, [it was eventually reverted back to original implementation](https://github.com/espressif/arduino-esp32/pull/6218) because of errors with larger values. With the provided test (0...180 to 1000...2000 and back), error rate is 0. With 0 and 10000 as min and max - error rate becomes 99%, mapping value and mapping value back are off-by-one :/

Suppose in servo case it makes sense, but not in general func context. Or, calc should take a different value (and btw, divide-by-2 effect is not explained even in servo case besides that it is 'better' :/)
https://github.com/esp8266/Arduino/blob/521ae60a89e64bb0d1eb7a0b7addf620ced5cad3/libraries/Servo/src/Servo.cpp#L36-L38

func|result|expected|
-|-|-
map(7500, 0, 10000, 10000, 0)|2501|2500
map(2501, 10000, 0, 0, 10000)|7499|7499
map(7499, 0, 10000, 10000, 0)|2502|2501
map(2502, 10000, 0, 0, 10000)|7498|7498
...and so on...|...|...

TD-er:
Shouldn't a simple fix like this be enough to prevent these crashes?
```c++
if (divisor == 0) return in_min;
```

Increasing resolution and int overflows with large values are much harder to fix.
But my gut feeling tells me there should be 3 separate cases:
- dividend > divisor
- dividend < divisor
- dividend == divisor

TD-er:
OK, just thought of something like this (still untested, just some gut feeling programming)

```c++
long map(long x, long in_min, long in_max, long out_min, long out_max) {
  const long divident = out_max - out_min;
  const long divisor = in_max - in_min;
  if (divisor == 0) return in_min;

  const long delta = x - in_min;

  if (divident == divisor) {
    return out_min + delta;
  }
  if (divident == (-1 * divisor)) {
    return out_max + delta;
[...]

PR #8957, opened 2023-07-17
Fix Map function - IntegerDivideByZero #8938

TD-er:
Fixes: #8938

Also increase accuracy of mapping and fixing map issues with large values. Still only using integer operations.

Test function to see absolute error compared to floating point map function using `double` precision floating point values.

```c++
# include <iostream>
# include <string>

long test_map(long x, long in_min, long in_max, long out_min, long out_max) {
    const double out_length = out_max - out_min;
    const double in_length  = in_max - in_min;

    if (in_length == 0) { return in_min; }

    if (out_length == 0) { return out_min; }

    const double delta = x - in_min;
    const double value = (delta * out_length + (in_length / 2)) / in_length + out_min;
//    return std::round(value);
    return value;
}

long check(long x, long in_min, long in_max, long out_min, long out_max) {
    const long floatvalue = test_map(x, in_min, in_max, out_min, out_max);
    const long map_value = map(x, in_min, in_max, out_min, out_max);
    return floatvalue - map_value;
}

int main()
{
    long input = 0;

    for (size_t i = 0; i < 15; ++i) {
        input = (i / 3) * 2500 - 1 + i % 3;
        const long values[] = {
            check(input, 0,        1000,   255,      0),
            check(input, 0,        1000,   0,        255),
            check(input, 10000,    0,      100,      10100),
            check(input, 10000,    0,      10100,    100),
            check(input, 10000,    0,      0,        1024),
            check(input, 10000,    0,      1024,     0),
            check(input, 1000,     0,      0,        10240),
            check(input, 1000,     0,      10240,    0),
            check(input, 0,        1000,   0,        10240),
            check(input, 0,        1000,   10240,    0),
            check(input, 0,        10000,  10240,    0),
            check(input, 10234567, -12345, 10234567, 100),
            check(input, 10000,    0,      10234567, 0),
            check(input, 50,       1,      10234567, 0)
        };
        std::cout << input << ":";
        constexpr size_t nrvalues = sizeof(values) / sizeof(values[0]);

        for (size_t i = 0; i < nrvalues; ++i) {
            std::cout << "\t" << values[i];
        }
        std::cout << "\n";
    }
}
```

Output:
```
-1:	0	-1	0	0	0	-1	0	2	2	0	0	-1	2	2
0:	0	0	0	0	0	0	0	0	0	0	0	-1	0	1
1:	-1	0	0	0	-1	0	1	0	0	1	1	-1	0	0
2499:	1	-1	0	0	0	0	1	1	1	1	0	0	0	0
2500:	1	1	0	0	0	0	1	0	0	1	0	0	1	0
2501:	0	0	0	0	0	0	2	0	0	2	1	0	0	1
4999:	1	-1	0	0	0	0	1	1	1	1	0	0	0	0
5000:	1	0	0	0	0	0	1	0	0	1	0	0	1	1
5001:	0	0	0	0	0	0	2	0	0	2	1	0	1	0
7499:	1	-1	0	0	0	0	1	1	1	1	0	0	1	1
7500:	1	1	0	0	0	0	1	0	0	1	0	0	0	0
7501:	0	0	0	0	0	0	2	0	0	2	1	0	1	0
9999:	1	-1	0	0	0	-1	1	1	1	1	0	0	1	0
10000:	1	0	0	0	0	0	1	0	0	1	0	0	0	0
10001:	0	0	0	0	-1	0	2	0	0	2	2	0	0	1
```

### DIFF
--- a/cores/esp8266/WMath.cpp
+++ b/cores/esp8266/WMath.cpp
@@ -72,11 +72,67 @@ long secureRandom(long howsmall, long howbig) {
 }
 
 long map(long x, long in_min, long in_max, long out_min, long out_max) {
-    const long dividend = out_max - out_min;
-    const long divisor = in_max - in_min;
-    const long delta = x - in_min;
+    const long out_length = out_max - out_min;
+    const long in_length  = in_max - in_min;
 
-    return (delta * dividend + (divisor / 2)) / divisor + out_min;
+    if (in_length == 0) { return in_min; }
+
+    if (out_length == 0) { return out_min; }
+
+    long delta = x - in_min;
+
+    if (out_length == in_length) {
+        return out_min + delta;
+    }
+
+    if ((out_length < 0) && (in_length < 0)) {
+        return map(x, in_max, in_min, out_max, out_min);
+    } else if (out_length < 0) {
+        return map(in_max - delta, in_min, in_max, out_max, out_min);
+    } else if (in_length < 0) {
+        return map(in_max - delta, in_max, in_min, out_min, out_max);
+    }
+
+    // We now know in_min < in_max and out_min < out_max
+    // Make sure x is within range of in_min ... in_max
+    if ((x < in_min) || (x > in_max)) {
+        long shift_factor = 0;
+
+        if (x < in_min) {
+            const long before_min = in_min - x;
+            shift_factor = -1 - (before_min / in_length);
+        } else if (x > in_max) {
+            const long passed_max = x - in_max;
+            shift_factor = 1 + (passed_max / in_length);
+        }
+
+        if (shift_factor != 0) {
+            const long in_shift  = shift_factor * in_length;
+            const long out_shift = shift_factor * out_length;
+            in_min += in_shift;
+            in_max += in_shift;
+            out_min += out_shift;
+            out_max += out_shift;
+            delta = x - in_min;
+        }
+    }
+
+    if (out_length > in_length) {
+        // Map to larger range
+        const long factor = out_length / in_length;
+        const long error_mod =    out_length % in_length;
+        const long error = (delta * error_mod) / in_length;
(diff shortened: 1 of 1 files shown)


d-a-v:
Why is `in_min` returned when input range is 0? Why not `out_min` or `(out_min + out_max)/2` ?
User always expects to get a value of the same order as `out_min..out_max`.

The [arduino reference](https://www.arduino.cc/reference/en/language/functions/math/map/) is 
```c
long map(long x, long in_min, long in_max, long out_min, long out_max) {
  return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
}
```

The issue is the division by 0, why not [just handling this case](https://github.com/dok-net/arduino-esp8266/blob/cd5cbae1f6881961c1f7b7bb11641f00008df01b/cores/esp8266/WMath.cpp#L77) ?
How would this compare with your `test_map()` (with fixed return when input range is 0)

TD-er:
In the latest commit this was alread addressed, as I also noticed this odd choice of mine :)

```c++
    if (in_length == 0 || out_length == 0) { 
        return out_min; 
    }
```

> The issue is the division by 0, why not [just handling this case](https://github.com/dok-net/arduino-esp8266/blob/cd5cbae1f6881961c1f7b7bb11641f00008df01b/cores/esp8266/WMath.cpp#L77) ?

Because there were more issues with map, as it was experiencing integer overflow issues quite easily.
You don't need to get really high values to experience those overflow issues.
Only with `delta` and `dividend` (renamed to `out_range` in my code as I constantly got confused) approaching 31 bits in size would already result in overflow issues.

> How would this compare with your test_map() (with fixed return when input range is 0)

Just tested it and it will just return what has been given as out_min value.
Thus before the 'swap' of values.
`map(X, N, N, out_min, out_max)` will return `out_min`.

Issue #8581, opened 2022-05-24
core 2.7.4 to 3.0.2 problem in interrupt and timer1


GitHubLionel:
Hi

So my project : ESP12E Wemos D1 mini, 4 Mo, core 2.7.4, Sloeber env. on Windows7

Problem : My code work perfectly with core 2.7.4 and crash with 3.0.2. The crash is always around the same place : in the interrupt of the timer1. In short, I made a dimer : when I have the zero cross interrupt, I start the timer1 to activate the SSR relay for a delay.

I really made hundred of try but always the same problem.

Any ideas are welcome :)

Regards,
Lionel

The relevant code is :
```cpp
#define SSR_TURN_ON_us    50
uint8_t SSR_PIN;
int8_t LED_PIN = -1;
static bool Top_CS_ZC = false;
static volatile bool Tim_Interrupt_Enabled = false;
static volatile bool Is_SSR_enabled = false;
static volatile float SSR_COUNT = 0.0;

// Send SSR pulse after delay
void IRAM_ATTR onTimerSSR(void)
{
	if (Is_SSR_enabled)
	{
		if (Top_CS_ZC)
		{
			digitalWrite(SSR_PIN, HIGH);
			Top_CS_ZC = false;
			Timer_SSR.setInterval(SSR_TURN_ON_us);
			Timer_SSR.startTimer();
		}
		else
		{
			digitalWrite(SSR_PIN, LOW);
			if (LED_PIN != -1)
				digitalWrite(LED_PIN, HIGH);
		}
	}
}

// Zero cross was fired
void IRAM_ATTR onZC(void)
{
	Count_CS_ZC++;

	if (Is_SSR_enabled && Tim_Interrupt_Enabled)
	{
		if (LED_PIN != -1)
			digitalWrite(LED_PIN, LOW);

		Top_CS_ZC = true;
		 Timer_SSR.setInterval(SSR_COUNT);
		Timer_SSR.startTimer();     <------------------- here is the line 120
	}
}

void IRAM_ATTR startTimer(void)
{
	timer1_write(_timerCount);
	timer1_enable(TIM_DIV, TIM_EDGE, _loop);
}

```

The dump crash stack is :
```
0x40101654: interrupt_handler(void*, void*) at ...\esp8266\3.0.2\cores\esp8266\core_esp8266_wiring_digital.cpp line 175
[...]

PR #8958, opened 2023-07-17
optionally move float emulation code into iram

d-a-v:
following #7180 and #7188

Floating point operations are currently in IROM, they cannot be used from inside ISRs.

This PRs allows arduino builder & IDE to select where to link soft floating point code.
This was already possible for platformIO, nothing changes on this side.

From an undetermined date and until now, they have been in IROM(?) and are still there.

This is suitable for external libraries that are doing floats in ISRs.
(AccelStepper is an example)


d-a-v:
fpiram

tools/boards.txt.py
```diff
@@ -1682,6 +1682,17 @@ def sdk ():
 
 ################################################################
 
+def float_in_iram ():
+    return { 'iramfloat': collections.OrderedDict([
+                        ('.menu.iramfloat.no', 'Not allowed in ISR (more IRAM)'),
+                        ('.menu.iramfloat.no.build.iramfloat', '-DFP_IN_IROM'),
+                        ('.menu.iramfloat.yes', 'Allowed in ISR (less IRAM)'),
+                        ('.menu.iramfloat.yes.build.iramfloat', '-DFP_IN_IRAM'),
```
mcspr, line 1690: May lose IRAM too in the 1st? Just mention in the 2nd option

: Not allowed in ISR
: Allowed in ISR (+1KiB IRAM) *(or however much, we definitely know here)*

mcspr: Or, as VTABLES one. Just to signify they don't go away, just change mapping

: Not allowed in ISR (Flash)
: Allowed in ISR (IRAM)

mcspr:
Also linking to https://github.com/esp8266/Arduino/issues/8581, but I am still as confused as when closing that. So, if FP is in IROM, what was the point of 7180 then? If is not the default, just an optional flag, 

There is a weird mix of FP funcs coming from both ROM libgcc.a and from our own libgcc.a, the full list for reference https://gcc.gnu.org/onlinedocs/gccint/Soft-float-library-routines.html
Note that PIO has a special filter for .ld generation, and may skip define flags. Will fix that if we figure out the logic of fp funcs >_<

mcspr:
For a quick test, trying to build with latest 3.1.2 and 2.7.4. Note that this comes from _cmpdf2.o, where #6294 update added some .o files to the .irom filter
```diff
+    *libgcc.a:_divsf3.o(.literal .text)
+    *libgcc.a:_fixsfsi.o(.literal .text)
+    *libgcc.a:_cmpdf2.o(.literal .text)
+    *libgcc.a:_cmpsf2.o(.literal .text)
```

Works as expected, so filters weren't somehow broken
```
~/d/arduino8958> nm -C .pio/build/3_1_2/firmware.elf | grep ledf2
4020a7f4 T __ledf2
```
```
~/d/arduino8958> nm -C .pio/build/2_7_4/firmware.elf | grep ledf2
40102334 T __ledf2
```

Otherwise unmatched stuff still goes to IRAM, where PR would also force it into IROM (!)
```
> nm -C .pio/build/3_1_2/firmware.elf | grep -f funcs.txt | grep 401
4010658c T __fixdfdi
401068f0 T __floatdidf
401068e4 T __floatundidf
```

Remaining parts of the lib are PROVIDE'd from the chip itself
(__adddf3 and below in the eagle.rom.addr.v6.ld)
```
4000c538 A __adddf3
4000c180 A __addsf3
4000cb94 A __divdf3
4000c8f0 A __muldf3
4000c3dc A __mulsf3
4000c688 A __subdf3
4000c268 A __subsf3
```
Explicitly excluded in esp-quick-toolchain post-build script, and via .ld filters

mcspr:
> Note that PIO has a special filter for .ld generation

```diff
diff --git a/tools/platformio-build.py b/tools/platformio-build.py
index 83b7d783..512a0cf9 100644
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -275,17 +275,23 @@ else:
 #

 current_vtables = None
-fp_in_irom = ""
+current_fp = None
 for d in flatten_cppdefines:
     if str(d).startswith("VTABLES_IN_"):
         current_vtables = d
-    if str(d) == "FP_IN_IROM":
-        fp_in_irom = "-DFP_IN_IROM"
+    if str(d).startswith("FP_IN_"):
+        current_fp = "-D" + std(d)
+
 if not current_vtables:
     current_vtables = "VTABLES_IN_FLASH"
     env.Append(CPPDEFINES=[current_vtables])
 assert current_vtables

+if not current_fp:
+    current_fp = "FP_IN_IROM"
+    env.Append(CPPDEFINES=[current_fp])
+assert current_fp
+
 #
 # MMU
 #
@@ -363,9 +369,10 @@ app_ld = env.Command(
     join("$BUILD_DIR", "ld", "local.eagle.app.v6.common.ld"),
     join(FRAMEWORK_DIR, "tools", "sdk", "ld", "eagle.app.v6.common.ld.h"),
     env.VerboseAction(
-        "$CC -CC -E -P -D%s %s %s $SOURCE -o $TARGET"
+        "$CC -CC -E -P -D%s -D%s %s $SOURCE -o $TARGET"
         % (
             current_vtables,
+            current_fp,
             # String representation of MMU flags
             " ".join(
                 [
@@ -373,7 +380,6 @@ app_ld = env.Command(
                     for f in mmu_flags
                 ]
             ),
-            fp_in_irom,
         ),
         "Generating LD script $TARGET",
     ),
```

`if not current_fp:` branch sets default as suggested by the PR.

Repeating the note above - PR changes locations of **all** fp funcs to IROM, not just selected few b/c `-DFP_IN_IROM` becomes default (not true right now, various fp funcs can be either in IROM or IRAM)

d-a-v:
If default was float in iram until v2.7, should we revert this default 3.x change back to as it was ?
I don't remember that it was done on purpose.

d-a-v:
Here is the new menu proposal: default is now in iram, text simplified per review
iramfloat

mcspr:
> If default was float in iram until v2.7, should we revert this default 3.x change back to as it was ?
I don't remember that it was done on purpose.

Tasmota does not seem to use 3.x.x (if ever used it?), so direct comparison does not really work with their builds. .ld change was true for 2.7.4, but 3.x moved even more into IROM. With your changes we move conversions into IROM by default (to, from float / double to signed / unsigned pairs. see gcc link for reference)

I also tried ESPeasy beta builds - IROM opt builds and saves 552 bytes. IRAM does not, too much in there already to fit extra funcs. I wonder if conversions actually break something in some sensor code somewhere that accidentally converts to or from float though

PR #8968, opened 2023-08-03
ESP8266HTTPUpdate: Get available firmware version from update server

hmueller01:
This addresses #8967


hmueller01:
@d-a-v Would you mind merging this, or do you have any comments / improvements on this?

hmueller01:
I kindly like to ask, if someone can merge this. I use this in several projects of mine and always have to use this branch instead of the HEAD. This should not break anything as it more or less just adds an additional function and some very minor changes (like the before noted enum, the move `_httpClientTimeout`, `_followRedirects` from `private`to `protected` which helps later extention of the class and an information output).

d-a-v (approved):
LGTM

PR #8969, opened 2023-08-05
Add virtual beginMulticast(...) stub to UDP class

lmartorella:
Hi,
It seems that (at least) the Arduino core and the ESP32 core both has the base `beginMulticast` virtual definition in the base `UDP` class:

```cpp
     virtual uint8_t beginMulticast(IPAddress, uint16_t) { return 0; }  // initialize, start listening on specified multicast IP address and port. Returns 1 if successful, 0 on failure
``` 

In order to be compatible with libraries like [arduino-libraries/ArduinoMDNS](https://github.com/arduino-libraries/ArduinoMDNS), I've added the same base virtual function at `UDP` level here too.

With these changes, the above library works on a ESP8266 like a charm.

I'm not sure if the different already existing signature of the `WiFiUDP::beginMulticast` should be left here or not. To be backward compatible, I've added both signatures, but not sure if the one with the different signature should be tagged as deprecated or not.

Thanks!
 L

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiUdp.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiUdp.cpp
@@ -86,16 +86,10 @@ uint8_t WiFiUDP::begin(uint16_t port)
 
 uint8_t WiFiUDP::beginMulticast(IPAddress multicast, uint16_t port)
 {
-    IPAddress local = WiFi.localIP();
-    return _beginMulticast(local, multicast, port);
+    return beginMulticast(IP_ADDR_ANY, multicast, port);
 }
 
 uint8_t WiFiUDP::beginMulticast(IPAddress interfaceAddr, IPAddress multicast, uint16_t port)
-{
-    return _beginMulticast(interfaceAddr, multicast, port);
-}
-
-uint8_t WiFiUDP::_beginMulticast(IPAddress interfaceAddr, IPAddress multicast, uint16_t port)
 {
     if (_ctx) {
         _ctx->unref();
--- a/libraries/ESP8266WiFi/src/WiFiUdp.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiUdp.cpp
@@ -84,7 +84,18 @@ uint8_t WiFiUDP::begin(uint16_t port)
     return (_ctx->listen(IPAddress(), port)) ? 1 : 0;
 }
 
+uint8_t WiFiUDP::beginMulticast(IPAddress multicast, uint16_t port)
+{
+    IPAddress local = WiFi.localIP();
+    return _beginMulticast(local, multicast, port);
+}
+
 uint8_t WiFiUDP::beginMulticast(IPAddress interfaceAddr, IPAddress multicast, uint16_t port)
+{
+    return _beginMulticast(interfaceAddr, multicast, port);
+}
+
+uint8_t WiFiUDP::_beginMulticast(IPAddress interfaceAddr, IPAddress multicast, uint16_t port)
 {
     if (_ctx) {
         _ctx->unref();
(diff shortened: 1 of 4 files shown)


libraries/ESP8266WiFi/src/WiFiUdp.cpp
```diff
@@ -84,7 +84,18 @@ uint8_t WiFiUDP::begin(uint16_t port)
     return (_ctx->listen(IPAddress(), port)) ? 1 : 0;
 }
 
+uint8_t WiFiUDP::beginMulticast(IPAddress multicast, uint16_t port)
+{
+    IPAddress local = WiFi.localIP();
```
mcspr, line 89: `IP_ADDR_ANY`?

libraries/ESP8266WiFi/src/WiFiUdp.h
```diff
@@ -47,7 +48,8 @@ class WiFiUDP : public UDP, public SList<WiFiUDP> {
   // Finish with the UDP connection
   void stop() override;
   // join a multicast group and listen on the given port
-  uint8_t beginMulticast(IPAddress interfaceAddr, IPAddress multicast, uint16_t port);
+  virtual uint8_t beginMulticast(IPAddress interfaceAddr, uint16_t port);
+  uint8_t beginMulticast(IPAddress interfaceAddr, IPAddress multicast, uint16_t port) __attribute__((deprecated));
```
mcspr, line 52: Do we need to deprecate, really, if methods do not overlap?
Virtual now works via `UDP` pointer for your case, and `WiFiUDP` user continues to use this one.
(localIP() might not be the only thing happening, any IP available to lwip stack could be used)

libraries/ESP8266WiFi/src/WiFiUdp.cpp
lmartorella: I'm not an expert of multicast UDP and I don't have an opinion about that, so that's sounds good to me.
I have done my search for other usages in the framework, and found both `IP_ADDR_ANY` and `localIP`. BTW since the [doc](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/udp-class.html) is currently suggesting to use `localIP()`, I believe that this was the correct way. Changing that.

libraries/ESP8266WiFi/src/WiFiUdp.h
lmartorella: Yes, ideally we could leave the two calls since they are clean overloads.
However it seems that the signature with two parameters is out there on the AVR core from 2017:

https://github.com/arduino/ArduinoCore-avr/commit/5a05bf01f4ffbe16670eb8d713688d059c3bd084

and the STM32 followed the same standard in 2019:

https://github.com/stm32duino/Arduino_Core_STM32/pull/401

However the change that introduced the new argument in the ESP8266 core was way older, back to 2015:

https://github.com/esp8266/Arduino/commit/fea362a3b53507ece698717b7f869f6befefd6b4

(and it was using `WiFi.localIP()` internally in the commit immediately before the change, not `IP_ADDR_ANY`).

I don't know, if the idea behind the Arduino core is to let developers to write the most "portable" code, I think we should push for a standard, and avoid APIs that are very specific for a specific platform only, especially at these levels of abstraction. For instance, why not having both the overloads at the base `UDP` level? 
Ideally, it would be more logic having the overload with 3 arguments declared as the virtual "abstract" one, and then the other smaller overloads directly defined at `UDP` base class as non-virtual functions using the default argument values.

So I feel that we should actually trying to converge to a single portable API between cores, making easier for library devs to write the code only once (and telling the ESP8266 devs that a more standard API is available for portability).

BTW let me remove the warning, everything could be done in a separate thread if really needed.
Thx!  L

PR #8970, opened 2023-08-06
Add url redirect

hmueller01:
Add redirect function like [ESPAsyncWebServer](https://github.com/me-no-dev/ESPAsyncWebServer).
Sorry for the wrong commits/reverts. Obviously my master wasn't clean.


hmueller01:
Is it possible to merge this, also there is a build problem on Linux? Seems to be not the fault of this PR ...

d-a-v:
What about using this API [in our examples](https://github.com/search?q=repo%3Aesp8266%2FArduino%20send\(302&type=code) ?

hmueller01:
@d-a-v Sorry, but I do not understand your comment. These are examples that do a redirect. My proposal was to add this out-of-the-box of the ESP8266WebServer ...

d-a-v:
Sorry for lack of clarity.
You added  a new API.
I propose you to update our examples to use it.
They would become examples of how to use your new API, examples are meant for that.
Lots of users take them as basis to start their project.
My link above shows you where you *could* make these changes.

hmueller01:
Oh, sry for being so stupid. Sure, I'll do this in the next few days.

hmueller01:
Looking through the examples I found this:
```
  server.send(302, "text/plain", "");  // Empty content inhibits Content-length header so we have to close the socket ourselves.
  server.client().stop();  // Stop is needed because we sent no content length
```
Why is it needed to stop the client? Should this be done in `redirect()` as well, if I do not send content?

d-a-v:
> Why is it needed to stop the client? Should this be done in `redirect()` as well, if I do not send content?

It seems to be a good idea.

hmueller01:
@d-a-v Ok, I finished my work on this. Also did some tests and it looks good.

hmueller01:
Fixed. All checks have passed. 👍

hmueller01:
I just merged with master and it still looks clean.
@d-a-v Is this ready form merging now (and adding to a new release you just mentioned)?

mcspr:
(sry for delay)

I don't think you have to rename params this much.
Request object is the one we care about the most, either 'friend'ing the Server class or simply using different signature to access the vector would probably be a shorter solution?
New Server method can keep its name

d-a-v (approved):

PR #8978, opened 2023-08-23
`#include <exception>` for a direct dependency

jpinkul:
I was recently trying to compile this project with a more recent version of GCC and hit an issue with this file. With older versions the `std::set_terminate` definition was being included via a transitive include of `<exception>` from the C++ standard library implementation. This is no longer the case with newer versions of `libstdc++`. This fixes the issue by using a direct include instead of relying on an implementation specific transitive include.


earlephilhower:
This seems fine, but using exceptions unpatched from a generic GCC will result in the exception unwind tables living in RAM.  It'll work, but you're going to spend a large portion of usable RAM on them.  See the esp-quick-toolchain for some patches against the current GCC.

PR #8985, opened 2023-09-08
updating examples in ESP8266WiFi

ktauchathuranga:
in this example update I added a basic Telnet server example and a basic Telnet client example.
with that, it should be easy to understandd the concept and adapt to the projects!


d-a-v:
Did you check this [example](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/WiFiTelnetToSerial/WiFiTelnetToSerial.ino) ?

PR #8995, opened 2023-09-23
WiFiServer - operator bool() and method end()

JAndrassy:
I did a research how different Arduino networking libraries implement the Arduino networking API.
https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md#server-class

For most libraries it is simpler to implement operator bool() than the method status() with its return value constants. So this PR adds bool operator for WiFiServer in the ESP8266Wifi library.

In the research I find out that end() is the most popular name for the function to stop the server and Arduino now used it too in the new WiFiS3 library.

### DIFF
--- a/libraries/ESP8266WiFi/src/WiFiServer.h
+++ b/libraries/ESP8266WiFi/src/WiFiServer.h
@@ -100,6 +100,8 @@ public:
   uint16_t port() const;
   void close();
   void stop();
+  void end();
+  explicit operator bool();
 
   using ClientType = WiFiClient;
 
--- a/libraries/ESP8266WiFi/src/WiFiServer.cpp
+++ b/libraries/ESP8266WiFi/src/WiFiServer.cpp
@@ -173,6 +173,14 @@ void WiFiServer::stop() {
     close();
 }
 
+void WiFiServer::end() {
+    close();
+}
+
+WiFiServer::operator bool() {
+  return (status() != CLOSED);
+}
+
 err_t WiFiServer::_accept(tcp_pcb* apcb, err_t err) {
     (void) err;
     DEBUGV("WS:ac\r\n");


earlephilhower:
That's a darn odd failure in CI.  I have no idea what's going on there w/host builds and don't think it's related to this at all.

That said, maybe add those methods to wired Ethernet too?  https://github.com/esp8266/Arduino/blob/master/cores/esp8266/LwipIntfDev.h

Would you also like to do the same PR for https://github.com/earlephilhower/arduino-pico ? I could add the add'l functions, but since you've identified the problem I don't want to steal your thunder.  We now have wired and WiFi support there, ~stolen~ borrowed from here. :laughing:

JAndrassy:
@earlephilhower I already do to many PR and write issues as a result of the research. Most of the [PR] and [issue] links in the document are my PR and issues. Please do the changes in the Pico core.
 
btw I didn't include the libraries from pico core in the research since the WiFi is only for Pico W and the libraries for wired Ethernet were not ready when I started :-)

libraries/ESP8266WiFi/src/WiFiServer.h
```diff
@@ -100,6 +100,8 @@ class WiFiServer {
   uint16_t port() const;
   void close();
   void stop();
+  void end();
+  virtual operator bool();
```
mcspr, line 104: `virtual` / non-virtual here?
I see at least esp32 has a normal one, note of the difference

mcspr, line 104: `explicit` and `const`?

This is implicit conversion, and c++ (also implicit) defaults for operator declaration have many repercussions
https://en.cppreference.com/w/cpp/language/implicit_conversion

`explicit` restricts this to the `if (obj)` construct mentioned in the Arduino reference
https://www.arduino.cc/reference/en/libraries/ethernet/ifserver/

mcspr (commented):
Same q as above, don't we want this in `Server` base?
Otherwise, same thing must be copied to Lwip server implementations

libraries/ESP8266WiFi/src/WiFiServer.h
JAndrassy: it doesn't have to be virtual. I removed `virtual`

JAndrassy: yes. `explicit` is better.
strange `Serial.println(server);` stopped working while `Serial.println(!server);` works.

d-a-v (approved):

PR #9006, opened 2023-10-16
Fix memory leaks if handshake fails

igor-epam:


igor-epam:
> This way we lose last bearssl error since you delete the context obj with `_freeSSL()` call. You might want to save result of `br_ssl_engine_last_error` some place beforehand
> 
> https://github.com/esp8266/Arduino/blob/1662248b394740d82824f504a021d4dad3d314b6/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L1347

@mcspr Thank you for the hint. It was done similarly to existing code: https://github.com/esp8266/Arduino/blob/1662248b394740d82824f504a021d4dad3d314b6/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L1188

d-a-v:
@igor-epam What do you think of saving the content of `br_ssl_engine_last_error` in a "releaseSSL()" method that would also call `_freeSSL()` ?

PR #9007, opened 2023-10-16
Renaming the cities of Ukraine

rublin:
Fix the naming of some cities in the Ukrainian way


d-a-v:
We use this database: https://github.com/nayarsystems/posix_tz_db

Any local change would be overwritten when making a new release.
Can you address your request to them ?

([this script does the update](https://github.com/esp8266/Arduino/blob/master/tools/TZupdate.sh))

rublin:
> We use this database: https://github.com/nayarsystems/posix_tz_db
> 
> Any local change would be overwritten when making a new release. Can you address your request to them ?
> 
> ([this script does the update](https://github.com/esp8266/Arduino/blob/master/tools/TZupdate.sh))

Sure, no problem. I'm going to address the request to them.

mcspr:
Europe/Simferopol, Europe/Kyiv and Europe/Kiev are from raw tzdata, see https://www.iana.org/time-zones
(tzdb-2023c.tar.lz at the time of writing this; see `europe` file inside the archive. [also note that script above only knows about Europe/Kiev, there are supposed to be both entries](https://github.com/nayarsystems/posix_tz_db/blob/3edd201fa1131db7d09fd90cbb66d2f33d8915b3/gen-tz.py#L347))

I'd also suggest that we might want to simplify the script a bit and use python and pytz since it already has access to zoneinfo data and we can filter and produce header in the same pass
Something like this - https://github.com/xoseperez/espurna/blob/dev/code/scripts/tzdata.py - plus, there would be no need to maintain the tz name list

d-a-v:
@mcspr good idea.
Your link seems to be broken.

PR #9008, opened 2023-10-17
WiFi.BSSID and scan result BSSID with parameter as in WiFi libraries by Arduino

JAndrassy:
https://www.arduino.cc/reference/en/libraries/wifi/wifi.bssid/


d-a-v (approved):
Thanks

PR #9011, opened 2023-10-28
add Stream::readStringUntil function that uses string terminator

Arian8j2:
currently there is a
```c++
String readStringUntil(char terminator)
```
function, but there is no equivalent of it that uses string terminator. this pr implements
```c++
String readStringUntil(const char* terminator, uint32_t count = 1)
```
i found this useful for small html parsing or truncating.

### DIFF
--- a/cores/esp8266/Stream.cpp
+++ b/cores/esp8266/Stream.cpp
@@ -262,6 +262,32 @@ String Stream::readStringUntil(char terminator) {
     return ret;
 }
 
+String Stream::readStringUntil(const char* terminator, uint32_t count) {
+    String ret;
+    int c;
+    uint32_t termCount = 0;
+    size_t termLen = strlen(terminator);
+    size_t termIndex = 0;
+    size_t index = 0;
+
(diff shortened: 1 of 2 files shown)


d-a-v:
It is not a breaking change and does not override the Arduino API, so looks good.
However the `count` name is not intuitive.
Is it a ` int until_total_number_of_occurrences = 1` ?

Arian8j2:
> Is it a ` int until_total_number_of_occurrences = 1` ?

yes

d-a-v:
Can you please add a comment or rename the variable ?

Arian8j2:
> Can you please add a comment or rename the variable ?

done

d-a-v (approved):
Approving: Next step is to simplify [our calls to `readStringUntil()`](https://github.com/search?q=repo%3Aesp8266%2FArduino%20readStringUntil&type=code) from our libraries (especially but not only `parsing-impl.h`)

*edit* (or improve `Stream::sendUntil()` with this same new API) - see #9005)

PR #9020, opened 2023-11-07
WiFiSTA - allow using DHCP again after disconnecting static IP

JAndrassy:
reset `_useStaticIp` to false in `WiFi.disconnect()`

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp
@@ -400,6 +400,8 @@ bool ESP8266WiFiSTAClass::disconnect(bool wifioff, bool eraseCredentials) {
         WiFi.enableSTA(false);
     }
 
+    _useStaticIp = false;
+
     return ret;
 }
 


d-a-v (approved):
LGTM

mcspr:
https://github.com/esp8266/Arduino/blob/1c624dd/doc/esp8266wifi/station-class.rst#config
> If you pass all three parameter as 0.0.0.0 (local_ip, gateway and subnet), it will re enable DHCP

In our docs this is mentioned right at the end. Might be more discover-able if moved to the top? Plus, what updated disconnect() behaviour from this PR.

JAndrassy:
@mcspr 
Now I know about config(0,0,0,0). it looks like a workaround. 
Do you think resetting back to DHCP  in `disconnect`is not right? That some sketch may want after invoking disconnect() join the network with same previously configured static IP? 

I try to find the common Arduino WiFi API so there are as few differences as possible between various WiFi libraries. But `config(0, 0, 0, 0)` would be hard 'hard to sell'. Having `disconnect()` reset to DHCP is much simpler.

mcspr:
I only thought about docs... But, considering some examples, that definitely breaks something like this? Where user would expect config() to persist and does not to always pair config() and begin(). Not sure how other cores and libs behave here
```
setup() {
  WiFi.config()
}
loop() {
  WiFi.begin(foo, bar)
  ... work ...
  WiFi.disconnect()
}
```

For config, might be another method?
```
config() {
  config(0, 0, 0, 0)
}
```

JAndrassy:
I tested 7 WiFi libraries. 
ESP32 WiFi has the same _useStaticIP so they got the same PR from me and merged it. The new Arduino WiFiS3 library has a fw written with esp32 Arduino so it has the problem because esp32 has it.
WiFiNINA's nina-fw is written with ESP-IDF but copied code from esp32 Arduino so they too have _useStaticIP.
With two libraries disconnect() worked to allow DHCP at next begin. New Arduino WiFiC3 library and my old WiFiEspAT library.
In WiFi101 `begin` hangs after `disconnect()` even going from DHCP to static IP.  `end()` worked.

JAndrassy:
I am too tired now, but I will revert this tomorrow. In esp32 too. config(0) gives more control to the user of the library. I will try to steer other libraries to config(0), if they don't support it already.

PR #9021, opened 2023-11-08
WiFiSTA - method setDNS as in WiFi libraries by Arduino

JAndrassy:
Arduino doc: https://www.arduino.cc/reference/en/libraries/wifi/wifi.setdns/

I [compared more than a dozen of Arduino Networking libraries](https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md) to find the common API for Arduino networking. Then I wrote [a test sketch](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/WiFiTest/WiFiTest.ino) which checks how a library complies to this common API.

ESP8266 Arduino WiFi library was based on the first WiFi library by Arduino. Then ESP32 Arduino WiFi library was based on ESP8266 Arduino WiFi library. Arduino then created more WiFi libraries and the API diverged, but at least with the most recent library they looked at the ESP32 WiFi library for new method names (for example dnsIP(n)).

While doing the research and tests I do PR in Arduino networking libraries repositories with in most cases simple modifications to unify the API where it doesn't break the library (for example I will not ask esp8266 and esp32 maintainers to make WiFi.begin blocking :-) ).


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -46,6 +46,7 @@ class ESP8266WiFiSTAClass: public LwipIntf {
         //to detect Arduino arg order, and handle it correctly. Be aware that the Arduino default value handling doesn't
         //work here (see Arduino docs for gway/subnet defaults). In other words: at least 3 args must always be given.
         bool config(IPAddress local_ip, IPAddress gateway, IPAddress subnet, IPAddress dns1 = (uint32_t)0x00000000, IPAddress dns2 = (uint32_t)0x00000000);
```
d-a-v, line 48: [If you can also fix this :\]](https://github.com/esp8266/Arduino/pull/9022#discussion_r1387252542)

JAndrassy: should I change it for `config` too to INADDR_ANY?

d-a-v (approved):

PR #9023, opened 2023-11-09
LwipIntfDev - method end() to enable repeated begin

JAndrassy:
While it is not a common use, I think it should be possible to do Ethernet.end() and then Ethernet.begin() with a different static IP or return to DHCP after using a static IP.

I tested it with [LegacyEthernetTest](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/LegacyEthernetTest/LegacyEthernetTest.ino) with hardwareStatus and linkStatus commented out (they don't work before begin here) and with hostByName commented out and with the getters/setters PR.

### DIFF
--- a/cores/esp8266/LwipIntfDev.h
+++ b/cores/esp8266/LwipIntfDev.h
@@ -68,6 +68,7 @@ public:
 
     // default mac-address is inferred from esp8266's STA interface
     boolean begin(const uint8_t* macAddress = nullptr, const uint16_t mtu = DEFAULT_MTU);
+    void    end();
 
     const netif* getNetIf() const
     {
@@ -138,6 +139,7 @@ protected:
     int8_t   _intrPin;
     uint8_t  _macAddress[6];
     bool     _started;
+    bool     _scheduled;
     bool     _default;
 };
 
@@ -229,6 +231,7 @@ boolean LwipIntfDev<RawDev>::begin(const uint8_t* macAddress, const uint16_t mtu
     if (!netif_add(&_netif, ip_2_ip4(&ip_addr), ip_2_ip4(&netmask), ip_2_ip4(&gw), this,
                    netif_init_s, ethernet_input))
     {
+        RawDev::end();
         return false;
     }
 
@@ -242,10 +245,11 @@ boolean LwipIntfDev<RawDev>::begin(const uint8_t* macAddress, const uint16_t mtu
             break;
 
         case ERR_IF:
+            RawDev::end();
             return false;
 
         default:
-            netif_remove(&_netif);
+            end();
             return false;
         }
     }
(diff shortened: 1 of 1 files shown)


cores/esp8266/LwipIntfDev.h
```diff
@@ -272,22 +274,39 @@ boolean LwipIntfDev<RawDev>::begin(const uint8_t* macAddress, const uint16_t mtu
...
-            100))
+    if (_intrPin < 0 && !_scheduled)
     {
-        netif_remove(&_netif);
-        return false;
+        if (schedule_recurrent_function_us(
+                [&]()
+                {
+                    this->handlePackets();
+                    return true;
```
d-a-v, line 283: You might want to return `_scheduled` here, which would stop and remove the scheduled function when it is set to `false` in `::end()`.
Then maybe `_started` can be used in place of this new `_scheduled`.
(For sanity, the one or the other would have to be `true` before the call to `schedule_recurrent_function_us()`)

JAndrassy: I didn't know how to remove the scheduled function. I rework it.

JAndrassy: done. I like it

cores/esp8266/LwipIntfDev.h
```diff
@@ -272,22 +273,43 @@ boolean LwipIntfDev<RawDev>::begin(const uint8_t* macAddress, const uint16_t mtu
...
+    static bool _scheduled = false;
+
+    if (_intrPin < 0 && !_scheduled)
+    {
+        _scheduled = schedule_recurrent_function_us(
             [&]()
             {
                 this->handlePackets();
+                if (!_started)
+                {
```
d-a-v, line 285: You can use `_started` in place of `_scheduled` which become useless.
Inside, `handlePackets()` should be called *after* checking that scheduled function must be removed.

JAndrassy: if `begin` is executed immediately after `end`, `started` changes to true before the scheduled function is removed. the function is then scheduled again and the previous schedule is not removed.

with `_scheduled` the function stays scheduled if begin is executed before `_scheduled` changes to false. the schedule is removed if it executes before `begin` is invoked from the sketch.

d-a-v: Right, with `_started` alone there would be a risk of having two scheduled functions when `begin()` is immediately called after `end()`.

There is still an issue with `_scheduled` because it is static. This prevents to use more than one instance.

And `handlePackets()` must be called *after* the `(!_started)` otherwise it would be called after `Rawdev::end()` is called.

cores/esp8266/LwipIntfDev.h
```diff
@@ -272,22 +274,41 @@ boolean LwipIntfDev<RawDev>::begin(const uint8_t* macAddress, const uint16_t mtu
...
                 return true;
             },
-            100))
-    {
-        netif_remove(&_netif);
-        return false;
+            100);
+        if (!_scheduled)
+        {
+            netif_remove(&_netif);
```
d-a-v, line 293: This should also be added
```cpp
RawDev::end();
_started = false;
```
also master-missing `RawDev::end();` should be added above where `return false` are.

JAndrassy: so `end()`

d-a-v: (also lines  [234](https://github.com/esp8266/Arduino/pull/9023/files#diff-ced83b441ac52d743d09c392b44aa2a6f9ec68e68ca1f8fe788fdd74ae125daaR234), [247](https://github.com/esp8266/Arduino/pull/9023/files#diff-ced83b441ac52d743d09c392b44aa2a6f9ec68e68ca1f8fe788fdd74ae125daaR247) and [251](https://github.com/esp8266/Arduino/pull/9023/files#diff-ced83b441ac52d743d09c392b44aa2a6f9ec68e68ca1f8fe788fdd74ae125daaR251))

JAndrassy: this way?

d-a-v (approved):

PR #9024, opened 2023-11-10
EthernetCompat - static IP auto gw,mask,dns as in Arduino libraries

JAndrassy:
The parameters of `begin` in EthernetCompat.h are optional but automatic values for gw IP, mask and DNS IP were not calculated.

In the `begin(mac, local_ip, arg1)`, arg1 is assumed to be the DNS IP, because it can be an address outside of the local network. Automatic gw IP based on local_ip with .1 at the end is more likely to be a right auto value than DNS IP. Basically this is the reason for the Arduino ordering of begin parameters.

### DIFF
--- a/libraries/lwIP_Ethernet/src/EthernetCompat.h
+++ b/libraries/lwIP_Ethernet/src/EthernetCompat.h
@@ -38,10 +38,29 @@ public:
     // Arduino-Ethernet API compatibility, order can be either:
     // mac, ip, gateway, netmask, dns (esp8266 or natural order)
     // mac, ip, dns, gateway, netmask (Arduino legacy)
-    boolean begin(const uint8_t* macAddress, const IPAddress& local_ip = IPADDR_NONE,
-                  const IPAddress& arg1 = IPADDR_NONE, const IPAddress& arg2 = IPADDR_NONE,
-                  const IPAddress& arg3 = IPADDR_NONE)
+    boolean begin(const uint8_t* macAddress, IPAddress local_ip = IPADDR_NONE,
+                  IPAddress arg1 = IPADDR_NONE, IPAddress arg2 = IPADDR_NONE,
+                  IPAddress arg3 = IPADDR_NONE)
     {
+        if (local_ip.isSet() && local_ip.isV4())
+        {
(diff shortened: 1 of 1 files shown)


d-a-v:
We already have a [logic reordering network parameters](https://github.com/esp8266/Arduino/blob/31c1592ad6e1153dfaa9957f5b64e0f1920d4d5a/cores/esp8266/LwipIntf.cpp#L61-L94) and it is [used in `config()`](https://github.com/esp8266/Arduino/blob/31c1592ad6e1153dfaa9957f5b64e0f1920d4d5a/cores/esp8266/LwipIntfDev.h#L145-L156) which is called just below your proposal.

Did you check if the operations are similar ?

Also, `const&` are there for IPAddress because IPv6 addresses are more than one single `uint32`.

JAndrassy:
this doesn't do a reordering. with no param entered it creates arduino ordering, which is then reordered in the mentioned funtion

d-a-v:
> IPAddress with IPv6 is not copyable?

They are copyable, but when IPv6 is enabled they are bigger than a single reference or a single IPv4.
So it takes less memory / less time to transfer them as constref. It is not a big deal, there are many places where they are passed by copy. In this PR you remove the constref so I explain why they were there: I find it to be a good reason to recall that they can be bigger.

d-a-v:
> this doesn't do a reordering. with no param entered it creates arduino ordering, which is then reordered in the mentioned funtion

OK I misread it.
The goal is to ensure that there is not nothing for IPv4 dns and gateway, by copying static address and setting `.1` on the last digit (why not `.254`?). Out of curiosity is it specified somewhere in their doc or is it only something that is done over many of their implementations from the beginning ?

Why are you not doing this in our global `config()` (link above) so this arduino logic would spread over all type of interface ?

JAndrassy:
https://www.arduino.cc/reference/en/libraries/ethernet/ethernet.begin/
these Arduino docs are now imported to the web page from GitHub
https://github.com/arduino-libraries/Ethernet/blob/master/docs/api.md

d-a-v (approved):
a step towards API coherence & compatibility
thanks

PR #9026, opened 2023-11-11
WiFiServer - allow constructor without parameters

JAndrassy:
with `begin(port)` available, the constructor doesn't have to get `port`. I hope you like the default port number 23.


d-a-v:
After re-reading the official API, you're telling that
- Official and esp8266 Server constructor always have a port argument
- `Server::begin()` can have a port argument or no argument

So you're adding a constructor without argument for coherency with `Server::begin(port)`, is that so ?

JAndrassy:
yes. the newest Arduino libraries have it so I consider it now as part of the official API. (And users ask for it for years now)

https://github.com/arduino/ArduinoCore-renesas/blob/3a1b8a33fb61d0a286f400922576e6736bab859a/libraries/WiFiS3/src/WiFiServer.h#L37
https://github.com/arduino/ArduinoCore-renesas/blob/main/libraries/WiFi/src/WiFiServer.h + https://github.com/arduino/ArduinoCore-renesas/blob/main/libraries/lwIpWrapper/src/lwipServer.h

d-a-v (approved):
default port 23 instead of 80 on other implementations is one of the spices used for the taste of this core.

JAndrassy:
> default port 23 instead of 80 on other implementations is one of the spices used for the taste of this core.

I know :-)

PR #9031, opened 2023-11-15
WiFi - static IP auto gw,mask,dns as in Arduino libraries

JAndrassy:
a version of `config` with two parameters local_ip and dns IP. dns IP is optional. gateway IP is calculated, dns IP is calculated if not provided.  then the other `config` is invoked. 

this now allows, these versions
```
config(ip); <- new
config(ip, dns); <- new
config(ip, dns, gw);
config(ip, gw, mask);  
config(ip, dns, gw, mask);
config(ip, gw, mask, dns);
config(ip, dns, gw, mask, dns2);
config(ip, gw, mask, dns, dns2);
```
it also allows `WiFi.config(INADDR_NONE)` to return to DHCP as in other WiFi libraries.


d-a-v (approved):

TD-er:
I'll link my comment here also, as I'm really not happy with this new `config` function.
https://github.com/esp8266/Arduino/commit/9a4e17876e782f58577829e068d4005c9d4e9b66#r134164761

IMHO, it is an inconvenience function, not a convenience function because it makes incorrect assumptions about what will be universally applicable.

PR #9040, opened 2023-11-29
LwipIntfDev - config static IP auto gw,mask,dns as in Arduino libraries

JAndrassy:
so one more, This one is for 'modern' Ethernet libraries W5100lwIP, W5500lwIP and ENC28J60lwIP used without EthernetCompat.
I wrote [a test](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/ModernEthernetTest/ModernEthernetTest.ino) for what I consider a modern Ethernet library (one with config for static IP as in WiFi libraries) and to my surprise these libraries matched my idea. Only the automatic dns, gw, subnet values were not supported here again.

So here is the third PR for auto gw,mask,dns as in Arduino libraries. First was in 'begin' for EthernetCompat, second was `config` with two parameters in WiFiSTA.
I had to move the clearing of localIP to `EthernetCompat.end`, because in library with `config`, `disconnect`/`end` should not clear static IP settings. (config(INADDR_NONE) should))

### DIFF
--- a/cores/esp8266/LwipIntfDev.h
+++ b/cores/esp8266/LwipIntfDev.h
@@ -63,8 +63,13 @@ public:
         memset(&_netif, 0, sizeof(_netif));
     }
 
+    //The argument order for ESP is not the same as for Arduino. However, there is compatibility code under the hood
+    //to detect Arduino arg order, and handle it correctly.
     boolean config(const IPAddress& local_ip, const IPAddress& arg1, const IPAddress& arg2,
                    const IPAddress& arg3 = IPADDR_NONE, const IPAddress& dns2 = IPADDR_NONE);
+
+    // two and one parameter version. 2nd parameter is DNS like in Arduino. IPv4 only
+    boolean config(IPAddress local_ip, IPAddress dns = INADDR_ANY);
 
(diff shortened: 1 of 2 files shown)


d-a-v:
Thanks for checking and keeping compatibility of the network API across every arduino implementations.

> to my surprise these libraries matched my idea. 

(what were you thinking ?)

> I had to move the clearing of localIP to `EthernetCompat.end`, because in library with `config`, `disconnect`/`end` should not clear static IP settings. (config(INADDR_NONE) should))

Can you recall where these two requirements are stated (one should not, the other should) ?
Even if you're not arduino official, a link from our documentation to your [NetApiHelpers](https://github.com/JAndrassy/NetApiHelpers/) might help anyway.

JAndrassy:
> Can you recall where these two requirements are stated (one should not, the other should) ?

It is related to my PR which was then reverted because I found out that all libraries have it as exp8266 WiFi. 
relevant comment https://github.com/esp8266/Arduino/pull/9020#issuecomment-1806789080

d-a-v:
Thanks
Given @mcspr's suggestion in the 9020 comment, we'll update the documentation ourselves.

PR #6696, opened 2019-11-01
Add Uri with support for regexUri and globUri

Bmooij:
Add support for regex uri and glob uri.
This pull request is a combination of:
https://github.com/esp8266/Arduino/pull/5214
https://github.com/esp8266/Arduino/pull/5467

There is an issue when importing `<regex>`

PR #9041, opened 2023-11-29
use uriglob to catch all android captive portal requests

dbu:
i observed that i get lots of `/generate_204_<uuid>` requests when building the captive portal and connecting with an android phone. since #6696, the web server supports uri globs (and other features) to match requests. with the glob we can also redirect the requests with the uuid suffix.


dbu:
good point. i adjusted the comment (and hopefully fixed the build error by aligning the doc comments)

d-a-v:
CI wants an additional space (or you can run it yourself and commit its changes, see above)

dbu:
thank you a lot for maintaining this awesome library!

PR #9042, opened 2023-12-02
doc: config(0,0,0) to restore dhcp + mention portable network API

d-a-v:
reference: https://github.com/esp8266/Arduino/pull/9040#issuecomment-1835753590
@JAndrassy @mcspr


JAndrassy:
so NetApiHelpers library has the test sketches, but the API research and documentation is in repository
https://github.com/JAndrassy/Arduino-Networking-API

d-a-v:
@JAndrassy are you OK with the rewording about the mention of the network API research ?

PR #9045, opened 2023-12-06
document WiFi.config Arduino parameters versions

JAndrassy:


doc/esp8266wifi/station-class.rst
```diff
@@ -120,6 +122,19 @@ The following IP configuration may be provided:
    (like e.g. *www.google.co.uk*) and translate them for us to IP
    addresses
 
+For Arduino networking API compatibilty the ESP8266WiFi library supports for IPv4 additional versions of the ``config`` function.
```
mcspr, line 125: ...4 additional versions...?

mcspr (approved):
Thanks!

doc/esp8266wifi/station-class.rst
JAndrassy: two are new and two are just the 'good old' parameter ordering detection

mcspr: Sry I wasn't clear enough.
Right now - 'library supports for IPv4 additional versions of the `config` function'
Did you mean the above? 'library supports 4 additional versions'

mcspr: 'For Arduino networking API compatibility this library also provides 4 additional versions of the ``config`` function. These versions only support IPv4 addresses.'
'library supports for' -> 'library supports', my main issue here :)
can also mention as 'IPv4-only versions'

d-a-v: fixed by #9050

PR #9050, opened 2023-12-08
WiFi config(): warning for legacy idioms

d-a-v:
reference: https://github.com/esp8266/Arduino/pull/9031#issuecomment-1839256347
@TD-er 
@JAndrassy 

reference: https://github.com/esp8266/Arduino/pull/9045#discussion_r1420859947
@mcspr

### DIFF
--- a/doc/esp8266wifi/station-class.rst
+++ b/doc/esp8266wifi/station-class.rst
@@ -99,11 +99,11 @@ config
 
 Disable `DHCP <https://en.wikipedia.org/wiki/Dynamic_Host_Configuration_Protocol>`__ client (Dynamic Host Configuration Protocol) and set the IP configuration of station interface to user defined arbitrary values. The interface will be a static IP configuration instead of values provided by DHCP.
 
-Note that to reenable DHCP, all three parameters as 0.0.0.0 (local_ip, gateway and subnet as ``INADDR_ANY``) must be passed back to config() and re-connecting is needed.
+Note that to reenable DHCP, all three parameters (local_ip, gateway and subnet) as IPv4 ``0U`` (= 0.0.0.0) must be passed back to config() and re-connecting is needed.
 
 .. code:: cpp
 
-    WiFi.config(local_ip, gateway, subnet)             (discouraged, insufficiently accurate, for Arduino API portability)
+    WiFi.config(local_ip, gateway, subnet)             (for Arduino API portability, discouraged as chosen defaults may not match the local network configuration)
     WiFi.config(local_ip, gateway, subnet, dns1)
     WiFi.config(local_ip, gateway, subnet, dns1, dns2)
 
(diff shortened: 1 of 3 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -49,6 +49,7 @@ class ESP8266WiFiSTAClass: public LwipIntf {
 
         // two and one parameter version. 2nd parameter is DNS like in Arduino
         // IPv4 only
+        [[deprecated("It is discouraged to use this 1 or 2 parameters network configuration legacy function config(ip[,dns]) because it sets the router (and ommited dns) to an arbitrary address")]]
```
TD-er, line 52: It doesn't set the router, but either sets the gateway, or makes assumptions about the router's IP-address.

doc/esp8266wifi/station-class.rst
```diff
@@ -122,16 +122,16 @@ The following IP configuration may be provided:
...
+For Arduino networking API compatibility, the ESP8266WiFi library supports IPv4-only additional versions of the ``config`` function:
 
 .. code:: cpp
 
-    WiFi.config(local_ip)
-    WiFi.config(local_ip, dns)
-    WiFi.config(local_ip, dns, gateway)
+    WiFi.config(local_ip)                        (discouraged, for retro compatibility: dns=gateway=···.1)
+    WiFi.config(local_ip, dns)                   (discouraged, for retro compatibility: gateway=···.1)
+    WiFi.config(local_ip, dns, gateway)          (discouraged, for retro compatibility: subnet/netmask quite unavoidable)
```
TD-er, line 131: "subnet/netmask quite unavoidable" is a bit unclear.
Maybe better: "Assumes subnet mask of 255.255.255.0"

And "retro compatibility" with what?
I know it is retro Arduino compatibility.
But maybe that's a bit too long to repeat?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -49,6 +49,7 @@ class ESP8266WiFiSTAClass: public LwipIntf {
 
         // two and one parameter version. 2nd parameter is DNS like in Arduino
         // IPv4 only
+        [[deprecated("It is discouraged to use this 1 or 2 parameters network configuration legacy function config(ip[,dns]) because it is insufficiently accurate")]]
```
TD-er, line 52: "insufficiently accurate" is a bit confusing.
Maybe ".. as chosen defaults may not be universally applicable"
Or ".. as chosen defaults may not match the local network configuration"

I think the last one may be most telling to a user that his program might work at his desk, but not for the neighbor he's making the device for.

doc/esp8266wifi/station-class.rst
```diff
@@ -122,16 +122,16 @@ The following IP configuration may be provided:
...
-    WiFi.config(local_ip)
-    WiFi.config(local_ip, dns)
-    WiFi.config(local_ip, dns, gateway)
+    WiFi.config(local_ip)                        (discouraged, insufficiently accurate, for Arduino API portability)
+    WiFi.config(local_ip, dns)                   (discouraged, insufficiently accurate, for Arduino API portability)
+    WiFi.config(local_ip, dns, gateway)          (discouraged, insufficiently accurate, for Arduino API portability)
     WiFi.config(local_ip, dns, gateway, subnet)
 
-Versions where some of ``dns``, ``gateway`` and ``subnet`` parameters are not specified use a default value. Default ``subnet`` is 255.255.255.0. Default ``gateway`` and ``dns`` are derived from ``local_ip`` by changing the last number to 1. To return to DHCP you can use ``WiFi.config(INADDR_NONE);``.
+Versions where some of ``dns``, ``gateway`` and ``subnet`` parameters are not specified use a default value. Default ``subnet`` is 255.255.255.0. Default ``gateway`` and ``dns`` are derived from ``local_ip`` by changing the last number to 1 (discouraged). To return to DHCP you can use ``WiFi.config(INADDR_NONE);``.
```
TD-er, line 134: instead of "(discouraged)" you could write "(discouraged as it may not apply to any local network)"

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
d-a-v: done

JAndrassy:
I want to add that long before Arduino I always did setup my networks this way so I like the Arduino defaults and I always use config(ip) if I use static IP in a sketch.

d-a-v:
> I want to add that long before Arduino I always did setup my networks this way so I like the Arduino defaults and I always use config(ip) if I use static IP in a sketch.

So you're going to see more logs during compilation on esp8266.

It it OK for both of you @JAndrassy @TD-er ?

TD-er:
Yep looks OK to me.

PR #9072, opened 2024-01-18
Add 2 functions to adjust timeout value in WiFiClient.h and WiFiClient.cpp

RobertGnz:
Default value of timeout is 5000 ms but currently it can't be changed in case the user think it is inadequate for its application.
For ESP32 we already have the possibility to specifie the timeout value as an argument of the connect function.
The aim of the request is to provide the possibilty to specifie a value for ESP8266 too.

Usage example:
int oldValue = getTimeout();
setTimeout( newValue );
client.connect( args );
setTimeout( oldValue );

WiFiClient.h: 
 // Default timeout is set to 5000 ms 
  // setTimeout gives the possibility to adjust this value.
  // getTimeout return the timeout current value.  
  bool setTimeout( int );
  int getTimeout();

WiFiClient.cpp:
// In case you need to increase/decrease timeout current value
bool WiFiClient::setTimeout( int timeout_ms )
{
   if ( timeout_ms <= 0 || timeout_ms > 3600000 ) return(false); // More than 0 and less than 1 hour
   _timeout = timeout_ms;
   if (!_client) return(true);
   else _client->setTimeout(_timeout);
   return(true);
}

int WiFiClient::getTimeout()
{
	return( _timeout );
}

### DIFF
--- a/doc/esp8266wifi/client-class.rst
+++ b/doc/esp8266wifi/client-class.rst
@@ -29,6 +29,35 @@ Default input value 0 means that effective value is left at the discretion of th
 
 ``stop()`` returns ``false`` in case of an issue when closing the client (for instance a timed-out ``flush``). Depending on implementation, its parameter can be passed to ``flush()``.
 
+abort
+~~~~~
+
+.. code:: cpp
+
+   void abort();
+
+
+Originally proposed in `#8738 <https://github.com/esp8266/Arduino/pull/8738>`__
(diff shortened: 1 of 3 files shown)


mcspr:
You are pointing to the wrong branch in your repo ^
https://github.com/RobertGnz/Arduino/branches

RobertGnz:
> You are pointing to the wrong branch in your repo ^ https://github.com/RobertGnz/Arduino/branches

OK I made a new request hoping its good now.

PR #9076, opened 2024-02-01
ESP8266WebServer - fix possible memory leak in request argument handling

everslick:
* fix possible leak of _postArgs array in case of returning early from _parseForm().
* don't use _postArgs member, but instead use a new local variable postArgs instead.
* same for _postArgsLen member vs.local postArgsLen.
* remove useless NULL pointer check before delete().

fix #9075


mcspr:
Should this be combined with #9075 and also remove the member itself?

PR #9077, opened 2024-02-02
ESP8266WebServer - free _currentArgs after request has been handled

everslick:
This can reduce permanent RAM consumption significantly, if the last request had lots of arguments, but it assumes, that we do not need the arguments after we have handled a request.


mcspr (approved):
Fair assumption, don't see any examples handling args after the request.

PR #9084, opened 2024-02-05
Generate TZ.h from zoneinfo

mcspr:
Using tzdata pypi package that ships zoneinfo blobs
Can't use raw data package from IANA, need it built first
Minor tweaks to [Espurna script](https://github.com/xoseperez/espurna/blob/dev/code/scripts/format_tzdata.py) that generated .md

More data than the original, not limited to ZONE/...
Zoneinfo source can be overwritten using path args, if needed
(but, probably not needed, since it would always be preferable to pull the latest version)

Also, allow to override certain names and tz strings independent
 of tzdata values (resolves #9007)


d-a-v:
Convenience proposal:

Ubuntu 22.04 has python 3.10 by default, [`datetime.UTC`](https://docs.python.org/3/library/datetime.html#datetime.UTC) is an alias introduced with python 3.11

```diff
--- a/tools/format_tzdata.py
+++ b/tools/format_tzdata.py
@@ -123,7 +123,7 @@ def markdown(zones):
 
 def header(zones):
     print(f"// ! ! ! DO NOT EDIT, AUTOMATICALLY GENERATED ! ! !")
-    print(f"// File created {datetime.datetime.now(tz=datetime.UTC)}")
+    print(f"// File created {datetime.datetime.now(tz=datetime.timezone.utc)}")
     print(f"// Based on IANA database {tzdata.IANA_VERSION}")
     print(f"// Re-run <esp8266 arduino core>/tools/{sys.argv[0]} to update")
     print()
```

d-a-v (changes requested):
Approving !

... after TZUpdate.sh is removed.

This change is much better for 
- aliases
- removing dependency from the other external repo

Issue #8780, opened 2022-12-31
DNS lookup problem


brunoudi:
### Basic Infos

- [ x ] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [ x ] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [ x ] I have tested that the issue is present in current master branch (aka latest git).
- [ x ] I have searched the issue tracker for a similar issue.
- [ x ] If there is a stack dump, I have decoded it.
- [ x ] I have filled out all fields below.

#### Platform

- Hardware:				[ESP-12]
- Core Version:			[3.0.2]
- Development Env:		[Arduino IDE]
- Operating System:		[Windows]

### Settings in IDE

- Module:  			[Generic ESP8266 Module]
- Flash Mode:			[DOUT]
- Flash Size: 		[4MB]
- lwip Variant:   [v2 Lower Memory]
- Reset Method:		[dtr]
- Flash Frequency:	[40Mhz]
- CPU Frequency:		[80Mhz]
- Upload Using:		[SERIAL]
- Upload Speed:   [115200]

### Problem Description

I have a separate router and modem. 
After temporary data loss (modem disconnected, internet reconnection, etc), without WiFi connection loss some ESP DNS problem is no longer able to resolve IP addresses.

I have enabled the DEBUG level to WIFI.

### Sketch

```cpp

#include <ESP8266WiFi.h>

const char* ssid     = "my-ssid";
const char* password = "my-password";

void setup() {
  Serial.begin(115200);

  Serial.println();
  Serial.println();
  Serial.print("Connecting to ");
  Serial.println(ssid);

  WiFi.mode(WIFI_STA);
  WiFi.begin(ssid, password);

  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
  }

  Serial.println("");
  Serial.println("WiFi connected");
  Serial.println("IP address: ");
  Serial.println(WiFi.localIP());
}

void loop() {
  IPAddress ipResult;
  int result = WiFi.hostByName("www.google.com", ipResult);
  Serial.print("Result: "); Serial.println(result);
  Serial.print("Ip: "); Serial.println(ipResult);
  Serial.println();
  delay(5000);
}

```
Turn on the router WITHOUT DATA (modem connection), and start the ESP.

The output, as expected, is this:

```
[...]

PR #9092, opened 2024-02-25
Added support to override host header.

nilo85:
As [dns is not working](https://github.com/esp8266/Arduino/issues/8780) on my network I decided to workaround the issue for now and figured this might actually be a feature we want here =)

### DIFF
--- a/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
+++ b/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp
@@ -119,6 +119,9 @@ bool HTTPClient::begin(WiFiClient &client, const String& host, uint16_t port, co
     return true;
 }
 
+void HTTPClient::setHostHeader(const String& host) {
+    _hostHeader = host;
+}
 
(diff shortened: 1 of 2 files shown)


nilo85:
I no longer need this hack, however, there might be situations you need to override host header (proxy use etc) so might still make sense to support

mcspr:
Perhaps the client should allow `addHeader` to override `Host:`?
Meaning, when headers are prepared it would check whether it was added or not; and if not, use the default `_host`. Right now it is restricted, don't see why it should be tbh

nilo85:
@mcspr I suppose, however to be completely http compliant, you can pass multiple values for same header, most other http implementations do setHeader and addHeader, where setHeader sets a single header and replaces whatever would have been there before, and addHeader would send multiple values.

However in this case, judging by the code I suspect the restriction here is to kep the code simple as the host header is hardcoded  in the request call. 

I would be fine rejecting this PR as I no longer need it, but possible it could make sense to consider adopting more http client features in the future, if needed... I suspect for most applications using these microchips, explicit proxy settings etc is overkill so I personally would consider it a corner case

PR #9104, opened 2024-03-17
Add missing update error codes

DRSDavidSoft:
This PR adds the two missing error states when beginning an Update. There were debugging logs for this but the error state was not set which would lead to confusion


mcspr (changes requested):
Missing getErrorString() update for new error codes

DRSDavidSoft:
Good catch, added

mcspr (approved):

PR #9083, opened 2024-02-05
CONT stack overflow postmortem

mcspr:
- check in cont_run() and cont_suspend() whether a1 points after cont_t::stack[0]
- in case a1 is broken, postmortem will still report proper context in
  proper stack boundaries. previously it used sys ctx values

additionally
- as suggested in https://github.com/esp8266/Arduino/issues/9069, change stack smashing to a single line that
  does not mention any Exceptions
- reduce overall stack dump length when there are know garbage values i.e. cont stackguard
- decoder.py addr search regexp should no longer skip stack lines ending with ' <'
- update decoder.py parsing so it notices both stack smashing and alloc errors

Noticed in esp-idf cfg as `CONFIG_FREERTOS_CHECK_STACKOVERFLOW_PTRVAL=y`, where Freertos it is a so called 'Method 1' of overflow checking happening right before ctx switch
https://www.freertos.org/Stacks-and-stack-overflow-checking.html
(also ref. #8666)

simple example like this one usually already offsets a1 when yield() happens, so the newly added check would cause a reset
```cpp
#include <Arduino.h>

void setup() {
  Serial.begin(115200);
}

void loop() {
  int buf[1048]; // oops
  yield();
  printf("%p %p %p %p\n",
    &buf[0], &buf[1], &buf[2], &buf[3]);
}
```

and it *should* generate more useful info as the result
```
Stack overflow detected
sp: 3fffef28 end: 3fffffd0 offset: 0000

ctx: cont
0x4010015c: ets_post at /home/runner/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/core_esp8266_main.cpp:243
0x402017f5: esp_suspend_within_cont at /home/runner/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/core_esp8266_main.cpp:121
(inlined by) __yield at /home/runner/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/core_esp8266_main.cpp:193
0x40201052: loop at /home/runner/dev/stack_overflow_test/src/main.cpp:10
0x402016da: loop_task(ETSEventTag*) at /home/runner/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/cores/esp8266/core_esp8266_main.cpp:279
0x40105245: call_user_start_local at ??:?
[...]

PR #9108, opened 2024-03-24
decoder.py: allowing to use it live

d-a-v:
Changes:

- avoid ctx repetition
- hide "DECODE IT"
- immediately print stack at closing marker
- check for executable decoder at startup

live command line example:
```
    pyserial-miniterm /dev/ttyUSB0 115200 | \
        path/to/esp8266/Arduino/tools/decoder.py \
            --tool addr2line \
            --toolchain-path path/to/esp8266/Arduino/tools/xtensa-lx106-elf/bin
            path/to/build.elf
```

### DIFF
--- a/tools/decoder.py
+++ b/tools/decoder.py
@@ -126,7 +126,7 @@ def decode_lines(format_addresses, elf, lines):
     def format_address(address):
         return "\n".join(format_addresses(elf, [address]))
 
-    lastepc = 0
+    lastaddr = {}
     for line in lines:
         # ctx could happen multiple times. for the 2nd one, reset list
         # ctx: bearssl *or* ctx: cont *or* ctx: sys *or* ctx: whatever
@@ -146,9 +146,9 @@ def decode_lines(format_addresses, elf, lines):
                 name, addr = pair.split("=")
                 if name in ["epc1", "excvaddr"]:
                     addr = addr.rstrip(',')
-                    if lastepc != addr:
+                    if not name in lastaddr or lastaddr[name] != addr:
                         if addr != '0x00000000':
-                            lastepc = addr
+                            lastaddr[name] = addr
                         output = format_address(addr)
                         if output:
                             print(f"{name}={output}")
--- a/tools/decoder.py
+++ b/tools/decoder.py
@@ -105,6 +105,7 @@ def decode_lines(format_addresses, elf, lines):
     STACK_LINE_RE = re.compile(r"^[0-9a-f]{8}:\s\s+")
 
     CUT_HERE_STRING = "CUT HERE FOR EXCEPTION DECODER"
+    DECODE_IT = "DECODE IT"
     EXCEPTION_STRING = "Exception ("
     EPC_STRING = "epc1="
 
@@ -125,6 +126,7 @@ def decode_lines(format_addresses, elf, lines):
     def format_address(address):
         return "\n".join(format_addresses(elf, [address]))
 
+    lastepc = 0
     for line in lines:
         # ctx could happen multiple times. for the 2nd one, reset list
         # ctx: bearssl *or* ctx: cont *or* ctx: sys *or* ctx: whatever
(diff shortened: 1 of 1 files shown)


tools/decoder.py
```diff
@@ -138,15 +144,20 @@ def format_address(address):
             stack_addresses.setdefault(last_stack, [])
             for addr in addrs:
                 stack_addresses[last_stack].append(addr)
+            stack_addresses = print_all_addresses(stack_addresses)
```
d-a-v, line 147: This line can be moved next to `in_stack = False` below.
I can't decide.

d-a-v:
#9083 should be merged first

tools/decoder.py
```diff
@@ -181,6 +189,9 @@ def select_tool(toolchain_path, tool, func):
     path = f"xtensa-lx106-elf-{tool}"
     if toolchain_path:
         path = os.path.join(toolchain_path, path)
+        if not os.path.isfile(path):
+            print(f"error: not existing tool (path -- tool): '{path}'", file=sys.stderr)
+            sys.exit(1)
```
mcspr, line 194: Already handled when trying to subprocess.Popen(path, ...)?
For both cases, with or without toolchain path.

tools/decoder.py
```diff
@@ -143,9 +145,13 @@ def format_address(address):
             for pair in pairs:
                 name, addr = pair.split("=")
                 if name in ["epc1", "excvaddr"]:
-                    output = format_address(addr)
-                    if output:
-                        print(f"{name}={output}")
+                    addr = addr.rstrip(',')
```
mcspr, line 148: Where does the comma show up here?

tools/decoder.py
```diff
@@ -125,6 +126,7 @@ def print_all_addresses(addresses):
     def format_address(address):
         return "\n".join(format_addresses(elf, [address]))
 
+    lastepc = 0
```
mcspr, line 129: Why no repeats?

tools/decoder.py
d-a-v: Agreed.
This one checks on script startup and helps user to build the command line up by not failing later when the first stack dump occurs.

tools/decoder.py
d-a-v: It aesthetically avoid this by removing the second one from the display.
```
Fatal exception 29(StoreProhibitedCause):
epc1=0x40206b4b: _printf_i at /workdir/repo/newlib/newlib/libc/stdio/nano-vfprintf_i.c:189 (discriminator 1)
Exception (29) - StoreProhibited: A store referenced a page mapped with an attribute that does not permit stores
epc1=0x40206b4b: _printf_i at /workdir/repo/newlib/newlib/libc/stdio/nano-vfprintf_i.c:189 (discriminator 1)     (<- repetition)
sp: 3ffffda0 end: 3fffffd0 offset: 0150
```

tools/decoder.py
d-a-v: With the below input, there is `epc1=0x40206b4b,` (<- comma)
and two lines below `epc1=0x40206b4b` (<- no comma)
Both are showing the same decoded info:
```
epc1=0x40206b4b: _printf_i at /workdir/repo/newlib/newlib/libc/stdio/nano-vfprintf_i.c:189 (discriminator 1)
```

input:
```
Fatal exception 29(StoreProhibitedCause):
epc1=0x40206b4b, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000000, depc=0x00000000

To make this dump useful, DECODE IT - https://tinyurl.com/8266dcdr

Exception (29):
epc1=0x40206b4b epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>
```

mcspr: Maybe ignore or silence it instead?
(same for below as it turns out)

Totally forgot this existed... quick search shows it around `user_uart_wait_tx_fifo_empty`
```
4027a381:       ffbf21                  l32r    a2, 4027a280
```

But not sure if it is really that func or part of some other stuff. Wonder if it can be turned off in favour of Core postmortem func

tools/decoder.py
mcspr: https://docs.python.org/3/library/shutil.html#shutil.which
e.g.
```python
if not shutil.which(path):
    raise FileNotFoundError(path)
```

tools/decoder.py
```diff
@@ -181,6 +189,8 @@ def select_tool(toolchain_path, tool, func):
     path = f"xtensa-lx106-elf-{tool}"
     if toolchain_path:
         path = os.path.join(toolchain_path, path)
+        if not shutil.which(path):
+            raise FileNotFoundError(path)
```
mcspr, line 193:
```suggestion
        
    if not shutil.which(path):
        raise FileNotFoundError(path)
```

PR #9109, opened 2024-03-25
Updater - fixed signature verification for compressed binaries

bakadave:
### Overview

This pull request introduces a fix to the signature verification, ensuring that the hash for compressed signed binaries is correctly calculated.

### Problem Description

Previously, Arduino Core attempted to read from flash memory without proper consideration for the 4-byte alignment requirement when calculating the hash for the signature verification. This did not present an issue when uncompressed binaries are checked as all compiled binaries are 4-aligned _(unconfirmed, just an educated guess)_, and signature verification appears to work well in these cases.
When uploading a compressed binary (based on [this](https://arduino-esp8266.readthedocs.io/en/latest/ota_updates/readme.html#compression)) the gzip algorithm makes no attempt to produce a 4-aligned file. The rest of the signing results in a valid signed binary regardless, however when calculating the hash for the verification process there is a ~75% chance that the hash will include some bytes from the signature, thus compromising the whole signature verification process.

### Solution

* Implementing a conditional check to identify unaligned read requests.
  -   For aligned reads, operations proceed as normal.
  -   For unaligned reads, the read size is adjusted to the nearest 4-byte boundary, ensuring alignment. Only the relevant bytes (excluding the padded ones for alignment) are processed after reading into a temporary buffer.
* These adjustments ensure ESP.flashRead operates within the requirements set by the ESP SDK.

### DIFF
--- a/cores/esp8266/Updater.cpp
+++ b/cores/esp8266/Updater.cpp
@@ -284,24 +284,8 @@ bool UpdaterClass::end(bool evenIfRemaining){
     _hash->begin();
     for (uint32_t offset = 0; offset < binSize; offset += sizeof(buff)) {
       auto len = std::min(sizeof(buff), binSize - offset);
-
-      if (len % 4 == 0) {
-            ESP.flashRead(_startAddress + offset, reinterpret_cast<uint32_t *>(&buff[0]), len);
-            _hash->add(buff, len);
-        }
-        else {
-            // Calculate padding needed to make len 4-byte aligned
-            uint32_t padLen = (len + 3) & ~3; // Rounds up to nearest multiple of 4
-            
-            // Temporary buffer to satisfy 4-byte alignment requirement
-            uint8_t tempBuff[padLen] = {0};
-            
-            // Read into the temporary buffer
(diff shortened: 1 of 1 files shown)


mcspr:
...or it can be swapped with u8 flashRead variant that does this?

The gist of the issue here is that flashRead cannot use unaligned flash size, which fails?

bakadave:
> The gist of the issue here is that flashRead cannot use unaligned flash size, which fails?

Yes.

> ...or it can be swapped with u8 flashRead variant that does this?

If there is one that would also work.

bakadave:
So far it seems like @mcspr's take on this is correct, simply dropping the `reinterpret cast` causes the u8 flashRead method to be used instead of the u32 and the whole problem goes away.

I'm somewhat inexperienced with pull requests on such a large project, what's the next step here?

d-a-v:
> what's the next step here?

Your last commit is enough.

bakadave:
just realised that the `alignas` specifier was redundant

bakadave:
@mcspr thank you for pointing it out

mcspr:
tnx!

PR #9114, opened 2024-03-29
WiFi STA encryptionType() getter added

JAndrassy:
Arduino WiFi API has this function.
in ESP8266WiFi library scan result has  encryptionType() but STA didn't have it.
I found this way to implement it.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
@@ -240,6 +240,7 @@ void ESP8266WiFiGenericClass::_eventCallback(void* arg)
 
     if (event->event == EVENT_STAMODE_AUTHMODE_CHANGE) {
         auto& src = event->event_info.auth_change;
+        WiFi.authMode = src.new_mode;
         if ((src.old_mode != AUTH_OPEN) && (src.new_mode == AUTH_OPEN)) {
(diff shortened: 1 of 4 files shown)


libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
```diff
@@ -240,6 +240,7 @@ void ESP8266WiFiGenericClass::_eventCallback(void* arg)
 
     if (event->event == EVENT_STAMODE_AUTHMODE_CHANGE) {
         auto& src = event->event_info.auth_change;
+        WiFi.authMode = src.new_mode;
```
mcspr, line 243: Is this is a side-effect of attempted authmode (i.e. wpa3 downgrade to wpa2) or event that always happen every connection?

libraries/ESP8266WiFi/src/ESP8266WiFiSTA.h
```diff
@@ -94,6 +95,9 @@ class ESP8266WiFiSTAClass: public LwipIntf {
         static void enableInsecureWEP (bool enable = true) { _useInsecureWEP = enable; }
 
     protected:
+        friend class ESP8266WiFiGenericClass;
+
+        uint8_t authMode;
```
mcspr, line 100: -1 default?
reset to -1 on disconnect?

libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp
JAndrassy: after every connection. 
src.old_mode is 0

PR #9116, opened 2024-03-31
WiFi wl_enc_type enum update

JAndrassy:
the `wl_enc_type` enum is from the first Arduino WiFi library and is part of the Arduino WiFi API. 
It is updated in newer WiFi libraries by Arduino.

### DIFF
--- a/cores/esp8266/wl_definitions.h
+++ b/cores/esp8266/wl_definitions.h
@@ -60,10 +60,14 @@ typedef enum {
 } wl_status_t;
 
 /* Encryption modes */
-enum wl_enc_type {  /* Values map to 802.11 encryption suites... */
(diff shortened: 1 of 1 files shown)


mcspr:
Not upstream everywhere, yet
https://github.com/arduino/ArduinoCore-renesas/blob/3dd08de860915a3577c232d3c0c2a66d33e04df7/libraries/WiFiS3/src/WiFiTypes.h#L20-L32 UNO R4 deployed these

Still pending for older cores
https://github.com/arduino-libraries/WiFi101/pull/350
https://github.com/arduino-libraries/WiFiNINA/pull/259

I would guess this should be ok here, actively saving from undefined enum errors and architecture ifdefs

JAndrassy:
in Arduino's Mbed Core my PR with this was already merged. https://github.com/arduino/ArduinoCore-mbed/pull/776

PR #9117, opened 2024-03-31
Mock - update func signatures for latest glibc

mcspr:
```
/usr/include/stdlib.h:102:15: error: declaration of ‘double atof(const char*) noexcept’ has a different exception specifier
...etc...
```
```
/usr/include/string.h:506:15: error: declaration of ‘size_t strlcpy(char*, const char*, size_t) noexcept’ has a different exception specifier
...etc...
```
```console
> rpm -q -f /usr/include/string.h /usr/include/stdlib.h
glibc-headers-x86-2.38-17.fc39.noarch
glibc-headers-x86-2.38-17.fc39.noarch
```

### DIFF
--- a/cores/esp8266/stdlib_noniso.h
+++ b/cores/esp8266/stdlib_noniso.h
@@ -22,38 +22,35 @@
 #ifndef STDLIB_NONISO_H
 #define STDLIB_NONISO_H
 
+#include <stdlib.h>
+
 #ifdef __cplusplus
-extern "C"{
+extern "C" {
 #endif
 
-int atoi(const char *s);
+#ifdef __cplusplus
+#define __STDLIB_NONISO_NOEXCEPT noexcept
+#else
+#define __STDLIB_NONISO_NOEXCEPT
+#endif
 
-long atol(const char* s);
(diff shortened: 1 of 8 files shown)


cores/esp8266/stdlib_noniso.h
```diff
@@ -46,8 +46,6 @@ char* ulltoa (unsigned long long val, char* str, int slen, unsigned int radix);
 
 char* dtostrf (double val, signed char width, unsigned char prec, char *s);
 
-void reverse(char* begin, char* end);
-
```
mcspr, line 50: only for internal use, does not need to be public

tests/host/common/mock.h
```diff
@@ -55,19 +55,23 @@
 #define D7 7
 #define D8 8
 
+#include <stdlib.h>
 #include <stddef.h>
 
+#ifndef HAVE_NONISO
+#include <stdlib_noniso.h>
```
mcspr, line 62: promised by TODO

overall libc stdlib comparison is tricky, don't see any feature flags with glibc :shrug:
matching attributes with latest version breaks old one and vice versa, so these have to be hidden unless asked for

Issue #9139, opened 2024-05-16
Exceptions with experimental::SPI0Command 


mhightower83:
### Basic Infos

- [X] This issue complies with the [issue POLICY doc](https://github.com/esp8266/Arduino/blob/master/POLICY.md).
- [X] I have read the documentation at [readthedocs](https://arduino-esp8266.readthedocs.io/en/latest) and the issue is not addressed there.
- [X] I have tested that the issue is present in current master branch (aka latest git).
- [X] I have searched the issue tracker for a similar issue.
- [X] If there is a stack dump, I have decoded it.
- [x] I have filled out all fields below.

#### Platform

- Hardware:					ESP-12
- Core Version:			[685f2c97ff4b3c76b437a43be86d1dfdf6cb33e3]
- Development Env:		Arduino IDE
- Operating System:		Ubuntu

### Settings in IDE

- Module:  			NodeMCU 1.0
- Flash Mode:			dio
- Flash Size: 		4MB
- lwip Variant:   v2 Higher Bandwidth
- Reset Method:		nodemcu
- Flash Frequency:	40Mhz
- CPU Frequency:		80Mhz
- Upload Using:		SERIAL
- Upload Speed:   921600 (serial upload only)

### Problem Description

`experimental::SPI0Command` causes an Exception on return to ICACHE address space. The issue was seen with flash non-volatile write operations like Write Status Register, Erase Sector, etc.

In `_SPICommand`, Adding a call to `Wait_SPI_Idlep` before restoring interrupts resolves the problem. I'll follow up with a PR.
```cpp
  PRECACHE_END();
  if (!spiIfNum) {
      // w/o a call to Wait_SPI_Idlep, 'Exception 0' or other exceptions (saw
      // 28) may occur later after returning to iCache code. This issue was
      // observed with non-volatile status register writes.
      //
      // My guess: Returning too soon to uncached iCache executable space. An
      // iCache read may not complete properly because the Flash or SPI
      // interface is still busy with the last write operation. In such a case,
      // I expect new ICACHE "code or literals" Reads result in zeros. This
      // would explain an Exception 0 for code, and Exception 20, 28, and 29
      // where a literal was misread as 0 and then used as a pointer.
      Wait_SPI_Idlep((SpiFlashChip *)fchip);
      xt_wsr_ps(saved_ps);
  }
  return (timeout>0 ? SPI_RESULT_OK : SPI_RESULT_TIMEOUT);
}
```

[...]

PR #9140, opened 2024-05-16
Fixes exceptions resulting from using SPI0Command

mhightower83:
Resolves exceptions occurring when using `experimental::SPI0Command` for flash write operations such as: Write Status Register-1, Sector Eraser, etc.
 * https://github.com/esp8266/Arduino/issues/9139#issue-2299471911

Moved PRECACHE_END to ensure `Wait_SPI_Idlep` and `xt_wsr_ps` are included in the iCache.

Added SPIUCSSETUP to give more settling time for #CS.

### DIFF
--- a/cores/esp8266/core_esp8266_flash_quirks.cpp
+++ b/cores/esp8266/core_esp8266_flash_quirks.cpp
@@ -66,12 +66,13 @@ void initFlashQuirks() {
             newSR3=SR3;
             if (get_flash_mhz()>26) { // >26Mhz?
                // Set the output drive to 100%
+               // These definitions are for the XM25QH32B part. On a XM25QH32C
+               // part, the XM25QH32B's 100% is C's 25% driver strength.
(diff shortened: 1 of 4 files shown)


mhightower83:
Let's put this on hold. I need to look closer at the sequence of calls used when doing write enabled by other examples. Some examples poll the status register until bit WEL is set before continuing.

mcspr:
> Let's put this on hold. I need to look closer at the sequence of calls used when doing write enabled by other examples. Some examples poll the status register until bit WEL is set before continuing.

Still on hold?

Is this related to a specific flash chip? Or does this happen at random (or not so forced random) on any hardware?

mhightower83:
It looks ready. I haven't found any more issues.
>  Is this related to a specific flash chip? Or does this happen at random (or not so forced random) on any hardware?

This may be me overanalyzing and being over-cautious. The BootROM's `SPI_write_enable()`  is consistently used to enable WEL. This appears to be the case in NONOS SDK and RTOS examples. And, that function likes to spin until the WEL bit is set. This kind of coding reminds me of some of the anomalies workarounds I have seen.

PR #9173, opened 2024-07-27
LwIpIntfDev.end() - check _started to prevent crash

JAndrassy:
LwIpIntfDev.end() crashed if begin() was not executed or failed


earlephilhower:
@JAndrassy do you want to do the RP2040 version of this patch, or would you rather I dup this myself?

JAndrassy:
> @JAndrassy do you want to do the RP2040 version of this patch, or would you rather I dup this myself?

If you can do it. I came across this problem as I develop something so I don't want to switch to RP2040 now.

PR #9189, opened 2024-08-05
Add typehints, fix Pathlib, fStrings

rowingdude:
To add readability and perhaps relative safety, I added typehints and formalized f-strings. 

I also got rid of the os.path in favor of Pathlib, whose use was inconsistent.

### DIFF
--- a/cores/esp8266/base64.cpp
+++ b/cores/esp8266/base64.cpp
@@ -26,25 +26,17 @@
 extern "C" {
 #include "libb64/cencode.h"
 }
-#include "base64.h"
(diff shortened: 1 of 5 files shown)


mcspr:
We can't deal with new syntax in *distributed scripts* i.e. the ones that get installed with the esp8266 package in IDE / PlatformIO, unless distribution is also updated to use py3.7. Windows version is pinned to it, macOS is 3.8 (I think?). Linux is the one without any requirements, we just assume recent distro is used.

And pls don't make every change in one PR :/

PR #9190, opened 2024-08-05
Update upload.py - Move code to functions and update operations

rowingdude:
- Migrate to Python 3.6+ syntax
- Add type hinting
- Use pathlib for file operations
- Implement argparse for CLI
- Refactor into smaller functions
- Improve input validation and error handling


mcspr:
Ok, but how does this work? Changing arguments from positional to `--`... would break when attempting an upload?

PR #9196, opened 2024-08-30
Fix #9194 - Allow more control over crash dump

### LINKED ISSUE #9194: Feature Request: Don't print stack trace to Serial on crash
### Basic Infos

- [N/A] If there is a stack dump, I have decoded it.

#### Platform

- Hardware:				ESP8266
- Core Version:			3.1.2 (latest)
- Development Env:		VS Code + Microsoft Arduino Extension
- Operating System:		Windows

### Settings in IDE

### Problem Description

Hi there, I'm trying to change the behavior of the `postmortem_report()` call not to print anything to the Serial upon a crash. The serial port in question is connected to a peripherical and is not used for debugging purposes.

So far, my efforts in #9193 have been unsuccessful. I am unable to override the `postmortem_report()` call via my code.

This got me thinking, can you possibly please introduce a new flag that would prevent the default function from being defined, or avoid printing to the serial? Currently the `custom_crash_callback()` callback is called _after_ the `postmortem_report()` has finished printing to the Serial port.

The addition of this would be awesome.

Thank you for the project!

DRSDavidSoft:
@mcspr This is a proof of concept PR for an alternative solution to #9193 and also closes #9194. Could you please verify? I would appreciate any input and feedback. Thanks!

### Description
This PR adds more granular control over how the crash dump is delivered, and some more features:
- `custom_panic_callback(panic_file, panic_line, panic_func, panic_what)`: Allows the user to additionally listen for panics (useful to record them in flash memory)
- `ets_uart_putc1()` → `crash_ets_uart_putc1()`: renamed function that prints to the Serial port. This allows the function to be overridden, possibly applying transformations on the text (useful if transmitting over a Bus-like UART, such as RS-485), remove clutter or insert additional comments (such as uptime), or completely inhibit the dump (useful is uart is connected to a peripheral device)
- `NO_CUT_HERE`: Introduce a macro that allows disabling the "CUT HERE" message and the divider lines. Can be used to minify the number of transmitted/recorded characters over the wire or storage. Can be used with the `-D` flag.
- `ets_println()` instead of `ets_putc('\n')`: This makes the new lines use the "cooked" interface instead of the "raw" one, which is also used on the static `ets_printf_P()` function, which allows:
    - uniform line endings (always "\r\n", instead of just the ones used within `printf_P(...)`
    - making the line endings also silent when the `crash_ets_uart_putc1()` is overriden
- `rst_info.exccause` used instead of local `exccause` for the `6` exception (divide by zero): unless I'm missing something, this allows the custom_crash_handler at the end of the postmortem function to also receive a correct exception number. Useful for projects such as EspCrashSave and EspCrashSaveSpiffs

### Additional notes
The naming of some functions might not be the optimal value, if you would like to suggest a better one please either modify this PR or submit a comment.

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -118,7 +118,7 @@ static inline void ets_println() { ets_printf_P(PSTR("\n")); }
 
 static void cut_here() {
 #ifdef NO_CUT_HERE
-	return;
+    return;
 #endif
 
     // https://tinyurl.com/8266dcdr => https://arduino-esp8266.readthedocs.io/en/latest/faq/a02-my-esp-crashes.html#exception
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -75,18 +75,33 @@ extern int umm_last_fail_alloc_line;
 
 static void raise_exception() __attribute__((noreturn));
 
-extern void __custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) {
+extern void __custom_crash_callback(struct rst_info * rst_info, uint32_t stack, uint32_t stack_end) {
     (void) rst_info;
     (void) stack;
     (void) stack_end;
 }
 
-extern void custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) __attribute__ ((weak, alias("__custom_crash_callback")));
+extern void __custom_panic_callback(const char* panic_file, int panic_line, const char* panic_func, const char* panic_what) {
+    (void) panic_file;
+    (void) panic_line;
+    (void) panic_func;
+    (void) panic_what;
+}
 
+extern void __ets_uart_putc1(char c) {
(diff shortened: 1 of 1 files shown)


mcspr:
> custom_panic_callback

Probably just one callback would be ok? We do have custom 'USER' reasons which are private now, but obviously can be moved to a public header. See `rst_reason_sw`. Suppose, this can be extended with something like `esp_panic_info()` returning a struct of those pointers?

> NO_CUT_HERE

Any reason not to fully hide the whole *internal* printf with a similar flag? Or, at least hide the part of postmortem & funcs calling postmortem that print stuff, structuring it a bit different in the main func especially.
(or simply replacing such internal printf with an empty macro on demand?)

> rst_info.exccause used instead of local exccause for the 6 exception (divide by zero): unless I'm missing something, this allows the custom_crash_handler at the end of the postmortem function to also receive a correct exception number. Useful for projects such as EspCrashSave and EspCrashSaveSpiffs

see https://github.com/esp8266/Arduino/pull/7496
This helps out exception decoder by pretending the hw does have instructions that would've triggered the exception. If only to be consistent with the output... I think the only issue is exc code would be different between rst info struct kept in memory between reboots and the one passed to the callback?

PR #9227, opened 2025-01-28
Tweaked `base64_encode_value` in `cores/esp8266/libb64/cencode.cpp`

LinRaymond2006:
Referred to the original libb64 implementation and optimized time complexity of `base64_encode_value` from O(n) to O(1). The existing method, which uses if-else statement, can slow down the execution speed.

### DIFF
--- a/cores/esp8266/libb64/cencode.cpp
+++ b/cores/esp8266/libb64/cencode.cpp
@@ -23,19 +23,12 @@ void base64_init_encodestate_nonewlines(base64_encodestate* state_in){
 }
 
 char base64_encode_value(const char n) {
-  char r;
+  unsigned char r;
 
-  if (n < 26)
-    r = n + 'A';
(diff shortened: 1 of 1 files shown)


mcspr:
So there are two places that encode b64 for our libs - HTTPClient and WebServer. Both client and server encode authorization header. Server also encodes file hash for etag. Which one is the problem here?

Pretty sure you've seen the history of the file. Original implementation was from libb64, but the lookup table placed in flash instead of ram (like it is here; since it is not marked with PROGMEM which forces it into .irom section, it never gets loaded into ram). https://github.com/esp8266/Arduino/pull/6607 traded flash space for some math. PR trades it back for ram. Any significant reason to make the change?

LinRaymond2006:
Oh I haven't notice that. Sorry for the confusion.

PR #9232, opened 2025-02-25
Add callback to allow Static IP per SSID in WiFiMulti

Sojourneer:
Callback before WiFi.begin allows user to configure selected SSID.

### DIFF
--- a/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
+++ b/libraries/ESP8266WiFi/src/ESP8266WiFiMulti.cpp
@@ -329,9 +329,10 @@ wl_status_t ESP8266WiFiMulti::connectWiFiMulti(uint32_t connectTimeoutMs)
             if (ssid == entry.ssid) {
                 DEBUG_WIFI_MULTI("[WIFIM] Connecting %s\n", ssid.c_str());
 
+                if(SSID_selected_callback) SSID_selected_callback(ssid.c_str());
                 // Connect to WiFi
                 WiFi.begin(ssid, entry.passphrase, channel, bssid);
-
+                
                 // Wait for status change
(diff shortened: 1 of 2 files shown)


mcspr:
Meaning, there are other use-cases besides config() call?

fwiw, default arguments passed at the end of the method can be omitted and .add() call stays the same. IPAddress has default constructor, ::isSet() would notify whether IPAddress was set by the user or not.

Sojourneer:
> Meaning, there are other use-cases besides config() call?
> 
In my use-case, I am setting the config parameters based on the MAC address.  Yes, that could be done upfront rather than when the SSID is decided.
But it seems cleaner, as well as more efficient, to have the config calculated when needed rather than preemptively.  I have a library that takes a JSON config (IP, as well as optionally MQTT etc.) and applies it in the callback to call the config per the found SSID.  Again, decoupling that from the code for the physical connection is beneficial.

mcspr:
Ok, agreed

Style suggestion then - use `onSsidSelected(...)` as one word without underscores. This is also the style used for other event-related things in the WiFi class.
Can you also update the example .ino (sketch) and documentation, or do you want me to do it?

Sojourneer:
Ok, I'll do it, and change the name as you recommend.  Do I need to redo the pull (I'm a noobie at this process)?

mcspr:
Just push more commits to Sojourneer:master

mcspr:
...if you are OK with my changes, that is. note the naming change connected -> selected, plus doc updates

mcspr:
> Sorry, noobie to this process; what do I need to do? I don't mind a rename.

See `Files changed 4` at the top, or `git pull` locally your own master. If everything is ok from your pov, I'll merge

PR #9238, opened 2025-03-25
update Mercury Board

iotgupta:
Updating Mercury Board terminology and pins.

Old boards (only 5 qty) not in production, not in market.

This updated version is final production version.


mcspr:
@iotgupta see my comment at #9254

PR #9239, opened 2025-04-02
Integrate String library (Arduino API)

jsnkan:


mcspr:
See https://github.com/arduino/ArduinoCore-API/
Should Core API be used, would make sense to c/p most of it. But, not the parts we locally override.
Plus, there is `api/deprecated/WString.h` linking back to `api/String.h`

Any specific reason, though?

PR #9271, opened 2025-08-27
Postmortem - allow more control over crash dump

### LINKED ISSUE #9194: Feature Request: Don't print stack trace to Serial on crash

mcspr:
ref. #9193 

resolves #9194
resolves #9196

@DRSDavidSoft

this got somewhat forgotten... my draft patch was just replacing the restart, but I figured it would be better to also replace/refactor internals to make things more consistent

- persistent & public REASON_USER_... for user callback rst_info reason
  possibly a minor breakage for anyone tracking reason==254 specifically, but very easily noticed
- public postmortem callback functions, don't overly hide the implementation
- put postmortem func one level down in the function chain, allow to
  override both the function itself or individual internals
  also separates rst\_info modifications and output, since the idea is to strip .o from the needless PSTR(...) strings when output is disabled (~1..2KiB)
- stack offset for -fexceptions, skip unwind data to the actual point of origin
  as top comment states, tested by throwing a dummy exception with just what(). this would stop at the `~Foo()` (hm) and the line throwing it

### DIFF
--- a/cores/esp8266/core_esp8266_postmortem.cpp
+++ b/cores/esp8266/core_esp8266_postmortem.cpp
@@ -517,9 +517,8 @@ void __unhandled_exception(const char *what) {
     raise_exception();
 }
 
-[[noreturn]]
-static inline void __panic_impl(const char *, const char *, int, const char *, const char *)
-    __attribute__((always_inline));
+static inline void __panic_impl(user_rst_reason, const char *, int, const char *, const char *)
+    __attribute__((noreturn, always_inline));
 static inline void __panic_impl(user_rst_reason reason, const char *file, int line, const char *func, const char *what) {
     s_user_reset_reason = reason;
     s_panic_file = file;


cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -99,31 +84,109 @@ static void ets_printf_P(const char *str, ...) {
...
+
+    if (c == '\n') {
+        USF(1) = '\r';
+    }
+    USF(1) = c;
+}
+
+void crash_system_prepare_rst_info(rst_info *rst_info, uint32_t user_reset_reason) noexcept {
+    auto *rtc_sys = reinterpret_cast<uint32_t *>(0x60001100ul);
+    memset(rst_info, 0, sizeof(*rst_info));
```
mcspr, line 118: slightly cheaper than `system_rtc_...` and no need to remember its logic

cores/esp8266/core_esp8266_postmortem.h
```diff
@@ -0,0 +1,108 @@
...
+
+// Following functions could be overridden, simply declare them in your own code with a custom functionality
+// Ensure that overrides are either declared with `extern "C"` *or* located in .c file(s)
+// Comments below describe the default function.
+
+#ifdef __cplusplus
+#define ESP8266_POSTMORTEM_NOEXCEPT noexcept
+#else
+#define ESP8266_POSTMORTEM_NOEXCEPT
+#endif
```
mcspr, line 51: General idea here is to restrict crash opportunities in the functions. Even when exceptions are on.

cores/esp8266/core_esp8266_postmortem.cpp
```diff
@@ -99,31 +84,109 @@ static void ets_printf_P(const char *str, ...) {
...
+void IRAM_ATTR uart1_write_char_d(char c) {
+    while (((USS(1) >> USTXC) & 0xff) >= 0x7e) { }
+
+    if (c == '\n') {
+        USF(1) = '\r';
+    }
+    USF(1) = c;
+}
+
+void crash_system_prepare_rst_info(rst_info *rst_info, uint32_t user_reset_reason) noexcept {
```
mcspr, line 116: Output is optional, so rst\_info has to be modified earlier

DRSDavidSoft:
> this got somewhat forgotten... my draft patch was just replacing the restart, but I figured it would be better to also replace/refactor internals to make things more consistent

Thank you, this is excellent, I've been waiting for something like this for ages! 🚀 

I have some comments I've been meaning to post, I've been doing some tests, would update later soon

